PR #199, opened 2016-05-24
Add Assert Data

sdtwigg:
Allow user to insert Bits into their assert message.
printf allows the use of format specifiers for inserting data from Bits
at hardware simulation time into the message. This just extends assert
to allow the same.


aswaterman:
what's up with the first 4 commits being in this PR?

sdtwigg:
I'm actually unsure given those were already merged....

sdtwigg:
Weird commits resolved. Had to trick github/git into properly believing the old commits were merged.

aswaterman:
I support merging this but can you make it apply cleanly against master?

ucbjrl:
Closed by #318 , #319

PR #200, opened 2016-05-24
Fixup to Chisel connections, literal creation, direction, etc.

sdtwigg:
First set of changes:
Require io be wrapped in IO, add Binding to track how/if Data is bound

* Adds internal.Binding, which is a marker on Elements (subtype/leaves of
Data), that tracks whether a Data has been placed onto the hardware graph
as a register, port, literal, wire, operation result, etc.
* This requires the io of a module actually be explicitly placed on the graph
AS it is being created. (Since the module will want to connect to its own IO,
module close is too late to do the binding.)
* The binding check will also enforce that a bundle constructor does not
actually create logic that is interpreted as part of the 'core bundle'.
* Subsequent CLs will expand upon the Binding system to carry more data
and execute more checks.
* Now, frontend emissions of undefined declarations should be classed as bugs
that should be caught by this system.

Second set of changes:
First, a listing of API changes:
For Literals,
-- You can no longer create a literal with UInt|SInt|Bool.apply(...)
++ Literal UInts/SInts are created from scala Int/BigInt by e.g. 0.asUInt,
myInt.asUInt(5). The latter means forced width to be 5.
++ Literal UInt can also be created from String bu asUInt
++ Literal Bools are created from scala Boolean by e.g. true.asBool
++ The UInt|SInt|Bool factory objects have Lit methods (which is what the old
literal apply methods were renamed to. These are primarily intended for
internal and transitional use. The .as* functions should be preferred.
++ The U|S|B methods added to Int|BigInt|String (identical to asUInt, asSInt, asBool) have been added and marked as deprecated.

For working with Chisel Data,
-- You can no longer supply a direction to UInt|SInt|Bool
-- You can no longer use .asInput|.asOutput|.flip. They have been marked as deprecated.
++ Instead, wrap in Input(), Output(), Flipped() call. These functions can only
be called on unbound data (avoiding extranneous copies)
++ In conjunction with literal changes, UInt(#), SInt(#) now always creates an
unbound variant of that width ie. equivalent to UInt(width=#), etc.

Particularly special notice on cloneType and newType,
-- cloneType is now protected and should not be called by the user
-- The user must still define cloneType (usually) for their cloned Bundles
however the contract has simiplified in that the clone merely must be
'structurally' similar to this.
++ Calling newType will invoke cloneType to get the structural copy and then
return that copy with the directions cleaned up for you.
++ This cleans up a somewhat nasty bug where, for example, direction changes
could get lost when a Vec of a directioned Data was created.
++ Reg and Wire now requires that outType/t variable be unbound (to avoid potential user errors). newType is still called on it so the same Type can be used to create multiple registers and wires.

On connection semantics and directionality,
++ Internal nodes like Reg and Wire that have no defined notion of input or
output direction now lose any direction information when going from Unbound to
Bound. It may be enforced later that the types be explicitly undirectioned.
++ <> is now back to being a bidirectional connect mean to operate between
Module IOs only
++ <> requires Bundles on both sides to have exactly the same fields
++ := is now back to being a monodirectional connect that assumes all
subelements of the RHS are sources for corresponding subelement sinks on the
LHS.
++ := requires the source Bundle to have at least the fields of the sink
Bundle.

In ChiselUtils,
++ EnqIO and DeqIO have been reworked and are no longer distinct classes and
just factory methods for creating an appropriate DecoupledIO
++ Decoupled factory object renamed to DecoupledIO
++ DecoupledIO fire() method renamed to firing
++ DecoupledIO has enq, noenq, deq, nodeq, similar to old EnqIO, DeqIO methods
++ ValidIO type renamed to Valid, no longer forces Output

Second, some notes on implementation changes.
Direction is no longer stored on the Data but on the Binding attached to the
data. The concept of Binding has been enriched to also introduce a 'Binder'
function that a caller users to describe how they want a Data's nodes to be
bound dependent on the specifics of the UnboundBinding. For example, PortBinder
will create a Binding with the same direction information of the
UnboundBinding.

The MonoConnect and BiConnect operators will only emit FIRRTL connections
for Elements, never Aggregates. This is to avoid the issues with FIRRTL
connection semantics. Relatedly, Bundles are never marked as flip or input.
Only Elements of Bundles (or Vec chains of Elements in Bundles) can be marked
as flip. Essentially, all flips have no been distributed through the Bundles to
the root nodes.

Vec types now contain a sample_element that tracks all the binding changes to
the actual elements. This sample_element is used when emitting the type of the
Vec to FIRRTL and when creating a dynamically indexed port. This is needed to
handle the case of an empty Vec having no elements but still requiring those
descriptions. For consistency, this sample_element is used even when the Vec is
not empty.

cloneTypeWidth behavior is somewhat weird. However, this is private and only
used to create operators and should not be an immediate issue. This will
eventually be removed (since cloneTypeWidth previously and currently will not
create appropriate Muxes of Aggregates).

The use of DynamicVariable in Builder has been more carefully hidden from the
rest of the internals by making dynamicContext private and adding accessors to
the other global state.

I want to watch the overhead in cloneType and newType to make sure too much
time isn't being spent making copies, manipulating them, copying again, as so
on as people built more complex types. (Not having Input, Output, Flipped
create a copy was inspired by a desire to avoid these copies)

### DIFF
--- a/src/main/scala/Chisel/Aggregate.scala
+++ b/src/main/scala/Chisel/Aggregate.scala
@@ -12,9 +12,9 @@ import internal.firrtl._
 /** An abstract class for data types that solely consist of (are an aggregate
   * of) other Data objects.
   */
-sealed abstract class Aggregate(dirArg: Direction) extends Data(dirArg) {
+sealed abstract class Aggregate extends Data {
   private[Chisel] def cloneTypeWidth(width: Width): this.type = cloneType
-  def width: Width = flatten.map(_.width).reduce(_ + _)
+  private[Chisel] def width: Width = flatten.map(_.width).reduce(_ + _)
 }
 
 object Vec {
@@ -22,10 +22,10 @@ object Vec {
     *
     * @note elements are NOT assigned by default and have no value
     */
-  def apply[T <: Data](n: Int, gen: T): Vec[T] = new Vec(gen.cloneType, n)
+  def apply[T <: Data](n: Int, gen: T): Vec[T] = new Vec(gen, n)
 
   @deprecated("Vec argument order should be size, t; this will be removed by the official release", "chisel3")
-  def apply[T <: Data](gen: T, n: Int): Vec[T] = new Vec(gen.cloneType, n)
+  def apply[T <: Data](gen: T, n: Int): Vec[T] = Vec(n, gen)
 
   /** Creates a new [[Vec]] composed of elements of the input Seq of [[Data]]
     * nodes.
@@ -94,29 +94,41 @@ object Vec {
   * @note Vecs, unlike classes in Scala's collection library, are propagated
   * intact to FIRRTL as a vector type, which may make debugging easier
   */
-sealed class Vec[T <: Data] private (gen: => T, val length: Int)
-    extends Aggregate(gen.dir) with VecLike[T] {
+sealed class Vec[T <: Data] private (gen: T, val length: Int)
+    extends Aggregate with VecLike[T] {
   // Note: the constructor takes a gen() function instead of a Seq to enforce
   // that all elements must be the same and because it makes FIRRTL generation
   // simpler.
+  private val self: Seq[T] = Vector.fill(length)(gen.newType)
 
-  private val self = IndexedSeq.fill(length)(gen)
+  /**
+  * sample_element 'tracks' all changes to the elements of self.
+  * For consistency, sample_element is always used for creating dynamically
+  * indexed ports and outputing the FIRRTL type.
+  *
+  * Needed specifically for the case when the Vec is length 0.
+  */
+  private[Chisel] val sample_element: T = gen.newType
 
-  override def <> (that: Data): Unit = this := that
+  // allElements current includes sample_element
+  // This is somewhat weird although I think the best course of action here is
+  // to deprecate allElements in favor of dispatched functions to Data or
+  // a pattern matched recursive descent
+  private[Chisel] final def allElements: Seq[Element] =
+    (sample_element +: self).flatMap(_.allElements)
 
   /** Strong bulk connect, assigning elements in this Vec from elements in a Seq.
     *
     * @note the length of this Vec must match the length of the input Seq
     */
-  def <> (that: Seq[T]): Unit = this := that
+  def <> (that: Seq[T]): Unit = {
+    require(this.length == that.length)
+    for ((a, b) <- this zip that)
+      a <> b
+  }
 
   // TODO: eliminate once assign(Seq) isn't ambiguous with assign(Data) since Vec extends Seq and Data
-  def <> (that: Vec[T]): Unit = this := that.asInstanceOf[Data]
-
-  override def := (that: Data): Unit = that match {
-    case _: Vec[_] => this connect that
-    case _ => this badConnect that
-  }
+  def <> (that: Vec[T]): Unit = this bulkConnect that.asInstanceOf[Data]
 
   /** Strong bulk connect, assigning elements in this Vec from elements in a Seq.
     *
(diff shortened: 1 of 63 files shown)


aswaterman:
looks like this PR got mangled with unrelated commits

sdtwigg:
Given the major directory changes and potential for future changes (and how long it takes to merge in those sorts of changes), I have not yet gotten around to reconciling that. Also, there are some commits in chisel3 that we refuse to merge (like changing the names of override_clock and override_reset to _clock and _reset).

I don't have time in the near future to clean up this PR so I have to leave it as-is for now. It at least serves as a proof of concept.

sdtwigg:
I fiddled with it some more and purged the unnecessary commits. There is still the other PR in there although that one, I think is non-controversial enough to not be an issue.

As stated before, I don't have time presently to bring this into conformance at the head of master. Since there are  a bunch of things here that should likely be split out into less of an omnibus commit, I am tempted to just leave this sit until there is time to do so.

aswaterman:
I think that makes sense.  In any case, we need to deprecate these features before outright removing them--people besides us are actually developing against chisel3...

kammoh:
Hi,
any plans on merging this to master?

ucbjrl:
branch [gsdt](/ucb-bar/chisel3/tree/gsdt) contains the connection and IO wrapping code from this PR and a compileOptions structure to selectively disable the "stricter" checks in this PR. As of 8/19/16, the gsdt branch builds rocket-chip and passes the asm and bmark tests after a single change to uncore/src/main/scala/devices/Prci.scala:

       val interrupts = new PRCIInterrupts {
         val mtip = Bool()
         val msip = Bool()
    -  }.asOutput
    +  }
    +  interrupts.asOutput

Without that change, compilation with the gsdt branch of chisel3 provokes the following error:

    [error] .../uncore/src/main/scala/devices/Prci.scala:27: value asOutput is not a member of uncore.devices.PRCIInterrupts{val mtip: chisel3.core.Bool; val msip: chisel3.core.Bool}
    [error] possible cause: maybe a semicolon is missing before `value asOutput'?
    [error]   }.asOutput
    [error]     ^
    [error] one error found
    [error] (uncore/compile:compileIncremental) Compilation failed

sdtwigg:
That is ... strange. PRCIInterrupts is a Bundle <: Data so it should have
implicitly converted to the variant that exposed the method (as .asOutput
presumably works in other contexts just fine, like on the very next line)

This is (slightly dangerously) abusing the fact that asOutput doesn't
replicate the input argument (which is subject to later change).

This change would be less objectionable (the best one would be knowing why
the implicit conversion to add the method isn't applying):
val interrupts = {
  val result = new PRCIInterrupts {
     val mtip = Bool()
     val msip = Bool()
  }
  result.asOutput
}

aswaterman:
Wait.  asOutput does replicate the input argument (or at least it did).
Changing back to the mutating version is a bad idea.

ucbjrl:
I suspect the failure of the implicit may be tangled with the block argument to the constructor - at some level the type isn't recognized as Bundle <: Data. I suspect this is a compiler bug.

ucbjrl:
asOutput() in the gsdt branch is (slightly) less mutating than the master branch version. Stephen's version was non-mutating, but the requirement to preserve the "firrtl" direction re-introduced the mutability found in the master branch.

sdtwigg:
The direction setters, currently as a very temporary measure, are not
calling newType on their arguments. This was to prevent the need for
certain spurious cloneType definitions (and also originally out a
performance fear that seems less problematic now). The plan was to
reintroduce the direction modifiers using newType once automatic cloneType
generator macro annotation was deployed. (That is half done. The case for
automatically generating cloneType for named things is done. There was a
bug in the scala typer that caused it to choke for anonymous bundles.)

This anticipated change was why I called the reliance 'dangerous'.

PS: I imagine it is a compiler bug in that the presence of a structural
type is making scalac typer reluctant to try to apply implicit conversions.

ucbjrl:
Thanks for the clarification @sdtwigg .
 
525e5688f69f474240fe5a21d6210beb04cfef29 temporarily restores direction override immutability pending the finished version.

ucbjrl:
Closed per #265. Remaining issues to be submitted as separate pull requests.

PR #201, opened 2016-05-24
Reg: include a convenient object to hold the value of a signal

terpstra:
A useful way of extending bus signals. I used this in the hasti rework.

### DIFF
--- a/src/main/scala/Chisel/BitPat.scala
+++ b/src/main/scala/Chisel/BitPat.scala
@@ -54,7 +54,7 @@ object BitPat {
     *
     * @note the BitPat must not have don't care bits (will error out otherwise)
     */
-  implicit def bitPatToUInt(x: BitPat): UInt = {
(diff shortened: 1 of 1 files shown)


sdtwigg:
The idea is fine. My main critique is the argument naming scheme. I acknowledge that Reg.scala already is inconsistent (Reg/RegNext vs RegEnable). But, I think ideally this would use the same scheme as RegEnable: so change 'in' to 'updateData'. Particularly since Reg and its derivatives advises the use of named parameters, it is more important to have consistency there.

(PS: The growing inconsistency will likely require some sort of unification across Reg, RegNext, RegEnable, etc. but that would have to come as a separate PR.)

terpstra:
I have no attachment to the parameter names. Feel free to change them. I
also wasn't sure if HoldUnless or HoldIf was more appropriate. I went with
HoldUnless because that uses an enable, like the other Reg objects.

PR #203, opened 2016-06-01
Move BitPat out of core

ducky64:
Also remove the very unsafe implicit conversions between BitPat <=> UInt. Add implicit conversions from Unt to BitPatConvertible, which allows BitPat operations on UInts using the pimp-my-library / extension methods pattern.

It seems that rocket-chip is failing to build, probably for unrelated reasons.


aswaterman:
Merged.

Issue #202, opened 2016-05-31
Data#toBits() deprecation says use asBits(), but data does not have asBits()


chick:
Chisel tutorial Adder has 

```
  val io = new Bundle {
    val Sum  = UInt(OUTPUT, n)
  }
  val sum   = Wire(Vec(n, Bool()))
  ...
  io.Sum := sum.toBits.toUInt()

```

chick:
Adding ```def asBits(): Bits``` abstract method, finds Vec, Bundle, and Clock as classes subclassing Data, but having no asBits() which toBits deprecation method tells them they should be using.

chick:
Deprecation message changed
```
  @deprecated("Best alternative, .asUInt()", "chisel3")
```

Issue #149, opened 2016-04-20
Tests use deprecated functions


sdtwigg:
A lot of the tests in chiselTests are using deprecated functions. It would be good (and a partial bonus to documentation) if this was fixed.

ducky64:
Also fixed.

PR #212, opened 2016-06-06
Tobits deprecation

chick:
Fixes last deprecation existing in chiselTests
Fixes issue #202 and #149


ducky64:
LGTM. Though you do have to make it merge-able again...
It would be nice if we could give a better description of asUInt, especially with regards to ordering in an Aggregate to UInt conversion.

Issue #217, opened 2016-06-09
access to Bundle elements


ucbjrl:
The testers (via IOAccessor) need access to Bundle elements as types (so the hwio testers can find the ready/valid signals which gate stepping through the input/output test vectors). As of commit 07fa5622ccc995f925d6d967d2a386540c9064cc namedElts is private[core]. The expedient fix is to make this private[chisel], but we should discuss what should be the "right" solution.

(This is a typical problem. Testers/debuggers do have valid reasons for reaching behind the public API.)

ducky64:
Testers / debuggers have reasons to reach beyond the _hardware construction_ API, but it should still use a _public_ API, reasoning being that someone who's not developing Chisel core may also want to write testers / debuggers, and the necessary functions should be exposed to them.

ducky64:
Also, it seems like hardware construction time reflection is currently used (I heard one Chisel project apparently queries IO pins for directionality), so exposing Bundle contents may be part of that. But possibly done in a more structured way than just exposing internal data types.

ucbjrl:
@donggyukim points out "elements" is part of the public API.

PR #218, opened 2016-06-09
Increase the scope of Bundle.namedElts to private[chisel] for testers.

ucbjrl:
This is a temporary fix to re-enable testers (IOAccessor). See #217


ucbjrl:
@donggyukim points out that the public API provides "elements" for this.

PR #221, opened 2016-06-15
Generate better node names when names collide

aswaterman:
Rather than using a global counter, memoize the last returned value for
colliding names to generate smaller sequence numbers.


jackkoenig:
lgtm

PR #224, opened 2016-06-22
renamechisel3 - "chisel" -> "chisel3"

ucbjrl:
This renames the package to chisel3 in order to co-exist with the Chisel compatibility object on case-insensitive file systems.
External packages should use the Chisel object in order to preserve backward compatibility with the existing Chisel2 code.


ducky64:
Looks good, but has a merge conflict...

PR #225, opened 2016-06-28
Fixed impl

da-steve101:
An implementation of Fixed for chisel3.
Just transforms over SInt so shouldn't affect FIRRTL


da-steve101:
As a side note, perhaps Bits should not be sealed? I can understand UInt/SInt/Bool but I think Bits would commonly want to be extended ( custom floating point implementations etc ). Fixed could even go in a separate chisel-utils library if this was the case. I think Fixed should be in chisel but it is an option worth considering.

da-steve101:
This is currently broken

da-steve101:
Seems to be working ok. Still not sure if fixed should be put into firrtl or just SInt based implementation as  here. Tend towards layer on SInt as it is essentially the same. Changed the fromBits method to be a connect instead of ":=". I think fromBits should force a connection.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -213,7 +213,7 @@ sealed abstract class Bits(dirArg: Direction, width: Width, override val litArg:
   /** Reinterpret cast to an UInt.
     *
     * @note value not guaranteed to be preserved: for example, a SInt of width
-    * 3 and value -1 (0b111) would become an UInt with value 7
```
aswaterman: "0b" here means "base 2," not a width.

aswaterman:
If it's possible to implement it w/o changing FIRRTL, it seems simpler not to change FIRRTL.

Relatedly, can we do this as a library on top of Chisel, instead of making it part of Chisel?

ucbjrl:
We're doing similar development for Chisel DSP support. Stay tuned.

da-steve101:
@aswaterman You can make it external to chisel, but would require changing a few things. Main one is Bits not being sealed (which I really think it shouldn't be). I would also be in favour of unsealing UInt/SInt/Bool but the argument isn't as strong.

chick:
@da-steve101 have you taken a look at the FixedPoint type now baked into firrtl and chisel3 and used extensively in dsptools.  I think it covers everything that is in here.

ducky64:
Superseded by FixedPoint

PR #226, opened 2016-06-28
Added RegNext and RegInit to chisel3

da-steve101:
Added a wrapper around Reg for the RegInit and RegNext chisel2 functions


aswaterman:
these already exist: https://github.com/ucb-bar/chisel3/blob/master/src/main/scala/chisel/util/Reg.scala

da-steve101:
Ah ok, woops not importing it. thanks

PR #228, opened 2016-06-29
Add a tree reduce function to Vec

da-steve101:
Added a few tree reduce operations.
I find I need this structure reduce more than reduceLeft/reduceRight and i think it makes sense to add it to Vec.

I also see a use case for a single layer reduce. For example, to implement some form of time multiplexing on a vec:

``` scala
val outputVec = inputVec.treeLayerReduce(
                         ( a : T, b : T ) => RegNext(Mux( myCond, a, b )),
                         ( a : T ) => RegNext( a ) )
```

Thoughts?

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -170,12 +170,20 @@ sealed class Vec[T <: Data] private (gen: => T, val length: Int)
   for ((elt, i) <- self zipWithIndex)
     elt.setRef(this, i)
 
-  /** Applys a single pass of a reduce operation in a parallel tree structure */
-  def treeLayerReduce( redOp : ( T, T ) => T, layerOp : ( T ) => T ) : Vec[T] = macro VecTransform.treeLayerReduce
+  /** Pairs the outputs together to transform something such as [ 1, 2, 3, 4, 5 ] to [ redOp(1,2), redOp(3,4), layerOp(5) ]
+    * @example The following example could be used in timeseries multiplexing
+    * {{{
+    * val outputVec = inputVec.pair(
+    *   ( a : T, b : T ) => RegNext(Mux( myCond, a, b )),
+    *   ( a : T ) => RegNext( a )
+    * )
+    * }}}
+    */
+  def pair( redOp : ( T, T ) => T, layerOp : ( T ) => T ) : Vec[T] = macro VecTransform.pair
 
-  def do_treeLayerReduce( redOp : ( T, T ) => T, layerOp : ( T ) => T )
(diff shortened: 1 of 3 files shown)


aswaterman:
I think the general notion is a good idea, but the single-layer vs. multi-layer distinction isn't consistent with scala reductions.  It's closer to fold, but not exactly the same as that, either.

I'd propose adding fold and reduce, which more closely match their counterparts in the scala collections.  (Also, I'd suggest merely documenting the binary tree nature of the reduction, rather than making it part of the name of the function.)

da-steve101:
Just as I was making the request I was thinking I didn't like the function names.
With scala having reduceLeft and reduceRight, perhaps treeReduce -> reduceTree?
Can't think of a good name for the single layer operation. Perhaps pairItems?

aswaterman:
I'm more inclined toward "reduce" over "reduceTree," but I don't feel that strongly about it.

The single-layer case just isn't consistent with the reduce pattern, though... maybe it should be split into two phases (e.g., map followed by reduce) or handled with fold (where an identity element is explicitly passed).

da-steve101:
Changed to reduce. I think pair is a bit better for the single layer.
I don't think it should be split up. It would have an intermediate state which would be messy to handle
```scala
Vec( UInt(1), UInt(2), UInt(3), UInt(4), UInt(5) ).grouped(2) => {
  IndexedSeq( ( UInt(1), UInt2 ), ( UInt(3), UInt(4) ), ( UInt(5) ) )
}
```

da-steve101:
I removed the single layer function. I think any name will be ambiguous and can easily be constructed in a clearer way using grouped(2).map( ... )

aswaterman:
I agree.

However, I still think it's quite confusing that reduce doesn't mirror that of the scala collections.  Its only argument is a function that combines (T, T) => T.  In the case that the input has just one element, it simply returns that element.  If you want to do something to each element prior to combining them, you can map over them first.

da-steve101:
In the case of a fully pipelined adder tree, I don't see how you could map the adding of 10 numbers as the second stage would need a register not there initially.
I think a default value of an identity function is better but unfortunately its not supported :/
See [SI-5920](https://issues.scala-lang.org/browse/SI-5920)
Replacing the scala reduce is the wrong approach in my opinion as the sequential behaviour could be expected. I think along the lines of reduceLeft and reduceRight, reduceTree is an appropriate name.
I have defined two macros for reduceTree, with and without the second argument. Its a little ugly, but I don't see a better way to do it.

aswaterman:
I actually favor having two implementations over the default argument in this case.

I'd like @sdtwigg to weigh in, but I'm in favor of merging.

jackkoenig:
ok to test

jackkoenig:
Replaced by #485

Issue #247, opened 2016-07-25
Vec.fill is call-by-name while Vec.apply is call-by-value?


seldridge:
The differences between the `fill` and `apply` methods of `Vec` have created some confusion for me. Specifically `fill` is call-by-name, while `apply` is call-by-value (https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala#L27, https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala#L92). One of the results of this is that `fill` can be used to instantiate a vector of modules, but `apply` cannot.

This conflicts with Chisel2 where both methods are call-by-name (https://github.com/ucb-bar/chisel/blob/master/src/main/scala/Vec.scala#L52, https://github.com/ucb-bar/chisel/blob/master/src/main/scala/Vec.scala#L86).

Is there a reason for this? This seems highly unintuitive to me as I generally view `fill` and `apply` (and `tabulate`, etc.) as all equivalent in terms of behavior, however, this introduces an odd disconnect.

I suspect this is the root cause of #148 and #242 which both use `Vec.apply`.

An overly verbose MWE:

``` scala
import chisel3._
import chisel3.util._

class PassThroughBundle extends Bundle {
  val x = Bool(INPUT)
  val y = Bool(OUTPUT)
}

class PassThrough extends Module {
  val io = new PassThroughBundle
  io.y := io.x
}

class VecModNoParamTest extends Module {
  val inside = Vec.fill(2)(Module(new PassThrough).io) // <- This works
  // val inside = Vec(2, Module(new PassThrough).io)   // <- This doesn't!
  val n = inside.length
  val io = new Bundle {
    val data = inside(0).cloneType
    val sel = UInt(width = log2Up(n))
  }

  (0 until n).map(i => inside(i).x := io.data.x)
  io.data.y := inside(io.sel).y
}
```

sdtwigg:
So, it is somewhat confusing because Vec contains two types of functions:
1. factory methods for Vecs: Vec.apply(Int, Data) . These just define hardware data structures, but do not actually create hardware directly. You just use these to build up Bundle descriptions, and pass as gen arguments to registers, wires, and other datatypes like Decoupled and Valid.
[...]

PR #248, opened 2016-07-25
Make Vec.apply gen parameter call-by-name

seldridge:
Possible fix for #247


sdtwigg:
The call-by-name isn't worth it because cloneType is called on the argument immediately. You don't get any benefits from using call-by-name.

seldridge:
Yes, you're right, this isn't even fixing the issue... I should have actually checked that. Bloody hell.

PR #251, opened 2016-08-08
Support Module name overrides

aswaterman:


aswaterman:
I didn't end up using getClass.getSimpleName, because that produces a less-informative name for anonymous classes.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -21,16 +21,25 @@ private[chisel3] class Namespace(parent: Option[Namespace], keywords: Set[String
     if (this contains tryName) rename(n) else tryName
   }
 
+  private def sanitize(s: String): String = {
+    // TODO what character set does FIRRTL truly support? using ANSI C for now
+    def legalStart(c: Char) = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_'
+    def legal(c: Char) = legalStart(c) || (c >= '0' && c <= '9')
+    val res = s.filter(legal _)
```
jackkoenig: Prefer: 
val res = s filter legal

src/main/scala/chisel3/internal/firrtl/Emitter.scala
```diff
@@ -82,7 +82,7 @@ private class Emitter(circuit: Circuit) {
     */
   private def emit(m: Component): String = {
     // Generate the body.
-    val moduleName = m.id.getClass.getName.split('.').last
+    val moduleName = m.id.desiredName
```
jackkoenig: Correct me if I'm wrong, but if name is the legalized form of desiredName, shouldn't that be what we use here?

aswaterman: Yeah, this is no more incorrect than it used to be, but it's not actually that simple.  Using the legalized name breaks deduplication.  I'll see if I can fix it.

sdtwigg: I need to do an internal cleanup first, but I will try to upstream our fix for this issue.

(Was okay with waiting when your PR was just going to go through but now there isn't much sense in having you continue to duplicate developing the same annoying edge case resolutions.)

aswaterman:
I've already corrected that specific issue.  When you get around to the cleanup, you can reconcile your version with what's now upstream... thanks for the help.

PR #255, opened 2016-08-12
Add support for per-Module compilation options.

ucbjrl:
Nothing uses these now, but when we integrate Stephen's PR200, we'll need a way to selectively enable some strict connection checks on a file by file basis. We plan to do this using package imports which will define suitable compilation options.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -97,12 +97,16 @@ private[chisel3] trait HasId {
   private[chisel3] def getRef: Arg = _ref.get
 }
 
-private[chisel3] class DynamicContext {
+private[chisel3] class DynamicContext(optionMap: Option[Map[String, String]] = None) {
   val idGen = new IdGen
   val globalNamespace = new Namespace(None, Set())
   val components = ArrayBuffer[Component]()
   var currentModule: Option[Module] = None
   val errors = new ErrorLog
+  val compileOptions = new CompileOptions(optionMap match {
+    case Some(map: Map[String, String]) => map
+    case None => Map[String, String]()
+  })
 }
 
 private[chisel3] object Builder {
@@ -115,6 +119,7 @@ private[chisel3] object Builder {
   def globalNamespace: Namespace = dynamicContext.globalNamespace
   def components: ArrayBuffer[Component] = dynamicContext.components
 
+  def compileOptions = dynamicContext.compileOptions
   def pushCommand[T <: Command](c: T): T = {
     dynamicContext.currentModule.foreach(_._commands += c)
     c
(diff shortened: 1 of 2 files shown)


aswaterman:
Wouldn't it be cleaner to pass the CompileOptions around instead of the Map?

ucbjrl:
The idea is that internally, clients refer to Builder.compileOptions.xxx. We could certainly generate this data structure at a higher level and pass it into Builder, but it may change on a Module by Module basis, so conceivably be reconstructed for each Module.

sdtwigg:
So, all of this analysis is going to run under the assumption that when a designer designs a Module, they do so under a certain assumption on the options. (e.g. If you design using ChiselOptionsA then you don't care about working/are unlikely to work under ChiselOptionsB)

Thus, I am not sure about this approach since it seems like people trying to mix Modules designed under different options sets are likely to have a bad time. I think your implicit-based approach you outlined in our e-mail conversation makes a lot more sense.

sdtwigg:
As an seemingly-unrelated-but-actually-related comment, some of the functions, in particular connections, will likely want to be added via extension methods (and thus importing) anyway. So, you might not need the ChiselOptions object yet.

This allows those connections to be typechecked based on the type of the LHS and RHS matching. This fairly cleanly solve the issue that Data can't actually define a typesafe connection operator (since any Data can't be arbitrarily connected to any Data). Instead you add the connection operators via implicit conversion (so the box can capture the more specific Data subtype) and only issue the connection when a Connectable[LHS, RHS] typeclass is resolvable.
See gama: https://github.com/sdtwigg/gama/blob/master/chiselfrontend/src/main/scala/frontend/implementation/ConnectTo.scala and (apparently I never actually checked in the changes that used implicit classes to add := and <>)
(Also, I think you can still have := and <> defined in Data and avoid any implicit conversion penalties by using the Self type member which is set by the concrete subtypes. However, this gets tricky if you want a concrete class to override a concrete class since overriding a type member is... odd.)

So, one would imagine the strict version not wanting an unrestricted := or <> defined in Data at all. Also, a loose version can just define an implicit Connectable for all of Data on all of Data that does the current string-based connections. (A caveat to this is that people can't reliably supply their own special snowflake connection behavior via overrides. e.g. in Chisel2, mySInt := myUInt would prepend a zero for you. It would work for direct connections but not when those types are in Bundles. However, the current behavior of deferring all aggregate connections to firrtl means you can't do this in chisel3 anyway so perhaps no one will notice the deficiency anyways!)

ucbjrl:
I think we have to ensure that CompileOptions do not change the semantics of the Module, otherwise we're building a Tower of Babel. CompileOptions is probably too general a name. We're really talking about error detection.

Modules designed under different options (error detection) had better be compatible.

ucbjrl:
Superseded by #265

PR #257, opened 2016-08-12
Make "def width" a private API; expose isWidthKnown instead

### LINKED ISSUE #256: Chisel does not like parameters in modules that are called width
This has been cropping up for a while now. Chisel gives really inscrutable error messages if you ever name a parameter to a module or bundle "width". It's easy to work around, but it does cause a lot of confusion when you encounter it. If you try to instantiate a module in Chisel that is declared as follows ...

    class AdjustableWidth(width: Int) extends Module {
      val io = new Bundle {
        val out = UInt(OUTPUT, width=width)
      }

      io.out := UInt(1)
    }

You get the error message

    java.lang.reflect.InvocationTargetException
    Caused by: java.lang.UnsupportedOperationException: empty.reduceLeft
        at scala.collection.TraversableOnce$class.reduceLeft(TraversableOnce.scala:178)
        at scala.collection.mutable.ArrayBuffer.scala$collection$IndexedSeqOptimized$$super$reduceLeft(ArrayBuffer.scala:48)
        at scala.collection.IndexedSeqOptimized$class.reduceLeft(IndexedSeqOptimized.scala:74)
        at scala.collection.mutable.ArrayBuffer.reduceLeft(ArrayBuffer.scala:48)
        at scala.collection.TraversableOnce$class.reduce(TraversableOnce.scala:206)
        at scala.collection.AbstractTraversable.reduce(Traversable.scala:104)
        at chisel3.core.Aggregate.width(Aggregate.scala:19)

However, just changing it by renaming the parameter makes it work just fine.

    class AdjustableWidth(w: Int) extends Module {
      val io = new Bundle {
        val out = UInt(OUTPUT, width=w)
      }

      io.out := UInt(1)
    }

Full example at https://github.com/zhemao/chiselbugs/blob/master/src/main/scala/WidthParameter.scala

aswaterman:
Resolves #256.

Since width was used to determine whether getWidth would succeed, I added

    def isWidthKnown: Boolean

but another option would be to expose something like

    def widthOption: Option[Int]

...thoughts?

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -373,5 +373,6 @@ class Bundle extends Aggregate(NO_DIR) {
 }
 
 private[core] object Bundle {
-  val keywords = List("flip", "asInput", "asOutput", "cloneType", "toBits")
+  val keywords = List("flip", "asInput", "asOutput", "cloneType", "toBits",
+                      "widthOption")
 }
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -16,7 +16,7 @@ import chisel3.internal.sourceinfo.{SourceInfo, DeprecatedSourceInfo, VecTransfo
   */
 sealed abstract class Aggregate(dirArg: Direction) extends Data(dirArg) {
   private[core] def cloneTypeWidth(width: Width): this.type = cloneType
-  def width: Width = flatten.map(_.width).reduce(_ + _)
+  private[core] def width: Width = flatten.map(_.width).reduce(_ + _)
 }
 
 object Vec {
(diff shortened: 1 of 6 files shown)


ucbjrl:
I vote for the first - isWidthKnown. It seems like the scenarios under which one would use this, would involve triggering additional code (optimizations) having nothing to do with the actual value itself. The second case (Option[Int]) seems more geared to providing a "default" value in the case the value is None. I suspect the latter usage will rarely arise in practice.

aswaterman:
For the second case, I was thinking that instead of writing code like

    if (x.isWidthKnown)
      doSomething(x.getWidth)
    else
      doSomethingElse

you'd write something like

    x.widthOption match {
      case Some(w) => doSomething(w)
      case None => doSomethingElse
    }

aswaterman:
so I think it's just a matter of taste

ucbjrl:
Hmmm. Or:

    x.widthOption match {
      case Some(32) => doSomethingWithTheOneTrueWidth(32)
      case Some(w) => doSomething(w)
      case None => doSomethingElse
    }

jackkoenig:
Porque no los dos?

I think isWidthKnown has a lot of utility but returning Option[Int] allows people to do more functional and "Scala-y" things like Andrew's example.

aswaterman:
good point

PR #259, opened 2016-08-16
Public APIs for chisel object names

donggyukim:
From discussion with @chick, @ucbjrl, and @azidar, this change provides chisel obejct names for FIRRTL annotation and chisel testers. Chisel's internal `HasId` now extends `Signald` having four public methods:
* `signalName`: returns the CHIRRTL name of the signal
* `pathName`: returns the full path name of the signal from the top module
* `parentPathName`: returns the full path of the signal's parent module instance from the top module
* `parentModName`: returns the signal's parent **module(not instance)** name. This is not final until CHIRRTL is actually emitted because duplication checks are done in this stage.

To annotate chisel objects and pass them to FIRRTL, users may have containers of `SignalId` and create annotation for signals with `signalName` and `parentModName` after CHIRRTL is emitted by Chisel Driver.  Also, path names are useful for various cases including chisel testers.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -63,6 +63,9 @@ extends HasId {
   /** Legalized name of this module. */
   final val name = Builder.globalNamespace.name(desiredName)
 
+  /** Signal name (for simulation). */
+  override def signalName(component: Component) = name
+
   /** IO for this Module. At the Scala level (pre-FIRRTL transformations),
     * connections in and out of a Module may only go through `io` elements.
     */
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -78,7 +78,7 @@ extends HasId {
 
 
   /** Signal name (for simulation). */
-  override def signalName =
+  override def instanceName =
     if (_parent == None) name else _component match {
       case None => getRef.name
       case Some(c) => getRef fullName c
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -31,7 +31,9 @@ object Module {
     m._commands.prepend(DefInvalid(childSourceInfo, m.io.ref)) // init module outputs
     dynamicContext.currentModule = parent
     val ports = m.computePorts
-    Builder.components += Component(m, m.name, ports, m._commands)
+    val component = Component(m, m.name, ports, m._commands)
+    m._component = Some(component)
+    Builder.components += component
     pushCommand(DefInstance(sourceInfo, m, ports))
     m.setupInParent(childSourceInfo)
   }
@@ -63,8 +65,24 @@ extends HasId {
   /** Legalized name of this module. */
   final val name = Builder.globalNamespace.name(desiredName)
 
+  /** FIRRTL Module name */
+  private var _modName: Option[String] = None
+  private[chisel3] def setModName(name: String) = _modName = Some(name)
+  def modName = _modName match {
+    case Some(name) => name
+    case None => throwException("modName should be called after circuit elaboration")
+  }
+
+  /** Keep component for signal names */
+  private[chisel3] var _component: Option[Component] = None
+
+
   /** Signal name (for simulation). */
-  override def signalName(component: Component) = name
+  override def signalName =
+    if (_parent == None) name else _component match {
+      case None => getRef.name
+      case Some(c) => getRef fullName c
+    }
 
   /** IO for this Module. At the Scala level (pre-FIRRTL transformations),
     * connections in and out of a Module may only go through `io` elements.
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -55,10 +55,29 @@ private[chisel3] class IdGen {
   }
 }
 
-private[chisel3] trait HasId {
+/** Public API to Nodes.
+  * currently, the node's name, the full path name, and a reference to its parent.
+  */
+trait SignalID {
+  def signalName(component: Component): String
+  def signalPathName(component: Component, separator: String = "_"): String
+  def signalParent: Module
+}
+
+private[chisel3] trait HasId extends SignalID {
   private[chisel3] def _onModuleClose {} // scalastyle:ignore method.name
   private[chisel3] val _parent = Builder.dynamicContext.currentModule
   _parent.foreach(_.addId(this))
+
+  // Implementation of public methods.
+  override def signalParent = _parent.get
+  override def signalName(component: Component) = _ref.get.fullName(component)
+  override def signalPathName(component: Component, separator: String = "_"): String = {
+    _parent match {
+      case Some(p) => p.signalPathName(component, separator) + separator + signalName(component)
+      case None => signalName(component)
+    }
+  }
 
   private[chisel3] val _id = Builder.idGen.next
   override def hashCode: Int = _id.toInt
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -59,28 +59,17 @@ private[chisel3] class IdGen {
   * currently, the node's name, the full path name, and references to its parent Module and component.
   * These are only valid once the design has been elaborated, and should not be used during its construction.
   */
-trait SignalID {
-  def signalName(component: Component): String
-  def signalPathName(component: Component, separator: String = "."): String
-  def signalParent: Module
-  def signalComponent: Option[Component]
+trait SignalId {
+  def signalName: String
+  def pathName: String
+  def parentPathName: String
+  def parentModName: String
 }
 
-private[chisel3] trait HasId extends SignalID {
+private[chisel3] trait HasId extends SignalId {
   private[chisel3] def _onModuleClose {} // scalastyle:ignore method.name
   private[chisel3] val _parent = Builder.dynamicContext.currentModule
   _parent.foreach(_.addId(this))
-
-  // Implementation of public methods.
-  override def signalParent = _parent.get
-  override def signalName(component: Component) = _ref.get.fullName(component)
-  override def signalPathName(component: Component, separator: String = "_"): String = {
-    _parent match {
-      case Some(p) => p.signalPathName(component, separator) + separator + signalName(component)
-      case None => signalName(component)
-    }
-  }
-  override def signalComponent: Option[Component] = None
 
   private[chisel3] val _id = Builder.idGen.next
   override def hashCode: Int = _id.toInt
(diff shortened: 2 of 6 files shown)


chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -95,6 +106,33 @@ private[chisel3] trait HasId {
...
+    def constructName(arg: Arg): String = arg match {
+      case Node(id) => id.signalName
+      case ILit(n) => n.toString
+      case ULit(n, w) => n.toString
+      case SLit(n, w) => n.toString
+      case Ref(name) => name
+      case Slot(imm, name) => s"${imm.id.signalName}.$name"
+      case Index(imm, value) => s"${constructName(imm)}[${constructName(value)}]"
+      case ModuleIO(mod, name) => name
+    }
```
aswaterman: Any reason that constructName is a match, rather than a virtual function on Arg that each of these classes must implement?  This seems extremely similar to the Arg.fullName, so they should either be merged, or at least defined in the same place.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -55,7 +55,18 @@ private[chisel3] class IdGen {
...
 
-private[chisel3] trait HasId {
+/** Public API to access Node/Signal names.
+  * currently, the node's name, the full path name, and references to its parent Module and component.
+  * These are only valid once the design has been elaborated, and should not be used during its construction.
+  */
+trait SignalId {
+  def signalName: String
+  def pathName(seperator: String = "."): String
+  def parentPathName(seperator: String = "."): String
```
aswaterman: Is separator ever changed at runtime?  Can SignalId just have a member like

    private def separator = "."

instead of the default argument to pathName/parentPathName?

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
donggyukim: That `fullName` requires a `Component` as its argument, so I didn't want to handle it in `Module` or somewhere else. Probably I can use it instead here.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
donggyukim: Yes, it can be changed, but I can make the consumer post-process it.

ucbjrl:
retest this please

src/test/scala/chiselTests/AnnotatingExample.scala
```diff
@@ -0,0 +1,157 @@
...
+  val annotate = MyBuilder.myDynamicContext.annotationMap
+
+  annotate(AnnotationKey(subModule2, AllRefs))     = s"SomeSubMod was used"
+
+  annotate(AnnotationKey(x, JustThisRef)) = "I am register X"
+  annotate(AnnotationKey(io.a, JustThisRef)) = "I am io.a"
+  annotate(AnnotationKey(io.bun.nested_1, AllRefs)) = "I am io.bun.nested_1"
+  annotate(AnnotationKey(io.bun.nested_2, JustThisRef)) = "I am io.bun.nested_2"
+
+  when (x > y)   { x := x -% y }
```
ducky64: Is this logic relevant to the test case? If not (and it doesn't seem this has relevance to naming / annotations), consider simplifying the test case so things being tested are orthogonal (separate annotation test failures from logic test failures).

src/test/scala/chiselTests/AnnotatingExample.scala
```diff
@@ -0,0 +1,157 @@
...
+  def build[T <: Module](f: => T): Map[String, String] = {
+    myDynamicContextVar.withValue(Some(new MyDynamicContext)) {
+      Driver.emit(() => f)
+      processAnnotations(myDynamicContextVar.value.get.annotationMap)
+    }
+  }
+}
+
+object MyDriver extends BackendCompilationUtilities {
+  def doStuff[T <: Module](gen: () => T): Map[String, String] = MyBuilder.build(Module(gen()))
```
ducky64: 'doStuff' as a name is basically content-free. It doesn't give the reader any idea of what it does, any function could be named 'doStuff' and it would still be accurate.

src/test/scala/chiselTests/AnnotatingExample.scala
```diff
@@ -0,0 +1,157 @@
...
+      k match {
+        case (AnnotationKey(signal, JustThisRef)) =>
+          f"Just this ref $k%60s -> $v%40s  component $signal"
+        case (AnnotationKey(signal, AllRefs)) =>
+          f"All refs      $k%60s -> $v%40s  component $signal"
+        case  _ =>
+          s"Unknown annotation key $k"
+      }
+    }.toList.sorted.mkString("\n")
+    println(s"Sorted list of annotations\n$list")
```
ducky64: Shouldn't the style checker complain about uses of println?

In general, use of println during a test should be avoided, because after you're finished debugging it, no one else is going to check the console output. Everything that needs to be checked should be encoded in the test.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -95,6 +106,27 @@ private[chisel3] trait HasId {
   private[chisel3] def setRef(parent: HasId, index: Int): Unit = setRef(Index(Node(parent), ILit(index)))
   private[chisel3] def setRef(parent: HasId, index: UInt): Unit = setRef(Index(Node(parent), index.ref))
   private[chisel3] def getRef: Arg = _ref.get
+
+  // Implementation of public methods.
+  def signalName = _parent match {
+    case None => getRef.name
```
ducky64: Is there ever a case where a signal wouldn't have a parent (other than because of bugs)? Should it assert out rather than trying to make something up?

I also assume that this is trying to be consistent with how CHIRRTL names are generated. If so, it would be nice to just reference the CHIRRTL generated name, as opposed to possibly duplicating logic here (like concatenating path and signal name in pathName implementation).

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -63,6 +65,19 @@ extends HasId {
   /** Legalized name of this module. */
   final val name = Builder.globalNamespace.name(desiredName)
 
+  /** FIRRTL Module name */
+  private[chisel3] var _modName = name
```
ducky64: Is introducing mutable state the best way to go here? While it's convenient from an implementation standpoint, it also makes program correctness harder to reason about ("when is this legal?" - especially since the value can change, even if it looks legitimate from the start).

From a higher level system architecture standpoint, would it make sense to query FIRRTL names from some kind of map rather than placing it directly in the module? Like firrtlNameMap.getName(myModule)?

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
aswaterman: myTopLevelModule.signalName is the only case this should happen, but presumably that should be a legal thing to ask.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
aswaterman: The main reason this is problematic is that it's initialized to name, rather than being an Option[String] initialized to None.  With that change, I don't see a clear advantage or disadvantage to firrtlNameMap vs. the var.  (This is kind of like the old debate about the refMap.)

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
donggyukim: On Fri, Aug 19, 2016 at 12:55 PM, Richard Lin <notifications@github.com>
wrote:

> In chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
> <https://github.com/ucb-bar/chisel3/pull/259#discussion_r75542313>:
>
> > @@ -95,6 +106,27 @@ private[chisel3] trait HasId {
> >    private[chisel3] def setRef(parent: HasId, index: Int): Unit = setRef(Index(Node(parent), ILit(index)))
> >    private[chisel3] def setRef(parent: HasId, index: UInt): Unit = setRef(Index(Node(parent), index.ref))
> >    private[chisel3] def getRef: Arg = _ref.get
> > +
> > +  // Implementation of public methods.
> > +  def signalName = _parent match {
> > +    case None => getRef.name
>
> Is there ever a case where a signal wouldn't have a parent (other than
> because of bugs)? Should it assert out rather than trying to make something
> up?
>
> Yes, It shoudn't happen, so now exceptions are thrown in this case.

> I also assume that this is trying to be consistent with how CHIRRTL names
> are generated. If so, it would be nice to just reference the CHIRRTL
> generated name, as opposed to possibly duplicating logic here (like
> concatenating path and signal name in pathName implementation).
>
 As far as I know, there are no counter part names in CHIRRTL/FIRRTL for
pathName.

sdtwigg:
One concern I have with the general annotation strategy (which pathName and parentPathName in particular expose as a concern) is a superposition problem. This is related but more complex than simple aliasing. It is related (but not solely caused by) to the fact that annotations make it so that you can't rely on just the graph description to see if two modules are equal (even though many portions of chisel assume so.)

Annotations of the form parentModName.signalName are fine and simplest to handle directly. For a given named instance of something (wire, reg, module, etc.), it is intuitive to know to look at that entry in the map for that thing. The issue demonstrates for handling annotations of the form (parent.parentModName).(parent.signalName).signalName (and so on for parent.parent until you reach some ModuleDef). pathName is the special case of needing to reach the 'top' ModuleDef. Since the same ModuleDef can be used to describe multiple instances of a module, the annotations of specific nodes can be thought of as being 'in superposition.' 

```
e.g. Circuit Top has 4 modules: Top, A, B, C that instantiate like a diamond
Both A and B instantiate a C and Top instantiates both A and B
Annotations refer to some range analysis:
  C.in says in is known to be <16
  Top.instA.instC.in says in is known to be >= 8
  Top.instB.instC.in says in is known to be < 8
```

Passes really need to know to 'decompress' the module definition listing into the actual module instances and analyze those separate instances with the proper annotations. Note that some annotations may apply to all, some, or only one instance depending on where the annotation is rooted. A parentModName.signalName always applies to all. A pathName applies to one unless the top has been modified). And, some other nested reference rooted in non-Top may refer to say half the instances.

I particularly label this as a concern because FIRRTL already has bugs where it fails to handle deduplicated modules properly in the width inferer. Going back to the above example, if C has an input with unknown width, width constraints from its use in both A and B are merged. (This actually leads to frightening results when that input then drives an output of unknown width. Now, one of those module instances is going to experience a silent/implicit truncation that is not obvious from local inspection.)

Actually, the symmetry between the problems can be further illustrated by doing width inference in two passes that transfer information via the annotations. The first pass analyzes the circuit and adds annotations to each node representing constraints. You expect most the annotations to be on/refer to parentModName.signalName.[signalPart] (for internal nodes) or (parent.parentModName).(parent.signalName).[IOpart] (for a submodule IO). Then, the second pass reads all those and resolves them. The bug is that it is 'flattening' all the latter references to just parentModName.signalName, even when the parent.parentModName are different.

Even if you legislate that all annotations must be rooted at the top definition (or add an analysis that replicates annotations such that they all are rooted at the top), this only solves an aliasing problem. You CAN further solve the superposition problem by replicate and renaming module instances until no annotated thing is in superposition. However, this could result in a lot of repeated work (and to some degree means Chisel shouldn't have bothered deduplicating in the first place). Also, since any pass could introduce new annotations (that could cause a superposition issue) or adjust module instantiations, you would need to run this analysis after every pass.

sdtwigg:
In writing up examples in my prior comment I noted an oddity in naming:

Given that Module <: HasId and therefore SignalId, Module has a use for signalName (it is in fact the name of it as the instance). However, it is somewhat odd for a Module to have a signalName. Instead I propose:
signalName -> instanceName
SignalId -> InstanceId.

It makes sense for wires and regs to have an instanceName as they are merely 'instances' of hardware registers or wires (like how in general software values could be considered instances of some type).

PR #267, opened 2016-08-25
Make sure RRArbiter's lastGrant has an initial value

zhemao:
The round robin arbiter uses a register to determine which of its inputs gets priority. It's set based on which one was last accepted. The initial value does not really matter, but we shouldn't rely on random initialization to set it in simulation, since that could be turned off.

### DIFF
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -55,7 +55,7 @@ abstract class LockingArbiterLike[T <: Data](gen: T, n: Int, count: Int, needsLo
 
 class LockingRRArbiter[T <: Data](gen: T, n: Int, count: Int, needsLock: Option[T => Bool] = None)
     extends LockingArbiterLike[T](gen, n, count, needsLock) {
-  lazy val lastGrant = RegEnable(io.chosen, io.out.fire())
+  lazy val lastGrant = RegEnable(io.chosen, UInt(0), io.out.fire())
   lazy val grantMask = (0 until n).map(UInt(_) > lastGrant)
   lazy val validMask = io.in zip grantMask map { case (in, g) => in.valid && g }
 


aswaterman:
Since the initial value doesn't matter, why spend the hardware to reset it?

zhemao:
It's in the standard library and FIRRTL gives the option to not initialize the register value. If you used that option in simulation, there would be an X and the arbiter wouldn't function correctly. Maybe then the solution is to simply disallow turning off randomization in simulation.

aswaterman:
That will be a problem with any register that is not reset; this is just a band-aid (and one that adds HW cost).

If you turn off randomization in FIRRTL, you need to tell your simulator not to X-initialize.

aayupov:
We do also have a problem with the arbiter selector not being initialized. Although, you provided an insight why it is intentionally not initialized, it clearly can break some downstream simulation flows when you cannot tell simulator to not X-initialize. We also would like to have a reproducible behavior in simulation. So for those, who may need it, can we have a parameter to the arbiter to initialize it?

aswaterman:
@aayupov The problem here is that lots of designs in Chisel, including other utilities in the standard library, deliberately do not initialize registers whose initial values don't actually matter.  It's inconsistent to add a flag to fix this for only one module.  (Another example is the Queue datapath, where correct designs can fall victim to X-pessimism.  However, few people would be willing to pay for the entire Queue datapath to be reset, just to please their simulation environments.)

Perhaps this should be treated as a backend issue, rather than as a Chisel issue.  It might make sense to have an optional Firrtl pass to reset all flops.

pentin-as:
Maybe, make it possible to set the initial value of lastGrant in RRArbiter, and then set it to 0 in PTW?

poemonsense:
Should we re-consider this PR to avoid X propogation issues? I think the `lastGrant` here is one of the most popular X-prop sources in Chisel designs that are used across a wide range of projects. It does not take much to initialize this register, given that it's not wide registers in the datapath.

Previously the PLRU state in rocket-chip is not initialized as well and brings X-prop issues. However, by the end of 2020, they have added reset values to the registers. So do these arbiters.

seldridge:
> Should we re-consider this PR to avoid X propogation issues? I think the lastGrant here is one of the most popular X-prop sources in Chisel designs that are used across a wide range of projects. It does not take much to initialize this register, given that it's not wide registers in the datapath.

The main problem is that not initializing `lastGrant` is a special case of a valid Chisel or Verilog design pattern. We can patch it for `lastGrant`, but it doesn't fix the underlying problem of what to do about Verilog X-pessimism (other than clearly indicating that randomization or VCS xprop may be required and that the user needs to understand how to setup their simulation environment to avoid x-pessimism).

That said, Chisel likely needs better APIs for describing _intentionally_ uninitialized registers and intentional, legal reads of these registers. Compilers can then provide better static or dynamic warnings/errors when bad things happen or add assertions to dynamically catch issues. Some type of uninitialzed sanitizer, relying on instrumentation, could also be built.

> Perhaps this should be treated as a backend issue, rather than as a Chisel issue. It might make sense to have an optional Firrtl pass to reset all flops.

This is actually doable today if a user is willing to switch to the MLIR-based FIRRTL Compiler. That supports a documented, but not publicized [`FullAsyncResetAnnotation`](https://circt.llvm.org/docs/Dialects/FIRRTL/FIRRTLAnnotations/#fullasyncresetannotation). This can be used to indicate that a specific port is an async reset and to use it as the async reset for all registers under that point in the hierarchy. (This annotation may change in the future! I anticipate some variant of this API always existing, though.)

E.g.:

```
circuit Foo:
  module Foo:
    input clock: Clock
    input reset: AsyncReset
    input d: UInt<1>
    output q: UInt<1>

    reg r: UInt<1>, clock
    r <= d
    q <= r
```
```json
[
 {
   "class":"sifive.enterprise.firrtl.FullAsyncResetAnnotation",
   "target":"~Foo|Foo>reset"
  }
]
```
Becomes:
```verilog
module Foo(
  input  clock,
         reset,
         d,
  output q);

  reg r;
  always @(posedge clock or posedge reset) begin
    if (reset)
      r <= 1'h0;
    else
      r <= d;
  assign q = r;
endmodule
```

poemonsense:
> That said, Chisel likely needs better APIs for describing intentionally uninitialized registers and intentional, legal reads of these registers. 

I agree. Uninitialized registers is a fundamental issue of simulators and we have to come up with some intelligent methods to address this issue. If Chisel allows the users to specify registers that are intentionally left uninitialized, it would be much better. For example, if the user can annotate some registers and let the Chisel compiler to generate a list, which would be further used for deposit/force in the Verilog simulator, the issue may be solved.

Why we really want this: Verilog is the signoff language for hardware at least in 2022. Now Chisel generates a macro that initializes all registers, from which the designer cannot even tell how many registers are left uninitialized and whether they really affects the functionality. Chisel do need to provide an elegant way for the users to know whether the uninitialized registers would cause errors. This is the current workflow from us: mark all registers to be uninitialized and find all the registers causing x-prop. Then we have to manually create a deposit/force file that cannot be generated with the default open-source transforms. Or we have to initialize these registers in Chisel.

I agree that some registers can be intentially left uninitialized and this is a real hardware design requirement. I think the main issue is that the users must know the range of registers that are intentially left uninitialized and do not affect functionality. Chisel should not make decisions for the users to randomly initialize all registers. However, since Chisel currently don't have the transforms/annotations to allow the users to specify the list of uninitialized registers, should we provide a temporary workaround to resolve the X issues?

That is, in 2022, to please the verilog simulators and the signoff workflows, Chisel users must address the X issues. Since we don't have these intelligent methods now, should we provide low but effective patches to temporarily fix this issue?

Though we now want these stupid patches to initialize the registers, we will transfer to CIRCT in the visible feature and would like to create useful transforms to address the issues we met when using Chisel, as we have did with the scala compiler. These two workarounds are not mutually exclusive.

mwachs5:
Can we just make this library component take an argument if it wants to be initialized like this? And extend the arbiter
 test that it works with both forms?

ekiwi:
> That said, Chisel likely needs better APIs for describing _intentionally_ uninitialized registers and intentional, legal reads of these registers.

Doesn't `Reg(...)` already declare an  _intentionally_ uninitialized register? I generally use `RegInit(...)` unless I am pretty sure that the initial value does not matter. Is this maybe a problem with how these two functions are named? I would be curious if there are any actual bugs with register initialization in the Chisel library that people found. Most PRs and issues seem to be about false positives.

poemonsense:
It is not bug but causes x-propagation in VCS and other simulators with X states. Since this arbiter usually lies in the control path, instead of the data path, this uninitialized register always causes the system to abort due to lots of X in the circuit.

ekiwi:
> It is not bug but causes x-propagation in VCS and other simulators with X states. Since this arbiter usually lies in the control path, instead of the data path, this uninitialized register always causes the system to abort due to lots of X in the circuit.

Yes, I understand. Have you considered writing a firrtl pass that allows you to add reset values to registers after the Chisel is elaborated? You could use an aspect to select which registers you want to target.

seldridge:
> Doesn't Reg(...) already declare an intentionally uninitialized register?

Sorry, unclear by me. I'm worried about separation of an intentional read of an uninitialized register vs. an incorrect (unintentional) read of an uninitialized register.

E.g., separation of reads of `lastGrant` vs reads of a program counter.

pineapplehunter:
I am having the same issue here with Icarus backend.
I understand that on real hardware this is not an issue, but it wold be much easier to debug if there is a default value IMO.
if someone really doesn't want the extra overhead, can there be an argument to explicitly **not** initialize the register? Make the default initialized and a option to make it undefined for advanced usage.

ekiwi:
> I am having the same issue here with Icarus backend.

Is this with chiseltest? We really need to enable randomization by default.

pineapplehunter:
> Is this with chiseltest? 
Yes, it is.
I was using Icarus because the results are closer to Vivado's behavior simulation.

PR #270, opened 2016-08-30
Add Printable

jackkoenig:
Printable is a new type that changes how printing of Chisel types is represented
It uses an ordered collection rather than a format string and specifiers
Main Features:
 - Custom String Interpolator for Scala-like printf
 - String-like manipulation of "hardware strings" for custom pretty-printing
 - Default pretty-printing for Chisel data types

A few notes and questions:
- This does *not* remove the old printf functionality
- This partially resolves @ccelio's need for hardware strings, it allows for strings to be constructed naturally although it will emit the full string for each printf that results
- There is a bit of an inconsistency between the new naming API and this which uses FullName and Name. FullName corresponds to instanceName, and Name is just the local name of a field (eg. within a bundle)
- Should this actually be a part of Chisel? Here are the reasons it's not currently a library
  - want the ability to put names of Chisel things in resulting printfs (useful for logging)
  - want default pretty-printing of Chisel types (Vec prints like Scala Seq, Bundle prints like Scala map)
- It is a bit cumbersome to type `p"blah blah Decimal(myWire) blah blah"` so I suggest that if this is merged we come up with shorthand. Decimal, Hexadecimal, Binary, and Character are just the underlying case class names, we could create other ways to create those Printables

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -169,6 +169,17 @@ sealed class Vec[T <: Data] private (gen: => T, val length: Int)
 
   for ((elt, i) <- self zipWithIndex)
     elt.setRef(this, i)
+
+  /** Default "pretty-print" implementation
+    * Analogous to printing a Seq
+    * Results in "Vec(elt0, elt1, ...)"
+    */
+  def toPrintable: Printable = {
+    val elts =
+      if (length == 0) List.empty[Printable]
+      else self flatMap (e => List(e.toPrintable, PString(", "))) dropRight 1
+    PString("Vec(") + Printables(elts) + PString(")")
+  }
 }
 
 /** A trait for [[Vec]]s containing common hardware generators for collection
@@ -369,6 +380,21 @@ class Bundle extends Aggregate(NO_DIR) {
         Builder.error(s"Parameterized Bundle ${this.getClass} needs cloneType method")
         this
     }
+  }
+
+  /** Default "pretty-print" implementation
+    * Analogous to printing a Map
+    * Results in "Bundle(elt0.name -> elt0.value, ...)"
+    */
+  def toPrintable: Printable = {
+    val elts =
+      if (elements.isEmpty) List.empty[Printable]
+      else {
+        elements.toList.reverse flatMap { case (name, data) =>
+          List(PString(s"$name -> "), data.toPrintable, PString(", "))
+        } dropRight 1 // Remove trailing ", "
+      }
+    PString("Bundle(") + Printables(elts) + PString(")")
   }
 }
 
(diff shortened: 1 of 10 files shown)


jackkoenig:
One really powerful feature of all of this is the ability to give a custom pretty printing for custom Bundles.

First, consider the default printing of custom types:
```scala
class MyPacket extends Bundle {
  val valid = Bool()
  val addr = UInt(width = 32)
  val data = UInt(width = 64)
  override def cloneType = (new MyPacket).asInstanceOf[this.type]
}
class MyTester extends BasicTester {
  val (count, done) = Counter(Bool(true), 2)
  val myBun = Wire(new MyPacket)
  myBun.addr := UInt(100)
  myBun.data := UInt(1000)
  when (count === UInt(0)) {
    myBun.valid := Bool(false)
  } .otherwise {
    myBun.valid := Bool(true)
  }
  printf(p"@$count: $myBun\n")
  when (done) { stop() }
}
```

This will print
```
@0: Bundle(valid -> 0, addr ->        100, data ->                 1000)
@1: Bundle(valid -> 1, addr ->        100, data ->                 1000)
```

We can change it up a little by adding a custom .toPrintable to our custom Bundle, and even do some basic "hardware string" manipulation
```scala
class MyPacket extends Bundle {
  val valid = Bool()
  val addr = UInt(width = 32)
  val data = UInt(width = 64)
  override def cloneType = (new MyPacket).asInstanceOf[this.type]
  override def toPrintable = {
    p"(${this.getClass.getSimpleName}) ${FullName(this)}: " +
    p"[0x${Hexadecimal(addr)}]: $data"
  }
}
class MyTester extends BasicTester {
  val (count, done) = Counter(Bool(true), 2)
  val myBun = Wire(new MyPacket)
  myBun.addr := UInt(100)
  myBun.data := UInt(1000)
  when (count === UInt(0)) {
    myBun.valid := Bool(false)
  } .otherwise {
    myBun.valid := Bool(true)
  }
  val prefix = p"@$count: "
  when (myBun.valid) {
    printf(prefix + p"$myBun\n")
  } .otherwise {
    printf(prefix + p"INVALID\n")
  }
  when (done) { stop() }
}
```

This prints the following
```
@0: INVALID
@1: (MyPacket) myBun: [0x00000064]:                 1000
```

I think this is pretty neat. Unfortunately it does lack the ability to do this conditional "hardware string" manipulation in functions (you can't put whens in the toPrintable method), but eventually we can probably add that feature eventually.

aswaterman:
This is cool, but I agree we'd want to sort out the shorthand API.

ben-k:
This is great!  It's been on my Chisel wishlist for years: ucb-bar/chisel#178

ucbjrl:
Unfortunately, this seems to break chisel-testers and firrtl-interpreter. Details to come.

PR #274, opened 2016-09-07
Add IrrevocableIO alternative to DecoupledIO

hcook:
… that promises not to change the contents of .bits on a cycle after .valid is high and .ready is low. This is useful for detecting whether queues have been correctly inserted in front of modules that require these additional semantics.

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -7,33 +7,59 @@ package chisel3.util
 
 import chisel3._
 
-/** An I/O Bundle with simple handshaking using valid and ready signals for data 'bits'*/
-class DecoupledIO[+T <: Data](gen: T) extends Bundle
+/** An I/O Bundle containing 'valid' and 'ready' signals that handshake
+  * the transfer of data stored in the 'bits' subfield.
+  * The base protocol implied by the directionality is that the consumer
+  * uses the flipped interface. Actual semantics of ready/valid are
+  * enforced via use of concrete subclasses.
+  */
+abstract class ReadyValidIO[+T <: Data](gen: T) extends Bundle
 {
   val ready = Bool(INPUT)
   val valid = Bool(OUTPUT)
   val bits  = gen.cloneType.asOutput
   def fire(dummy: Int = 0): Bool = ready && valid
+}
+
+/** A concrete subclass of ReadyValidIO signalling that the user expects a
+  * "decoupled" interface: 'valid' indicates that the producer has
+  * put valid data in 'bits', and 'ready' indicates that the consumer is ready
+  * to accept the data this cycle. No requirements are placed on the signalling
+  * of ready or valid.
+  */
+class DecoupledIO[+T <: Data](gen: T) extends ReadyValidIO[T](gen)
+{
   override def cloneType: this.type = new DecoupledIO(gen).asInstanceOf[this.type]
 }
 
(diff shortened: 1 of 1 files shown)


aswaterman:
There's also the constraint that valid will stay high once it's asserted, right? Should probably be pedantic about the description in the comment.

sdtwigg:
The subclasses seems improper since the subclassing contract fails, for example, when this is used for 'enqueue' sides of Modules. The underling issue is that IrrevocableIO is indicating (via subclassing) that it can be used any time you expect a DecoupledIO.
This is fine for a producer dequeue side: A Module that can consume a DecoupledIO can easily consume an IrrevocableIO. 
This is not fine for a consumer enqueue side. You may have a function (or class, etc.) that takes as input a consumer port and connects a producer to it. Since IrrevocableIO <: DecoupledIO, you are allowed to give it an IrrevocableIO. However, the function thinks it has a DecoupledIO and will not be bound to the same contract.

```scala
class SensitiveSink extends Module {
  val io = IO(new Bundle {
    val enq = Flipped(IrrevocableIO(UInt(16)))
  })
  // do something
}

def driveFromUnreliableSource(enq: DecoupledIO[UInt]) = {
  val parity = Reg(init = false.asBool)
  val bits = Reg(init = 1.asUInt(16))
  enq.valid := parity
  enq.bits := Mux(parity, bits, 0.asUInt)
  when(enq.valid && enq.ready) { bits := bits +% 1.asUInt }
}

class Test extends BasicTester {
  val sink = Module(new SensitiveSink)
  driveFromUnreliableSource(sink.io.enq) // will compile and elaborate fine
}
```

aswaterman:
Is your proposal to not make it descend from DecoupledIO but instead be a separate part of the type hierarchy?

sdtwigg:
Yes

hcook:
I totally see your point. I was scared of making backwards-incompatible changes to QueueIO, but it looks like there are only 3 breakages in current RocketChip, one of which will be fixed by adding another factory apply method to Queue.

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -7,23 +7,63 @@ package chisel3.util
...
+{
+  override def cloneType: this.type = new DecoupledIO(gen).asInstanceOf[this.type]
+}
+
+/** Factory adds a decoupled handshaking protocol to a data bundle. */
+object Decoupled
+{
   def apply[T <: Data](gen: T): DecoupledIO[T] = new DecoupledIO(gen)
+
+  def apply[T <: Data](irr: IrrevocableIO[T]): DecoupledIO[T] = {
```
aswaterman: can you add scaladoc for the two apply methods?  the new one should explain the use case and the invariant

aswaterman:
I think this is ok now

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -164,22 +205,35 @@ extends Module(override_reset=override_reset) {
...
+      entries: Int = 2,
+      pipe: Boolean = false): IrrevocableIO[T] = {
+    val deq = apply(enq, entries, pipe)
+    val irr = Wire(new IrrevocableIO(deq.bits))
+    irr.bits := deq.bits
+    irr.valid := deq.valid
+    deq.ready := irr.ready
+    irr
+  }
 }
```
terpstra: While we're at it, could we have 'flow' added as another defaulted argument on the Queue.methods?

PR #280, opened 2016-09-16
Decoupled: cast DecoupledIO to IrrevocableIO as an input

terpstra:
We can cast an output IrrevocableIO to DecoupledIO.
Likewise, we can cast an input DecoupledIO to an IrrevocableIO.


aswaterman:
lgtm

PR #281, opened 2016-09-18
Driver: Check the simulation exit code

SebastianBoe:
Fail the test if the exit code of the simulation is non-zero.

Simulation timeouts return a non-zero exit code, so after this commit
tests that time out will be correctly reported as failures.


ucbjrl:
Extend ExpectedFailure tester.

ucbjrl:
Closed by #320

SebastianBoe:
Thanks! #320 resolves my problem.

PR #283, opened 2016-09-19
remove trailing whitespace for annotations

sbeamer:
Is this a typo, or is there a reason one would want a space following the source annotations?


aswaterman:
that might be a vestige from when source line annotations preceded the code, rather than following it.

colinschmidt:
@jackkoenig says this is fine to merge

PR #285, opened 2016-09-20
Add Cookbook tests

jackkoenig:
These tests are intended to be the examples in the Chisel3 Wiki Cookbook.

The primary idea is that the Cookbook examples should always be working so I want to generate the actual documentation on the Wiki from these examples. I am looking into tools to help do this, but haven't been able to get one working yet.

https://github.com/ucb-bar/chisel3/wiki/Cookbook


jackkoenig:
Changed to Hex

PR #288, opened 2016-09-21
Make implicit clock name consistent

### LINKED ISSUE #258: Name of implicit clock is inconsistent
In the emitted FIRRTL, we call the implicit clock "clk."  In Chisel code, one refers to the implicit clock by accessing Module.clock.  Either's fine, but the discrepancy is uncool.  Changing either one will break some users' code (though changing the FIRRTL output is probably most disruptive).  Should make a decision sooner than later.

aswaterman:
In the Chisel frontend, the implicit clock is named clock, but in the
generated FIRRTL, it is named clk.  There is no reason for this
discrepancy, and yet fixing it is painful, as it will break test harnesses.
Better to take the pain now than later.

Resolves #258.


hcook:
better to get it over with. at least "clock" and "reset" match each other now.

PR #290, opened 2016-09-21
Expose FIRRTL asClock construct

aswaterman:
This allows people to write clock-management circuitry in Chisel, rather
than mandating the use of BlackBoxes.


aswaterman:
NO

aswaterman:
jk

PR #291, opened 2016-09-22
Scaladocs for utils

ducky64:
Add scaladocs for chisel3.util functions which didn't have them, and improves scaladocs for some others.

Except Lookup, that was undocumented and I can't figure out what it's supposed to do.

Treat this as a first pass ("basically usable") towards nicer scaladoc. More work needed (later) to make formatting / tone / style more consistent.


chick:
lgtm.
Appears safe, all just comment changes.  Examples given look good.

PR #293, opened 2016-09-23
Add support for parameterized BlackBoxes

### LINKED ISSUE #289: BlackBoxes silently drop hardware
The Chisel below

``` scala
class MyBlackBox extends BlackBox {
  val io = new Bundle {
    val in = Bool().asInput
    val out = Bool().asOutput
  }

  val (_, done) = Counter(io.in, 100)
  io.out := done
}

  val io = new Bundle { }
  val bb = Module(new MyBlackBox)
}
```

This emits the following Firrtl

```
circuit MyModule : 
  extmodule MyBlackBox : 
    output out : UInt<1>
    input in : UInt<1>


  module MyModule : 
    input clk : Clock
    input reset : UInt<1>
    output io : {}

    io is invalid
    inst bb of MyBlackBox @[BlackBox.scala 22:18] 
    bb.out is invalid
    bb.in is invalid
```

In Chisel 2, it made sense to allow hardware in BlackBoxes because it could be used in the C++ emulator. We don't have an equivalent in Firrtl so this silent dropping is problematic.

jackkoenig:
Also restrict black boxes to not allow hardware inside of them since it was
being silently dropped anyway.

Resolves #289 

This PR obviously depends on ucb-bar/firrtl#314 in order to pass tests. Also, thanks to the new Firrtl feature where extmodules can overwrite their name, Chisel emits BlackBoxes that will always get their desired names.

The current API proposed by this PR is to make blackbox parameters be an argument to the BlackBox abstract class of type Map[String, Any] where Any is dynamically checked to be only be Int, Long, BigInt, Double, or String. This leads to the use in one of the test cases of:

``` scala
class BlackBoxConstant(params: Map[String, Any]) extends BlackBox(params) {
  val io = new Bundle {
    val out = UInt(width = params("WIDTH").asInstanceOf[Int]).asOutput
  }
}

class BlackBoxWithParamsTester extends BasicTester {
  val blackBoxOne  = Module(new BlackBoxConstant(Map("WIDTH" -> 1, "VALUE" -> 1)))
  val blackBoxFour  = Module(new BlackBoxConstant(Map("WIDTH" -> 3, "VALUE" -> 4)))
  ...
}
```

Alternatively, params could be a def overridden in the BlackBox body itself, then users could pass their parameters however they see fit and transform them to the Map[String, Any] in the body of the BlackBox.

I'm not the biggest fan of the above syntax and am certainly open to suggestions.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
@@ -3,10 +3,19 @@
 package chisel3.core
 
 import chisel3.internal.Builder.pushCommand
-import chisel3.internal.firrtl.{ModuleIO, DefInvalid}
+import chisel3.internal.firrtl._
+import chisel3.internal.throwException
 import chisel3.internal.sourceinfo.SourceInfo
 // TODO: remove this once we have CompileOptions threaded through the macro system.
 import chisel3.core.ExplicitCompileOptions.NotStrict
+
+/** Parameters for BlackBoxes */
+sealed abstract class Param
+case class IntParam(value: BigInt) extends Param
+case class DoubleParam(value: Double) extends Param
+case class StringParam(value: String) extends Param
+/** Unquoted String */
+case class RawParam(value: String) extends Param
 
 /** Defines a black box, which is a module that can be referenced from within
   * Chisel, but is not defined in the emitted Verilog. Useful for connecting
@@ -16,12 +25,9 @@ import chisel3.core.ExplicitCompileOptions.NotStrict
   * {{{
   * ... to be written once a spec is finalized ...
   * }}}
+  * @note The parameters API is experimental and may change
   */
-// REVIEW TODO: make Verilog parameters part of the constructor interface?
-abstract class BlackBox extends Module {
-  // Don't bother taking override_clock|reset, clock/reset locked out anyway
-  // TODO: actually implement this.
-  def setVerilogParameters(s: String): Unit = {}
+abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param]) extends Module {
 
   // The body of a BlackBox is empty, the real logic happens in firrtl/Emitter.scala
   // Bypass standard clock, reset, io port declaration by flattening io
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -41,8 +41,16 @@ object Module {
                      sourceInfo.makeMessage(" See " + _))
     }
     Builder.currentModule = parent // Back to parent!
+
     val ports = m.computePorts
-    val component = Component(m, m.name, ports, m._commands)
+    // Blackbox inherits from Module so we have to match on it first TODO fix
+    val component = m match {
+      case bb: BlackBox =>
+        DefBlackBox(bb, bb.name, ports, bb.params)
+      case mod: Module =>
+        mod._commands.prepend(DefInvalid(childSourceInfo, mod.io.ref)) // init module outputs
+        DefModule(mod, mod.name, ports, mod._commands)
+    }
     m._component = Some(component)
     Builder.components += component
     // Avoid referencing 'parent' in top module
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -181,7 +181,10 @@ private[chisel3] object Builder {
   // TODO(twigg): Ideally, binding checks and new bindings would all occur here
   // However, rest of frontend can't support this yet.
   def pushCommand[T <: Command](c: T): T = {
-    forcedModule._commands += c
+    forcedModule match {
+      case _: BlackBox => throwException("Cannot add hardware to a BlackBox")
+      case m => m._commands += c
+    }
     c
   }
   def pushOp[T <: Data](cmd: DefPrim[T]): T = {
--- a/chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -215,8 +215,14 @@ case class Connect(sourceInfo: SourceInfo, loc: Node, exp: Arg) extends Command
 case class BulkConnect(sourceInfo: SourceInfo, loc1: Node, loc2: Node) extends Command
 case class ConnectInit(sourceInfo: SourceInfo, loc: Node, exp: Arg) extends Command
 case class Stop(sourceInfo: SourceInfo, clock: Arg, ret: Int) extends Command
-case class Component(id: Module, name: String, ports: Seq[Port], commands: Seq[Command]) extends Arg
 case class Port(id: Data, dir: Direction)
 case class Printf(sourceInfo: SourceInfo, clock: Arg, pable: Printable) extends Command
+abstract class Component extends Arg {
+  def id: Module
+  def name: String
+  def ports: Seq[Port]
+}
+case class DefModule(id: Module, name: String, ports: Seq[Port], commands: Seq[Command]) extends Component
+case class DefBlackBox(id: Module, name: String, ports: Seq[Port], params: Map[String, Param]) extends Component
 
 case class Circuit(name: String, components: Seq[Component])
--- a/src/main/scala/chisel3/Driver.scala
+++ b/src/main/scala/chisel3/Driver.scala
@@ -119,13 +119,13 @@ trait BackendCompilationUtilities {
         "-Wno-WIDTH",
         "-Wno-STMTDLY",
         "--trace",
-        "-O0",
+        "-O1",
         "--top-module", topModule,
         "+define+TOP_TYPE=V" + dutFile,
         s"+define+PRINTF_COND=!$topModule.reset",
         s"+define+STOP_COND=!$topModule.reset",
         "-CFLAGS",
-        s"""-Wno-undefined-bool-conversion -O0 -DTOP_TYPE=V$dutFile -include V$dutFile.h""",
+        s"""-Wno-undefined-bool-conversion -O1 -DTOP_TYPE=V$dutFile -include V$dutFile.h""",
         "-Mdir", dir.toString,
         "--exe", cppHarness.toString)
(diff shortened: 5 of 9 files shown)


aswaterman:
Looking forward to trying this out soon.  @mwachs5

mwachs5:
Yes, looking forward to this!

 Not sure I understand your comment "Alternatively, params could be a def overridden in the BlackBox body itself, then users could pass their parameters however they see fit and transform them to the Map[String, Any] in the body of the BlackBox."

Do you mean that is currently an option, or are you looking for feedback on whether that would be a nicer API? I'm not sure it's that big a difference, since you can always do arbitrary processing and construction of the parameters in a subclass, as you've demonstrated.

I haven't used the crazier parameter types being discussed on the firrtl list.

jackkoenig:
I'm just looking for feedback for a possibly nicer API. You're right that it's not a huge difference, but by passing the params to the BlackBox superclass constructor, you don't have a whole lot of "space" if you will to do the conversion. Let's pretend we're using CDE and the API is instead to override a def:

```scala
class BlackBoxConstant(implicit p: Parameters) extends BlackBox {
  val io = new Bundle {
    val out = UInt(width = p("WIDTH")).asOutput
  }
  override def params = Map("WIDTH" -> p("WIDTH"), "VALUE" -> p("VALUE"))
}

class BlackBoxWithParamsTester extends BasicTester {
  implicit val p = ...
  val blackBoxOne  = Module(new BlackBoxConstant))
  val blackBoxFour  = Module(new BlackBoxConstant))
  ...
}
```

Basically, we could instead just use our own parameter system and convert to the BlackBox parameter system inside the BlackBox body itself rather than needing the Map[String, Any] available to be passed to the BlackBox superclass constructor.

mwachs5:
I think I prefer this later interface. This way the parameters are kept with the module they are describing.

It's also more aligned with how you handle 'io'. we don't pass that in to the BlackBox constructor, so why should parameters be different.

sdtwigg:
If the initialization code of the BlackBox doesn't need params (by quick inspection, it doesn't) then it seems unnecessary to pass as a constructor parameter. You may want to define it as the empty map (perhaps as a def) and have users override to supply their own, like with name.

This scheme also avoids needing to obfuscate the code where you force the user to upcast the parameter  to Any and then downcast to 'whatever'.

aswaterman:
That seems awkward, especially given that the common case is that BlackBoxes will take parameters.

aswaterman:
If I'm parsing this thread right, @sdtwigg's opinion is about purity of style, and so is mostly a distraction from the usability discussion at hand.  OTOH, I think @mwachs5's argument stems from the notion that parameters, like io, are part of a module's interface; their types/names should be part of the module declaration, whereas their values should be bound at instantiation time.  @jackkoenig's second proposal partially achieves that goal, but doesn't illustrate the opportunity for type safety, and doesn't illustrate that it can be done with DRY.

I still favor the approach of passing args to the BlackBox constructor.  I think it's akin to parameterizing a Module on type.  Consider Queue.  Ignoring the Scala type system for a thought experiment, you could consider making it an abstract class, which you make concrete with "def tpe = foo."  That would achieve the same thing as passing tpe to the Queue constructor, but seems more awkward.

Similarly, it's not uncommon to parameterize Module based on an integer parameter, like the width of a signal.  To me, it's most natural to pass that as an arg to the constructor.

jackkoenig:
If I'm understanding your point correctly, Andrew, then here's an updated version of the original example:

```scala
class BlackBoxConstant(value: Int) extends BlackBox(
    Map("WIDTH" -> log2Up(value), "VALUE" -> value) {
  require(value >= 0)
  val io = new Bundle {
    val out = UInt(width = log2Up(value)).asOutput
  }
}

class BlackBoxWithParamsTester extends BasicTester {
  val blackBoxOne  = Module(new BlackBoxConstant(1))
  val blackBoxFour  = Module(new BlackBoxConstant(4))
  ...
}
```

Something more like that?

aswaterman:
I think that looks pretty clean, as far as BlackBoxes are concerned.

sdtwigg:
I argue the example illustrates the downside of forcing the parameters to be resolved for the BlackBox module.

Both the parameter map and blackbox IO had to replicate the parameter processing (log2Up(value)). While the computation could be abstracted to a function in the companion object, there is no good caching place (since the class has yet to be constructed) leading to inefficiency. Trying to extract from the map itself requires the type casts, which at the very least are fairly lengthy. This all just seems so unnecessary since the BlackBox constructor doesn't even use the params.

You can use early initializer syntax to bypass but that ends up looking far worse....
```scala
class BlackBoxConstant(value: Int) extends {
  require(value >= 0)
  val width = log2Up(value)
} with BlackBox(Map("WIDTH" -> width, "VALUE" -> value))
{
  val io = new Bundle {
    val out = UInt(width = width).asOutput
  }
}
```
However, this is particularly risky because the internals of Chisel are not designed to handle general use of early initializers. (Here, it is probably fine. In general Modules, you run a high risk of error.)

jackkoenig:
Alright I've updated this to support unquoted string parameters (RawParam), and to use implicit conversions rather than the hacky matching thing before. This will still fail until the Firrtl-side is merged, and I'm still passing a map to the BlackBox abstract class constructor. Now it's a Map[String, Param] instead of Map[String, Any]

jackkoenig:
Bumping this since Firrtl master now supports it.

Kind of ironically given #332, this requires Verilator being invoked with O1 or higher. The BlackBoxWithParamsTester gets the following warning and error:

```
-Info: Command Line disabled gate optimization with -Og/-O0.  This may cause ordering problems.
%Error: Internal Error: ../V3File.h:156: : /tmp/BlackBoxWithParamsTester7512079037470064518/VBlackBoxWithParamsTester_BlackBoxTypeParam__z1549.cpp: Underflow of indentation
%Error: Internal Error: See the manual and http://www.veripool.org/verilator for more assistance.
```

I would love an alternative fix but I'm really not sure what the deal is here.

The API is still to pass Map to BlackBox constructor. There's some disagreement on this so we should decide what to do so I can either implement what's different or go ahead and merge.

aswaterman:
-O1 enables several individual -f options.  You could try to figure out
which option is actually necessary and enable only that one.

jackkoenig:
I'm having trouble finding any documentation about specific optimizations but I found this concerning gem:

```
-Ooptimization-letter

Rarely needed. Enables or disables a specific optimizations, with the optimization selected based
on the letter passed. A lowercase letter disables an optimization, an upper case letter enables it.
This is intended for debugging use only; see the source code for version-dependent mappings of
optimizations to -O letters.
```

Note: **see the source code for version-dependent mappings of optimizations to -O letters.**

aswaterman:
See https://github.com/riscv/riscv-old-gcc/blob/riscv-gcc-6.1.0/gcc/opts.c#L442

-O1 enables all the ones marked OPT_LEVELS_1_PLUS.

wallento:
@jackkoenig Do you have the example you used packed somewhere? I can have a look at the verilator issue

jackkoenig:
Hi @wallento, I made a git repo from the failing test:  https://github.com/jackkoenig/verilator-issue

If you do just `make` it will build and run the test. However if you do:

```
make clean
make VOPT=0
```

You will get the following error:

```
-Info: Command Line disabled gate optimization with -Og/-O0.  This may cause ordering problems.
%Error: Internal Error: ../V3File.h:156: : obj_dir/VBlackBoxWithParamsTester_BlackBoxTypeParam__z1549.cpp: Underflow of indentation
%Error: Internal Error: See the manual and http://www.veripool.org/verilator for more assistance.
```

I can also provide instructions on how to reproduce the error from chisel3 tests but I thought this way would be easier.

wallento:
Thanks, @jackkoenig. I am onto it, it is an emitter issue with the parameter generation. For types it ends up using an invisible character (`\x0` to make it worst). File name generation and emitting into that file hence fail epically. This only happens with `-O0` as it then emits the whole hierarchy, with `-O1` it already optimized that out.

I think I can fix this today, will submit a verilator patch and push it to a public repo for testing.

wallento:
I solved the issue and submitted the patch to verilator here: http://www.veripool.org/issues/1101-Verilator-Fix-SmallName-for-ParamTypeDType

Please use my Verilator repository at branch `fix-typeparamname` for testing: https://github.com/wallento/verilator/tree/fix-typeparamname

Looking forward to your feedback. I hope that will be merged into the next version of verilator then.

aswaterman:
thanks @wallento!

jackkoenig:
Thanks so much! I look forward to this fix being part of a release. In the meantime, do you think we should just stick to `-O1` or should we specifically add in whatever optimizations prevent this from happening?

wallento:
I would only use `-O0` when really necessary. I still need to understand the issue why you are moving to it fully, but I think we are talking about a huge loss of performance here. I will look up over the weekend if there is a combination of `-O0` and switches for the current verilator build to workaround.

jackkoenig:
We certainly don't *need* to. We moved away from `-O2` because we were running out of memory on Travis runs, but it really is only a bandaid fix. As for performance, these Verilator runs are really small unit tests where the build time far dominates run time. Anyone using Verilator for actual simulation should be invoking it themselves, not using this repo's helper functions for its own testing. I'm intending to remove this functionality altogether from chisel3 and firrtl in the near future.

wallento:
Ah, I see. For your case you should consider trying `VM_PARALLEL_BUILDS=1 make -j8` that will speed up build times significantly, as here: https://github.com/ucb-bar/rocket-chip/blob/master/emulator/Makefrag-verilator#L69

wallento:
The patch is applied to the verilator source tree by the way and will be in 3.889.

aswaterman:
Nice!

jackkoenig:
Poking this again, it's a very desirable feature and it's blocking #347.

The current API is that the abstract class Blackbox can accept a Map[String, Param] and Params box BigInt, Double, or String.

This has some limitations if you want to convert from some custom parameter system to this system, you might have some code duplication in the body of the blackbox for using the same parameters to create the ports. There's also the awkwardness that maybe instead of a Map[String, Param], parameters should have a String "name" field and it should just be Seq[Param]. Also, a possible alternative to passing the parameters to the Blackbox constructor is to override some method parameters.

aswaterman:
Who are you poking exactly?

jackkoenig:
Anyone in general, I think this is a desirable feature but as a public API it requires some agreement

colinschmidt:
I think this API is good enough for a first pass and we should just mark it as experimental and merge it.

PR #301, opened 2016-09-28
Generate scaladoc for all subprojects.

ucbjrl:
Use aggregate to run project tasks in subprojects
Use sbt-unidoc to generate combined scaladoc for project and subprojects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,10 +6,18 @@ site.includeScaladoc()
 
 ghpages.settings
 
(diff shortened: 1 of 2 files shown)


ducky64:
lgtm

ucbjrl:
Unfortunately, due to the way sbt (doesn't) deal with subproject plugins, this requires adding the plugin to parent project plugins. We may wish to revisit this.

PR #302, opened 2016-09-29
Massive rename of CompileOptions.

ucbjrl:
Massage CompileOption names in an attempt to preserve default (Strict) CompileOptions in the absence of explicit imports.
NOTE: Since the default is now strict, we may encounter errors when we generate connections for clients (i.e., in Vec.do_apply() when we wire up a sequence).
We should really thread the CompileOptions through the macro system so the client's implicits are used.


ducky64:
There's a lot of commented out code (mostly in terms of CompileOption imports) - you should just remove them (unless there's a reason for keeping them, but not using them?)

Also, why the separation between ExplicitCompileOptions and ImplicitCompileOptions, even for users of CompileOptions? Would it make sense to have a CompileOptions interface / trait, and Strict / NonStrict be a type of that? (though probably naming it to separate the CompileOptions interface / trait from the CompileOptions hierarchy for Strict / NonStrict)

ucbjrl:
The commented out code is there to remind me that it may need to be re-enabled, and in fact, rocket-chip no longer compiles with an internal default of Strict. We need to either revert to NonStrict or thread the CompileOptions through the macros so we can inherit them internally (does that make sense?).

The separation of ExplicitCompileOptions and ImplicitCompileOptions is due to my understanding (likely incorrect) of the implicit resolution hierarchy and the subsequent belief that a "more specific" implicit definition overrides a "less specific" one.

Please correct.

ducky64:
Yeah, it seems that Scala looks for the most specific implict when resolving implicits in the current scope (priority 1). However, the materialize implicit is priority 2 (companion objects of a type), so we don't need the type hierarchy for the default compile option.
http://docs.scala-lang.org/tutorials/FAQ/finding-implicits.html

One place where this could get tricky is that since the chisel2 compatibility package would import the non-strict compile option implicit, it would make it impossible to import the strict compile option implicit. Though using strict in chisel2 compatibility mode probably doesn't make a lot of sense anyways.

ucbjrl:
Good. I'll fix that and re-enable NotStrict for the internally generated connections until we manage to thread the appropriate implicits through the macro system.

I agree that using Strict in Chisel2 compatibility mode doesn't make much sense.

aswaterman:
Yeah, I don't think strict-chisel2 makes sense.

ducky64:
lgtm (assuming it compiles)

Note that you still have a commented-out import of chisel3.CompileOptions in chiselFrontend, where the chisel3 package object shouldn't yet be visible. But presumably those comments will either be completely removed, or put back in (and fixed, because it won't compile).

ucbjrl:
Thanks Richard. I was going to clean it up a bit before merging but you were to fast for me. I deleted those commented-out imports.

rocket-chip master ( ucb-bar/rocket-chip@e928b741ced0992468f621e95e54f1a666b47269 ) compiles and passes `make run`

Issue #306, opened 2016-09-30
firrtlDirection needs cloning


ucbjrl:
In #265, @sdtwigg points out this value isn't being cloned correctly, and style issues of getters and setters for this value

ucbjrl:
Closed by #307

PR #307, opened 2016-09-30
clone firrtlDirection when cloning - Issue #306

ucbjrl:
Fix for #306.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Data.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Data.scala
@@ -155,7 +155,19 @@ abstract class Data extends HasId {
   private[chisel3] def toType: String
   private[core] def width: Width
 
+  /** cloneType must be defined for any Chisel object extending Data.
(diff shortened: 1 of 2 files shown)


sdtwigg:
Double check to ensure it works even though intermediate elements are not getting their directions manually fixed up. (The fact that the root Bundle wholly reconstructs the intermediates with Flipped means it likely shouldn't be a problem since the reconstruction should automatically fix them up. Actually, because of this it is odd that the leaves could be a problem except in very pathological cases. When I originally made the comment, I had a different understanding of where firrtlDirection was being stored.)

I see cloneType isn't protected. If people are using that instead of chiselCloneType, then there could be problems. (I didn't notice that the newType from the PR was renamed to chiselCloneType).

ucbjrl:
There is definitely room for improvement here. Adding an issue to track this - #321.

Client cloneType methods should only call element cloneType methods. chiselCloneType should probably be renamed topCloneType and should only be called for the top-most clone call (i.e., it should never be called from within cloneType).

PR #311, opened 2016-10-04
Add CompileOptions implicits to all Module constructors - fix #310.

### LINKED ISSUE #310: Compatibility mode can't handle Modules with reset/clock passed in
Instantiating this module in compatibility mode will trigger a binding error:

    class Test(reset: Bool) extends Module(_reset = reset) {
      val io = new Bundle {
        val done = Bool(OUTPUT)
      }

      io.done := Bool(false)
    }

Without the reset argument, it succeeds.

The same issue holds for passing in a clock.

ucbjrl:
class Module has three other constructors (for variations on whether clock, reset, or both are explicitly provided). These also require an implicit CompileOptions parameter.


aswaterman:
thanks @ucbjrl -- i'll give this a shot

PR #313, opened 2016-10-05
Buildinfo

ucbjrl:
Generate BuildInfo and print it out at Driver object creation. As Chisel versions proliferate, we need to be able to tie issues to specific versions.


build.sbt
```diff
@@ -101,6 +101,25 @@ lazy val chiselFrontend = (project in file("chiselFrontend")).
   dependsOn(coreMacros)
 
 lazy val chisel = (project in file(".")).
+  enablePlugins(BuildInfoPlugin).
+  settings(
+    // We should really be using name.value, but currently, the package is "Chisel" (uppercase first letter)
+    buildInfoPackage := /* name.value */ "chisel3",
+    buildInfoOptions += BuildInfoOption.BuildTime,
+    buildInfoKeys := Seq[BuildInfoKey](buildInfoPackage, version, scalaVersion, sbtVersion),
+    // Move the managed source directory where git won't complain about it,
```
ducky64: The plugin documentation https://github.com/sbt/sbt-buildinfo says the generated source is under `target/`, which should be ignored already? And as long as it doesn't drop the file in one of the existing source folders (is it?), adding an entry in .gitignore may be a better solution?

Also, this comment block needs reformatting. And what's up with the commented out line of code?

src/test/scala/chiselTests/BuildInfoTests.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import org.scalatest.FlatSpec
+
+//import Chisel._
+
+class BuildInfoTests extends FlatSpec {
+  behavior of "BuildInfoTests"
+
+  it should "provide correct BuildInfo" in {
+    println(Chisel.Driver.chiselVersionString)
```
ducky64: `println` does not a good unit test make...

src/main/scala/chisel3/Driver.scala
```diff
@@ -135,4 +136,8 @@ object Driver extends BackendCompilationUtilities {
   }
 
   def targetDir(): String = { target_dir getOrElse new File(".").getCanonicalPath }
+
+  val version = BuildInfo.version
+  val chiselVersionString = BuildInfo.toString
+  println(chiselVersionString)
```
ducky64: Can we put the print as part of an actual function? And use the same logging / printing infrastructure as everything else? Or better yet, put it in the generated FIRRTL as a comment?

build.sbt
ucbjrl: Good points. These changes were due to the misunderstanding that we needed to export the generated BuildInfo with the source. We do want to record it as part of the release, but that can be done separately.

src/test/scala/chiselTests/BuildInfoTests.scala
ucbjrl: And this test is unnecessary. Removed.

src/main/scala/chisel3/Driver.scala
ucbjrl: Excellent suggestion. Done.

src/main/scala/chisel3/internal/firrtl/Emitter.scala
```diff
@@ -102,7 +102,7 @@ private class Emitter(circuit: Circuit) {
   private def unindent() { require(indentLevel > 0); indentLevel -= 1 }
   private def withIndent(f: => Unit) { indent(); f; unindent() }
 
-  private val res = new StringBuilder(s"circuit ${circuit.name} : ")
+  private val res = new StringBuilder(s";${Driver.chiselVersionString}\ncircuit ${circuit.name} : ")
```
ducky64: Consider putting the circuit part in a new line, so it doesn't look like it's part of the comment?
```
private val res = new StringBuilder(s";${Driver.chiselVersionString}\n")
res ++= "circuit ${circuit.name} : "
```

Some documentation, like http://www.scala-lang.org/docu/files/collections-api/collections_26.html also construct an empty StringBuilder instead and add everything afterwards, so this might be another option:
```
private val res = new StringBuilder
res ++= s";${Driver.chiselVersionString}\n"
res ++= "circuit ${circuit.name} : "
```
I like the latter, especially in terms of parallel structure.

ucbjrl: I concur.

ducky64:
lgtm

PR #315, opened 2016-10-05
Give <> and := legacy behavior in compatibility mode

aswaterman:
This makes <> and := do what they did two weeks ago (bidirectional connect with weak semantics for aggregates; unidirectional connect with weak semantics for elements), so that legacy code that imports Chisel._ will work seamlessly.

This means legacy code may emit incorrect FIRRTL, as was the case before; only new code that imports chisel3._ will be safe from this possibility.


aswaterman:
Done.

PR #316, opened 2016-10-05
Make asInput/asOutput/flip deprecation warnings dynamic

aswaterman:
Code that imports Chisel._ shouldn't see them.

Not sure if requireIOWrap is the right condition... or if cyan is a
good choice of color for deprecation warnings.


aswaterman:
OK, will do

aswaterman:
OK to merge now?

sdtwigg:
You also may be able to get away with adding the deprecate marking at the package that is aliasing this. Then, the old compatibility package just doesn't have the deprecation warning.

Issue #303, opened 2016-09-29
Better name propagation by macros, RFC


ducky64:
So, this has been a requested feature for a while (informally as well as formally, see #262). Basically, wire name propagation to FIRRTL right now only works for vals that are defined directly within a module (not inside a sub-scope, such as a if block, where it wouldn't be visible to reflection) and by explicit TransitName (which isn't widely used, and in my opinion has a somewhat clunky API and implementation).
## Basic Idea

The gist of this proposal is to use static annotations (a macro paradise feature) to rewrite each instance of
`val name = expr` into `val name = Namer(expr, name)`, where Namer is a passthrough function that sets the object's name field (if it has one, and does nothing otherwise).

Note that this is the same basic approach as used in @sdtwigg 's gama Chisel fork.
## Modules

The case for naming in Modules (and sub-blocks directly in Modules, like when or if blocks) is pretty simple, the transform above works. In cases of name aliasing, Chisel seems to be able to resolve it by adding a _1, _2, etc. One thing that might be nice is to have string interpolation into naming; for example:

```
for (i <- 1 to 10) {
  val myInt_i = i.asUInt
}
```

where i would be interpolated, so in FIRRTL you would get names like `myInt_1`, `myInt_2`, ..., `myInt_9`.
I'd like suggestions for a syntax for this though.

This would strictly be opt-in, users would need to annotate their modules with a static annotation, like `@module`:

```
@module
class myModule extends Module {
```

This also requires users to compile with the macro paradise plugin, though Chisel RTL that doesn't use the annotation would still continue to run fine without macro paradise.
## Functions

However, gets gets funner when dealing with propagating names to functions, since the usual pattern (I think, please correct me if I'm wrong) is:

```
val myQueue = Queue(blah)
```

where Queue is a function that instantiates a sub-module and might also contain some logic. But because the sub-module is instantiated in a different function, we would get the same generic name, q, that's defined in Queue(). That's obviously not helpful, and was the basis for the post-name hooks in TransitName. I'd like to replace that with macros too, hopefully making a syntactically consistent naming system that doesn't need changes in code.

_The rest of this section is more speculative and has not been backed up by a proof-of-concept implementation yet._

One possible way is to rewrite a function definition to take an implicit EnclosingName, and provide the implicit on val definitions:

```
object Queue {
  def apply(enq: ..., ...) {
```

into

```
object Queue {
  def apply(enq: ..., ...)(implicit enclosingName: EnclosingName) {
```

[...]

PR #327, opened 2016-10-18
Better name propagation by macros

ducky64:
Implementation for #303.
Doesn't use it anywhere (yet), so this only adds infrastructure. Hopefully also makes this less like to break things.
A (near) future PR would add this annotation to some core utils and replace TransitName.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -151,6 +151,7 @@ private[chisel3] class DynamicContext() {
   // Used to distinguish between no Module() wrapping, multiple wrappings, and rewrapping
   var readyForModuleConstr: Boolean = false
   val errors = new ErrorLog
+  val namingStack = new internal.naming.NamingStack
 }
 
 private[chisel3] object Builder {
@@ -163,6 +164,7 @@ private[chisel3] object Builder {
   def globalNamespace: Namespace = dynamicContext.globalNamespace
   def components: ArrayBuffer[Component] = dynamicContext.components
   def annotations: ArrayBuffer[ChiselAnnotation] = dynamicContext.annotations
+  def namingStack: internal.naming.NamingStack = dynamicContext.namingStack
 
   def currentModule: Option[Module] = dynamicContext.currentModule
   def currentModule_=(target: Option[Module]): Unit = {
@@ -221,4 +223,12 @@ private[chisel3] object Builder {
       Circuit(components.last.name, components, annotations.map(_.toFirrtl))
     }
   }
+}
+
+/** Allows public access to the naming stack in Builder / DynamicContext.
+  * Necessary because naming macros expand in user code and don't have access into private[chisel3]
+  * objects.
+  */
+object DynamicNamingStack {
+  def apply() = Builder.namingStack
 }
--- a/src/main/scala/chisel3/package.scala
+++ b/src/main/scala/chisel3/package.scala
@@ -285,5 +285,26 @@ package object chisel3 {    // scalastyle:ignore package.object.name
         */
       def range(args: Any*): (NumericBound[Int], NumericBound[Int]) = macro chisel3.internal.RangeTransform.apply
     }
+
+    import scala.language.experimental.macros
+    import scala.annotation.StaticAnnotation
+    import scala.annotation.compileTimeOnly
+
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class dump extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.DebugTransforms.dump
+    }
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class treedump extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.DebugTransforms.treedump
+    }
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class chiselName extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.chiselName
+    }
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class localName extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.localName
+    }
   }
 }
--- a/src/main/scala/chisel3/package.scala
+++ b/src/main/scala/chisel3/package.scala
@@ -302,9 +302,5 @@ package object chisel3 {    // scalastyle:ignore package.object.name
     class chiselName extends StaticAnnotation {
       def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.chiselName
     }
-    @compileTimeOnly("enable macro paradise to expand macro annotations")
-    class localName extends StaticAnnotation {
-      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.localName
-    }
   }
 }
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,8 @@ lazy val commonSettings = Seq (
   autoAPIMappings := true,
   scalaVersion := "2.11.7",
   scalacOptions := Seq("-deprecation", "-feature"),
+  libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
+  addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
   // Since we want to examine the classpath to determine if a dependency on firrtl is required,
   //  this has to be a Task setting.
   //  Fortunately, allDependencies is a Task Setting, so we can modify that.
@@ -78,7 +80,6 @@ lazy val chiselSettings = Seq (
 
   libraryDependencies ++= Seq(
     "org.scalatest" %% "scalatest" % "2.2.5" % "test",
-    "org.scala-lang" % "scala-reflect" % scalaVersion.value,
     "org.scalacheck" %% "scalacheck" % "1.12.4" % "test",
     "com.github.scopt" %% "scopt" % "3.4.0"
   ),
@@ -96,17 +97,11 @@ lazy val chiselSettings = Seq (
 
 lazy val coreMacros = (project in file("coreMacros")).
   settings(commonSettings: _*).
-  settings(
-    libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-    publishArtifact := false
-  )
+  settings(publishArtifact := false)
 
 lazy val chiselFrontend = (project in file("chiselFrontend")).
   settings(commonSettings: _*).
-  settings(
-    libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-    publishArtifact := false
-  ).
+  settings(publishArtifact := false).
   dependsOn(coreMacros)
 
 
--- a/chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
@@ -1,0 +1,130 @@
+// See LICENSE for license details.
+
+// This file contains part of the implementation of the naming static annotation system.
+
+package chisel3.internal.naming
+
+import scala.collection.mutable.Stack
+import scala.collection.mutable.ListBuffer
+
+import java.util.IdentityHashMap
+
+/** Recursive Function Namer overview
+  *
+  * In every function, creates a NamingContext object, which associates all vals with a string name
+  * suffix, for example:
+  *   val myValName = SomeStatement()
+  * produces the entry in items:
+  *   {ref of SomeStatement(), "myValName"}
+  *
+  * This is achieved with a macro transforming:
+  *   val myValName = SomeStatement()
+  * statements into a naming call:
+  *   val myValName = context.name(SomeStatement(), "myValName")
+  *
+  * The context is created from a global dynamic context stack at the beginning of each function.
+  * At the end of each function call, the completed context is added to its parent context and
+  * associated with the return value (whose name at an enclosing function call will form the prefix
+  * for all named objects).
+  *
+  * When the naming context prefix is given, it will name all of its items with the prefix and the
+  * associated suffix name. Then, it will check its descendants for sub-contexts with references
+  * matching the item reference, and if there is a match, it will (recursively) give the
+  * sub-context a prefix of its current prefix plus the item reference suffix.
+  *
+  * Note that for Modules, the macro will insert a naming context prefix call with an empty prefix,
+  * starting the recursive naming process.
+  */
+
+/** Base class for naming contexts, providing the basic API consisting of naming calls and
+  * ability to take descendant naming contexts.
+  */
+class NamingContext {
+  val descendants = new IdentityHashMap[AnyRef, NamingContext]()
+  val items = ListBuffer[(AnyRef, String)]()
+  var closed = false  // a sanity check to ensure no more name() calls are done after name_prefix
+
+  /** Adds a NamingContext object as a descendant - where its contained objects will have names
+    * prefixed with the name given to the reference object, if the reference object is named in the
+    * scope of this context.
+    */
+  def add_descendant(ref: AnyRef, descendant: NamingContext) {
+    // First set takes effect, subsequent ones are discarded
+    // TODO: is this the expected behavior?
+    if (!descendants.containsKey(ref)) {
+      descendants.put(ref, descendant)
+    }
+  }
+
+  /** Suggest a name (that will be propagated to FIRRTL) for an object, then returns the object
+    * itself (so this can be inserted transparently anywhere).
+    * Is a no-op (so safe) when applied on objects that aren't named, including non-Chisel data
+    * types.
+    */
+  def name[T](obj: T, name: String): T = {
+    assert(!closed, "Can't name elements after name_prefix called")
+    obj match {
+      case ref: AnyRef => items += ((ref, name))
+      case _ =>
+    }
+    obj
+  }
+
+  /** Gives this context a naming prefix (which may be empty, "", for a top-level Module context)
+    * so that actual naming calls (HasId.suggestName) can happen.
+    * Recursively names descendants, for those whose return value have an associated name.
+    */
+  def name_prefix(prefix: String) {
+    closed = true
+    for ((ref, suffix) <- items) {
+      // First name the top-level object
+      ref match {
+        case nameable: chisel3.internal.HasId => nameable.suggestName(prefix + suffix)
+        case _ =>
+      }
+
+      // Then recurse into descendant contexts
+      if (descendants.containsKey(ref)) {
+        descendants.get(ref).name_prefix(prefix + suffix + "_")
+      }
+    }
+  }
+}
+
+/** Class for the (global) naming stack object, which provides a way to push and pop naming
+  * contexts as functions are called / finished.
+  */
+class NamingStack {
+  val naming_stack = Stack[NamingContext]()
+
+  /** Creates a new naming context, where all items in the context will have their names prefixed
+    * with some yet-to-be-determined prefix from object names in an enclosing scope.
+    */
+  def push_context(): NamingContext = {
+    val context = new NamingContext
+    naming_stack.push(context)
+    context
+  }
+
+  /** Called at the end of a function, popping the current naming context, adding it to the
+    * enclosing context's descendants, and passing through the prefix naming reference.
+    * Every instance of push_context() must have a matching pop_context().
+    *
+    * Will assert out if the context being popped isn't the topmost on the stack.
+    */
+  def pop_return_context[T <: AnyRef](prefix_ref: T, until: NamingContext): T = {
+    assert(naming_stack.top == until)
+    naming_stack.pop()
+    if (!naming_stack.isEmpty) {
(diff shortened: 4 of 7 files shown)


ucbjrl:
retest this please

ucbjrl:
retest this please

ucbjrl:
retest this please

colinschmidt:
Is this now supposed to work with Chisel._ and not just chisel3._ ?

ducky64:
Yes, in theory.

ducky64:
Fixed the build.sbt, moved compiler plugins and the scala.lang.reflect library deps into commonSettings.

ducky64:
@sdtwigg poke

colinschmidt:
rocket-chip uses chisel3/224ee397db324067b35874bf22c76a63e9ca531b
and just building with this branch on rocket-chip/master results in the issue seen in https://github.com/ucb-bar/rocket-chip/pull/423

So maybe rebase this onto a more recent chisel3?

ducky64:
Rebased to chisel3/224ee397db324067b35874bf22c76a63e9ca531b

colinschmidt:
Thanks that solved this issue in rocket-chip master.

colinschmidt:
So I checked this out for the if/when naming and that seems to work well, but I haven't been able to get the function annotations working yet.

My chisel code looks like:

This module 
```
@chiselName
class IBox(implicit p: Parameters) extends VMUModule()(p) {
 ...
  val op = VMUDecodedOp(opq.io.deq.bits)
```
calls this function
```
  @chiselName
  def apply(op: VMUOp)(implicit p: Parameters): VMUDecodedOp = {
    val dec = Wire(new VMUDecodedOp)
...
    dec.mt := DecodedMemType(op.fn.mt)
```
which calls this function
```
object DecodedMemType {
  @chiselName
  def apply[T <: UInt](mt: T): DecodedMemType = {
    val b = (mt === MT_B)
    val h = (mt === MT_H)
    val w = (mt === MT_W)
    val d = (mt === MT_D)
    val bu = (mt === MT_BU)
    val hu = (mt === MT_HU)
    val wu = (mt === MT_WU)

    val dec = Wire(new DecodedMemType)
    dec.b := (b || bu)
    dec.h := (h || hu)
    dec.w := (w || wu)
    dec.d := d
    dec.signed := ((b || h) || (w || d))
    dec
  } 
} 
```
I was expecting all of the val b, val h, etc. to be named in the firrtl but the output firrtl looks like:
```
    node T_2329 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h00")) @[types-vmu.scala 71:17]                 
    node T_2330 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h01")) @[types-vmu.scala 72:17] 
    node T_2331 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h02")) @[types-vmu.scala 73:17]                 
    node T_2332 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h03")) @[types-vmu.scala 74:17]
    node T_2333 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h04")) @[types-vmu.scala 75:18]
    node T_2334 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h05")) @[types-vmu.scala 76:18]
    node T_2335 = eq(opqInst.io.deq.bits.fn.mt, UInt<3>("h06")) @[types-vmu.scala 77:18]                 
    wire T_2342 : {b : UInt<1>, h : UInt<1>, w : UInt<1>, d : UInt<1>, signed : UInt<1>} @[types-vmu.scala 79:19]
    T_2342 is invalid @[types-vmu.scala 79:19]
    node T_2348 = or(T_2329, T_2333) @[types-vmu.scala 80:17]
    T_2342.b <= T_2348 @[types-vmu.scala 80:11]
    node T_2349 = or(T_2330, T_2334) @[types-vmu.scala 81:17]
    T_2342.h <= T_2349 @[types-vmu.scala 81:11]
    node T_2350 = or(T_2331, T_2335) @[types-vmu.scala 82:17]
    T_2342.w <= T_2350 @[types-vmu.scala 82:11]
    T_2342.d <= T_2332 @[types-vmu.scala 83:11]
    node T_2351 = or(T_2329, T_2330) @[types-vmu.scala 84:23]
    node T_2352 = or(T_2331, T_2332) @[types-vmu.scala 84:35]
    node T_2353 = or(T_2351, T_2352) @[types-vmu.scala 84:29]                                            
    T_2342.signed <= T_2353 @[types-vmu.scala 84:16]                                                     
    op.mt <- T_2342 @[vmu.scala 80:12]
```

ducky64:
So the name propagation breaks here:
```
  @chiselName
  def apply(op: VMUOp)(implicit p: Parameters): VMUDecodedOp = {
    val dec = Wire(new VMUDecodedOp)
    ...
    dec.mt := DecodedMemType(op.fn.mt)
```
Because `dec.mt := DecodedMemType(op.fn.mt)` is not in the form of `val name = blah`, the naming annotation doesn't recognize it and is unable to provide a prefix for the contents of DecodedMemType.

In general, handling these cases is hard since the assignment is conditional and may only happen in some circumstances. Currently, those are just ignored as far as naming is concerned.

It may be possible to handle IO assignments that are static (top level scope, only one assignment) as a special case, since assigning it to a val, then assigning the val to a IO is pretty cumbersome. Though it's also probably very hard to detect all cases of reassignment by traversing the AST, so there will be edge cases. Thoughts?

sdtwigg:
Why doesn't it just default to calling the signals b, h, bu, hu?

Also, the original proposal involved two macros: name and propagateName so the user could decide if you wanted the function to just name things directly versus  naming things based on names applied to the return. (Although as noted above, even for the latter, it should still default to applying the proposed suffix.)

Similarly, if the user manually calls suggestName on the result of DecodedMemType (before assignment), will it propagate the suggested name into the internal function nodes?

ducky64:
For directly naming signals in functions without a prefix: it can be problematic if the function is called multiple times in the same module, then there would be a conflict. I also think if we wanted to specify that things be named without a prefix, that would have to happen in the calling scope rather than the function being called, since (in the general case) when you're writing the function you can't know ahead of time if it's going to be used in a situation that will cause a conflict.

suggestName won't work, this doesn't use the TransitName facility. I still think the postname hooks directly in Module is inappropriate coupling and very special-case functionality. The two options are for fixing this are:
- Define it as a val and use the resulting val. This requires two lines, so the resulting code is less elegant.
- The internal naming functions could be exposed and documented as part of the API. This can be an "escape hatch" that allows names and propagation in special cases to be handled. However, the syntax would still be bulky and could mean duplication of information (for example, the above problem might be handled with `dec.mt := Named(DecodedMemType(op.fn.mt), "dec.mt")`, but writing `dec.mt` twice is undesirable.
- Maybe a better smarter solution that I haven't thought of yet?

ducky64:
Moved chiselName annotations to chisel3.experimental and Chisel.experimental, rebase to latest.

ducky64:
Major changes:
- Added a `@blindName` for functions. This does not need anything to be returned to function, and is recursive. No annotation chain is needed to the Module level.
- `@chiselName` still behaves the same for methods (hierarchical naming if a return value is given and used by the calling function, assuming there are `@chiselName` annotations all the way up the function call stack to the Module level.
- `@chiselName` applies `@blindName` transforms for inner methods, unless they are explicitly annotated by `@chiselName`.
- Added naming collision tests. In cases of conflict, subsequent names have the _(number) attached.
- I'm going to hold off on exposing a `Named(...)` construct because of potential interplay issues with the macro-generated hierarchical naming context system.

@sdtwigg

sdtwigg:
A tempting thing to suggest is that, instead of blindName, call the second annotation "suggestNames" to imply it is equivalent to calling suggestName on (almost all) nested vals.

ducky64:
So the goal of blindName was to make it clear that it doesn't respect the call hierarchy and it would have higher potential for undesirable behavior (like collisions). Users should think before applying this over chiselName.

I think suggestName is pretty general; if anything it would make sense to replace chiselName (which is somewhat content-free) with suggestName.

ducky64:
Our resolution is to change blindName to localName, and to leave chiselName as is. Implementation behavior will not change. Changes pending...

ducky64:
Rename done, review ready.

azidar:
@jackkoenig please review.

ducky64:
Bump.

This PR is now 3 months old...

src/main/scala/chisel3/package.scala
```diff
@@ -281,5 +281,26 @@ package object chisel3 {    // scalastyle:ignore package.object.name
...
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.DebugTransforms.treedump
+    }
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class chiselName extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.chiselName
+    }
+    @compileTimeOnly("enable macro paradise to expand macro annotations")
+    class localName extends StaticAnnotation {
+      def macroTransform(annottees: Any*): Any = macro chisel3.internal.naming.NamingTransforms.localName
+    }
```
jackkoenig: For most of the stuff in these packages, it's just a type alias, should that not be the case here? Regardless, since these are the actual annotations, I think the ScalaDoc explaining them should live here. It's not clear looking at the ScalaDoc generated by this branch where I would find a description of chiselName or localName

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -222,3 +224,11 @@ private[chisel3] object Builder {
...
   }
 }
+
+/** Allows public access to the naming stack in Builder / DynamicContext.
+  * Necessary because naming macros expand in user code and don't have access into private[chisel3]
+  * objects.
+  */
+object DynamicNamingStack {
+  def apply() = Builder.namingStack
+}
```
ducky64: Macros expand such that this is generated into annotated functions, so it can't be private. I also know of no way to suppress warnings.

jackkoenig:
So, this works:

```scala
@chiselName
class MySimpleModule extends Module {
  def addOne(in: UInt): UInt = {
    val inc = Wire(in)
    inc := in +& 1.U
    inc
  }

  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := addOne(io.in)
}
```

And this works:

```scala
object Utils {
  @localName
  def addOne(in: UInt): UInt = {
    val inc = Wire(in)
    inc := in +& 1.U
    inc
  }
}

class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := Utils.addOne(io.in)
}
```

But this errors:

```scala
@chiselName
object Utils {
  def addOne(in: UInt): UInt = {
    val inc = Wire(in)
    inc := in +& 1.U
    inc
  }
}

class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := Utils.addOne(io.in)
}
```

error message:
```
[error] /Users/jack.koenig3/ucb-bar/chisel3-nameprop/src/test/scala/chiselTests/NamingAnnotationTest.scala:34: @chiselName annotation did not match exactly one valid tree, got:
[error] object Utils {
[error]   def addOne(in: UInt): UInt = {
[error]     val inc = Wire(in);
[error]     inc.:=(in.+&(1.U));
[error]     inc
[error]   }
[error] }
[error] @chiselName
[error]  ^
```

(and if I just annotate the function `addOne` with `@chiselName`, nothing happens)

Is annotating objects with `@chiselName` supposed to work?

ducky64:
Annotating objects is not supposed to work, you should explicitly annotate each function within.

Unlike on Modules, where you can guess that inner functions are probably helper functions that are called once, I don't think the same applies on general objects. So I prefer to leave it for the user to specify each method explicitly.

That being said, the error message could be more helpful (from a "what you did wrong" rather than "what the compiler saw that's wrong" standpoint)

jackkoenig:
Okay fair enough, yeah I'm not sure how easy it is to add a better error message to the macro but if possible that would be helpful.

Still possibly an issue?:
If I annotate just the function with `@chiselName` it doesn't do anything:

```scala
object Utils {
  @chiselName
  def addOne(in: UInt): UInt = {
    val inc = Wire(in)
    inc := in +& 1.U
    inc
  }
}

class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := Utils.addOne(io.in)
}
```

Gives
```
circuit MySimpleModule :
  module MySimpleModule :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : UInt<32>, out : UInt<32>}

    io is invalid
    wire _T_6 : UInt<32>
    _T_6 is invalid
    node _T_8 = add(io.in, UInt<1>("h01"))
    _T_6 <= _T_8
    io.out <= _T_6
```

jackkoenig:
By Module methods being "helper functions called once" do you mean *at least* once or *exactly* once?

ducky64:
For your example: you also need to annotate your top-level module. chiselName annotations must be unbroken in the call chain to work properly. 

Helper functions called once: no convention is enforced. The idea of localName is that functions aren't called more than once in the same scope so there's less ambiguity. But if they are, it will just resolve with the default name disambiguation: `_1`, `_2`, ...

jackkoenig:
```scala
object Utils {
  @chiselName
  def addOne(in: UInt): UInt = {
    val inc = Wire(in)
    inc := in +& 1.U
    inc
  }
}

@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := Utils.addOne(io.in)
}

object Test extends App {
  chisel3.Driver.execute(args, () => new MySimpleModule)
}
```

So in this case, `MySimpleModule` is the top-level and it still doesn't seem to be working. Anything else I'm missing?

ducky64:
So it also only recognizes direct val assign (`val myNamedThingy = something`) statements, it doesn't recognize Chisel assignment. The reasoning for this is that Chisel assignments aren't final, so in this case I can't infer that the result of Utils.addOne(...) will be io.out in all cases.

jackkoenig:
The idea was the `inc` would get propagated, which it does with `@localName` or if `addOne` is a method on the Module class, but does not work in the above example with `@chiselName`

ducky64:
I don't think it makes sense to infer a prefix when a complete chain doesn't exist for chiselName, since the idea of chiselName is that you want that prefix. For example, if you called addOne many times in your module, then you would just get spammed with incs trailed by numbers.

sdtwigg:
A  series of inc_1, inc_2, inc_3 would definitely be preferable to W25, W35, W45.

ducky64:
I see your point, though missing the context is also undesirable.

A possible change would be to have chiselName generate the longest contiguous prefix (from named node) of chiselName chained functions. This would also eliminate localName (which is equivalent to chiselName with no enclosing elements in the chain). The trade-off here is that it would generate more helpful, but possibly less consistent names that may change as code is moved around. One of the hopes was the chiselName was strict enough that you could use the name in downstream tools (for example, constraining module placement) and it would remain consistent through some types of refactoring (for example, moving blocks of code around, where the numerical suffix could change).

What are everyone else's thoughts?

jackkoenig:
I think I'm finally starting to understand chiselName. So in the following:

```scala
object Utils {
  @chiselName
  def addTwo(in: UInt): UInt = {
    val inc = in +& 1.U
    inc +& 1.U
  }
}

@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  val sum = Utils.addTwo(io.in)
  io.out := sum
}
```
I get the following Firrtl
```
circuit MySimpleModule :
  module MySimpleModule :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : UInt<32>, out : UInt<32>}

    io is invalid
    node sum_inc = add(io.in, UInt<1>("h01")) 
    node sum = add(sum_inc, UInt<1>("h01")) 
    io.out <= sum 
```
Pretty cool. I guess the thing that concerns me is that if I instead connect to the result of `addTwo` directly, ie.

```scala
  io.out = Utils.addTwo(io.in)
```

I all of the sudden lose everything:

```
    node _T_7 = add(io.in, UInt<1>("h01"))
    node _T_9 = add(_T_7, UInt<1>("h01"))
    io.out <= _T_9
```

I see what you mean about trying to provide guarantees, but it seems strange to me that everything should just get thrown away if it's not a perfect line all the way to the top. As a library writer, I would like for people to get decent names even if they're unwilling or unable to write the right code to create these chains. I think your idea of providing the longest unbroken chain is a good one, since you can have fairly sophisticated libraries generating large chains with nice names and yet have a user invoke the library in a bad way. At a bare minimum, this would make the naming so much more stable that I think the benefit there is worth the possibility that people may use libraries wrong and not get perfect naming guarantees.

ducky64:
How about this proposal:
- Make chiselName generate names according to longest unbroken chain
- Eliminate localName (it's redundant)
- No guarantees at this point in time, it can be looked into when there's a use case
  - (+) Practically users have little control over library writers and whether they provide a unbroken chiselName. The guarantee semantics may needs to be encoded elsewhere (not in naming annotation in libraries)
  - (-) If guarantees require changing the naming annotation that becomes more difficult if everyone uses chiselName.

jackkoenig:
I think this proposal makes a lot of sense

colinschmidt:
I think the name guarantee idea can be handled two ways. Either you can provide some additional annotation that hardcodes a name and errors during elaboration if it can't get that name. Or you could provide some sort of firrtl annotation that the user can consume in their backend.

Otherwise I like the elimination of the second macro and the streamlining this proposal seems to add.

ducky64:
Updated. localName eliminated. chiselName does best-effort (and will drop non-present links in the chain, instead of maximal contiguous chain, since some prefix is probably better than no prefix).

Note that the chiselName annotation still must be applied at the top level. Eventually, it might be rolled into one of the things that happens on a Module close, but chiselName is purely opt-in for now (in case something goes wrong).

jackkoenig:
This looks good. With the recent changes, would it make sense to allow chiselName on objects as well so that you could annotate just a utilities object and not have to do it on every single function? It could also annotate any modules in side for you as well. Would this work?

jackkoenig:
Also, this works on `if` blocks (thanks!!!!) but doesn't seem to work on for loops:

```scala
// This does NOT work!
@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })

  for (i <- 0 until 1) {
    val a = Wire(init = io.in)
    val b = Wire(init = a)
    io.out := b
  }
}

// This does work!
@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })

  if (true) {
    val a = Wire(init = io.in)
    val b = Wire(init = a)
    io.out := b
  } else {
    io.out := func(io.in)
  }
}

```

jackkoenig:
Does not work on anonymous functions either, for example:

```scala
@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })

  Seq(io.in).foreach { in =>
    val a = Wire(init = in)
    val b = Wire(init = a)
    io.out := b
  }
}
```

If supporting all of these cases is extremely difficult then obviously we can move ahead without them, just wanted to check.

sdtwigg:
scala desugars for loops into a bunch of function calls with the inner body inside the parameters so I wonder if `case q"(..$params) => $expr" => tree` is blocking that. (Interestingly, even though it does that desugaring really early in the compile process, quasiquotes still has matchers for 'for' and 'for-yield'. I wonder if it only intends for you to use those to build, not match, trees.)

jackkoenig:
Just for fun, it **does** work with match:

```scala
@chiselName
class MySimpleModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })

  io.in match {
    case in =>
      val a = Wire(init = in)
      val b = Wire(init = a)
      io.out := b
  }
}
```

ducky64:
Yeah, so the for loop is turning into (using the new `@dump` debugging annotation)
```scala
   0.until(1).foreach(((i) => {
     val a = Wire(init = io.in);
     val b = Wire(init = a);
     io.out.$colon$eq(b)
   }));
```
which for some reason isn't matching. I'll look into it. 

As far as recursively naming inner functions / objects / whatever: it's possible (and not even difficult to implement), but I'm not sure whether it's a good idea or not since it might be desirable to limit the scope of the annotation. Theoretically, it should be harmless (performance aside, since it is executing more code) when applied on non-Chisel code. Thoughts?

ducky64:
For some reason I wrote a rule preventing it from recursing into anonymous functions. That's gone now, so that and the for case should work.
It will be likely to produce name collisions though, but a future change may be to introduce some kind of variable interpolation syntax into the variable name, like `a_i` which would turn into `a_0`, `a_1`

jackkoenig:
I guess it would be nice to have a test for for loops and anonymous functions, but yeah this looks great. I'm fine with the decision to not do the recursive macro applying, seems reasonable enough. People might request it in the future but we can cross that bridge if/when we come to it!

ducky64:
Tests added! for loop test is pretty basic because it's difficult to get the expected disambiguated name in without significant plumbing, but the anonymous function test is more robust.

jackkoenig:
Looks great!

PR #328, opened 2016-10-19
FixedPoint number support for chisel3

chick:
FixedPoint numbers have a width and a binary position
Either, neither or both maybe inferred.
Firrtl will convert these to SInts during lowering passes
These changes are necessary for support of the DSP project

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -250,10 +250,11 @@ sealed abstract class Bits(width: Width, override val litArg: Option[LitArg])
 
   def do_asSInt(implicit sourceInfo: SourceInfo): SInt
 
-  /** Reinterpret cast to a SInt.
+  /** Reinterpret cast as a FixedPoint.
     *
     * @note value not guaranteed to be preserved: for example, an UInt of width
-    * 3 and value 7 (0b111) would become a SInt with value -1
+    * 3 and value 7 (0b111) would become a FixedInt with value -1, the interpretation
+    * of the number is also affected by the specified binary point.  Caution advised
     */
   final def asFixedPoint(that: BinaryPoint): FixedPoint = macro SourceInfoTransform.thatArg
 
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -876,6 +876,19 @@ object Mux {
   }
 }
 
+//scalastyle:off number.of.methods
+/**
+  * A sealed class representing a fixed point number that has a bit width and a binary point
+  * The width and binary point may be inferred.
+  *
+  * IMPORTANT: The API provided here is experimental and may change in the future.
+  *
+  * @param width       bit width of the fixed point number
+  * @param binaryPoint the position of the binary point with respect to the right most bit of the width
+  *                    currently this should be positive but it is hoped to soon support negative points
+  *                    and thus use this field as a simple exponent
+  * @param lit
+  */
 sealed class FixedPoint private (width: Width, val binaryPoint: BinaryPoint, lit: Option[FPLit] = None)
     extends Bits(width, lit) with Num[FixedPoint] {
   private[core] override def cloneTypeWidth(w: Width): this.type =
@@ -999,6 +1012,10 @@ sealed class FixedPoint private (width: Width, val binaryPoint: BinaryPoint, lit
 sealed trait PrivateType
 private case object PrivateObject extends PrivateType
 
+/**
+  * Factory and convenience methods for the FixedPoint class
+  * IMPORTANT: The API provided here is experimental and may change in the future.
+  */
 object FixedPoint {
   /** Create an FixedPoint type with inferred width. */
   def apply(): FixedPoint = apply(Width(), BinaryPoint())
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -250,6 +250,17 @@ sealed abstract class Bits(width: Width, override val litArg: Option[LitArg])
 
   def do_asSInt(implicit sourceInfo: SourceInfo): SInt
 
+  /** Reinterpret cast to a SInt.
+    *
+    * @note value not guaranteed to be preserved: for example, an UInt of width
+    * 3 and value 7 (0b111) would become a SInt with value -1
+    */
+  final def asFixedPoint(that: BinaryPoint): FixedPoint = macro SourceInfoTransform.thatArg
+
+  def do_asFixedPoint(that: BinaryPoint)(implicit sourceInfo: SourceInfo): FixedPoint = {
+    throwException(s"Cannot call .asFixedPoint on $this")
+  }
+
   /** Reinterpret cast to Bits. */
   @deprecated("Use asUInt, which does the same thing but returns a more concrete type", "chisel3")
   final def asBits(): Bits = macro SourceInfoTransform.noArg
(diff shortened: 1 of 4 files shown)


chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -853,3 +876,205 @@ object Mux {
...
+//  def Lit(value: BigInt, width: Int): SInt = Lit(value, Width(width))
+//  /** Create an SInt literal with specified width. */
+//  def Lit(value: BigInt, width: Width): SInt = {
+//
+//    val lit = SLit(value, width)
+//    val result = new SInt(lit.width, Some(lit))
+//    // Bind result to being an Literal
+//    result.binding = LitBinding()
+//    result
+//  }
```
jackkoenig: Im guessing these comments shouldn't be here?

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -853,3 +876,205 @@ object Mux {
...
+  /**
+    * How to create a bigint from a double with a specific binaryPoint
+    * @param x               a double value
+    * @param binaryPoint     a binaryPoint that you would like to use
+    * @return
+    */
+  def toBigInt(x: Double, binaryPoint    : Int): BigInt = {
+    val multiplier = math.pow(2,binaryPoint    )
+    val result = BigInt(math.round(x * multiplier))
+    // println(s"toBigInt:x = $x, width = $binaryPoint     multiplier $multiplier result $result")
```
jackkoenig: delete comment

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -853,3 +876,205 @@ object Mux {
...
+  /**
+    * converts a bigInt with the given binaryPoint into the double representation
+    * @param i            a bigint
+    * @param binaryPoint  the implied binaryPoint of @i
+    * @return
+    */
+  def toDouble(i: BigInt, binaryPoint    : Int): Double = {
+    val multiplier = math.pow(2,binaryPoint    )
+    val result = i.toDouble / multiplier
+    // println(s"toDouble:i = $i, fw = $binaryPoint    , multiplier = $multiplier, result $result")
```
jackkoenig: delete comment

chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -136,6 +146,43 @@ sealed case class KnownWidth(value: Int) extends Width {
...
+  type W = Int
+  def max(that: BinaryPoint): BinaryPoint = this.op(that, _ max _)
+  def + (that: BinaryPoint): BinaryPoint = this.op(that, _ + _)
+  def + (that: Int): BinaryPoint = this.op(this, (a, b) => a + that)
+  def shiftRight(that: Int): BinaryPoint = this.op(this, (a, b) => 0 max (a - that))
+  def dynamicShiftLeft(that: BinaryPoint): BinaryPoint =
+    this.op(that, (a, b) => a + (1 << b) - 1)
+
+  def known: Boolean
+  def get: W
```
jackkoenig: In my opinion, using the API of `if (bp.known) bp.get` is an antipattern compared to just returning `Option[W]`. Chisel provides the `.isWidthKnown` and `.getWidth` but I think these functions are historical more than desirable.

chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -136,6 +146,43 @@ sealed case class KnownWidth(value: Int) extends Width {
...
+  protected def op(that: BinaryPoint, f: (W, W) => W): BinaryPoint
+}
+
+sealed case class UnknownBinaryPoint() extends BinaryPoint {
```
jackkoenig: why case class instead of case object?

src/test/scala/chiselTests/FixedPointSpec.scala
```diff
@@ -0,0 +1,46 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import chisel3.testers.BasicTester
+import org.scalatest._
+
+//scalastyle:off magic.number
+
```
jackkoenig: Is this scalastyle comment used? If so I guess there shouldn't be a line between it and the class?

src/test/scala/chiselTests/FixedPointSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  assert(dut.io.out === FixedPoint.fromDouble(3.0, binaryPoint = 0))
+
+  stop()
+}
+class SBPSpec extends ChiselPropSpec {
+
+  property("should allow set binary point") {
+    //    val c = Driver.elaborate(() => new SBPTester)
+    //    val e = Driver.emit(c)
+    //    println(e)
```
jackkoenig: Delete comments

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -853,3 +876,205 @@ object Mux {
...
+  private[chisel3] def toType = s"Fixed$width$binaryPoint"
+
+  def := (that: Data)(implicit sourceInfo: SourceInfo): Unit = that match {
+    case _: FixedPoint => this connect that
+    case _ => this badConnect that
+  }
+
+  private[chisel3] def fromInt(value: BigInt, width: Int): this.type = {
+    throwException(s"Don't use $this.fromInt($value, $width): Use literal constructors instead")
+    //FixedPoint.fromBigInt(value, width, 0).asInstanceOf[this.type]
```
jackkoenig: Delete comment?

chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
chick: Hard to argue with you, but all the other classes in IR carry these methods.  I'd prefer to leave them in for symmetry.  Although I don't see any usages of these particular methods, they are in moderate use for other classes that have them

chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
chick: Hard to argue with you, but all the other classes in IR carry these methods. I'd prefer to leave them in for symmetry. Although I don't see any usages of these particular methods, they are in moderate use for other classes that have them.  If you have strong feelings I can remove them

chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
chick: done

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
chick: done

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
chick: done

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
chick: done

src/test/scala/chiselTests/FixedPointSpec.scala
chick: Done, though appearing where it does it is for entire file, not somehow linked to the class.  Magic numbers should not be considered anyway in test code, but IntelliJ doesn't seem to see the scalastyle-test-config.xml which does not carry this constraint, and since I rely on IntelliJ to do my style analysis, it is worth including for me

src/test/scala/chiselTests/FixedPointSpec.scala
chick: Done, I'm little torn on removing some of these.  Sometimes when things change I find that adebug line that I but some thought into in a previous debugging session, is useful to uncomment.  Ideally there would be some special comment type or annotation that could be hidden by default.

aswaterman:
Can we mark the APIs as experimental?  I suspect that as more people use fixed point, we'll want to evolve the API.

jackkoenig:
+1 to marking as experimental

@aswaterman Is there a standard way to do this? I can't seem to find any "deprecated-like" annotation. Do we just put "EXPERIMENTAL" at the top of the Scaladoc or something?

aswaterman:
I don't know... @deprecated("This experimental API may change") might be the way to go

ucbjrl:
lgtm - passes jenkins.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -250,6 +250,17 @@ sealed abstract class Bits(width: Width, override val litArg: Option[LitArg])
 
   def do_asSInt(implicit sourceInfo: SourceInfo): SInt
 
+  /** Reinterpret cast to a SInt.
```
azidar: Fix comment

Issue #331, opened 2016-10-19
Use verilator VPI interface


ucbjrl:
Predicting verilator's signal names is increasingly problematic. #329 (compiling with -O0) breaks the existing harness communication due to signal renaming.

jackkoenig:
Which harness breaks? The chisel3 tests pass using their testing harness, is it a problem in https://github.com/ucb-bar/chisel-testers?

donggyukim:
We should move this to chisel-tester.

wallento:
Hi, actually I think you want to use DPI for it instead of VPI.

donggyukim:
@wallento Can we access internal signals with DPI?

wallento:
Yes, you can. It depends a bit on what you want to achieve. Do I understand it correctly it is driving signals in a testbench? Then DPI works fine, VPI has the only advantage that you can mess in the entire hierarchy. DPI instead is the suggested way for ease of use and clarity from my point of view.

wallento:
If you want to randomly access all signals than VPI (or plain C++) and `/* verilator public */` are indeed needed.

oharboe:
I created a Perl post-processing hack to inject ```/* verilator public */``` after input/output signals that I need to reach from C++. This hack broke with 3.2.0, because the Verilog is very slightly different in 3.2.0.

It would be great if there was a robust way to specify that should ```/* verilator public */``` follow input or output signals to a module in the generated Verilog.

seldridge:
@oharboe: I don't have an example of doing this, but the [`DescriptionAnnotation`](https://www.chisel-lang.org/api/firrtl/latest/firrtl/DescriptionAnnotation.html) is likely the stable way of doing this coupled with https://github.com/freechipsproject/firrtl/pull/1172. The attributes you're talking about are generally needed...

wallento:
just a quick update on the verilator side: There will soon be support to specify public signals to be accessed via VPI in a separate file. Cheers!

jackkoenig:
Thanks for your hard work on this @wallento 🙂

PR #332, opened 2016-10-19
Try -O1 instead of -O0 for verilator cpp compiles - #329, #331.

ucbjrl:
O0 seems to cause too many changes to the generated .{h,cpp} files, breaking the test harness.
Once #331 is addressed, this commit can be reverted.

### DIFF
--- a/src/main/scala/chisel3/Driver.scala
+++ b/src/main/scala/chisel3/Driver.scala
@@ -119,13 +119,13 @@ trait BackendCompilationUtilities {
         "-Wno-WIDTH",
         "-Wno-STMTDLY",
         "--trace",
-        "-O0",
+        "-O1",
         "--top-module", topModule,
         "+define+TOP_TYPE=V" + dutFile,
         s"+define+PRINTF_COND=!$topModule.reset",
         s"+define+STOP_COND=!$topModule.reset",
         "-CFLAGS",
-        s"""-Wno-undefined-bool-conversion -O0 -DTOP_TYPE=V$dutFile -include V$dutFile.h""",
+        s"""-Wno-undefined-bool-conversion -O1 -DTOP_TYPE=V$dutFile -include V$dutFile.h""",
         "-Mdir", dir.toString,
         "--exe", cppHarness.toString)
     System.out.println(s"${command.mkString(" ")}") // scalastyle:ignore regex


jackkoenig:
I don't think this is the right way to do this. I've been looking into the chisel-testers RealGCD2 failure and as far as I can tell, I think it's purely coincidental that the test has been working at all.

I looked at the VRealGCD.h generated with O0 vs O1 and here's the important part of the diff

With O1

```cpp
VL_MODULE(VRealGCD2) {
  public:
    // CELLS
    // Public to allow access to /*verilator_public*/ items;
    // otherwise the application code can consider these internals.
    
    // PORTS
    // The application code writes and reads these signals to
    // propagate new values into/out from the Verilated model.
    VL_IN8(clock,0,0);
    VL_IN8(reset,0,0);
    VL_OUT8(io_in_ready,0,0);
    VL_IN8(io_in_valid,0,0);
    VL_OUT8(io_out_valid,0,0);
    //char  __VpadToAlign5[1];
    VL_IN16(io_in_bits_a,15,0);
    VL_IN16(io_in_bits_b,15,0);
    VL_OUT16(io_out_bits,15,0);
    
    // LOCAL SIGNALS
    // Internals; generally not touched by application code
    VL_SIG8(RealGCD2__DOT__p,0,0);
    VL_SIG8(RealGCD2__DOT__T_44,0,0);
    VL_SIG8(RealGCD2__DOT__T_46,0,0);
    //char  __VpadToAlign19[1];
    VL_SIG16(RealGCD2__DOT__x,15,0);
    VL_SIG16(RealGCD2__DOT__y,15,0);
    VL_SIG16(RealGCD2__DOT__ti,15,0);
    VL_SIG16(RealGCD2__DOT__T_39,15,0);
    VL_SIG16(RealGCD2__DOT__GEN_0,15,0);
    VL_SIG16(RealGCD2__DOT__GEN_1,15,0);
    VL_SIG(RealGCD2__DOT__GEN_9,31,0);
    VL_SIG(RealGCD2__DOT__GEN_10,31,0);
    VL_SIG(RealGCD2__DOT__GEN_11,31,0);
    VL_SIG(RealGCD2__DOT__GEN_12,31,0);
    VL_SIG(RealGCD2__DOT__T_49,16,0);
```

with O0

```cpp
VL_MODULE(VRealGCD2) {
  public:
    // CELLS
    // Public to allow access to /*verilator_public*/ items;
    // otherwise the application code can consider these internals.
    VRealGCD2_RealGCD2* __PVT__RealGCD2;

    // PORTS
    // The application code writes and reads these signals to
    // propagate new values into/out from the Verilated model.
    VL_IN8(clock,0,0);
    VL_IN8(reset,0,0);
    VL_OUT8(io_in_ready,0,0);
    VL_IN8(io_in_valid,0,0);
    VL_OUT8(io_out_valid,0,0);
    //char  __VpadToAlign5[1];
    VL_IN16(io_in_bits_a,15,0);
    VL_IN16(io_in_bits_b,15,0);
    VL_OUT16(io_out_bits,15,0);

    // LOCAL SIGNALS
    // Internals; generally not touched by application code
```

See that the LOCAL SIGNALS are missing in O0, but also note the comment "// Internals; generally not touched by application code"

I started poking around Verilator and it seems that if you want internal signals to be public, you're supposed to indicate as much in the Verilog code with `/*verilator public*/`. See http://www.veripool.org/projects/verilator/wiki/Faq. Grepping through chisel-testers does not show that this is being done so unless I'm missing something we're not actually marking any of the internal signals as public like we're supposed to. Also note that both of the header file snippets above have a section where `/*verilator public*/` signals should show up and the signals aren't there.

Verilator does have a --public option (http://www.veripool.org/projects/verilator/wiki/manual-verilator) but it is for "historical debug use only" so I guess we can't easily make all of the internal signals public.

ucbjrl:
Closed by ucb-bar/chisel-testers#51

PR #333, opened 2016-10-20
add verilator vcd command line option

vegaluisjose:
I have created a pull-request for rocket that support passing VCD filename from command line


jackkoenig:
Similar to ucb-bar/firrtl#344, the intent is to remove this support from this repo entirely but the requested feature should definitely be present in whatever we ultimately use

Issue #294, opened 2016-09-24
Standardize the One True Way for defining literals


ducky64:
With a plan for merging #265 on the roadmap, there are now quite a few (inconsistent!) ways to specify literals:
- UInt(value, width)
- UInt.Lit(value), UInt.Lit(value, width)
- value.asUInt(), value.asUInt(width)
- value.U

And several ways to construct value-less fields:
- UInt(width)
- UInt.width(width)

(and I'm probably missing a few, please add them...)

Since we'd like to break as much of the API as we need to before chisel3 (to minimize continued breakage in the future), let's try to standardize on one. The rest should be deprecated and/or moved into compatibility packages (so existing chisel2 compatibility code continues to work), and users can continue defining their own in their own codebases if they like, but officially, chisel3 should have just one.

I'm of the opinion that UInt(width) and UInt(value, width) should be mutually exclusive, because the first parameter isn't consistent in both cases, and they also do very different things.

Whether we keep UInt.Lit(value, [width]) or value.asUInt([width]) or value.U is a cosmetic and style question. .asUInt is structurally similar to Scala's toInt() methods, and the distinction between .U and .asUInt is just verbosity (though there's a good argument of favoring readability over code conciseness, before your code turns into symbols and alphabet soup)

Thoughts?

aswaterman:
IMO, 1.U is also more readable than 1.asUInt; the extra chars are clutter.
This is a DSL. The distinction between Int and UInt should not be so
strong. It would be ridiculous if it took fewer chars to express a literal
in the IR than in the language.

aswaterman:
BTW, I don't think the PR should be merged til the exact syntax is
resolved. This is about the API, after all!

ducky64:
That's true, but the other argument is that this could be a much longer debate (and arguably an issue that wasn't introduced by the PR, though it certainly contributed to the addition of styles). Note that there are some other issues still being worked out with the PR, so it's entirely possible this debate could be resolved before the PR is ready, and the results integrated.

More people should chime in: @jackkoenig @sdtwigg @chick @ucbjrl (probably many more I missed)
Including the DSP people, and whether their DSP DSL can be stylistically consistent: @grebe @shunshou

shunshou:
[...]

PR #338, opened 2016-10-21
Deprecate single arg UInt/SInt constructor

ucbjrl:
If compiling with "Strict" options in effect, deprecate the UInt/SInt single argument (lit) apply method and suggest one of the two competing alternatives.
Deprecate the Chisel2 `Direction` argument to UInt/SInt apply methods.

Have checkpoint() print warnings (so we actually see the deprecations).

Replace deprecated usage in utils and tests.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -550,12 +550,20 @@ private[core] sealed trait UIntFactory {
   // NOTE: This resolves UInt(width = 32)
   def apply(dir: Option[Direction] = None, width: Int): UInt = apply(Width(width))
   /** Create a UInt literal with inferred width.- compatibility with Chisel2. */
-  def apply(value: BigInt): UInt = apply(value, Width())
+  def apply(value: BigInt)(implicit opts: CompileOptions): UInt = {
+    if (opts.deprecateSingleArgumentFactoryMethods) {
+      Builder.deprecated("UInt.Lit(val) or val.U should be used over UInt(val)")
+    }
+    apply(value, Width())
+  }
   /** Create a UInt with a specified direction and width - compatibility with Chisel2. */
-  def apply(dir: Direction, width: Int): UInt = apply(dir, Width(width))
+  def apply(dir: Direction, width: Int)(implicit opts: CompileOptions): UInt = apply(dir, Width(width))(opts)
   /** Create a UInt with a specified direction, but unspecified width - compatibility with Chisel2. */
-  def apply(dir: Direction): UInt = apply(dir, Width())
-  def apply(dir: Direction, wWidth: Width): UInt = {
+  def apply(dir: Direction)(implicit opts: CompileOptions): UInt = apply(dir, Width())(opts)
+  def apply(dir: Direction, wWidth: Width)(implicit opts: CompileOptions): UInt = {
+    if (opts.deprecateOldDirectionMethods) {
+      Builder.deprecated("Direction(UInt(...)) should be used over UInt(DIRECTION, ...)")
+    }
     val result = apply(wWidth)
     dir match {
       case Direction.Input => Input(result)
@@ -700,7 +708,12 @@ object SInt {
   def width(width: Width): SInt = new SInt(width)
 
   /** Create an SInt literal with inferred width. */
-  def apply(value: BigInt): SInt = Lit(value)
+  def apply(value: BigInt)(implicit opts: CompileOptions): SInt = {
+    if (opts.deprecateSingleArgumentFactoryMethods) {
+      Builder.deprecated("SInt.Lit(val) or val.S should be used over SInt(val)")
+    }
+    Lit(value)
+  }
   /** Create an SInt literal with fixed width. */
   def apply(value: BigInt, width: Int): SInt = Lit(value, width)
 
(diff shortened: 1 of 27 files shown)


chick:
lgtm; looks like you found some deprecated usages while you were at it.  Good.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -550,12 +550,20 @@ private[core] sealed trait UIntFactory {
   // NOTE: This resolves UInt(width = 32)
   def apply(dir: Option[Direction] = None, width: Int): UInt = apply(Width(width))
   /** Create a UInt literal with inferred width.- compatibility with Chisel2. */
-  def apply(value: BigInt): UInt = apply(value, Width())
+  def apply(value: BigInt)(implicit opts: CompileOptions): UInt = {
+    if (opts.deprecateSingleArgumentFactoryMethods) {
```
ducky64: Might it make sense to factor this into a Builder method that takes an implicit CompileOptions?
So that these three lines would just be:
```
Builder.deprecatedSingleArgumentFactoryMethods("helpful message here")
```
(and have the `if` test be inside the method)

ducky64: Also, if we could make the compile options known at Scala-compile time, this could make deprecations more inline with Scala's native compile-time `@deprecated` annotation. With macro paradise (which will be pulled in with the naming annotations), this could also be an annotation.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -550,12 +550,20 @@ private[core] sealed trait UIntFactory {
...
+      Builder.deprecated("UInt.Lit(val) or val.U should be used over UInt(val)")
+    }
+    apply(value, Width())
+  }
   /** Create a UInt with a specified direction and width - compatibility with Chisel2. */
-  def apply(dir: Direction, width: Int): UInt = apply(dir, Width(width))
+  def apply(dir: Direction, width: Int)(implicit opts: CompileOptions): UInt = apply(dir, Width(width))(opts)
```
ducky64: If the argument is implicit do you need to explicitly pass it through?

src/test/scala/chiselTests/Direction.scala
```diff
@@ -35,4 +40,8 @@ class DirectionSpec extends ChiselPropSpec with ShouldMatchers {
      elaborate(new BadDirection)
     }
   }
+
+  property("Chisel2 directions should be deprecated") {
+    elaborate(new DeprecatedDirection)
```
ducky64: What is this testing? That the deprecation system doesn't just crash? Since it's not actually checking anything here, you should either document what this is doing, or perhaps remove it completely.
Though it may be possible to get a list of deprecation errors from elaborate (but that may require more plumbing?)

chick:
I don't have anything to add to the @ducky64 comments.  Overall looks ok to me.  Your call on taking it further or approving as is.

sdtwigg:
Rather than adding your own deprecation scheme, the compatibility layer should define its own UInt/SInt objects with the 'extra' functions marked as deprecated.

Alternatively, the compatibility layer leaves them unmarked and the 'real' objects have the functions marked as deprecated.

ducky64:
Problem is that apparently people want to control deprecation, in an old codebase, no one wants to get a thousand deprecation warnings if they have no intent to migrate to Chisel3, but if they do want to migrate, they want compatibility with deprecation so migration can be done incrementally without breaking the build. This PR puts that switch in CompileOptions.

I guess maybe another approach is to have a different compatibility layer package object intended for the migration process that deprecates things?

aswaterman:
The plan is to support the compatibility layer indefinitely, so as @ducky64 said, those should not generate deprecation warnings.

I think an additional layer for migration is a better solution, since it evaluates statically and separates the concerns from core.  But it's not clear that it's better-enough to warrant the effort.

sdtwigg:
From @aswaterman comments, in that case my latter proposal is more ideal.

The compatibility layer has the objects with functions unmarked and preserving legacy behaviors. Then, the 'real objects' have the functions marked (and hopefully eventually removed).

The compile options is a clever hack for behavioral variations that can't quite be handled with other mechanisms but shouldn't necessarily be the goto solution for every time there needs to be variations.

ucbjrl:
I agree with @sdtwigg and @aswaterman. I'll update the pr.

ducky64:
The idea is to provide for 3 use cases, indefinitely:
1. People who don't want to update legacy code (yet) - no deprecation warnings
2. People who want to (at some point) update legacy code incrementally, changing syntax but keeping things building - compatibility layer functionality with deprecation warnings
3. Awesome Chisel3 people - where these functions just wouldn't exist at all and would cause a compile time error.

I think your proposal only addresses cases 1 and 3?

sdtwigg:
I don't see a problem with ignoring #2. People who want to migrate slowly can add deprecation warnings themselves on whatever they want to migrate away from. I don't see the benefit of explicitly supporting (with new packages , layers, and associated implementation complexity) every possible feature permutation.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
ucbjrl: Yes. Otherwise the implicit is resolved at the site of the function definition with whatever is currently in scope, which doesn't include the value implicitly passed by the original caller. This may be a Scala bug.

ducky64:
I think it's worth considering Stephen's argument, whether this adds more complexity than it's worth. However, since the code is fairly isolated, it's not likely to produce wider correctness errors. The main continuing effect will probably just be lines-of-code maintenance burden.

Also, might it make sense to move the compatibility-specific methods into the compatibility package object instead of in core Bits.scala?

aswaterman:
I'm also amenable to that argument.

ducky64:
Since there's consensus on a path forward, I think the fixes should be implemented so this can be pushed through:
- Remove optional deprecations (unnecessary complexity)
- Move compatibility stuff into Chisel2 compatibility package (instead of in `chiselFrontend` or `core`)

ducky64:
With #294 resolved I'll implement this as part of deprecating all the other bad ways to build literals.

ducky64:
Superseded by #372 
It seems there's some other functionality in this pull request (some additions to checkpoint), if those are still needed they should be added in a separate pull request.

PR #339, opened 2016-10-24
fix SeqMem's read port creation

donggyukim:
Unfortunately, 41674d5e130f64d7489fdb8583b8f4ad88b64aeb broke read enable inference for SeqMems in `RemoveCHIRRTL` of firrtl, and also broke `inferReadWrite` because port creation generates additional wires to truncate index bits. This will help firrtl correctly infer read enables for SeqMems.


aswaterman:
@azidar is this the intended Firrtl behavior, i.e., that the address wire should be declared outside the when block, but the port should be declared within the when block?

If so, I agree this is the right fix.

azidar:
Yes, this is the intended Firrtl behavior.

PR #340, opened 2016-10-26
Add basic Chisel2 compatibility sanity checks.

ucbjrl:


ducky64:
The first part looks like a direct transformation of the Chisel compatibility package. Is it useful, and what kind of bugs is it expected to catch?
Potentially related: https://docs.google.com/document/d/13k8AsgYdF-2TJx9QIHHvPiuV3JMbsrMLkWmmjdYBLVg/edit

The Chisel2 RISC tester makes sense, but I don't think it's well targeted. What Chisel2 compatibility features in particular is it trying to test? Can we directly test those? I imagine individual tests that exercise the different connection semantics would be more useful, especially because a failure gives you a specific cause instead of "my omnibus processor test is on fire".

ucbjrl:
Good point on the compatibility package transform. I'll pull it.

ucbjrl:
The Chisel2 RISC tester is a "sanity" check, which is the issue this test is trying to address.

Testing individual connection semantics is a good idea. I'll add some.

jackkoenig:
@ucbjrl Can you check out why this is failing?

ucbjrl:
It's because of src/main/scala/chisel3/compatibility.scala:273

      require(!nodeType.widthKnown, "Bit width may no longer be specified for enums")

Introduced by Simplify Enum API (#385) edb19a0, and CompatibilitySpec.scala:126

    val add_op :: imm_op :: Nil = Enum(Bits(width = 8), 2)

Should we update the test or the new (incompatible) requirement?

ducky64:
I'd argue for changing the test. It makes no sense to specify a width in an enum since you can't control the values, and I don't think it was used outside chisel tests.

ucbjrl:
ping

src/test/scala/chiselTests/CompatibiltySpec.scala
```diff
@@ -0,0 +1,245 @@
...
+    u.getWidth shouldEqual width
+    u.dir shouldEqual(directionArgument)
+  }
+
+  it should "accept single argument U/SInt factory methods" in {
+    // Choose a random value
+    val value: Int = Gen.choose(0, Int.MaxValue).sample.get
+    val l = UInt(value)
+    l shouldBe a [UInt]
+    l shouldBe 'lit
```
jackkoenig: I don't know what this line does, but it's awesome

PR #341, opened 2016-10-26
Revamp VendingMachine.scala as cookbook example

jackkoenig:
- Add non-FSM implementation
- Add execution-driven test

Made a simple example of using FSM vs not, will turn in to a Wiki page


jackkoenig:
I should probably modify the non-FSM implementation to be similar in cost to the FSM implementation to counter the obvious criticisms...

jackkoenig:
Cleaned this up some more, any thoughts? The descriptions are kind of short for now but I imagine extending this example (or writing another one) to delve into the philosophical discussion of implicit vs explicit FSMs.

jackkoenig:
@chick What do you think about the updated version?

jackkoenig:
Alright so I changed the cookbook example to be simpler (and added it to the cookbook). I've also written a Verilog FSM and added another vending machine example. I intend to turn this code into a longer wiki discussion thta could also be a blog post.

PR #346, opened 2016-11-01
Add firrtl string to ExecutionOptionsManager.

grebe:
The driver currently writes emitted firrtl to a file and reads it back
in. This can be problematic if tests are being run in parallel on the
same module.

This patch passes the firrtl string to the driver directly, instead of
reading it back in from the written file.

Should I be removing references to the emitted firrtl file?


chick:
The fix looks good, I am independently looking at some changes to how it assigns work directories for all this stuff.  I'll merge this our augment tomorrow hopefully

chick:
I am closing this. This same fix is part of PR #393

PR #347, opened 2016-11-01
Remove de-duplication from Chisel

donggyukim:
@azidar @jackkoenig @chick @aswaterman @sdtwigg 
This PR is opened because module instance de-duplication will be implemented in FIRRTL.


chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -101,17 +101,11 @@ extends HasId {
   final val name = Builder.globalNamespace.name(desiredName)
 
   /** FIRRTL Module name */
-  private var _modName: Option[String] = None
-  private[chisel3] def setModName(name: String) = _modName = Some(name)
-  def modName = _modName match {
-    case Some(name) => name
-    case None => throwException("modName should be called after circuit elaboration")
-  }
+  def modName = name
```
aswaterman: Can we just delete modName?  It's not supposed to be part of the public API, anyway.

src/main/scala/chisel3/internal/firrtl/Emitter.scala
```diff
@@ -83,17 +80,10 @@ private class Emitter(circuit: Circuit) {
...
-        m.id setModName duplicate.name
-        ""
-      case None =>
-        defnMap((m.id.desiredName, defn)) = m
-        m.id setModName m.name
-        moduleDecl(m) + defn
-    }
+    val sb = new StringBuilder
+    sb append moduleDecl(m)
+    sb append moduleDefn(m)
```
jackkoenig: These should use dot notation
Infix notation should only be used for purely functional methods (http://docs.scala-lang.org/style/method-invocation.html#arity-1)

jackkoenig:
Even with approval, wait to merge this until the accompanying changes are merged to Firrtl so this can be retested and pass

azidar:
I believe the changes have been made to Firrtl - can this be merged now?

jackkoenig:
retest this please

colinschmidt:
This breaks de-duplication in Hwacha. Chisel3/master firrtl/master works as expected but chisel3/remove-deduplication and firrtl/master fails to deduplicate properly. I'll investigate a little more and hopefully get closer to a root cause to help track down the issue.

colinschmidt:
I found two Arbiter instances that are identical except for their name. Do we need to reopen the FIRRTL PR/issue ?

colinschmidt:
See https://github.com/ucb-bar/firrtl/issues/371

ucbjrl:
The breakage was caused by building with the onetrueliteral pr. Building
with master looks ok.

PR #349, opened 2016-11-02
Changed T to _T for generated names

azidar:


jackkoenig:
Per discussion in Chisel meeting:  Verilator provides an argument "--trace-underscore" if people do want to trace generated signals so we probably don't need to worry about adding any flags to Chisel or Firrtl

azidar:
Added the following to the commit message:
Prefix temporary names with underscores so Verilator won't trace them
Use verilator argument "--trace-underscore" if you want to trace these
signals

PR #354, opened 2016-11-08
[stevo]: set regnext return width to be same as input

stevobailey:
Chisel couldn't find bitwidth when I tried to use

Reverse(ShiftRegister(value, 2))

because Reverse calls getWidth. This fixes that issue. If a change to RegNext isn't desirable, I'll request a change to ShiftRegister instead.

### DIFF
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -11,13 +11,13 @@ object RegNext {
     *
     * Essentially a 1-cycle delayed version of the input signal.
     */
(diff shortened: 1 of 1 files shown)


aswaterman:
This changes the semantics of RegNext.  Previously, something like the following

    val foo = RegNext(x)
    when (bar) { foo := y }

would result in width(foo)=max(width(x), width(y)), but this change will force width(foo)=width(x).

Changing ShiftRegister to use Reg(t = value, next = value) seems like a better solution to me.

PR #355, opened 2016-11-08
[stevo]: allow bitwidth inference on output of shift register

stevobailey:
Chisel couldn't find bitwidth when I tried to use

Reverse(ShiftRegister(value, 2))

because Reverse calls getWidth. This fixes that issue.  See #354 for the old fix attempt, where I changed RegNext instead of ShiftRegister


stevobailey:
superseded by #370

PR #359, opened 2016-11-10
Rangemacro: add more tests

grebe:
@chick @ducky64 , I added some more basic tests


src/test/scala/chiselTests/RangeSpec.scala
```diff
@@ -8,10 +8,64 @@ import org.scalatest.{Matchers, FreeSpec}
 class RangeSpec extends FreeSpec with Matchers {
   "Ranges can be specified for UInt, SInt, and FixedPoint" - {
     "to specify a UInt" in {
-      val x = UInt(range"[0, 7)")
+      val x = UInt(range"[0, 8)")
       x.getWidth should be (3)
 
       println(range"[4,32)")
```
ducky64: Consider getting rid of prinlns in tests unless there's no other way to verify the results.

ducky64: A better way might be to check equality to the raw tuple generated from the range macro.

src/test/scala/chiselTests/RangeSpec.scala
```diff
@@ -8,10 +8,64 @@ import org.scalatest.{Matchers, FreeSpec}
 class RangeSpec extends FreeSpec with Matchers {
   "Ranges can be specified for UInt, SInt, and FixedPoint" - {
     "to specify a UInt" in {
-      val x = UInt(range"[0, 7)")
+      val x = UInt(range"[0, 8)")
```
ducky64: Do you want to check edge case [0, 0] in here? Or is the behavior still under debate?

grebe: I did check it, but for some reason put it in the test that was checking for illegal behavior. I moved it

src/test/scala/chiselTests/RangeSpec.scala
grebe: I'll add a test that checks the macros by themselves, without factory methods

ducky64:
Do you also want to test interpolated variables, both literals (like range"[0, ${1+1}]" and variables?

grebe:
Is there a way to test invalid macro invocations? I.e. range"(-1 1)" should be illegal because it is missing a comma.

ducky64:
I don't think it's possible to test invalid macro invocations, since those would fail at compile time, before the test suite even starts running.

ducky64:
Also, LGTM, merging.

sdtwigg:
ScalaTest assertTypeError is designed to test if code fails to typecheck
(supplied code must still be parsable).

It is commonly used to check macro invocations and typelevel things.

ducky64:
Cool, thanks! Added some compile-time failure checks. 04a1d71

PR #360, opened 2016-11-10
Throw exceptions for cloneType failures - fix #358

### LINKED ISSUE #358: Erroneous BindingException in both Chisel._ and chisel3._
Given the following:

```scala
import chisel3._

class TestIO(w: Int) extends Bundle {
  val a = Vec(4, UInt(width = w)).asInput

  //override def cloneType = (new TestIO(w)).asInstanceOf[this.type]
}

class Test extends Module {
  val io = IO(new TestIO(32))
}

class TestTop extends Module {
  val io = IO(new Bundle {})

  val subs = Vec.fill(2) { Module(new Test).io }
}
```

and

```scala
import Chisel._

class TestIO(w: Int) extends Bundle {
  val b = Vec(4, UInt(width = w)).asInput

  //override def cloneType = (new TestIO(w)).asInstanceOf[this.type]
}

class Test extends Module {
  val io = new TestIO(32)
}

class TestTop extends Module {
  val io = new Bundle {}

  val subs = Vec.fill(2) { Module(new Test).io }
}
```

With cloneType commented out, both versions get the error: 

`chisel3.core.Binding$BindingException: : Already bound to PortBinding(example.Test@6,Some(input))`

Both work as expected if cloneType is uncommented.

I believe there are two issues:

1. This should be an error with no implemented cloneType, not a BindingException
1. BindingExceptions shouldn't show up at all in Chisel._

ucbjrl:


jackkoenig:
This looks good. Might it also be wise for throwException to print the errors in the ErrorLog if there are any? I think such a change would help solve all problems of errors getting covered up by thrown exceptions.

ucbjrl:
I think it would be a good idea to output accumulated errors before existing in response to an exception. I think it should be a separate issue - #361

ucbjrl:
Agreed. Tests should be legal Chisel code.

ucbjrl:
@jackkoenig, could you re-review this please?

PR #362, opened 2016-11-11
Add checks for forgetting Module() wraps or double Module() wrapping

### LINKED ISSUE #192: Curious behavior when wrapping Module instantiation in multiple Module.apply invocations
The code below seems like a very reasonable thing to write:

``` scala
class Minuser extends Module {
  val io = new Bundle {
    val a = UInt(width = 32).flip()
    val b = UInt(width = 32).flip()
    val c = UInt(width = 32)

    c := a -% b
  }
}

object Driver extends App {
  val s = Chisel.Driver.emit(() => Module(new Minuser))
  println(s)
}
```

but Chisel.Driver.emit calls elaborate which invokes Module.apply, so this is essentially double calling the apply method.

This results in the following, somewhat surprising FIRRTL:

```
circuit Minuser : 
  module Minuser : 
    input clk_7 : Clock
    input reset_8 : UInt<1>
    output io_9 : {flip a_3 : UInt<32>, flip b_2 : UInt<32>, c_1 : UInt<32>}

    io_9 is invalid
    io_9 is invalid
    node T_5 = sub(io_9.a_3, io_9.b_2)
    node T_6 = tail(T_5, 1)
    io_9.c_1 <= T_6
```

Is there any way we can deal with this?

jackkoenig:
Implemented by changing currentModule to moduleStack
Fixes #192 

Deleted DeqIOSpec because it wasn't really doing anything and it was missing a Module() wrap anyway. When I added the Module() wrap the test just times out.


jackkoenig:
Note that this might break other repos' tests because even in this repo people sometimes forget to Module wrap or double Module wrap and things just kinda happened to work out.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -25,16 +25,22 @@ object Module {
...
+    val m = bc.setRefs() // This will invoke bc which pushes it to the top of the Module stack!!!
     m._commands.prepend(DefInvalid(childSourceInfo, m.io.ref)) // init module outputs
-    Builder.currentModule = parent // Back to parent!
+    // Since the above only calls the constructor for the Module once, if you double wrap then we
+    // end up popping twice. If you forget to Module wrap, then a outter Module apply will trigger this
+    if (Builder.moduleStack.isEmpty || Builder.moduleStack.head != m) {
+      throwException("Error: Top of Module stack is not the Module expected. " +
+                     "You probably forgot a Module() wrapper or double Module() wrapped.")
+
+    }
```
aswaterman: I suggest splitting these cases apart and giving the more precise error message in either case.

aswaterman: Ah, I see.

sdtwigg:
I think you can do this without a Stack by just having a global boolean variable.

Module() asserts variable is false (and then sets it to true) and Module constructor asserts variable is true (and then sets it to false). If I recall, this is what was done in Chisel 2 with modStackPushed.

sdtwigg:
Amendment: Module() should check twice. Before setting the variable and after constructing the Module. The latter check catches trying to re-Module an already instantiated Module (which is like double wrapping but different due to the fact Module() is call-by-name):
```scala
val myMod = Module(new Blah)
val myMod2 = Module(myMod)
```

aswaterman:
Agreed, but the intent is arguably clearer this way, and it's conceivable that some future implementation feature would want the explicit stack anyway.  (And, of course, the performance implications are irrelevant.)

jackkoenig:
Alright done. Switch implementation to boolean because it can also detect rewrapping with Module. Also switched back to currentModule because the stack was no longer being used (can always be reimplemented if it's useful). Also added useful function on SourceLine for printing error messages.

PR #364, opened 2016-11-14
Change noenq in ReadyValid to use an uninitialized Wire instead of zero

jackkoenig:
Prevent unnecessary hardware from being generated.

Also, why are these functions defined in a value class instead of just on the class itself?


aswaterman:
Does anyone have any idea where this function came from?

ucbjrl:
It was part of Stephen's PR 200.

aswaterman:
I don't remember discussing that aspect of the PR. @sdtwigg is this meant to replace EnqIO/DeqIO? Do people like/use it?

sdtwigg:
I don't like or use it; however, it got placed in DecoupledIO when EnqIO and DeqIO were folded into just being DecoupledIO. It was there before that PR

aswaterman:
Hmm... ok.  I note with amusement that within chisel3, the only place they are used is in a test module named BrokenVectorPacketModule.

I'll approve the PR, because it's unrelated to whether these methods should exist.

PR #366, opened 2016-11-15
Add Record as superclass of Bundle for custom Aggregate types

jackkoenig:
This proposal for programmatic bundles is to just use Bundle.apply to generate Bundles. I haven't thought through all of the implications so here is a PR to start discussion. I also need to write more tests, most of the ones I've added so far just check standard Bundles.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -335,50 +335,93 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId {
     SeqUtils.oneHotMux(indexWhereHelper(p))
 }
 
+/** Base class for Aggregates based on key values pairs of String and Data
+  *
+  * Record should only be extended by libraries and fairly sophisticated generators.
+  * RTL writers should use [[Bundle]].
+  */
+abstract class Record extends Aggregate {
+
+  /** The collection of [[Data]]
+    *
+    * This underlying datastructure is a ListMap because the elements must
+    * remain ordered for serialization/deserialization. Elements added later
+    * are higher order when serialized (this is similar to [[Vec]]). For example:
+    * {{{
+    *   // Assume we have some type MyRecord that creates a Record from the ListMap
+    *   val record = MyRecord(ListMap("fizz" -> UInt(16.W), "buzz" -> UInt(16.W)))
+    *   // "buzz" is higher order because it was added later than "fizz"
+    *   record("fizz") := "hdead".U
+    *   record("buzz") := "hbeef".U
+    *   val uint = record.asUInt
+    *   assert(uint === "hbeefdead".U) // This will pass
+    * }}}
+    */
+  val elements: ListMap[String, Data]
+
+  /** Name for Pretty Printing */
+  def className: String = this.getClass.getSimpleName
+
+  private[chisel3] def toType = {
+    def eltPort(elt: Data): String = {
+      val flipStr: String = if(Data.isFirrtlFlipped(elt)) "flip " else ""
+      s"${flipStr}${elt.getRef.name} : ${elt.toType}"
+    }
+    elements.toIndexedSeq.reverse.map(e => eltPort(e._2)).mkString("{", ", ", "}")
+  }
+
+  private[chisel3] lazy val flatten = elements.toIndexedSeq.flatMap(_._2.flatten)
+
+  // NOTE: This sets up dependent references, it can be done before closing the Module
+  private[chisel3] override def _onModuleClose: Unit = { // scalastyle:ignore method.name
+    val _namespace = Builder.globalNamespace.child
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name)) }
+  }
+
+  private[chisel3] final def allElements: Seq[Element] = elements.toIndexedSeq.flatMap(_._2.allElements)
+
+  // Helper because Bundle elements are reversed before printing
+  private[chisel3] def toPrintableHelper(elts: Seq[(String, Data)]): Printable = {
+    val xs =
+      if (elts.isEmpty) List.empty[Printable] // special case because of dropRight below
+      else elts flatMap { case (name, data) =>
+             List(PString(s"$name -> "), data.toPrintable, PString(", "))
+           } dropRight 1 // Remove trailing ", "
+    PString(s"$className(") + Printables(xs) + PString(")")
+  }
+  /** Default "pretty-print" implementation
+    * Analogous to printing a Map
+    * Results in "$className(elt0.name -> elt0.value, ...)"
+    */
+  def toPrintable: Printable = toPrintableHelper(elements.toList)
+}
+
 /** Base class for data types defined as a bundle of other data types.
   *
   * Usage: extend this class (either as an anonymous or named class) and define
   * members variables of [[Data]] subtypes to be elements in the Bundle.
   */
-class Bundle extends Aggregate {
-  private val _namespace = Builder.globalNamespace.child
+class Bundle extends Record {
+  override def className = "Bundle"
 
-  // TODO: replace with better defined FIRRTL weak-connect operator
-  /** Connect elements in this Bundle to elements in `that` on a best-effort
-    * (weak) basis, matching by type, orientation, and name.
+  /** The collection of [[Data]]
     *
-    * @note unconnected elements will NOT generate errors or warnings
-    *
-    * @example
+    * Elements defined earlier in the Bundle are higher order upon
+    * serialization. For example:
     * {{{
-    * // Pass through wires in this module's io to those mySubModule's io,
-    * // matching by type, orientation, and name, and ignoring extra wires.
-    * mySubModule.io <> io
+    *   class MyBundle extends Bundle {
+    *     val foo = UInt(16.W)
+    *     val bar = UInt(16.W)
+    *   }
+    *   // Note that foo is higher order because its defined earlier in the Bundle
+    *   val bundle = Wire(new MyBundle)
+    *   bundle.foo := 0x1234.U
+    *   bundle.bar := 0x5678.U
+    *   val uint = bundle.asUInt
+    *   assert(uint === "h12345678".U) // This will pass
     * }}}
     */
-
-  lazy val elements: ListMap[String, Data] = ListMap(namedElts:_*)
-
-  /** Returns a best guess at whether a field in this Bundle is a user-defined
-    * Bundle element without looking at type signatures.
-    */
-  private def isBundleField(m: java.lang.reflect.Method) =
-    m.getParameterTypes.isEmpty &&
-    !java.lang.reflect.Modifier.isStatic(m.getModifiers) &&
-    !(Bundle.keywords contains m.getName) && !(m.getName contains '$')
-
-  /** Returns a field's contained user-defined Bundle element if it appears to
-    * be one, otherwise returns None.
-    */
-  private def getBundleField(m: java.lang.reflect.Method): Option[Data] = m.invoke(this) match {
-    case d: Data => Some(d)
-    case Some(d: Data) => Some(d)
-    case _ => None
-  }
-
-  /** Returns a list of elements in this Bundle.
-    */
-  private[core] lazy val namedElts = {
+  final lazy val elements: ListMap[String, Data] = {
     val nameMap = LinkedHashMap[String, Data]()
     val seen = HashSet[Data]()
     for (m <- getPublicFields(classOf[Bundle])) {
@@ -391,20 +434,17 @@ class Bundle extends Aggregate {
         }
       }
     }
-    ArrayBuffer(nameMap.toSeq:_*) sortWith {case ((an, a), (bn, b)) => (a._id > b._id) || ((a eq b) && (an > bn))}
+    ListMap(nameMap.toSeq sortWith { case ((an, a), (bn, b)) => (a._id > b._id) || ((a eq b) && (an > bn)) }: _*)
   }
-  private[chisel3] def toType = {
-    def eltPort(elt: Data): String = {
-      val flipStr: String = if(Data.isFirrtlFlipped(elt)) "flip " else ""
-      s"${flipStr}${elt.getRef.name} : ${elt.toType}"
-    }
-    s"{${namedElts.reverse.map(e => eltPort(e._2)).mkString(", ")}}"
-  }
-  private[chisel3] lazy val flatten = namedElts.flatMap(_._2.flatten)
-  private[chisel3] override def _onModuleClose: Unit = // scalastyle:ignore method.name
-    for ((name, elt) <- namedElts) { elt.setRef(this, _namespace.name(name)) }
 
-  private[chisel3] final def allElements: Seq[Element] = namedElts.flatMap(_._2.allElements)
+  /** Returns a field's contained user-defined Bundle element if it appears to
+    * be one, otherwise returns None.
+    */
+  private def getBundleField(m: java.lang.reflect.Method): Option[Data] = m.invoke(this) match {
+    case d: Data => Some(d)
+    case Some(d: Data) => Some(d)
+    case _ => None
+  }
 
   override def cloneType : this.type = {
     // If the user did not provide a cloneType method, try invoking one of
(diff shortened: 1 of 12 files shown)


jackkoenig:
Also given that namedElts is what is used by most internal things, should we make elements final? I'm not sure what use there is in overriding it.

aswaterman:
Add experimental warning to Bundle.apply?

aswaterman:
Yeah.

chick:
lgtm

jackkoenig:
After further discussion, we would like to type parameterize these "parameterizable bundles". We can't type parameterize Bundle because that will break all Chisel code, so I'm working on an alternative

sdtwigg:
It is concerning to have this exposed as an ArrayBuffer (neither part of the mutability is good).

Also, I think it would be better to have the user extend the trait. That way their fields can still retain type information. e.g.

```scala
// The oft-requested sequence of elements with varying widths
// invariant because can be source or sink of connection
class HeterogenousVec[T<: Data](elems: Seq[T]) extends CustomBundle {
  val fields = elems.map(_.chiselCloneType)

  // doesn't need to be lazy
  override val namedElts: ArrayBuffer[(String, Data)] = ArrayBuffer(
    (for((e, i) <- elems.zipWithIndex) yield s"field$i" -> e):_*
  )
  override protected def cloneType: this.type =
    (new HeterogenousVec(elems)).asInstanceOf[this.type]
}
```

```scala
// This demonstrates a bundle with 'extra' (differently typed) fields
// that are separately added to namedElts.
class MuxInput[T<: Data](elems: Seq[T]) extends CustomBundle {
  val select = UInt(log2Up(elems.length))
  val fields = elems.map(_.chiselCloneType)

  override val namedElts: ArrayBuffer[(String, Data)] = ArrayBuffer(
    (for((e, i) <- elems.zipWithIndex) yield s"field$i" -> e):_*, ("select", select)
  )
  override protected def cloneType: this.type =
    (new MuxInputPacket(elems)).asInstanceOf[this.type]
}
```

accesses through fields expose the type as T.

jackkoenig:
I have proposed a new solution that is essentially what @sdtwigg described. I had independently come to a similar conclusion and with minor modification it matches the spirit of his comment.

This is currently RFC. Documentation and more tests are needed before merging. This is a fairly significant API change.

jackkoenig:
To summarize the new proposal:

Bundle now extends abstract class Record which extends Aggregate. Record takes much of the code originally in Bundle, but leaves elements (ListMap[String, Data]) and cloneType unimplemented. Bundle implements these fields (elements via reflection, cloneType with the same "best effort" attempt as before). Library writers and power users can extend Record directly to build more interesting custom Chisel Data types. Bundle is still the workhorse of RTL. Since Bundle extends Record, the two are interoperable with Record becoming the type generally used within Chisel. The current RecordSpec tests illustrate programmatically creating bundles from ListMaps directly. @sdtwigg's examples above should also work (and should be written as tests before merging).

jackkoenig:
Also note the more complicated cloneType in RecordFromMap in RecordSpec. This extra complication supports the notion that Record should only be extended when necessary, most users should stick to Bundles.

terpstra:
Could someone please rebase this commit against current master? :)

aswaterman:
Yeah, it appears that only a whitespace change caused the merge conflict... one sec.

aswaterman:
@terpstra there you go

terpstra:
I used this PR to write this:

```scala
final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
  def apply(x: Int) = elts(x)
  def length = elts.length

  private val contents = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, n) }:_*)
  def elements = contents
  override def cloneType: this.type = (new HeterogeneousBag(elts)).asInstanceOf[this.type]
}
```

I was not sure if it needed to have a custom dashcode or eq operation. In any case, it totally worked for what I needed! I was able to put a bunch of differently shaped items into a Bag in my IO Bundle and it did not squash them together like a Vec. Awesome!

jackkoenig:
I'm glad this is useful! I think there are a couple of changes that *probably* should be made to HeterogenousBag:

* `cloneType` should be called on all elements upon construction rather than just cloneType (I know I did the same in the original test, I've been meaning to update it). This is consistent with how Bundle and Vec do it currently.
* With the above, `apply` should index into the already cloneTyped underlying data structure (so as to get the correct references)

@aswaterman or @sdtwigg is this correct? I am hoping to standardize some guidelines for Record use before we unleash it.

jackkoenig:
Questions for the public API:

1. Should Record be available to the compatibility wrapper? People could always use chisel3._ to make the custom datastructure and then just instantiate it in their Chisel._ code.
1. Should Record emit a Firrtl Bundle to keep the underlying structure (and name of the aggregate)? @azidar suggested this to me. I tried out HeterogeneousBag and it emitted flat names _0 ... _N. Note that this would result in names with two underscores in the final Verilog in the case of HeterogeneousBag.
1. Should the type of `Module.io` be changed to Record? Or even Data?
1. Should HeterogeneousBag be made part of chisel util? This type of aggregate is the one of the most commonly requested features, it might be wise to get it right once so others don't have to.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -320,50 +320,59 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId {
...
+/** Base class for Aggregates based on pairs of (String, Data)
   *
-  * Usage: extend this class (either as an anonymous or named class) and define
-  * members variables of [[Data]] subtypes to be elements in the Bundle.
+  * Only used to share implementation code, not exposed as part of API
   */
-class Bundle extends Aggregate {
+abstract class Record extends Aggregate {
+
+  def elements: ListMap[String, Data]
```
sdtwigg: This should be val elements.

Ideally even scala.immutable.ListMap[String, Data] but that might not be feasible atm given other weirdness in current Bundle.

aswaterman:
You shouldn't need to cloneType in both places @jackkoenig; I think the following would suffice:

    final case class HeterogeneousBag[T <: Data](eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
      private val elts = eltsIn map cloneType
      def apply(x: Int) = elts(x)
      def length = elts.length
   
      private val contents = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, n) }:_*)
      def elements = contents
      override def cloneType: this.type = (new HeterogeneousBag(elts)).asInstanceOf[this.type]
    }

jackkoenig:
I agree @aswaterman, I should've been more clear, I meant that cloneType should be called upon construction *rather than* upon calling cloneType on the Aggregate. Also, this means HeterogeneousBag *probably* should not be a case class since unapplying and getting `eltsIn` is probably not what you want.

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+    override def cloneType = (new BundleFoo).asInstanceOf[this.type]
+  }
+  class BundleBar extends Bundle {
+    val bar = UInt.width(32)
+    override def cloneType = (new BundleBar).asInstanceOf[this.type]
+  }
+
+  class MyModule(output: Bundle, input: Bundle) extends Module {
+    val io = IO(new Bundle {
+      val in = Input(input.cloneType)
```
sdtwigg: Remove call to cloneType. Input and Output already clone their inputs.

aswaterman:
1. No, for the reason you mentioned.
2. Maybe? In this case, it would make the code less readable, right?  Unless Firrtl were to support field names 0, 1, ...
3. I guess that would follow if you do point 2.
4. Let's play around with the API in rocket-chip more before deciding... but possibly.

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -0,0 +1,45 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import scala.collection.immutable.ListMap
+
+class RecordSpec extends ChiselFlatSpec {
+  class RecordFromMap(val elements: ListMap[String, Data]) extends Record {
+    override def cloneType =
```
sdtwigg: I think this is a little nicer:
override def cloneType = (new RecordFromMap(
  for((field, elem) <- elements) yield field -> elem.cloneType
)).asInstanceOf[this.type]

or at least,

override def cloneType = (new RecordFromMap(
  elements.map(e => e._1 -> e._2.cloneType)
)).asInstanceOf[this.type]

sdtwigg: The particulars of Aggregate cloning may require your cloneType to be calling chiselCloneType. Otherwise, an intermediate Aggregate may fail to get the proper direction.

sdtwigg:
First, no one (but chiselCloneType) should be calling cloneType. cloneType makes an imperfect clone whereas chiselCloneType manipulates the result of cloneType to closer-to-perfect clone of the type. When defining a new datatype, it may be tempting (and legal!) to prefer cloneType for performance reasons; however, this can lead to oddities now that chisel3 tries to track the direction of intermediate Aggregates. Thus, care should be taken to ensure directions are properly clones in the Bundle of Bundle of ... cases.

It is incorrect for Record to clone its 'elements' for the same reason Bundle does not clone its elements. Presumably, the user who is extending record is assigning those elements to some val and thus a cloneType on (or right before setting) 'elements' would effectively invalidate all those references. Users that extend Record should be calling chiselCloneType on the fields. Note that Input and Output already call chiselCloneType for you.

1. No, the compatibility system is already fairly complex. It would be nice not to have to specify or fix how new features (relying on new semantics) interact with old code (importing old semantics).

2. Yes, it already should be emitting a FIRRTL Bundle. A HeterogenousBag should result in a FIRRTL Bundle with fields called field_0, field_1, field_2, .... (and each may have a different type). For the () comment, I was under the impression that FIRRTL was incapable of retaining the Aggregate name. (e.g. FIRRTL never actually sees DecoupledIO).

3. Yes, definitely Record. In almost all places in core Chisel, Bundle should be replaced by Record. Bundle now is just a convenience that uses reflection to get element names for you and attempts to define a cloneType method for you. (It might even be helpful to split both of those off into individual helper traits so a user can get one behavior without the other.) Also, yes, I don't see why io mechanically HAS to be a Bundle|Record instead of Data. (That said, I'd imagine 99%+ of Modules would have an io <: Record)

4. Just make sure you really like the name first. These things tend to be hard to sever.

sdtwigg:
I'm concerned by any code examples that suggest Chisel Data subtypes (or Module) can be defined as a case class. From http://docs.scala-lang.org/tutorials/tour/case-classes.html , Chisel.Data immediately violates assumptions 1 (immutable by default) and 3 (compared by structural equality).

Because Chisel (currently) overrides hashCode and equals in HasId, the case class won't actually introduce structural equality and thus things should still work in implementation. However, anyone who reads the case class part and expects value equality will be in for an unfortunate surprise. Also, the override of those two functions is technically a (rather archaic) implementation detail and thus not something users should be relying on.

If all that is desired from the case class is the object apply, it would likely be best just to define that directly.

terpstra:
Hey, I am happy to see a modified Bag become part of the chisel, but I needed this for a specific case where cloneType of the inputs is not ok.

terpstra:
I just wanted to say that as a Chisel user, I am really looking forward to this feature. I already have several places where I will want to use it immediately. When can I expect something like this to be merged to master? 👍

jackkoenig:
I've updated this PR slightly, and I think the current example of `CustomBundle` illustrates how `Record` should be extended. To summarize the currently proposed changes:
* Add Record as subclass of Aggregate
  * `val elements: ListMap[String, Data]` is underlying datastructure
    * I tested the change from `def elements` to `val elements` in rocket-chip and there doesn't seem to be a problem
  * `Bundle` becomes a subclass of `Record` that uses reflection to populate `elements`
    * `namedElts` is no more
  * Record has public function `className` for use in pretty printing
  * Record maps to Bundles in Firrtl, which means if you use numbers for names, you will get a Firrtl bundles with fields named things like `_0`, `_1`, ...

colinschmidt:
I believe a utility like https://gist.github.com/colinschmidt/bc0b15ec90422a87574498e9c9c72ba4 is still easy to implement with these changes but I just wanted to point out one of my use cases of ```elements```

jackkoenig:
As far as I can tell, this PR doesn't affect the gist much except you might want to change Bundle to Record (but I think it should still work as is).

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -320,50 +320,60 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId {
...
 
-  /** Returns a best guess at whether a field in this Bundle is a user-defined
-    * Bundle element without looking at type signatures.
-    */
-  private def isBundleField(m: java.lang.reflect.Method) =
-    m.getParameterTypes.isEmpty &&
-    !java.lang.reflect.Modifier.isStatic(m.getModifiers) &&
-    !(Bundle.keywords contains m.getName) && !(m.getName contains '$')
+  private[chisel3] override def _onModuleClose: Unit = // scalastyle:ignore method.name
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name)) }
```
sdtwigg: Move the namespace declaration into here. It is unclear to me why it is retained as a class field.

sdtwigg: PS: For future changes, _onModuleClose doesn't quite capture what the point of the function is (setting up dependent references). Particularly because this can (and probably should) be done before the Module is 'closed'.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -408,33 +415,19 @@ class Bundle extends Aggregate {
...
-      else {
-        elements.toList.reverse flatMap { case (name, data) =>
-          List(PString(s"$name -> "), data.toPrintable, PString(", "))
-        } dropRight 1 // Remove trailing ", "
-      }
-    PString("Bundle(") + Printables(elts) + PString(")")
-  }
 }
 
 private[core] object Bundle {
```
sdtwigg: Are these fields still needed?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -376,27 +386,24 @@ class Bundle extends Aggregate {
...
+  /** Returns a field's contained user-defined Bundle element if it appears to
+    * be one, otherwise returns None.
+    */
+  private def getBundleField(m: java.lang.reflect.Method): Option[Data] = m.invoke(this) match {
+    case d: Data => Some(d)
+    case Some(d: Data) => Some(d)
+    case _ => None
+  }
 
   override def cloneType : this.type = {
```
sdtwigg: These uses of Record didn't have to be changed to Bundle. (Although, am ambivalent on whichever word.)

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -320,50 +320,60 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId {
...
   *
-  * Usage: extend this class (either as an anonymous or named class) and define
-  * members variables of [[Data]] subtypes to be elements in the Bundle.
+  * Record should only be extended by libraries and fairly sophisticated generators.
+  * RTL writers should use [[Bundle]].
   */
-class Bundle extends Aggregate {
+abstract class Record extends Aggregate {
+
+  val elements: ListMap[String, Data]
```
sdtwigg: Should document somewhere here how the elements definition order affects asUInt and fromUInt deserialization/serialization.

terpstra:
Still really looking forward to this PR... Any idea when it will get merged? It will make a lot of things that we have to hack around a lot cleaner.

shunshou:
Also looking forward to finally having a legit HetVec thing in Chisel, now that I'm back to making generators... Glad more people are finally agreeing this is actually useful...

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
jackkoenig: I changed them back

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
jackkoenig: I don't think they are used, but I think removing them is a separate PR since it kind of goes along with modifying the Namespace

jackkoenig:
I've addressed Stephen's comments including adding documentation for serialization order of the elements for both Record and Bundle as well as tests that illustrate the API. Any feedback on the documentation phrasing (or I guess the "chosen" serialization order for Record since I guess it could be the opposite order) is appreciated. @aswaterman @sdtwigg and anyone else who cares please take a look! I think this is about ready.

jackkoenig:
One thing I did **not** do is make Module.io be a Record, it is still a Bundle. Should it be a Record?

sdtwigg:
Yes, Module.io should be a Record. This ends up being particularly important for BlackBoxes.

jackkoenig:
Done

jackkoenig:
Testing this with rocket-chip: https://travis-ci.org/ucb-bar/rocket-chip/builds/193501120

jackkoenig:
I don't know why the rocket-chip Travis builds are insisting on running with an old version of Firrtl and thus failing, but I've run all of the test suites locally on the rocket-chip test-record branch and they pass

jackkoenig:
This is ready for merging assuming people are happy with it

ucbjrl:
This breaks `IOAccessor` in chisel-testers. Should it (and `parseBundle`) be changed to expect a `Record` instead of a `Bundle`?

jackkoenig:
I think so. `Bundle` is now just a special case of `Record` that populates `elements` via reflection.

PR #369, opened 2016-11-16
Add ranging API

ducky64:
Adds the range specifier macro, allowing range to be specified in mathematical notation using string interpolators like:
`range"[0, 2)"`
or
`range"[$myInt, 2)"`
or
`range"[${myInt + 4}, 2)"`

The specifier itself is parsed and (syntactically) validated at compile time using macros.

Currently, the propagation of range information to FIRRTL isn't there yet, it just converts the ranges to a bitwidth. Reasoning (discussed at the Chisel meeting) for pushing this now is to allow people to use the API and leave optimizations to later. The actual more optimized range propagation will be a future pull request.

For those who haven't been following Chisel dev, ranges are a way to specify data widths at a more granular level, allowing more optimal width propagation. For example, it might not make sense to have `1.U +& 1.U +& 1.U` be 3 bits wide when only 2 bits are needed to represent `3.U`.


coreMacros/src/main/scala/chisel3/internal/RangeTransform.scala
```diff
@@ -0,0 +1,102 @@
...
+
+    var nextStringIndex: Int = 1
+    var nextArgIndex: Int = 0
+    var currString: String = strings(0)
+
+    /** Mutably gets the next numeric value in the range specifier.
+      */
+    def getNextValue(): c.Tree = {
+      currString = currString.dropWhile(_ == ' ')  // allow whitespace
+      if (currString.isEmpty()) {
```
azidar: No () on isEmpty

ducky64:
Added requested test and fixed isEmpty.

Also, StringContext parses as two separate lists of strings and interpolated variables (indexed such that the interpolated variable elements are between string elements), which makes it difficult to parse using a regex. For example, s"12${56}78" would become two lists: ["12", "78"] (format string) and [56] (interpolated element). I agree that the parser implementation isn't as elegant as I'd like, but I don't think it's possible to get rid of the mutable index and using a tokenizer-like approach (where you eat the input as you traverse it) seems to make sense.

azidar:
makes sense. One final thing, is it possible to mark (at the very least, with comments) that these are experimental features?

ducky64:
Yeah, I think it makes sense to mark these as experimental. What do you think about requiring users to import a chisel3.experimental package, much like how Scala requires that for the use of macros? Eventually, that import would become a noop when these get mainlines back into chisel3._

Otherwise a comment is fine.

azidar:
I think an experimental package is an excellent idea

ducky64:
Added experimental package.

chick:
lgtm

jackkoenig:
Per range discussion in #294 

I appreciate this as being marked experimental, but it's not clear to me why it can't be done experimentally as a library rather than core feature of Chisel. This could be a perfect candidate annotations propagating from Chisel to Firrtl and then a custom range inference transform could be run before Width Inference.

ducky64:
This feature requires (or will eventually require, once completely implemented) changes to core data types (UInt and creating a Width-like type) and FIRRTL generation (allowing ranges instead of width), so putting it in a separate library would be pretty difficult.

jackkoenig:
Could you elaborate on what changes to core Chisel data types are required? Obviously the annotations API isn't done but we've been working very hard on it and I think it will be available very soon.

Unless I'm missing something important, this sounds like an exemplar of how annotations should be used. Extra information that does not change the semantics of the circuit but can improve QoR.

aswaterman:
I don't think that's a good enough reason in itself. It could be
implemented in a research fork of Chisel/Firrtl, and merged later after a
long, deliberate process.

ducky64:
In core Chisel, I think this requires changes to UInt (and friends) so it can support a range width instead of bit widths. I was also under the impression that ranging would be a core FIRRTL feature (rather than annotation), though @azidar would know more. This feature has also been discussed at several meetings, and I don't think there was any significant pushback.

Anyways, if anyone thinks that Chisel is modular enough that this could be cleanly implemented as a library, that would certainly be interesting from a software engineering standpoint. Though it also makes sense for this to be in core Chisel if ranging is going to be in core FIRRTL.

Also, yes, this can be a fork, though that causes significant maintenance and compatibility issues, especially if the ranging API is going to be widely used, as would be expected in DSP contexts.

jackkoenig:
I believe it could be implemented as follows:

The library provides a way to specify UInts/whatever with ranges. The library outputs to Chisel a UInt with the minimum width necessary to represent the max of the range; it also issues a "Range Annotation" associated with the given port, register, or wire. A Firrtl transformation run before width inference can then perform range inference based on these annotations.

One of the big motivations in making Chisel3 was to reduce Chisel down to a set of core features upon which people could layer libraries. Now it seems like this policy has been reversed to instead expand the API a bunch. I strongly suspect that a library will suffice here which is why I'm so hesitant to add features that Chisel will have to support forever.

azidar:
I agree we can try and do it as an external library. However, it's prudent to merge now as experimental, and we can easily remove it later.

ducky64:
Waiting on approval.

Plan will be:
Merge now so people can start using it. Note that experimental carries a much lower commitment to continued support.
If the backend can be implemented as a FIRRTL annotation (I imagine it will take some time to try), this can be broken back out in a library. Users would need to add a new dependency to sbt as well as (possibly) change an import statement where this is used.

PR #370, opened 2016-11-17
[stevo]: shift register enable signal now enables ALL registers in th…

stevobailey:
…e chain.

Previously, the enable signal only worked on the first register. Later ones still shifted every cycle. Now they are all controlled by en.

This also fixes the problem in PR #355, so I'll close that one in favor of this one.

### DIFF
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -55,9 +55,7 @@ object ShiftRegister
     */
   def apply[T <: Data](in: T, n: Int, en: Bool = Bool(true)): T = {
     // The order of tests reflects the expected use cases.
-    if (n == 1) {
-      RegEnable(in, en)
-    } else if (n != 0) {
+    if (n != 0) {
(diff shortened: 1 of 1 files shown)


aswaterman:
shouldn't the enable be pipelined down, rather than simply fanning out to all of the stages?

stevobailey:
But then "disabling" the shift register just causes the value in the first register to shift down, filling all the registers, as is currently the case. When would this be the desired behavior? 

My intended use of the enable signal is to enable or disable the shifting for a cycle.

aswaterman:
I see your point.  I guess what I'm describing is Pipe, not ShiftRegister.

davidbiancolin:
Why not also remove the n == 1 case now that this recurses more regularly?

aswaterman:
Agreed.

ccelio:
This is an unacceptable change to the Chisel API.  In the future please adhere to a deprecation schedule.

aswaterman:
In retrospect, I completely agree, @ccelio.  Sorry about that.

PR #371, opened 2016-11-17
Check that concrete implementations of Module make io a [lazy] val

jackkoenig:
Also add test to ensure Chisel allows users to rely on lazy semantics of a lazy val io

This already errors, but in Chisel._ it was giving a None.get in the emission. chisel3._ gives the error message that IO is already defined (and this change doesn't overrule that) which imo isn't as useful of a message in this case, but instances of this problem should be rare.

Since io being a def is a public API, who knows how many projects rely on it. rocket-chip's uncore uses this API. Power users can do it without messing up, this check will just dynamically check to be sure.


jackkoenig:
Yes but this public API has already been used by rocket-chip and conceivably other people, changing it to a val breaks those use cases.

sdtwigg:
But, the implementation very particularly requires that io remain stable.

If people are defining these things with defs then the only way their code can actually be working properly (and/or sanely) is if those defs were always returning precisely the same thing. Thus, the code can trivially be converted to a [lazy] val.

jackkoenig:
In the case of rocket-chip, `def io` is used for mixins, the concrete implementations use [lazy] val.

When I changed `io` to be a `val` in Module, I got errors like the following in rocket-chip:
```scala
[error] /scratch/jack/rocket-chip-clean/src/main/scala/uncore/agents/Trackers.scala:639: overriding value io in class Module of type chisel3.core.Bundle;
[error]  method io in trait RoutesInParent of type => uncore.agents.HierarchicalXactTrackerIO needs to be a stable, immutable value
[error] abstract class VoluntaryReleaseTracker(val trackerId: Int)(implicit p: Parameters) extends XactTracker()(p)
[error]                ^
```

So, trying to change these instances of `def io` to be `vals`:

Consider: https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/uncore/agents/Trackers.scala#L416

```scala
trait RoutesInParent extends HasBlockAddressBuffer
    with HasXactTrackerStates {
  def io: HierarchicalXactTrackerIO
  type AddrComparison = HasCacheBlockAddress => Bool
  def exactAddrMatch(a: HasCacheBlockAddress): Bool = a.conflicts(xact_addr_block)
  def routeInParent(iacqMatches: AddrComparison = exactAddrMatch,
            irelMatches: AddrComparison = exactAddrMatch,
            oprbMatches: AddrComparison = exactAddrMatch,
            iacqCanAlloc: Bool = Bool(false),
            irelCanAlloc: Bool = Bool(false),
            oprbCanAlloc: Bool = Bool(false)) {
    io.alloc.iacq.matches := (state =/= s_idle) && iacqMatches(io.iacq())
    io.alloc.irel.matches := (state =/= s_idle) && irelMatches(io.irel())
    io.alloc.oprb.matches := (state =/= s_idle) && oprbMatches(io.oprb())
    io.alloc.iacq.can := state === s_idle && iacqCanAlloc
    io.alloc.irel.can := state === s_idle && irelCanAlloc
    io.alloc.oprb.can := state === s_idle && oprbCanAlloc
    io.alloc.addr_block := xact_addr_block
    io.alloc.idle := state === s_idle
  }
}
```

When I tried changing io in Module to be a val, and this def to be a lazy value, Scala complained: 

```
[error] /scratch/jack/rocket-chip-clean/src/main/scala/uncore/agents/Trackers.scala:418: lazy values may not be abstract
[error]   lazy val io: HierarchicalXactTrackerIO
[error]            ^
```

Perhaps there's an easy solution to this problem. In any case, this is a powerful pattern that we should support.

sdtwigg:
You can just continue chaining as val io. It is just the ultimate concrete definition must have lazy val. (Assuming a superclass is even accessing the io in its constructor).

aswaterman:
I still vote for keeping @jackkoenig's version of the change.  People will rarely get this wrong, so having it be a runtime check instead of a compile-time check doesn't hurt productivity too much.  And there's no easy way to make this backwards compatible in the compatibility wrapper.

aswaterman:
You could use ne instead of != in case someone gets cute and overrides equals.  Not that that will ever happen.

sdtwigg:
Yes, != and ne would be equivalent unless someone overrode equals in Chisel.

https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala#L76

PR #372, opened 2016-11-17
Standardize the One True Way of specifying literals

ducky64:
Implementation for #294, supersedes #338 

Does exactly what the title says, removes the deprecated ways of specifying literals / data type constructors from core and moves them into the package objects. The extended constructors will be eventually removed from the chisel3 package, while the one in Chisel (compatibility) will stay indefinitely.

In a nutshell, the new ways of constructing a literal / data type are:
- `value.U`, `value.asUInt` (auto-determined width)
- `value.U(width)`, `value.asUInt(width)` (explicit width)
- `UInt()` (inferred width)
- `UInt(width)`

where `value` is either a Int, BigInt, or String, and `width` is a Width (Widths can be created from Ints using x.W)

and similar constructors for Bits, SInt (x.S), and Bool (x.B)

A few regex conversion examples are provided that can automate the transformation: (in package.scala):
- `Bool\((true|false)\)` => $1.B
- `UInt\(width\s*=\s*(\d+|[_a-zA-Z][_0-9a-zA-Z]*)\)` => UInt($1.W)
- `(UInt|SInt|Bits).width\((\d+|[_a-zA-Z][_0-9a-zA-Z]*)\)` => $1($2.W)
- `(U|S)Int\((-?\d+|0[xX][0-9a-fA-F]+)\)` => $2.$1
- `UInt\((\d+|0[xX][0-9a-fA-F]+),\s*(?:width\s*=)?\s*(\d+|[_a-zA-Z][_0-9a-zA-Z]*)\)` => $1.U($2.W)
- `(UInt|SInt|Bool)\(([_a-zA-Z][_0-9a-zA-Z]*)\)` => $2.as$1
- `(UInt|SInt)\(([_a-zA-Z][_0-9a-zA-Z]*),\s*(?:width\s*=)?\s*(\d+|[_a-zA-Z][_0-9a-zA-Z]*)\)` => $2.as$1($3.W)

This also enables deprecation warnings in Chisel core and all the tests have been rewritten to the new style.

The old way of specifying Inputs/Outputs (`UInt(INPUT, ...)`) have also been removed completely from the chisel3 package. A compatibility shim remains in the Chisel package


ducky64:
Since this is a fairly sweeping change that modifies the API, I'd like approvals from people spanning at least two different organizations.

There's not much actually changed code here if you ignore the mostly mechanical replacements in utils and tests.

aswaterman:
I suspect that somewhere you've introduced a case where the default CompileOptions are getting pulled in, rather than the CompileOptions in the user's scope.

aswaterman:
Try commenting out CompileOptions.materialize, and you'll see that several places within core and ChiselUtil now fail to compile.  This means they're incorrectly failing to pass the user's CompileOptions through.

ducky64:
As for removal of CompileOptions.materialize, that behavior shows up on chisel3 master as well (and when it's uncommented out, rocket successfully elaborates). So while that may be a bug, it seems orthogonal to whatever bug is being introduced in this PR. Which is strange, since the literal constructors don't take compile options.

ucbjrl:
I'm investigating ...

ducky64:
Found it. Was a mistake in converting Log2 to the new style and there seems to have been no unit tests of that util (this is sadly true for many utils in general). Rocket seems to build as well as master now.

aswaterman:
Thanks for the detective work.

azidar:
Looks good to merge. I'd wait until the end of today, and if there are no objections, we should merge.

azidar:
In particular, does @sdtwigg have any thoughts.

terpstra:
This PR still breaks rocket-chip/master.
I just tried to bump chisel3 in rocket-chip and this causes run-asm-tests to hang.
Is there some sort of API change that chisel user's need to make?
I'll keep digging.

aswaterman:
It is not supposed to break anything that imports Chisel (as opposed to chisel3).  If that is the case, there is a regression here.

ducky64:
Try with #380, there was a bug in this PR where asUInt(int) may be misinterpreted as a chained apply.
(and if it works for you, feel free to also review that PR so it can be merged, it's only a few lines long...)

aswaterman:
@terpstra can you try #380?  I'll review the PR if it rectifies the problem.

aswaterman:
cool, thanks @ducky64 @terpstra

PR #374, opened 2016-11-17
Eliminate some doc warnings

ducky64:
So when you do `sbt publish-local` or `sbt doc` you can have three less warnings.

The one that I couldn't fix is in Data.getWidth in Data.scala, where it has `@throws java.util.NoSuchElementException` that scaladoc can't generate the link for (since NoSuchElementException isn't in Chisel). One possible workaround is to disable all link warnings, but a proper clean fix would probably need to be in the scaladoc generator itself.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Binding.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Binding.scala
@@ -140,10 +140,9 @@ object Binding {
       }
     }
 
-    /** Diagnose a binding error caused by a missing IO() wrapper.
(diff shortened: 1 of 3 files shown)


ducky64:
Something about an unmoored doc comment, I couldn't find too much in-depth information anywhere.
It seems https://github.com/clulab/processors/pull/94/files had the same issue and their solution was to turn it into non-scaladoc-like comments. I'm guessing it's because this is an inner function which scaladoc doesn't generate files for, so it can't figure out what the comment should be attached to.

aswaterman:
Ah, I see.

PR #376, opened 2016-11-21
Have checkpoint report (and clear) non-fatal errors.

ucbjrl:


ducky64:
Style: can you use a common function for the two? It seems one handles a fatal path while the other isn't fatal, shouldn't the printing be common code? Right now it looks like two wildly separate code paths.

ucbjrl:
I'm afraid trying to come up with common code to handle the printing will just make things more complicated with little benefit. Feel free to prove me wrong.

ucbjrl:
retest this please

PR #380, opened 2016-11-22
Disallow chained apply

### LINKED ISSUE #377: Int.asUInt(Int) Doesn't Work
```scala
class Test extends Module {
  val io = IO(new Bundle {
    val out = UInt(64.W)
  })
  io.out := 0.asUInt(64)
}
```

I forgot the `.W` on the width of the literal `0`, this just gives me a literal of width 1 instead of 64

```
circuit Test :
  module Test :
    input clock : Clock
    input reset : UInt<1>
    output io : {out : UInt<64>}

    io is invalid
    io is invalid
    io.out <= UInt<1>("h00") @[Test.scala 8:10]
```

Also, io is invalidated twice, perhaps a separate issue.

ducky64:
Should fix #377


terpstra:
Yes. Applying this PR fixes the problem I had with #372 in rocket-chip.

PR #383, opened 2016-11-23
Add PerformancePathologySpec for finding performance bugs

jackkoenig:
Add DeduplicationSpec as first example of this kind of test

This will fail until https://github.com/ucb-bar/firrtl/pull/375 is merged and issue pointed out in https://github.com/ucb-bar/firrtl/pull/375#issuecomment-262149561 is resolved.

The basic idea is that certain things should scale linearly. This PR provides a simple mechanism for writing tests that check some possible performance pathology by putting an upper bound on how long it should take. It's a very rough way to do this, and is certainly not a replacement for real performance regression testing. However, it could serve as a useful sanity check.


colinschmidt:
Glad to see de-dupe and performance tested!

src/test/scala/chiselTests/DeduplicationSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+  def timeout: Int = 10
+  def warmupConfig = () => new DedupQueue(8)
+  def benchmarkConfig = () => new DedupQueue(128)
+
+  // Check Verilog result
+  it should "actually do deduplication" in {
+    val ModuleRegex = """\s*module\s+(\w+)\b.*""".r
+
+    val modules = verilogResult split "\n" collect { case ModuleRegex(name) => name }
+    assert(modules.size === 2, "Deduplication must actually happen!")
```
colinschmidt: Since deduplication is now in firrtl I think this should be moved there.

Issue #373, opened 2016-11-17
Enum deficiencies


ducky64:
- Enum values can't be used in testing, since they only return a Chisel-land data type rather than a Scala-land Int.
- nodeType may be meaningless / boilerplate - I don't see a case for anything other than UInt. That nodeType exists also forces Chisel to keep the `Bits.fromInt` method which isn't exposed or used elsewhere.
- Enum values aren't type-safe, they return generic Chisel types like UInt which aren't treated differently. Whether this is desirable is an open question, much like the Scala debate between Enumeration and case objects

Thoughts?

aswaterman:
Re: point 2: I'd favor creating Enum(n) which implicitly does the same thing as Enum(UInt(log2Ceil(n)), n) currently does.  You could then deprecate Enum(nodeType, n) in new-chisel3.  I agree, it's not really obvious what an SInt enum would do, since there's no facility to start at anything besides 0.

(In C, having signed enums is occasionally useful, because you're allowed to set the start number to something negative.  I don't see a compelling reason to follow their lead in this case, though.)

aswaterman:
Re: the other points: despite its downsides, it's still useful.

ducky64:
Yes, we definitely do need to keep the current Enum (perhaps with nodeType deprecated, and assertion-checked to be a UInt so we can remove Bits.fromInt), but it would be interesting to think of what a more featured enum system would look like.

ducky64:
So in the attempt to deprecate old style enums I came across some more issues:
- The Symbol-based enum generators (`def apply[T <: Bits](nodeType: T, l: Symbol *)` and `def apply[T <: Bits](nodeType: T, l: List[Symbol])` are *really old* code, and are used in neither chisel3 tests nor rocket-chip. Since they're not compile-time-checked and (in my opinion) strictly inferior to the list one, I'm thinking of removing those completely (without even a deprecation phase).
- Technically, enums can take a nodeType of specified width, though this doesn't see use in rocket either, and it's unclear whether this would ever be a good idea (since the user has no control over the numeric values). I'm thinking of dropping this feature and having chisel3 assert out if a width is specified.

Thoughts?

mwachs5:
I have to say that as a new Chisel user, the first point, 
"Enum values can't be used in testing, since they only return a Chisel-land data type rather than a Scala-land Int." immediately bit me, and made me wonder what the point of Enums is. 

[...]

PR #385, opened 2016-11-23
Simplify Enum API

ducky64:
Get rid of some cruft exposed in #373 

This also allows `Bits.fromtInt(...)` to be removed. Yay!

All old APIs (with some new restrictions, rocket still works fine) are preserved without deprecation in `Chisel._`, aside from the non-compile-time-checkable Map[] enum constructor which probably should have been deprecated during chisel2. The Map[] enums have been removed from chisel3._ without deprecation.

The new restriction is that `nodeType` (legacy API) may only be of `UInt` type with unspecified width. Note that `Bits()` creates a `UInt`, and if you can't control the enum values, it makes little sense to specify a bitwidth.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -54,8 +54,6 @@ sealed abstract class Bits(width: Width, override val litArg: Option[LitArg])
   // Arguments for: self-checking code (can't do arithmetic on bits)
   // Arguments against: generates down to a FIRRTL UInt anyways
 
-  private[chisel3] def fromInt(x: BigInt, w: Int): this.type
-
   private[chisel3] def flatten: IndexedSeq[Bits] = IndexedSeq(this)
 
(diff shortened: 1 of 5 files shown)


ducky64:
Scala can't check a map access during compile time (unlike the List-based version, which turn into vals that the compiler understands). This just nags people who are still using that to do something safer, I have no plans to actually remove those elements from the compatibility layer.

If you're super against it I'll remove the deprecation tags.

aswaterman:
I was just trying to understand the rationale.  I guess I agree they should be deprecated, even in the compatibility layer.

ducky64:
I'll document the rationale and that those are not to be removed.

seldridge:
I'm only just seeing this from a rocket-chip bump on my end. 

I'm against the deprecation tags here. Unless something has been fixed upstream with Scala, the Tuple-based approach for Enums hits Scala's limit of 22 on Tuple sizes. The only way that I know of to get around this is to use the List-based version (ignoring the fact that I idiotically have a legacy state machine with > 22 states...).

Thanks for keeping this support in, nonetheless!

PR #389, opened 2016-12-01
Fix literal width

### LINKED ISSUE #388: Weird Width Decisions
Consider:

```scala
class Literals extends Module {
  val io = IO(new Bundle {
    val out1 = Output(UInt(32.W))
    val out2 = Output(UInt(32.W))
  })

  io.out1 := "h_dead_beef".U
  io.out2 := "hdeadbeef".U

}
```

This results in the following Firrtl:

```
circuit Literals :
  module Literals :
    input clock : Clock
    input reset : UInt<1>
    output io : {out1 : UInt<32>, out2 : UInt<32>}

    io is invalid
    io is invalid
    io.out1 <= UInt<40>("h0deadbeef") @[Literals.scala 10:11]
    io.out2 <= UInt<32>("h0deadbeef") @[Literals.scala 11:11]
```

It looks like the underscores in "h_dead_beef" are contributing to the chosen width.

jackkoenig:
Fixes #388 

Also add sanity checks

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/package.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/package.scala
@@ -55,14 +55,17 @@ package chisel3 {
     implicit class fromStringToLiteral(val str: String) {
       /** String to UInt parse, recommended style for constants.
         */
-      def U(): UInt = UInt.Lit(parse(str), parsedWidth(str))  // scalastyle:ignore method.name
+      def U(): UInt = str.asUInt() // scalastyle:ignore method.name
       /** String to UInt parse with specified width, recommended style for constants.
         */
-      def U(width: Width): UInt = UInt.Lit(parse(str), width)  // scalastyle:ignore method.name
+      def U(width: Width): UInt = str.asUInt(width) // scalastyle:ignore method.name
 
       /** String to UInt parse, recommended style for variables.
         */
-      def asUInt(): UInt = UInt.Lit(parse(str), parsedWidth(str))
+      def asUInt(): UInt = {
+        val bigInt = parse(str)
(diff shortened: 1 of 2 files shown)


ucbjrl:
retest this please

chiselFrontend/src/main/scala/chisel3/core/package.scala
```diff
@@ -55,14 +55,17 @@ package chisel3 {
...
         */
-      def U(width: Width): UInt = UInt.Lit(parse(str), width)  // scalastyle:ignore method.name
+      def U(width: Width): UInt = str.asUInt(width) // scalastyle:ignore method.name
 
       /** String to UInt parse, recommended style for variables.
         */
-      def asUInt(): UInt = UInt.Lit(parse(str), parsedWidth(str))
+      def asUInt(): UInt = {
+        val bigInt = parse(str)
+        UInt.Lit(bigInt, Width(bigInt.bitLength))
```
ucbjrl: This causes DecodeSpec failures. The (suppressed) message is:

      Message: requirement failed: The literal value 0 was elaborated with a specificed width of 0 bits, but at least 1 bits are required.

and this is because `bigInt.bitLength` is reporting a value of 0 for `0`. According to [BigInt](http://www.scala-lang.org/api/current/scala/math/BigInt.html) bitLength "Returns the number of bits in the minimal two's-complement representation of this BigInt, excluding a sign bit."

The line should be:

        UInt.Lit(bigInt, Width(math.max(bigInt.bitLength, 1)))

jackkoenig: Fixed and test added

PR #390, opened 2016-12-01
ScalaCheck may shrink to fail DecoderSpec.

ucbjrl:
We either need to set constraints on the nPairs generator so it doesn't generate an empty list, or ignore empty lists when running tests. This was triggered by a recent Jenkins failure:

    [info] - BitPat wildcards should be usable in decoding *** FAILED ***
    [info]   IllegalArgumentException was thrown during property evaluation.
    [info]     Message: requirement failed
    [info]     Occurred when passed generated values (
    [info]       arg0 = List() // 3 shrinks
    [info]     )

### DIFF
--- a/src/test/scala/chiselTests/Decoder.scala
+++ b/src/test/scala/chiselTests/Decoder.scala
@@ -33,6 +33,9 @@ class DecoderTester(pairs: List[(String, String)]) extends BasicTester {
 }
 
 class DecoderSpec extends ChiselPropSpec {
+  // Inhibit shrinking for this test, otherwise, we may end up generating an empty List of pairs which is guaranteed to fail.
+  import org.scalacheck.Shrink
+  implicit val noShrink: Shrink[Int] = Shrink.shrinkAny
 
   // Use a single Int to make both a specific instruction and a BitPat that will match it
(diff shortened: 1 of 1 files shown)


jackkoenig:
Doesn't ScalaCheck only shrink if it finds a failing input to begin with? Agreed though that this shouldn't be shrinking since we're explicitly generating lists of size 4. I *think* we could alternatively use whenever inside of the forAll property to prevent it trying to shrink it, might be wrong though.

ucbjrl:
This is due to #389. The (suppressed) message is:

      Message: requirement failed: The literal value 0 was elaborated with a specificed width of 0 bits, but at least 1 bits are required.

and this is because `bigInt.bitLength` is reporting a value of 0 for `0`. According to [BigInt](http://www.scala-lang.org/api/current/scala/math/BigInt.html) bitLength "Returns the number of bits in the minimal two's-complement representation of this BigInt, excluding a sign bit."

I've added a similar comment to that pr.

ucbjrl:
This turns out to be incorrect or insufficient and shrinking isn't inhibited. Since the error will be fixed by fixing #389, I'm closing this pr.

PR #391, opened 2016-12-01
Have FixedPoint use asFixedPoint in fromBits.

grebe:
In reference to ucb-bar/firrtl#389

### DIFF
--- a/src/test/scala/chiselTests/FixedPointSpec.scala
+++ b/src/test/scala/chiselTests/FixedPointSpec.scala
@@ -17,6 +17,22 @@ class FixedPointSpec extends FlatSpec with Matchers {
 
     initialDouble should be(finalDouble)
   }
+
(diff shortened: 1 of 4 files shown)


src/test/scala/chiselTests/FixedPointSpec.scala
```diff
@@ -20,7 +20,7 @@ class FixedPointLiteralSpec extends FlatSpec with Matchers {
 }
 
 class FixedPointFromBitsTester extends BasicTester {
-    val uint = 3.U
+    val uint = 3.U(4.W)
```
grebe: This width is necessary, but it shouldn't be. This shouldn't block the PR b/c it is a separate problem though

chick:
@azidar  can you remove your block on this, or comment on what else is needed

PR #397, opened 2016-12-02
Add pointer to ScalaDoc.

ucbjrl:


ucbjrl:
Should we move this to github? It was suggested the firrtl equivalent should live on github.

jackkoenig:
If we can do it as a github page or whatever I don't see why not. I think there is also a tool that generates Markdown from Scaladoc but I think it's less powerful than the html version

ucbjrl:
Check out [sbt-ghpages](https://github.com/sbt/sbt-ghpages). I believe we already load the required plugins.

PR #398, opened 2016-12-02
utils scaladoc examples for BitPat through CircuitMath

ducky64:
Arbiter already has examples.

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -35,16 +35,10 @@ object BitPat {
   }
 
   /** Creates a [[BitPat]] literal from a string.
-   *
+    *
     * @param n the literal value as a string, in binary, prefixed with 'b'
     * @note legal characters are '0', '1', and '?', as well as '_' and white
     * space (which are ignored)
-    *
-    * @example {{{
-    * "b10101".U === BitPat("b101??") // dynamically evaluates to true.B
-    * "b10111".U === BitPat("b101??") // dynamically evaluates to true.B
-    * "b10001".U === BitPat("b101??") // dynamically evaluates to false.B
-    * }}}
     */
   def apply(n: String): BitPat = {
     val (bits, mask, width) = parse(n)
@@ -85,8 +79,13 @@ object BitPat {
 }
 
 /** Bit patterns are literals with masks, used to represent values with don't
-  * cares. Equality comparisons will ignore don't care bits (for example,
-  * BitPat(0b10?1) === 0b1001.asUInt and 0b1011.asUInt.
+  * care bits. Equality comparisons will ignore don't care bits.
+  *
+  * @example {{{
+  * "b10101".U === BitPat("b101??") // dynamically evaluates to true.B
+  * "b10111".U === BitPat("b101??") // dynamically evaluates to true.B
+  * "b10001".U === BitPat("b101??") // dynamically evaluates to false.B
+  * }}}
   */
 sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) {
   def getWidth: Int = width
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -82,9 +82,9 @@ object BitPat {
   * care bits. Equality comparisons will ignore don't care bits.
   *
   * @example {{{
-  * "b10101".U === BitPat("b101??") // dynamically evaluates to true.B
-  * "b10111".U === BitPat("b101??") // dynamically evaluates to true.B
-  * "b10001".U === BitPat("b101??") // dynamically evaluates to false.B
+  * "b10101".U === BitPat("b101??") // evaluates to true.B
+  * "b10111".U === BitPat("b101??") // evaluates to true.B
+  * "b10001".U === BitPat("b101??") // evaluates to false.B
   * }}}
   */
 sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) {
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -39,13 +39,24 @@ object BitPat {
     * @param n the literal value as a string, in binary, prefixed with 'b'
     * @note legal characters are '0', '1', and '?', as well as '_' and white
     * space (which are ignored)
+    * 
+    * @example {{{
+    * "0b00101".U === BitPat("b001??") // dynamically evaluates to true.B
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/util/Bitwise.scala
```diff
@@ -73,7 +103,14 @@ object Reverse {
       Cat(doit(in(half-1,0), half), doit(in(length-1,half), length-half))
     }
   }
+  
   /** Returns the input in bit-reversed order. Useful for little/big-endian conversion.
+    *  
+    * @example {{{
+    * Reverse("b1101".U)  // equivalent to "b1011".U
+    * Reverse("b1101".U(8))  // equivalent to "b10110000".U
```
jackkoenig: This should be `"b1101".U(8.W)` since due to aliasing with apply we don't allow Ints as arguments

Also, stylistically, I prefer using `.asUInt(width.W)` when giving a width to a literal (and `.U` when not), but that's just my opinion

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -39,13 +39,24 @@ object BitPat {
     * @param n the literal value as a string, in binary, prefixed with 'b'
     * @note legal characters are '0', '1', and '?', as well as '_' and white
     * space (which are ignored)
+    * 
+    * @example {{{
+    * "0b00101".U === BitPat("b001??") // dynamically evaluates to true.B
+    * "0b00111".U === BitPat("b001??") // dynamically evaluates to true.B
+    * "0b00001".U === BitPat("b001??") // dynamically evaluates to false.B
```
jackkoenig: Is there a reason these have leading zeros before the b?

ducky64:
Darn, I thought I caught all of those. Anyways, fixed now, thanks!

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -39,13 +39,24 @@ object BitPat {
     * @param n the literal value as a string, in binary, prefixed with 'b'
     * @note legal characters are '0', '1', and '?', as well as '_' and white
     * space (which are ignored)
+    *
+    * @example {{{
+    * "0b10101".U === BitPat("b101??") // dynamically evaluates to true.B
+    * "0b10111".U === BitPat("b101??") // dynamically evaluates to true.B
+    * "0b10001".U === BitPat("b101??") // dynamically evaluates to false.B
```
jackkoenig: The leading zeros need to be removed, eg.`"b10101".U`. These examples cause runtime exceptions.

ducky64:
I see what you mean now. derp.

jackkoenig:
Haha no worries

jackkoenig:
One thing I think we should keep in mind is that while documenting the apply methods is really important, when people look at the API documentation, the first thing they see is the class, not the companion object. People new to Scala won't think to click on the companion object so I think it would be prudent to put our examples on the class itself in addition to or rather than on the apply methods. See List for example:  http://www.scala-lang.org/api/current/scala/collection/immutable/List.html

jackkoenig:
Sorry, I just thought of this when I used this branch to generate the ScalaDoc and am looking at it.

ducky64:
So I think a main problem is that the apply methods could have different APIs, so it doesn't really make sense to have it in the class. Furthermore, a lot of classes really shouldn't be instantiated by users.

What do you think about adding something to the top of the class scaladoc to the effect of "Users: perhaps you were looking for [the companion object]?" or "Users: are you looking for how to use [link to companion object] instead?
(but that's probably something for another PR)

jackkoenig:
I mean, [specifically looking at Scala List as an example](http://www.scala-lang.org/api/current/scala/collection/immutable/List.html), they put examples of using companion object apply and unapply in the ScalaDoc for the class. Object List and its apply method do not have any examples, just explanation of what the arguments are. The primary issue I'm trying to address is that many people when they open the API aren't trying to find the list of all functions or apply methods and people who aren't super familiar with Scala may not even be able to tell the difference between different apply methods (or know what a companion object is!). Rather, when someone opens the API and clicks on BitPat (which takes them to the class), they are most likely wanting to know: "What is this and/or how do I use it?"

Obviously when there are multiple apply methods with different APIs, we have to be careful how we write the examples. Using Vec as an example, on the class Vec (which has a private constructor so you can't even instantiate the class directly), we could have the example:

```scala
// Create a register of type Vec
val myRegOfVec = Reg(Vec(8, UInt(32.W)))

// Create a ROM (note that this creates a Wire)
val rom = Vec(Seq(0.U, 1.U, 2.U, 3.U))

// etc.
```

I think that concrete examples like that are useful and should be placed on the class itself, not on the apply methods since I don't think even intermediate Chisel users are going to get much out of object Vec:
screen shot 2016-12-06 at 1 59 33 pm

ducky64:
Yeah, that's fair. Will move all examples from object apply into class scaladoc.

jackkoenig:
As a completely unrelated question, is it possible for us to hide the macro do_ functions in the ScalaDoc?

ducky64:
Possibly, depending on how ScalaDoc parsing works. One method may be to use a StaticAnnotation (macro) to generate the do_ behind the scenes, such that if ScalaDoc is parsing the raw file it will not see the macro expansions. This would be a general code style improvement, but the implementation details are still uncertain.
Otherwise, a quick search turns up no method for excluding class methods / members, though apparently packages can be excluded (https://issues.scala-lang.org/browse/SI-5533).

ducky64:
How's this? I've moved the examples for the apply methods into the class scaladoc, and pruned documentation from the apply method where it doesn't add any additional information.

src/main/scala/chisel3/util/Bitwise.scala
```diff
@@ -23,14 +33,31 @@ object FillInterleaved {
...
 
-/** Returns the number of bits set (i.e value is 1) in the input signal.
+/** Returns the number of bits set (value is 1 or true) in the input signal.
+  *
+  * @example {{{
+  * PopCount(Seq(true.B, false.B, true.B, true.B))  // dynamically evaluates to 3.U
+  * PopCount(Seq(false.B, false.B, true.B, false.B))  // dynamically evaluates to 1.U
+  *
+  * PopCount("b1011".U)  // dynamically evaluates to 3.U
+  * PopCount("b0010".U)  // dynamically evaluates to 1.U
```
jackkoenig: I think it would be better to just say what they evaluate to without "dynamically"

jackkoenig:
Looks great!

src/main/scala/chisel3/util/Bitwise.scala
ducky64: Good point. It was meant as a generalization that this could generate dynamic circuitry, but I guess the example below addresses that. Removed unnecessary dynamic references.

PR #404, opened 2016-12-12
Add Cookbook examples Reg of Vec and FSM

jackkoenig:


src/test/scala/cookbook/RegOfVec.scala
```diff
@@ -0,0 +1,26 @@
...
+import chisel3._
+
+/* ### How do I create a Reg of type Vec?
+ *
+ * For information, please see the API documentation
+ * (https://chisel.eecs.berkeley.edu/api/index.html#chisel3.core.Vec)
+ */
+class RegOfVec extends CookbookTester(2) {
+  // Reg of Vec of 32-bit UInts
+  val regOfVec = Reg(Vec(4, UInt(32.W)))
```
jackkoenig: I was planning to put that in the api documentation which this links to, do you think it should be here as well?

edwardcwang: I guess if we really wanted to we could just include both (one version with just the reg and another with a initialized reg). I got the impression that initializing Reg was good practice but I might be mistaken.

jackkoenig: I can add it here, it just requires a little more explanation since it requires using a Vec.apply method that constructs a Wire (which is kind of confusing in the *type* vs *value* distinction that I'm trying to make more explicit). It's probably wise to have that example though since it is a cookbook after all.

jackkoenig:
@edwardcwang Alright I added an example with a small explanation, thoughts?

edwardcwang:
It looks good, am I allowed to approve?

jackkoenig:
Sure, I'll Squash and merge after approval

PR #406, opened 2016-12-14
Default to no dequeue in Queue factory methods

jackkoenig:
In my opinion, Queue factory methods should default the Queue output to no dequeuing

Most of the time Queue.apply is used the result is instantly bulk connected to something else, this has no effect on those uses (except I guess an extra line of Firrtl that goes away).

However, other use cases where some Decoupled input is being buffered and then used, it seems natural that it should default to "no dequeue" the output of the Queue.

eg.
```scala
val myQ = Queue(io.input)
myQ.nodeq() // equivalent to myQ.ready := false.B  <- This line seems super unnecessary
when (cond) {
  val bits = myQ.deq()
  // Do something with bits
}

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -248,6 +248,7 @@ object Queue
     q.io.enq.valid := enq.valid // not using <> so that override is allowed
     q.io.enq.bits := enq.bits
     enq.ready := q.io.enq.ready
+    q.io.deq.ready := false.B  // Default to no dequeue
     TransitName(q.io.deq, q)
   }
 
@@ -266,6 +267,7 @@ object Queue
     irr.bits := deq.bits
     irr.valid := deq.valid
     deq.ready := irr.ready
+    irr.ready := false.B
     irr
   }
 }

### CODE REVIEW

**Comment by aswaterman** [comment]
It's a little weird that Queue returns a Decoupled whose ready signal defaults to false, but other APIs that return Decoupled don't.

**Comment by jackkoenig** [comment]
What else returns Decoupled? `git grep ':\s*Decoupled'` on chisel3 only shows things in Decoupled.scala, I guess the actually Decoupled companion object returns DecoupledIO but I don't see anything else

**Comment by terpstra** [comment]
This would be very asymmetric with use of Decoupled. It's a very common idiom to write:
```scala
  val foo = Wire(some_decoupled)
  // connect source
  foo.valid := something
  // connect sink
  foo.ready := something else
```
So why should a Queue() behave differently from a Wire?

jackkoenig:
This change would not preclude using Queue in the same way since `foo.ready := somethingElse` will overrule the default assignment of ready to false.

There is also already a huge asymmetry between Wire and Queue in that `Wire(someDecoupled)` creates a Wire with the same type as `someDecoupled` while `Queue(someDecoupled)` will construct a Queue of type `someDecoupled.bits` and connect `someDecoupled` to the queue's input.

sdtwigg:
It isn't common for other classes to initialize themselves with defaults so it seems odd to start here. Upon review, people would have to remember that 'that Queue' is special and doesn't need a default (since it is already baked in).

jackkoenig:
The consensus seems to be that this is either unintuitive and/or inconsistent so I'm going to close this PR

PR #407, opened 2016-12-14
Private vals

jackkoenig:
After looking at the public API for Queue (see all of the public vals https://chisel.eecs.berkeley.edu/api/index.html#chisel3.util.Queue), I wanted to clean it up a little bit.

So it turns out that Chisel already names private vals, Scala just gives them really crazy names (like chisel3$util$Queue$$ram) so I wrote a function to clean these names up. I have not been able to figure out a way to detect if something is a private val via java reflection so this cleanup is my hacky alternative. Suggestions welcome.

I also changed all of the public fields in util that I could find that have no business being public (since accessing them externally would result in invalid Firrtl)

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -199,8 +199,13 @@ extends HasId {
           }
         case _ => // Do nothing
       }
+    /** Scala generates names like chisel3$util$Queue$$ram for private vals
+      * This extracts the part after $$ for names like this and leaves names
+      * without $$ unchanged
+      */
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -199,8 +199,16 @@ extends HasId {
           }
         case _ => // Do nothing
       }
+    /** Scala generates names like chisel3$util$Queue$$ram for private vals
+      * This extracts the part after $$ for names like this and leaves names
+      * without $$ unchanged
+      */
+    def cleanName(name: String): String = {
+      val split = name.split("\\$\\$")
```
chick: Could be single lined as 
```
name.split("""\$\$""").lastOption.getOrElse(name)
```

PR #422, opened 2016-12-21
Add Analog type

jackkoenig:
Used for stitching Verilog inout through Chisel Modules (from BlackBox
to BlackBox)

This isn't quite done, but I'm PRing it because there is a lot to discuss. All this PR really does is piggyback on the Analog type in Firrtl https://github.com/ucb-bar/firrtl/pull/295, tactically inserting Wires whenever those would be required in the resulting Firrtl. This leads to a few problems and exposes some of the restrictions present in the Firrtl implementation. The tests illustrate some of the problems but I will try to summarize here.

The current proposal is to add a new type `Analog` that extends `Element`. It is intended to allow stitching inouts from Blackboxes through Chisel designs. 

Questions/Problems:

1. Is `Analog` the right name? Are there other features this type should ultimately support? This is especially relevant if other `Analog` things have direction.
1. What is the future of this feature? Will Chisel ever "fully" support inout? Should we provide some "Analog driver generator" that splits an Analog into an `input`, `output`, and `enable` (or `drive`) ?
1. Should MonoConnect (`:=`) be supported? I've always thought some directionality was implied there but Analog does not have direction
1. What should happen if you assign a direction to an Analog type?
    * eg. `val bus = Input(Analog(32.W))`
    * Should it become a UInt of the appropriate direction?
        * This might make some sense since Analog does not really fit into Chisel's type and direction system
    * Should it be an error?
        * This would prevent people taking some Bundle with an Analog in it and wrapping in `Input(...)` or `Output(...`) (although I presume `Flipped(...)` would still work)
    * Perhaps Analog should not be a type but rather a direction? (This would open a whole different can of worms)
1. What do last connect semantics mean for Analog?
    For example:
    ```scala
    val busWire = Wire(Analog(32.W))
    busWire <> mod1.io.bus
    busWire <> mod2.io.bus
    busWire <> mod3.io.bus
    ```
    Will connect mod1's, mod2's, and mod3's busses together.
    ```scala
    mod1.io.bus <> mod2.io.bus
    ```
    Will connect mod1's and mod2's busses together.
    ```scala
    mod1.io.bus <> mod2.io.bus // This gets overruled by the next connection!
    mod2.io.bus <> mod3.io.bus
    ```
    Will **only** connect mod2's and mod3's busses together
1. It is impossible to connect two Analog Ports within a Module
    For example:
    ```scala
    class InoutConnector extends Module {
      val io = IO(new Bundle {
        val bus1 = Input(Analog(32.W))
        val bus2 = Input(Analog(32.W))
      })
      io.bus1 <> io.bus2
    }
    ```
    ~~The only possible way to express this in Verilog as far as I know is with a "via" Module~~, eg.
    ```verilog
    module via (w, w);
      inout [31:0] w;
    endmodule

    module InoutConnector(
      input   clock,
      input   reset,
      inout  [31:0] io_bus1,
      inout  [31:0] io_bus2
    );
      via via_inst(io_bus1, io_bus2);
    endmodule
    ```
    However, this does not seem to be supported by Verilator since it requires redefinition of a port. This *does* seem to be supported by VCS though.
    Related discussion: https://groups.google.com/forum/#!topic/comp.lang.verilog/5enmkDcC4Uk
    SystemVerilog `alias` should work for this but is not supported in Verilator, see below comment.
1. Conditional connections of Analog should probably be an error, however in the tests I provide an example where this should error but it actually passes. 
    * This is due to the connection needing to create a wire (which is thus within the same when scope) which in Firrtl makes the connection unconditional.
    * This can be solved by:
        1. Change the semantics of Firrtl so the wire isn't necessary, then just emit the connection so Firrtl can error
        1. Have Chisel check the when scope and error if an Analog is connected within a when scope.
    * Either way, this creates the restriction that any bundle with an analog cannot be bulk connected in a when scope


jackkoenig:
Per (6), the following seems to work in Verilator but **not** in VCS

```verilog
module InoutConnector(
  input   clock,
  input   reset,
  inout  [31:0] io_bus1,
  inout  [31:0] io_bus2
);
  wire [31:0] io_bus1_p;
  wire [31:0] io_bus2_p;
  assign io_bus1_p = io_bus2;
  assign io_bus2_p = io_bus1;
  assign io_bus1 = io_bus1_p;
  assign io_bus2 = io_bus2_p;
endmodule
```

There is also the `alias` SystemVerilog 2005 construct but this is not supported by Verilator

edwardcwang:
General comment: doesn't this seem to be a bit like a BlackboxWire or AbstractWire? An idea would be to do things like generator, breaking tri-state into enable + data, etc by having some kind of abstract wire and then extending on top of that for various different types of wire, which might be nicer in terms of abstraction/cleanliness.

scottj97:
I checked Cadence Incisive and Mentor Questa and found behavior identical to VCS; that is, the via and alias mechanisms both worked, but the assign method led to x's on the bus.

aswaterman:
The ifdef-VERILATOR approach seems most unsavory.  Maybe it's pragmatic in the short term.  Maybe we should try to get Verilator to implement alias.

@edwardcwang indeed, tri-state buses are only one of the potential uses.  IMO, the Analog name captures the notion pretty well.

sdtwigg:
The assign method was derived from this Verilator post:
http://www.veripool.org/boards/3/topics/1041-Verilator-Implementing-SystemVerilog-alias-

so it looks like they had spent some time (3 years ago) looking at it. Odd that the Verilator AST seems to already have a type for alias but didn't finish the feature request.

ducky64:
My thoughts on your questions, in terms of design decisions:
1. Analog seems like a very general supertype for all signals. Wire also would work, though unfortunately we already have a `Wire()` construct and shouldn't add to the confusion.
2. My interpretation is that Analog is a general escape hatch for all the fancy things you could do, but chisel doesn't (yet) support. I imagine you could implement all your digital logic in Analog too - so if we add inout/enable/etc as base chisel types in the future, current implementations using Analog to implement those would coexist, just as less clean code.
3. I think this depends on the resolution to 4.
4. Make it match conventions used in current data types - which (I think) is to require directions at least on IO boundaries. Perhaps add a Undirectioned / Bidirectional construct (where only bidirectional - bidirectional connects are legal) while the other connects stay the same?
5. This is tricky. I think in chisel, last-connect semantics are usually used for mux generation when you do a reassign in a conditional, but that obviously doesn't make sense for Analog since chisel doesn't know enough to generate switch hardware. Last-connect also only works when you have some idea of directionality (and <> is interpreted as a non-directional connect). Ignoring the issue of directional connects in question 3, I would expect the third example to hook everything up together (or error out, if you think there's potential for confusion).
6. idk
7. idk

Implementation thoughts:
- We should re-examine moving flatten out of Data. As you've noticed it doesn't make sense at all for Analog, and barely any sense for Bits types.
- Consider formalizing the definition for the width parameter in comments.

src/test/scala/chiselTests/Inout.scala
```diff
@@ -0,0 +1,132 @@
...
+    assertTesterPasses(new InoutTester {
+      val mods = Seq(Module(new InoutReaderBlackBox), Module(new InoutReaderWrapper))
+      val busWire = Wire(writer.io.bus)
+      busWire <> writer.io.bus
+      busWire <> mods(0).io.bus
+      busWire <> mods(1).io.bus
+      mods.foreach(check(_))
+    }, Seq("/InoutBlackBox.v"))
+  }
+  // TODO This is not supported in Firrtl, should this be supported?
```
ducky64: Was there given rationale as to why this isn't supported in FIRRTL?

src/test/scala/chiselTests/Inout.scala
```diff
@@ -0,0 +1,132 @@
...
+      val reader = Module(new InoutReaderBlackBox)
+      val connector = Module(new InoutConnector)
+      connector.io.bus1 <> writer.io.bus
+      reader.io.bus <> connector.io.bus2
+      check(reader)
+    }, Seq("/InoutBlackBox.v"))
+  }
+  // TODO What should happen here?
+  // This creates a wire in order to connect the Analog types
+  // Since it is created within the when scope the connection is actually unconditional
```
ducky64: I'd argue for having this error out because Chisel doesn't have enough information to generate switch hardware, and it creates potential for confusion since chisel essentially is dropping the `when` which is almost certainly not the right (intuitive) thing.

src/test/scala/chiselTests/Inout.scala
```diff
@@ -0,0 +1,132 @@
...
+  })
+  io.bus1 <> io.bus2
+}
+
+// Parent class for tests connecing up InoutReaders and InoutWriters
+abstract class InoutTester extends BasicTester {
+  final val BusValue = "hdeadbeef".U
+
+  final val (cycle, done) = Counter(true.B, 2)
+  when (done) { stop(); stop() } // Double stop for Verilator
```
ducky64: Unless this double stop pertains to Analog in particular, this should be filed as an issue and get fixed at the root.

yunsup:
Here are my thoughts, feedback is welcome:

1. I'm not sure whether Analog is the best name, but can't think of a better name.
2. No. The way inout wires are split into input/output/enable seems too technology specific to make Chisel do the job. Resorting to users supplying their own blackboxes for a given technology to break an inout to whatever port seems like an acceptable solution.
3. I'm going back and forth on this one. I can argue that we can make a MonoConnect of an Analog wire be the mechanism of forcing a direction, and therefore casting it to a UInt. For example:
    ```
    val busWire = Wire(Analog(32.W))
    o := busWire (becomes an input)
    busWire := i (becomes an output)
    ```
    But at the same time, this can be a source of Zs and therefore Xs when the inout wire is treated as an input with no active drivers. I think there was ways to check whether there was a driver or not in Verilog -- if that's the case we can probably inject random values to prevent X propagation problems.
4. I vote for Analog types not having directionality, and generating Chisel errors.
5. Last connect should win
6. We have to support it correctly on both Verilator and commercial Verilog simulators. So we'll have to ifdef the code until we find a solution that works for both simulators. There's actually another case where we need the same connector (see the following example):
    ```
    val busWire0 = Wire(Analog(32.W))
    val busWire1 = Wire(Analog(32.W))
    busWire0 <> busWire1
    ```
7. Conditional connections should be an error

jackkoenig:
I've revamped this PR a bit. It works with https://github.com/ucb-bar/firrtl/pull/415 (but unfortunately that Firrtl implementation is not synthesizable by design compiler so I'm looking into it). In the mean time, I've improved the proposed Analog API a big.

(Note these numbers do not correspond to the ones before)
1. Analog cannot have direction
    * It will error if you try to make it an Input or Output
    * It **can** be Flipped
1. Only Ports and Wires can be of type Analog
1. Analog must be defined with a Width (there is no zero-argument constructor)
1. `attach(...)` is the primitive for attaching any number of Analog elements
1. Analog is only pass-through, no casting to or from other types (fromBits will error)
1. Analog **can** be bulk connected, but only **once per module**
    * This is to make it still legal to bulk connect IO but avoids confusing last connect semantics
    * "Once per module" sounds strange but it basically means that both sides of a Port can be bulk connected
1. While bulk connecting Analog types technically works in the compatibility wrapper, it does not work if you bulk connect Aggregates with Analog elements
    * This is a technical limitation, `Chisel._` issues Firrtl connect on the aggregate while `chisel3._` walks the hierarchy and issues connects for each of the leaves. Firrtl only supports attaching Analogs so in `chisel3._` we can intercept the bulk connect and issue an attach--we can't do this in `Chisel._`
1. Conditionally attaching (including as derived from bulk connect) will error

shunshou:
@jackkoenig do you have an eta on when this branch will be usable? I'd like to use this for CRAFT pads and a simple analog black box, which I think physical design people would like ASAP.

alonamid:
When you hook up an Analog with an input or output of a different module (not in bulk connect), will you be able to use :=, or will you use attach()?

jackkoenig:
@shunshou I'm trying to get this in ASAP. No one has raised any complaints about the current API so I believe you should feel free to use it. I'm trying to make sure that the Verilog Firrtl is emitting synthesizes since the tools don't seem to handle inout in a consistent or general way.

@alonamid `:=` is not supported since imo it implies directionality and Analog is directionless. You can use `attach` or `<>` on the actual Analog type itself.

jackkoenig:
This will fail until https://github.com/ucb-bar/firrtl/pull/415 is merged but I think it is all set and ready to go! See the latest comment there for some updates on how Analog is emitted.

jackkoenig:
ucb-bar/firrtl#415 has been merged so this is ready for primetime. Multiple people are already using it. Anyone want to review? Most of the code is tests.

ducky64:
Will review this shortly.

src/test/scala/chiselTests/AnalogSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+    elaborate { new Module {
+      val io = IO(new Bundle {
+        val a = Flipped(Analog(32.W))
+      })
+    }}
+  }
+
+  // There is no binding on the type of a memory
+  // Should this be an error?
+  ignore should "NOT be a legal type for Mem" in {
```
ducky64: Why is this ignore?

src/test/scala/chiselTests/AnalogSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+    a [ChiselException] should be thrownBy {
+      elaborate { new Module {
+        val io = IO(new Bundle {})
+        val mem = Mem(16, Analog(32.W))
+      }}
+    }
+  }
+
+  it should "NOT be bindable to Mem ports" in {
```
ducky64: Practically this is equivalent to the above, both because it's going to fail at the Mem creation stage, and that if a Analog Mem is allowed then so should a Analog Mem port.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+// Has multiple ports for driving and checking but only one shared bus
+class AnalogBlackBoxIO(n: Int) extends Bundle {
+  require(n > 0)
+  val bus = Analog(32.W)
+  val port = Vec(n, new AnalogBlackBoxPort)
+}
+
+// Assigns bus to out
+// Assigns in.bits + index to bus when in.valid
+class AnalogBlackBox(index: Int) extends BlackBox(Map("index" -> index)) {
```
ducky64: Can we use #453 so Verilog implementations stay with the BlackBox instead of being added as tester parameters?

src/test/scala/chiselTests/AnalogSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+      attach(busWire, writer.io.bus)
+      attach(busWire, mods(0).io.bus)
+      attach(mods(1).io.bus, busWire)
+      mods.foreach(check(_))
+    }, Seq("/AnalogBlackBox.v"))
+  }
+
+  // This does not currently work in Verilator unless Firrtl does constant prop and dead code
+  // elimination on these wires
+  ignore should "work with intermediate wires attached to each other" in {
```
ducky64: Will FIRRTL at least error out if this happens?
(you don't want a silent failure)

chiselFrontend/src/main/scala/chisel3/core/BiConnect.scala
```diff
@@ -42,6 +44,9 @@ object BiConnect {
     BiConnectException(s": Right Record missing field ($field).")
   def MismatchedException(left: String, right: String) =
     BiConnectException(s": Left ($left) and Right ($right) have different types.")
+	def AttachAlreadyBulkConnectedException(sourceInfo: SourceInfo) =
```
jackkoenig: Mistake, I'll fix it

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -1011,3 +1012,33 @@ object FixedPoint {
   }
 
 }
+
+/** Data type for Verilog inout */
+final class Analog private (width: Width) extends Element(width) {
```
jackkoenig: Sure, I'll add it to the ScalaDoc. It is the number of wires analogously to the width of a UInt

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -1011,3 +1012,33 @@ object FixedPoint {
...
+  require(width.known, "Since Analog is only for use in BlackBoxes, width must be known")
+
+  // Used to enforce single bulk connect of Analog types, multi-attach is still okay
```
jackkoenig: To prevent unintuitive behavior surrounding last connect semantics. Basically:
```scala
mod1.someBundle <> mod2.someBundle
mod1.someBundle <> mod3.someBundle
```
Last connect semantics say that the latter will overrule the former, but in the case of any Analog members, they would stay connected. Firrtl doesn't even allow attach in connections but Analog would be virtually unusable in Chisel without it so the most reasonable contract I could come up with was that you *can* bulk connect, but not more than once because last connect doesn't really make sense

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
jackkoenig: Yeah I'll look into it

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+
+// Wraps up n blackboxes, connecing their buses and simply forwarding their ports up
+class AnalogBlackBoxWrapper(n: Int, idxs: Seq[Int]) extends Module {
+  require(n > 0)
+  val io = IO(new AnalogBlackBoxIO(n))
+  val bbs = idxs.map(i => Module(new AnalogBlackBox(i)))
+  io.bus <> bbs.head.io.bus // Always bulk connect io.bus to first bus
+  io.port <> bbs.flatMap(_.io.port) // Connect ports
+  if (n > 1) { // attach all buses if there are more than 1
+    attach(bbs.map(_.io.bus):_*)
```
jackkoenig: Yeah probably

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+  )
+  // Connect all ports to top
+  io.ports <> mods.flatMap(_.io.port)
+  // Attach first 3 Modules
+  attach(mods.take(3).map(_.io.bus):_*)
+}
+
+
+// This tester is primarily intended to be able to pass the dut to synthesis
+class AnalogIntegrationTester(mod: => AnalogDUTModule) extends BasicTester {
```
jackkoenig: Agreed, I've wanted the chisel tests to be able to depend on the chisel-testers, which it should be okay in sbt since test and main are separate but I'm not sure

src/test/scala/chiselTests/AnalogSpec.scala
jackkoenig: This is ignore because it doesn't fail, there's no Mem binding so this code elaborates just fine

src/test/scala/chiselTests/AnalogSpec.scala
jackkoenig: Similar but not exactly equivalent. I added this because the above does not fail, but this does

src/test/scala/chiselTests/AnalogSpec.scala
jackkoenig: Firrtl doesn't error out because this code can be simulated in most tools and synthesized. However, the emitted Verilog has 
```verilog
`ifdef verilator
  `error "Helpful error message"
`endif
```
So that if you try to compile with Verilator you will get a reasonable error message.

jackkoenig: I could change this test to be something like "assertVerilatorFails { "

chiselFrontend/src/main/scala/chisel3/core/Attach.scala
```diff
@@ -0,0 +1,45 @@
...
+    * @param elts The components to attach
+    *
+    * @example
+    * {{{
+    * val a1 = Wire(Analog(32.W))
+    * val a2 = Wire(Analog(32.W))
+    * attach(a1, a2)
+    * }}}
+    */
+  def apply(elts: Analog*)(implicit sourceInfo: SourceInfo): Unit = {
```
jackkoenig: You should use `<>` if doing a bulk connect like normal. You should use attach if you're trying to attach more than 2 Analog. I'm not sure that there is a general concept here since attach is explicitly bidirectional and there's no analogous concept for Bits. You could have multi connect I guess but that has an explicit single source and multiple sinks.

jackkoenig:
`AnalogSpec` does not have any driving but `AnalogIntegrationSpec` actually does https://github.com/ucb-bar/chisel3/pull/422/files/04e2be437565087e310cc1d62bfbdd067190b836#diff-fb5ba5874d9c1b21c7cad48dc9a6c312R121

chiselFrontend/src/main/scala/chisel3/core/Attach.scala
ducky64: I think multi-connect is something that might make sense, though with Bits directionality issues would be caught at the Chisel level rather than in Verilog.

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
jackkoenig: Added documentation about this and width to the ScalaDoc

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -118,6 +120,7 @@ extends HasId {
   private[chisel3] val _commands = ArrayBuffer[Command]()
   private[core] val _ids = ArrayBuffer[HasId]()
   Builder.currentModule = Some(this)
+  Builder.whenDepth = 0
```
jackkoenig: I added the check. I am of the opinion that whenDepth should be set in the Module itself to keep all side effects of creating a Module in the same place since currentModule, etc. are all set there. I know they *have* to be set inside of the Module but I think keeping them in the same place is more consistent.

chiselFrontend/src/main/scala/chisel3/core/Attach.scala
jackkoenig: I'm open to having some kind of multi-connect. It would be cool if the API could be something like
`val (a, b) ::= mySource` but I'm not sure how possible this is (is is possible to have a function defined on a class invoke an unapply like thing? Or could this just be piggy-backed on unapply without unintended strange consequences?)

Anyhow, I think this is orthogonal to attach because they are very different.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
jackkoenig: I tried but there seems to be a bug in #453 or in the accompanying Firrtl implementation. I will look into and file/fix that separately but I don't think this should be blocked by it.

jackkoenig:
@ducky64 Improved and rebased on master

chiselFrontend/src/main/scala/chisel3/core/Attach.scala
ducky64: That's fine, though let's remember to discuss this later, preferably before we move this out of experimental?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: The other part of that argument is that whenDepth belongs with Builder, and for the sake of parallelism (in the grammatical sense) it might make sense to zero it in the same function where you store it and restore it?

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
ducky64: That's fine, can you open an issue so we remember to fix it?

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
ducky64: Not blocking, so can you create a follow-on issue (and this might be wrapped into a larger discussion of whether attach could be applied to all Data)?

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -1020,3 +1021,51 @@ object FixedPoint {
   }
 
 }
+
+/** Data type for representing bidirectional bitvectors of a given width
```
ducky64: Do you want to limit this to bitvectors, or could this be a way to thread actual analog signals through Chisel? The more generalized definition of width would just be (parallel) wires in an analog bus.

jackkoenig: In its current incarnation, it really just maps to bidirectional bitvectors in Verilog. If we do generalize it to simply wires in an analog bus, then I do think we should change this description!

Issue #421, opened 2016-12-21
Does fromBits work?


grebe:
It is looking like my fix for FixedPoint.fromBits is imperfect (see #391). In trying to figure out how everything works, I wrote this test:

```
class SIntFromBitsTester extends BasicTester {
  val sint = SInt(2.W)
  val uint = 3.U
  val b: SInt = sint.fromBits(uint)
  val sint_answer = (-1).S(2.W)
  
  assert(b === sint_answer)
  stop()
}
```

This fails with `firrtl.passes.CheckTypes$InvalidConnect: Type mismatch. Cannot connect b to UInt<2>("h3")` because the emitted firrtl tries to directly connect a UInt to an SInt. The offending lines of emitted firrtl are:

```
    wire b : SInt<2>
    b is invalid
    b <= UInt<2>("h03")
```

Am I wrong to think the test should work as written? It seems fromBits will only work for Bundles/Vectors entirely composed of UInts.

ducky64:
Seems like SInt doesn't overload fromBits, and it's defaulting to the Bits.fromBits implementation which doesn't do anything special in terms of FIRRTL generation.

Side notes: not sure why Data has a base fromBits implementation (the current one should probably to into Aggregate/Bundle/Vec, especially since Bits overrides it completely). We might also need more SInt in the stock Chisel3 test suite...

grebe:
It is true that sint doesn't overload frombits, but even if it did there
would be a problem. A bundle with an sint would give the same error because
frombits doesn't recursively call itself, it just bitslices and makes the
assignment directly. I feel like frombits should call itself recursively,
but it's also possible to fix this in the connect operator

On Dec 22, 2016 1:28 AM, "Richard Lin" <notifications@github.com> wrote:

Seems like SInt doesn't overload fromBits, and it's defaulting to the
Bits.fromBits implementation which doesn't do anything special in terms of
FIRRTL generation.

Side notes: not sure why Data has a base fromBits implementation (the
current one should probably to into Aggregate/Bundle/Vec, especially since
Bits overrides it completely). We might also need more SInt in the stock
Chisel3 test suite...

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ucb-bar/chisel3/issues/421#issuecomment-268728662>, or mute
the thread

ducky64:
Appears fixed?

PR #424, opened 2016-12-24
Make fromBits work with types other than UInt

grebe:
My attempt to address #421 

Data now has no default implementation for do_fromBits. The previous default implementation is moved to Aggregate, and each flavor of Bits has it's own implementation too.

I'm not sure if fromBits should be legal for a clock, or if it should just throw and exception. Right now it just makes a clock wire and uses the := operator.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -21,6 +21,21 @@ sealed abstract class Aggregate extends Data {
     pushCommand(BulkConnect(sourceInfo, this.lref, that.lref))
 
   override def do_asUInt(implicit sourceInfo: SourceInfo): UInt = SeqUtils.do_asUInt(this.flatten)
+  def do_fromBits(that: Bits)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): this.type = {
+    var i = 0
+    val wire = Wire(this.chiselCloneType)
+    val bits =
+      if (that.width.known && that.width.get >= wire.width.get) {
+        that
+      } else {
+        Wire(that.cloneTypeWidth(wire.width), init = that)
+      }
+    for (x <- wire.flatten) {
+      x := x.fromBits(bits(i + x.getWidth-1, i))
+      i += x.getWidth
+    }
+    wire.asInstanceOf[this.type]
+  }
 }
 
 object Vec {
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -935,12 +935,6 @@ sealed class FixedPoint private (width: Width, val binaryPoint: BinaryPoint, lit
     res
   }
   //TODO(chick): Consider "convert" as an arithmetic conversion to UInt/SInt
-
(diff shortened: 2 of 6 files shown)


ducky64:
Implementation looks sane, but (sorry to turn everything into an API discussion) I think this raises the question whether fromBits is a good API to have. We already have the functionally-equivalent Bits.asSInt | asUInt (in fact, your implementation just delegates to those), and it makes no sense to try to make a Clock from Bits (other than Bool.asClock). Realistically, I think the only compelling reason for fromBits is Aggregate types (where you want chisel to handle packing and unpacking consistently), so it might make more sense to have fromBits (and maybe flatten?) as an Aggregate operation?

sdtwigg:
Our approach was the following:
1. asUInt or from[blah] on a Clock is a runtime-exception (iirc FIRRTL doesn't support this at all anyway)
2. Instead of fromBits, Data has an "asTypeOf[T <: Data](model: T): T" since most commonly people were using fromBits to perform a cast and this made all casts look like as[blah]. This also made asTypeOf invert itself. (There are potentially some odd cases when unknown-width SInts are involved; however, it was generally advised to use only known width things under threat of emitting errors later.)
3. asTypeOf uses connectFromUInt(target: UInt) for each element, which basically acts like your proposed fromBits (but without having to worry about which Bits it is). connectFromUInt also doesn't create any wires.

grebe:
@ducky64 I agree with you that fromBits is probably not a good API. In addition to the points you bring up, I don't like how it uses `this.type`. I think the approach @sdtwigg outlined looks like a much better API and I would vote for moving to that. Was this part of a previous PR that didn't make it in for some reason?

I still think it is probably worth merging something similar to this PR. Making an API change requires debate that will be further slowed by the holidays and the retreat, and this fixes a bug that is causing problems for us now (specifically with unpacking and packing for our axi4s ports in craft).

sdtwigg:
Yes, you should definitely merge this as a bugfix since the other proposals would likely end up using it anyway.

I think I had done this API change in late July. It was after my initial large pull request during that period when there was no interest in my pull request (but also long enough ago that I had actually thought this was part of that PR). As an amusing anecdote, I think it pushing the exact bugfix you are doing now (SInt just didn't work) a few weeks earlier which triggered the desire for an API re-examination.

ducky64:
For this PR itself, I think it would be good to have a test that checks that when you do (some UInt).fromBits, the resulting UInt is of the same width. At least truncation behavior can be checked using the data value, while you might need DataMirror to check width increases.

@sdtwigg can you open an issue proposing a fromBits API change? I think your API is less confusing and more consistent with the other conversions. It should also be possible to keep Bundle.fromBits in the compatibility layer.

ducky64:
@grebe if you add the width tests (larger UInt will be truncated to fit within a smaller UInt it's being converted to; smaller UInt will be width expanded when converted to a larger UInt - this can be checked using DataMirror), I'll approve this.

grebe:
Not sure what DataMirror is for or if I used it correctly, but here is my attempt. Does this look OK?

grebe:
Also, I believe #391 is not addressed by this and is still needed for FixedPoint to work

grebe:
#391 got merged and broke something, so I just pushed a fix. I also added a test case.

PR #436, opened 2017-01-11
Make EnqIO (output of queue) IrrevocableIO, reflecting queue guarantees

albert-magyar:
Two changes:

`IrrevocableIO` is now a subtype of `DecoupledIO`, which was already assumed and implied by the documentation of the up/downconversions between them.

`EnqIO` (the object used to create the output of  `Queue`) now returns an instance of type `IrrevocableIO`, which reflects a guarantee provided by queues. This seems more-or-less to be central to the "queue-ness" of a module.

Understandably, this creates a new burden that must be maintained in future implementations to avoid breaking the API. This enables users of `AdvTester` to use queues properly on the outputs of their designs.

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -103,7 +103,7 @@ object Decoupled
   * Additionally, once 'valid' is raised it will never be lowered until after
   * 'ready' has also been raised.
   */
-class IrrevocableIO[+T <: Data](gen: T) extends ReadyValidIO[T](gen)
+class IrrevocableIO[+T <: Data](gen: T) extends DecoupledIO[T](gen)
 {
   override def cloneType: this.type = new IrrevocableIO(gen).asInstanceOf[this.type]
 }
@@ -129,7 +129,7 @@ object Irrevocable
 }
 
 object EnqIO {
-  def apply[T<:Data](gen: T): DecoupledIO[T] = Decoupled(gen)
+  def apply[T<:Data](gen: T): IrrevocableIO[T] = Irrevocable(gen)
 }
 object DeqIO {
   def apply[T<:Data](gen: T): DecoupledIO[T] = Flipped(Decoupled(gen))


hcook:
I'm pretty suspicious of the subtyping. We definitely wrote it that way the first time and then changed it because we came up with a case that was problematic.  I think it had something to do with [the directionality of the interface](https://github.com/albert-magyar/chisel3/blob/dda64c1dee16b5da15ac690bd3cd6759c3d5c032/src/main/scala/chisel3/util/Decoupled.scala#L46). If you have a function you're going to map over the inputs to something that expects Irrevocable behavior, you don't want that to be able to accept DecoupledIOs...

As far as the output of queue goes, we then left it alone for backwards compatibility reasons.

albert-magyar:
Not sure what the problem would be in that function example, since `(IrrevocableIO) => ...` won't accept `DecoupledIO` in either case?

It also looks like `EnqIO` and `DeqIO` are backwards, but that's a somewhat separate issue.

aswaterman:
@hcook's right; the discussion leading to that conclusion is here: https://github.com/ucb-bar/chisel3/pull/274

albert-magyar:
It looks like the issue Stephen raised is with `(DecoupledIO) => ...` accepting `IrrevocableIO`, not the other way around.

I understand that having the constrained `IrrevocableIO` subtype the less constrained `DecoupledIO` doesn't feel very Liskov substitution-y, but at least the runtime check will catch violations in simulation. The issue with attempting to push it more to the type system by eliminating `IrrevocableIO <: DecoupledIO` is that there are still huge holes.

Since `IrrevocableIO <: ReadyValidIO`, the example Stephen posited is still possible (substituting `ReadyValidIO` for `DecoupledIO`) under the current type hierarchy. The hole exists in some form as long as `IrrevocableIO` subtypes any subtype of `Bundle` rather than using composition and a has-a relationship.

**Much bigger issue**: even without the subtyping, the following is perfectly legal (compile and elaboration) under current Chisel3.

```
def ConnectWrong[T <: Data] (source: DecoupledIO[T], sink: IrrevocableIO[T]) = {
  source <> sink
}
```

ducky64:
We're keeping the same type hierarchy and fixing the advanced tester to issue warnings. Re-open if you disagree...

Issue #438, opened 2017-01-15
Reg + Type parameterization


shunshou:
Ran into 

```

inferred type arguments [Object] do not conform to method apply's type parameter bounds [T <: chisel3.core.Data]
[error]   io.out.bits.y := Reg(next = io.in.bits.y)

```

error, which this comment in the code alluded to [assuming you happened to see the comment]:

```
@note this may result in a type error if called from a type parameterized
    * function, since the Scala compiler isn't smart enough to know that null
    * is a valid value. In those cases, you can either use the outType only Reg
    * constructor or pass in `null.asInstanceOf[T]`.
    */
```

Can you not just make `def apply[T <: Data](t: T = null, next: T = null, init: T = null)` private and case out all of the possibilities via RegNext, RegInit, RegEnable [edit: Ok seems they're pretty cased out], etc. to save people the trouble of debugging stuff like this?

ducky64:
I think that's a valid possible solution to the Reg problem (eliminate the nulls, which are *terrible*) by having only a single-argument Reg constructor (analogous to Wire, which only creates a synthesizable node), and requiring the use of RegNext and friends for the next=... and init=... arguments.

This would be an API changing thing (so it's a good thing 3.0.0 is still on the horizon), and the unclean Reg could stay in compatibility layer for those stuck at 2.x semantics.

(side note: many of these notes documenting non-intuitive behavior was from the initial push to 3.0, where we prioritized API compatibility and avoided solutions which could cause existing code to break. The compatibility layer makes that less necessary now.)

Thoughts? @sdtwigg (you must hate the use of nulls, did you have a solution?), @azidar, @jackkoenig ?

shunshou:
@sdtwigg was the one that reminded me of RegNext. :P 

Regarding the notes: I assume the errors you'd run into would basically be the same each time. It'd be much more useful to create some searchable list of errors and solutions i.e.

Search "inferred type arguments [Object] do not conform to method apply's type parameter bounds [T <: chisel3.core.Data]", "Reg"

Result: Don't use Reg(next = blah)

Rather than making people scratch their heads... Maybe people would think to search for Reg specifically; maybe they'd think they just wrote their code incorrectly (esp. for people who are new to type parameterization)... but that just costs design time...

ducky64:
[...]

PR #439, opened 2017-01-16
Add shift register with reset

stevobailey:
Open to discussion. It resets all registers in the shift to the same value. This is sufficient for my needs.

### DIFF
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -66,7 +66,7 @@ object ShiftRegister
     *
     * @param in input to delay
     * @param n number of cycles to delay
-    * @param reset reset value
+    * @param resetData reset value for each register in the shift
     * @param en enable the shift
     */
   def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool = true.B): T = {
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -61,4 +61,20 @@ object ShiftRegister
       in
     }
   }
+
+  /** Returns the n-cycle delayed version of the input signal with reset initialization.
+    *
+    * @param in input to delay
+    * @param n number of cycles to delay
+    * @param reset reset value
+    * @param en enable the shift
+    */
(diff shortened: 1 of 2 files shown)


edwardcwang:
1) Is there a corresponding test in `src/test/scala/chiselTests/Reg.scala`?

2) Seems reasonable to squash the second commit into the first since it modifies the previous one.

ducky64:
Might be worth waiting until #438 is resolved, so we can keep semantics consistent. That this overloads a function with a slightly-different-but-not-that-much-different type signature (with reset vs. without reset) kind of "smells" off.

Squash protip: there's a "squash and merge" button on GitHub. (see end of https://github.com/blog/2141-squash-your-commits)

stevobailey:
1. No, but I didn't see a shift register test there either...there's a ShiftRegisterTester in Vec.scala (but it tests a Vec + Counter, not a ShiftRegister).
2. As Richard says, squash and merge can be done once approved.
3. Okay, we'll wait for #438

grebe:
I wouldn't advocate waiting for #438 - that issue has been around forever in one form or another. ShiftRegister already exists- whatever happens with the issue will impact it regardless. This new apply just parallels the two already existing implementations of RegEnable.

ducky64:
I'm pushing for a resolution for #438 within a week and implementation (if necessary) to follow shortly after.

stevobailey:
Bump. Richard, it doesn't look like #455 changes the API, so this should still be valid. Can we merge?

ducky64:
This doesn't break anything, but it does an API item which we would have to maintain continued support for. I'd like broader agreement on this before merging.
@jackkoenig @azidar

ducky64:
This doesn't break any existing APIs, but it does add a new API item, which I think should have wider consensus. @azidar @jackkoenig ?

azidar:
I'm fine with this.

grebe:
Does the approver do the merge or the pull-requester?

ducky64:
I would prefer the requester does the merge (in case someone has a last-minute-fix moment), though it's kind of been a free-for-all.

ucbjrl:
I believe we encourage the requester to do the merge (assuming they have
push access), so they can deal with any conflicts that arise post approval.

stevobailey:
Okay I don't have push access, so I can't merge.

PR #444, opened 2017-01-19
Mark Annotation and FixedPoint as experimental

chick:
Moved interfaces into experimental package
Fix tests and other references to these constructs

### DIFF
--- a/src/test/scala/chiselTests/AnnotatingDiamondSpec.scala
+++ b/src/test/scala/chiselTests/AnnotatingDiamondSpec.scala
@@ -3,7 +3,7 @@
 package chiselTests
 
 import chisel3._
-import chisel3.experimental._
+import chisel3.experimental.ChiselAnnotation
 import chisel3.internal.InstanceId
 import chisel3.testers.BasicTester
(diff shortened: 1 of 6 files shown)


src/test/scala/chiselTests/AnnotatingDiamondSpec.scala
```diff
@@ -3,6 +3,7 @@
 package chiselTests
 
 import chisel3._
+import chisel3.experimental._
```
ducky64: Consider just importing the relevant parts of experimental.

stevobailey:
Why? I seemed to have missed a discussion about this.

chick:
This came out of a *cleanup the chisel api* meeting.  I probably jumped the gun on approving it, I was thinking the cleanup would be just in dsptools.

chick:
Could be task for next hacking session, shouldn't take very long to fix imports

stevobailey:
Okay, so this isn't something that will be moved out of experimental anytime soon? If not, then yeah we can go ahead and add the import statements everywhere on Monday

PR #448, opened 2017-01-23
Use the same firrtl for all sub-projects - fixes #447.

### LINKED ISSUE #447: chisel3 sub-projects may use a different version of firrtl
It looks like each sub-project gets its own `unmanagedBase`, potentially using a different version of firrtl from the root project.

ucbjrl:
Use the root project's `unmanagedBase` (".../lib") in all sub-projects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,8 @@ lazy val commonSettings = Seq (
   scalacOptions := Seq("-deprecation", "-feature"),
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
   addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
+  // Use the root project's unmanaged base for all sub-projects.
+  unmanagedBase := (unmanagedBase in root).value,
   // Since we want to examine the classpath to determine if a dependency on firrtl is required,
   //  this has to be a Task setting.
   //  Fortunately, allDependencies is a Task Setting, so we can modify that.
@@ -104,6 +106,8 @@ lazy val chiselFrontend = (project in file("chiselFrontend")).
   settings(publishArtifact := false).
   dependsOn(coreMacros)
 
+// Hack to avoid circular dependency on chisel.
+val root = Project("chisel", file("."))
 
 lazy val chisel = (project in file(".")).
   enablePlugins(BuildInfoPlugin).
--- a/build.sbt
+++ b/build.sbt
@@ -106,8 +106,8 @@ lazy val chiselFrontend = (project in file("chiselFrontend")).
   settings(publishArtifact := false).
   dependsOn(coreMacros)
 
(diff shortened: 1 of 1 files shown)


colinschmidt:
This seems to work for me. No SBT expert but from what I learned debugging this issue before it looks good.

build.sbt
```diff
@@ -109,6 +111,8 @@ lazy val chiselFrontend = (project in file("chiselFrontend")).
   ).
   dependsOn(coreMacros)
 
+// Hack to avoid circular dependency on chisel.
+val root = Project("chisel", file("."))
 
 lazy val chisel = (project in file(".")).
```
jackkoenig: I am no sbt expert either, but it looks like we now have `chisel` and `root` as the same project. Is root a magical val that projects using chisel will set and thus override or overrule this assignment thus unmanageBase in root will still work?

ucbjrl: I need to verify that higher level projects may re-define root.

jackkoenig: I have no idea if projects actually do redefine it, I was just speculating about how this works! Does `unmanagedBase in root` mean this `val root` or does it mean like the actual "root" project?

colinschmidt:
Does this mean that rocket-chip needs to put the firrtl jar somewhere else? I.e. in top-level lib directory rather than chisel3/lib?

ucbjrl:
Yes. It's a simple change to rocket-chip's Makefrag:

    % git diff Makefrag
    diff --git a/Makefrag b/Makefrag
    index b613ae5..8845435 100644
    --- a/Makefrag
    +++ b/Makefrag
    @@ -23,8 +23,8 @@ FIRRTL ?= java -Xmx2G -Xss8M -XX:MaxPermSize=256M -cp $(FIRRTL
     $(FIRRTL_JAR): $(shell find $(base_dir)/firrtl/src/main/scala -iname "*.scala")
        $(MAKE) -C $(base_dir)/firrtl SBT="$(SBT)" root_dir=$(base_dir)/firrtl b
        touch $(FIRRTL_JAR)
    -       mkdir -p $(base_dir)/chisel3/lib
    -       cp -p $(FIRRTL_JAR) $(base_dir)/chisel3/lib
    +       mkdir -p $(base_dir)/lib
    +       cp -p $(FIRRTL_JAR) $(base_dir)/lib
     
     src_path := src/main/scala
     default_submodules := . hardfloat chisel3
    (END)

but it does have to be coordinated with this change to chisel3.

build.sbt
ucbjrl: The updated definition (using `RootProject`) should work without requiring any re-definition on the part of the root project. It should always refer to the actual "root" project.

colinschmidt:
Ok should probably submit a PR for that change. I'm testing it locally now.

jackkoenig:
Resolved conflict in `build.sbt` and rebased on master, this is required to bump rocket-chip so will merge pending jenkins passing

Issue #427, opened 2017-01-02
Special Top Module class to rename/delete clock/reset


Martoni:
It seems to be impossible to delete or rename clock and reset signals of Module declared with Chisel3. But it can be useful for some design where clock come from differential input buffer for example.

I had this problem with a Xilinx FPGA Kit [where a special buffer must be instantiated to use clock](http://stackoverflow.com/questions/41427717/how-to-delete-clock-signal-on-chisel3-top-module/41430368#41430368).

It could be useful to have a special top module where we can decide if we use clock and reset and where we can rename io signal. Like BlackBox.

ben-k:
I'll second this - it currently seems to require external post-processing of the generated verilog to address, and seems like it will be a very common issue in non-toy designs.

It might be better addressed in firrtl, though.

edwardcwang:
Another alternative: wrap the generated 'top-level' Verilog into another Verilog file and just connect the wires.

sdtwigg:
We addressed this by redoing the Module hierarchy (BlackBox already was working as a hack so this was a good opportunity for reconciliation)

The hierarchy is:
BaseModule
BaseModule -> BlackBox (user-extendable)
BaseModule -> UserDefinedModule
UserDefinedModule -> ImplicitModule (user-extendable, Module is type aliased to this)
UserDefinedModule -> RawModule

A BaseModule is basically any module that can be instantiated as a submodule. Thus, all it has is an io. (Note that the 'io' field might not be the true full io)
A UserDefinedModule is a module that actually holds an implementation in Chisel. Thus, it has both an io and commands. It can also have submodules/definitions and therefor all 'parent' fields are actually UserDefinedModule

ImplicitModule is an old Chisel.Module and thus has a clock/reset pin and the io is actually put in an "io" subfield of the true io. The Module wrapper handled propagating clock/reset from ImplicitModule parent to ImplicitModule child.
RawModule is new and has no clock/reset pins. The io is the true io (excepting any ports added reflectively). No automatic propagation is done into this module or from this module into its children. Thus, an ImplicitModule in a RawModule must have its clock and reset manually connected.

[...]

PR #469, opened 2017-01-31
Module Hierarchy Refactor

ducky64:
Implementation for the proposal in #427 

Major changes:
- Implemented the new module hierarchy in #427 (abstract BaseModules, BlackBoxes, UserModules without implicit io, clock, reset, and ImplicitModule which is chisel's old Module)
- UserModules have ios defined with IO(...), that is, IO(...) registers its contents as a module IO. Arbitrary numbers of IO are allowed.
- All IOs are named by reflection, so they must be of the form `val myIo = IO(...)`. IO names are no longer hardcoded.
- Refactoring to make it easier to reason about correctness and for better abstraction boundaries.
- Added getReset, getClock, forcedReset, forcedClock to Builder. These return empty (or error out) when in a UserModule without an implicit clock / reset. This also paves the way for withClock blocks.
- Yes, rocket chip still builds.

Debate points:
- Should `io` be required in ImplicitModule? Perhaps `io` should be more a convention (or even have IOs turned into the wild wild west)?
- Blackbox `io` handling is inconsistent with Module `io`. Blackboxes don't emit the `io` prefix when elaborating; this was likely to allow it to do something useful before we even considered having IOs not in `io.`. Now that this can change - should it? (as in, instead of putting things in an `io` bundle, just dump it directly into the BlackBox as `IO(...)`?
- This will break override `io` (except in compatibility mode, the auto-wrap handles things properly), because it seems that every time you define `val io = IO(...)`, even if overridden, it calls the IO and creates a port. The overridden ones are unable to get a name through reflection and fails.
- Should it fail if it can't name an IO through reflection? It can have a generic name, like `port_1`, ... On the other hand, some use cases may require good names, like if you had a top-level Verilog file instantiating a Chisel-generated module. Additionally, we rely on reflection for Bundle elements.


sdtwigg:
A lot of places take a BaseModule (like in connect signature) when they should be taking the more descriptive UserModule.

Also, mandating reflection works and having multiple IOs seems a bit too risky: would likely be best to still have 1 IO. Different Module types just may interpret that IO differently (for example, in a raw module, that IO is the true IO whereas in ImplicitModule that IO is a subpart io of the true io which includes clock and reset).

As noted, the current approach with IO() is inconsistent with BlackBoxes (and would make it tricky to have programmatic IOs using Record....)

ducky64:
Yeah, I guess the context_mod should be UserModules since it really doesn't make sense to be issuing connects in a BlackBox. Will fix.

Mandating reflection: we already do that in Bundles, so there is precedent. It also uses very similar infrastructure. Whether we should have multiple IO (or whether it's worth having another necessary reflection based system) is a larger question. There have been requests for not needing the `.io` fields.

Record support is a good point for Blackbox IOs. Though style-wise current Blackbox IOs are already inconsistent with Module IOs (and it shows through the code structure).

sdtwigg:
My discussions with other people who use Scala and have done DSL work (but are not involved with Chisel) have reaffirmed avoiding using reflection (particularly runtime reflection) for things.

Also, as discovered, the Bundle reflection is already somewhat fragile and people are more likely to have complex Modules than complex Bundles.

sdtwigg:
Addendum: This is a good opportunity to bring BlackBox IO's in line with other Module IO's. The desires are also similar: for both you want fine-grained control of IO naming.

ducky64:
We can have the old Blackbox behavior preserved in the compatibility layer.

Though what should the new behavior be? We don't want the `io` prefix so Chisel Blackboxes can interface with native Verilog modules without a wrapper layer, but reflection-based IO naming (along with the lack of Record-like programmatic generation) is also undesirable. If we just dump the io contents into the top level, that risks a conflict with the ImplicitModule clock and reset signals.

sdtwigg:
Submodules of BaseModule decide their IO naming structure.

For BlackBox and RawModule, we have them dump their io contents as the raw port names.
For ImplicitModule, the whole io Record is just labeled io (as if it was a field of the effective true io).

ducky64:
Isn't that what we have now?

sdtwigg:
I haven't worked through your code but my impression is that:

```scala
class TestModuleR extends RawModule {
  val io = IO(new Bundle {
    val clk = Input(Clock())
    val in = Input(Bool())
    val out = Output(Bool())
  })
}
```

in my proposal would result in 3 ports called "clk", "in", and "out"
whereas in your proposal results in 3 ports called "io_clk", "io_in", "io_out" where the 'io' is being derived reflectively. (you would instead need 3 vals called clk, in, and out all wrapped in IO)

ducky64:
Ah, I see, RawModule is a sibling class to ImplicitModule rather than users inheriting UserModule.

It could be done, though it's also inconsistent with ImplicitModule, which I expect to see the most usage because of the implicitly threaded clock and reset.

Would be nice to have some way to unify all three, though that might involve mangling the clock and reset naming for ImplicitModule in the event of a conflict. Which all things considered, might not be all that bad?

ducky64:
Updated with latest resolution.
ExtModule and MultiIOModule (actually an alias for chisel3.core.ImplicitModule) implemented.
Module is an alias for chisel3.core.LegacyModule, which is an ImplicitModule that enforces that the only IOs are io, clock, and reset. While MultiIOModule requires reflection for naming IOs, (Legacy)Module does not.

ducky64:
Also added exposed RawModule (which is just a UserModule). I'm not sure what kind of tests we want to see there, it's more absence of features than presence of any feature.

ducky64:
Bump on review...

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -18,19 +22,16 @@ object Module {
...
 
+  def do_apply[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = {
     if (Builder.readyForModuleConstr) {
       throwException("Error: Called Module() twice without instantiating a Module." +
                      sourceInfo.makeMessage(" See " + _))
     }
     Builder.readyForModuleConstr = true
-    val parent: Option[Module] = Builder.currentModule
+
+    val parent: Option[BaseModule] = Builder.currentModule
```
sdtwigg: The currentModule should always be a Option[UserModule].

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -39,8 +40,7 @@ object Module {
     //   - unset readyForModuleConstr
     //   - reset whenDepth to 0
     //   - set currentClockAndReset
-    val m = bc.setRefs()
-    m._commands.prepend(DefInvalid(childSourceInfo, m.io.ref)) // init module outputs
+    val module: T = bc  // Module is actually invoked here
```
sdtwigg: Technically, "bc is actually evaluated here"
You are already well underway of the Module invocation.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -39,8 +40,7 @@ object Module {
...
 
     if (Builder.whenDepth != 0) {
```
sdtwigg: Why isn't whenDepth tracked on the current module?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -79,34 +71,122 @@ object Module {
...
-    // Bind each element of the iodef to being a Port
-    Binding.bind(iodef, PortBinder(this), "Error: iodef")
-    iodef
+  Builder.currentModule = Some(this)
+  Builder.whenDepth = 0
+
+  //
+  // Module Construction Internals
+  //
+  protected var _closed = false
```
sdtwigg: Isn't this redundant with !_component.isEmpty?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -79,34 +71,122 @@ object Module {
...
+
+  // Fresh Namespace because in Firrtl, Modules namespaces are disjoint with the global namespace
+  private[core] val _namespace = Namespace.empty
+  protected val _ids = ArrayBuffer[HasId]()
```
sdtwigg: Why is this exposed?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -79,34 +71,122 @@ object Module {
...
+      * without $$ unchanged
+      */
+    def cleanName(name: String): String = name.split("""\$\$""").lastOption.getOrElse(name)
+
+    for (m <- getPublicFields(rootClass)) {
+      nameRecursively(cleanName(m.getName), m.invoke(this))
+    }
+
+    // For Module instances we haven't named, suggest the name of the Module
+    _ids foreach {
```
sdtwigg: Why is backup naming of modules done here? This seems like it should be done when registers, wires, etc. also need backup naming.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -120,32 +200,24 @@ extends HasId {
...
-
-  // Fresh Namespace because in Firrtl, Modules namespaces are disjoint with the global namespace
-  private[core] val _namespace = Namespace.empty
-  private[chisel3] val _commands = ArrayBuffer[Command]()
-  private[core] val _ids = ArrayBuffer[HasId]()
-  Builder.currentModule = Some(this)
-  Builder.whenDepth = 0
-  if (!Builder.readyForModuleConstr) {
-    throwException("Error: attempted to instantiate a Module without wrapping it in Module().")
+  protected def IO[T<:Data](iodef: T): iodef.type = {
```
sdtwigg: Again, allowing arbitrary fields to be io's will now encourage users to try to access other fields. (Or, be confused at what fields can or cannot be safely accessed).

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
-  // Setup ClockAndReset
-  Builder.currentClockAndReset = Some(ClockAndReset(clock, reset))
+/** Abstract base class for Modules that contain Chisel RTL.
+  */
+abstract class UserModule(implicit moduleCompileOptions: CompileOptions)
+    extends BaseModule {
+  //
+  // RTL construction internals
+  //
+  protected val _commands = ArrayBuffer[Command]()
```
sdtwigg: Why is this exposed?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
-        this
-      }
-      case None => this
+/** Abstract base class for Modules, which behave much like Verilog modules.
+  * These may contain both logic and state which are written in the Module
+  * body (constructor).
+  *
+  * @note Module instantiations must be wrapped in a Module() call.
+  */
+abstract class ImplicitModule()(implicit moduleCompileOptions: CompileOptions)
```
sdtwigg: Extra parens

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
+    extends UserModule {
+  // Implicit clock and reset pins
+  val clock = IO(Input(Clock()))
+  val reset = IO(Input(Bool()))
+
+  // Setup ClockAndReset
+  Builder.currentClockAndReset = Some(ClockAndReset(clock, reset))
+
+  private[core] def initializeInParent() {
+    // Don't generate source info referencing parents inside a module, since this interferes with
```
sdtwigg: This comment doesn't make sense though. This function isn't being called IN the parent. The clock and reset connections should mean the parent has references to the child source code (which should be fine since presumably all modules are instantiating that submodule)

sdtwigg: Also, Chisel no longer has de-duplication anyway....

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
-      * This extracts the part after $$ for names like this and leaves names
-      * without $$ unchanged
-      */
-    def cleanName(name: String): String = name.split("""\$\$""").lastOption.getOrElse(name)
-    for (m <- getPublicFields(classOf[Module])) {
-      nameRecursively(cleanName(m.getName), m.invoke(this))
-    }
+  // IO for this Module. At the Scala level (pre-FIRRTL transformations),
+  // connections in and out of a Module may only go through `io` elements.
+  def io: Record
```
sdtwigg: This should be restricted to being a stable identifier (val).

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
-      case m: Module => m.suggestName(m.desiredName)
-      case _ =>
+  // Allow access to bindings from the compatibility package
+  protected def _ioPortBound() = _ports contains io
+
+  protected override def nameIds(rootClass: Class[_]): HashMap[HasId, String] = {
+    val names = super.nameIds(rootClass)
+
+    // Allow IO naming without reflection
+    names.put(io, "io")
```
sdtwigg: Shouldn't these get names before everything else?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
+  private[core] override def initializeInParent() {
+    // Don't generate source info referencing parents inside a module, since this interferes with
+    // module de-duplication in FIRRTL emission.
+    implicit val sourceInfo = UnlocatableSourceInfo
+
+    pushCommand(DefInvalid(sourceInfo, io.ref))
+
+    override_clock match {
+      case Some(override_clock) => clock := override_clock
+      case _ => clock := Builder.forcedClock
```
sdtwigg: What if the parent module can't resolve forcedClock?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
     }
 
-    // All suggestions are in, force names to every node.
-    _ids.foreach(_.forceName(default="_T", _namespace))
-    _ids.foreach(_._onModuleClose)
-    this
+    override_reset match {
+      case Some(override_reset) => reset := override_reset
+      case _ => reset := Builder.forcedReset
```
sdtwigg: What if the parent module can't resolve forcedReset?

src/test/scala/chiselTests/BundleEltsSpec.scala
```diff
@@ -0,0 +1,84 @@
+// See LICENSE for license details.
```
sdtwigg: Remove this test or fix it.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -2,8 +2,12 @@
```
sdtwigg: This file should be split into multiple files.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -157,91 +229,160 @@ extends HasId {
...
+
+  private[core] def initializeInParent() {
+    // Don't generate source info referencing parents inside a module, since this interferes with
+    // module de-duplication in FIRRTL emission.
+    implicit val sourceInfo = UnlocatableSourceInfo
+
+    for (port <- _ports) {
+      pushCommand(DefInvalid(sourceInfo, port.ref))
     }
+
```
sdtwigg: What if the parent module cannot resolve forcedClock or forcedReset?

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
```diff
@@ -17,6 +17,84 @@ case class StringParam(value: String) extends Param
...
+  * import chisel3.experimental._
+  *
+  * // Example with Xilinx differential buffer IBUFDS
+  * class IBUFDS extends ExtModule(Map("DIFF_TERM" -> "TRUE", // Verilog parameters
+  *                                    "IOSTANDARD" -> "DEFAULT"
+  *                      )) {
+  *   val O = IO(Output(Clock()))
+  *   val I = IO(Input(Clock()))
+  *   val IB = IO(Input(Clock()))
+  * }
```
edwardcwang: Furthermore, why are all the I/Os clock? There's nothing clock-related with IBUFDS.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Also fixed with `_commands` accessor. And `_ports` fixed too.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Fixed

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Removed. Not sure what was going on there, probably was separated out of another code block where the comment was relevant.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Then you get a crash, just like what would happen if you tried to instantiate a Reg inside a non-ImplicitModule.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Also agree, but this preserves existing behavior.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: The standard names get priority (HashMap updated after the reflection naming).

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Ditto, crash as would be the case if you create a Reg where you can't resolve forcedClock.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: Ditto

src/main/scala/chisel3/compatibility.scala
```diff
@@ -152,16 +152,43 @@ package object Chisel {     // scalastyle:ignore package.object.name
   object Bool extends BoolFactory
   val Mux = chisel3.core.Mux
 
-  type BlackBox = chisel3.core.BlackBox
+  import chisel3.core.Param
+  abstract class BlackBox(params: Map[String, Param] = Map.empty[String, Param]) extends chisel3.core.BlackBox(params) {
```
ducky64: Done (this auto wraps IO)

src/test/scala/chiselTests/BundleEltsSpec.scala
ducky64: I'm not sure where this came from, probably git add'd something from another branch. Gone now.

sdtwigg:
I find these arguments to 'consistency' are weak, particularly because you are exposing very new functionality. Also, the argument is not compelling. I can similarly argue that it is inconsistent that ImplicitModule and BlackBox IO must be accessed wholly through io but RawModule and ExtModule io must access through arbitrary members.

Similarly, the user of a module must now comb through the implementation of RawModule and ExtModule to find which fields are actually ports. (This is even setting aside designers being 'clever' and parametrically making certain fields IOs versus internal nodes as demonstrated in the prior issue.)

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
sdtwigg: Is there a strong demand for a BlackBox that automatically has io on all members?
That you use this case to justify the feature is bizarre.

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
```diff
@@ -17,6 +17,84 @@ case class StringParam(value: String) extends Param
...
+  * @note The parameters API is experimental and may change
+  */
+abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Param]) extends BaseBlackBox {
+  private[core] override def generateComponent(): Component = {
+    require(!_closed, "Can't generate module more than once")
+    _closed = true
+
+    val names = nameIds(classOf[ExtModule])
+
+    // Name ports based on reflection
```
sdtwigg: I refuse to accept the claim that use of reflection has not been a problem. One came up with Bundles (which classically have less implementation trickery than modules since users would come to understand that reflection was being used to inspect them) literally days before you proposed expanding the use of reflection (#456).

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
```diff
@@ -52,22 +130,29 @@ case class RawParam(value: String) extends Param
   * }}}
   * @note The parameters API is experimental and may change
   */
-abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param]) extends Module {
+abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param]) extends BaseBlackBox {
+  def io: Record
```
sdtwigg: You are refactoring the module API, now is the time to address improper existing behavior. (Regardless, what prior modules are able to reliably get away with a non-stable io?)

chiselFrontend/src/main/scala/chisel3/core/Module.scala
sdtwigg: All of the code that set closed was also setting component. In what times (outside this function) is expected that _closed != _component.isDefined?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
sdtwigg: A reg not being instantiated in a non-implicit module is a problem because there is no way to later supply that reg a clock. In this case, there is since the module already exposes clock and reset pins.

ducky64:
With regards to the multi IO behavior, we can have another discussion about it during the next Chisel meeting, but the resolution from the last discussion was that it does have benefits and it's worth trying.

With regards to consistency: I'd argue that (assuming people like the Multi IO idea) the MultiIOModule and ExtModule style will be the new style.

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
ducky64: The idea is that if there is no required `.io` and have multi top-level IO support, then you don't need to generate the `io.` prefix in the FIRRTL/Verilog.

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
ducky64: And we will fix them as they come up, just like with any Chisel bugs. I don't know of any fundamental issues with reflections that would preclude some functionality.

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
ducky64: I'm trying to preserve the user-facing behavior of chisel3.Module / chisel3.core.LegacyModule. The main changes are the inclusion of new options like chisel3.MultiIOModule and chisel3.RawModule, which are new APIs.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: The new chose checks `_closed == true` are some points during the generateComponent phase, before `_component` is set.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: I would argue the stylistic way to do this is to use withClockAndReset, rather than assigning the clock and reset pins as connect statements.

Do you have a good example use case where you connect the clock and reset ports, but withClockAndReset is a bad fit?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
sdtwigg: Precisely, since this function is the one both setting _closed and _component anyway, there is no time outside execution of this function where _closed and _component.isDefined are different.

Multithreading isn't in play here since the current function already can't be called by multiple concurrent threads.

ducky64: It's more a defense against Chisel core being wrong rather than preventing mis-use,

src/main/scala/chisel3/package.scala
```diff
@@ -310,6 +310,11 @@ package object chisel3 {    // scalastyle:ignore package.object.name
     val withClock = chisel3.core.withClock
     val withReset = chisel3.core.withReset
 
+    type BaseModule = chisel3.core.BaseModule
+    type MultiIOModule = chisel3.core.ImplicitModule
+    type RawModule = chisel3.core.UserModule
+    type ExtModule = chisel3.core.ExtModule
```
jackkoenig: Move to experimental

src/test/scala/chiselTests/ExtModule.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.scalatest._
+import chisel3._
+import chisel3.experimental._
+import chisel3.testers.BasicTester
+import chisel3.util._
+//import chisel3.core.ExplicitCompileOptions.Strict
+
+// Avoid collisions with regular BlackBox tests by putting ExtModule blackboxes
+// in their own scope.
+package ExtModule {
```
jackkoenig: Should this just be a new package inside of chiselTests including the directory structure?

src/test/scala/chiselTests/ExtModule.scala
```diff
@@ -0,0 +1,72 @@
...
+package chiselTests
+
+import java.io.File
+
+import org.scalatest._
+import chisel3._
+import chisel3.experimental._
+import chisel3.testers.BasicTester
+import chisel3.util._
+//import chisel3.core.ExplicitCompileOptions.Strict
```
jackkoenig: Remove comment

src/main/scala/chisel3/package.scala
ducky64: ... this is in experimental ...
(GitHub diff apparently can't handle inner classes / objects / scopes when telling you the enclosing scope of a block of code)

src/test/scala/chiselTests/ExtModule.scala
ducky64: Fixed.

src/test/scala/chiselTests/ExtModule.scala
ducky64: I just need this to deconflict the scope, and the current implementation seems lightweight and fine.

ducky64:
@sdtwigg @jackkoenig issues fixed, inheritance composition with multiple top level IOs test added. Turns out the new stuff was in experimental all along, just GitHub doesn't do a good job of showing inner enclosing scopes.

Martoni:
Have you got a simple example to test this pull request ? I tried it with a simple blinking led project : https://github.com/Martoni/blp/tree/master/platforms/kintex7/chisel3

But I can't manage to generate verilog:
https://groups.google.com/forum/#!topic/chisel-users/m-ibJXZjm3E

ducky64:
@Martoni oversight in forgetting to refactor the Driver methods; fixed now.
In your example, you will also need to encapsulate your Module-inside-RawModule instantiation in a `withClockAndReset` block (needs to also be imported from `chisel3.experimental`).

sdtwigg:
Heh, I made the same oversight internally.

Martoni:
Thanks @ducky64, I pulled the last commits. But I can't find how to encapsulate my modules correctly in withClockAndReset. (How to do reset if I don't have one ?).

ducky64:
You can pass in any Bool for the reset signal, including `false.B`

Martoni:
Hi,
Thanks @ducky64, I managed to compile my blink design : 
https://github.com/Martoni/blp/tree/master/platforms/kintex7/chisel3

Works for me.

ducky64:
Rebased to latest master.

sdtwigg:
This is probably fine as long as the controversial parts actually remain in experimental.

ducky64:
Hooray!
... waiting for tests ...

ducky64:
retest this please

ucbjrl:
retest this please

ucbjrl:
retest this please

ducky64:
That's actually it for outstanding mega-PRs right now, the rest of them are mostly around 100 lines changes. So the entire PR train is in the station ... _for now_ ...

Martoni:
That works for me :)

PR #470, opened 2017-01-31
Fix namespace

jackkoenig:
Fixing one subtle bug led me to another which led to this fix (and a couple of minor other improvements). As all 3 commits are kind of orthogonal, I request a rebase and merge rather than squash.

Copy-paste most important commit message here:

Make Module and Bundle properly use empty namespaces
Also fix default suggested name of Module instances (now based on desired name
rather than actual assigned name).

Previously, Module and Bundle namespaces were "children" of the Module
definition namespace. This could lead to collisions that would give unexpected
names for module instances or Bundle elements. In particular, otherwise
identical modules that instantiate other identical modules in such a way that
the instance cannot be named via reflection would not be deduplicated because
the names of the instances would collide with the names of the modules in the
Builder.globalNamespace.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -374,7 +374,10 @@ abstract class Record extends Aggregate {
 
   // NOTE: This sets up dependent references, it can be done before closing the Module
   private[chisel3] override def _onModuleClose: Unit = { // scalastyle:ignore method.name
-    val _namespace = Builder.globalNamespace.child
+    // Since Bundle names this via reflection, it is impossible for two elements to have the same
+    // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
+    // which can cause collisions
+    val _namespace = Namespace.empty
     for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name)) }
   }
 
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -113,7 +113,8 @@ extends HasId {
     Port(iodef)
   }
 
-  private[core] val _namespace = Builder.globalNamespace.child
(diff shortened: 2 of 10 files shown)


chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -374,7 +374,7 @@ abstract class Record extends Aggregate {
 
   // NOTE: This sets up dependent references, it can be done before closing the Module
   private[chisel3] override def _onModuleClose: Unit = { // scalastyle:ignore method.name
-    val _namespace = Builder.globalNamespace.child
+    val _namespace = Namespace.empty
```
aswaterman: Comment on why Namespace is (not actually) necessary here?

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -113,7 +113,7 @@ extends HasId {
     Port(iodef)
   }
 
-  private[core] val _namespace = Builder.globalNamespace.child
+  private[core] val _namespace = Namespace.empty
```
aswaterman: Add comment noting that in Firrtl, the global and instance namespaces are disjoint

jackkoenig:
@ucbjrl @chick This will break chisel-testers because it removes the extra 'c' in ChiselExecutionSuccess. I figure this needs to happen sooner or later. Is this okay with you to merge?

sdtwigg:
Why do namespaces even have the parenting feature?
Is that even being used?

aswaterman:
It's historical.  I either misunderstood Firrtl's guarantees or was working around a Firrtl bug; I forget.  At this stage, it would be fine to nuke it altogether... up to y'all.

sdtwigg:
Given that it currently grants a false guarantee, it should probably just be removed since it is not worth fixing as it is not being used. (The parent doesn't check that an a element isn't in any of its descendants before being added so you can still end up with the same identifier in a child and parent namespace.)

aswaterman:
Agreed

jackkoenig:
Done.

Issue #460, opened 2017-01-26
Disallow CompileOptions materialize inside chisel3


ducky64:
Tracking issue.

Implement by making materialize a macro to prevent usage in the same package it's compiled into (as some with SourceInfo materialize). Scala may remove that restriction eventually (and break this "feature") but it'll be nice to have in the meantime.

PR #476, opened 2017-02-01
Don't allow CompileOptions to materialize in Chisel core

ducky64:
Implemented using macros.

The implicit needs to be defined low priority in the resolution order (https://stackoverflow.com/questions/5598085/where-does-scala-look-for-implicits) so that the user could overload it (instead of triggering an ambiguous implicit error). The companion object must be defined in the same source file (so we can't move the definition to the chisel3 package, for instance) so this relies on what is an implementation limitation of the Scala macro system.

Implementation for #460


aswaterman:
thanks for figuring this out!

PR #227, opened 2016-06-29
Added vec IO tests for #104

da-steve101:
Add test cases for #104, #109 fixes Vec( vecSize, UInt( INPUT ) ) but not Vec.fill( vecSize ) { UInt(INPUT) }
Also fails for RegInit( Vec(vecSize, UInt( 0 ) )

PR #109, opened 2016-02-25
Fix init in deq io

chick:
This fix is specifically to enable continued work on Chisel tutorials that use DeqIO and EnqIO
There are more general issues in  [Chisel Emits firrtl files with undeclared wires #104](https://github.com/ucb-bar/chisel3/issues/104) that can be addressed later.  This PR eliminates assignment statements in DeqIO/EnqIO which is generally agreed to be unsupported.

Issue #104, opened 2016-02-23
Chisel emits firrtl files with undeclared wires


chick:
Creating a circuit with an IO of the following form

```
class Packet extends Bundle {
  val header = UInt(width = 1)
}
class VectorPacketIO(n: Int) extends Bundle {
  val ins  = Vec(n, new DeqIO(new Packet()))
  val outs = Vec(n, new EnqIO(new Packet()))
}
```

coupled with the fact that DeqIO and EnqIO contain initialization code emits a .fir file that firrtl blows up with the following messages

```
[DEBUG] Starting High Form Check
Exception in thread "main" firrtl.passes.PassExceptions: 
firrtl.passes.CheckHighForm$UndeclaredReferenceException: VectorPacketIOUnitTester4043176957748550505.fir@8.4: [module BrokenVectorPacketModule] Reference T_7 is not declared.
firrtl.passes.CheckHighForm$UndeclaredReferenceException: VectorPacketIOUnitTester4043176957748550505.fir@9.4: [module BrokenVectorPacketModule] Reference T_15 is not declared.
```

There is a spec file chiselTests/VectorPacketIO.scala that demonstrates this error in the chisel3 branch DeqIO_contains_init_code.

chick:
I can fix the problem by removing the initialization code in DeqIO and EnqIO and moving it to a method, that the circuit must invoke but I am not sure if that is the correct approach.  This is a distillation of the Router example in the chisel-tutorials. Please advise

aswaterman:
@jackbackrack came up with the approach of emitting firrtl line by line so might have something to say about what to do about this.

aswaterman:
@chick I think there are a few problems here.  It's not well defined when it's legal to create Chisel nodes, so it's not even obvious whether this should be considered incorrect.  Regardless, this kind of issue clearly shouldn't manifest as an opaque FIRRTL error.

A possibly acceptable hack is to auto-declare Wires on first reference.  This requires the Chisel3 frontend be aware of FIRRTL scoping, which would be a significant departure--but possibly necessary.

sdtwigg:
[...]

PR #480, opened 2017-02-02
Added vec IO tests for #104

ucbjrl:
This is a reworking of da-steve101's #227:

  Add test cases for #104, #109 fixes Vec( vecSize, UInt( INPUT ) )

with all tests fixed and the Vec.fill() test with pure Chisel types (not values) expected to fail with a binding error.


chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -43,12 +43,16 @@ object Vec {
...
+
+  def do_apply[T <: Data](n: Int, gen: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
+    if ( gen.isLit ) Vec( Seq.fill(n)(gen) )
+    else new Vec(gen.chiselCloneType, n)
+  }
 
   @deprecated("Vec argument order should be size, t; this will be removed by the official release", "chisel3")
-  def apply[T <: Data](gen: T, n: Int)(implicit compileOptions: CompileOptions): Vec[T] =
-    apply(n, gen)
+  def apply[T <: Data](gen: T, n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] =
```
ducky64: For deprecated methods I usually use UnlocatableSourceInfo / DeprecatedSourceInfo, because it's not worth the pain of threading source info through

src/test/scala/chiselTests/Vec.scala
```diff
@@ -2,16 +2,81 @@
...
+
+class IOTesterMod( vecSize : Int ) extends Module {
+  val io = IO(new Bundle {
+    val in = Input(Vec( vecSize, UInt() ))
+    val out = Output(Vec( vecSize, UInt() ))
+  })
+  io.out := io.in
+}
+
+class LitTester(w: Int, values: List[Int] ) extends BasicTester {
```
ducky64: You should style your arguments consistently. I think we typically have no spaces after the begin parentheses or before the end parentheses, like `class MyClass(myVal: Type)`. But this like is just ridiculous, where you mix space and non-space...

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ucbjrl: But we still need to thread the CompileOptions, so is there any benefit to not threading the SourceInfo as well?

src/test/scala/chiselTests/Vec.scala
ucbjrl: Good point. Will clean up.

src/test/scala/chiselTests/Vec.scala
```diff
@@ -8,69 +8,69 @@ import chisel3.testers.BasicTester
...
 }
 
-class IOTesterModFill( vecSize : Int ) extends Module {
+class IOTesterModFill(vecSize : Int) extends Module {
   // This should generate a BindingException when we attempt to wire up the Vec.fill elements
   //  since they're pure types and hence unsynthesizeable.
   val io = IO(new Bundle {
-    val in = Input(Vec.fill( vecSize ) {UInt() })
-    val out = Output(Vec.fill( vecSize ) { UInt() })
+    val in = Input(Vec.fill(vecSize) {UInt() })
```
ducky64: style and spacing...

src/test/scala/chiselTests/Vec.scala
```diff
@@ -8,69 +8,69 @@ import chisel3.testers.BasicTester
...
+  val dut = Module(new IOTesterMod(values.length))
   dut.io.in := v
   for ((a,b) <- dut.io.out.zip(values)) {
     assert(a === b.U)
   }
   stop()
 }
 
-class IOTesterModFill( vecSize : Int ) extends Module {
+class IOTesterModFill(vecSize : Int) extends Module {
```
ducky64: do we have spaces before / after the `:` or not? Because this isn't even consistent within the same file...

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ducky64: Sure.

sdtwigg:
This PR was not named properly and the API change it introduced is concerning. The name implies the PR just added tests (so not of particular interest) but the PR redefined the instantiation of Vec.

The old Vec(Int, T) very reliably always returned a Vec as a type that had to be used somewhere else. Now, this is no longer guaranteed. Similarly, with https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Reg.scala#L11 , it seems like this line of code would be broken now, depending on CompileOptions:
```scala
Reg(Vec(1, 1.U))
```

mwachs5:
Errr, what? Yes that is concerning. What would that `Reg` code do now?

ucbjrl:
I agree with @sdtwigg concerns. We should revert the redefinition of the instantiation of Vec.

sdtwigg:
@mwachs5 throw an exception rather than instantiate a Reg of Vec of 1-bit UInt

PR #483, opened 2017-02-03
Add naming transforms to chisel utils

ducky64:
Won't work in all cases but is better than nothing.

Annotations aren't added in places where it obviously won't work (like recursive functions without a val definition). More discussion is needed on how to handle those cases (perhaps with variable interpolation?), but that's a later thing, preferably as we get more use cases.

Rocket builds and tests fine. Note that when compiling code with this, you may get warnings with "dependencies are invalidated unconditionally" - I think this means forced recompilation as a result of macros and should only show up once.

### DIFF
--- a/coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
+++ b/coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
@@ -82,9 +82,12 @@ class NamingTransforms(val c: Context) {
         tree  // don't recurse into inner classes
       case q"$mods def $tname[..$tparams](...$paramss): $tpt = $expr" => {
         val Modifiers(_, _, annotations) = mods
+        // don't apply naming transform twice
         val containsChiselName = annotations.map({q"new chiselName()" equalsStructure _}).fold(false)({_||_})
-        if (containsChiselName) {
-          tree  // don't apply the transform multiple times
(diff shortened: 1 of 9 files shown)


coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
```diff
@@ -82,9 +82,12 @@ class NamingTransforms(val c: Context) {
         tree  // don't recurse into inner classes
       case q"$mods def $tname[..$tparams](...$paramss): $tpt = $expr" => {
         val Modifiers(_, _, annotations) = mods
+        // don't apply naming transform twice
         val containsChiselName = annotations.map({q"new chiselName()" equalsStructure _}).fold(false)({_||_})
-        if (containsChiselName) {
-          tree  // don't apply the transform multiple times
+        // transforming overloaded initializers causes errors, and the transform isn't helpful
+        val isInitializer = tname == TermName("<init>")
+        if (containsChiselName || isInitializer) {
```
jackkoenig: Is an initializer a constructor? Just curious

ducky64: Yeah, this caused problem in one of the classes with multiple constructors (Queue, I think).

jackkoenig: Cool, well good to fix!

PR #490, opened 2017-02-07
Rename SeqMem to SyncReadMem - fixes #102.

### LINKED ISSUE #102: Call it SyncMem
Rename the new memory from sequential to synchronous memory (e.g., SyncMem). That's the term that is used for memory chips. E.g., SSRAM is a synchronous static RAM, in SDRAM the S is for synchronous.

Basically there are two types of memories (or better memory interfaces) available: asynchronous and synchronous. The term sequential is not use for memories. And if your really want to, I would say both asynchronous and synchronous are 'sequential'.

ucbjrl:
Retain un-deprecated SeqMem in compatibility mode, deprecate in chisel3.
Fixes #102.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Mem.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Mem.scala
@@ -15,7 +15,7 @@ object Mem {
   @deprecated("Mem argument order should be size, t; this will be removed by the official release", "chisel3")
   def apply[T <: Data](t: T, size: Int): Mem[T] = do_apply(size, t)(UnlocatableSourceInfo)
 
-  /** Creates a combinational-read, sequential-write [[Mem]].
+  /** Creates a combinational/asynchronous-read, sequential/synchronous-write [[Mem]].
     *
     * @param size number of elements in the memory
     * @param t data type of memory element
@@ -100,7 +100,7 @@ sealed abstract class MemBase[T <: Data](t: T, val length: Int) extends HasId wi
   }
 }
 
-/** A combinational-read, sequential-write memory.
+/** A combinational/asynchronous-read, sequential/synchronous-write memory.
   *
   * Writes take effect on the rising clock edge after the request. Reads are
   * combinational (requests will return data on the same cycle).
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/core/Mem.scala
```diff
@@ -143,7 +143,7 @@ object SeqMem {
   * @note when multiple conflicting writes are performed on a Mem element, the
   * result is undefined (unlike Vec, where the last assignment wins)
   */
-sealed class SeqMem[T <: Data](t: T, n: Int) extends MemBase[T](t, n) {
+sealed class SyncReadMem[T <: Data](t: T, n: Int) extends MemBase[T](t, n) {
   def read(addr: UInt, enable: Bool): T = {
```
jackkoenig: Is this not a backwards incompatible change in the compatibility layer as well? What if someone matches on the type of a Mem? Obviously they will get a compiler error but yeah.

ducky64:
"Does rocket still build with this?"

chiselFrontend/src/main/scala/chisel3/core/Mem.scala
ucbjrl: Don't these changes accomplish that (from src/main/scala/chisel3/compatibility.scala):
```
  val SeqMem = chisel3.core.SyncReadMem
  type SeqMem[T <: Data] = chisel3.core.SyncReadMem[T]
```

jackkoenig: It does!
```scala
scala> class Foo(val a: Int)
defined class Foo

scala> type Bar = Foo
defined type alias Bar

scala> val bar = new Bar(3)
bar: Foo = Foo@33433fc5

scala> bar.isInstanceOf[Bar]
res0: Boolean = true

scala> bar match {
     | case _: Bar => print("Success!")
     | }
Success!
```
It doesn't work if you try to do reflection though:
```scala
scala> bar.getClass
res5: Class[_ <: Foo] = class Foo
```
But this is probably okay

ucbjrl:
rocket-chip builds and passes [jenkins build 1116](http://a17.millennium.berkeley.edu/jenkins/job/rocket-chip_with_chisel3/1116/)

PR #493, opened 2017-02-07
Update comments describing Decoupled/ReadyValid - fix #437.

### LINKED ISSUE #437: Are `EnqIO` and `DeqIO` backwards?
I see that `EnqIO` is used on the _output_ of `Queue` and vice-versa. Is this intentional? Chisel2 did not use `EnqIO` or `DeqIO` in `Queue` at all (as far as I can tell).

https://github.com/ucb-bar/chisel3/blob/3215df07a97b58babb1deb3fab0928198b1daad2/src/main/scala/chisel3/util/Decoupled.scala#L131-L146

ucbjrl:
It seems there is a valid use case for EnqIO/DeqIO and updating the comments may clear some of the confusion and encourage their usage.


src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -11,9 +11,10 @@ import chisel3.core.ExplicitCompileOptions.NotStrict
 
 /** An I/O Bundle containing 'valid' and 'ready' signals that handshake
   * the transfer of data stored in the 'bits' subfield.
-  * The base protocol implied by the directionality is that the consumer
-  * uses the flipped interface. Actual semantics of ready/valid are
-  * enforced via use of concrete subclasses.
+  * The base protocol implied by the directionality is that the producer uses
+  * the interface as-is (outputs bits) while the consumer
+  * uses the flipped interface (inputs bits).
```
edwardcwang: Seems like the comment could be better reflowed here?

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -131,21 +134,36 @@ object Irrevocable
...
 class QueueIO[T <: Data](gen: T, entries: Int) extends Bundle
 {
-  /** I/O to enqueue data, is [[Chisel.DecoupledIO]] flipped */
+  /* These may look inverted, because the names (enq/deq) are from the perspective of the client,
+   *  but internally, the queue implementation itself sits on the other side
+   *  of the interface so uses the flipped instance.
+   * These could be replaced with:
+   *   val enq = Flipped(EnqIO(gen))
+   *   val deq = Flipped(DeqIO(gen))
+   *  once issue #492 is resolved.
```
jackkoenig: This comment should be changed

edwardcwang: What should it be changed to?

PR #494, opened 2017-02-07
Remove explicit import of NotStrict - fixes #492

### LINKED ISSUE #492: remove import chisel3.core.ExplicitCompileOptions.NotStrict
There is a:
```
// TODO: remove this once we have CompileOptions threaded through the macro system.
import chisel3.core.ExplicitCompileOptions.NotStrict
```
in 7 chisel files.
Is it time?

I'd like to replace the following in QueueIO:
```
  val enq = DeqIO(gen)
  val deq = EnqIO(gen))
```
with
```
  val enq = Flipped(EnqIO(gen))
  val deq = Flipped(DeqIO(gen))
```
to help reduce some of the confusion, but the NotStrict compile options introduce a connection error when cloneChiselType() fails to preserver directionality (due to 09282f3).

ucbjrl:
DON'T merge this until #476 is merged.


ucbjrl:
This currently (2/7/17) fails in rocket-chip. See [jenkins rocket-chip build 1115](http://a17.millennium.berkeley.edu/jenkins/job/rocket-chip_with_chisel3/1115/console).

colinschmidt:
Wow I was just trying to write some new code in rocket-chip today and encountered a very similar error. I don't think undirectioned bundles are given the "default" direction correctly.

colinschmidt:
Mine occured in a more complicated bundle though:

```
Caused by: chisel3.internal.ChiselException: Connection between left (uncore.tilelink.ClientUncachedTileLinkIO@3e641d) and source (uncore.tilelink.ClientUncachedTileLinkIO@487385) failed @.grant.bits.data: Locally unclear whether Left or Right (both internal)
```

chiselFrontend/src/main/scala/chisel3/core/Mem.scala
```diff
@@ -36,27 +36,30 @@ sealed abstract class MemBase[T <: Data](t: T, val length: Int) extends HasId wi
   /** Creates a read accessor into the memory with static addressing. See the
     * class documentation of the memory for more detailed information.
     */
-  def apply(idx: Int): T = {
+  def apply(idx: Int)/*(implicit compileOptions: CompileOptions)*/: T = {
+    // We can't force this method to take an implicit parameter, since its signature comes from collection.IndexedSeq
+    // How do we thread the caller's CompileOptions through here?
+    import chisel3.core.ExplicitCompileOptions.NotStrict
```
ucbjrl: What's the correct way to do this? Do we need to make this `apply` method a macro so we can materialize the client's `compileOptions` and pass them along?

ucbjrl:
I've think I've correctly threaded the compileOptions through all the required methods (all tests pass), but is there a better way to do this (i.e., as part of sourceInfo setup)?

aswaterman:
Since #476 has been merged, is this good to go (once conflicts are resolved)?

chiselFrontend/src/main/scala/chisel3/core/Mem.scala
ducky64: If you expect a chained apply, it does need to be a macro, otherwise the compiler gets really confused.

chiselFrontend/src/main/scala/chisel3/core/Mem.scala
```diff
@@ -30,16 +30,30 @@ object Mem {
   }
 }
 
-sealed abstract class MemBase[T <: Data](t: T, val length: Int) extends HasId with VecLike[T] {
+/* The following definition of ChiselVecLike is to work around the issue that you cannot
```
ucbjrl: @ducky64 could you review this please? There may be a better way.

ducky64: So what we've done in this cases like this is to have the implementation (do_apply) in a separate function which is abstract, and the macro transform defined at the abstract class. Since the macro only does a naming change (without resolving which subclass / etc) and the type signatures of both the macro method and function implementation are the same, the same macro should work in all subclasses.

So basically I would define the apply macro in VecLike. Vec's apply method may need to be changed to take compile options it ultimately doesn't need, but that should be fine.

ucbjrl: I'm not sure I completely grok the suggestion. We need a concrete `apply(x: Int): T` method that the macro can override. I've added a suitable definition to `VecLike`. Is this what was intended?

ducky64: So basically move the `apply` macro into VecLike (so the transform happens with all VecLike subtypes), and define an abstract `do_apply` in VecLike (which is what the macro renames things into). `do_apply` is just a normal function, and can be overridden normally.

ucbjrl:
@ducky64, could you re-review this please?

ucbjrl:
retest this please.

PR #498, opened 2017-02-09
Fix random failures in CompatibilitySpec

jackkoenig:
0 failed because log2Ceil requires > 0, 1 failed because (1 << 1) - 1 = 1 which is a power of 2, and 32 (+more >32 probably) failed the isPow2 tests because integer overflow.


aswaterman:
These tests are a little questionable since they just repeat their definitions.

jackkoenig:
I believe the intent is merely to ensure that these calls work in the compatibility wrapper

ucbjrl:
Thank you Jack. I stared at that code on at least two separate occasions,
and convinced myself it couldn't fail.

I didn't spot the log2Ceil() problem at all.

 Missed it by that much.

PR #500, opened 2017-02-09
Escape % in assertion messages

jackkoenig:
Note that currently assertions with % in the message or in the line or with % in the filename will all fail to compile


colinschmidt:
Are there any other things we need to escape?

jackkoenig:
I don't think so, regular escape characters come already escaped in Scala. `%` is the only format specifier so it should be the only thing from Scala strings we have to explicitly escape.

Side note, I need to implement assertions accepting Printables so people can print useful diagnostic information easily.

PR #504, opened 2017-02-12
Add firrtl as git submodule

jackkoenig:


jackkoenig:
This currently fails because jenkins needs to do a `git submodule update --init`

I removed the option of Firrtl as a library dependency, but I am open to the idea of adding it back where basically Firrtl is used as a library dependency if directory `./firrtl` does not exist. Conditional `Project.dependsOn(...)` requires switching to a `build.scala` though (to the best of my knowledge). Another reason we may not want to do this is that if someone includes chisel in a project and only does `git submodule update --init` and not `git submodule update --init --recursive`, then they would use the version of chisel they have in the repo, but yet still pull Firrtl from the library dependency. It seems likely to me that anyone who checks out chisel3 is in full development mode and would rather also use Firrtl as a git submodule as well.

ben-k:
It would be nice to avoid having to `git submodule update --init --recursive` in rocket-chip to get RTL to build, as this grabs all of the `riscv-tools` recursion and so takes an order of magnitude longer than `git submodule update --init`.

This is obviously a nit, but it's also the most common use case for chisel3-as-submodule, so maybe worth accounting for?

edwardcwang:
So in that case you would modify rocket-chip to support a publish-local'ed chisel3?

aswaterman:
@ben-k we could include a shell script in rocket-chip that fetches only the requisite submodules and avoids fully recursing into riscv-tools.  (The travis setup already works this way.)

jackkoenig:
@edwardcwang `publish-local` doesn't really work because it publishes a version that is global to your machine when what we're really trying to do is have a version for a specific checkout of rocket-chip or another project.

@ben-k Yeah the `git submodule update --init --recursive` is just to illustrate that this PR would require people including chisel3 as a submodule would need to get firrtl inside of chisel3 as well. I tried really hard to allow the firrtl dependency to live above chisel3 but I think sbt won't let that happen in any consistent way.

PR #505, opened 2017-02-13
Fixed point factory stuff

chick:
creation of FixedPoint types and literals was not consistent with current practices
for other types.  This commit brings them in-line, deprecated previous usages.
Also fixed a couple of return type  problems with shift operators on FixedPoint


shunshou:
Should `def apply(dir: Direction, width: Width, binaryPoint: BinaryPoint)` still be used?

Also, please also specify a "one-true" way to represent a fixed point literal with fixed width (not just binary point!) -- this is important for stuff like LUTs

chick:
@shunshou I'd love to get rid of the FixedPoint factory methods that take a direction.  I don't see any usages but I was worried there might be some, I guess I'll just do it then.  
As for OTW, I'm torn between
``` 3.3.F(10.BP,16.W)``` or ``` 3.3.F(16.W, 10.BP)``` 
I'm leaning to the second because it's more in line with the class argument structure, but I think ease of use should trump that.  Which would you prefer

shunshou:
I'm satisfied w/ your W before BP thing.

shunshou:
Can someone double check why the tests are failing? and if it's ok plz merge? I'd like to change over some of my dsptools code now that I have time to work on Chisel-y things again...

chick:
@shunshou I can't get to Jenkins right now, but it is probably a firrtl update that has not propagated yet.  I'll look into it

chick:
retest this please

PR #509, opened 2017-02-15
Add support for clock and reset scoping

jackkoenig:
withClockAndReset allows multi-clock within a Module

At long last, multi-clock within a Module!

I had written a test that inspected the Firrtl but realized that was not nearly as good as execution tests.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
@@ -12,7 +12,14 @@ import chisel3.internal.sourceinfo.{SourceInfo}
 private[chisel3] final case class ClockAndReset(clock: Clock, reset: Bool)
 
 object withClockAndReset {  // scalastyle:ignore object.name
-  def apply[T](clock: Clock, reset: Bool)(block: => T)(implicit sourceInfo: SourceInfo): T = {
+  /** Creates a new Clock and Reset scope
+    *
+    * @param clock the new implicit Clock
+    * @param reset the new implicit Reset
+    * @param block the block of code to run with new implicit Clock and Reset
+    * @return the result of the block
+    */
+  def apply[T](clock: Clock, reset: Bool)(block: => T): T = {
     // Save parentScope
     val parentScope = Builder.currentClockAndReset
     Builder.currentClockAndReset = Some(ClockAndReset(clock, reset))
@@ -21,5 +28,27 @@ object withClockAndReset {  // scalastyle:ignore object.name
     Builder.currentClockAndReset = parentScope
     res
   }
+}
+
+object withClock {  // scalastyle:ignore object.name
+  /** Creates a new Clock scope
+    *
+    * @param clock the new implicit Clock
+    * @param block the block of code to run with new implicit Clock
+    * @return the result of the block
+    */
+  def apply[T](clock: Clock)(block: => T): T =
+    withClockAndReset(clock, Module.reset)(block)
+}
+
+object withReset {  // scalastyle:ignore object.name
+  /** Creates a new Reset scope
+    *
+    * @param reset the new implicit Reset
+    * @param block the block of code to run with new implicit Reset
+    * @return the result of the block
+    */
+  def apply[T](reset: Bool)(block: => T): T =
+    withClockAndReset(Module.clock, reset)(block)
 }
 
--- a/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
@@ -1,0 +1,25 @@
+// See LICENSE for license details.
+
+package chisel3.core
+
+import scala.language.experimental.macros
+
+import chisel3.internal._
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl._
+import chisel3.internal.sourceinfo.{SourceInfo}
+
+private[chisel3] final case class ClockAndReset(clock: Clock, reset: Bool)
+
+object withClockAndReset {  // scalastyle:ignore object.name
+  def apply[T](clock: Clock, reset: Bool)(block: => T)(implicit sourceInfo: SourceInfo): T = {
+    // Save parentScope
+    val parentScope = Builder.currentClockAndReset
+    Builder.currentClockAndReset = Some(ClockAndReset(clock, reset))
+    val res = block // execute block
+    // Return to old scope
+    Builder.currentClockAndReset = parentScope
+    res
+  }
+}
+
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -151,6 +151,7 @@ private[chisel3] class DynamicContext() {
   // Used to distinguish between no Module() wrapping, multiple wrappings, and rewrapping
   var readyForModuleConstr: Boolean = false
   var whenDepth: Int = 0 // Depth of when nesting
+  var currentClockAndReset: Option[ClockAndReset] = None
   val errors = new ErrorLog
   val namingStack = new internal.naming.NamingStack
 }
@@ -186,6 +187,16 @@ private[chisel3] object Builder {
   def whenDepth_=(target: Int): Unit = {
     dynamicContext.whenDepth = target
   }
+  def currentClockAndReset: Option[ClockAndReset] = dynamicContext.currentClockAndReset
+  def currentClockAndReset_=(target: Option[ClockAndReset]): Unit = {
+    dynamicContext.currentClockAndReset = target
+  }
+  def forcedClockAndReset: ClockAndReset = currentClockAndReset match {
+    case Some(clockAndReset) => clockAndReset
+    case None => throwException("Error: No implicit clock and reset.")
+  }
+  def forcedClock: Clock = forcedClockAndReset.clock
+  def forcedReset: Bool = forcedClockAndReset.reset
 
   // TODO(twigg): Ideally, binding checks and new bindings would all occur here
   // However, rest of frontend can't support this yet.
--- a/chiselFrontend/src/main/scala/chisel3/core/Assert.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Assert.scala
@@ -51,12 +51,12 @@ object assert { // scalastyle:ignore object.name
   }
 
   def apply_impl_do(cond: Bool, line: String, message: Option[String], data: Bits*)(implicit sourceInfo: SourceInfo) {
-    when (!(cond || Builder.forcedModule.reset)) {
+    when (!(cond || Builder.forcedReset)) {
       message match {
         case Some(str) => printf.printfWithoutReset(s"Assertion failed: $str\n    at $line\n", data:_*)
         case None => printf.printfWithoutReset(s"Assertion failed\n    at $line\n", data:_*)
       }
-      pushCommand(Stop(sourceInfo, Node(Builder.forcedModule.clock), 1))
+      pushCommand(Stop(sourceInfo, Node(Builder.forcedClock), 1))
     }
   }
 
@@ -76,8 +76,8 @@ object assert { // scalastyle:ignore object.name
 object stop { // scalastyle:ignore object.name
   /** Terminate execution with a failure code. */
   def apply(code: Int)(implicit sourceInfo: SourceInfo): Unit = {
-    when (!Builder.forcedModule.reset) {
-      pushCommand(Stop(sourceInfo, Node(Builder.forcedModule.clock), code))
+    when (!Builder.forcedReset) {
+      pushCommand(Stop(sourceInfo, Node(Builder.forcedClock), code))
     }
   }
 
--- a/chiselFrontend/src/main/scala/chisel3/core/Mem.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Mem.scala
@@ -27,7 +27,7 @@ object Mem {
     // TODO(twigg): Remove need for this Binding
 
     val mem = new Mem(mt, size)
-    pushCommand(DefMemory(sourceInfo, mem, mt, size)) // TODO multi-clock
+    pushCommand(DefMemory(sourceInfo, mem, mt, size))
     mem
   }
 }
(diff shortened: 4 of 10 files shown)


chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -186,6 +187,16 @@ private[chisel3] object Builder {
...
+  def currentClockAndReset: Option[ClockAndReset] = dynamicContext.currentClockAndReset
+  def currentClockAndReset_=(target: Option[ClockAndReset]): Unit = {
+    dynamicContext.currentClockAndReset = target
+  }
+  def forcedClockAndReset: ClockAndReset = currentClockAndReset match {
+    case Some(clockAndReset) => clockAndReset
+    case None => throwException("Error: No implicit clock and reset.")
+  }
+  def forcedClock: Clock = forcedClockAndReset.clock
+  def forcedReset: Bool = forcedClockAndReset.reset
```
aswaterman: What's the idea behind the "forced" terminology, instead of "current" or "implicit"?

aswaterman:
hooray

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: I was mirroring the `currentModule` and `forcedModule`, but my interpretation is that `currentClockAndReset` is an `Option` because you can be outside of a clock and reset scope (esp. once we have `RawModule`). `forcedClock` and `forcedReset` are documenting their use that you must have a clock and a reset in scope. It also provides a nice shorthand to just get the current Clock or Reset with the check built-in

aswaterman: ack

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import chisel3.internal._
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl._
+import chisel3.internal.sourceinfo.{SourceInfo}
+
+private[chisel3] final case class ClockAndReset(clock: Clock, reset: Bool)
+
+object withClockAndReset {  // scalastyle:ignore object.name
+  def apply[T](clock: Clock, reset: Bool)(block: => T)(implicit sourceInfo: SourceInfo): T = {
```
ducky64: Would it make sense to also have just `withClock` and `withReset` (so the user doesn't need to thread those though manually, especially since `clock` and `reset` aren't seen in user-level Module code?

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
```diff
@@ -0,0 +1,25 @@
...
+    // Save parentScope
+    val parentScope = Builder.currentClockAndReset
+    Builder.currentClockAndReset = Some(ClockAndReset(clock, reset))
+    val res = block // execute block
+    // Return to old scope
+    Builder.currentClockAndReset = parentScope
+    res
```
ducky64: I can see why returning a value might be useful, but there's also the pathological case of trying to return a function defined in the withClockAndReset scope. Invoking it outside would cause it to execute in the outer clock and reset scope, which might not be what is expected.
That being said, I think that use case would look obviously wrong, and even without this feature a user could just overwrite a var in an outer scope.
Are there other edge cases that we haven't thought of yet that might lead to unintuitive behavior?

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
aswaterman: The user should be able to get at the implicit clock and reset *somehow*.

jackkoenig: Yeah we do need a way to get at the current `clock` and `reset` as a user since it might be relevant in a function or something

jackkoenig: I'm also happy to add `withClock` and `withReset` since people will often want to change one but not the other

mwachs5: `clock` and `reset` are currently available in user-level Module code, right? Is this a change in that functionality? And if they are still available, what does the following code do:

``` 
  
 val io = new Bundle { val otherClock = Clock(INPUT)); }

  WithClock(io.otherClock) {

  val subModule = Module(new SubModule())
  subModule.clock := clock

}
```

What clock does subModule actually get in the end? Or is something written above illegal now?

jackkoenig: Yes, currently `clock` and `reset` are available in user-level Module code. This PR does nothing to change that, however, your example illustrates the problem. Inside of a Module, `clock` refers to the implicit Clock input port to the Module. Thus, in your example, `subModule.clock` gets connected to the `clock` from the Module, **not** `io.otherClock`

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
jackkoenig: Yeah you make a solid point. I decided to return a value because I really like that style of coding. In particular, this would allow people to do things like:

```scala
val myModule = withClockAndReset(someClock, someReset) { Module(new MyModule)) }
```
And thus neatly get a Module with a different clock and/or reset in a single line. It is important that this return value doesn't set people up for failure. @sdtwigg @aswaterman can y'all think of any other reasons why returning a value might be a bad idea? I'm not too too concerned about people trying to return functions but if others think that's reason enough I can switch to returning Unit.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
mwachs5: Ok. I think I like that better anyway. 

Is there a name for the thing inside the withClock block?

jackkoenig: To my knowledge, there currently isn't any way to access the clock or reset of the current Module inside of a function. Obviously this PR provides much more power for changing them so people are more likely to do so, just mentioning that this isn't removing anything that existed

mwachs5: I'm fine with clocks being required to be more explicit vs implicit. Users should *not* be using this stuff much. 

Is there another PR or issue for this relating to AsyncReset? Is that going to be done in a similar manner somehow?

jackkoenig: This PR is intended to lead in to `AsyncReset`. The idea is that `AsyncReset` will be a new type that if you pass it as the reset to `withClockAndReset` it will construct `AsyncResetRegs` (working name, suggestions welcome) and Modules where the implicit reset is an `AsyncReset`.

mwachs5: > To my knowledge, there currently isn't any way to access the clock or reset of the current Module inside of a function. Obviously this PR provides much more power for changing them so people are more likely to do so, just mentioning that this isn't removing anything that existed

This is where the horrendous (in my opinion) `AsyncScope` hack came from... you instantiate an AsyncScope then ask it what its clock is. This creates an empty module for no good reason.

But yeah. I'd prefer that if people really need to know the clock inside their function, then it should be passed as a parameter to the function.

mwachs5: One more question about my example above...

does the module get the right name? Will it be called subModule even though it's inside a function (I think, I don't fully grasp all that is going on behind the scenes here).

aswaterman: Since AsyncScope is legal, people will use it to materialize the implicit clock and reset.  Can't ban it, so might as well embrace it.  I'd suggest adding them as methods on the Module object (i.e. Module.clock gives the current module's implicit clock).

ducky64: Passing in clocks to functions explicitly does make sense (though it's inconsistent with how Regs/etc have clocks implicitly passed). However, there's also an argument for getting the implicit clock to make moving code easier across withClock blocks, since otherwise you have to manually keep track of the clock across those blocks, even within the same function scope.

aswaterman: IMO, whether the clock is passed into a function or taken implicitly should be a matter of style--there are clearly cases where both are appropriate.

mwachs5: Yeah. Just like in Verilog there are good ways and bad ways to do multi-clock stuff. Not doing insane things and using a small set of well tested techniques will be up to the user/ whoever reviews PRs with code that has these in it.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
aswaterman: The example is such a compelling use case that I think we should just say "just don't do that" to @ducky64's example.

If it returns Unit, people will just use vars to achieve the same thing badly.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
jackkoenig: Alright so my current plan is to do the following:
* Add `Module.clock` and `Module.reset` (object Module) to get the current implicit clock and reset
* Add `withClock` and `withReset` to remove boilerplate for those common cases
* Mark `with*` functions experimental because I forgot to do that. Can't really mark the `Module.[clock|reset]` functions experimental though

Any thoughts on the return value below?

ducky64: Should  you name it implicitClock or implicitReset (or similar)? `Module.clock` and `.reset` are module-level IO signals and overloading their functionality (if that's even possible) may not be a good idea?

Everything else looks good.

ducky64: Also when the module hierarchy refactor happens, withClock/withReset blocks can be used outside ImplicitModule (or just Module)

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
colinschmidt: I agree that it should return a value. If you can't use the module return example its going to a big loss to the usefulness of this construct.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
mwachs5: RE adding the Module.clock... If I call a function inside a withClock block and then inside that function it calls Module.clock... what clock are we talking about then?

Concrete example:

```
object ClockConnector {
  def apply(Clock ck) {
    ck := Module.clock
  }
}

... 

val FooModule extends Module {
  val io = new Bundle {val otherClock = Clock(INPUT) }
  val yetAnotherClock = Wire(Clock())

  withClockAndReset(io.otherClock) {
    ConnectClock(yetAnotherClock)  
  }
```

what is `yetAnotherClock` connected to?

jackkoenig: I'm fine with `.implicitClock` and `.implicitReset`. Also keep in mind we're not talking about overloading the IO signals functionality because this is object `Module.clock`, ie. inside of a given `Module`, `this.clock` or just `clock` returns the input port `clock` to that `Module` while `Module.clock` (or `Module.implicitClock`) would return the actual current implicit clock set by the most recent `withClock`. 

Any easy confusion here though might be reason enough to not call them `.clock` and `.reset`

jackkoenig: `yetAnotherClock` would get connected to `io.otherClock`

aswaterman: I propose using Module.clock and Module.reset since they don't actually collide with anything.  It makes more sense that the implicit clock is always named "clock", rather than sometimes "clock" and other times "implicitClock".

jackkoenig: Makes sense to me. What should `Module.clock` and `Module.reset` return? Should they return `Option[...]` or just return the clock or reset and error if there is none? We could also have `Module.clockOption` and `Module.resetOption` for those use cases (analogous to Scala `.head` and `.headOption`)

ducky64: I think there should be some way of just getting the clock and reset signal without Option[...], much like how you can declare a Reg.

In principle, it's nice to have clockOption, though I'm not sure if it's practically useful? Especially since there's no similar construct for Reg, and what are you going to do (other than re-throw exceptions) if there isn't a clock?

aswaterman: I don't think the important use case is to make a decision based upon whether there is an implicit clock--you really just want the implicit clock.  So throwing an exception seems like the right behavior, and as @ducky64 points out, matches what would happen if you declare a Reg without an implicit clock present.

jackkoenig: @ducky64 I currently have `(implicit sourceInfo: SourceInfo)` at the end but since nothing in the function needs sourceInfo I'm guessing that's unnecessary right? The code in the block should get the sourceInfo on it's own?

jackkoenig:
I've added `withClock`, `withReset`, `Module.clock`, and `Module.reset`. I also moved `withClock`, `withReset`, and `withClockAndReset` to `chisel3.experimental`.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
ducky64: Yeah, the code in the block will get source info where it is defined. The structure of source info should mean that it won't materialize in chisel core, so if you forget to thread it through an internal construct it should be caught at compile time.

jackkoenig: Cool, I removed it in the most recent iteration

jackkoenig:
I know you already approved but @ducky64, would you mind glancing over the new changes?

ducky64:
Still looks good.

jackkoenig:
Alright I'm going to merge

PR #510, opened 2017-02-15
Use firrtl elses in elsewhen/otherwise case emission

albert-magyar:
This changes Chisel3's firrtl emission to use nested firrtl `when` and `else` statements for chisel `when`, `elsewhen`, and `otherwise` blocks.

There is something of a naming confusion with 132b80edee2fb8e730d3b6f5eb5f36051a819525, which adds a counter to keep track of the depth of the chisel `when` "level." This branch keeps track of the firrtl `when` nesting with a similarly named depth, which does not hold the same value.

The two features could potentially be combined, especially given that `Builder.whenDepth` is only ever checked for nonzero values.

Tested with ucb-bar/rocket-chip@abe344a1a4713ec766128f3dcd5314d187f3bdbf, ucb-bar/firrtl@8f3c7d357fd6c97121d5adbb0ca09ce34f763374


chiselFrontend/src/main/scala/chisel3/core/When.scala
```diff
@@ -27,35 +27,36 @@ object when {  // scalastyle:ignore object.name
...
 
-/** Internal mechanism for generating a when. Because of the way FIRRTL
-  * commands are emitted, generating a FIRRTL elsewhen or nested whens inside
-  * elses would be difficult. Instead, this keeps track of the negative of the
-  * previous conditions, so when an elsewhen or otherwise is used, it checks
-  * that both the condition is true and all the previous conditions have been
-  * false.
-  */
-final class WhenContext(sourceInfo: SourceInfo, cond: Bool, prevCond: => Bool, block: => Unit) {
+final class WhenContext(sourceInfo: SourceInfo, cond: Option[() => Bool], block: => Unit, depth: Int = 0) {
```
azidar: Add comment explaining new WhenContext behavior

azidar: Maybe label `depth` as `firrtlDepth`

src/main/scala/chisel3/internal/firrtl/Emitter.scala
```diff
@@ -75,8 +82,9 @@ private class Emitter(circuit: Circuit) {
           // Firrtl extmodule can overrule name
           body ++= newline + s"defname = ${bb.id.desiredName}"
           body ++= newline + (bb.params map { case (n, p) => emitParam(n, p) } mkString newline)
-        case mod: DefModule => for (cmd <- mod.commands) {
-          body ++= newline + emit(cmd, mod)
+        case mod: DefModule => {
+          val procMod = mod.copy(commands = processWhens(mod.commands))
```
azidar: Add comment explaining the post-processing of the whens

albert-magyar:
ucb-bar/firrtl#601 now transforms the code generated under this PR to have the same expanded mux tree structure as the prior when emission style.

jackkoenig:
I have run DC with recent rocket-chip + FPU and as far as I can tell, this PR does not cause a QoR regression. Reason why results are different from past experiments is unclear, but I think we should move forward with this. Not only does it enable the invalidate API, it also improves coverage.

albert-magyar:
With the added emission patch (#601) or just as is?

jackkoenig:
As is, I just merged this branch into the branch of chisel3 being used by rocket-chip, no changes to firrtl.

jackkoenig:
There have been lots of changes to rocket-chip in the last few weeks, especially to critical paths. I suspect these changes may have removed the problematic path(s).

PR #511, opened 2017-02-15
Add scaladoc examples for Vec and Bundle

chick:
Added some simple example for Vec and Bundle creation.
Comments welcome

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -155,17 +155,12 @@ object Vec {
 /** A vector (array) of [[Data]] elements. Provides hardware versions of various
   * collection transformation functions found in software array implementations.
   *
-  * @tparam T type of elements
-  * @note when multiple conflicting assignments are performed on a Vec element,
-  * the last one takes effect (unlike Mem, where the result is undefined)
-  * @note Vecs, unlike classes in Scala's collection library, are propagated
-  * intact to FIRRTL as a vector type, which may make debugging easier
-  *
   * Careful consideration should be given over the use of [[Vec]] vs [[Seq]] or some other scala collection. In
   * general [[Vec]] only needs to be used when there is a need to express the hardware collection in a [[Reg]]
   * or IO [[Bundle]] or when access to elements of the array is indexed via a hardware signal.
+  *
   * Example of indexing into a [[Vec]] using a hardware address and where the [[Vec]] is defined in an IO [[Bundle]]
-  * @example
+  *
   *  {{{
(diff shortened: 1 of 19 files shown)


chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -160,6 +160,21 @@ object Vec {
   * the last one takes effect (unlike Mem, where the result is undefined)
   * @note Vecs, unlike classes in Scala's collection library, are propagated
   * intact to FIRRTL as a vector type, which may make debugging easier
+  *
+  * Careful consideration should be given over the use of [[Vec]] vs [[Seq]] or some other scala collection. In
+  * general [[Vec]] only needs to be used when there is a need to express the hardware collection in a [[Reg]]
+  * or IO [[Bundle]] or when access to elements of the array is indexed via a hardware signal.
+  * Example of indexing into a [[Vec]] using a hardware address and where the [[Vec]] is defined in an IO [[Bundle]]
+  *  {{{
```
ducky64: Consider use of the `@example` tag?

ducky64: Do you need the full Module, or can the example be more minimal?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -410,6 +425,34 @@ abstract class Record extends Aggregate {
   *
   * Usage: extend this class (either as an anonymous or named class) and define
   * members variables of [[Data]] subtypes to be elements in the Bundle.
+  *
+  * Example of an anonymous IO bundle
```
ducky64: `@example` tag?

ducky64:
Did you miss tagging the second example? In Bundle?

chick:
@ducky64 as a matter of fact I did miss second @example, tag fixed now

jackkoenig:
How does it actually look with the example tags? I think it might move them away so that text in between them may not make sense

ducky64:
You can preview by building the scaladoc?

jackkoenig:
The `@example` tag doesn't seem to be doing anything. And the first example in Bundle still doesn't have it

jackkoenig:
Actually, I was behind, the tag does move the example elsewhere, so maybe the tag isn't right for the style here? [Scala List](https://github.com/scala/scala/blob/v2.12.1/src/library/scala/collection/immutable/List.scala#L36) seems to use the tag sometimes but not always depending on what they're trying to do

chick:
@jackkoenig That is weird stuff, looking into it

PR #515, opened 2017-02-17
Builderreflectionfix

shunshou:
Should supersede #456 @sdtwigg @ducky64


ducky64:
Make sure you do a squash merge, so that the broken commit doesn't make it into the main tree (in case there's someone out there playing with bisect or something)

PR #517, opened 2017-02-17
Remove check on potentialy empty string for Record names.

mwachs5:
@terpstra -- apparently I don't have the ability to push to Chisel3. 

This does what I'd expect, but seems perhaps FIRRTL can't handle numbers as names either.
Note the difference between how a Vec (the uarts) and a Record (the spis) are output to FIRRTL.

```
    io.uarts[0] <- uart_0.io.port @[UARTPeriphery.scala 36:8]
    io.uarts[1] <- uart_1.io.port @[UARTPeriphery.scala 36:8]
    io.spis.0 <- spi_0.io.port @[SPIPeriphery.scala 32:8]
    io.spis.1 <- spi_1.io.port @[SPIPeriphery.scala 32:8]
...
    uart_pins_0.io.uart <- RocketChipTop.io.uarts[0]
    uart_pins_1.io.uart <- RocketChipTop.io.uarts[1]
    spi_pins_0.io.spi <- RocketChipTop.io.spis.0 
    spi_pins_1.io.spi <- RocketChipTop.io.spis.1 
```

FIRRTL gives these errors on the SPI lines:

```
line 265272:12 no viable alternative at input 'io.spis.0'
line 265273:12 no viable alternative at input 'io.spis.1'
...
line 266421:52 no viable alternative at input '0'
line 266422:52 no viable alternative at input '1'
```

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -26,8 +26,8 @@ private[chisel3] class Namespace(keywords: Set[String]) {
     def legalStart(c: Char) = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_'
     def legal(c: Char) = legalStart(c) || (c >= '0' && c <= '9')
     val res = s filter legal
-    val headOk = if (leadingDigitOk) legal(res.head) else legalStart(res.head)
-    if (res.isEmpty || !headOk) s"_$res" else res
+    val headOk = (!res.isEmpty) && (leadingDigitOk || legalStart(res.head))
+    if (headOk) res else s"_$res"
   }
 
   def contains(elem: String): Boolean = names.contains(elem)


terpstra:
Well, shit. Maybe firrtl should elide back-to-back _s? Might cause a namespace collision, though.

How bad is it if we leave it as foo__0 and foo__1?

mwachs5:
Well, in this case it means the top level signals of your Module are changing. So, that's up to you how annoying you find that...

terpstra:
I was planning to use Record in diplomacy, so those double __s would infect nearly everything.

mwachs5:
@jackkoenig does this sound like the correct diagnosis of this problem? Or is it actually something else going on here...

jackkoenig:
Yeah Firrtl doesn't allow numbers as the first character in identifiers but for names of fields we should probably allow it. Thoughts @azidar?

jackkoenig:
When are you seeing records with empty names?

mwachs5:
The names aren't empty. This PR was to attempt to get rid of the difference between a list of things as Vec and a list of Record (e.g. HeterogeneousBag). The names end up with an extra '_' in the Record case, e.g. in the above example the old behavior was:

```
io_uarts_0_tx
io_uarts_0_rx
io_uarts_1_tx
io_uarts_1_rx
...
io_spis__0_sck
io_spis__0_cs_0
io_spis__0_cs_1
io_spis__0_cs_2
io_spis__1_sck
io_spis__1_cs_0
```

We want to use the HeterogenousBag way because the size of CS varies from one SPI to another (UARTs are all the same in this example). But it's annoying when the top level name of your thing depends on whether it was a Vec or a Record.

jackkoenig:
Yeah I don't see any reason why Firrtl can't accept digits as the first character in names of Bundles, I'll look into it

azidar:
The original reason for not allowing numbers as bundle fields was so there wasn't a name conflict when expanding a vec versus a bundle. However, since the Uniquify pass in Firrtl solved this, I don't think there is a fundamental issue any longer (as long as the parser can still handle it).

jackkoenig:
The parser doesn't currently support it but I've started looking into it.

mwachs5:
I assigned you to review it because I was merging it to *your* original branch. This is trying to merge into your branch.

PR #528, opened 2017-02-24
Remove log2Up and log2Down from API

aswaterman:
These methods are a scourge.  They are mathematically incorrect (log2(1) should yield 0, not 1), and so their use results in corner-case bugs that never fail to surprise good programmers.  For an example that occurred just this evening, see: https://github.com/ucb-bar/rocket-chip/pull/563

For historical context, log2Up was created within rocket-chip to work around zero-width wire limitations in chisel1.  A chisel maintainer imported them into chisel, apparently without much regard for the fact that they were deliberately defined incorrectly.  So, while we must keep them in the compatibility layer, we should take this compatibility-breaking opportunity to purge them.

This PR moves these methods to the compatibility layer and purges all of their use in the chisel codebase itself.  In most cases within Chisel itself, the arguments to log2Up are easily shown to be at least 2, so replacing them was easy.

However, I noticed something concerning in the process: when I replaced log2Up in ShiftRegisterTester, I got a Verilator compile error.  I suspect this is a FIRRTL bug that @azidar's recent work will fix, and it's totally unrelated to this PR, so I worked around it in the mean time (by using (log2Ceil(x) max 1) instead of log2Ceil(x)).

### DIFF
--- a/src/main/scala/chisel3/compatibility.scala
+++ b/src/main/scala/chisel3/compatibility.scala
@@ -215,12 +215,24 @@ package object Chisel {     // scalastyle:ignore package.object.name
     val TesterDriver = chisel3.testers.TesterDriver
   }
 
-
-  val log2Up = chisel3.util.log2Up
   val log2Ceil = chisel3.util.log2Ceil
-  val log2Down = chisel3.util.log2Down
   val log2Floor = chisel3.util.log2Floor
   val isPow2 = chisel3.util.isPow2
+
+  /** Compute the log2 rounded up with min value of 1 */
+  object log2Up {
+    def apply(in: BigInt): Int = {
+      require(in >= 0)
+      1 max (in-1).bitLength
+    }
+    def apply(in: Int): Int = apply(BigInt(in))
+  }
+
+  /** Compute the log2 rounded down with min value of 1 */
+  object log2Down {
+    def apply(in: BigInt): Int = log2Up(in) - (if (isPow2(in)) 0 else 1)
+    def apply(in: Int): Int = apply(BigInt(in))
+  }
 
   val BitPat = chisel3.util.BitPat
   type BitPat = chisel3.util.BitPat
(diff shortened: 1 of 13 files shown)


edwardcwang:
@ducky64 We'll probably need to update the chisel3-cheatsheet if this gets merged, since log2Ceil isn't on there but log2Up and log2Down are.

ducky64:
No deprecation phase in the main chisel3._ package? Having deprecation is nice in that it doesn't immediately break everyone's code, and you can recommend a replacement there using the nag warning.

aswaterman:
I'm agnostic to that matter @ducky64 ; I'll leave that decision (and implementation of it) to you and @jackkoenig.

jackkoenig:
I changed it to deprecated and rebased on master (since these commits are are independent)

jackkoenig:
Good to merge upon tests passing @ducky64 ?

ducky64:
I assume no one will ever import both chisel3.util._ and Chisel._ simultaneously? (there are log2Up / log2Down objects in both of those, which may collide)

Otherwise looks fine.

edwardcwang:
I'm not sure about the chisel3.util._ and Chisel._ assumption - could see it happen if someone is confused and tries to combine snippets from both Chisel._ (e.g. older examples etc) and chisel3.util.

ducky64:
So new code shouldn't be using Chisel._, if they are then that's a major failure of documentation on our side. Also, Chisel._ imports most of utils, preserving behavior from before chisel was split into frontend and utils.

jackkoenig:
I just checked, you get the same error on master if you try importing `chisel3.util._` and `Chisel._` at the same time as you do on this branch. It seems that using `val blah = chisel3.object` counts as a collision just the same.

jackkoenig:
Given that there's no difference in the error, what do you think @ducky64 ?

ducky64:
Yeah, this is fine. I think it's a fundamental problem that we really can't solve.

Issue #522, opened 2017-02-22
Revisit Vec wiring once the directionality / vec / binding refactors happen.


ucbjrl:
Instances of:
```
Vec.fill(n)(Module(new FullAdder()).io)
```
have to be replaced (if possible) with:
```
Array.fill(n)(Module(new FullAdder()).io)
```
since Vec wiring is now mono-connect, due to lack of direction on cloned `Vec` elements.

See 375e2b6a0a456c55298d82837d28986de6211ebc

sdtwigg:
I strongly doubt the mutability of the Array is actually desired so should advocate using a more appropriate Scala data structure.

Since random-indexing is highly likely, amusingly Vector is probably the best choice. To avoid confusion of Vec vs. Vector can use IndexedSeq, which is basically equivalent.

jackkoenig:
Here's a branch with a test that now fails but used to succeed:

https://github.com/ucb-bar/chisel3/tree/vec-of-modules

ducky64:
Ok, traced through this.

The issue seems to be that Vecs create an internal Wire, but Wire bindings are undirectioned, so the BiConnect operation gets really confused.

PR #530, opened 2017-02-24
Revert "Change Vec creation to check if gen is lit (and hence needs t…

ucbjrl:
…o be declared)"

This reverts commit dc86e7e1734d6abacb739b488df1de231e6b41b2.
This may address #522 - using chiselCloneType (instead of cloneType) to preserve directionality.

### DIFF
--- a/src/test/scala/chiselTests/Vec.scala
+++ b/src/test/scala/chiselTests/Vec.scala
@@ -12,7 +12,7 @@ class LitTesterMod(vecSize: Int) extends Module {
   val io = IO(new Bundle {
     val out = Output(Vec(vecSize, UInt()))
   })
-  io.out := Vec(vecSize, 0.U)
+  io.out := Vec(Seq.fill(vecSize){0.U})
 }
 
 class RegTesterMod(vecSize: Int) extends Module {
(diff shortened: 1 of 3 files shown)


ducky64:
First problem is that the tests (compile!) aren't passing...

For Aggregate.scala, this doesn't look like a direct (line-by-line) revert, though I think it does match the old behavior.

Do we need to delete the tests, or can we preserve those?

ucbjrl:
I think the fix is simple. Hang on ...

donggyukim:
What's the priority of this PR? It should be merged ASAP for ucb-bar/riscv-boom#20.

jackkoenig:
This should be merged ASAP, It looks good to me, but I don't entirely understand the switch from using SourceInfoTransform to CompileOptionsTransform macro on indexWhere and lastIndexWhere.

Also the removal from the macro on `apply[T <: Data](n: Int, gen: T)`, I'm guessing because these create types so their source locators aren't that important?

@ducky64 can you review?

src/test/scala/chiselTests/Vec.scala
```diff
@@ -32,6 +32,15 @@ class IOTesterMod(vecSize: Int) extends Module {
   io.out := io.in
 }
 
+class OneBitUnitRegVec extends Module {
+  val io = IO(new Bundle {
+    val out = Output(UInt(1.W))
+  })
+  val oneBitUnitRegVec = Reg(Vec(1, 1.U))
```
ducky64: Should the second Vec argument be a UInt(1.W) instead of a literal?

ucbjrl: I think @ducky64  is correct. The second `Vec` argument should be a `UInt(1.W)`.

jackkoenig: Aren't we trying to check Vecs with lit gens here though?

PR #531, opened 2017-02-25
Digit record field

jackkoenig:
This incorporates the fixes from @terpstra and @mwachs5. I merged their commits because they kinda went together while I think the other 2 are separate. In particular, @ducky64, can you look at the first commit which changes and uses NamedModuleTester?

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -411,7 +411,7 @@ abstract class Record extends Aggregate {
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name)) }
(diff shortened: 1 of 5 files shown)


ducky64:
Only comment is that expectParentName sounds like a very roundabout way of getting the name of a Module, is there no way to specify expectModuleName?

jackkoenig:
Actually I think I can use just `Module.name`: https://github.com/ucb-bar/chisel3/blob/563f348ca3e4a984ad23caba9e980dab8fa808bb/chiselFrontend/src/main/scala/chisel3/core/Module.scala#L148

For some reason I thought final naming occurred in emission so this wouldn't be possible, but I think that was either just for blackboxes or was removed when we added `defname` to Firrtl `extmodules`

jackkoenig:
Done, better @ducky64? Using the parent name of the io was a bit wonky, I figured it was fine since it was just a test but yeah, this is much better.

jackkoenig:
I always hated how it was done, glad you figured out a better way in the NamingAnnotationTests.

jackkoenig:
I'm going to rebase to maintain some of the separate commits

PR #542, opened 2017-03-02
Fixes format strings in assertions. Fixes #540

### LINKED ISSUE #540: Add assertion example to Wiki
The Wiki has examples of how to print, but not write assertions. My assertions recently stopped compiling, perhaps because of #500 ? What is the right way to do something like:

`assert (a === b, "A (%x) is not equal to B (%x) !", a, b)`

If I do this I get errors like:

```
Caused by: java.lang.IllegalArgumentException: requirement failed: Too many arguments! More format specifier(s) expected!
    fmt = "A (%%x) is not equal to B(%%x)!"
...
```

jackkoenig:


mwachs5:
Yes this fixed my issue

chiselFrontend/src/main/scala/chisel3/core/Assert.scala
```diff
@@ -51,12 +51,15 @@ object assert { // scalastyle:ignore object.name
   }
 
   def apply_impl_do(cond: Bool, line: String, message: Option[String], data: Bits*)(implicit sourceInfo: SourceInfo) {
+    val escLine = line.replaceAll("%", "%%")
     when (!(cond || Builder.forcedReset)) {
       val fmt = message match {
-        case Some(str) => s"Assertion failed: $str\n    at $line\n"
-        case None => s"Assertion failed\n    at $line\n"
+        case Some(str) =>
+          val msg = if (data.isEmpty) str.replaceAll("%", "%%") else str
```
ducky64: Does this make sense to have it do different things based on whether data is passed in? Can we instead make it consistent between the two cases, but error out here with an informative message if the number of format arguments doesn't match the number of data passed in?

jackkoenig: Right so the basic problem is that people want to be able to write assertions with format strings and arguments and yet also with just normal strings. We **could** require that assert messages are format strings ([there's already a helper to do the error checking](https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Printable.scala#L37)). My main issue with this is that it strikes me as surprising behavior. Assertions with just simple messages are by far the most common case. However, I agree that it is undesirable to have the same function have entirely different functionality based on if there are any `Data` elements.

One idea is to add another `apply` method that accepts just `Bool` and `String` and another that accepts `Bool`, `String`, `Data`, and `Data*` to differentiate the two forms.

sdtwigg: The functions will still have the same name so the effect is the same. Having another function signature seems superfluous as you are just having the compiler do the dispatching for you.

In either case this will continue to work:
assert(???: Bool, "1 % 0 is undefined.")

jackkoenig: Right, so this PR is intended to make the following all work right:
* `assert(cond, "1 % 0 is undefined")`
* `assert(cond, "a %% b == %d", a % b)` (recall that the whole line is put in the assertion message)

sdtwigg: Oh, I know and agree. I am saying that the current code is fine and separating out a function for each case seems superfluous.

mwachs5:
As the user, if you tell me I need to use %%'s or \%s in one place and not in another, I don't care :) Just tell me the right way to do it.

mwachs5:
Can this be merged?

jackkoenig:
@ducky64 So what do you think? Think we can merge this?

ducky64:
I'm still not convinced that it's a great idea to have it automatically infer % = %% when there's no arguments - the pathological case is when you start with no arguments to print and have the single %, but then add an argument and the format string is interpreted differently. It's not a show-stopper, especially for experienced users, but it is a minor possibly annoying gotcha.

ducky64:
One thing we might consider is to keep consistent with how C/C++ and Verilog handles printing with no arguments - do they automatically allow this % shorthand special case or do they require %% uniformly?

aswaterman:
C/C++ printf is variadic, and the callee doesn't know how many arguments were passed.  It relies on the number of % format specifiers to know the argument count.  It is not tolerant to the % shorthand.  gcc warns about it:

    waterman@gamma00:~$ cat test.c
    #include <stdio.h>
    void foo() { printf("% "); }
    waterman@gamma00:~$ gcc test.c
    test.c: In function ‘foo’:
    test.c:2:21: warning: conversion lacks type at end of format [-Wformat=]
     void foo() { printf("% "); }

jackkoenig:
Alright so what do y'all think we should do?
`assert` accepts just a string (% shorthand)
`assertf` accepts printf like constructs? (Must use %% for percent)

Also how does `p"..."` play into this? Currently for `printf` you can either pass printf style variatic constructs or a Printable via `p"..."`

ucbjrl:
retest this please

ducky64:
I vote for assertion being like printfs. This can be elaboration-time checked (that all your %s are matched with print arguments) so it's unlikely to produce subtle bugs.

aswaterman:
I think the backwards incompatibility with asserts that formerly contained % is acceptable, since the error is noisy and easily fixed.

jackkoenig:
I've made format messages format strings, ie. % must be escaped as %%

jackkoenig:
I'm thinking the documentation shoudl just be in printf and then assert should refer to it, I'm fine with doing it as part of this PR, taking a look

mwachs5:
Whatever happened with this PR? I still get errors when I try to print variables inside my assert statements.

mwachs5:
For Future-me , note that you CAN currently write an assertion like:

```
assert (a > b, "a (" + a ") not > b (" + b + ")" )
```

jackkoenig:
This is actually a minor API change: `assert(myBool, "Message with %")` will fail, so I'm rescheduling this for 3.1

jackkoenig:
Alright, between https://github.com/freechipsproject/chisel3/pull/737 and this, I finally actually wrote the format string documentation. hooray...

PR #546, opened 2017-03-09
Fixed oneHotMux can't handle only one input

colin4124:
When there is only one select signal, the original one-hot Mux ignores the select signal and always return the input data.

I think the one-hot Mux in that case should return input data when select signal is set or return zero when select signal is clear.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
@@ -55,7 +55,9 @@ private[chisel3] object SeqUtils {
 
   def do_oneHotMux[T <: Data](in: Iterable[(Bool, T)])(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
     if (in.tail.isEmpty) {
-      in.head._2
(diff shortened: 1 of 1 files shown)


aswaterman:
The behavior is implementation defined when not exactly one input is hot. So a correct and minimal-hardware implementation is to ignore the select signal when there's only one input.

colin4124:
Okay, I got it. Thanks for your explaining.

shunshou:
@aswaterman I think this behavior should be better documented. I ran into the exact same problem. I kind of expected the Mux1H to be implemented something like x & selXFilled | y & selYFilled | z & selZFilled, etc. so, if the Seq had length one, I was still expecting x & selXFilled instead of x. I was originally getting the behavior (chisel2) with some less pretty foldLeft and thought I could replace it with prettier Mux1H... but guess not... :\

aswaterman:
I'm a bit surprised it's surprising, since the contract for one-hot inputs
is almost always that the behavior is undefined if the input is not exactly
one-hot. But I agree, the docs should be very clear about this.

Importantly, the docs should not conflate the implementation with the
interface. The contract should be that if one select signal is high, the
corresponding output is returned, else the result is
implementation-defined. This preserves the option to change the
implementation for target-specific optimizations.

PR #559, opened 2017-03-24
Fixed fix, allow Mux of different binary points and widths

### LINKED ISSUE #557: FixedPoint Mux (with different binary points)
That should be legal as long as the binary point values are "known", which is most of my use case. It's really annoying to write more complicated DSP b/c this isn't supported.

ducky64:
Resolves #557 

The implementation is ugly, but I can't think of a better way and this is a feature that makes sense. The larger problem seems to be that Chisel is duplicating width inferencing work with FIRRTL, but without all the information. The larger discussion might be whether that is a good practice, as removing width propagation in Chisel can remove a lot of complexity and special cases.


shunshou:
@ducky it removes a lot of complexity, but it's still extremely useful to do getWidth on signals whose width are known for certain functions.

Issue #558, opened 2017-03-24
Mux1H Bug (HELP!) OneHotMux broken when used with FixedPoint


shunshou:
Please fix ASAP! I *reallyreallyreally* would like this to work.

```
package dspblocks.fft
import chisel3._
import chisel3.experimental._
import dsptools.numbers._
import dsptools.numbers.implicits._
import org.scalatest.{FlatSpec, Matchers}
import dsptools.{DspTesterOptionsManager, DspTester}

class Borked[T <: Data:RealBits](dspDataType: => T) extends Module {
  val io = IO(new Bundle {
    val sels = Input(Vec(4, Bool()))
    val out = Output(DspComplex(dspDataType))
  })

  val one = dspDataType.fromDoubleWithFixedWidth(1.0)
  val C52 = dspDataType.fromDoubleWithFixedWidth(-1.5388417840003967)
  val C75 = dspDataType.fromDoubleWithFixedWidth(-0.34087294340133667)
  val A5 = Mux1H(Seq(
    io.sels(0) -> one,
    io.sels(1) -> C52,
    io.sels(2) -> C75,
    io.sels(3) -> one
  ))
  io.out.real := A5
}

class BorkedSpec extends FlatSpec with Matchers {
  behavior of "Borked"
  it should "not suck" in {
    val opt = new DspTesterOptionsManager 
    dsptools.Driver.execute(() => new Borked(dspDataType = FixedPoint(28.W, 24.BP)), opt) { c =>
      new BorkedTester(c)
    } should be (true)
  }
}

class BorkedTester[T <: Data:RealBits](c: Borked[T]) extends DspTester(c) {
  poke(c.io.sels(0), false.B)
  poke(c.io.sels(1), false.B)
  poke(c.io.sels(3), false.B)
  poke(c.io.sels(2), true.B)
  peek(c.io.out.real)
}
```

I spent a really long time figuring out why my circuit was working with DspReal but not with FixedPoint (and running into annoying issues like Mux not supporting different BP's, etc.), and I just *can't* afford to be spending this time debugging stuff on the Chisel side. 

---

This works:

```
package dspblocks.fft
import chisel3._
import chisel3.experimental._
import dsptools.numbers._
import dsptools.numbers.implicits._
import org.scalatest.{FlatSpec, Matchers}
import dsptools.{DspTesterOptionsManager, DspTester}

class Borked[T <: Data:RealBits](dspDataType: => T) extends Module {
  val io = IO(new Bundle {
[...]

PR #560, opened 2017-03-24
Andrew's fix for one-hot mux with differing fixed point targets

chick:
This should fix @shunshou problem described in #558

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
@@ -53,13 +53,15 @@ private[chisel3] object SeqUtils {
     */
   def oneHotMux[T <: Data](in: Iterable[(Bool, T)]): T = macro CompileOptionsTransform.inArg
 
+  //scalastyle:off method.name
   def do_oneHotMux[T <: Data](in: Iterable[(Bool, T)])(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
     if (in.tail.isEmpty) {
       in.head._2
     } else {
-      val masked = for ((s, i) <- in) yield Mux(s, i.asUInt, 0.U)
       val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
+      val masked = for ((s, i) <- in) yield Mux(s, Wire(t = output, init = i).asUInt, 0.U)
       output.fromBits(masked.reduceLeft(_|_))
     }
   }
+
 }


shunshou:
FYI, I have the workaround of using Wires everywhere, so I'm ok holding off on this merge until Andrew's review gets addressed. I'm moving on to control logic, but will be getting back to DSP data path things for other parts of my design by ~Monday, which is when I'd like something working.

ducky64:
This smells like a band-aid fix and might be masking a deeper problem. Is there a reason that we haven't seem this bug with UInt before? Does a similar structure (but with UInt rather than FixedPoint) produce the same bug? If not, then this might be a FixedPoint-specific error.

I'd like to see a minimal test case that produces this bug, so we can treat the root cause instead of the symptoms.

aswaterman:
@ducky64 I believe it may be some of both.  While Mux1H is correct for UInt inputs, by inspection it looks wrong for SInts of different widths (they'll be zero-extended).

ducky64:
Ok, I see.

In this case, you're not using Wire because you want a Wire, you're using Wire because you want a value-preserving cast. It might make sense to have an explicit cast operator (much like `.asTypeOf`), though that would only make sense in special cases (like SInt/UInt). By inspection, it also seems like it would choke on SInts of different widths, where at least one (and hence the output) is of unknown width.

shunshou:
Yeah that should be the case. I'm sure there are probably other utilities that have the same problem, but the use case with Signed anythings has never really been exercised b/c rocket-chip is all unsigned stuff.

shunshou:
Pinging people -- any progress? @chick and whoever else...

sdtwigg:
Since the FIRRTL is hard to read for this:
What does asUInt on a FixedPoint do?
What does FixedPoint.fromBits(UInt) do?
  The original intent was that asUInt/fromBits were blind bit-reinterpretations of their values. (Note the old Mux1H also isn't quite right with SInt since will convert zero-extending to sign-extending.)

I think this form does a better job of matching everything up by first ensuring everything is consistently sized and then doing cleverness. I am exploiting the fact that Vec will be ensure all connections to it are maximally sized to all then bit-extend properly into it. I'm not sold the proposed solution makes sure things work when a binary point or width is unknown.
```scala
val sameSize = Wire(Vec(cloneSupertype(...), in.length))
for(((_, in_e), same_e) <- in zip sameSize) { in := same_e }
val masked = for (((s, _), same_e) <- sameSize) yield Mux(s, same_e.asUInt, 0.U)
output.fromBits(masked.reduce(_|_))
```

Right now is similar. However, eventually, if you could defer the fromBits to post-firrtl-inference could have:
```scala
// sameSize is totally 'uninferred' here.
val sameSize = Wire(Vec(cloneSupertype(...).wipeAllParams, in.length))
for(((_, in_e), same_e) <- in zip sameSize) { in := same_e }
val masked = for (((s, _), same_e) <- sameSize) yield Mux(s, same_e.asUInt, 0.U)
// head is fine since all elements of sameSize are identical
sameSize.head.fromBits(masked.reduce(_|_))
```

This would solve a latent issue where cloneSupertype doesn't work for semi-homogenous aggregate types (Same type-'structure' but different widths, bps, etc.). Also, as a bonus, far fewer wires (not including firrtl vec exploding....)

PS: I def agree with Andrew that these forms should only be used when the scenario warrants (sizing is different or unknown.)

chick:
@shunshou I'll give Stephen's suggestion a try

chick:
@shunshou @aswaterman @sdtwigg I believe the issue that triggered this PR was ultimately due to width information being lost in fixed point literal creation. I have addressed that in new PR #564 . I am closing this PR, but wondered if the discussion here on OneHotMux should be moved into its own issue.

sdtwigg:
I think so because cloneSupertype still isn't quite right at least for
aggregates and thus I am not confident there won't be problems in the future

shunshou:
@chick fundamentally, the fixed point lit width issue is separate from the fact that Mux1H doesn't properly sign extend. Both should be addressed, and it'd be preferable to not have to specify FixedPoint Lit and still get the right result.

PR #562, opened 2017-03-27
Support Vec(0) fields in Bundles, just like Option[Data]; add test

aswaterman:
This also allows asUInt/asTypeOf to work properly on those Bundles,
even though zero-width wire support is lacking.


hcook:
This fixes my issue #562

PR #564, opened 2017-03-28
Creating FixedPoint literals was throwing away width

chick:
Creating FixedPoint literals was throwing away width even when specifically provided.
This caused one hot muxing problems in dsptools
FixedPoint spec fixed based on error uncovered by this change


shunshou:
Can someone say that this is ok and merge? Lgtm.

shunshou:
@grebe can you please update your rocket-chip chisel3 to this version?

PR #566, opened 2017-03-29
Add annotation dumping to dumpFirrtl (used by rocket)

grebe:
Rocket's `generateFirrtl` calls `dumpFirrtl`, which doesn't dump annotations. This PR adds a new function that both `dumpFirrtl` and `execute` call so they will both dump annotations in more or less the same way.

For now we are overriding `generateFirrtl` to also spit out annotations, but it would be nice if `dumpFirrtl` did it by default.

### DIFF
--- a/src/main/scala/chisel3/Driver.scala
+++ b/src/main/scala/chisel3/Driver.scala
@@ -12,6 +12,7 @@ import firrtl._
 import firrtl.util.{ BackendCompilationUtilities => FirrtlBackendCompilationUtilities }
 
 import _root_.firrtl.annotations.AnnotationYamlProtocol._
+import _root_.firrtl.annotations.Annotation
 
 /**
   * The Driver provides methods to invoke the chisel3 compiler and the firrtl compiler.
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/Driver.scala
```diff
@@ -94,12 +95,25 @@ object Driver extends BackendCompilationUtilities {
 
   def emit[T <: Module](ir: Circuit): String = Emitter.emit(ir)
 
-  def dumpFirrtl(ir: Circuit, optName: Option[File]): File = {
-    val f = optName.getOrElse(new File(ir.name + ".fir"))
-    val w = new FileWriter(f)
-    w.write(Emitter.emit(ir))
+  def dumpFirrtl(ir: Circuit, optName: Option[File], optAnnoName: Option[File] = None): File = {
```
grebe: Is giving `optAnnoName` a default value too much of a dirty trick? Rocket calls a two argument version of `dumpFirrtl`

colinschmidt:
Hmm I solved this by extending Generator with a dumpAnnotations method in addition to the dumpFirrtl. I understand the desire to always spit out annotations as they become more necessary for useful operation of chisel, but I don't know if this is the right way to solve this problem.

jackkoenig:
I remember we discussed this but I don't remember what the conclusion was....

edwardcwang:
I recall us discussing adding a separate function like dumpAnnotations instead of using an optional argument to dumpFirrtl or something.

edwardcwang:
I recall us discussing adding a separate function like dumpAnnotations instead of using an optional argument to dumpFirrtl or something.

chick:
@grebe should this be acted upon or has something else taken care of it.

grebe:
dumpAnnotations is fine for rocket, I'll close this.

PR #568, opened 2017-03-31
Make Module instantiations draw clock from Builder instead of parent

### LINKED ISSUE #567: withClock and friends doesn't apply to child modules correctly
In a module I essentially have the following

```
withClockAndReset(clock = newClock, reset = newReset) {
  val mod = Module(new ModuleOfSomeVariety)
  ...
}
```

and the module doesn't get newClock or newReset. According to @ducky64, setupInParent gets the parents clock whereas withClockAndReset only modifies the Builder's clock.

jackkoenig:
Fixes #567


aswaterman:
I just stumbled into this bug :-(

aswaterman:
:(

PR #570, opened 2017-04-03
Use input element to decide if Vec of values has direction

### LINKED ISSUE #522: Revisit Vec wiring once the directionality / vec / binding refactors happen.

### LINKED ISSUE #569: Working code fails tests
I had a working code written a few weeks ago which uses Vec.fill as follows:

`val multipliers = Vec.fill(n)(Module(new FPMul()).io)`

The same code doesn't work today neither using FIRRTL nor Verilator. 

Do I need to change something to make the code work again?

Update:
I just looked at the Verilog code and it seems to be okay. So I guess Vec.fill() is doing what it is supposed to do. I don't know where the problem is.

Thanks,
Matt

jackkoenig:
Using the sample_element of the created wire is incorrect because Wires have no
direction so the Wire constructed for a Vec of Module IO was constructed
incorrectly. Fixes #569 and resolves #522.

Once I figured out that this was the problem, I was confused how you could index into a wire that was of type Vec of Module io and still correctly bulk connect to the directioned inputs. It turns out that this works by using the direction of the other side of the bulk connect as a hint. This has the following two nasty consequences that should be fixed in the Binding refactor:
1. You cannot bulk connect the Bundle ports of two modules that are elements of Vecs.
1. If you try to bulk connect some Bundle port of a module (not in a Vec) to the Bundle port of another module (**is** in a Vec), and the Bundles have the same field names but mismatched directions, this will lead to an silently incorrect successful connection. It doesn't even error in Firrtl.

(1) errors with a reasonable error message in Chisel, it just says it cant figure out the directions.
(2) is the scary one, I'll file a separate issue

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -73,7 +73,9 @@ object Vec {
 
     def doConnect(sink: T, source: T) = {
       // TODO: this looks bad, and should feel bad. Replace with a better abstraction.
-      val hasDirectioned = vec.sample_element match {
+      // NOTE: Must use elts.head instead of vec.sample_element because vec.sample_element has
+      //       WireBinding which does not have a direction
+      val hasDirectioned = elts.head match {
         case t: Aggregate => t.flatten.exists(_.dir != Direction.Unspecified)
         case t: Element => t.dir != Direction.Unspecified
       }
--- a/src/test/scala/chiselTests/Vec.scala
+++ b/src/test/scala/chiselTests/Vec.scala
@@ -147,6 +147,42 @@ class ZeroEntryVecTester extends BasicTester {
   stop()
 }
 
+class PassthroughModuleIO extends Bundle {
+  val in = Input(UInt(32.W))
+  val out = Output(UInt(32.W))
+}
+
+class PassthroughModule extends Module {
+  val io = IO(new PassthroughModuleIO)
+  io.out := io.in
+}
+
+class PassthroughModuleTester extends Module {
+  val io = IO(Flipped(new PassthroughModuleIO))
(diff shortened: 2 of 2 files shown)


chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -73,7 +73,9 @@ object Vec {
 
     def doConnect(sink: T, source: T) = {
       // TODO: this looks bad, and should feel bad. Replace with a better abstraction.
-      val hasDirectioned = vec.sample_element match {
+      // NOTE: Must use elts.head instead of vec.sample_element because vec.sample_element has
+      //       WireBinding which does not have a direction
+      val hasDirectioned = elts.head match {
```
sdtwigg: This won't work if elts is empty. Also, in the Vec.apply case, aren't the elts also bound to Wire?

jackkoenig: This method already requires that elts are nonEmpty.

It's not really clear what Vec(Seq.empty) should mean although I do think constructing a Vec[Nothing] such that so long as the user does not try to index the Vec, it's fine for generators to do `Vec.fill(0){...}`

As for elts already being bound to a Wire, I guess this does mean you couldn't make multi-dimensional Vecs of Module IO. I think we should continue discussion #571 for a complete solution. In the meantime, I believe that this PR fixes a really big problem even if it doesn't completely solve the fundamental issue.

PR #573, opened 2017-04-03
Fix one hot mux

chick:
Answers some of Angie's problems for #558 Mux1H
- Implements FixedPoint as MuxTree, this is not optimized
- catches some cases of inferred with aggregates not working
- adds tests that illustrate a number of situations

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
@@ -6,6 +6,8 @@ import scala.language.experimental.macros
 
 import chisel3.internal.sourceinfo._
 
+//scalastyle:off method.name
+
 private[chisel3] object SeqUtils {
   /** Concatenates the data elements of the input sequence, in sequence order, together.
     * The first element of the sequence forms the least significant bits, while the last element
@@ -39,7 +41,8 @@ private[chisel3] object SeqUtils {
     */
   def priorityMux[T <: Data](in: Seq[(Bool, T)]): T = macro CompileOptionsTransform.inArg
 
-  def do_priorityMux[T <: Data](in: Seq[(Bool, T)])(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+  def do_priorityMux[T <: Data](in: Seq[(Bool, T)])
+                               (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
     if (in.size == 1) {
       in.head._2
     } else {
@@ -48,21 +51,65 @@ private[chisel3] object SeqUtils {
   }
 
   /** Returns the data value corresponding to the lone true predicate.
+    * This is elaborated to firrtl using a structure that should be optimized into and and/or tree.
     *
     * @note assumes exactly one true predicate, results undefined otherwise
+    *       FixedPoint values or aggregates containing FixedPoint values cause this optimized structure to be lost
     */
   def oneHotMux[T <: Data](in: Iterable[(Bool, T)]): T = macro CompileOptionsTransform.inArg
 
-  def do_oneHotMux[T <: Data](in: Iterable[(Bool, T)])(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+  //scalastyle:off method.length cyclomatic.complexity
+  def do_oneHotMux[T <: Data](in: Iterable[(Bool, T)])
+                             (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
     if (in.tail.isEmpty) {
       in.head._2
-    } else {
+    }
+    else {
+      val compares = in.map { case (_, element) => in.head._2.typeEquivalent(element) }
+      val allSame = in.forall { case (_, element) => in.head._2.typeEquivalent(element) }
+      if(allSame) {
+        val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
+        val masked = for ((s, i) <- in) yield Mux(s, i.asUInt(), 0.U)
+        output.fromBits(masked.reduceLeft(_|_))
+      }
       in.head._2 match {
-        case _: FixedPoint =>
+        case _: SInt =>
+          // SInt's have to be managed carefully so sign extension works
           val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
-          val masked = for ((s, i) <- in)
-            yield Mux(
-              s, i.asTypeOf(output).asInstanceOf[FixedPoint], 0.U.asTypeOf(output).asInstanceOf[FixedPoint])
+
+          val sInts: Iterable[(Bool, SInt)] = in.collect { case (s: Bool, f: SInt) =>
+            (s, f.asTypeOf(output).asInstanceOf[SInt])
+          }
+
+          val masked = for ((s, i) <- sInts) yield Mux(s, i, 0.S)
+
+          output.fromBits(masked.reduceLeft(_|_))
+
+        case _: FixedPoint =>
+          // Fixed point cannot use the standard methodology here
+          val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
+
+          //          val masked = in.collect { case (s: Bool, f: FixedPoint) => (s, f) }.map { case (s, f) =>
+          //            Mux(
+          //              s, f, 0.U.asTypeOf(output).asInstanceOf[FixedPoint])
+          //          }
+
+          val fixedPoints = in.collect { case (s: Bool, f: FixedPoint) => (s, f) }
+          def makeMux(inputs: List[(Bool, FixedPoint)]): FixedPoint = {
+            inputs match {
+              case Nil => 0.U.asTypeOf(output).asInstanceOf[FixedPoint]
+              case (select, value) :: tail  =>
+                Mux(select, value, makeMux(tail))
+            }
+          }
+
+          makeMux(fixedPoints.toList).asTypeOf(output)
+        //          val masked = for ((s, i) <- in)
+        //            yield Mux(
+        //              s, i.asTypeOf(output).asInstanceOf[FixedPoint], 0.U.asTypeOf(output).asInstanceOf[FixedPoint])
+        case aggregate: Aggregate =>
+          val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
+          val masked = for ((s, i) <- in) yield Mux(s, i.asUInt(), 0.U)
           output.fromBits(masked.reduceLeft(_|_))
         case _ =>
           val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
--- a/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
@@ -57,9 +57,18 @@ private[chisel3] object SeqUtils {
     if (in.tail.isEmpty) {
       in.head._2
     } else {
-      val masked = for ((s, i) <- in) yield Mux(s, i.asUInt, 0.U)
-      val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
-      output.fromBits(masked.reduceLeft(_|_))
+      in.head._2 match {
+        case _: FixedPoint =>
+          val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
+          val masked = for ((s, i) <- in)
+            yield Mux(
+              s, i.asTypeOf(output).asInstanceOf[FixedPoint], 0.U.asTypeOf(output).asInstanceOf[FixedPoint])
+          output.fromBits(masked.reduceLeft(_|_))
+        case _ =>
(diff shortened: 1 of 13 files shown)


chick:
Some further discussion may be necessary here, but this can serve as straw man.  
This implements Mux1H for Fixed as a mux tree because fixedPoint does not support or operator and cast to UInt loses width and binary point info
Special cases SInt so that sign extension on inferred widths works, optimized still in this case
Throws exceptions for  a couple of situations with aggregates containing inferred widths or unmatched types, not sure what is the right behavior here.

shunshou:
My current fix uses the default Mux1H while special casing out FixedPoint. When FixedPoint inputs are matched, I check to see their widths and binary points are known. Then I take the max of the two and construct wires so and feed the inputs to them so that their widths are matched. It's ok if I then feed to normal Mux1H bc of their fromBits option.

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,68 @@ private[chisel3] object SeqUtils {
...
       in.head._2
-    } else {
-      val masked = for ((s, i) <- in) yield Mux(s, i.asUInt, 0.U)
-      val output = cloneSupertype(in.toSeq map {_._2}, "oneHotMux")
-      output.fromBits(masked.reduceLeft(_|_))
+    }
+    else {
+      val output = cloneSupertype(in.toSeq map { _._2}, "oneHotMux")
+
+      in.head._2 match {
```
ducky64: Does it make more sense to mux on `output`?

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,68 @@ private[chisel3] object SeqUtils {
...
+            (s, f.asTypeOf(output).asInstanceOf[SInt])
+          }
+
+          val masked = for ((s, i) <- sInts) yield Mux(s, i, 0.S)
+
+          output.fromBits(masked.reduceLeft(_ | _))
+
+        case _: FixedPoint =>
+          // FixedPoint doesn't support *or* so use Muxes
+          val output = cloneSupertype(in.toSeq map { _._2 }, "oneHotMux")
```
ducky64: Why another call to cloneSupertype, this looks like it does the exact same thing as above?

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,68 @@ private[chisel3] object SeqUtils {
...
+    }
+    else {
+      val output = cloneSupertype(in.toSeq map { _._2}, "oneHotMux")
+
+      in.head._2 match {
+        case _: SInt =>
+          // SInt's have to be managed carefully so sign extension works
+
+          val sInts: Iterable[(Bool, SInt)] = in.collect { case (s: Bool, f: SInt) =>
+            (s, f.asTypeOf(output).asInstanceOf[SInt])
```
ducky64: Are you sure `asTypeOf` still needs the `asInstanceOf`?

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,68 @@ private[chisel3] object SeqUtils {
...
+          }
+
+          val masked = for ((s, i) <- sInts) yield Mux(s, i, 0.S)
+
+          output.fromBits(masked.reduceLeft(_ | _))
+
+        case _: FixedPoint =>
+          // FixedPoint doesn't support *or* so use Muxes
```
ducky64: You can do a FP -> FP (width / precision expanded) -> bits -> FP conversion? Otherwise you're just generating a mux chain (not even a tree!), and this function really doesn't give you anything special.

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,68 @@ private[chisel3] object SeqUtils {
...
+              case Nil => 0.U.asTypeOf(output).asInstanceOf[FixedPoint]
+              case (select, value) :: tail =>
+                Mux(select, value, makeMux(tail))
+            }
+          }
+
+          makeMux(fixedPoints.toList).asTypeOf(output)
+
+        case aggregate: Aggregate =>
+          val allSameType = in.forall { case (_, element) => in.head._2.typeEquivalent(element) }
```
ducky64: isn't this already checked by cloneSupertype?

src/test/scala/chiselTests/OneHotMuxSpec.scala
```diff
@@ -0,0 +1,345 @@
...
+  ))
+}
+class SIntOneHotTester extends BasicTester {
+  val dut = Module(new SIntOneHot)
+  dut.io.selectors(0) := false.B
+  dut.io.selectors(1) := true.B
+  dut.io.selectors(2) := false.B
+  dut.io.selectors(3) := false.B
+
+  printf("out is %d\n", dut.io.out)
```
ducky64: please no printfs in test, no one will actually look at the console output, and most likely it will not make sense with the other 100+ tests in parallel

src/test/scala/chiselTests/OneHotMuxSpec.scala
```diff
@@ -0,0 +1,345 @@
...
+    val selectors = Input(Vec(4, Bool()))
+    val out = Output(SInt(8.W))
+  })
+
+  io.out := Mux1H(Seq(
+    io.selectors(0) -> (-3).S,
+    io.selectors(1) -> (-5).S,
+    io.selectors(2) -> (-7).S,
+    io.selectors(3) -> (-11).S
+//      io.selectors(0) -> (-1).S,
```
ducky64: eliminate dead code?

src/test/scala/chiselTests/OneHotMuxSpec.scala
```diff
@@ -0,0 +1,345 @@
...
+  })
+
+  io.out := Mux1H(Seq(
+    io.selectors(0) -> 2.U,
+    io.selectors(1) -> 4.U,
+    io.selectors(2) -> 8.U,
+    io.selectors(3) -> 11.U
+  ))
+}
+class SIntOneHotTester extends BasicTester {
```
ducky64: Do you really need the Tester/ModuleBeingTested abstraction? Can you not just drop everything in the tester and save yourself what appears to be an unnecessary Module boundary?

shunshou:
I still say there's some utility to being able to query known propagated
widths in Chisel.

It's likely that the tools are smart enough to know that you have a bunch
of or's rather than full-fledged muxes, but again, this should be something
to be verified. As the hardware designer, I need to reemphasize for the nth
time that I really really really care about good QOR and would rather write
100 lines over 10 lines if it means I can get a post CAD tool design that's
2x smaller. I.e. If I don't care about priority, don't generate something
that prioritizes inputs.

ducky64:
Actually, it might not be an either-or situation (wrt querying propagated widths and delegating to firrtl). Getting delegated widths is type-level information, which (at least in the examples here) is easily handled with Chisel facilities. The problem where this screwed up is generating the circuit, and it's basically impossible to get const-propagated circuit values in Chisel anyways.

shunshou:
Ok I buy that.

chick:
@shunshou Where is the current fix you referenced above.  I agree that the fixed point should work that way when widths and binary points are known

shunshou:
https://github.com/ucb-art/FFTGen/blob/chisel3_transition/src/main/scala/Chisel3/Mux1H_Hack.scala

There are a couple of things going on in there:

1) Splitting Complex into real and imaginary parts before doing Mux1H. This is necessary due to a weird Verilator bug.
2) Handling when there's only 1 element. Mux1H returns the element itself. I still want it to return 0 if the condition is false.
3) A hack-ish patch to special case out FixedPoint.

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
chick: Yes, this call was superfluous.

src/test/scala/chiselTests/OneHotMuxSpec.scala
chick: Fixed

src/test/scala/chiselTests/OneHotMuxSpec.scala
chick: Done

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
chick: Yes, removed

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
chick: Yes, done

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
chick: Yes needed to get Or, or at least I couldn't figure out how to get it to work without it

ducky64: "get Or"?

chick: I needed the *or* method to be defined on the elements of the sInts, asTypeOf(output) gave me a T which is subtype of data and does not have *or*

ducky64: Ah, I see, because it returns the type of `output`, which isn't what you matched on and the typechecker does't know any more than it is type T <: Data. What if you instead did the match on `output` instead of `in.head._2`?

chick: I did change it to match on output but that did not help

sdtwigg: So, the issue here is that the match established that T <: SInt but not that T is SInt. If the user subtyped SInt (for whatever reason) and then passed the subtype in as an input, then T would be that subtype and this type case should except (since SInt isn't a subtype of that type). You can try making SInt final (is sealed but sealed can have subtypes) but I don't think scalac will take that into account when doing the analysis.

Consider how the similar scenario: def makeUnknown[T <: UInt](x: T): T = UInt()
doesn't work for makeUnknown(Bool())

chick:
@ducky64 ping

chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala
```diff
@@ -48,18 +52,70 @@ private[chisel3] object SeqUtils {
...
+          val (intWidths, binaryPoints) = in.toSeq.map { case (_, o) =>
+            val fo = o.asInstanceOf[FixedPoint]
+            require(fo.binaryPoint.known, "Mux1H requires width/binary points to be defined")
+            (fo.getWidth - fo.binaryPoint.get, fo.binaryPoint.get)
+          }.unzip
+
+          if (intWidths.distinct.length == 1 && binaryPoints.distinct.length == 1) {
+            buildAndOrMultiplexor(in)
+          }
+          else {
```
ducky64: What kind of style is
```scala
}
else {
```
?

chick: I am not having much luck finding any guidance on where to put ```else```, but I have found on the order of 200 such usages throughout the chisel family.

ducky64: I think the usual style is
```
} else {
```
or more rarely (and I don't think often used, if at all, in the chisel3 codebase),
```
}
else
{
```

chick: 368  occurrences
```
} else {
```
235 occurrences 
```
}
else {
```
50 occurrences 
```
}
else
{
```

shunshou: I hope the 235 occurrences of 
```
} 
else {
```

weren't all from me b/c that's what I've been using. ~_~

ducky64: The Spark style guide (was was a while ago suggested for Chisel) examples have `} else {` (same line), though don't seem to have a formal rule: https://github.com/databricks/scala-style-guide
Google's C++ style guide is similar (lists the example without a formal rule): https://google.github.io/styleguide/cppguide.html
Google's Java style guide makes it a formal rule: https://google.github.io/styleguide/javaguide.html
Scala's style guide has neither an example nor a rule: http://docs.scala-lang.org/style/

chick: @shunshou I tend to use ```else {``` also, so it's not all you.  I have to say I don't have strong feelings about this one, except I don't like else by itself.  ```else {``` seems to have a little more symmetry with ```if ... {``` but that's the only argument.  I'll try the other. I don't know if this can be caught by the style checker

PR #574, opened 2017-04-07
Change Enum to emit minimum widths of 1

### LINKED ISSUE #554: 1-entry enum compile error
To keep track of the conversation on d4ecf002be3490e27a1cd50bd00a12837beaffb1

jackkoenig:
Fixes #554 

Tried to changed ULit to allow zero-width wires but this seems to have exposed some bugs in Firrtl so I think a workaround is fine.


donggyukim:
This works for me. Thanks!

PR #575, opened 2017-04-08
Fix macOS install instructions for homebrew package manager

yep:
On macOS, running `brew cask install sbt` from the install instructions results in:

   Error: No available Cask for sbt

Looks like `sbt` moved from cask to the official homebrew repository. Correct command seems to be `brew install sbt`

Check with official [sbt documentation](http://www.scala-sbt.org/download.html)


ucbjrl:
ok to test

PR #590, opened 2017-04-23
Fixed point width inference was wrong when binary points didn't align.

shunshou:
@chick I'm not sure how this escaped some of the mixed width/binary point tests... Basically, if I was subtracting a <<9>><<0>> number from a <<22>><<21>> number, the immediate output was right... but when the output was passed through subsequent ops, getWidth was wrong (therefore cloneType was wrong... and everything after that was wrong).


shunshou:
@ucbjrl can I merge this now?

ucbjrl:
Be my guest.

PR #592, opened 2017-04-25
Dropimportnotstrict492 - More updates to get things through rocket-chip.

ucbjrl:
Add missing implicit compileOptions to do_pad() and do_zext().
Add SourceInfoTransform macros to Vec methods in order to avoid apply chain issues.
Since utils methods are no longer NotStrict, Pipe objects need access to the client's compile options. There may be more.


src/main/scala/chisel3/util/Valid.scala
```diff
@@ -34,7 +36,7 @@ object Valid {
 object Pipe
 {
   @chiselName
-  def apply[T <: Data](enqValid: Bool, enqBits: T, latency: Int): Valid[T] = {
+  def apply[T <: Data](enqValid: Bool, enqBits: T, latency: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Valid[T] = {
```
ducky64: Source info shouldn't be propagated through helper functions, only chisel internals.
Source info support for libraries is something that needs more thought and discussion.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -243,11 +246,11 @@ sealed class Vec[T <: Data] private (gen: => T, val length: Int)
   def apply(idx: Int): T = self(idx)
 
   @deprecated("Use Vec.apply instead", "chisel3")
-  def read(idx: UInt)(implicit compileOptions: CompileOptions): T = apply(idx)
+  def read(idx: UInt)(implicit compileOptions: CompileOptions): T = do_apply(idx)(DeprecatedSourceInfo, compileOptions)
 
   @deprecated("Use Vec.apply instead", "chisel3")
   def write(idx: UInt, data: T)(implicit compileOptions: CompileOptions): Unit = {
-    apply(idx)(compileOptions).:=(data)(DeprecatedSourceInfo, compileOptions)
+    do_apply(idx)(DeprecatedSourceInfo, compileOptions).:=(data)(DeprecatedSourceInfo, compileOptions)
```
ducky64: Can we just not thread compile options through deprecated functions?
These should be removed eventually anyways.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -223,7 +223,10 @@ sealed class Vec[T <: Data] private (gen: => T, val length: Int)
 
   /** Creates a dynamically indexed read or write accessor into the array.
     */
-  def apply(idx: UInt)(implicit compileOptions: CompileOptions): T = {
+  override def apply(x: UInt): T = macro SourceInfoTransform.xArg
+
+  def do_apply(x: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+    val idx = x
```
ducky64: You could also just change all instances of `idx` to `x`.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ucbjrl: Thanks @ducky64. I hope d9eff59 addresses this.

PR #595, opened 2017-04-26
Add implicit CompileOptions to Record and Bundle. Fixes #495

### LINKED ISSUE #495: Problem with chisel3 bulk connect of compatibility layer defined bundles
Consider the following two Chisel files:

```scala
import Chisel._

  val a = UInt(width = 32)
}

class ChiselModuleA extends Module {
  val io = new Bundle {
    val foo = new MyBundle
  }
}
class ChiselModuleB extends Module {
  val io = new Bundle {
    val foo = (new MyBundle).flip
  }
}
```

```scala
import chisel3._

class InterfaceTop extends Module {
  val io = IO(new Bundle {})
  val a = Module(new ChiselModuleA)
  val b = Module(new ChiselModuleB)
  b.io.foo <> a.io.foo
}
```

Building `InterfaceTop` fails with the error:
`chisel3.internal.ChiselException: Connection between left (MyBundle@16) and source (MyBundle@c) failed @.a: Left or Right unavailable to current module.`

It's actually not that `.a` is unavailable, it's actually because it isn't defined with an explicit direction. This is perfectly legal in compatibility mode, the default is Output. But if we try to connect the Bundle in `chisel3._` code, it doesn't work.

Related to issues in #494

jackkoenig:
Helps distinguish between Records/Bundles defined in Chisel._ vs.
chisel3._. Also override compilationOptions when bulk connecting
Records/Bundles defined in Chisel._. This allows Records/Bundles defined
in Chisel._ code to be correctly bulk connected in chisel3._ code.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -362,7 +362,7 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId {
   * Record should only be extended by libraries and fairly sophisticated generators.
   * RTL writers should use [[Bundle]].  See [[Record#elements]] for an example.
   */
-abstract class Record extends Aggregate {
+abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
   /** The collection of [[Data]]
     *
@@ -464,7 +464,7 @@ abstract class Record extends Aggregate {
   *   }
   * }}}
   */
-class Bundle extends Record {
+class Bundle(implicit compileOptions: CompileOptions) extends Record {
   override def className = "Bundle"
 
   /** The collection of [[Data]]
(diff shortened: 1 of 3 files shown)


jackkoenig:
Which implicit compile option is used when a Record method that takes a
compile option is called?

ducky64:
According to https://stackoverflow.com/questions/5598085/where-does-scala-look-for-implicits, first looks in current scope. I think if there are multiple options in any of those priorities, the compiler will error out with an ambiguous implicit. I guess it's helpful that Record methods don't have implicit arguments, though superclass methods do.

jackkoenig:
It seems that this isn't exactly doing what I wanted, so I'm taking another look.

jackkoenig:
As discussed last week, not going to let the perfect be the enemy of the good. I'm going to merge this when updating passes Jenkins.

This currently issues a Firrtl bulk connect whenever a Chisel._ Record/Bundle is used in chisel3._ code. The better proposed solution is to still do the standard chisel3._ bulk connect walk through leaf elements, but be a bit more permissive for Chisel._ bundles (eg. no explicit direction is Output).

jackkoenig:
There appears to be a compiler issue with rocket-chip, debugging.

jackkoenig:
False alarm, we good

Issue #529, opened 2017-02-24
Can no longer create Vecs of size 0 with a literal type


jchang0:
I think this is due to due to commit #480: Added vec IO tests for #104 

where the construction method for Vec(Seq) construction requires Seq to be not empty.

Actually, it would be useful for Vecs to be able to create an empty vec even with the Vec(Seq) construction.  Useful when Seq is constructed from parameter, and a 0 length sequence is a possible parameterization, and don't want to write a special case handler.

jackkoenig:
The real question is what is the type of an empty Vec? In Scala, all empty Seqs are equal despite having different types. 
eg.

```
scala> val x = Seq.empty[Int]
x: Seq[Int] = List()

scala> val y = Seq.empty[String]
y: Seq[String] = List()

scala> x == y
```

So what should we do in Chisel?

aswaterman:
I agree this is a bug.  They're equal in Scala because the type doesn't matter in that case.  Seems like the same should be true of Vec.

As for the implementation, it should be made possible to pass ```gen = null.asInstanceOf[T]``` when size==0.

PR #614, opened 2017-05-23
Add support for empty Vecs - addresses #529

ucbjrl:


aswaterman:
Great, we need this.  I think there are a few more things to check:

- Right now it looks like this would error if you try to emit Firrtl, because ```toType``` will dereference null.  Presumably ```Wire(Vec(Seq()))``` would trigger this.  Not sure what the right fix is.
- It looks like ```cloneType``` needs to be special-cased to avoid dereferencing null in ```gen.cloneType```.
- ```apply(UInt)``` will also dereference null; it should probably throw an IndexOutOfBounds exception instead, as ```apply(0)``` does in this case.  (Actually, returning ```apply(0)``` in this case is a good way to do that.)
- ```typeEquivalent``` needs to be special-cased to return true if both lengths are 0, or else the comparisons on ```sample_element``` will dereference null.
- ```allElements``` should return an empty list if length is 0 (```sample_elements``` is again the problem here).  Maybe this (and the previous one) are best handled by making ```sample_element``` an Option instead.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -66,29 +66,32 @@ object Vec {
...
-    def doConnect(sink: T, source: T) = {
-      // TODO: this looks bad, and should feel bad. Replace with a better abstraction.
-      // NOTE: Must use elts.head instead of vec.sample_element because vec.sample_element has
-      //       WireBinding which does not have a direction
-      val hasDirectioned = elts.head match {
-        case t: Aggregate => t.flatten.exists(_.dir != Direction.Unspecified)
-        case t: Element => t.dir != Direction.Unspecified
+    if (elts.isEmpty) {
+      // Support empty Vecs.
+      new Vec(null.asInstanceOf[T], 0)
```
ducky64: null.asInstanceOf[T] raises serious red flags.

Possible alternative solutions:
- Make sample_element an Option[T]
- Remove sample_element entirely, using instead head of Vec elements, and handling the empty case properly

ducky64: Related: making this an Option will probably address some of @aswaterman 's cases at compile time. Using Vec elements head is also an option as long as we consistently check all accesses for empty. sample_element can also return elements head as an Option, which might be cleaner.

ucbjrl:
Could this be solved by converting `Vec(Seq[T]())` into `new Vec(gen[T], 0)` where `gen[T]` constructs an instance of `T`? I realize this is problematic where parameterized types are concerned.

Failing that, can we squirrel away the type of `T` so we can still use it in those cases where we were relying on the `sample_element` to provide `T`?

We still need to be able to deal with optional `sample_element`s or empty `elements`.

edwardcwang:
Related to #946

ucbjrl:
Closed since it seems there is another solution has crept in and #529 has been closed for almost two years.

PR #616, opened 2017-05-24
Update internal Pipe wiring - fixes #615

### LINKED ISSUE #615: chisel3.util.pipe is broken
Hello!
After last release, simple instantiation of pipe module gives error
example
```Scala
class Sample extends Module {
  val io = IO(new Bundle {
    val a = Input(Bool())
    val b = Input(UInt(32.W))
  })
  val pipe = Module(new Pipe(UInt(32.W), 32))
  pipe.io.enq.valid <> io.a
  pipe.io.enq.bits <> io.b
}
object Sample extends App {
  Driver.execute(Array("-td", "./"), () => new Sample)
}
```
dat code gives an error
Exception in thread "main" chisel3.internal.ChiselException: Connection between left (chisel3.core.Bool@cc) and source (chisel3.core.Bool@c2) failed @: Locally unclear whether Left or Right (both internal)
	at chisel3.internal.throwException$.apply(Error.scala:13)
	at chisel3.core.Data.bulkConnect(Data.scala:222)
	at chisel3.core.Data.$less$greater(Data.scala:273)
	at chisel3.util.Pipe$.apply(Valid.scala:41)
...
etc

ucbjrl:
Replace ambiguous bi-connect ("<>") with mono-connect (":=") for internal Pipe wiring.

### DIFF
--- a/src/main/scala/chisel3/util/Valid.scala
+++ b/src/main/scala/chisel3/util/Valid.scala
@@ -38,8 +38,8 @@ object Pipe
   def apply[T <: Data](enqValid: Bool, enqBits: T, latency: Int)(implicit compileOptions: CompileOptions): Valid[T] = {
     if (latency == 0) {
       val out = Wire(Valid(enqBits))
-      out.valid <> enqValid
-      out.bits <> enqBits
+      out.valid := enqValid
+      out.bits := enqBits
       out
     } else {
       val v = RegNext(enqValid, false.B)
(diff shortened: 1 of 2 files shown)


ucbjrl:
It reproduces the original failure reported by SpaceCowboyMax. It will fail if the mono-connections are replaced with the original bulk/bi connections.

aswaterman:
@ucbjrl We learned a few years ago that a common case of Pipe (register retiming) depends on the registers being in the same module as the unit being retimed, which is why the companion object instantiates them directly.

The Module named Pipe only exists for backwards compatibility, to support old code that writes ```Module(new Pipe)``` instead of ```Pipe```.  Maybe the Module should be moved into the compatibility layer.

ducky64:
Agree, we should document the rationale as well as deprecate the class (if possible) / move it into compatibility. Might be a separate PR though.

PR #620, opened 2017-05-30
Dont touch

jackkoenig:
Adds annotation for ensuring things don't get DCE-d.

The same annotation should be used to prevent CSE and ConstProp of certain signals, but this has not yet been implemented in Firrtl.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
@@ -2,7 +2,7 @@
 
 package chisel3.core
 
-import chisel3.internal.InstanceId
+import chisel3.internal.{Builder, InstanceId}
 import firrtl.Transform
 import firrtl.annotations.{Annotation, CircuitName, ComponentName, ModuleName}
 
@@ -28,3 +28,40 @@ case class ChiselAnnotation(component: InstanceId, transformClass: Class[_ <: Tr
     }
   }
 }
+
+/** Marks that a signal should not be removed by Chisel and Firrtl optimization passes
+  *
+  * @example {{{
+  * class MyModule extends Module {
+  *   val io = IO(new Bundle {
+  *     val a = Input(UInt(32.W))
+  *     val b = Output(UInt(32.W))
+  *   })
+  *   io.b := io.a
+  *   val dead = io.a +% 1.U // normally dead would be pruned by DCE
+  *   dontTouch(dead) // Marking it as such will preserve it
+  * }
+  * }}}
+  *
+  * @note Calling this on Data creates an annotation that Chisel emits to a separate annotations
+  * file. This file must be passed to Firrtl independently of the .fir file.
+  * [[chisel3.Driver.execute]] will do this automatically.
+  */
+object dontTouch { // scalastyle:ignore object.name
+  /** Marks a signal to be preserved in Chisel and Firrtl
+    *
+    * @note Requires the argument to be bound to hardware
+    * @param data The signal to be marked
+    * @return Unmodified signal `data`
+    */
+  def apply[T <: Data](data: T)(implicit compileOptions: CompileOptions): T = {
+    if (compileOptions.checkSynthesizable) {
+      Binding.checkSynthesizable(data, s"$data")
+    }
+    // TODO unify with firrtl.transforms.DontTouchAnnotation
+    val anno = ChiselAnnotation(data, classOf[firrtl.Transform], "DONTtouch!")
+    Builder.annotations += anno
+    data
+  }
+}
+
--- a/src/main/scala/chisel3/Driver.scala
+++ b/src/main/scala/chisel3/Driver.scala
@@ -146,10 +146,16 @@ object Driver extends BackendCompilationUtilities {
     af.write(circuit.annotations.toArray.toYaml.prettyPrint)
     af.close()
 
-    /* create custom transforms by finding the set of transform classes associated with annotations
-     * then instantiate them into actual transforms
-     */
-    val transforms = circuit.annotations.map(_.transform).toSet.map { transformClass: Class[_ <: Transform] =>
+    /** Find the set of transform classes associated with annotations then
+      * instantiate an instance of each transform
+      * @note Annotations targeting firrtl.Transform will not result in any
+      *   transform being instantiated
+      */
+    val transforms = circuit.annotations
+                            .map(_.transform)
+                            .distinct
+                            .filterNot(_ == classOf[firrtl.Transform])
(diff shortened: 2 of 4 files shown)


src/test/scala/chiselTests/DontTouchSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import chisel3._
+import chisel3.experimental.dontTouch
+import firrtl.{FirrtlExecutionSuccess, Transform}
+
+class HasDeadCodeChild(withDontTouch: Boolean) extends Module {
+	val io = IO(new Bundle {
+		val a = Input(UInt(32.W))
+		val b = Output(UInt(32.W))
+    val c = Output(Vec(2, UInt(32.W)))
```
edwardcwang: Indentation looks off here?

src/test/scala/chiselTests/DontTouchSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+		dontTouch(io.c)
+	}
+}
+
+class HasDeadCode(withDontTouch: Boolean) extends Module {
+	val io = IO(new Bundle {
+		val a = Input(UInt(32.W))
+		val b = Output(UInt(32.W))
+	})
+  val inst = Module(new HasDeadCodeChild(withDontTouch))
```
edwardcwang: Here as well?

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
```diff
@@ -28,3 +28,40 @@ case class ChiselAnnotation(component: InstanceId, transformClass: Class[_ <: Tr
...
+  *
+  * @example {{{
+  * class MyModule extends Module {
+  *   val io = IO(new Bundle {
+  *     val a = Input(UInt(32.W))
+  *     val b = Output(UInt(32.W))
+  *   })
+  *   io.b := io.a
+  *   val dead = io.a +% 1.U // normally dead would be pruned by DCE
+  *   dontTouch(dead) // Marking it as such will preserve it
```
edwardcwang: Just out of curiosity, is there a reason for using a function as opposed to an literal "@..." annotation? I suppose you could pass more parameters or something but in this case there are no extra parameters.

src/test/scala/chiselTests/DontTouchSpec.scala
jackkoenig: fixed

src/test/scala/chiselTests/DontTouchSpec.scala
jackkoenig: fixed

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
jackkoenig: We *could* use a Scala annotation instead, although unless I'm mistaken the macro would have to unwrap into a function call like this one anyway. Thoughts @ducky64

src/main/scala/chisel3/Driver.scala
```diff
@@ -149,7 +149,11 @@ object Driver extends BackendCompilationUtilities {
     /* create custom transforms by finding the set of transform classes associated with annotations
      * then instantiate them into actual transforms
      */
-    val transforms = circuit.annotations.map(_.transform).toSet.map { transformClass: Class[_ <: Transform] =>
+    val transforms = circuit.annotations
+                            .map(_.transform)
+                            .distinct
+                            .filterNot(_ == classOf[firrtl.Transform])
```
chick: I'm not sure what that filterNot is doing, removing transforms that firrtl implements perhaps.  If so shouldn't that be done in firrtl, seems like firrtl.Transform might be too specific.

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
edwardcwang: Or perhaps we could even support both methods (function and Scala annotation)?

src/main/scala/chisel3/Driver.scala
jackkoenig: This whole chunk of code instantiating transforms from annotations probably should be moved to Firrtl and we should rethink how we decide what transforms to instantiate (since it's possible someone will annotate a transform that can't be instantiated in this way). 

Specifically, this filterNot is preventing annotations that target firrtl.Transform (like dont touch) don't cause it to try to instantiate an instance of firrtl.Transform since it is an abstract class without a constructor. 

So long as this code stays in chisel this workaround is kind of mandatory because without it we get java invocation target exceptions when we try to have dont touch annotations.

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
chick: Seems like if it is going to be a Chisel annotation, it should look like a chisel annotation rather than a scala one

src/main/scala/chisel3/Driver.scala
chick: I see.

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
ducky64: As far as I know, we can't do a `@something`-style macro annotation inline, those can be applied to only types and classes.
You could make `dontTouch` a macro transform, though you don't gain anything in this case.

src/main/scala/chisel3/Driver.scala
ducky64: Is this something you want to throw a warning / error with helpful message on instead of just silently dropping?

edwardcwang: Or perhaps we could write the explanation above as a comment so we can remember why this is here?

chiselFrontend/src/main/scala/chisel3/core/ChiselAnnotation.scala
edwardcwang: I see; that makes sense.

src/main/scala/chisel3/Driver.scala
jackkoenig: So the issue here is that it's convenient for custom transforms to be included automatically if there's an Annotation that targets them. It's obvious that people do not intend for a transform to be instantiated when an annotation targets a built-in Transform that is run anyway, let alone when the targeted transform is firrtl.Transform which is the abstract superclass of all Transforms!

I'll add a comment to make this more clear to future readers, but improving the Transform auto instantiation API is beyond the scope of this PR imo.

PR #623, opened 2017-06-02
Too small fixes that caused me problems testing the interval stuff

chick:
@azidar not sure how this works for you, but I needed these fixes to run


chick:
Failing build check are due to this branch not using a corresponding branch of firrtl.

PR #628, opened 2017-06-14
Make deprecated debug point to and call dontTouch

jackkoenig:


colinschmidt:
Nice!
I think the message should say something like "Use DontTouch instead of debug in chisel3"
Now the deprecation message doesn't say what was deprecated just the solution which might be hard to apply if you don't know to get rid of debug.

jackkoenig:
Agreed, done.

jackkoenig:
@ucbjrl Do you know why Jenkins isn't running?

ucbjrl:
I suspect it's due to the move from ucb-bar to freechipsproject. I'll look
into it over the weekend.

ucbjrl:
I updated the repo URL in the Jenkins GitHub Pull Request Builder job and
things appear to be working.

jackkoenig:
I'm actually not sure if I think this is the right thing to do anymore. Debug hasn't done anything and has been deprecated for like 2 years, I think maybe we should just remove it in 3.1

jackkoenig:
I think we should move debug to compatibility layer and keep it deprecated but doing nothing there and just delete from chisel3.

PR #632, opened 2017-06-22
Add an assembly.sbt for chisel3

da-steve101:


ucbjrl:
ok to test

jackkoenig:
Why is this plugin needed here? Are you wanting to invoke Chisel on the command-line in a similar way to how rocket-chip invokes Firrtl?

da-steve101:
yeah not that important i guess, i just did it to make a jar with latest

Issue #578, opened 2017-04-13
The definitive binding refactor thread, request for comments


ducky64:
This is a summary of all the issues related to the bindings refactor (issues #440, #445, #419, #399, #368, #282, parts of #535, #543), and the proposed path forward. Work on this should begin soon.

## Terminology & Conventions

These will primarily be documentation clarifications.

- Terminally-bound nodes will be called _hardware_
- Non-terminally bound nodes will be called _chisel types_
- "conventional" (non-flipped) Bundle directionality is as a dataflow source or master (when relevant, for some buses).
  - This is consistent with current Decoupled convention.

## Directionality

- Nodes can have these directions: Input, Output, Bidirectional (for Aggregates, below, or Analog), or Unspecified (default).
- `Output(...)` and `Input(...)` will (effectively) clear direction of internal nodes, so the top-level node can be set to an input/output.
- Aggregate rules, including Bundles and Vecs:
  - All or no elements must be Unspecified.
  - Considered Bidirectional if there are both Input, Output, or Bidirectional elements, otherwise either Input, Output, or Unspecified.
  - Can only contain chisel types
- Internally: Deduplicate the directionality data source by using only hierarchical directionality (dropping leaf directionality). Leaf directionality can be resolved using parent links.
  - Since top-level bindings are required, that becomes a convenient place to create parent links.
  - Chisel (compatibility mode) didn't require top-level bindings, but Chisel also skips the directionality system. firrtlDirection (hierarchical directionality) is sufficient to emit FIRRTL.
  - For performance reasons, leaf directionality can still be cached.
  - Bonus: this may eliminate the need for `_onModuleClose`
- Parent links will be a private chisel3 feature for now.

## Binding operations

- `Wire(t)` must have `t` of chisel type
- `Reg(t), Mem(t), Vec(t)` must have `t` of chisel type that is not Bidirectional (either Input, Output, or Unspecified)
- `Vec(t)` additionally is defined to create a chisel type.
- `IO(t)`: must have `t` of chisel type with direction.
- Hardware instantiation binding operations will return a new node (rather than binding directly to the input chisel type node), which I think is current behavior.
- `Vec(args*)`, `Vec(elts)` deprecated, replace with `VecWire(elts)` (or similar, someone suggest good names). `elts` must be a hardware node, and this returns a hardware node.

## Node operations

- `.chiselCloneType` and `.cloneType` are to be deprecated.
- `.getType` operator introduced, legal on hardware nodes.
[...]

PR #635, opened 2017-06-28
More of the bindings refactor

ducky64:
Rest of the Bindings refactor
- Separate out initialization constructors, with deprecation:
  - Wire(init=...) => WireInit
  - Vec(elts) => VecInit
- Require Vec, Wire, and Mem constructors to take a chisel type, not a hardware type.
- Fixes to test suite to make it pass, because it was creating Wires from hardware types

Tests with rocket-chip pending, will fix the use cases in chisel test suite soon so they stop spewing deprecated warnings.

It's likely that rocket-chip will need some fixes to match the new semantics. That list to come soon...

Part 2 (of probably 3?) in a series of PRs for #578

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -76,7 +76,7 @@ sealed abstract class Aggregate extends Data {
   private[core] override def connectFromBits(that: Bits)(implicit sourceInfo: SourceInfo,
       compileOptions: CompileOptions): Unit = {
     var i = 0
-    val bits = Wire(UInt(this.width), init=that)  // handles width padding
+    val bits = WireInit(UInt(this.width), that)  // handles width padding
     for (x <- flatten) {
       x.connectFromBits(bits(i + x.getWidth - 1, i))
       i += x.getWidth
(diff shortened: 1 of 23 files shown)


ducky64:
rocket builds on this now, with one change to src/main/scala/uncore/axi4/Deinterleaver.scala
```diff
diff --git a/src/main/scala/uncore/axi4/Deinterleaver.scala b/src/main/scala/uncore/axi4/Deinterleaver.scala
index 3778c91..e7cb257 100644
--- a/src/main/scala/uncore/axi4/Deinterleaver.scala
+++ b/src/main/scala/uncore/axi4/Deinterleaver.scala
@@ -46,7 +46,7 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
         val qs = Seq.tabulate(endId) { i =>
           val depth = edgeOut.master.masters.find(_.id.contains(i)).flatMap(_.maxFlight).getOrElse(0)
           if (depth > 0) {
-            Module(new Queue(out.r.bits, beats)).io
+            Module(new Queue(out.r.bits.cloneType, beats)).io
           } else {
             Wire(new QueueIO(out.r.bits, beats))
           }

```

edwardcwang:
Why is the clonetype needed?

ducky64:
The new semantics require a chisel type on the Mem constructor in chisel3._. What happens here is that rocket invokes a Queue constructor (under Chisel._) passing in a hardware type, but the Queue internals are chisel3._ which requires the stricter semantics.

This is a larger problem in the compatibility layer, since many things are only validated at the chiselFrontend level (UInt(...), Reg(...), ...), but a higher-level library (like chisel3.util) may be passing arguments down but not validating them or passing through compile options.

edwardcwang:
Is it possible to introduce an overload for Queue that takes in a hardware type and calls clonetype to get the Chisel type automatically?

jackkoenig:
Even if it leads to a little code duplication we really need to not force `Chisel._` users to change their code

ducky64:
I could patch Queue specifically, but the more general problem here is that libraries (like chisel utils) are loose in what they accept and don't propagate (or even take) compile options. I'm not sure as to the best way to solve this.

Also, I think a fix in rocket is fine since it's only one use, rather than generally accepted practice (as in, I think the convention people are following is to pass in chisel types where chisel types are expected, and this was an instance where a break in convention was not caught and made it through.

edwardcwang:
> I could patch Queue specifically, but the more general problem here is that libraries (like chisel utils) are loose in what they accept and don't propagate (or even take) compile options. I'm not sure as to the best way to solve this.

What other libraries in utils have the same issue, and does it make sense to pre-emptively work around them (e.g. deal with compatibility errors like chisel vs hardware type).

> Also, I think a fix in rocket is fine since it's only one use, rather than generally accepted practice (as in, I think the convention people are following is to pass in chisel types where chisel types are expected, and this was an instance where a break in convention was not caught and made it through.

Is this convention valid in `Chisel._`-land though?

ducky64:
It's an issue that potentially affects everything in utils, but this seems to be the only problem in rocket.

This PR doesn't add strict semantics to Bundle construction, though. That might cause more issues, and a more robust workaround may be needed.

ducky64:
Resolution: add CompileOptions to Queue and hope that this is the only code that this change breaks, also add a big nasty dynamic deprecation warning when this is detected

ducky64:
@jackkoenig does this fix things?
(it seems the original rocket-chip breakage has been addressed upstream, in that the code isn't being called anymore...)

jackkoenig:
I pushed some changes, Mems weren't checking compile options properly.

In doing those changes I got an idea, what if `requireIsChiselType` also took compile options and didn't actually do the check for compatibility code? it would provide a more concise way for libraries to accept both chisel3 and Chisel code than what we're doing in Queue.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -326,6 +250,74 @@ sealed class Vec[T <: Data] private (gen: => T, val length: Int)
...
+    * @param n number of elements in the vector (the function is applied from
+    * 0 to `n-1`)
+    * @param gen function that takes in an Int (the index) and returns a
+    * [[Data]] that becomes the output element
+    */
+  def tabulate[T <: Data](n: Int)(gen: (Int) => T): Vec[T] = macro VecTransform.tabulate
+
+  def do_tabulate[T <: Data](n: Int)(gen: (Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] =
+    apply((0 until n).map(i => gen(i)))
+}
```
jackkoenig: So, if I recall correctly, we removed Vec.fill because it confused people when they were trying to create Chisel Types. Now that we're splitting Vec and VecInit, might it be time to bring it back?

ducky64: That sounds cleaner than VecInit(Seq.fill(...)).

ducky64:
requireIsChiselType taking compile options: I really don't want to thread compile options through the code any more than is necessary, it just bloats the argument lists.

I also don't want library writers using compile options, and it shouldn't be part of the chisel external API. (similarly with SourceInfo, at least until we develop abstractions that make sense for propagating library entry locators and internals locators). requireIsChiselType is also experimental, and I think we should look for better abstractions than having library writers haphazardly sprinkling require guards in their code.

PR #643, opened 2017-07-06
Ensure IO is non-null before attempting to autoWrapPorts.

ucbjrl:


ducky64:
What was this meant to fix? I think it makes sense to have it fail noisily if IO isn't defined?

ucbjrl:
The problem is due to attempting to validate any connections that precede the IO definition in compatibility mode. Failing with a null pointer exception seems wrong. But I suppose so would a reference to an (as yet) undefined IO port.

Perhaps we should require IO ports to be defined before any connections are made?

PR #651, opened 2017-07-22
Added MemShiftRegister to chisel utils

da-steve101:
Added a shift register implemented with a memory instead

### DIFF
--- a/src/main/scala/chisel3/util/MemShiftRegister.scala
+++ b/src/main/scala/chisel3/util/MemShiftRegister.scala
@@ -75,7 +75,7 @@ class MemShiftRegister[T <: Data]( genType : T, n : Int ) extends Module {
     when ( io.en ) {
       myMem.write( cntr._1, io.in )
     }
-    io.out := myMem.read( readAddr, io.en )
+    io.out := myMem.read( readAddr, true.B )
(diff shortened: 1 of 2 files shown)


ucbjrl:
Ok to test.

ucbjrl:
Please add some comments (scaladoc style) describing usage and parameters.

edwardcwang:
This seems very similar to a Queue - is there a way you could use that construction instead?

da-steve101:
I could use the Queue instead of a memory to implement it but i don't understand why I would. I would still need a wrapper around a Queue to implement the shift register and there would be additional overhead. Unless you are saying why add this to chisel when you can use a Queue?

edwardcwang:
Well, how large are these shift registers which you would like to implement? It just seems a bit like such a shift register with delay n is essentially a Queue with depth n and always enqueueing.

da-steve101:
If i used a Queue, instead of 1 Counter I would have 2. And then also have the logic to detect Full/Empty. I currently am working on an application that has unusually large SR but I agree that it wouldn't be that common.

da-steve101:
Perhaps an official "chisel-utils" repo should be created similar to testers and this should go in there?

Issue #419, opened 2016-12-16
cloneType vs. chiselCloneType


ducky64:
cloneType doesn't seem to preserve IO directionality, while chiselCloneType does. However, the names are highly unintuitive. 

Related to #368, it might be a good idea to have a `.clone` (which preserves directionality) and `.asType` / `.getType` which doesn't. (and in my opinion, the behavior is clearer too)

jackkoenig:
I agree that `.cloneType` vs `.chiselCloneType` is confusing. I don't think that `.clone` is a good candidate because of it's Java meaning and all of the baggage that comes with that, but I do think that `.asType` or `.getType` are good names. Picking these names is pretty tough though, and it might be wise to drop `.cloneType` unless it does the exact same thing it does now or did in the past...

ducky64:
If we go with the stricter type system in #368 I can see there being a use for two types of clone, one that preserves directionality / "bindings" (I've found it useful to be able to clone another block's IO) and another that does not (for instance, to use a IO data type in a Reg).

aswaterman:
Do avoid the name clone. That was the chisel1 name for cloneType and it was problematic for the reason @jackkoenig mentioned.

ucbjrl:
@sdtwigg used `newType` in the original #200 pr. My renaming that to `chiselCloneType`, although well-intentioned, was probably a mistake: I wanted to mark it _for internal use only_.

Do we need three type generators:
- `newUnboundType` (corresponding to the current `cloneType`
- `newUnboundTypeWithWidth` (corresponding to the `cloneTypeWithWidth`
- `newBoundType` (corresponding to the current `chiselCloneType`

I'm hoping we don't need to expose all three to user code. `cloneType` exists only because we can't reproduce the client's type construction environment without help from the client. Since it's now part of the public API (and we don't intend to change its semantics), it should probably retain its current name. I'm quite happy changing the name of the other two, and possibly using `newUboundType` internally, passing the buck to `cloneType` to do the heavy lifting.

ducky64:
Is there ever a case where we want clone type while dropping widths? While widths aren't part of the types as far as the type system is concerned, I think that's just an implementation detail.

[...]

Issue #193, opened 2016-05-17
Invalid use of `this.type` for clone methods


jix:
I noticed that `Data.cloneType` has a return type of `this.type` ([Data.scala:64](/ucb-bar/chisel3/blob/a26a0ad4504e99182e3548db8f407c3dad7302b0/chiselFrontend/src/main/scala/Chisel/Data.scala#L64)) which in scala isn't just the class of the receiver object, but a special type consisting of only the receiver object. It does not make much sense for a clone method to return the exact same instance it was invoked on, so at least it is misleading. See also [this discussion about using this.type for clone](http://www.scala-lang.org/old/node/6835.html).

It also requires explicit type casts for implementers (e.g. in [Decoupled.scala:15](/ucb-bar/chisel3/blob/a26a0ad4504e99182e3548db8f407c3dad7302b0/src/main/scala/Chisel/util/Decoupled.scala#L15)). Given scala's type system those casts would be invalid. They only work because at runtime the weaker JVM's type system is used to check casts.

AFAICT, the scala compiler would also be free to ignore the actual return value, and just use the receiver object instead. I don't think this is likely to happen though.

The intended type signature is probably `def cloneType: this.getClass`, except that this is not valid scala. The usual work-around involves using a trait (`trait CloneableType[T]`) which is implemented by derived classes (`class MyBusBundle(busWidth: Int) extends CloneableType[MyBusBundle]`).

I assume this is not what you'd want in Chisel, given the excessive verboseness when implementing parametrized bundles. It probably also wouldn't work with the reflection code that gives a generic cloneType implementation for non-parametrized bundles.

I have some vague ideas on a different approach, but wanted to ask whether a fix for this would be welcome, before spending time on it. The idea is to change uses of `foo.cloneType` to `Data.cloneType(foo)` or just `cloneType(foo)` with the right import, and making `foo.cloneType` return just `Data`.

The implementation would be along the lines of this:

``` scala
abstract class Data {
    ...
    def cloneType: Data
    ...
}
object Data {
    def cloneType[T <: Data](data: T): T = data.cloneType.asInstanceOf[T]
}
```

It would still be possible to write implementations of `cloneType` that return a Data object of the wrong type, causing an error at runtime. But this is the same as before with the explicit and incorrect cast to `this.type` needed for implementations.

There are some other invalid uses of `this.type` for clone like methods. The same would apply to them.

[...]

PR #653, opened 2017-07-28
the cloneType and chiselCloneType hot mess 🔥

ducky64:
Addresses #419 

Overview:
- cloneType is now marked (through comments only) as an internal API.
- chiselCloneType deprecated (and changed to cloneTypeFull internally, analogous to cloneTypeWidth).
- chiselTypeOf(data) introduced as the external API to get a chisel type from a hardware object
- Intended usage: cloning is an implementation detail, and chisel types and hardware objects both should act as immutable types, with operations like Input(...), Reg(...), etc returning a copy and leaving the original unchanged. Hence, the clone operations are all deprecated.
- Deletes what appears to be an unused Bundle companion object.
- Input(...), Output(...), Flipped(...) require the object to be unbound

Remaining issues:
- Several uses of chiselCloneType remain in utils and tests for compatibility reasons, chiselCloneType does not check that the input is hardware bound while chiselTypeOf(...) does check. This goes into strictness issues with #635 and needs more discussion.
- It would be nice for cloneType to return type Data, and have cloneTypeFull / etc fix up the types, because `this.type` is a nasty hack (see #193) and to avoid the `.asInstanceOf[this.type]` boilerplate needed in every cloneType implementation. It seems rocket-chip still has things that call cloneType directly so this causes compile errors. Only a small handful, though, so maybe this is something we could do? Otherwise, Scala does not seem to try implicit conversions on the wrong return type, so PML isn't a solution here.
- cloneType is not deprecate-able because it is simultaneously an internal and external API. Ideally, if the usages in rocket-chip are few, those could be fixed (with either chiselCloneType or chisel3-semantics chiselTypeOf) and the function (possibly could be?) made protected. (note, this will protect things within our control like Data.cloneType, but not user defined functions that don't explicitly protect their cloneType functions, especially Bundle sub-types)
- Cases where the user needs to define cloneType is where it get hairy, because that's where the intended abstraction does break down and users have to think about object uniqueness. Prime examples: ComplexAssign and RecordSpec test cases. Bundle annotations (or more advanced reflection-based cloning) might address some (perhaps even most) cases. Record especially might be difficult, because the elements can be defined however. This will probably be the biggest issue, especially since Bundle constructors can be deceptively wrong (what if you put a chisel type argument directly as a Bundle element and don't redefine the Bundle clone so the elements get cloned?)

This should not be merged yet, this is just a starting point to think about these issues.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -89,10 +89,10 @@ object Vec {
     *
     * @note elements are NOT assigned by default and have no value
     */
-  def apply[T <: Data](n: Int, gen: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = new Vec(gen.chiselCloneType, n)
+  def apply[T <: Data](n: Int, gen: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = new Vec(gen.cloneTypeFull, n)
 
   @deprecated("Vec argument order should be size, t; this will be removed by the official release", "chisel3")
-  def apply[T <: Data](gen: T, n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = new Vec(gen.chiselCloneType, n)
+  def apply[T <: Data](gen: T, n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = new Vec(gen.cloneTypeFull, n)
 
   /** Creates a new [[Vec]] composed of elements of the input Seq of [[Data]]
     * nodes.
@@ -588,9 +588,3 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     */
   override def toPrintable: Printable = toPrintableHelper(elements.toList.reverse)
 }
-
-private[core] object Bundle {
-  val keywords = List("flip", "asInput", "asOutput", "cloneType", "chiselCloneType", "toBits",
-    "widthOption", "signalName", "signalPathName", "signalParent", "signalComponent")
-}
-
--- a/chiselFrontend/src/main/scala/chisel3/core/Data.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Data.scala
@@ -137,6 +137,15 @@ private[core] object cloneSupertype {
   }
 }
 
+/** Returns the chisel type of a hardware object, allowing other hardware to be constructed from it.
+  */
+object chiselTypeOf {
+  def apply[T <: Data](target: T): T = {
+    requireIsHardware(target)
+    target.chiselCloneType.asInstanceOf[T]
+  }
+}
+
 /**
 * Input, Output, and Flipped are used to define the directions of Module IOs.
 *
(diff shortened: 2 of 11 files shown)


edwardcwang:
Should note (at least for now) the necessity of overriding cloneType for user-defined types (e.g. subclasses of Bundle).

ducky64:
Resolution:
- For internal APIs that need to be exposed to Chisel._ compatibility (like chiselCloneType), qualify them into some namespace to prevent people from just using them.
- No resolution for the cloneType/chiselCloneType mess when the nice "all chisel types are immutable and you don't need to worry about object reference" abstraction breaks, because we don't know how to fix it. Goal is to reduce the number of cases where the user needs to override cloneType by incremental whackamole.

jackkoenig:
Alright I've merged master into this (@ducky64 if you could check my work that would be great because it was nontrivial). Master is now bumped on rocket-chip for the first time in months so we can more easily check this against big code bases!

ducky64:
@jackkoenig changes look good.

jackkoenig:
Okay so the only problem I see with this is pretty minor but I want to see if there's a better solution. Basically, this PR makes it to where if you are using chiselCloneType (be it in compatibility or regular chisel3 code) you get inferred existential types. This is just an annoying warning, but I don't understand why it now comes up but didn't before. It's also problematic if you default to erroring on warnings (you know, good practice).

An example that illustrates the problem:

```scala
import chisel3._

class MyTestSubModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(32.W))
    val out = Output(UInt(32.W))
  })
  io.out := io.in
}

class MyTestModule extends Module {
  val child = Module(new MyTestSubModule)
  val io = IO(child.io.chiselCloneType)
  io <> child.io
}

object Test extends App {
  chisel3.Driver.execute(args, () => new MyTestModule)
}
```

This simply passes through the IO of a child module. On master, there is no warning, on this branch we get the existential type warning for the io of the parent module (MyTestModule).

Weirdly, if we change the io of the parent module to flip the child io type twice, the warning goes away:
```scala
class MyTestModule extends Module {
  val child = Module(new MyTestSubModule)
  val io = IO(Flipped(Flipped(child.io.chiselCloneType)))
  io <> child.io
}
```

Is there any way we can remove this warning without forcing users to change their code or build configuration?

ducky64:
Strange... chiselCloneType was moved to compatibility using PML, so instead of returning `this.type`, it's now:
```scala
implicit class cloneTypeable[T<:Data](val target: T) extends AnyVal {
   def chiselCloneType: target.type = {
     DataMirror.internal.chiselTypeClone(target).asInstanceOf[target.type]
   }
 }
```

Also interestingly, IO is defined as
```scala
protected def IO[T<:Data](iodef: T): iodef.type = {
```
while Flipped is defined as
```scala
def apply[T<:Data](source: T): T = {
```

Does redefining IO to return T (instead of iodef.type) help? Or does that also break things? I'm not sure why IO is defined that way...

jackkoenig:
Changing IO to return `T` instead of `iodef.type` helps for this specific case, but not for a minorly modified version:

```scala
class MyTestModule extends Module {
  val child = Module(new MyTestSubModule)
  val io = IO(new Bundle {
    val port = child.io.chiselCloneType
  })
  io.port <> child.io
}
```

However, changing `chiselCloneType` to return `T` instead of `target.type` did fix it. The question is why do they return the structural type instead of just the type parameter type?

ducky64:
I'm not sure why IO returns a structural type, but if it works with parameter types, we should change it to that.
I think the reason chiselCloneType returns a structural type is because the old style (this.cloneType) returns this.type, and this was the minimal change. But again, I think the type parameter type should be preferred if it works. A long time ago, the uses of this.type was flagged as an issue which we're slowly solving.

PR #660, opened 2017-08-02
Disable aggregation until we can apply it selectively

ucbjrl:
or sbt-scoverage is updated to deal with dependencies.
Unfortunately, enabling this revives the publishing of the coreMacros and chiselFrontend jars and we don't want that.


jackkoenig:
Ahh bummer, so coverage will still be split between the different subprojects?

ucbjrl:
Yes, it is quite disappointing. My attempts to work around it have so far failed so I think it's better to pull it than start spreading problematic jars.

PR #667, opened 2017-08-07
Give default direction to children of Vecs in compatibility code

jackkoenig:
Basically, in compatibility mode direction assigning, Vecs were being treated like Records/Bundles defined in `chisel3._` code and thus not actually assigning default directions to their children.


jackkoenig:
@ucbjrl, is Jenkins down?

ucbjrl:
retest this please

jackkoenig:
Thanks!

PR #670, opened 2017-08-11
README: Java package

schoeberl:
For a start from a plain Ubuntu you need Java first before sbt


edwardcwang:
Note that you don't strictly even need root to run Chisel (assuming the system JDK is already installed) - sbt and verilator both don't require root.

PR #676, opened 2017-08-23
Added API to get Verilog from Chisel

azidar:


src/main/scala/chisel3/Driver.scala
```diff
@@ -95,6 +95,20 @@ object Driver extends BackendCompilationUtilities {
 
   def emit[T <: RawModule](ir: Circuit): String = Emitter.emit(ir)
 
+  def emitVerilog[T <: RawModule](gen: () => T): String = {
```
ducky64: Why are two signatures needed? And can you just make one call the other (so we can be a little more DRY)?

src/main/scala/chisel3/Driver.scala
```diff
@@ -95,6 +95,20 @@ object Driver extends BackendCompilationUtilities {
...
+    execute(Array[String](), gen) match {
+      case ChiselExecutionSuccess(_, _, Some(firrtl.FirrtlExecutionSuccess(_, verilog))) => verilog
+      case _ => sys.error("Cannot get Verilog!")
+    }
+  }
+
+  def emitVerilog[T <: RawModule](gen: => T): String = {
```
edwardcwang: Do we want scaladocs here?

src/main/scala/chisel3/Driver.scala
```diff
@@ -95,6 +95,18 @@ object Driver extends BackendCompilationUtilities {
 
   def emit[T <: RawModule](ir: Circuit): String = Emitter.emit(ir)
 
+  /** Elaborates the Module specified in the gen function into Verilog
+    *
+    *  @param gen a function that creates a Module hierarchy
+    *  @return the resulting String containing the design in Verilog
+    */
+  def emitVerilog[T <: RawModule](gen: => T): String = {
+    execute(Array[String](), { () => gen }) match {
```
edwardcwang: Do we ever care about wanting to pass arguments to emitVerilog? (e.g. optional argument)

src/main/scala/chisel3/Driver.scala
```diff
@@ -95,6 +95,18 @@ object Driver extends BackendCompilationUtilities {
...
+      case ChiselExecutionSuccess(_, _, Some(firrtl.FirrtlExecutionSuccess(_, verilog))) => verilog
+      case _ => sys.error("Cannot get Verilog!")
```
edwardcwang: Does it make sense to use ChiselException so that this can be caught?

src/main/scala/chisel3/Driver.scala
chick: Seems like an options manager or an array of string should be passed in.  overloaded definitions for each one

PR #684, opened 2017-08-29
Stop Queue while reset is high

jkorinth:
`enq.ready` and `deq.valid` of `Queue` should never be true while reset is high.

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -233,8 +233,8 @@ class Queue[T <: Data](gen: T,
     maybe_full := do_enq
   }
 
-  io.deq.valid := !empty
-  io.enq.ready := !full
+  io.deq.valid := !reset & !empty
+  io.enq.ready := !reset & !full
   io.deq.bits := ram(deq_ptr.value)
 
   if (flow) {
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -233,8 +233,12 @@ class Queue[T <: Data](gen: T,
     maybe_full := do_enq
   }
 
-  io.deq.valid := !reset & !empty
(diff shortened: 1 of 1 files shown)


jkorinth:
Ok, I see. But `io.deq.valid` depends on `!maybeFull` and `io.enq.ready` on `maybeFull`, so I'm not sure how to accomplish that without using `reset`? As far as I can tell, `when (reset) { io.deq.valid := false.B; io.enq.ready := false.B }` generates the same FIRRTL as the commit above.

jkorinth:
__Update__: I was wrong, the generated code _does_ differ subtly and it seems preferable to use `when (reset)`. Fixed in 6f10be0e873bfede0d4eda1f64cf9eec07624120.

mwachs5:
When (reset) should just be the same as RegInit. If that's what you mean, please use that primitive.

mwachs5:
And if that's not what you mean, then you need an asynchronous reset or multiple clock cycles to flush out those signals.

jkorinth:
I did not use `RegInit` because it would introduce an additional delay of one cycle, which I assume may cause regressions. How would you specify an asynchronous reset in Chisel?

mwachs5:
This PR makes me think that you are using the Queue primitive incorrectly. The blocks which are looking at the ready/valid signals should also be using reset  (not as part of their combinational outputs). So you should not need to condition these outputs on reset at all. 

Can you share how you are trying to use the Queue?

aswaterman:
As @mwachs5 said, this Queue (like most things in the chisel standard library) assume the thing(s) it's connected to are in reset at the same time. Other uses are outside the scope of this library. For your use case, you can wrap Queue to gate deq.valid with reset, or use Queue differently.

jkorinth:
Ok. I'd still argue that the current behavior is a breach of the ready/valid contract: `Queue` signals readiness to consume an input even though it cannot (since its reset line is still high). If the producer's reset is offset by even just one cycle, data loss occurs.

PR #693, opened 2017-09-15
Incorporate lessons learned from latest publishing.

ucbjrl:
Changes to sbt incorporated from release branch. This is not strictly required, but is an attempt to keep master and release branches relatively in sync.


build.sbt
```diff
@@ -92,12 +89,17 @@ lazy val chiselSettings = Seq (
...
-    Resolver.sonatypeRepo("releases")
-  ),
+val defaultVersions = Map("firrtl" -> "1.1-SNAPSHOT")
+
+lazy val chiselSettings = Seq (
+  name := "chisel3",
+
+  // Provide a managed dependency on X if -DXVersion="" is supplied on the command line.
+  libraryDependencies ++= (Seq("firrtl").map {
+  dep: String => "edu.berkeley.cs" %% dep % sys.props.getOrElse(dep + "Version", defaultVersions(dep)) }),
```
jackkoenig: Should this line be tabbed?

ducky64: Agree, this should be tabbed

ducky64: Can you also merge the dep into the rest of them? Unless you anticipate dependencies on more than just firrtl, that map really isn't doing much.

ucbjrl: Right. It was mostly to preserve style with the other repos where the map includes more/all of the BIG4.

Issue #418, opened 2016-12-16
Bundle literal constructors


ducky64:
This has been worked around in many ways, from assigning all the parts separately, to manually writing functions that can be invoked in one line but do the assignments inside, to compile-time-unsafe Maps. Theoretically, it should be possible to use macros to generate a companion object that constructs a bundle literal.

For example, if I have
```
@bundle
class MyBundle extends Bundle {
  val a = Bool()
  val b = UInt(8.W)
}
```
I want to be able to do
```
wire := MyBundle(a=true.B, b=0x42.U)
```
(ideally, like above, it would require a full set of keyword arguments, though how to accomplish this is uncertain...)

Also, if something could be done that addresses #417, this would be super useful in a testing context since it would not require the compile-time-unsafe string maps.

ducky64:
Superseded by #805

Issue #417, opened 2016-12-16
Statically usable literals


ducky64:
Currently, once you create a UInt literal, you can't get the literal back out. For RTL construction purposes, this isn't much of a issue (since you don't usually need it back), but because Chisel's literal from string constructor supports more radix types than base Scala (notably, binary), it helps to be able to use the same syntax for example when writing testvectors.
One possible proposal is for UIntLit to extend UInt, where UIntLit has an additional method to get the value. UIntLit would be returned by all literal constructors.

aswaterman:
Chisel2 had isLit and litOf methods defined on Data (or maybe Bits?). The API wasn't ideal, but the feature was a useful feature from time to time.  It was also even useful for RTL, as sometimes an implementation can be improved on the basis of constness. It's the same rationale behind gcc offering __builtin_constant_p.

sdtwigg:
For constructing binary and hexadecimal numbers succinctly, you can just have a StringContext for them, e.g.: h"abcd" or b"1111" that returns a BigInt.

Similarly, you could add an inspection function to see if some UInt or SInt is currently bound to being a literal. (It is really odd how Data currently has the literal methods.)

ducky64:
So the StringContext is a method, as would be PML on strings (i.e. "1111".b), but Chisel needs a standardized way to do this on tests, otherwise everyone else is going to write their own and it's going to be a giant mess. I think the most intuitive API would allow these to be specified the same way as Chisel literals (where we can do like "b111".U). Considering that bundle literals are also a desired feature, it should interoperate with that too, when (if?) it happens.

sdtwigg:
No, you should actually read the documentation: http://www.scala-lang.org/api/2.12.0/scala/StringContext.html

ducky64:
I know how StringContext works (b"1111"). PML ("1111".b) was another way that's been commonly suggested. Yes, they're different (I assume this is what you mean).

For the purposes of this discussion, what I want to resolve is standardizing literal specification in a way that's accessible for a testbench, and in a way that would both be consistent with these literals can currently be specified ("b1111".U) as well as potentially upcoming features (like Bundle literals).

sdtwigg:
[...]

PR #694, opened 2017-09-16
Aggregates can now be marked as literals.

grebe:
This is inspired by dsptools DspReal(), which is implemented as a Bundle with a UInt inside. Bundles currently can't be literals, which causes problems because it is totally natural to want a literal DspReal().

This is an initial stab at making literal bundles work in this context. It doesn't go anywhere towards addressing #417, although I think it does address #418 to an extent. This is also related to ucb-bar/dsptools#55 and perhaps ucb-bar/dsptools#87.

The big difference between how lits need to work for Aggregates and UInt/SInt/etc. is that UInt/SInt/etc aren't meant to be subclassed, whereas Aggregates are (especially Bundle and Record). Some API needs to be provided for users to mark things as literals (checking that they actually are).

Currently, I check by calling isLit() on all children. This is fine for UInt/SInt/etc., but is problematic for Aggregates because they currently don't mark themselves as lits. I added a private var that gets set to something when you try to mark an Aggregate as a lit, but the actual value is ignored because I don't do anything with it. I'm not a fan of doing this, but not familiar enough with the backend to know what a better alternative would be.

@ducky64 @jackkoenig @chick

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -34,8 +34,8 @@ sealed abstract class Aggregate extends Data {
   }
 
   protected[core] override def hasBinding = {
-   checkBinding
-   super.hasBinding
+    checkBinding
+    super.hasBinding
   }
 
   protected[core] override def binding = {
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -15,23 +15,45 @@ import chisel3.internal.sourceinfo._
   * of) other Data objects.
   */
 sealed abstract class Aggregate extends Data {
-  private var _litArg: Option[LitArg] = None
-  override def litArg = _litArg
+  protected def childLitArgs = getElements.map(_.litArg).foldLeft[Option[Seq[LitArg]]](Some(Seq[LitArg]())) {
+    case (Some(elems), Some(lit)) => Some(elems :+ lit)
+    case _                        => None
+  }
+  override def litArg = {
+    // TODO[grebe] make a lit for Aggregates
+    childLitArgs.map { x => ULit(0, 1.W) }
+  }
+
+  private var checkedBinding = false
+  private def checkBinding = {
+    // check if this aggregate is a literal, and bind it if it hasn't been bound yet
+    if (!checkedBinding && getElements.length > 0 && childLitArgs.isDefined) {
+      this.bind(LitBinding()) //, SpecifiedDirection.Unspecified)
+    }
+    checkedBinding = true
+  }
+
+  protected[core] override def hasBinding = {
+   checkBinding
+   super.hasBinding
+  }
+
+  protected[core] override def binding = {
+    checkBinding
+    super.binding
+  }
 
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection) {
     binding = target
 
-    target match {
-      case LitBinding() => _litArg = Some(ULit(0, 1.W)) // TODO[grebe] something better here
-      case _ =>
+    if (isLit) {
+      getElements.foreach { elem => require(elem.isLit()) }
+      return ()
     }
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
     for (child <- getElements) {
-      target match {
-        case LitBinding() => require(child.isLit())
-        case _ => child.bind(ChildBinding(this), resolvedDirection)
-      }
+      child.bind(ChildBinding(this), resolvedDirection)
     }
 
     // Check that children obey the directionality rules.
@@ -459,6 +481,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
 
   private[chisel3] final def allElements: Seq[Element] = elements.toIndexedSeq.flatMap(_._2.allElements)
 
+  protected override def childLitArgs = elements.map({ case (_, data) => data.litArg }).foldLeft[Option[Seq[LitArg]]](Some(Seq[LitArg]())) {
+    case (Some(elems), Some(lit)) => Some(elems :+ lit)
+    case _                      => None
+  }
+
   override def getElements: Seq[Data] = elements.toIndexedSeq.map(_._2)
 
   // Helper because Bundle elements are reversed before printing
@@ -564,7 +591,7 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     // - A one-parameter constructor for a nested Bundle, with the enclosing
     //   parent Module as the argument
     val constructor = this.getClass.getConstructors.head
-    try {
+    val newBundle = try {
       val args = Seq.fill(constructor.getParameterTypes.size)(null)
       constructor.newInstance(args:_*).asInstanceOf[this.type]
     } catch {
@@ -581,6 +608,10 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
         Builder.exception(s"Parameterized Bundle ${this.getClass} needs cloneType method")
         this
     }
+
+    require(!newBundle.isLit, s"Bundle ${this.getClass} has a bad cloneType: it should never return a literal")
+
+    newBundle.asInstanceOf[this.type]
   }
 
   /** Default "pretty-print" implementation
@@ -597,11 +628,3 @@ private[core] object Bundle {
     "widthOption", "signalName", "signalPathName", "signalParent", "signalComponent")
 }
 
-object Aggregate {
-  def LitBind(agg: Aggregate): Unit = {
-    agg.getElements.foreach { case x =>
-      require(x.isLit(), "Attempting to bind an aggregate as a lit with non-literal members")
-    }
-    agg.bind(LitBinding())
-  }
-}
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -15,12 +15,23 @@ import chisel3.internal.sourceinfo._
   * of) other Data objects.
   */
 sealed abstract class Aggregate extends Data {
+  private var _litArg: Option[LitArg] = None
+  override def litArg = _litArg
+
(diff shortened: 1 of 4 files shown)


src/test/scala/chiselTests/BundleLiterals.scala
```diff
@@ -0,0 +1,66 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import org.scalatest._
+
+class BundleWithLit(lit: Option[Int]=None) extends Bundle {
+  val internal = lit.map(_.U).getOrElse(UInt(4.W))
+  override def cloneType = new BundleWithLit(lit).asInstanceOf[this.type]
```
grebe: Is this a mistake? I just realized that cloneType of a lit should not be a lit. This should probably be `new BundleWithLit().asInstanceOf[this.type]`

src/test/scala/chiselTests/BundleLiterals.scala
```diff
@@ -0,0 +1,66 @@
...
+    val out = Output(BundleWithBundleWithLit())
+  })
+  io.out := BundleWithBundleWithLit(3)
+}
+
+class BundleLiteralsSpec extends FlatSpec with Matchers {
+  behavior of "Bundle literals"
+
+  it should "build the module without crashing" in {
+    println(chisel3.Driver.emitVerilog( new BundleWithLitModule ))
```
edwardcwang: You may want to compare firrtl output, instead of just emitting Verilog.

src/test/scala/chiselTests/BundleLiterals.scala
```diff
@@ -0,0 +1,66 @@
...
+
+class BundleWithBundleWithLit(lit: Option[Int]=None) extends Bundle {
+  val internal = lit.map(BundleWithLit(_)).getOrElse(BundleWithLit())
+  override def cloneType = new BundleWithBundleWithLit(lit).asInstanceOf[this.type]
+}
+
+object BundleWithBundleWithLit {
+  def apply(x: Int) = {
+    val ret = new BundleWithBundleWithLit(Some(x))
+    Aggregate.LitBind(ret)
```
edwardcwang: Seems strange to have to call this in user code; why is it here?

src/test/scala/chiselTests/BundleLiterals.scala
```diff
@@ -0,0 +1,66 @@
...
+  def apply() = {
+    new BundleWithBundleWithLit()
+  }
+}
+
+class BundleWithLitModule extends Module {
+  val io = IO(new Bundle {
+    val out = Output(BundleWithLit())
+  })
+  io.out := BundleWithLit(3)
```
edwardcwang: Is this the intended bundle literal syntax? Also, what about bundles with multiple members?

src/test/scala/chiselTests/BundleLiterals.scala
grebe: This is patterned after `UIntFactory`, which isn't user code b/c you don't subtype `UInt`, but users do subtype bundles. I suppose you could have an aggregate try to detect if its members are literals, but that seemed to be a bigger change so I didn't take that step.

ducky64:
Agree with @edwardcwang, the frontend definitely needs more thought - from the testcases, the user has to do a lot of work to write bundle literal constructors. As for this just being a backend implementation, I need to cache back in the bindings and bundles system, but as far as I can remember, is there a reason a literal Bundle needs to be bound as such?

Literal bundle constructors (possibly implemented using macros) is on the docket of stuff for 3.1.0, and will likely be a part of Testers2. As a timeframe estimate, both are probably at least two months out, there's a list of more pressing stuff in terms of Chisel3 engineering dev.

src/test/scala/chiselTests/BundleLiterals.scala
grebe: I don't know if I'd say *the* literal syntax, but it is the kind of syntax we use with DspReal().
You're companion object could take multiple parameters

src/test/scala/chiselTests/BundleLiterals.scala
grebe: Fair enough, I just did a "it's doing something" test to start with

grebe:
I'm trying to remember the exact issues I was seeing in dsptools, but our solution is to create a wire and assign to it with literals. Sometimes you don't want to create the wire- I can't remember what situations cause problems (maybe if you're passing around a literal bundle just to get a width out of it or call cloneType in an IO? Like `new TypeParameterizedModule(proto=DspReal(0.3))`). This means that it is somewhat confusing whether something will create a wire or not.

It's more than just the annoyance of creating new wires, there's a reason UInts/SInts/etc. treat literals specially.

I agree that the example of the frontend is clunky. I think it can be made easier, e.g. a `def MarkAsLit[T <: Aggregate](in: T): T` so your companion object would just have `def apply(x: Int) = MarkAsLit(new Whatever(x))` (doesn't eliminate the boilerplate or fragility, but makes it less ugly).

Do you see literal bundle constructors as changing how lits work at the core? It seems to me that whatever is implemented will have to do something to set litArg to `Some(...)`. Perhaps it will do so by rewriting literal bundles to be of a subtype of Bundle that sets litArg, rather than the hackish private var trick.

ducky64:
I'm not sure what the implementation for bundle literals should look like yet, that's something that needs more thought, possibly a fresh look into refactoring the entire literals system if warranted.

src/test/scala/chiselTests/BundleLiterals.scala
```diff
@@ -0,0 +1,66 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import org.scalatest._
+
+class BundleWithLit(lit: Option[Int]=None) extends Bundle {
+  val internal = lit.map(_.U).getOrElse(UInt(4.W))
```
chick: What is the OrElse doing?

src/test/scala/chiselTests/BundleLiterals.scala
chick: I think that's right, that cloneType should not return a Lit.

src/test/scala/chiselTests/BundleLiterals.scala
grebe: Providing a default width

grebe:
@ducky64 I agree that if you are going to tackle the entire bundle literals problem (with implications in chisel, testers, and firrtl) that's a large task that should be done thoughtfully and with a process. I should be clear that what I'm pushing for here is to use the existing bindings infrastructure to be able to bind bundles as literals and nothing more ambitious than that.

I just pushed a cleaner version of the original idea. The examples are pretty clean, I think. It adds another nuance to how cloneType can fail (if it returns something that is bound), but this is already something that can happen and cause problems if you aren't careful.

One slightly nuanced thing is that empty Bundles should not be marked as literal. They're used all over the place in our testers, which will result in a module complaining that it's IO is hardware when it should be chisel.

For whatever reason a Vec test is failing, I haven't tracked down why. I don't think it's a big deal, I think it's probably an easy fix if you understand how Vecs work.

src/main/scala/chisel3/testers/BasicTester.scala
```diff
@@ -13,7 +13,7 @@ import internal.sourceinfo.SourceInfo
 
 class BasicTester extends Module() {
   // The testbench has no IOs, rather it should communicate using printf, assert, and stop.
-  val io = IO(new Bundle())
+  val io = IO(new Bundle {val dummy = Input(UInt(0.W))})
```
grebe: Oops, forgot to clean this up, this is no longer necessary.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -15,9 +15,42 @@ import chisel3.internal.sourceinfo._
   * of) other Data objects.
   */
 sealed abstract class Aggregate extends Data {
+  protected def childLitArgs = getElements.map(_.litArg).foldLeft[Option[Seq[LitArg]]](Some(Seq[LitArg]())) {
+    case (Some(elems), Some(lit)) => Some(elems :+ lit)
+    case _                        => None
+  }
+  override def litArg = {
+    // TODO[grebe] make a lit for Aggregates
+    childLitArgs.map { x => ULit(0, 1.W) }
```
ducky64: Things should fail noisily if this is an unexpected case instead of propagating bad values

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -15,9 +15,42 @@ import chisel3.internal.sourceinfo._
...
+  private def checkBinding = {
+    // check if this aggregate is a literal, and bind it if it hasn't been bound yet
+    if (!checkedBinding && getElements.length > 0 && childLitArgs.isDefined) {
+      this.bind(LitBinding()) //, SpecifiedDirection.Unspecified)
+    }
+    checkedBinding = true
+  }
+
+  protected[core] override def hasBinding = {
+    checkBinding
```
ducky64: This looks kind of hacky, kind of like a game of whackamole to insert checks into all the right places, it would make more sense for higher-level code to apply the binding as it is being set to a literal.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -570,6 +608,10 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
         Builder.exception(s"Parameterized Bundle ${this.getClass} needs cloneType method")
         this
     }
+
+    require(!newBundle.isLit, s"Bundle ${this.getClass} has a bad cloneType: it should never return a literal")
```
ducky64: This require won't fire if the user overrides cloneType (essentially always given your current examples), right?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
edwardcwang: The problem is that bind() is `private[chisel3]` which makes it impossible to do that in user-land right now.

ducky64:
Superseded by #820 (with more discussion in #805), does that get closer to addressing your use cases?

grebe:
Yeah, this should be closed.

PR #695, opened 2017-09-19
Generate aggregate coverage but publish a single Jar

ucbjrl:


ucbjrl:
**NOTE**: This will require an update to rocket-chip's project/plugins to include sbt-coverage (freechipsproject/rocket-chip#1004).

PR #713, opened 2017-11-03
Add issue and pull_request templates.

ucbjrl:

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [x] Other enhancement

* **Impact**
  - [x] no functional change
  - [ ] API modification

* **Development Phase**
  - [x] proposal
  - [ ] implementation

* **Release Notes**

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -7,11 +7,6 @@ This repository's issues are reserved for feature requests and bug reports.
   - [ ] Feature request
   - [ ] Other enhancement
 
-* **Development phase**
-  - [ ] Request
-  - [ ] Proposal
-  - [ ] Implementation
-
 * **If the current behavior is a bug, please provide the steps to reproduce the problem:**
   * **What is the current behavior?**
   * **What is the expected behavior?**
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,23 @@
+**Note: for support questions, please use stackoverflow**.
+This repository's issues are reserved for feature requests and bug reports.
+
+* **Type of issue**
+  - [ ] bug report
+  - [ ] feature request
+  - [ ] other enhancement
+
+* **Development Phase**
+  - [ ] request
+  - [ ] proposal
+  - [ ] implementation
+
+* **If the current behavior is a bug, please provide the steps to reproduce the problem**
+  * **What is the current behavior?**
+  * **What is the expected behavior?**
+  * **Please tell us about your environment:**
+    - version: `3.0-SNAPSHOT`
+    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
+
+* **What is the use case for changing the behavior?**
+
+* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,23 +1,25 @@
-**Note: for support questions, please use stackoverflow**.
+**Note: for support questions, you are strongly advised to use Stack Overflow**.
+
 This repository's issues are reserved for feature requests and bug reports.
 
 * **Type of issue**
-  - [ ] bug report
-  - [ ] feature request
-  - [ ] other enhancement
+  - [ ] Bug report
+  - [ ] Feature request
+  - [ ] Other enhancement
 
-* **Development Phase**
-  - [ ] request
-  - [ ] proposal
-  - [ ] implementation
+* **Development phase**
+  - [ ] Request
+  - [ ] Proposal
+  - [ ] Implementation
 
-* **If the current behavior is a bug, please provide the steps to reproduce the problem**
+* **If the current behavior is a bug, please provide the steps to reproduce the problem:**
   * **What is the current behavior?**
   * **What is the expected behavior?**
   * **Please tell us about your environment:**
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
+**Note: for support questions, please use stackoverflow**.
```
ducky64: Do people really ask support questions using pull requests?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
...
+This repository's issues are reserved for feature requests and bug reports.
+
+* **Type of issue**
```
ducky64: This section doesn't seem applicable to a pull request. Maybe a field like "related issue (if applicable)".

Category might make more sense, with options being "bug fix", "new feature", and "other (describe)"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
...
+  - [ ] bug report
+  - [ ] feature request
+  - [ ] other enhancement
+
+* **Development Phase**
```
ducky64: What does this mean for a pull request that has associated code? Are you trying to categorize these as "implementation sketch" vs. "production / merge-ready code"?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
...
+  - [ ] request
+  - [ ] proposal
+  - [ ] implementation
+
+* **Impact**
+  - [ ] no functional change
+  - [ ] API addition (no impact on existing code)
+  - [ ] API modification
+
+* **If the current behavior is a bug, please provide the steps to reproduce the problem**
```
ducky64: Is this section relevant for a PR with code?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+
+* **Type of issue**
+  - [ ] Bug report
+  - [ ] Feature request
+  - [ ] Other enhancement
+
+* **Development phase**
+  - [ ] Request
+  - [ ] Proposal
+  - [ ] Implementation
```
ducky64: Is an implementation expected with a issue (as opposed to a pull request)?

.github/PULL_REQUEST_TEMPLATE.md
ucbjrl: Here's Danny's email comment on this (he's welcome to replace this with a real comment):

> I think a "request" will be generally in the form of a suggestions to improve a feature or a bug report where as a "proposal" would be a proposed solution to a "request" that needs more discussion. 
> 
> For example, if some one says "It would be great if we could add annotations, here are what I want to use them for", that would be a "request". And then if some one else says "here's a approach to an annotation system", that would be a "proposal". 
> 
> Also, some of the current pull requests in the Chisel repo such as PR#694 and PR#645 should be categorized as "proposal" rather than "implementation" because the high level approach is still under discussion. Pull requests such as PR#485 should be categorized as "implementation" because people agree on the approach and the PR essentially just needs a low level code review/test coverage to be merged in.
> 
> Hopefully good usage of this can get people to discuss their use cases and resolve their differences at the proposal stage and then one party can make a implementation that everyone will be happy with.

dt27182:
I think all 3 categories "Type of Issue", "Development Phase", and "Impact" should be present for both issues and PRs. While not all values are possible for both issues and PRs, just having "Type of Issue" for issues and "Development Phase" and "Impact" for PRs isn't sufficient to describe either.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+**Note: for support questions, you are strongly advised to use Stack Overflow**.
```
mwachs5: Should you provide a link/suggested tag for for Chisel on Stack Overflow?

ducky64:
If you want to keep all the categories in there, can you phrase it in a way that won't cause confusion if it doesn't apply? For example, something like "list categories or link the relevant issue" or "this is applicable when..."

dt27182:
I think the only "Development Phase" has values that don't apply to both issues and PRs. Issues can't be "implementation" and PR's generally can't be "request". Removing those two check boxes from the issues and PRs template respectively should be enough. All the values for "Type of Issue" and "Impact" seem like they apply to both issues and PRs. @ducky64 do you see any other values that don't apply?

ducky64:
Kind of, but there's probably going to be quite a few edge cases. For example, the "impact" field makes sense for a proposal, but might not be obvious during a bug report. I guess the options in "type of issue" makes sense for a PR, but it should be re-labeled as "type of change"?

dt27182:
Ok I see, the impact field might be hard to determine for some thing like a bug fix request. How about just adding a "unknown" options to that field? I think issues still do need to have the impact field because some proposals aren't associated with code and will be filed as issues. 

I agree with the suggestion to relabel "type of issue" as "type of change" for PRs.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+* **If the current behavior is a bug, please provide the steps to reproduce the problem:**
+  * **What is the current behavior?**
+  * **What is the expected behavior?**
+  * **Please tell us about your environment:**
+    (examples)
+    - version: `3.0-SNAPSHOT`
+    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
+
+* **What is the use case for changing the behavior?**
+
```
dt27182: I think **DevelopmentPhase** should still be in here, with "request" and "proposal" as possible values.

ucbjrl:
retest this please

PR #714, opened 2017-11-03
Update InvalidateAPISpec tests.

ucbjrl:
Use DontCare on `outs` not `ins` in InvalidateAPISpec.
Add additional InvalidateAPISpec tests demonstrating selective override of Strict/NotStrict behavior.

### DIFF
--- a/src/test/scala/chiselTests/InvalidateAPISpec.scala
+++ b/src/test/scala/chiselTests/InvalidateAPISpec.scala
@@ -66,13 +66,13 @@ class InvalidateAPISpec extends ChiselPropSpec with Matchers {
     val nElements = 5
     class ModuleWithoutDontCare extends Module {
       val io = IO(new Bundle {
-        val ins = Input(Vec(nElements, Bool()))
+        val outs = Output(Vec(nElements, Bool()))
       })
-      io.ins <> DontCare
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -57,7 +57,7 @@ ifneq (,$(RM_DIRS))
 endif
 
 scaladoc:
-	$(SBT) $(SBT_FLAGS) doc
+	$(SBT) $(SBT_FLAGS) unidoc
```
edwardcwang: What's the effect of this change? Is there a difference between "sbt doc" and "sbt unidoc"?

ucbjrl: `sbt unidoc` generates scaladoc for all subprojects. If you look at the scaladoc generated by `sbt doc`, you'll find it's missing the definitions from coremacros and more importantly, chiselFrontend.

edwardcwang: Good point; would be nice to put this into the commit message.

ucbjrl: I've updated the PR body to mention this.

ucbjrl:
I suppose this really should be broken up into two commits:
 - one for the unidoc change
 - one for the InvalidateAPISpec changes.

edwardcwang:
Seems like a good idea.

PR #718, opened 2017-11-09
BoringUtils / Synthesizable Cross Module References

seldridge:
Placeholder PR for enabling synthesizable cross module references via annotations tied to FIRRTL's WiringTransform. This currently only adds some sample annotators for defining sources and sinks in a hierarchy (which have proven to be good enough primitives for what I'm doing). Feedback is welcome on this API. This depends on an updated WiringTransform that will need to be merged first.

I did not test this explicitly---just lifted the tested/working annotators into this new trait. This is more here so that I remember to get this upstreamed...

This needs:
  - [x] A test
  - [x] Updates to Firrtl's Wiring Transform (PR: freechipsproject/firrtl/pull/648)
  - [x] Bug fix for Firrtl's Euler Tour to support single module circuits (PR: freechipsproject/firrtl/pull/736)
  - [x] Write a custom `ChiselAnnotation` subclass
  - ~Possible name change to something more descriptive, e.g., `HasCrossModuleReferences`~
  - [x] Add don't touch annotation to source

Possibly needs:
  - More tests?


edwardcwang:
ok to test

seldridge:
This should be good to test once freechipsproject/firrtl#736 makes it into a release.

seldridge:
ping @azidar

We can discuss, but this should be good to go. 3.1.1 should be fine as there's no API breakage here. 3.2.0 if you want to put API additions in there.

src/main/scala/chisel3/util/BoringUtils.scala
```diff
@@ -0,0 +1,59 @@
...
+  *   val b = Wire(Bool())
+  *   addSink(b, "unique_identifier")
+  * }
+  * class ModuleC extends Module with BoringUtils {
+  *   val c = Wire(Bool())
+  *   addSink(c, "unique_identifier")
+  * }
+  * }}}
+  */
+trait BoringUtils extends BaseModule {
```
jackkoenig: Since `annotate` is now an object with an apply method, is it necessary to make this be a trait that has to be mixed in by a given module?

src/main/scala/chisel3/util/BoringUtils.scala
```diff
@@ -0,0 +1,59 @@
+// See LICENSE for license details.
+
+package chisel3.util
```
jackkoenig: I suspect this should be marked experimental, what do you think @seldridge?

Tbf we need to do a better job of "promoting" things from experimental which we haven't done yet

src/main/scala/chisel3/util/BoringUtils.scala
```diff
@@ -0,0 +1,59 @@
...
+  */
+trait BoringUtils extends BaseModule {
+  self: BaseModule =>
+
+  /** Add a named source cross module reference
+    *
+    * @param component source circuit component
+    * @param name unique identifier for this source
+    */
+  def addSource(component: NamedComponent, name: String): Unit = {
```
jackkoenig: I've always thought it would be cool if such an API could be a little less Stringly-typed. For example:
```scala
object bore {
  def apply(source: Data, sinks: Data*): Unit = {
    def genName: = source. instanceName // delayed just like .toNamed
    annotate(new ChiselAnnotation with RunFirrtlTransform {
      def toFirrtl = SourceAnnotation(source.toNamed, genName)
      def transformClass = classOf[WiringTransform]
    })
    for (sink <- sinks) {
      annotate(new ChiselAnnotation with RunFirrtlTransform {
        def toFirrtl = SinkAnnotation(sink.toNamed, genName)
        def transformClass = classOf[WiringTransform]
      })
    }    
  }
}
```

jackkoenig: I'm not really sure how to deal with the fact that the names need to be distinct though... There could be a thread-safe datastructure that keeps track? The main issue with that is there is no way to clear it after Chisel elaboration finishes so multiple Chisel invocations could impact each other through that structure. Perhaps Chisel could provide an `endOfElaboration` hook...

seldridge:
I need to address @jackkoenig feedback. I've been away, but should have some time to get this cleaned up this week.

src/main/scala/chisel3/util/BoringUtils.scala
seldridge: Experimental seems fine with me. Where should this live, then? Should it be `package chisel3.experimental` or should it live someplace else and use the `experimental` package object?

jackkoenig: I think maybe a new package `chisel3.util.experimental`? I think we should keep `chisel3.experimental` for things that need to be in core.

seldridge: That seems reasonable to me.

src/main/scala/chisel3/util/BoringUtils.scala
seldridge: The stringly typed nature is annoying, I agree. That was entirely motivated by the inability of (in my understanding) Chisel to support arbitrary cross module references. You'd like to be able to describe, "Connect this source component to this sink component," with a single method. To do that, you need some way of being able to specify a component outside of the current module scope. Otherwise, it appears to me like the annotator sketch above would be limited to doing connections in one module scope (or, provide redundancy with `:=`). Maybe I'm missing something... or some way of talking about specific `Data` objects.

Is there a known way to do this?

Also, the parenthetical above may motivate an actual "bore connector", e.g., `:=>`, `:=b=`, `:=>=`.

I like this annotator and I'm amenable to a singleton managing state. Something like Rocket's elaboration artefacts (and it's British spelling) seems to make sense as a model for how to add hooks.

jackkoenig: Perhaps I am misunderstanding, but with typical Chisel code there's no reason why you *can't* express a cross-module reference. For example:
```scala
class Bar extends Module {
  val io = IO(new Bundle {})
  val myReg = RegInit(8.U)
}
class Foo extends Module {
  val io = IO(new Bundle {})
  val bar = Module(new Bar)
}
class Top extends Module {
  val io = IO(new Bundle { val out = Output(UInt(8.W)) })
  val foo = Module(new Foo)
  bore(foo.bar.myReg, io.out)
}
```
Now of course, you can't access fields if inner scopes or functions, but anything that makes it to a public val can be referenced, which due to reflection-based naming, is obviously how much Chisel code looks!

An obviously if you use `foo.bar.myReg` in `Top` in any *Chisel* operations, it will error, but you can certainly do such a cross-module annotation for the purposes of BoringUtils

seldridge: You're right! I don't know why that didn't occur to me. All the `val`s are public. Let me think about this... This pattern would make it so that the boring has to happen in the lowest common ancestor module or higher. I've always assumed that you'd want to annotate at the component. Cross module references wouldn't work there, but there may be a non-stringly typed way...

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -39,7 +40,9 @@ object BoringUtils {
           def toFirrtl = SourceAnnotation(component.toNamed, name)
           def transformClass = classOf[WiringTransform] },
         new ChiselAnnotation {
-          def toFirrtl = DontTouchAnnotation(component.toNamed) })
+          def toFirrtl = DontTouchAnnotation(component.toNamed) },
+        new ChiselAnnotation {
+          def toFirrtl = NoDedupAnnotation(component.toNamed.module) })
```
jackkoenig: I would think we would still want deduplication for Modules that all have the same boring annotations right? Given that we don't really have a way to distinguish NoDedup to YouCanDedupSomeOfUsTogether though, I think that's follow on work

seldridge: I modified the test case (just now) to show what I perceive as the issue with needing deduplication. If you have two instances of a source and you only want to wire one, then without deduplication they look the same. That manifests as a `WiringException` complaining that a source to sink mapping can't be determined due to multiple sources being equidistant to the same sink. The same problem exists for sinks. You wind up wiring all modules as opposed to just the one you want.

The use of `addSource`/`addSink` doesn't manifest that way typically from a usability perspective (though it still can). A user would have to mixin a trait to a module or do some `{ ... }` stuff when constructing the module (which most people won't do). This only occurred to me when I started playing with your `bore` object example.

Any better way to handle this? (Other than instance annotations...) Am I seeing a non-issue here?

seldridge:
@jackkoenig: this is good for another look. A few concerns are below. Notes about this are included in the `BoringUtils` documentation:

- The `BoringUtils` namespace persists across elaborations. While this is expected of global, mutable state... namespace persistence can result in non-determinism in boring names. You can see this if you elaborate a circuit that uses `BoringUtils.bore` twice, e.g., bore with desired name `x` you'll get `x` the first time and `x_1` the second time. Functionally, the names are irrelevant, but it's clearly wrong. I think this motivates an RFC discussing elaboration hooks (pre-elaboration, post-elaboration), driver state registration, and state encapsulation via a `DynamicVar` or similar (which I think matters for parallel elaboration).
- `addSource` and `addSink` can be dangerous, but it's an API that's useful if you want to do  connections with minimal modifications. One example is connecting two RoCCs. `addSink`/`addSource` let me do this without a `bore` in `RocketTileModuleImp` or with any concern for the hierarchy through which I'm boring. Not insurmountable, but it enables easier use by libraries.

PR #719, opened 2017-11-13
Fix chisel3.Driver example documentation

seldridge:
The chisel3.Driver example was mixing in case classes as opposed to the traits that were (assumedly) intended. This corrects that.


edwardcwang:
ok to test

ucbjrl:
ok to test

PR #720, opened 2017-11-21
Change switch to emit when, elsewhen, instead of when, when

jackkoenig:
While theoretically emitting independent whens for each "is" *could* make better hardware, Firrtl currently just emits a bad mux tree with lots of coverage holes. This change matches common use where each "is" is mutually exclusive and users would prefer then same result as when, elsewhen. 

@terpstra

### DIFF
--- a/src/main/scala/chisel3/util/Conditional.scala
+++ b/src/main/scala/chisel3/util/Conditional.scala
@@ -23,22 +23,21 @@ object unless {  // scalastyle:ignore object.name
 /** Implementation details for [[switch]]. See [[switch]] and [[chisel3.util.is is]] for the
   * user-facing API.
   */
-class SwitchContext[T <: Bits](cond: T) {
-  private var whenContext: Option[WhenContext] = None
-  def is(v: Iterable[T])(block: => Unit) {
+class SwitchContext[T <: Bits](whenContext: Option[WhenContext], cond: T) {
+  def is(v: Iterable[T])(block: => Unit): SwitchContext[T] = {
     if (!v.isEmpty) {
       // def instead of val so that logic ends up in legal place
       def p = v.map(_.asUInt === cond.asUInt).reduce(_||_)
       whenContext match {
-        case Some(w) =>
-          whenContext = Some(w.elsewhen(p)(block))
-        case None =>
-          whenContext = Some(when(p)(block))
+        case Some(w) => new SwitchContext(Some(w.elsewhen(p)(block)), cond)
+        case None => new SwitchContext(Some(when(p)(block)), cond)
       }
+    } else {
+      this
     }
   }
-  def is(v: T)(block: => Unit) { is(Seq(v))(block) }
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/SwitchSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import chisel3._
+import chisel3.util._
+import chisel3.testers.BasicTester
+
+class MultiSwitchFireTester extends BasicTester {
+  val wire = WireInit(0.U)
+  switch (0.U) {
+    is (0.U) { wire := 1.U } // Only this one should happen
+    is (0.U) { wire := 2.U }
```
sdtwigg: Oddly, this test immediately demonstrates that the new switch has different behavior than the old switch. The old one would run overlapping when conditions (and the last one would win). In the case where 'is' are with literals, this shouldn't change behavior (unless sloppy code) but could be problematic if you ever used a non-literal in the is block.

src/main/scala/chisel3/util/Conditional.scala
```diff
@@ -24,10 +24,16 @@ object unless {  // scalastyle:ignore object.name
   * user-facing API.
   */
 class SwitchContext[T <: Bits](cond: T) {
+  private var whenContext: Option[WhenContext] = None
```
sdtwigg: is it possible to fold over all the ises without tracking mutable state (for an indeterminate time)?

src/test/scala/chiselTests/SwitchSpec.scala
jackkoenig: Maybe I wasn't clear in the message . This is intentionally changing behavior and I wrote the test this way because it demonstrated the change.

jackkoenig: Although I mucked up the message below, I'll fix that

jackkoenig:
Alright I've changed it to be immutable (although that involves changing SwitchContext, doubtful anyone instantiates that directly but it is a public API), and enforce only mutually exclusive literals are allowed as is conditions.

I am less convinced that this is an API change we *need* to make. The original motivation was to close coverage holes in the emitted Verilog, but changing this only solves one (admittedly) common case. A more general solution would be to write a Firrtl transform to do some mux restructuring based on mutually exclusive cases. I think it wouldn't be that hard to do and would solve a greater number of coverage holes in Chisel designs.

There is still an argument for this API change based on the fact that switch is weird and it really doesn't seem like it should be allowing multiple cases to fire. While Verilog cases *do* allow multiple cases to fire, it's because of wildcards in the comparisons that switch currently does not support (we should consider allowing BitPat). We can introduce extensions to switch (like maybe switch.priority and switch.parallel) to better match expected Verilog semantics and be more explicit about what the behavior is. Note that if this PR is accepted, we could map switch to parallel cases in Verilog as an optimization (and we don't risk the same simulation-synthesis mismatch that Verilog can have).

jackkoenig:
@ducky64 can you take another quick look at this? I made some updates since your review

PR #723, opened 2017-11-23
Auto clone type 🦆

ducky64:
Update: the IO cloning change will be in a separate PR, and has been removed from this one.

Two changes for the PR of one!
- ~~Changes the behavior of IO(...) to return a clone of the input, consistent with all other binding operations. In Chisel compatibility mode, the original object is bound in-place (mutably) and returned - this is necessary for compatibility.~~
- Extend automatic cloneType to handle cases where the constructor has arguments that are all `val`s (this is necessary both for a guarantee of immutability, and for the arguments to be object fields that can be accessed by reflection).

Major backwards incompatible changes:
- ~~All Bundles need to have their arguments be vals (so auto clone type works). Previously, Bundles only used in IO(...) didn't need cloneType, because IO(...) didn't clone inputs.~~

Some issues with autoclonetype, which are not encountered in practice:
- It will fail on cloning a instance of Bundle (not a Bundle subclass) because of the implicit compileOptions. But one shouldn't be instantiating a Bundle directly anyways, that makes no sense. Workaround: define an empty Bundle subclass.
- It cannot handle anonymous inner classes with arguments (is this even possible?), or Bundles in an anonymous outer class. Both of these are due to limitations in Scala reflection, and none are encountered in practice. Java reflection handles the cases where there are no arguments.
- cloneType no longer tries to invoke constructors with all null arguments. I'm not convinced that is a good idea at all, especially since this can result in some very unintuitive null pointer errors.
- cloneType outer class instance detection is very heuristic, and only works on Modules. In particular, it will not work on Bundles defined inside objects.

This hasn't been tested on rocket-chip yet, someone probably should test it. IO(...) behavior is not changed in compatibility mode, but cloneType changes (in particular, not falling back to invoking everything with nulls) might break things in compatibility - though this potentially could be addressed with compileOptions if needed. Additionally, chisel3 code may need to be updated, hopefully just defining constructor arguments as vals. Which probably should have been done in the first place, since you need immutability guarantees to clone things reliably.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -5,7 +5,6 @@ package chisel3.core
 import scala.collection.immutable.ListMap
 import scala.collection.mutable.{ArrayBuffer, HashSet, LinkedHashMap}
 import scala.language.experimental.macros
-import scala.util.Try
 
 import chisel3.internal._
 import chisel3.internal.Builder.pushCommand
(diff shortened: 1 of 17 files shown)


ucbjrl:
Great job @albert-magyar and @ducky64. I'm concerned about such a significant change at this late date but I agree it does significantly improve the interface.

It fails the rocket-chip regression tests:
```bash
[error](run-main-0) java.lang.reflect.InvocationTargetException
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at freechips.rocketchip.util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:52)
	at freechips.rocketchip.util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:49)
	at chisel3.core.Module$.do_apply(Module.scala:45)
	at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:92)
	at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:92)
	at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:244)
	at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:242)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
	at chisel3.internal.Builder$.build(Builder.scala:242)
	at chisel3.Driver$.elaborate(Driver.scala:92)
Caused by: java.lang.Exception: Unable to use ParamaterizedBundle.cloneType on class freechips.rocketchip.rocket.HellaCacheBundle, probably because class freechips.rocketchip.rocket.HellaCacheBundle() takes more than one argument.  Consider overriding cloneType() on class freechips.rocketchip.rocket.HellaCacheBundle
```

ducky64:
Ah, that's (probably) a complicated issue. My guess to what's happening is that the IO(...) was being invoked in a chisel3._ context, and under new (cloning) semantics, but the Bundle was defined in a Chisel._ context, which doesn't expect new semantics. (the stack trace doesn't seem to provide enough information: the generator class is shown, but not the Module going through it). Further complicating things is that rocket-chip src/main/scala/util/Misc.scala defines a ParameterizedBundle which overrides cloneType with its own version, which isn't able to handle arguments.

ccelio:
> Backwards incompatible changes:

> All Bundles need to have their arguments be vals (so auto clone type works). Previously, Bundles only used in IO(...) didn't need cloneType, because IO(...) didn't clone inputs.

What does that mean? Is this suggesting that stuff like this is no longer possible?

https://github.com/ucb-bar/riscv-boom/blob/e71168e1b6412a1df85eedcabe2eecbae40bcf11/src/main/scala/tage-table.scala#L87-L91

https://github.com/ucb-bar/riscv-boom/blob/e71168e1b6412a1df85eedcabe2eecbae40bcf11/src/main/scala/tage-table.scala#L104-L123

Hwacha has other examples too.

ducky64:
I don't think that would be an issue.
The difference would be code like this:
```
val myBundleChiselType = new Bundle() {...}
val io = IO(myBundleChiselType)
```

Under old semantics, both `myBundleChiselType ` and `io` would refer to the same object, and IO(...) effectively mutates `myBundleChiselType` into a hardware object

Under new semantics, they would be unique objects, `myBundleChiselType` continues to refer to a chisel type, while `io` is a clone of that, which is bound to IO hardware type.

Under both cases, you can continue updating (through chisel assign) `io`.

ucbjrl:
@ducky64, I didn't provide the complete stack trace. Check the failing job's console [output](http://a17.millennium.berkeley.edu/jenkins/job/mc-rocket-chip/242/SUITE=RocketSuiteA,slave=ferry/console).

albert-magyar:
Correct me if I'm wrong, @jackkoenig, but this puts us in a quandary as we _cannot fix_ this rocket-chip error without changing the rocket-chip codebase or avoiding cloning in `IO`.

Philosophically, this error is only arising because rocket-chip is using the unfrozen pre-release chisel3 API, so shouldn't code changes be acceptable? Otherwise, we're stuck in a corner forcing compatibility with pre-release chisel3.

This rocket-chip error is not related to the auto `cloneType` implementation. As @ducky64 pointed out, this is intrinsic to the fact that `IO` now calls `cloneType`, which is overridden by an implementation in rocket-chip that is not general enough to clone bundles that previously didn't need to be cloned.

ucbjrl:
rocket-chip may not be the only client with this problem. We need to come up with a migration plan for existing clients (something that will hopefully work with the current (poor) API and the new one), and give people an opportunity to implement it.

I think we need to give some warning that an API that's been in use for over three years is about to change, albeit for the better.

Consider 3.0.0 the **last** major release to support non`val`s in an `IO()` bundle.

albert-magyar:
We should at least get auto `cloneType` in for the release if possible,
even if `IO` doesn't clone

ucbjrl:
In principle I agree, but I'd rather not hold up today's release. The auto
cloneType implementation isn't API impacting. There's nothing that prevents
us releasing 3.0.1 next week with auto cloneType plus whatever fixes are
prompted by the 3.0.0 release.

In fact, I would argue that having a stream of non-API impacting
improvements solidifies Chisel's credibility.

ducky64:
Added support for second parameter list (so for the rocket example, eliminating the custom limited clonetype and making all the parameters vals should work) and dropped the cloning in IO (which will come in a separate PR - though this fixes all Bundles in tests to be autoclonetype comaptible). API breakages should be much more limited, if any.

Does anyone want to test this with rocket again?

ucbjrl:
Passes rocket-chip regression [tests](http://a17.millennium.berkeley.edu/jenkins/job/mc-rocket-chip/245/).

albert-magyar:
Did you test with the IO cloning behavior enabled?

ducky64:
Yes, it passes chisel regressions with IO cloning

ducky64:
So one question is going to be what's the best way to deal with bundle class hierarchies? For example:
```scala
class BaseBundle(i: Int) extends Bundle {
  val inner = UInt(i.W)
}
class SubBundle(val i: Int, val i2: Int) extends BaseBundle(i) {
  val inner2 = UInt(i2.W)
}
```
is one way that could work, but this involves making everything in the leaf class a val. Another way would be:
```scala
class BaseBundle(val i: Int) extends Bundle {
class SubBundle(i: Int, val i2: Int) extends BaseBundle(i) {
```
where only the base class needs to have it as a val, and the leaf class is just passing through arguments. This seems nicely deduplicated and clean, though the nonparallel structure (between non-val `i` and val `i2` in the SubBundle constructor would look odd). Note: doesn't work with current autoclonetype implementation because it needs to explicitly look for accessors in base classes.

The scala compiler would complain about this:
```scala
class BaseBundle(val i: Int) extends Bundle {
class SubBundle(val i: Int, val i2: Int) extends BaseBundle(i) {
```
because there's a name collision without overriding, and for
```scala
class BaseBundle(val i: Int) extends Bundle {
class SubBundle(override val i: Int, val i2: Int) extends BaseBundle(i) {
```
it's unclear what is what, especially doing something really odd like:
```scala
class BaseBundle(val i: Int) extends Bundle {
class SubBundle(override val i: Int, val i2: Int) extends BaseBundle(i+1) {
```
It turns out that SubBundle instance `i` has value 8, but `inner` has width 9, so this is something that should be avoided.

azidar:
Support all cases, but encourage declaring val in super class constructor (aka code block 2).

ducky64:
Added support for those cases, this is now pretty much done and just waiting on code review.
IO(...) change will follow separately in another PR.

One interesting thing to measure would be whether this has an adverse effect on elaboration runtime - if so, it should be possible to remember the clone function for a particular type. But that should be driven by profiling results, and be part of another PR.

ducky64:
@jackkoenig poke ... and remember that the IO change needs to follow this, and that's even more likely to break code ...

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+    val outerClassInstance = Option(clazz.getEnclosingClass) match {
+      case Some(outerClass) =>
+        val outerInstance = try {
+          clazz.getDeclaredField("$outer").get(this)  // doesn't work in all cases, namely anonymous inner Bundles
         } catch {
-          case _: java.lang.reflect.InvocationTargetException | _: java.lang.IllegalArgumentException =>
-            Builder.exception(s"Parameterized Bundle ${this.getClass} needs cloneType method. You are probably using " +
-              "an anonymous Bundle object that captures external state and hence is un-cloneTypeable")
-            this
+          case _: NoSuchFieldException => this.outerModule match {
```
jackkoenig: alternatively, 
```scala
this.outerModule.getOrElse(reflectError(...))
```

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+            case Some(outerModule) => outerModule
+            case None => reflectError(s"Unable to determine instance of outer class $outerClass")
+          }
+        }
+        if (!outerClass.isAssignableFrom(outerInstance.getClass)) {
+          reflectError(s"Automatically determined outer class instance $outerInstance not assignable to outer class $outerClass")
```
jackkoenig: I'm having trouble seeing when this occurs, is there a test for it?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+          } else {
+            None
+          }
+        case _ => None
+      }
+      clone match {
+        case Some(clone) =>
+          clone.outerModule = this.outerModule
+          if (!clone.typeEquivalent(this)) {
+            reflectError(s"Automatically cloned $clone not type-equivalent to base $this." +
```
jackkoenig: Is there a test case for this?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+            " Constructor argument values were not inferred, ensure constructor is deterministic.")
+          }
+          return clone.asInstanceOf[this.type]
```
jackkoenig: Is it possible to factor this so that there isn't a return here?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+        case None =>
+      }
+    }
+
+    // Get constructor parameters and accessible fields
+    val mirror = runtimeMirror(clazz.getClassLoader)
+    val classSymbol = try {
+      mirror.reflect(this).symbol
+    } catch {
+      case e: scala.reflect.internal.Symbols#CyclicReference => reflectError(s"Scala cannot reflect on $this, got exception $e." +
```
jackkoenig: Is there a test case for this?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+          " This is known to occur with inner classes on anonymous outer classes." +
+          " In those cases, autoclonetype only works with no-argument constructors, or you can define a custom cloneType.")
+    }
+
+    val decls = classSymbol.typeSignature.decls
+    val ctors = decls.collect { case meth: MethodSymbol if meth.isConstructor => meth }
+    if (ctors.size != 1) {
+      reflectError(s"found multiple constructors ($ctors)." +
```
jackkoenig: Test case for this as well?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+    val ctorParams = ctorParamss match {
+      case Nil => List()
+      case ctorParams :: Nil => ctorParams
+      case ctorParams :: ctorImplicits :: Nil => ctorParams ++ ctorImplicits
+      case _ => reflectError(s"internal error, unexpected ctorParamss = $ctorParamss")
     }
+    val ctorParamsNames = ctorParams.map(_.name.toString)
+
+    // Special case for anonymous inner classes: their constructor consists of just the outer class reference
+    // Scala reflection on anonymous inner class constructors seems broken
```
jackkoenig: Does Scala reflection work for this in 2.12?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+    // (which could lead to data conflicts, since it's likely the user didn't know to re-bind them).
+    // This is not guaranteed to catch all cases (for example, Data in Tuples or Iterables).
+    val boundDataParamNames = ctorParamsNameVals.map{ case (paramName, paramVal) => paramVal match {
+      case paramVal: Data => paramVal.hasBinding match {
+        case true => Some(paramName)
+        case false => None
+      }
+      case _ => None
+      }
+    }.filter(_.isDefined).map(_.get)
```
jackkoenig: This whole block could be better written as:
```scala
    val boundDataParamNames = ctorParamsNameVals.collect {
      case (paramName, paramVal: Data) if paramVal.hasBinding => paramName
    }
```
What a great example of when to use collect 😁

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -557,30 +557,168 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+    if (!boundDataParamNames.isEmpty) {
```
jackkoenig: !x.isEmpty == x.nonEmpty

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ducky64: Unsure, but it appears we're still building with (or supporting) 2.11, so  the code has to stay in?

jackkoenig: Oh yeah I wasn't suggesting removing it, I was just curious

jackkoenig:
Updates look good. Maybe you answered this and I just missed it but is it possible to support direct cloning of [anonymous inner Bundles were the outer class is a Bundle](https://github.com/sifive/sifive-blocks/blob/9052a079d404ebbfda5f01765b909c20503504ad/src/main/scala/devices/spi/SPIMedia.scala#L12)? I guess I've only come across one case of this being a problem in the wild, but it is something that currently works.

ducky64:
I don't think you can clone an anonymous inner sub-bundle directly (in your example, doing `chiselTypeOf(io.link.cs)` would likely fail), though you would be able to do `chiselTypeOf(io.link)` because the constructor of the outer type handles the construction of the inner type, so there is no need to directly clone the inner type.

The old version only worked because it tries a `null` for the outer instance reference, and most use cases were lucky in that they don't ever reference the outer instance. We theoretically could keep that behavior, but I don't think having a null lurking around internally is a good idea.

With Bundles, it may also be possible to get the enclosing Bundle if it's a bound type, and to see if that fits. But that's quite a few restrictions...

jackkoenig:
Great work @ducky64 and @albert-magyar!

ucbjrl:
This breaks `MissingCloneBindingExceptionSpec` when compiled with Scala 2.12.4:
```
[info] MissingCloneBindingExceptionSpec:
[info] chiselTests.MissingCloneBindingExceptionSpec *** ABORTED ***
[info]   "Unable to automatically infer cloneType on chiselTests.MissingCloneBindingExceptionSpec$TestTop$2$$anon$1@7: constructor has parameters (Set(x$1)) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method." did not include substring "needs cloneType method" (MissingCloneBindingExceptionSpec.scala:31)
[info] ScalaTest
[info] Run completed in 624 milliseconds.
[info] Total number of tests run: 0
[info] Suites: completed 0, aborted 1
[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
[info] *** 1 SUITE ABORTED ***
[error] Error: Total 1, Failed 0, Errors 1, Passed 0
[error] Error during tests:
[error] 	chiselTests.MissingCloneBindingExceptionSpec
[error] (chisel/test:testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 58 s, completed Jan 5, 2018 11:30:20 AM
```
The breakage occurs with the initial commit 7c3c18de2ffd56af51b99030c7ae7d3a321aed5f.

ducky64:
How are you running the 2.12.4 test?
Everything appears to pass on just master?

ucbjrl:
Any one of:
```bash
% sbt "++2.12.4" clean "testOnly chiselTests.MissingCloneBindingExceptionSpec"
% sbt "++2.12.4" clean test
% sbt +clean +test
```
in the master branch should demonstrate the problem.

ducky64:
I can't reproduce, because everything is just refusing to work on 2.12.4 with errors like these:
```
[info]   java.lang.NoClassDefFoundError: Could not initialize class chisel3.testers.TesterDriver$
[info]   at chiselTests.ChiselRunners.runTester(ChiselSpec.scala:23)
[info]   at chiselTests.ChiselRunners.runTester$(ChiselSpec.scala:22)
```

Also, I have no idea why that test is failing, `"needs cloneType method"` isn't even in the code anymore.
https://github.com/freechipsproject/chisel3/blob/master/src/test/scala/chiselTests/MissingCloneBindingExceptionSpec.scala

ucbjrl:
Hmmm. I didn't check the actual error message clearly enough, and indeed the report has changed:
```bash
[info] MissingCloneBindingExceptionSpec:
[info] chiselTests.MissingCloneBindingExceptionSpec *** ABORTED ***
[info]   "Unable to automatically infer cloneType on chiselTests.MissingCloneBindingExceptionSpec$TestIO$1@b: Unable to determine instance of outer class class chiselTests.MissingCloneBindingExceptionSpec, guessed chiselTests.MissingCloneBindingExceptionSpec$Test$2@6, but is not assignable to outer class class chiselTests.MissingCloneBindingExceptionSpec" did not include substring "make all parameters immutable" (MissingCloneBindingExceptionSpec.scala:31)
```
I can reproduce this on OSX 10.12.6 and Ubuntu 4.4.0-92-generic. I'll dig further.

ucbjrl:
It looks like outer class/instance access is sufficiently different in Scala 2.12 that it thwarts `cloneType`'s attempts to determine the outer class for the purpose of cloning a Chisel type, at least for this case.

ducky64:
Yeah, my guess is that the `outer` variable is no longer being set. In any case, the structure isn't common, so I think I'll change the structure to be more common (move bundle definition into the module)

PR #724, opened 2017-11-27
Fixed X-pessimism in RRArbiter

pentin-as:
I launch Rocket without randomize. Virtual memory tests fall due to X-Pessimism in RRArbiter. If I do this correction, it works.


pentin-as:
Yes.  I added the init value to one of the registers. It's all.

stevenmburns:
You might want to look at the discussion in [#267](https://github.com/freechipsproject/chisel3/pull/267)

PR #731, opened 2017-12-08
Add WireInit.apply that accepts DontCare

### LINKED ISSUE #728: DontCare Influences Type Inference
Consider the following:
```scala
class DontCareTest extends Module {
  class Bun extends Bundle {
    val foo = UInt(8.W)
  }
  val io = IO(new Bundle {
    val uint = Output(UInt())
    val bun = Output(Valid(new Bun))
  })
  val uint = WireInit(UInt(8.W), DontCare)
  io.uint := uint

  val bun = WireInit(Valid(new Bun), DontCare)
  bun.valid := false.B
  io.bun := bun
}
```

This gives the following error:
```
[info] Compiling 1 Scala source to /Users/jack.koenig3/ucb-bar/chisel3-raw/target/scala-2.12/classes...
[error] /Users/jack.koenig3/ucb-bar/chisel3-raw/src/main/scala/DontCare.scala:16: value valid is not a member of chisel3.core.Data
[error]   bun.valid := false.B
[error]       ^
[error] one error found
```

The definition of `WireInit.apply` being:
```scala
def apply[T <: Data](t: T, init: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T
```

Now changing the type of `init` to `Data` should fix this particular instance but this problem will pop up all over the place, including in libraries. Is there a way to tell the type system to ignore `DontCare` for type inference? @ducky64 @sdtwigg

jackkoenig:
Prevents DontCare from affecting type inference
Fixes #728 

Just making `init` be of type `Data` would lose some static type safety, eg. `WireInit(new MyBundle, 0.U)` fails statically with this PR but wouldn't with changing the type of `init` to `Data`.


src/main/scala/chisel3/package.scala
```diff
@@ -32,6 +32,10 @@ package object chisel3 {    // scalastyle:ignore package.object.name
     @deprecated("Wire(t, init) is deprecated, use WireInit(t, init) instead", "chisel3")
     def apply[T <: Data](t: T, init: T)(implicit compileOptions: CompileOptions): T =
       chisel3.core.WireInit(t, init)
+
+    @deprecated("Wire(t, init) is deprecated, use WireInit(t, init) instead", "chisel3")
```
ducky64: I don't think it's necessary to maintain compatibility in this case - DontCare came after this so I doubt there's legacy code where this is important.

jackkoenig: That's fair, I agree with not adding this "compatibility"

edwardcwang: At the same time, this can help reduce surprise for people looking at legacy examples of code.

jackkoenig:
I'm not *adding* extra static safety, I'm merely preserving the safety that is already there. If you try:
```scala
val x: UInt = WireInit(UInt(4.W), 2.S)
```
This gives a compiler error on master and on this implementation but if init is changed to type Data then it will fail at runtime instead.

ucbjrl:
retest this please

ducky64:
I think the special-casing of the DontCare construct is a "code smell" - it doesn't generalize if we add more of these. That being said, it's probably fine as far as the user-visible API goes, and any refactoring / cleanup / expansion in the future would need to handle this case well anyways.

jackkoenig:
@ducky64 So you think this is okay? Mind giving it a review?

jackkoenig:
Yeah if we want to remove the duplication and make init just be of type Data, there's no API change so it's nbd. We should definitely consider that if/when we add more things like `DontCare` though.

PR #747, opened 2017-12-20
Add compileOptions to Module.apply, use for invalidating submod ports

### LINKED ISSUE #746: Submodule ports not properly invalidating in compatibility code
Instantiating an Arbiter (and possibly chisel3 modules in general) inside of compatibility code can lead to the ports of the submodule not being properly invalidated.

* **Type of issue**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **If the current behavior is a bug, please provide the steps to reproduce the problem:**
    (examples)
    - version: `3.0-SNAPSHOT`
    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`

Consider:
```scala
import Chisel._

class InvalidateBugTop extends Module {
  val io = IO(new Bundle {
    val in = Vec(2, Decoupled(UInt(32.W)).flip())
    val out = Decoupled(UInt(32.W))
  })
  val arb = Module(new Arbiter(UInt(32.W), 2))
  arb.io.in <> io.in
  io.out <> arb.io.out
}


object InvalidateBugDriver extends App {
  chisel3.Driver.execute(args, () => new InvalidateBugTop)
}
```

The resulting Firrtl does not have an invalidate on the io of the Arbiter:
```
  module InvalidateBugTop :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : {flip ready : UInt<1>, valid : UInt<1>, bits : UInt<32>}[2], out : {flip ready : UInt<1>, valid : UInt<1>, bits : UInt<32>}}

    clock is invalid
    reset is invalid
    io is invalid
    inst arb of Arbiter @[InvalidateBug.scala 9:19]
    arb.clock <= clock
    arb.reset <= reset
    arb.io.in <- io.in @[InvalidateBug.scala 10:13]
    io.out <- arb.io.out @[InvalidateBug.scala 11:10]
```
Combined with https://github.com/freechipsproject/firrtl/pull/706, this can emit incorrect Verilog (working on an example)

Fix expected behavior in compatibility mode

* **Impact**
  - [x] no functional change
  - [ ] API modification
  - [ ] unknown

* **Development Phase**
  - [x] request
  - [ ] proposal

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

Strangely, I tried making a similar example with a Queue as the submodule instead of an Arbiter, but it correctly invalidated the Queue's io.

jackkoenig:
Fixes #746
Also add test for https://github.com/freechipsproject/firrtl/issues/705

Basically, for invalidating ports of a submodule, we need to use the compile options for where the module is constructed, not defined. I added this to Module.apply and used it in initializeInParent

* **Type of change**
  - [x] Bug ~~report~~ **fix**
  - [ ] Feature request
  - [ ] Other enhancement

* **Impact**
  - [x] no functional change
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

Fix bug where chisel3.Modules instantiated in Chisel.Modules would not have their inputs properly invalidated

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
@@ -80,7 +80,7 @@ abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Para
     component
   }
 
-  private[core] def initializeInParent() {
+  private[core] def initializeInParent(instanceCompileOptions: CompileOptions): Unit = {
     implicit val sourceInfo = UnlocatableSourceInfo
 
     for (x <- getModulePorts) {
@@ -165,7 +165,7 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     component
   }
 
-  private[core] def initializeInParent() {
+  private[core] def initializeInParent(instanceCompileOptions: CompileOptions): Unit = {
     for ((_, port) <- io.elements) {
       pushCommand(DefInvalid(UnlocatableSourceInfo, port.ref))
     }
--- a/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
@@ -80,7 +80,7 @@ abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Para
     component
   }
 
-  private[core] def initializeInParent(instanceCompileOptions: CompileOptions): Unit = {
+  private[core] def initializeInParent(parentCompileOptions: CompileOptions): Unit = {
     implicit val sourceInfo = UnlocatableSourceInfo
 
     for (x <- getModulePorts) {
@@ -165,7 +165,7 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     component
   }
 
-  private[core] def initializeInParent(instanceCompileOptions: CompileOptions): Unit = {
+  private[core] def initializeInParent(parentCompileOptions: CompileOptions): Unit = {
     for ((_, port) <- io.elements) {
       pushCommand(DefInvalid(UnlocatableSourceInfo, port.ref))
     }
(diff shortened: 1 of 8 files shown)


chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
```diff
@@ -80,7 +80,7 @@ abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Para
     component
   }
 
-  private[core] def initializeInParent() {
+  private[core] def initializeInParent(instanceCompileOptions: CompileOptions): Unit = {
```
ducky64: Would it make more sense to call this parentCompileOptions? (and similarly elsewhere where this is used)

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
```diff
@@ -80,7 +80,7 @@ abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Para
...
     implicit val sourceInfo = UnlocatableSourceInfo
 
     for (x <- getModulePorts) {
```
ducky64: Is invalidating black boxes necessary?

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
jackkoenig: yeah I think parentCompileOptions is better, I'll change it

chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
jackkoenig: I'm not sure, I didn't change any of the BlackBox behavior in this PR, we should discuss for 3.1 though

PR #751, opened 2017-12-23
Memoize autoclonetype

jackkoenig:
**Note the target of this PR is autoclonetype**

Improves performance and allows potentially non-autoclonetypeable clones
of clonetypeble parents to be cloned.

The current autoclonetype implementation does not support cloning anonymous inner bundles of other bundles. I think this is actually an important thing to support since clonetype is called when doing things like muxing, `WireInit` or `RegNext`.

When bound, inner bundles have `ChildBinding` which points to the parent. This provides us with the correct outer instance. The problem is that when you call WireInit or any of that, the bound Bundle is cloned, and then cloned again! The second cloning fails because the cloned instance is not bound. However, if we memoize cloneType, that actually makes it possible to clone it anyway. This does mean that each clone of a cloned Bundle actually uses references from the original one, but it should give the same result. This has the nice side effect of avoiding lots of unnecessary reflection and it should improve performance.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -556,15 +556,32 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     case _ => None
   }
 
-  override def cloneType : this.type = {
+  // Memoize autoclonetype
+  private var _cloner: Option[() => this.type] = None
+
+  private def reflectError(desc: String): Nothing = {
+    Builder.exception(s"Unable to automatically infer cloneType on $this: $desc").asInstanceOf[Nothing]
+  }
+
+  override def cloneType : this.type = _cloner match {
+    case Some(f) =>
+      val clone = f()
+      if (!clone.typeEquivalent(this)) {
+        reflectError(s"Automatically cloned $clone not type-equivalent to base $this." +
+        " Constructor argument values were not inferred, ensure constructor is deterministic.")
+      }
+      clone._cloner = _cloner
+      clone.outerModule = this.outerModule // TODO Is this right???
+      clone
+    case None =>
+      _cloner = Some(autocloneType)
+      cloneType
+  }
+
+  private def autocloneType: () => this.type = {
     // This attempts to infer constructor and arguments to clone this Bundle subtype without
     // requiring the user explicitly overriding cloneType.
     import scala.language.existentials
-
-    def reflectError(desc: String): Nothing = {
-      Builder.exception(s"Unable to automatically infer cloneType on $this: $desc").asInstanceOf[Nothing]
-    }
-
     import scala.reflect.runtime.universe._
 
     // Check if this is an inner class, and if so, try to get the outer instance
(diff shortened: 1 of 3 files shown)


jackkoenig:
One big question I have: `clone.outerModule = this.outerModule` is used throughout autoclonetype, is this correct? What if you call `WireInit` on a port of a submodule? Those Bundles should have different outerModules right?

ducky64:
outerModule is meant to track a guess at the outer class of an inner class bundle, so any necessary external state is captured and kept consistent during a clone. So all clones should have the same outerModule.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -702,16 +712,17 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
-    if (!clone.typeEquivalent(this)) {
-      reflectError(s"Automatically cloned $clone not type-equivalent to base $this." +
-          " Constructor argument values were inferred: ensure that variable names are consistent and have the same value throughout the constructor chain," +
-          " and that the constructor is deterministic.")
+    val constructor = classMirror.reflectConstructor(ctor)
+    () => {
+      val clone = constructor.apply(ctorParamsVals:_*).asInstanceOf[this.type]
+
+      if (!clone.typeEquivalent(this)) {
+        reflectError(s"Automatically cloned $clone not type-equivalent to base $this." +
```
ducky64: Don't you handle the error messages in the cloneType (instead of the autoCloneType) function? I think this will still do the right thing (this message will take priority because it fires first, as it should), but it's not great from a code structure / separation of concerns standpoint

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -574,7 +591,10 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
         clazz.getDeclaredField("$outer").get(this)  // doesn't work in all cases, namely anonymous inner Bundles
       } catch {
         case _: NoSuchFieldException =>
-          this.outerModule.getOrElse(reflectError(s"Unable to determine instance of outer class $outerClass"))
+          // If this Bundle has a ChildBinding, it's an anonymous inner Bundle of another Bundle
+          this.bindingOpt.collect { case ChildBinding(p) => p }
```
ducky64: Are there cases where this static priority system gets it wrong? For example, a Bundle that is directly a Module inner class but is also a child of another Bundle, so it would think the outer class is a Bundle instead of (properly) the Module? Code written that way might be a bit odd, but not that odd...

jackkoenig:
The primary motivation for memoization wasn't for the performance improvement but actually to support cloning anonymous inner bundles. For example:
```scala
class BundleWithAnonymousInner extends Bundle {
   val inner = new Bundle {
     val in = UInt(8.W)
   }
 }
  val io = IO(new Bundle {
    val cond = Input(Bool())
    val in = Input(Vec(2, new BundleWithAnonymousInner))
    val out = Output(new BundleWithAnonymousInner)
  }
  out.inner := Mux(io.cond, io.in(0).inner, io.in(1).inner)
}
```
This works currently but fails in the current `autoclonetype` branch. My code above that uses the ChildBinding to determine the outer object for cloning the anonymous inner bundle which works. The problem is this freshly cloned bundle then gets cloned again and since it's unbound it doesn't have the ChildBinding to determine the correct outer object. I saw memoization as a way to fix that, but instead I could use the `outerModule` pattern.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
jackkoenig: That's a good point, I had waffled on whether to go static priority or not but it seems it probably shouldn't be. Looking into how to fix.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
jackkoenig: I'll think about how to factor this in the best way

ducky64:
Yeah, I think memoization is more difficult to do well (from a code structure view - it might make sense later, as a feature expansion), but storing and propagating an outer reference seems more simple.

jackkoenig:
See #753 instead

PR #753, opened 2017-12-29
Add support for autoclonetype of bound, anonymous inner Bundles

jackkoenig:
Also change Data.outerModule to Bundle._outerInst since it is only used
in autoclonetype. _outerInst is also Option[Object] instead of
Option[BaseModule] because the outer object could also be a Bundle.

**Note that the target of this PR is autoclonetype**
Replaces #751 

This still provides the functionality I wanted but without attempting to address the more complex problem of how to memoize autoclonetype. I did benchmark memoize-autoclonetype on a large Chisel design and it had virtually no effect so I think doing the simpler thing for now is fine.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -556,6 +556,11 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     case _ => None
   }
 
+  // If this Data is an instance of an inner class, record a *guess* at the outer object.
+  // This is only used for cloneType, and is mutable to allow autoCloneType to try guesses other
+  // than the module that was current when the Bundle was constructed.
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -556,6 +556,11 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     case _ => None
   }
 
+  // If this Data is an instance of an inner class, record enclosing class
+  // This is only used for cloneType, and is mutable to allow autoCloneType to preserve the
+  // original outerInstance.
+  private var _outerInst: Option[Object] = Builder.currentModule
```
ducky64: Comments might want to note that this is only a guess, because Builder.currentModule may be the wrong thing.

jackkoenig:
Comment updated! Thanks!

Issue #669, opened 2017-08-08
Auto clonetype


ducky64:
Needing to manually override and implement cloneType is problematic because it's boilerplate (and not necessarily clear how to write one at that) and it doesn't separate implementation and usage (it's an exposed implementation detail, that we're trying to fix with chisel3 but with limited success). So being able to do this automatically would be great.

There are two major approaches:
- `@bundle` macro annotation that automatically generates cloneType. You have to remember to add the `@bundle` annotation, but it doesn't change RTL functionality (so it's closer in spirit to the `@chiselName` annotations). Implementation is more complex (means, it will be a while before this happens), and it's also intertwined with bundle literal constructors.
- Better reflection. The main problem now seems to be that constructor values aren't visible to reflection unless they are declared as `val`s - and a potential workaround is to get users to define their constructor arguments as `val`s in the cloneType error message.

In both cases, there will probably need to be restrictions on what argument types auto-cloneType can work with - namely, being restricted to common immutable types (like `Int`) and Chisel types (like `UInt`, which will be cloned). I imagine this will probably address most places (>90%?) where cloneType is needed, and the user should always be able to override with their own cloneType. Ideally, it would also be possible for the user to specify additional immutable types that cloneType would recognize, though the implementation for that is less clear.

Thoughts?

@sdtwigg @azidar @jackkoenig

jackkoenig:
Is there a way with reflection to figure determine which vals were arguments to the constructor? My understanding is that making them vals will add public fields to the class but the actual arguments to the constructor are still private (because they have to be in Java). Perhaps there is some way in Scala reflection?

I am interested in a reflection-based implementation, at a bare minimum it would be better than what we currently have. I still think the macro is the ultimate goal, if only because it would also give us easy and type-safe bundle literal construction.

ducky64:
It looks like Scala reflection should be able to get names and types of constructor arguments:
https://stackoverflow.com/questions/27473440/scala-get-constructor-parameters-at-runtime

ducky64:
The good news: it appears possible to clone Bundles with val arguments using Scala reflection and matching accessor methods to constructor arguments. 

The bad news: anonymous inner classes are a giant (maybe even unsolvable?) can of worms.

## Anonymous Inner Classes

[...]

PR #754, opened 2018-01-02
Cloning IO with compatibility

ducky64:
Part 2 for #669, changes the API such that IO(...) clones. All Bundles will need to be clone-able, but auto clone type is expected to handle most cases.

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [x] Other enhancement

* **Impact**
  - [ ] no functional change
  - [x] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**


chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -200,29 +200,10 @@ abstract class BaseModule extends HasId {
...
-   * are problematic.
-   */
-  protected def IO[T<:Data](iodef: T): iodef.type = {
-    require(!_closed, "Can't add more ports after module close")
-    requireIsChiselType(iodef, "io type")
-
+  /** Chisel2 code didn't require the IO(...) wrapper and would assign a Chisel type directly to
+    * io, then do operations on it. This binds a Chisel type in-place (mutably) as an IO.
+    */
+  protected def _bindIoInPlace(iodef: Data) {
```
kevintownsend: As with the previous comment, I don't see the purpose of the underscore.
Add a return type. e.g.: ": Unit {"

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -191,7 +191,7 @@ abstract class BaseModule extends HasId {
     *
     * TODO: remove this, perhaps by removing Bindings checks in compatibility mode.
     */
-  def _autoWrapPorts() {}
+  def _compatAutoWrapPorts() {}
```
kevintownsend: Why the underscore?
Why isn't this marked deprecated?
Why are abbreviations being used? (compat)

ducky64: Underscore is used (perhaps inconsistently) for internal functions. User code should not use this, and these are subject to change at any time.

Ideally this would be a private / protected method, but because of how chisel is structured (and the much looser semantics of chisel2, which this provides the mechanism for compatibility), it needs to be called from outside.

This isn't marked deprecated because it's not deprecated, but an internal function that we don't (currently) plan to get rid of - and don't need deprecation warnings when chisel3 itself is compiling. While we'd love to get rid of it, there's no obvious implementation path.

chiselFrontend/src/main/scala/chisel3/core/Module.scala
ducky64: See above for purpose of underscore. Return type added.

ducky64:
Yeah, I don't know of any good way (in Chisel infrastructure) to solve that, sadly the stack trace doesn't give references to objects. The other problem is the inability to have Bundles in objects, which is sometimes useful for namespacing purposes - but I don't know of a way to solve for similar reasons.

ucbjrl:
This causes failures in:
chisel-tutorials:
```bash
[info] RouterTests:
[info] a router
[info] [0.000] Elaborating design...
[info] - should have its routing table loaded and changed and route a bunch of packets *** FAILED ***
[info]   chisel3.internal.ChiselException: Unable to automatically infer cloneType on examples.RouterIO@10: constructor has parameters Set(n) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
```
chisel-testers
```bash
[info]   chisel3.internal.ChiselException: Unable to automatically infer cloneType on chisel3.core.Bundle@1b: constructor has parameters Set(compileOptions) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
...
[info]   at chisel3.iotesters.InnerMemModule.<init>(MemPokeSpec.scala:12)
[info]   at chisel3.iotesters.OuterMemModule$$anonfun$4.apply(MemPokeSpec.scala:23)
[info]   at chisel3.iotesters.OuterMemModule$$anonfun$4.apply(MemPokeSpec.scala:23)

[info] RouterUnitTesterSpec:
[info] a router
[info] - should can have it's rout table loaded and changed and route a bunch of packets *** FAILED ***
[info]   chisel3.internal.ChiselException: Unable to automatically infer cloneType on examples.RouterIO@10: constructor has parameters Set(n) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
 ...
[info]   at examples.Router.<init>(Router.scala:53)
[info]   at examples.RouterUnitTester$$anonfun$14.apply(Router.scala:88)
[info]   at examples.RouterUnitTester$$anonfun$14.apply(Router.scala:88)
```
riscv-sodor
```bash
[info] Running Sodor.elaborate -td emulator/rv32_1stage/generated-src
[info] [0.002] Elaborating design...
[error] (run-main-0) chisel3.internal.ChiselException: Unable to automatically infer cloneType on Sodor.CoreIo@46c: constructor has parameters Set(conf) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
chisel3.internal.ChiselException: Unable to automatically infer cloneType on Sodor.CoreIo@46c: constructor has parameters Set(conf) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
...
	at Sodor.Core.<init>(core.scala:35)
	at Sodor.SodorTile$$anonfun$2.apply(tile.scala:26)
	at Sodor.SodorTile$$anonfun$2.apply(tile.scala:26)
	at chisel3.core.Module$.do_apply(Module.scala:47)
	at Sodor.SodorTile.<init>(tile.scala:26)
	at Sodor.Top$$anonfun$1.apply(top.scala:20)
	at Sodor.Top$$anonfun$1.apply(top.scala:20)

```
and various rocket-chip regressions, one if which is:
```bash
[info] [0.002] Elaborating design...
Interrupt map (2 harts 2 interrupts):
  [1, 2] => ExampleRocketSystem
[error] [(run-main-0) java.lang.reflect.InvocationTargetException
[error] [java.lang.reflect.InvocationTargetException
...
[error] [Caused by: java.lang.Exception: Unable to use ParamaterizedBundle.cloneType on class freechips.rocketchip.rocket.HellaCacheBundle, probably because class freechips.rocketchip.rocket.HellaCacheBundle() takes more than one argument.  Consider overriding cloneType() on class freechips.rocketchip.rocket.HellaCacheBundle
```

ducky64:
... well, I can't say I'm surprised ...

Yeah. this is going to break people's stuff. Which is going to be bad.

One potential compromise would be to turn this into a deprecation for 3.0.x (or even 3.1), so that IO tries to clone, then emits a runtime deprecation if it fails, with the intention to turn that deprecation into an error on the next release (3.2?). This has the advantage of marking multiple error sites simultaneously (without doing the full error system overhaul) while allowing a gentler transition, but only works if people actually pay attention to (and proactively fix) runtime deprecations (do they?)

jackkoenig:
I think that's a good idea. When I [deprecated the `ffaaf` option in Firrtl](https://github.com/freechipsproject/firrtl/pull/704/files) I added such a noisy warning and it got [removed relatively quickly from rocket-chip](https://github.com/freechipsproject/rocket-chip/pull/1221)!

PR #766, opened 2018-01-24
Remove unused BuildInfo.BuildTime.

richardxia:
`BuildInfo.BuildTime` appears to cause sbt to always recompile `BuildInfo` to get the latest time, which causes a recompile on any package that depends on chisel3. This is quite painful for doing development on any Chisel projects. BuildTime doesn't actually appear to be in use, and removing it fixes the unnecessary recompilation issue.

* **Type of change**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **Impact**
Changes the firrtl file header, removing the chisel build time.

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

Removed unused `BuildInfoOption.BuildTime`, which causes sbt to unnecessarily recompile Chisel and all dependent projects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -128,7 +128,6 @@ lazy val chisel = (project in file(".")).
   enablePlugins(ScalaUnidocPlugin).
   settings(
     buildInfoPackage := name.value,
-    buildInfoOptions += BuildInfoOption.BuildTime,
     buildInfoUsePackageAsPath := true,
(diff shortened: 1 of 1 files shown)


richardxia:
Thanks, getting rid of the unnecessary recompiles will make the development experience much nicer. I tried to look into what it would take to only include the `BuildTime` at release-time, but I think my sbt-foo is too weak to do that, so I hope you have better luck.

PR #767, opened 2018-01-27
Refactor Annotations

jackkoenig:
This makes it much easier to write annotations. As an example, check out identity annotation from the tests:
```scala
// Annotation used in Firrtl
case class IdentityAnnotation(target: Named, value: String) extends SingleTargetAnnotation[Named] {
  def duplicate(n: Named) = this.copy(target = n)
}
object identify {
  // Convenient constructor for this annotation in Chisel code on Chisel InstanceIds
  def apply(component: InstanceId, value: String): Unit = {
    val anno = LazyAnnotation(() => IdentityAnnotation(component.toNamed, value))
    annotate(anno)
    // We have a transform we want to run whenever something is identified
    annotate(RunTransformAnnotation(classOf[IdentityTransform]))
   }
 }
```

* **Related issue**

This is a companion to https://github.com/freechipsproject/firrtl/pull/721

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [x] Other enhancement

* **Impact**
  - [ ] no functional change
  - [x] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

    * Updates Chisel to new JSON-based Annotation system in Firrtl (see https://github.com/freechipsproject/firrtl/pull/721).
    * Deprecates `ChiselAnnotation` in favor of `LazyAnnotation`.
    * Deprecates `Module.annotate` in favor of `object annotate`
    * Adds `.toNamed` to `InstanceId` so that we can easily generate Firrtl references to Chisel objects.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -164,7 +164,7 @@ private[chisel3] class DynamicContext() {
   val idGen = new IdGen
   val globalNamespace = Namespace.empty
   val components = ArrayBuffer[Component]()
-  val annotations = ArrayBuffer[LazyAnnotation]()
+  val annotations = ArrayBuffer[ChiselAnnotation]()
   var currentModule: Option[BaseModule] = None
   // Set by object Module.apply before calling class Module constructor
   // Used to distinguish between no Module() wrapping, multiple wrappings, and rewrapping
@@ -188,7 +188,7 @@ private[chisel3] object Builder {
   def idGen: IdGen = dynamicContext.idGen
   def globalNamespace: Namespace = dynamicContext.globalNamespace
   def components: ArrayBuffer[Component] = dynamicContext.components
-  def annotations: ArrayBuffer[LazyAnnotation] = dynamicContext.annotations
+  def annotations: ArrayBuffer[ChiselAnnotation] = dynamicContext.annotations
   def namingStack: internal.naming.NamingStack = dynamicContext.namingStack
 
   def currentModule: Option[BaseModule] = dynamicContext.currentModule
@@ -263,7 +263,7 @@ private[chisel3] object Builder {
       errors.checkpoint()
       errors.info("Done elaborating.")
 
-      Circuit(components.last.name, components, annotations.map(_.get))
+      Circuit(components.last.name, components, annotations)
     }
   }
   initializeSingletons()
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -164,7 +164,7 @@ private[chisel3] class DynamicContext() {
   val idGen = new IdGen
   val globalNamespace = Namespace.empty
   val components = ArrayBuffer[Component]()
-  val annotations = ArrayBuffer[LazyAnnotation[_]]()
+  val annotations = ArrayBuffer[LazyAnnotation]()
   var currentModule: Option[BaseModule] = None
   // Set by object Module.apply before calling class Module constructor
   // Used to distinguish between no Module() wrapping, multiple wrappings, and rewrapping
@@ -188,7 +188,7 @@ private[chisel3] object Builder {
   def idGen: IdGen = dynamicContext.idGen
   def globalNamespace: Namespace = dynamicContext.globalNamespace
   def components: ArrayBuffer[Component] = dynamicContext.components
-  def annotations: ArrayBuffer[LazyAnnotation[_]] = dynamicContext.annotations
+  def annotations: ArrayBuffer[LazyAnnotation] = dynamicContext.annotations
   def namingStack: internal.naming.NamingStack = dynamicContext.namingStack
 
   def currentModule: Option[BaseModule] = dynamicContext.currentModule
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -8,6 +8,7 @@ import scala.collection.mutable.{ArrayBuffer, HashMap}
 import chisel3._
 import core._
 import firrtl._
+import _root_.firrtl.annotations.{CircuitName, ComponentName, ModuleName, Named}
 
 private[chisel3] class Namespace(keywords: Set[String]) {
   private val names = collection.mutable.HashMap[String, Long]()
@@ -66,6 +67,9 @@ trait InstanceId {
   def pathName: String
   def parentPathName: String
   def parentModName: String
+  /** Returns a FIRRTL Named that refers to this object in the elaborated hardware graph */
+  def toNamed: Named
+
 }
 
 private[chisel3] trait HasId extends InstanceId {
@@ -129,6 +133,11 @@ private[chisel3] trait HasId extends InstanceId {
     case Some(p) => p.name
     case None => throwException(s"$instanceName doesn't have a parent")
   }
+  // TODO Should this be public?
+  protected def circuitName: String = _parent match {
+    case None => instanceName
+    case Some(p) => p.circuitName
+  }
 
   private[chisel3] def getPublicFields(rootClass: Class[_]): Seq[java.lang.reflect.Method] = {
     // Suggest names to nodes using runtime reflection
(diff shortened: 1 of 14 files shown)


jackkoenig:
TODO
- [ ] Add ScalaDoc

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,6 +133,11 @@ private[chisel3] trait HasId extends InstanceId {
     case Some(p) => p.name
     case None => throwException(s"$instanceName doesn't have a parent")
   }
+  // TODO Should this be public?
```
ducky64: For Testers, we do need an API to get the FIRRTL name of a Chisel object. How that should be done (through DataMirror maybe?) and specifics (name qualified up to which level? Top? User-defined root?) are questions though.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -142,12 +151,20 @@ private[chisel3] trait HasId extends InstanceId {
     this.getClass.getMethods.sortWith(_.getName < _.getName).filter(isPublicVal(_))
   }
 }
+/** Holds the implementation of toNamed for Data and MemBase */
+private[chisel3] trait NamedComponent extends HasId {
+  /** Returns a FIRRTL ComponentName that references this object
+    * @note Should not be called until circuit elaboration is complete
```
ducky64: Perhaps not a now thing, but can we relax when this can be called? For example, it should be possible to get the name of any bound wire, assuming it's properly named (instead of T_, which is kind of meaningless anyways) - though we might need to ensure that these names can't be mangled in the future. Going up the module chain also presents difficulties in this case, you never know when an enclosing module might not have a proper name.

ducky64:
Lazy seems like "we're abusing this Scala language feature because we're too lazy [heh] to write the plumbing properly" and not balancing the usability / learnability concerns with the implementation effort, probably partially because we're Scala power users and kind of feel that users are also at a similar level.

I think the general convention for lazy evaluation is not that it gives you a different result because of execution order, but it gives you the same result with different performance characteristics. In other words, with few exceptions (for example, infinite lists), lazy is supposed to give the same result as the non-lazy version, which is not the case here.

For this PR in particular, I think Lazy is meant to be a way to get the target name, right? Can we not leave the serialization of the annotation (which is where the name is needed) to the FIRRTL emit stage, passing in a Chisel Data/Module/... reference at the annotation stage? My experience is that the `() => ...` notation is unnecessarily confusing, and those that don't understand Chisel execution order (which in my opinion is implementation rather than interface) won't understand the motivation and might instead think it's just a poorly designed API. This notation is also not foolproof, as you can pass in a function to an eagerly-evaluated value, and would likely create a very unintuitive error message.

jackkoenig:
I agree this is kind of an abuse of the term "lazy". It would probably be better to be more explicit and call it something like `DelayedAnnotation`.

I designed this to require the minimum amount of boiler plate for annotation writers. Using a reference to the Chisel object is *possible*, but in my opinion it leads to a worse overall API. I'll lay out the two ways I have thought of to do it that way:

1. Duplicate the Annotation
```scala
// Consider some Annotation with associated information
case class MyAnnotation(target: ComponentName, info: Map[String, Int]) extends SingleTargetAnno[ComponentName] {
  def duplicate(n: ComponentName) = this.copy(target = n)
}
// I could duplicate the datastructure to also support a Chisel Data
case class MyAnnotation2(target: Data, info: Map[String, Int]) {
  // I would also have to call out what other annotation it should correspond to, 
  // and AFAIK it's impossible to have static type safety that case class corresponds to the other one
  def companionAnno = classOf[MyAnnotation]
}
```

2. Extend Firrtl's Named to allow "Lazy" or "Delayed" Names

This one is probably better but has some potential issues. I could open up the Named in Firrtl (it's currently sealed) to allow Chisel types to extend it. If we want any type safety, we'd have `BaseModule` extend `ModuleName` and `Data` extends `ComponentName` and have some notion of a `Lazy` or `Delayed` Name. This then means that whenever the user uses that annotation, they have to check that the name is actually known. This would add boiler plate to use. Alternatively, I could overwrite the unapply methods on `ComponentName` and `ModuleName` and put the dynamic check there.

My main issue with 2. is that there's a bit of magic going on behind the scenes. I suspect that people could run into subtle issues trying to match on Chisel types in Firrtl transformations (which would actually work so long as you didn't serialize between Chisel and Firrtl!) The proposal on this PR is the least magical and most explicit way to do it. Sure, `() =>` is confusing to new users, but if someone is writing custom annotations, they're most certainly also writing custom transformations which is a whole larger bag of worms then learning what a function as an argument is.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: I think making it possible to eagerly access the names of things is desirable but it's a non-trivial undertaking. We would need to ensure that querying the name of something does not impact the result. For example:
```scala
class NamingTop extends RawModule {
  val w = Wire(UInt(8.W)).suggestName("foo")
  w.getName // ???
  val foo = IO(Input(UInt(8.W)))
  w := foo
}
```
Currently, `w` gets the name `foo_1` because ports always win. We would need to be sure that calling `getName` did not accidentally change this.

Doing this with reflection also sounds... dangerous. Perhaps we could make `@chiselName` disable the reflection-based naming and make eager-naming a feature of using `@chiselName`?

jackkoenig:
Updated this with a more explicit `ChiselAnnotation` trait that should be extended in Chisel libraries that use Annotations. You can still create anonymous ones if you don't want the boiler plate, but they do provide a reasonable API for corresponding Annotations in Chisel-land and Firrtl-land.

I also add RunFirrtlTransform to preserve behavior of Driver.execute instantiating Firrtl transforms associated with a given Annotation.

src/test/scala/chiselTests/AnnotatingDiamondSpec.scala
```diff
@@ -3,47 +3,50 @@
...
+  def duplicate(n: Named) = this.copy(target = n)
+}
+/** ChiselAnnotation that corresponds to the above FIRRTL annotation */
+case class IdentityChiselAnnotation(target: InstanceId, value: String) extends ChiselAnnotation {
+  def toFirrtl = IdentityAnnotation(target.toNamed, value)
+}
+object identify {
+  def apply(component: InstanceId, value: String): Unit = {
+    val anno = IdentityChiselAnnotation(component, value)
+    annotate(anno)
```
ducky64: Why is annotate called twice?

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
ducky64: I still can't say I'm a fan of an API that depends so much on mutable state where access can be unsafe. Are there other possible APIs, like ChiselAnnotation.toFirrtl taking in a object that can remap a component to a name?

jackkoenig: What is the difference between putting the code here in `NamedComponent` vs. in `ChiselAnnotation`? In either case, getting a Firrtl `Named` from a Chisel object relies on mutable state in that object that isn't set until the end of elaboration. This PR is not changing anything about that aspect of Chisel's implementation.

src/test/scala/chiselTests/AnnotatingDiamondSpec.scala
jackkoenig: Mistakenly left an old API, fixed

jackkoenig:
retest this please

jackkoenig:
Can any one reproduce the failure I'm seeing on Jenkins (`DontTouchSpec` failing)? I can't seem to reproduce it.

ucbjrl:
[info] DontTouchSpec:
[info] Dead code
[info] - should be removed by default *** FAILED ***
[info]   "module HasDeadCodeChild( // @[:@3.2]
[info]     input  [31:0] io_a, // @[:@6.4]
[info]     output [31:0] io_b, // @[:@6.4]
[info]     output [31:0] io_c_0, // @[:@6.4]
[info]     output [31:0] io_c_1 // @[:@6.4]
[info]   );
[info]     assign io_b = io_a;
[info]     assign io_c_0 = 32'h0;
[info]     assign io_c_1 = 32'h0;
[info]   endmodule
[info]   module HasDeadCode( // @[:@12.2]
[info]     input         clock, // @[:@13.4]
[info]     input         reset, // @[:@14.4]
[info]     input  [31:0] io_a, // @[:@15.4]
[info]     output [31:0] io_b // @[:@15.4]
[info]   );
[info]     wire [31:0] inst_io_a; // @[DontTouchSpec.scala 27:20:@17.4]
[info]     wire [31:0] inst_io_b; // @[DontTouchSpec.scala 27:20:@17.4]
[info]     wire [31:0] inst_io_c_0; // @[DontTouchSpec.scala 27:20:@17.4]
[info]     wire [31:0] inst_io_c_1; // @[DontTouchSpec.scala 27:20:@17.4]
[info]     wire [32:0] _T_10; // @[DontTouchSpec.scala 30:28:@22.4]
[info]     wire [31:0] _T_11; // @[DontTouchSpec.scala 30:28:@23.4]
[info]     wire [31:0] dead; // @[:@24.4]
[info]     HasDeadCodeChild inst ( // @[DontTouchSpec.scala 27:20:@17.4]
[info]       .io_a(inst_io_a),
[info]       .io_b(inst_io_b),
[info]       .io_c_0(inst_io_c_0),
[info]       .io_c_1(inst_io_c_1)
[info]     );
[info]     assign _T_10 = io_a + 32'h1; // @[DontTouchSpec.scala 30:28:@22.4
]
[info]     assign _T_11 = _T_10[31:0]; // @[DontTouchSpec.scala 30:28:@23.4]
[info]     assign dead = _T_11; // @[:@24.4]
[info]     assign io_b = inst_io_b;
[info]     assign inst_io_a = io_a;
[info]   endmodule
[info]   " included substring "io_c_0" (DontTouchSpec.scala:45)
[info] - should NOT be removed if marked dontTouch
[info] Dont touch
[info] - should only work on bound hardware

jackkoenig:
Well, ensuring the tests each had their own test directory fixed the problem 🤷‍♀️. Anyway for rocket-chip bumping reasons, I'm going to merge https://github.com/freechipsproject/chisel3/pull/788 first. Then I'll merge this.

Issue #765, opened 2018-01-24
[RFC] What does it mean to be a Bundle?


jackkoenig:
The standard pattern for using Chisel Data objects as generator parameters in Bundles is to pass them as a constructor argument like so:
```scala
class MyBundle[T <: Data](gen: T) extends Bundle {
  val foo = Input(gen)
}
```
In the modern world of autoclonetype; however, you are encouraged to make constructor arguments `vals` so that reflection can find them. If you make `gen` a `val`:
```scala
class MyBundle[T <: Data](val gen: T) extends Bundle {
  val foo = Input(gen)
}
```
Whoops, you've accidentally added a new element to your Bundle called `gen`! This is due to the fact that the elements of a Bundle are defined (by implementation) to be the public vals of type `Data` or `Some[Data]`.

Instead, you should use a private val:
```scala
class MyBundle[T <: Data](private val gen: T) extends Bundle {
  val foo = Input(gen)
}
```
While a little verbose this allows autoclonetype to automatically generate clonetype so everyone is happy.

This led to a discussion on what the fundamental definition of a Bundle is/should be. As far as I can tell there are two possibilities:

### 1. Bundle Elements are the public vals of the Scala class (the status quo)

As far as I know, this hasn't been much a point of confusion for new users; however, with autoclonetype encouraging people to make their constructor parameters `vals` it most certainly could become one. To make this better, I think there are a couple of things we should do:
1. Write a wiki page on the definition of a Bundle with examples
1. When a `chisel3.core.Binding$MixedDirectionAggregateException` caused by a val constructor parameter of type `Data` or `Some[Data]`, inform the user that it counts as an element and link to the wiki page.
1. Augment autoclonetype to detect elements that are also constructor parameters and implicitly call `cloneType` on them. This is needed since these constructor arguments are not just generator parameters, they belong to the Bundle so we need fresh objects.
    * Added bonus of this is it should enable single-line case class Bundles

There is one case where I'm not sure if we'll be able to [easily] provide a decent error message though. For a Bundle with no directions at all, you can end up with a `firrtl.passes.CheckInitialization$RefNotInitializedException` if you have a val constructor parameter of type `Data` that you didn't intend to be an element of the Bundle.

### 2. Bundle Elements should be vals in the *body* of a Bundle
[...]

PR #768, opened 2018-01-28
Autoclonetype will clone args that are of type data

ducky64:
Short-term (until 3.1.1/3.2) patch for #765 to work better with autoclonetype

This does not attempt to give an informative error message for `MixedDirectionAggregateException `, because the exception will not be raised in all cases (for example, if the Bundle contents were not directioned), it will require significant heuristic code to catch what seems to be a very uncommon edge case, sufficient debugging information is already printed, and an API addition for 3.1.1/3.2 should remove ambiguity.

Also makes ArbiterIO cloneable through a cloneType implementation.

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [X] Other enhancement

* **Impact**
  - [ ] no functional change
  - [X] API addition (no impact on existing code)
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [X] implementation

* **Release Notes**
Autoclonetype will work with Bundle that have fields defined in their constructors.
I personally don't think that's good style. as it raises the learning curve needed, but apparently it has users.
See #765 for proposals for 3.1.1/3.2, where we may add an explicit (but optional) Field(...) construct to Bundle.

### DIFF
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -15,8 +15,7 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   * @param n number of inputs
   */
 class ArbiterIO[T <: Data](private val gen: T, val n: Int) extends Bundle {
-  // gen is a val to allow autoclonetype to work, but private to not be detected as a Bundle field.
-  // No, it's not the most intuitive or clean API. We'll revisit it for 3.1.1 or 3.2.
+  // See github.com/freechipsproject/chisel3/issues/765 for why gen is a private val and proposed replacement APIs.
 
   val in  = Flipped(Vec(n, Decoupled(gen)))
   val out = Decoupled(gen)
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -14,11 +14,13 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   * @param gen data type
   * @param n number of inputs
   */
-class ArbiterIO[T <: Data](gen: T, n: Int) extends Bundle {
+class ArbiterIO[T <: Data](private val gen: T, val n: Int) extends Bundle {
+  // gen is a val to allow autoclonetype to work, but private to not be detected as a Bundle field.
+  // No, it's not the most intuitive or clean API. We'll revisit it for 3.1.1 or 3.2.
+
   val in  = Flipped(Vec(n, Decoupled(gen)))
   val out = Decoupled(gen)
   val chosen = Output(UInt(log2Ceil(n).W))
-  override def cloneType: this.type = new ArbiterIO(gen, n).asInstanceOf[this.type]
 }
 
 /** Arbiter Control determining which producer has access
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -18,6 +18,7 @@ class ArbiterIO[T <: Data](gen: T, n: Int) extends Bundle {
   val in  = Flipped(Vec(n, Decoupled(gen)))
   val out = Decoupled(gen)
   val chosen = Output(UInt(log2Ceil(n).W))
+  override def cloneType: this.type = new ArbiterIO(gen, n).asInstanceOf[this.type]
 }
 
 /** Arbiter Control determining which producer has access
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -703,10 +703,10 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
     }
 
     // Clone unbound parameters in case they are being used as bundle fields.
-    val ctorParamsVals = ctorParamsNameVals.map{ case (paramName, paramVal) => paramVal match {
-      case paramVal: Data => paramVal.cloneTypeFull
-      case paramVal => paramVal
-    } }
+    val ctorParamsVals = ctorParamsNameVals.map {
+      case (_, paramVal: Data) => paramVal.cloneTypeFull
+      case (_, paramVal) => paramVal
+    }
 
     // Invoke ctor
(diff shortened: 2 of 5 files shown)


src/main/scala/chisel3/util/Arbiter.scala
```diff
@@ -18,6 +18,7 @@ class ArbiterIO[T <: Data](gen: T, n: Int) extends Bundle {
   val in  = Flipped(Vec(n, Decoupled(gen)))
   val out = Decoupled(gen)
   val chosen = Output(UInt(log2Ceil(n).W))
+  override def cloneType: this.type = new ArbiterIO(gen, n).asInstanceOf[this.type]
```
jackkoenig: I actually think we should change the constructor arguments to `private val gen: T, val n: Int` to take advantage of autoclonetype rather than implementing a clonetype.

I ran into an interesting issue trying to get all of this working in rocket-chip. If you override clonetype and then someone extends the class, you get runtime exceptions. This occurs in rocket-chip with `QueueIO` extended [here](https://github.com/freechipsproject/rocket-chip/blob/e7de7f3e823ffbf95fef2a0b9a6f143783393b84/src/main/scala/util/ShiftQueue.scala#L16).

I think we should change all of the library types to use autoclonetype

ducky64:
Good points, converted to use autoclonetype. I don't like private vals (from my view of a new-user wtfs/min metric - which may or may not reflect reality), but I can see how clonetype + subclasses could cause even more unintuitive errors.
I've added a explanation about private val, because in the absence of knowledge about the details of autoclonetype or bundle rules, I would be going wtf at the code.
I haven't tested on rocket yet. Also, does #769 need to be merged first?

src/main/scala/chisel3/util/Arbiter.scala
```diff
@@ -14,7 +14,10 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   * @param gen data type
   * @param n number of inputs
   */
-class ArbiterIO[T <: Data](gen: T, n: Int) extends Bundle {
+class ArbiterIO[T <: Data](private val gen: T, val n: Int) extends Bundle {
+  // gen is a val to allow autoclonetype to work, but private to not be detected as a Bundle field.
```
kevintownsend: Can you just link to the issue? (and just make this a one line comment) (1) It's easier to search for parts of the code related to the issue. (2) It's easier to get updates if someone sees this in the code.

ducky64: Good point, shortened the comment and linked the relevant issue.

ducky64:
@jackkoenig making Decoupled use autoclonetype turns out to break rocket-chip, since there's a use of a bound `gen`...

ducky64:
@ucbjrl test seems to be failing because something is causing sbt to go crazy, apparently wiping the `coreMacros/target`, `chiselFrontend/target`, and `target` folder should solve it.
```
[error] java.lang.AssertionError: assertion failed: Modified names for chisel3.util.DecoupledIO is empty
[error] 	at scala.Predef$.assert(Predef.scala:219)
```

grebe:
I'd like to add that there are reasons you might not want `gen` to be private- e.g. we do matching on `gen` when generating IPXact for polymorphic bundles. This could all be structured in a way that deals with `gen` being private, but that seems like a pretty heavy constraint to work around something that is supposed to be making things easier.

Forgive me if previous discussions have raised this point, but couldn't we use the bindings system? If a `val` in a `Bundle` has no IO binding, just leave it out of the bundle. I suppose we could instead have an explicit hardware prototype binding that marks a chisel type as not belonging in a bundle. Some sort of `PolymorphicBundle[T <: Data](val proto: T) extends Bundle { ... }` could handle the binding for you.

ucbjrl:
retest this please

jackkoenig:
@grebe How do you access `gen` currently? The use of `private vals` as in this PR is the current behavior so I'm guessing you define accessors for the generator parameters?

I'm not sure how the bindings system could handle this, is your suggestion that if you instantiate `PolymorphicBundle` as:
`new PolymorphicBundle(UInt(32.W))` then `proto` will **not** be a field where as `new PolymorphicBundle(Input(UInt(32.W)))` then `proto` **will** be a field of the bundle?
That could work but what about the situation where you don't want any directions in the Bundle?

ducky64:
The issue with bindings is that Bundle fields aren't consistently bound. Some fields might have Input or Output bindings, but some might have none, for example
```scala
  val field = UInt(8.W)
}
```
(where `field` is an unbound Bundle field)
When a Bundle is wrapped in a binding operation (like `IO(new MuBundle)`), it relies on the Bundle's element detection to determine which are fields. So bindings can't be used to determine Bundle fields, unless we introduce some kind of Field(...) API for bundles, as being discussed in #765

ucbjrl:
retest this please.

grebe:
@jackkoenig I think right now we actually match on something that the prototype is used to make, i.e. if a field is `val in = Flipped(Decoupled(gen))` we mach on `in.bits`. That's a fair point- it's not like there's a clean way to do this right now.

@ducky64 is it legal to have an unbound element in a `Bundle`? Shouldn't that be some sort of error?

ducky64:
@grebe When you wrap the Bundle in `IO(...)` (or `Reg(...)`, or etc), that binds it and its fields, using the field detection scheme mentioned above. Before that point, its elements do not need to be bound, though some of them may have directionality assigned with `Input(...)` or `Output(...)`.

jackkoenig:
It's failing because of the sbt dependency bug 🙄 

```
[error] java.lang.AssertionError: assertion failed: Modified names for chisel3.util.DecoupledIO is empty
[error] 	at scala.Predef$.assert(Predef.scala:219)
[error] 	at sbt.internal.inc.NamesChange.<init>(Changes.scala:54)
[error] 	at sbt.internal.inc.IncrementalNameHashing.sameAPI(IncrementalNameHashing.scala:46)
[error] 	at sbt.internal.inc.IncrementalCommon.sameClass(IncrementalCommon.scala:205)
[error] 	at sbt.internal.inc.IncrementalCommon.$anonfun$changedIncremental$1(IncrementalCommon.scala:187)
```

ucbjrl:
retest this please

PR #769, opened 2018-01-29
Better support for autoclonetype of nested Bundles

ducky64:
Use a Bundle Stack in Builder to keep track of currently open bundles, and use those as autoclonetype candidates.
On each Bundle's construction, it appends itself onto the stack.
Before each element is added to the stack, previous elements are pruned if their classes are no longer on the stack trace.
This allows Bundles guesses at enclosing Bundles for autoclonetype.

The previous proposal for traversing Bundle elements and marking them with parents doesn't work, because if you have code like
```scala
        val io = IO(new Bundle {
          val inner = Input(new Bundle {
            val x = UInt(8.W)
          })
        })
```
the `Input(new Bundle{...})` is called (and attempts to bind the inner) before the outer finishes construction and before the outer `IO(...)` is called. So a solution to that problem requires that references to outers are recorded before those outers finish construction. A Bundle stack does that, and using the system stack trace is the only way I could think of to prune closed Bundles (Chisel doesn't do anything when Bundles finish construction).

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [X] Other enhancement

* **Impact**
  - [x] no functional change (since 3.0)
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [X] implementation

* **Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -582,24 +582,27 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
       val outerInstance = _outerInst match {
         case Some(outerInstance) => outerInstance  // use _outerInst if defined
         case None =>  // determine outer instance if not already recorded
-          val reflectOuter = try {
-            Some(clazz.getDeclaredField("$outer").get(this))  // doesn't work in all cases, namely anonymous inner Bundles
+          try {
+            // Prefer this if it works, but doesn't work in all cases, namely anonymous inner Bundles
+            val outer = clazz.getDeclaredField("$outer").get(this)
+            _outerInst = Some(outer)
+            outer
           } catch {
-            case _: NoSuchFieldException => None
-          }
-          val allOuterCandidates = Seq(
-              reflectOuter.toSeq,
-              _containingModule.toSeq,
-              _containingBundles
-            ).flatten
-          allOuterCandidates.filter(canAssignOuterClass(_)) match {
-            case outer :: Nil =>
-              _outerInst = Some(outer)  // record the guess for future use
-              outer
-            case Nil => reflectError(s"Unable to determine instance of outer class $outerClass," +
-                s" no candidates assignable to outer class types; examined $allOuterCandidates")
-            case candidates => reflectError(s"Unable to determine instance of outer class $outerClass," +
-                s" multiple possible candidates $candidates assignable to outer class type")
+            case _: NoSuchFieldException =>
+              // Fallback using guesses based on common patterns
+              val allOuterCandidates = Seq(
+                  _containingModule.toSeq,
+                  _containingBundles
+                ).flatten.distinct
+              allOuterCandidates.filter(canAssignOuterClass(_)) match {
+                case outer :: Nil =>
+                  _outerInst = Some(outer)  // record the guess for future use
+                  outer
+                case Nil => reflectError(s"Unable to determine instance of outer class $outerClass," +
+                    s" no candidates assignable to outer class types; examined $allOuterCandidates")
+                case candidates => reflectError(s"Unable to determine instance of outer class $outerClass," +
+                    s" multiple possible candidates $candidates assignable to outer class type")
+              }
           }
       }
       (outerClass, outerInstance)
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -588,7 +588,7 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
             _outerInst = Some(outer)
             outer
           } catch {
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -223,6 +225,36 @@ private[chisel3] object Builder {
...
+    val stackClasses = Thread.currentThread().getStackTrace()
+        .map(_.getClassName)
+        .reverse  // so stack frame numbers are deterministic across calls
+
+    // Prune the existing Bundle stack of closed Bundles
+    while (!dynamicContext.bundleStack.isEmpty &&
+        (dynamicContext.bundleStack.last._3 >= stackClasses.size ||
+            (stackClasses(dynamicContext.bundleStack.last._3) != dynamicContext.bundleStack.last._2))) {
+      dynamicContext.bundleStack.remove(dynamicContext.bundleStack.size - 1)
+    }
```
jackkoenig: I think this could be better expressed as something like:
```scala
while ({
  val (_, cname, pos) = dynamicContext.bundleStack.last
  !dynamicContext.bundleStack.isEmpty && pos >= stackClasses.size || stackClasses(pos) != cname
}) {
  dynamicContext.bundleStack.trimEnd(1)
}

File: chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala [hunk 2]
```diff
@@ -552,10 +552,15 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
-  private var _outerInst: Option[Object] = Builder.currentModule
+  // Memoize the outer instance for autoclonetype, especially where this is context-dependent
+  // (like the outer module or enclosing Bundles).
+  private var _outerInst: Option[Object] = None
+
+  // For autoclonetype, record possible candidates for outer instance.
+  // _outerInst should always take precedence, since it should be propagated from the original
+  // object which has the most accurate context.
+  private val _containingModule: Option[BaseModule] = Builder.currentModule
+  private val _containingBundles: Seq[Bundle] = Builder.updateBundleStack(this)
```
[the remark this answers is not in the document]
albert-magyar: I don't think it can be a def, but it should be a lazy val.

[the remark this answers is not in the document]
ducky64: Both need to be evaluated eagerly (when the Bundle is being constructed) to be valid. Especially the second.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
ducky64: A block in a while condition, while probably legal Scala, seems like quite odd style.
Anyways, I restructured the whole thing to sidestep the issue, it first calculates the length to prune using prefixLength, then trims that.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -570,23 +575,36 @@ class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+      val outerInstance = _outerInst match {
+        case Some(outerInstance) => outerInstance  // use _outerInst if defined
+        case None =>  // determine outer instance if not already recorded
+          try {
+            // Prefer this if it works, but doesn't work in all cases, namely anonymous inner Bundles
+            val outer = clazz.getDeclaredField("$outer").get(this)
+            _outerInst = Some(outer)
+            outer
+          } catch {
+            case _: NoSuchFieldException =>
```
jackkoenig: I have been unable to produce a small example (still working on it), but I have run into `IllegalAccessException`s on cases that this PR otherwise handles, eg.
`java.lang.IllegalAccessException: Class chisel3.core.Bundle$$anonfun$11 can not access a member of class mypackage.MyModule$$anon$2 with modifiers "private final"`

Note that the code that caused this issue was merely anonymous Bundles nested inside `Decoupled` and modifying the catch block to catch `IllegalAccessException`s fixed the problem and `autoclonetype` worked just fine.

Any ideas why this might occur?

ducky64: Played around with trying to reproduce this, was only able to get the error to come up when accessing `$outer` from outside the Bundle (in the testdriver class) object. So it seems like class-based access restrictions are being enforced even with reflection.

That being said, Bundle should only be reflecting on itself, which makes this really odd. I'm also surprised that autoclonetype is being called from some `class chisel3.core.Bundle$$anonfun$11` rather than the bundle object `class mypackage.MyModule$$anon$2` itself. Maybe some edge case with how Scala translates things down to Java?

jackkoenig:
I added a catch for IllegalAccessException, with that change (assuming you agree) I think this is good to go.

ducky64:
Yeah looks good, ideally we'd have a minimal test case so that can be included in the documentation / generate informative error messages, but it appears to be a very rare case.

PR #771, opened 2018-02-08
Make uncloneable IO deprecated instead of error, improve error handling

ducky64:
It appears #754 breaks more code than I thought, so this makes it a soft error (with a runtime deprecation warning - to get people to fix their stuff before we break it for real) for now.

@ucbjrl can you see if this fixes those breaking projects, and also produces informative error messages?

This additionally changes the autoclonetype errors to be more deterministic (reporting class names instead of object names) in the most common cases, to allow the deprecations manager to deduplicate warnings.

The error messages should look like:
```
[info] [0.001] Elaborating design...
[deprecated] class chiselTests.BundleWithIntArg (1 calls): Unable to automatically infer cloneType on class chiselTests.BundleWithIntArg: constructor has parameters (i, j) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
[warn] There were 1 deprecated function(s) used. These may stop compiling in a future release, you are encouraged to fix these issues.
[warn] Line numbers for deprecations reported by Chisel may be inaccurate, enable scalac compiler deprecation warnings by either:
[warn]   In the sbt interactive console, enter:
[warn]     set scalacOptions in ThisBuild ++= Seq("-unchecked", "-deprecation")
[warn]   or, in your build.sbt, add the line:
[warn]     scalacOptions := Seq("-unchecked", "-deprecation")
[info] [0.120] Done elaborating.
```
(turning on Scala deprecation doesn't actually help with line numbers in this case, since this is something that can only be caught at runtime)

Postfix for compatibility for #754 

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [X] Other enhancement

* **Impact**
  - [X] no functional change
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [X] implementation

* **Release Notes**

Uncloneable IOs will be a deprecation warning instead of an error for now.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Error.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Error.scala
@@ -46,7 +46,7 @@ private[chisel3] class ErrorLog {
     val depTag = s"[${Console.BLUE}deprecated${Console.RESET}]"
     val warnTag = s"[${Console.YELLOW}warn${Console.RESET}]"
     val errTag = s"[${Console.RED}error${Console.RESET}]"
-    deprecations foreach { case ((message, sourceLoc), count) =>
+    deprecations.foreach { case ((message, sourceLoc), count) =>
       println(s"$depTag $sourceLoc ($count calls): $message")
     }
     errors foreach println
--- a/chiselFrontend/src/main/scala/chisel3/internal/Error.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Error.scala
@@ -28,8 +28,16 @@ private[chisel3] class ErrorLog {
     println(new Info("[%2.3f] %s".format(elapsedTime/1e3, m), None))  // scalastyle:ignore regex
 
   /** Log a deprecation warning message */
-  def deprecated(m: => String): Unit = {
-    val thisEntry = (m, getUserLineNumber)
+  def deprecated(m: => String, location: Option[String]): Unit = {
+    val sourceLoc = location match {
+      case Some(loc) => loc
+      case None => getUserLineNumber match {
+        case Some(elt: StackTraceElement) => s"${elt.getFileName}:${elt.getLineNumber}"
+        case None => "(unknown)"
+      }
+    }
+
+    val thisEntry = (m, sourceLoc)
     deprecations += ((thisEntry, deprecations.getOrElse(thisEntry, 0) + 1))
   }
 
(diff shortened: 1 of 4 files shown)


chiselFrontend/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -28,8 +28,16 @@ private[chisel3] class ErrorLog {
...
-  def deprecated(m: => String): Unit = {
-    val thisEntry = (m, getUserLineNumber)
+  def deprecated(m: => String, location: Option[String]): Unit = {
+    val sourceLoc = location match {
+      case Some(loc) => loc
+      case None => getUserLineNumber match {
+        case Some(elt: StackTraceElement) => s"${elt.getFileName}:${elt.getLineNumber}"
+        case None => "(unknown)"
+      }
+    }
```
jackkoenig: I think this could better be expressed as:
```scala
val sourceLoc = location.orElse(getUserLineNumber)
                        .map(elt => s"${elt.getFileName}:${elt.getLineNumber})
                        .getOrElse("(unknown)")
```
but that's just like, my opinion man.

chiselFrontend/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -38,12 +46,8 @@ private[chisel3] class ErrorLog {
     val depTag = s"[${Console.BLUE}deprecated${Console.RESET}]"
     val warnTag = s"[${Console.YELLOW}warn${Console.RESET}]"
     val errTag = s"[${Console.RED}error${Console.RESET}]"
-    deprecations foreach { case ((message, stackElement), count) =>
-      val line = stackElement match {
-        case Some(stackElement) => s"$depTag ${stackElement.getFileName}:${stackElement.getLineNumber} ($count calls): $message"
-        case None => s"$depTag (unknown location) ($count calls): $message"
-      }
-      println(line)
+    deprecations foreach { case ((message, sourceLoc), count) =>
```
jackkoenig: Since foreach is not purely functional it should be called via dot notation

ducky64: Fixed

chiselFrontend/src/main/scala/chisel3/internal/Error.scala
ducky64: Couldn't get that to work (`value getFileName is not a member of java.io.Serializable`), I also think the nested match makes the logic more clear.

ducky64:
It's unclear from the error in #754 if the rocket regression was from IO(...), if not, it's outside the scope of this. 3 of 4 isn't too bad...

jackkoenig:
I have a rocket-chip PR for making it work with all of the recent changes: https://github.com/freechipsproject/rocket-chip/pull/1232

With this PR the only thing blocking bumping chisel in rocket-chip is the SBT partial recompilation issue

PR #774, opened 2018-02-09
Make Bundle abstract

jackkoenig:
I'm not sure why it wasn't abstract.

The basic motivation is that autoclonetype doesn't work on instances of Bundle because it can't figure out what to do about the CompileOptions. It **does** however, work on `new Bundle {}` because it's an anonymous subclass with no constructor arguments. I'm really not sure why the UIntOps test didn't fail in https://github.com/freechipsproject/chisel3/pull/754. We could of course special case this and make it work, but I'm honestly not sure why Bundle isn't abstract?

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [x] Other enhancement

* **Impact**
  - [ ] no functional change
  - [x] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

Make Bundle abstract

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -508,7 +508,7 @@ class AutoClonetypeException(message: String) extends ChiselException(message, n
   *   }
   * }}}
   */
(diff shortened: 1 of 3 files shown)


edwardcwang:
Can we merge this?

jackkoenig:
Yep

PR #776, opened 2018-02-12
run basic tests for travis confirmation

ucbjrl:
Run `sbt test` under travis.

* **Type of change**
   - [x] Feature request

* **Impact**
  - [x] no functional change

* **Development Phase**
  - [x] implementation


.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+
+cache:
+  directories:
+    $HOME/.ivy2
+    $INSTALL_DIR
+
+git:
+  depth: 10
+
+# Don't bother building and testing the PR branch itself. Just build the merge with master.
```
colinschmidt: If these are fast tests why not run them on PRs?

ucbjrl: The tests run in under a minute. The complete Travis job takes around 10 minutes.
By default (if you don't limit branches to "master"), Travis runs two jobs:
 - a test in the PR branch (the code prior to the merge)
 - a test in master (on what would be the result of the merge)

The first test is either a duplicate of the second (assuming master has been merged into the PR branch prior to submitting the PR), or a test on a configuration (an un-merged PR) that we don't care about. I suppose that should the *merged with master* test fail, a failure of the first test would indicate that the PR was faulty to begin with, but do we really care? The failed *merged with master* test is sufficient to indicate the PR can't be successfully merged as is and it is up to the submitter to fix it. They can run whatever tests they want to run locally before submitting further commits to the PR branch.

Requiring both tests to complete, can delay Travis' confirmation of a successful PR *merged with master* test by up to an hour.

ucbjrl:
@jackkoenig - this now installs Verilator (à la firrtl) and runs all tests.

edwardcwang:
Consensus: look at it again and merge it or close it if it's not worthwhile

ucbjrl:
No longer required: we're using CircleCI.

PR #778, opened 2018-02-16
Fixed UIntToOH(x, 1) invocation with x.width == 0

terpstra:
The previous code would give an error like this in firrtl:
firrtl.passes.CheckWidths$BitsWidthException:  @[OneHot.scala 49:27:secret.fir@5451186.4]: [module FastFuzzer_47] High bit 0 in bits operator is larger than input width 0.

* **Type of change**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **Impact**
  - [x] no functional change
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation


hcook:
I am running into this too :(

Also I note that the `0.U(0.W)` on line [47](https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/util/OneHot.scala#L47) is just plain broken right now

PR #783, opened 2018-02-21
Remove checkboxes from issue/pr templates.

ucbjrl:
* **Type of change**
  Other enhancement

* **Impact**
  no functional change

* **Development Phase**
   implementation

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,31 +1,26 @@
-**Note: for support questions, you are strongly advised to use [Stack Overflow](https://stackoverflow.com/questions/tagged/chisel)**.
-
+<!-- **Note: for support questions, you are strongly advised to use [Stack Overflow](https://stackoverflow.com/questions/tagged/chisel)**. -->
+<!--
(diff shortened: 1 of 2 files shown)


colinschmidt:
Nice. This would make it look more like my recent issue? I approve

ducky64:
One comment on the use of bullet points for headings: it means users need to be careful about making responses sub-bullets, which I don't think everyone will. For example, note the formatting (indents and dashes) required in "Please tell us about your environment".

Here's what it currently looks like in GitHub




* **Type of issue**: bug report | feature request | other enhancement


* **Impact**: no functional change | API addition (no impact on existing code) | API modification | unknown


* **Development Phase**: request | proposal

* **Other information**
 

* **If the current behavior is a bug, please provide the steps to reproduce the problem:**

ucbjrl:
Should we eliminate the bullet points? If we want a bot to tag the issue based on the categories in the PR, we need some sort of structured text to guide it.

ducky64:
I'm not sure bullet points will provide that structure, especially if users aren't careful about consistently formatting to that structure. Also, shouldn't the section headings (in bold too) be sufficient to extract structure from it?

PR #785, opened 2018-02-22
Canonicalize construction of Decoupled with no payload

jackkoenig:
Add new Decoupled.empty and Decoupled()

Per request, here is `Decoupled.empty` and `Decoupled()` for constructing DecoupledIO with no payload. It's unclear if maybe EmptyBundle should be canonicalized as well instead of being a member of Decoupled?

* **Type of change**
  - [ ] Bug report
  - [ ] Feature request
  - [x] Other enhancement

* **Impact**
  - [ ] no functional change
  - [x] API addition (no impact on existing code)
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

Canonicalize construction of Decoupled with no payload. Add new Decoupled.empty and Decoupled()


ducky64:
Can we not define a stock Bundle that just provides the ready/valid? It seems a bit odd to use Decoupled there, since you're immediately tossing out the bits field.

Related: maybe also an empty Data type (that isn't a Bundle subtype, maybe a Element type of width 0) would be more appropriate here? We don't need a Bundle specifically here, just a data type that doesn't actually contain data.

jackkoenig:
I think it's desirable for the type of `Decoupled.empty` to still be a `DecoupledIO` rather than some other stock Bundle, but I agree that some generic empty Data is probably better than a concrete Bundle subtype.

This raises the question thought what should the implementation of `Data.empty` or whatever be? If we create a new type, I worry that we'll have to special case it all over the place (like in BiConnect and MonoConnect). The nice thing about `EmptyBundle` is we don't have to do any of that, it just works.

ducky64:
Also, is there a good reason for `Decoupled.empty`? Assuming we have empty data types, wouldn't `Decoupled(Empty())` make more sense? It's not that many more characters, but completely consistent.

Special-casing it is a Chisel implementation issue, which we can be much more agile about then exposed APIs. The type structure is essentially an exposed API, and while I certainly hope no one tries to match on an Empty(Bundle) with a Bundle superclass, I expect that if we did end up refactoring it, some odd use case somewhere will break. Defining it as a subtype of Data (or Element) directly gives the least guarantees. I think it also makes the type hierarchy more comprehensible.

jackkoenig:
`Decoupled.empty` was just modeled after the [Scala collections GenericCompanion method "empty"](http://www.scala-lang.org/api/current/scala/collection/generic/GenericCompanion.html) that results in such methods as `List.empty`, `Map.empty`, etc. I'm not wedded to it, I think most people would prefer to write `Decoupled()` anyway.

I think an important consideration here is that Chisel actually has a need for many "empty" types. For example: `UInt(0.W)`, `Vec(0, ...)`, and `new Bundle {}`. Zero-width wire support is iffy and I think size 0 Vecs aren't currently supported, but they all greatly simplify writing generators and are needed to ensure type inference does something sensible for things that can be empty under certain configurations.

Of course it is more important to get the API right than it is to worry about implementation details. I'm just not sure about why we should add a new type that needs special casing when we have multiple concrete instances of such an "empty" type already existing. That being said, `Decoupled.empty` (or whatever) should return `DecoupledIO[Data]` for sure.

ducky64:
Yeah, I think we need better zero width support across Chisel. It's unclear if this particular use case is a good fit (style-wise) for any of the current mechanisms, or if it should have its own type (more in the spirit of self-documenting code than it merely does the right thing). I guess if I was hacking stuff together for this case I would use a 0 width UInt, though the proper API warrants more debate.

jackkoenig:
@ducky64 Setting aside the Empty Data implementation issue, this PR actually illustrates an autoclonetype problem. I just pushed a commit that removes the `cloneType` implementation from `EmptyBundle`. Now if you run `sbt "testOnly chiselTests.DecoupledSpec"`, autoclonetype fails on EmptyBundle because it cannot determine the outer object. Any idea what's going on?

jackkoenig:
Aha, so it seems to be related to whenever the anonymous inner bundle is wrapped in Decoupled, here's a small test case:

```scala
import chisel3._
import chisel3.util._

object MyObject {
  class MyBundle(val w: Int) extends Bundle {
    val foo = Input(UInt(w.W))
  }
}
class TestModule extends Module {
  import MyObject._
  val io = IO(Decoupled(new MyBundle(8)))
}

object Test extends App {
  println(chisel3.Driver.emit(() => new TestModule))
}
```

jackkoenig:
Possibly related https://stackoverflow.com/questions/17012294/recovering-a-singleton-instance-via-reflection-from-sealed-super-trait-when-typ/17016268#17016268

ducky64:
Looking at this again:
- I think the cleanest solution may be to make Decoupled a subtype of ReadyValid, which doesn't have the `bits` field.
- There may be an argument for an empty data type (but not EmptyBundle, since that sounds like a hack - the Bundle part is purely an implementation detail, right?), perhaps analogous structurally to DontCare. There's probably a good argument for better support of empty types too, like 0-width Bits and 0-length Vecs, but that's out of scope for this PR?

ducky64:
Resolution from the dev meeting: create an empty Chisel type, something that's a substitute for `Bundle { }` but not so hacky.

chick:
@ducky64 are you going to implement this last suggest solution. I wouldn't mind doing it but I'd need a bit more info on the properties of this empty chisel type.

ducky64:
I think the properties are going to be defined as we implement this and see what works. I'd imagine it's a type that can only be connected (either Mono or Bi Connect) with other Data of same type (Empty, maybe also zero-width UInt in the future?) and generates a nop connect, and generates either no FIRRTL or a zero width UInt (@azidar thoughts?).

I think this should be done after the connections refactor, since this will likely just pile more code on top of the connect logic and complicate the refactor.

azidar:
After discussion, go with the return type as `DecoupledIO[Data]`

albert-magyar:
@jackkoenig I bumped this and cleaned it up.

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -92,6 +92,14 @@ object Decoupled
   /** Wraps some Data with a DecoupledIO interface. */
   def apply[T <: Data](gen: T): DecoupledIO[T] = new DecoupledIO(gen)
 
+  private final class EmptyBundle extends Bundle
```
ducky64, line 95:
```suggestion
  // TODO: use a proper empty data type, this is a quick and dirty solution
  private final class EmptyBundle extends Bundle
```

Issue #772, opened 2018-02-08
Running coverage under sbt 1.0 does not cover source in chiselFrontEnd


chick:
* **Type of issue**
  - [X] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **Aggregated coverage not working with sbt 1.0**
When trying to get fuller coverage information from chiselFront end, I
change the sbt file to have
`aggregate := true` instead of `aggregate := false`
Then run
`make coverage`
This does not include the chisel.core and other things from chiselFrontend

As a test I went to a version of chisel3 that did not use sbt 1.0
via 
`git checkout 3.0.2`
After doing this the `make coverage` did include chisel.core code

* **What is the use case for changing the behavior?**
We are hoping to improve our coverage, but we are also moving forward on sbt.

* **Impact**
  - [ ] no functional change
  - [ ] API addition (no impact on existing code)
  - [ ] API modification
  - [ ] unknown

* **Other information** *
The following [Compatibility with sbt 1.0 #95](https://github.com/scoverage/sbt-coveralls/issues/95) may be relevant.

azidar:
Sounds like an sbt bug, @ucbjrl will submit a issue to sbt.

PR #786, opened 2018-02-22
Revert "Aggregate coverage - discover a working combination of aggreg…

ucbjrl:
…ate usage to enable coverage of subprojects. (#782)"

This is still broken - it publishes multiple jars.
This reverts commit e08da9127b3465f158145f97ef16eb9fc8d0b5a7.

**Related issue** (if applicable)
 #772 

**Type of change*:  bug report

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,16 +136,6 @@ lazy val chisel = (project in file(".")).
   // Prevent separate JARs from being generated for coreMacros and chiselFrontend.
   dependsOn(coreMacros % "compile-internal;test-internal").
   dependsOn(chiselFrontend % "compile-internal;test-internal").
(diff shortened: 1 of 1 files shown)


ucbjrl:
I'm pretty sure I convinced myself that it didn't generate the subproject
jars. But that was yesterday.
Today, it either generates them or complains it can't find them.

ucbjrl:
I'm not entirely crazy.
```bash
sbt clean test publishLocal
```
and
```bash
sbt "++2.12.4" clean test publishLocal
```
both work correctly with the original #782.
It's the equivalent single invocation:
```bash
sbt +clean +test +publishLocal
```
that fails.

In fact, a simple
```bash
sbt +publishLocal
```
fails.

PR #787, opened 2018-02-23
Add companion objects to macro classes

jackkoenig:
Workaround for unconditional invalidation of macro classes in
incremental recompilation

* **Related issue** 

It's unclear if this is a bug in sbt incremental recompilation or an issue with our use of macros. In either case, this fixes unconditional recompilation of all scala files for projects that use chisel3 in multiproject builds (eg. rocket-chip).

sbt issue describing the problem this is working around: https://github.com/sbt/sbt/issues/3966

* **Type of change**
  - [x] Bug(?) fix/workaround
  - [ ] Feature request
  - [ ] Other enhancement

* **Impact**
  - [x] no functional change
  - [ ] API modification

* **Development Phase**
  - [ ] proposal
  - [x] implementation

* **Release Notes**

Workaround for unconditional invalidation of macro classes in incremental recompilation

### DIFF
--- a/coreMacros/src/main/scala/chisel3/internal/RangeTransform.scala
+++ b/coreMacros/src/main/scala/chisel3/internal/RangeTransform.scala
@@ -9,6 +9,7 @@ import scala.language.experimental.macros
 import scala.reflect.macros.blackbox.Context
 import scala.reflect.macros.whitebox
 
+object RangeTransform
 class RangeTransform(val c: Context) {
   import c.universe._
   def apply(args: c.Tree*): c.Tree = {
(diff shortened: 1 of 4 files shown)


ducky64:
Is this recommended practice from official macro tutorials? If not, it might be a good idea to include in the comments that this is a workaround, so if sbt fixes the bug, we can remove the workarounds. Otherwise looks sane.

jackkoenig:
I haven't been able to find anything about this in macro tutorials or examples (although I should note that the vast majority do have companion objects for the macro classes, but that's because they put stuff in them).

I added a comment to each added companion object that this is a workaround and a link to the sbt issue.

PR #790, opened 2018-02-28
Adding a style guide

kevintownsend:
**Related issue**:

**Release Notes**


edwardcwang:
General question: is this style guide meant for the Chisel compiler/codebase itself, or for projects _using_ Chisel? I was initially under the impression that it was for the former, though some items in the guide seem to address the latter. (Furthermore, different organizations/projects may have different stylistic guidelines/preferences.)

doc/style.md
```diff
@@ -0,0 +1,274 @@
...
+  ...
+}
+// Companion object to MyCustomBundle
+object MyCustomBundle {
+  ...
+}
+
+```
+
+### Packages
```
colinschmidt: Why is there another packages sections?

doc/style.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Packages
+
+Package definitions must contain the full path to the package from scala. If
+you create a subpackage, it should go in a subdirectory.
+
+    package directory.name.to.get.you.to.your.source
+
+## Imports
+Wildcard ( ._ ) imports are not allowed, with the exception of Chisel._
```
colinschmidt: I always felt like after a certain number of explicit imports from the same class, it was convenient to use the wildcard. I see the point in avoiding that but its potnetially worth discussing, especially in a case where you only have a small number of imports.

doc/style.md
```diff
@@ -0,0 +1,274 @@
...
+### UpperCamelCase vs. lowerCamelCase
+
+There is more than one reasonable way to covert English prose into camel case.
+We follow the convention defined in the [Google Java style
+guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
+
+Note that the casing of the original words is almost entirely disregarded.
+Example:
+
+Prose form     | UpperCamelCase | lowerCamelCase | Incorrect
```
colinschmidt: Its a bit confusing why you have both versions of camel case so it might be worth reiterating that different kinds of variables use the two different styles.

kevintownsend: I put the this section last in the naming section. So people will know of examples of when to use lowerCamelCase and when to use UpperCamelCase.

doc/style.md
```diff
@@ -0,0 +1,269 @@
...
+```scala
+package class.under.test.class
+package tests
+```
+
+## Comments
+We use scaladoc to automatically generate documentation from the source code.
+
+```scala
+/**
```
ducky64: Totally bikeshedding: there are apparently (at least) 3 formatting styles that Scaladoc lists.
https://docs.scala-lang.org/style/scaladoc.html
We currently use "Scaladoc style, with gutter asterisks aligned in column three". I don't know the merits / tradeoffs of each, though standardizing on something is nice.

doc/style.md
```diff
@@ -0,0 +1,269 @@
...
+it contains, plus ".scala".
+
+## Packages
+
+Package definitions must contain the full path to the package from scala. If
+you create a subpackage, it should go in a subdirectory.
+
+    package directory.name.to.get.you.to.your.source
+
+As in Scala, packages follow the [Java package naming convention]
```
ducky64: link doesn't render properly in GitHub

doc/style.md
```diff
@@ -0,0 +1,269 @@
...
+    Queue(b, depth=3),
+    Queue(c, depth=4)
+  ),
+  depth=10
+)
+```
+
+## Naming Conventions
+
+Chisel follows the [Scala Naming Conventions]
```
ducky64: link doesn't render properly in GitHub

doc/style.md
ducky64: Consider adding a strikethrough on the incorrect entries?
Also, I think the main potentially nonobvious rule here is that acronyms are treated as a word for the purposes of camelcase, consider reproducing that here?

doc/style.md
```diff
@@ -0,0 +1,269 @@
...
+Test files should reside in a subdirectory called "tests".
+The tests package should be composed of the package class you are testing.
+
+```scala
+package class.under.test.class
+package tests
+```
+
+## Comments
+We use scaladoc to automatically generate documentation from the source code.
```
ducky64: Consider adding a quick sentence providing guidance for comments content, namely that they should help in understanding the code without restating it. But perhaps less vague.

doc/style.md
ducky64: Fair point, I'm, happy migrating to JavaDoc style (incrementally, for new code), since IDE support (at least Scala in Eclipse) seem better that way.

ducky64:
ok to test

doc/style.md
```diff
@@ -0,0 +1,272 @@
...
+val myMod = Module(new MyMod())
+myMod.io <> hookUp
+```
+
+Second, in a more programmatic inline style with factory methods. In this case
+Queue is actually returning the part of the IO bundle representing the queue's
+output. The factory method takes the input IO to the queue and an optional param
+for depth.
+
+    val queueOut = Queue(queueIn, depth=10)
```
ducky64: Should this be in ` ```scala ` for syntax highlighting?

ducky64:
I think the style guide should clarify on the scope. For example, some rules (like don't work in a package called chisel3) seems mainly applicable to RTL written in Chisel, but (obviously) not the Chisel core or compiler itself. Others, like CamelCase naming and package structure are pretty applicable anywhere.

I don't currently have answers to the suggestions I brought up earlier, but it might be something we discuss and add in a future change. I'm mostly happy with this, otherwise.

Also, general thoughts about individual imports? Should the exception (where users can import `chisel3._`) be expanded to utils, so `import chisel3.util._` instead of `import chisel3.util.{Decoupled, Counter, ValidIO, ...}`?

doc/style.md
kevintownsend: Done.

kevintownsend:
I lean towards not having wildcard imports. I think: `import chisel3.util.{Decoupled, Counter, ValidIO, ...}` is fine. Again, this is just a start, the style can be discussed and changed.

seldridge:
One thing to consider... 

I've always viewed the scalastyle xml as being the definitive version on what is allowed and not allowed. Things like disabling wildcard imports with regex ignores [can be done there](http://www.scalastyle.org/rules-1.0.0.html#org_scalastyle_scalariform_UnderscoreImportChecker). 

Whatever the decision on some of these points are, it should be reflected in the scalastyle checks.

kevintownsend:
scalastyle isn't comprehensive. Also, I think it is preferable to make quick changes to the style guide, then make the changes to scalastyle. That way if reviewers keep requesting changes to the style you only have to change style.md and not keep changing checkstyle and making sure it's correct.

PR #791, opened 2018-03-01
Tagbackendtests

ucbjrl:

This was incorporated from the `ucbjrltravis` branch (#776) and submitted as a separate PR since it has merit outside of any Travis usage, and by treating it distinctly from Travis support, significantly reduces the impact of that PR.
**NOTE**: This just tags tests and doesn't of itself change the tests that are run.

**Release Notes**
Tests requiring a backend  - vcs/verilator or firrtl-interpreter - are tagged as such, allowing their exclusion in quick and dirty smoke testing.


ducky64:
Why is there both `@RequiresBackend` and `taggedAs(TagRequiresBackend)`?
Can this be dynamically detected, because we already separate elaborate from test? (in other words, is there a way to achieve skipping tests without the user duplicating information?)

ucbjrl:
These aren't duplicates; they are alternatives. The problem is ScalaTest provides alternative mechanisms for tagging tests. This is an attempt to provide some hints for consistent treatment regardless of which mechanism is used.

I wish this could be determined dynamically, but I'm afraid that would require more knowledge of ScalaTest internals than I currently possess.

ducky64:
Resurrecting this PR.

Comments that should be addressed:
- Is there a reason we have `@RequiresBackend` as opposed to `@RequiresSimulator` / etc?
- firrtl-interpreter should basically run anywhere, shouldn't it? What are the issues with running the interpreter-based tests? And is running a test suite that bypasses simulation completely worth anything at all? I don't think the elaboration-only tests are anywhere near as comprehensive.
- Can you justify why three different styles are needed? ScalaTest provides them, but I'm not convinced using all of them is worth much. There seems to be `@RequiresBackend`, `taggedAs(TagRequiresBackend)`, and `TagRequiresBackend` (in property).  Cases 2 and 3 seem to be a case of us not standardizing on the `property(...)` and the `... should ...` style (which we should fix!), though I'm less convinced about the use of `@RequiresBackend` which only seems to save some characters by not spraying `taggedAs(...)` on the subtests.

PR #798, opened 2018-03-09
Testers2+interprocesstester

ucbjrl:

Add verilator/VCS (untested) backends.
Use TesterOptions to choose tester backend.
Run BasicTest on firrtl and verilator backends.
I've integrated the resource and backend code from chisel-testers preserving their commit history.
We can eliminate this if desired.


ducky64:
Um, I don't think now is the right time to add 2k LoC to new testers.
Yes, we do need Verilator / VCS / etc support sometime. But it should be done right. Definitely not before the new testers API and core abstractions have stabilized. And not with a 2k LoC code dump.

edwardcwang:
Moving discussion to https://github.com/ucb-bar/chisel-testers2/issues/28

PR #801, opened 2018-03-20
Fallback null insertion for autoclonetype

ducky64:
If autoclonetype is unable to determine an outer class, this attempts to insert a null (and give a deprecation warning), preserving old behavior (in some cases) where the new behavior doesn't work.

This doesn't provide full compatibility with old autoclonetype: this does not attempt null insertion in the general first argument (if it's not an outer class reference). Reasoning is that inserting a null for an explicit argument is probably not the right thing to do, and will likely cause a difficult-to-debug NullPointerException (whereas that would be unlikely for an outer class, which is not always referenced in Bundle subclass code).

@colinschmidt would this have prevented some of your autoclonetype errors when porting code?

**Related issue**: N/A

**Release Notes**


colinschmidt:
Just ran into another one of these problems. So I'm gonna say this would help me :)

PR #804, opened 2018-04-10
Annotations are Options, Immutable Option Parsing

seldridge:
This changes the way that options are parsed to align with a proposal for immutable option parsing in FIRRTL. This proposal uses immutable scopt parsing, specifically a `scopt.OptionParser[AnnotationSeq]` object. Each command line option maps to a single `firrtl.Annotation`. These will be consumed lazily to set `chiselOptions` (or `firrtlOptions` or `myFrontendOrBackendOptions`). This technically provides two equivalent methods for setting options:
  - Either through something which mimics a command line `Array[String]`
  - Through the explicit setting of `Seq[firrtl.Annotation]` written to a file

In consequence, `chiselOptions` is now immutable and the old approach of creating an options manager, explicitly calling a parse method, and then tweaking `chiselOptions` is disallowed.

@ucbjrl: This is a rough guide for how to migrate to use immutable parsing. Usually, it's just some fixes to the driver and a cleanup of any tests that may be using the options manager directly. I have not looked at the other repos (testers/treadle/firrtl-interpreter/rocket-chip) to see if there's any additional oddities there.

**Related issue**: 
- https://github.com/freechipsproject/firrtl/issues/764
- https://github.com/freechipsproject/firrtl/issues/765 (__depends on this__)

**Impact**: API change 

Any attempt to mutably modify `firrtlOptions` or `chiselOptions` will no longer work. The old method of providing options via an options manager is still supported, but deprecated. Attempts to create an options manager that do not conform to mandatory options required by FIRRTL (e.g., the name of the top module cannot be determined) will fail.

**Release Notes**

Switch to using immutable parsing of command line options

Add two new options for controlling whether or not specific
output files are generated:
  - `--dont-save-chirrtl`: Do not save CHIRRTL output
  - `--dont-save-annotations`: Do not save annotations file

Deprecates passing in an explicit options manager to
`chisel3.Driver.execute` in favor of passing in command line
arguments (`Array[String]`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ test_run_dir
 *~
 \#*\#
 .\#*
+utils
--- a/build.sbt
+++ b/build.sbt
@@ -102,7 +102,7 @@ lazy val chiselSettings = Seq (
   libraryDependencies ++= Seq(
     "org.scalatest" %% "scalatest" % "3.0.1" % "test",
     "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
-    "com.github.scopt" %% "scopt" % "3.6.0"
+    "com.github.scopt" %% "scopt" % "3.7.0"
   ),
 
   // Tests from other projects may still run concurrently.
--- a/build.sbt
+++ b/build.sbt
@@ -161,3 +161,7 @@ lazy val chisel = (project in file(".")).
     // published artifact) also see the stuff in coreMacros and chiselFrontend.
     exportJars := true
   )
+
+assemblyJarName in assembly := "chisel3.jar"
+test in assembly := {} // skip tests when building fat jar
+assemblyOutputPath in assembly := file("./utils/bin/chisel3.jar")
--- a/chiselFrontend/src/main/scala/chisel3/core/Annotation.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Annotation.scala
@@ -87,4 +87,3 @@ object dontTouch { // scalastyle:ignore object.name
     data
   }
 }
-
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -16,3 +16,4 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.7.0")
 
 addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")
 
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
--- a/src/main/scala/chisel3/ChiselExecutionOptions.scala
+++ b/src/main/scala/chisel3/ChiselExecutionOptions.scala
@@ -3,36 +3,63 @@
 package chisel3
 
 import firrtl.ExecutionOptionsManager
-import firrtl.annotations.NoTargetAnnotation
+import firrtl.annotations.{
+  Annotation,
+  NoTargetAnnotation }
 
-sealed trait ChiselOption
+/** Indicates that a subclass is an [[firrtl.annotation.Annotation]] with
+  * an option consummable by [[HasChiselExecutionOptions]]
+  *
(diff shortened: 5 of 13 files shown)


seldridge:
Ping @ducky64: per your request, can you take a look? (It looks like I'm unable to request reviewers...)

ducky64:
Looking at the tests and glancing over the implementation, this looks mostly sane? I do have some comments:
- My main concern is the two ways to set options: annotations and command line. I think learning curve could be an issue, since there are two styles to learn. Also, two places to look for possible configuration issues. At least, perhaps there should be guidelines for which to use where?
- I'm still unsure when it makes sense for command-line options to be preferred over Scala annotation declaration, since command-line options need to be stringly-typed, and code can be better organized in Scala.
- One interesting analogy may be C compiler options (command line, or encoded in build rules) vs. pragmas (specified in code). Of course, user transforms are also less of a thing there...
- Can you add a test case for specifying options through annotations in Scala (if this is a supported interface)?
- More system architecture level documentation would also be nice, perhaps ideally in ScalaDoc?

seldridge:
> My main concern is the two ways to set options: annotations and command line.

Annotations are the ground truth. Anything from the outside world must be converted to an annotation. The command line options only exist as helpers that create annotations. The internal `chiselOptions`, `commonOptions` only exist as data structures that a `Driver` constructs from a snapshot/view of the annotations. The real need for this (and rough equality of the CLI being able to generate all rudimentary annotations) is when building a standalone executable.

> ... when it makes sense for command-line options to be preferred over Scala annotation declaration

This is the standalone executable scenario. More specifically, Rocket Chip's usage of a standalone FIRRTL fat jar. This is more of a standard usage model for FIRRTL than it is for Chisel (though you can do the same for Chisel and this PR actually enables that). Outside of that case, it's up to the user. I agree with you on the stringly typed annoyance of CLI arguments, but that needs to exist for FIRRTL. The usage of annotations blackboxes the internals of a `Driver` (`chiselOptions`) while ensuring that there exists an external means of providing what the `Driver` wants.

> Can you add a test case for specifying options through annotations in Scala (if this is a supported interface)?

Can do. I'll mix in the usage somewhat in the test case and make sure that the ScalaDoc is up to date for the Driver. I was biased towards CLI arguments when writing this, but @chick was inversely biased when migrating Treadle (everything uses annotations). An explicit mix in the test cases would be good. That may already exist for FIRRTL, but I can make sure it exists here, too.

> More system architecture level documentation would also be nice, perhaps ideally in ScalaDoc?

I'm in agreement here. One point of potential confusion / documentation duplication is that most of the system architecture items related to this PR should likely live in FIRRTL. We'd need to then either publish one unified ScalaDoc or publish FIRRTL's ScalaDoc and then do cross linking (which supposedly works).

Thanks for the feedback, @ducky64!

ducky64:
My concern about two ways to do everything is the user-facing interface. Internally, I think the 'convert everything into an annotation' is quite clean, but from the user perspective, there's two places to check for whether something gets set, and there's two ways to learn (which can get confusing unless every source of documentation, including user-contributed documentation like StackOverflow, presents both cases, instead of just the one that the author thought was convenient at the time).

I think we should approach mixing CLI and annotations with caution, for those above reasons. I don't have a particularly good solution off the top of my head right now, but I think it warrants discussion.

chick:
I'd like to cast my vote for the two ways of doing something. One way of doing something is a great goal but I think it requires some compelling metaphor or user interface. For the amount of work that has been done in this area without discovering such a single method suggests there is not one. There are many software stacks I interact with that have multiple ways of doing things and as long as interfaces are well documented and the interaction between the different methods is clean and also documented this does not present a problem.

seldridge:
I'm opting to break this up into multiple PRs like with [F764](https://github.com/freechipsproject/firrtl/issues/764). I'll provide a back-link to the discussion here.

PR #810, opened 2018-04-20
Add Module.currentModule for getting a reference to the current Module

### LINKED ISSUE #809: Access to Current Module Object

**Other information**
 

Please see request at https://github.com/freechipsproject/rocket-chip/pull/1363

Not a bug

Cannot access the current Module

Would like to access the current Module

Access the current Module similarly to how `Module.clock` and `Module.reset` provide access to those values.

jackkoenig:
Resolves #809 and https://github.com/freechipsproject/rocket-chip/pull/1363

**Release Notes**

Add ability to get a reference to the current Module

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -75,6 +75,8 @@ object Module {
   def clock: Clock = Builder.forcedClock
   /** Returns the implicit Reset */
   def reset: Reset = Builder.forcedReset
+  /** Returns the current Module */
+  def currentModule: Option[BaseModule] = Builder.currentModule
 }
 
 /** Abstract base class for Modules, an instantiable organizational unit for RTL.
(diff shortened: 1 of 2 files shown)


ducky64:
So the related issue (https://github.com/freechipsproject/rocket-chip/pull/1363) says that there is a need to access the current module from within nested helpers, but doesn't give details. I'm curious what those are.

However, if we really do want the current module, I think this API looks sane.

jackkoenig:
@ducky64 An example use case is the [rocket-chip Reg(Field|Mapper)](https://github.com/freechipsproject/rocket-chip/blob/78dad3e89b17a0759cda42da05ef42297b21e580/src/main/scala/regmapper/RegField.scala) which constructs a TileLink memory map for memory-mapped registers in the design. Passing around the current module throughout some fairly sophisticated logic would be onerous, but it is useful to be able to attach information (ie. annotate) the current Module with this information.

ducky64:
@jackkoenig I don't see a reference to Module in the linked file (and I don't quite understand the example), but mostly what I want to make sure in this PR is that we're not using this as a less-perfect proxy for some other more narrow but more relevant feature, like associating annotations with wires.

As your understanding of the use case is better than mine, if this is actually the feature you want, I have no objections.

PR #814, opened 2018-04-23
Add chisel version to generated top-level page.

ucbjrl:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,12 @@ ifneq (,$(RM_DIRS))
 	$(RM) -r $(RM_DIRS)
 endif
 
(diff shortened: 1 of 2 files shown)


ducky64:
Is there a way to do this by configuring the doc generator instead of as a hackish post-processing string replacement operation?

ucbjrl:
It's actually a hackish _pre-processing_ string replacement. Although one would think sbt-unidoc would provide a mechanism to support this (they add the version to page titles), it doesn't appear to be documented. I have a slight preference for a visible, user-controlled string replacement hack, than using undocumented implementation details.

Issue #818, opened 2018-04-27
Chisel Standard Library


ducky64:
So the standard library we currently have (chisel3.utils) is kind of haphazard, incomplete, and in some cases, inelegant and confusing (I'm looking at you, Counter...)

This issue is just to start the conversation on what we want in a standard library. Ideally, these should be highly-reusable parts, and kind of following in Python's philosophy of "batteries included".

Some may already exist elsewhere (most notably, in [rocket-chip](freechipsproject/rocket-chip)'s various utilities), the goal here would be to make it easy to use those bits without importing the monster that is rocket. A standard library would also carry the expectation of stability.

## Standard library thoughts
I'll start off a list here. It's not complete or exhaustive, just a starting point for discussion. **Please add more suggestions in the comments!**

### Async utils
- Clock crossing FIFOs
- Clock dividers
- PLL templates, generating into FPGA blocks or blackboxes

### Communications utils
- Digital CDR blocks
- Encoders / decoders: 8b/10b, 64b/66b, NRZ, ...
- Standardized interfaces: JTAG, boundary scan, ...
- Standardized communication protocols: USB (2.0 | 3.0), SD card, Ethernet, PCIe, ...
- Standardized inter-chip protocols: I2C, SPI, UART, ...
- On-chip busses: TileLink, AXI, ...
- CRC generator

### Processing
- All the DSP stuff, with a standardized interface

### Physical interfaces
- FPGA blocks, including output drivers, input buffers, tristate drivers, and associated bundles (like a Tristate bundle)
- DDR/DDR2/etc interfaces

### Other
- Most of the stuff we have in the current chisel3.utils package, but probably cleaning up the interface a bit. API breakages can be avoided by deprecating what we have right now in favor of a different package, like chisel3.stdlib (or something)
- Efficient SRAM versions of things like registers and queues

## Issue Data

**Type of issue**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: request

**Other information**

seldridge:
I've got a small number of things that I can see about re-licensing (highlighted below).

Some ideas:
  - PRNGs
    - [A better LFSR](https://github.com/IBM/perfect-chisel/blob/master/src/main/scala/perfect/random/Lfsr.scala)
  - Checksums (you've got CRC, but others may be useful)
    - [Fletcher](https://github.com/IBM/perfect-chisel/blob/master/src/main/scala/perfect/util/Fletcher.scala) 
    - MD5?
    - SHA
[...]

PR #819, opened 2018-04-27
Add width utility functions to avoid incorrect usage of bare log2Ceil().

ucbjrl:
**Related issue**: #818 

**Release Notes**
Add helper functions `UnsignedBitsRequired` and `UnsignedWidthRequired` to provide an alternative to the potentially incorrect usage of `log2Ceil` to calculate wire widths.


edwardcwang:
Consensus from dev meeting: shorten names and general agreement to add something like this

ucbjrl:
Remove `apply(Data)` method
change name(s) to `signedBitLength`, `unsignedBitLength`

src/main/scala/chisel3/util/Math.scala
```diff
@@ -99,3 +98,34 @@ object isPow2 {
...
+    *
+    * Basically, `n.bitLength` unless n is 0 (since Java believes 0.bitLength == 0), in which case the result is `1`.
+    * @param in - the number to be encoded.
+    * @return - an Int representing the number of bits to encode.
+    */
+  def apply(in: BigInt): Int = {
+    require(in >= 0)
+    if (in == 0) {
+      // 0.bitLength returns 0 - thank you Java.
+      1
```
ducky64: Was this the behavior we wanted? (I'm sure we brought this up, I don't remember the resolution and I don't think it was written down)

src/test/scala/chiselTests/Math.scala
```diff
@@ -0,0 +1,38 @@
...
+
+class Math extends ChiselPropSpec {
+  import chisel3.util._
+  // Disable shrinking on error.
+  implicit val noShrinkListVal = Shrink[List[Int]](_ => Stream.empty)
+  implicit val noShrinkInt = Shrink[Int](_ => Stream.empty)
+
+  property ("unsignedBitLength is computed correctly") {
+    forAll(safeUIntWidth) { case (width: Int) =>
+      for ( offset <- List(-1, 0, 1)) {
```
ducky64: I'm not sure this test is testing anything since it's basically the duplicating the code used to implement unsignedBitLength. Consider just dropping in a few values and hand-coding the result.

src/test/scala/chiselTests/Math.scala
```diff
@@ -0,0 +1,38 @@
...
+          val t = if (n == 1) 1 else BigInt(n).bitLength
+          d shouldEqual (t)
+        }
+      }
+    }
+  }
+
+  property ("signedBitLength is computed correctly") {
+    forAll(safeUIntWidth) { case (width: Int) =>
+      for ( offset <- List(-1, 0, 1)) {
```
ducky64: ditto

src/main/scala/chisel3/util/Math.scala
```diff
@@ -99,3 +98,34 @@ object isPow2 {
...
+
+object signedBitLength {
+  /** Return the number of bits required to encode a specific value, assuming a sign bit is required.
+    *
+    * Basically, `n.bitLength` + 1.
+    * @param in - the number to be encoded.
+    * @return - an Int representing the number of bits to encode.
+    */
+  def apply(in: BigInt): Int = {
+    in.bitLength + 1
```
ducky64: Is this always true? With a signed int you can go one negative number lower than the max positive number right (for example, 16-bit signed is -32768 to 32767)?

aswaterman: BigInt.bitLength explicitly assumes two's complement, so this is correct.

ducky64: Yep - just tried this in the repl and `BigInt(32767).bitLength`, `BigInt(32768).bitLength` both return 15

src/main/scala/chisel3/util/Math.scala
ucbjrl: I believe so, but we can review at next dev meeting.

src/test/scala/chiselTests/Math.scala
ucbjrl: Corrected in ec3e030

src/test/scala/chiselTests/Math.scala
ucbjrl: Corrected in ec3e030

src/main/scala/chisel3/util/Math.scala
ucbjrl: Return 0, not 1.

jackkoenig:
@ducky64 thoughts on merging thisz?

jackkoenig:
Ping @ducky64

ducky64:
Is the zero width fix implemented? It doesn't appear so?

ucbjrl:
Thanks @ducky64 for pointing out the zero width fix hadn't yet been implemented. Should we be doing this for signed 0 as well? (Currently `signedBitLength(0) == 1`.)

ducky64:
The unsigned case looks fine. The signed zero case is a good question.

My thoughts for signed: 
0 bits: 0
1 bit: 0, -1
2 bits: 1, 0, -1, -2
though more thoughts are welcome

I'm still not a fan of the test style since it's not simple enough to be obvious at a glance (eg, explicit testvectors) but also doesn't give you the better correctness guarantees from a more formal approach. In fact, the test code is much harder to reason about than the main code, because there's so much complexity going on. That being said, I don't care enough about this to block it. Once we resolve the signed zero, I'll give this a green.

ucbjrl:
I agree with @ducky64, but I think there's enough confusion over this, that we shouldn't incorporate it. (I don't see how you can represent `0` without requiring any bits. Perhaps my lack of hardware knowledge is showing.)

jackkoenig:
Zero-width wires are fine and we have rules for removing them.

The advantage of representing `0.U` with zero bits is that you can perform operations on it without increasing the resulting width, eg.: `wire +& 0.U(1.W)` will have a resulting width 1 greater than that of `wire`, while `wire +& 0.U(0.W)` will have a width the same as that of `wire`. That being said, interval width inference could be an alternative mechanism for this feature and would work for literals of any width.

ucbjrl:
But why add the constant `0.U` in the first place? I would argue that this should be dealt with in optimizations for specific operations, not as a side-effect of the width required to store a specific constant. Similarly, `x  - 0.U`, `x + 0.S`, `x - 0.S`, `x * 1.U`, and `x * 1.S` should be optimized away.

aswaterman:
I’ll also add that zero-width zeros are particularly useful for Cat-like
operations in generator code. Avoids special-casing the case that there’s
only 1 of something, i.e., when log2(number of things) = 0.

ucbjrl:
Right. I keep forgetting the assumption that a zero-width wire has the value `0`. I won't let my confusion hinder the adoption of this PR.

ducky64:
Resolution from today's meeting: signed 0 is 0 bits, signed -1 is 1 bit, signed 1 is 2 bits.

PR #823, opened 2018-05-24
Fix UIntToOH for output widths larger than 2^(input width)

aswaterman:
UIntToOH(foo, bar) fails when bar > (2^foo.getWidth).

**Related issue**: https://github.com/freechipsproject/chisel3/issues/822

**Type of change**: bug report +/- feature request

**Release Notes**


aswaterman:
@hcook fyi

PR #825, opened 2018-05-24
Use Vec.apply instead of new Vec in VecInit.apply

jackkoenig:
The Vec constructor invokes the gen argument for each element in the
Vec. Since VecInit invokes cloneSupertype which touches every element of
the input Seq, this was an n^2 operation. Vec.apply accepts its
arguments by value so cloneSupertype is only called once. It then calls
cloneType on that once for each element in the Vec, which is constant
time reducing the overall complexity of VecInit to just n.

**Release Notes**
Speed up VecInit performance by a factor of n


jackkoenig:
Building on https://github.com/freechipsproject/chisel3/pull/824, this PR further reduces `VecInit(Seq.fill(16384)(true.B))` runtime from ~5s to <100ms

Issue #472, opened 2017-01-31
Chisel vals in objects can prevent deduplication


jackkoenig:
**Workaround:** Don't use `val` in objects that return Chisel types, use `def` instead

**How do I know if I have this problem?:** When you have N of some identical module that dedups down to 2 instead of 1 final module

```scala
object MyConsts {
  val amount = 1.U
}

class ShouldDedupIO extends Bundle {
  val in = Input(UInt(32.W))
  val out = Output(UInt(32.W))
}

class ShouldDedup extends Module {
  val io = IO(new ShouldDedupIO)
  io.out := io.in + MyConsts.amount
}

class DedupTop(n: Int) extends Module {
  val io = IO(new ShouldDedupIO)
  val insts = Seq.fill(n)(Module(new ShouldDedup))
  var ref = io.in
  for (inst <- insts) {
    inst.io.in := ref
    ref = inst.io.out
  }
  io.out := ref
}
```

ShouldDedup deduplicates down to 2 final Verilog modules rather than just 1. Looking at the resulting Firrtl tells us why:
```
  module ShouldDedup :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : UInt<32>, out : UInt<32>}

    io is invalid
    node _T_7 = add(io.in, UInt<1>("h01")) @[DedupSpec.scala 69:19]
    node _T_8 = tail(_T_7, 1) @[DedupSpec.scala 69:19]
    io.out <= _T_8 @[DedupSpec.scala 69:10]

  module ShouldDedup_1 :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : UInt<32>, out : UInt<32>}

    io is invalid
    node _T_6 = add(io.in, UInt<1>("h01")) @[DedupSpec.scala 69:19]
    node _T_7 = tail(_T_6, 1) @[DedupSpec.scala 69:19]
    io.out <= _T_7 @[DedupSpec.scala 69:10]

  module ShouldDedup_2 :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip in : UInt<32>, out : UInt<32>}

    io is invalid
    node _T_6 = add(io.in, UInt<1>("h01")) @[DedupSpec.scala 69:19]
    node _T_7 = tail(_T_6, 1) @[DedupSpec.scala 69:19]
    io.out <= _T_7 @[DedupSpec.scala 69:10]

  module ShouldDedup_3 :
    input clock : Clock
    input reset : UInt<1>
[...]

PR #826, opened 2018-06-01
Literals set the ref so that they no longer get named

### LINKED ISSUE #763: withClockAndReset generates incorrect FIRRTL causing UndeclaredReferenceException
Consider the following snippet:

```scala
package chiselTests

import chisel3._
import chisel3.util.Counter

import chisel3.core.UserModule
import chisel3.experimental.withClockAndReset

class ResetBugDemo extends UserModule {
  val io = IO(new Bundle {
    val reset_n = Input(Bool())
    val clock_1MHz = Input(Clock())
    val clock_20MHz = Input(Clock())
  })

  /*** BLOCK 1 ***/
  val system_reset = !io.reset_n
  val reset_1MHz = Wire(Bool())
  withClockAndReset(clock = io.clock_1MHz, reset = false.B) {
    val sync1_reset_1MHz = RegNext(system_reset)
    val sync2_reset_1MHz = RegNext(sync1_reset_1MHz)
    reset_1MHz := sync2_reset_1MHz
  }

  /*** BLOCK 2 ***/
  val clock_internal_5MHz = Wire(Bool())
  val hack_reset_signal = WireInit(false.B) // this line shouldn't be needed; we should be able to do reset=false.B
  withClockAndReset(clock = io.clock_20MHz, reset = false.B) { // <-- reset=false.B causes UndeclaredReferenceException
  //withClockAndReset(clock = io.clock_20MHz, reset = hack_reset_signal) {
    val num_cycles_per_period_5MHz: Int = 4
    val clock_internal_5MHz_r = RegInit(false.B)
    val flip_bit_5MHz = Counter(num_cycles_per_period_5MHz / 2).inc()
    when(flip_bit_5MHz) {
      clock_internal_5MHz_r := !clock_internal_5MHz_r
    }
    clock_internal_5MHz := clock_internal_5MHz_r
  }
}

class ResetBugDemoSpec extends ChiselPropSpec {
  property("should elaborate") {
    import chisel3.core.ExplicitCompileOptions.NotStrict
    chisel3.Driver.execute(Array[String](), () => new ResetBugDemo)
  }
}

```

In block 1, we are able to specify `reset=false.B` in withClockAndReset, yet in block 2 using the same `reset=false.B` construction causes a `UndeclaredReferenceException` in firrtl. It seems like block 2 is causing Chisel to generate some incorrect firrtl.

```
firrtl.passes.CheckChirrtl$UndeclaredReferenceException:  @[Counter.scala 31:40:@16.4]: [module ResetBugDemo] Reference _T_13 is not declared.
firrtl.passes.CheckChirrtl$UndeclaredReferenceException:  @[Counter.scala 26:33:@17.4]: [module ResetBugDemo] Reference _T_13 is not declared.
firrtl.passes.PassException: 2 errors detected!
```

* **Type of issue**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **If the current behavior is a bug, please provide the steps to reproduce the problem:** Run the above code
  * **What is the current behavior?** Chisel generates incorrect firrtl causing `UndeclaredReferenceException`
  * **What is the expected behavior?** Chisel should NOT generate incorrect firrtl

* **Impact**
  - [x] no functional change
  - [ ] API modification
  - [ ] unknown

* **Development Phase**
  - [x] request
  - [ ] proposal

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

Chisel revision 70ca35b9d7b3884e5f701d49bc5286f89701fd14.

Although we can work around the issue by explicitly using a WireInit for the reset, this does seem like a bug in Chisel that should be addressed.

Commenting out block 1 seems to have no effect on the bug.

jackkoenig:
Have literals set their ref so that a name isn't allocated

This has a few implications
* Constructing a literal no longer increments `_T_` suffixes
* Internally, wrapping a literal Bits in Node(...) will work
* Literal Bools work in withReset/withClockAndReset
* Data.ref is potentially redundant

**Related issue**:  Fixes #763.

@azidar beat me to fixing #472, but this would also have fixed that issue (I added a test in this PR as well)

**Release Notes**
Literal Bools work for reset in withClockAndReset and withReset
Literals no longer increment temporary names suffix

### DIFF
--- a/src/test/scala/chiselTests/MultiClockSpec.scala
+++ b/src/test/scala/chiselTests/MultiClockSpec.scala
@@ -142,6 +142,18 @@ class MultiClockSpec extends ChiselFlatSpec {
       assert(withReset(this.reset) { 5 } == 5)
     })
   }
+  it should "support literal Bools" in {
+    assertTesterPasses(new BasicTester {
+      val reg = withReset(true.B) {
+        RegInit(6.U)
+      }
+      reg := reg - 1.U
+      // The reg is always in reset so will never decrement
+      chisel3.assert(reg === 6.U)
+      val (_, done) = Counter(true.B, 4)
+      when (done) { stop() }
+    })
+  }
 
   "withClockAndReset" should "return like a normal Scala block" in {
     elaborate(new BasicTester {
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -67,6 +67,9 @@ sealed abstract class Bits(width: Width, override val litArg: Option[LitArg])
   // Arguments for: self-checking code (can't do arithmetic on bits)
   // Arguments against: generates down to a FIRRTL UInt anyways
 
+  // If this is a literal, setRef so that we don't allocate a name
+  litArg.foreach(setRef)
+
   // Only used for in a few cases, hopefully to be removed
   private[core] def cloneTypeWidth(width: Width): this.type
 
--- a/src/test/scala/chiselTests/DedupSpec.scala
+++ b/src/test/scala/chiselTests/DedupSpec.scala
@@ -41,6 +41,31 @@ class NestedDedup extends Module {
   io.out <> inst1.io.out
 }
 
+object DedupConsts {
+  val foo = 3.U
+}
+
+class SharedConstantValDedup extends Module {
+  val io = IO(new Bundle {
+    val in = Input(UInt(8.W))
+    val out = Output(UInt(8.W))
+  })
+  io.out := io.in + DedupConsts.foo
+}
+
+class SharedConstantValDedupTop extends Module {
+  val io = IO(new Bundle {
+    val in = Input(UInt(8.W))
+    val out = Output(UInt(8.W))
+  })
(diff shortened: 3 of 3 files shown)


src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -142,6 +142,18 @@ class MultiClockSpec extends ChiselFlatSpec {
       assert(withReset(this.reset) { 5 } == 5)
     })
   }
+  it should "support literal Bools" in {
```
ducky64: This test seems kind of random, I assume this fix is a side effect of the above change?

jackkoenig: #763 was the motivating issue so I added a test for it :) In looking into that issue, I realized that literal Bools not working in `withReset` was a side effect of the weirdness that this PR fixes

jackkoenig:
@ducky64 Yeah I think this will need to be done slightly differently in #820. Since the current LitBinding doesn't have any information about the literal value, I think this implementation is the right one in the meantime. I agree though that with Bundle literals it makes more sense to set the ref as part of the Binding.

PR #829, opened 2018-06-07
Direct to firrtl

jackkoenig:
I think this still needs a little work, in particular the function converting `Command`s to `Statement`s is a bit of a beast. I would appreciate suggestions on how to improve it! I'll provide some benchmarking numbers soon so that we can ensure this doesn't hurt performance (although for larger designs, skipping the parser should only help).

I should note that the "Converter" as currently implemented is kind of a transliteration of the existing Emitter. I would greatly appreciate thoughts on improvements. It is slightly more verbose than the Emitter because some emission logic lives within the Chisel IR and at least initially I decided to keep all conversion logic contained.

**Impact**: API addition (no impact on existing code) -- You can now get FIRRTL IR directly in Chisel!

**Release Notes**
Directly translate from ChiselIR to FIRRTL without using the FIRRTL ANTLR Parser

### DIFF
--- a/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -52,6 +52,7 @@ private[chisel3] object Converter {
     case ModuleIO(mod, name) =>
       if (mod eq ctx.id) fir.Reference(name, fir.UnknownType)
       else fir.SubField(fir.Reference(mod.getRef.name, fir.UnknownType), name, fir.UnknownType)
+    case u @ ULit(n, UnknownWidth()) => fir.UIntLiteral(n, fir.IntWidth(u.minWidth))
     case ULit(n, w) => fir.UIntLiteral(n, convert(w))
     case slit @ SLit(n, w) => fir.SIntLiteral(n, convert(w))
       val unsigned = if (n < 0) (BigInt(1) << slit.width.get) + n else n
--- a/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -1,0 +1,237 @@
+// See LICENSE for license details.
+
+package chisel3.internal.firrtl
+import chisel3._
+import chisel3.core.SpecifiedDirection
+import chisel3.experimental._
+import chisel3.internal.sourceinfo.{NoSourceInfo, SourceLine, SourceInfo}
+import firrtl.{ir => fir}
+import chisel3.internal.throwException
+
+import scala.annotation.tailrec
+import scala.collection.immutable.{Queue}
+
+private[chisel3] object Converter {
+  // TODO modeled on unpack method on Printable, refactor?
+  def unpack(pable: Printable, ctx: Component): (String, Seq[Arg]) = pable match {
+    case Printables(pables) =>
+      val (fmts, args) = pables.map(p => unpack(p, ctx)).unzip
+      (fmts.mkString, args.flatten.toSeq)
+    case PString(str) => (str.replaceAll("%", "%%"), List.empty)
+    case format: FirrtlFormat =>
+      ("%" + format.specifier, List(format.bits.ref))
+    case Name(data) => (data.ref.name, List.empty)
+    case FullName(data) => (data.ref.fullName(ctx), List.empty)
+    case Percent => ("%%", List.empty)
+  }
+
+  def convert(info: SourceInfo): fir.Info = info match {
+    case _: NoSourceInfo => fir.NoInfo
+    case SourceLine(fn, line, col) => fir.FileInfo(fir.StringLit(s"$fn $line:$col"))
+  }
+
+  def convert(op: PrimOp): fir.PrimOp = firrtl.PrimOps.fromString(op.name)
+
+  def convert(dir: MemPortDirection): firrtl.MPortDir = dir match {
+    case MemPortDirection.INFER => firrtl.MInfer
+    case MemPortDirection.READ => firrtl.MRead
+    case MemPortDirection.WRITE => firrtl.MWrite
+    case MemPortDirection.RDWR => firrtl.MReadWrite
+  }
+
+  // TODO memoize?
+  // TODO move into the Chisel IR?
+  def convert(arg: Arg, ctx: Component): fir.Expression = arg match {
+    case Node(id) => convert(id.getRef, ctx)
+    case Ref(name) => fir.Reference(name, fir.UnknownType)
+    case Slot(imm, name) => fir.SubField(convert(imm, ctx), name, fir.UnknownType)
+    case Index(imm, ILit(idx)) =>
+      fir.SubIndex(convert(imm, ctx), idx.toInt, fir.UnknownType)
+    case Index(imm, value) =>
+      fir.SubAccess(convert(imm, ctx), convert(value, ctx), fir.UnknownType)
+    case ModuleIO(mod, name) =>
+      if (mod eq ctx.id) fir.Reference(name, fir.UnknownType)
+      else fir.SubField(fir.Reference(mod.getRef.name, fir.UnknownType), name, fir.UnknownType)
+    case ULit(n, w) => fir.UIntLiteral(n, convert(w))
+    case slit @ SLit(n, w) => fir.SIntLiteral(n, convert(w))
+      val unsigned = if (n < 0) (BigInt(1) << slit.width.get) + n else n
+      val uint = convert(ULit(unsigned, slit.width), ctx)
+      fir.DoPrim(firrtl.PrimOps.AsSInt, Seq(uint), Seq.empty, fir.UnknownType)
+    // TODO Simplify
+    case fplit @ FPLit(n, w, bp) =>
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  // alt indicates to a WhenEnd whether we're closing an alt or just a regular when
+  // TODO we should probably have a different structure in the IR to close elses
+  private case class WhenFrame(when: fir.Conditionally, outer: Queue[fir.Statement], alt: Boolean)
+
+  // Whens markers are flat so scope must be inferred
+  // TODO refactor
+  def convert(cmds: Seq[Command], ctx: Component): fir.Statement = {
+    @tailrec
+    def rec(acc: Queue[fir.Statement],
```
chick: What does `rec` mean?

chick: You've probably considered this but could you create a trait like HasToStatement that most of the Commands could extend. Then the convert of command to statement could handle most of the cases with 
```
  case m: HasToStatement =>
     rec(acc :+ m.toStatement, scope)(cmds.tail)
```
It changes the encapsulation of information, since for example, DefWire has to know how to make a fir.DefWire but it sure cleans up the massive match. I can see it's a bit ugly though since toStatement would need access to the Converter object for info, clocks etc.

jackkoenig: Yeah I have thought about it a bit but your point is well taken and I ought to share some of my thinking.

In prototyping this, I wanted to touch as little code as possible which is why the Converter is wholly self-contained, I am open to some amount of encapsulation though. `toStatement` would have to take the `Component` as an argument, but that's how `def fullName` currently works for converting `Args` to `Expressions`.

A big question related to performance I have that goes back to `serialize` in FIRRTL is whether large match statements or inheritance are better for performance. Serialization functions in this case are almost certainly megamorphic which is generally terrible for performance, but when we moved serialize from a match statement to a method on `FIRRTLNode`, it still got faster. Perhaps the dynamic method lookup in megamorphic state is still faster in practice than large match blocks. I'm really not sure.

jackkoenig:
Currently the performance of this is terrible because the `Seq` of `Commands` is an `ArrayBuffer`, so repeatedly calling `.tail` on it is super bad. Simple solution is to convert to `List` before recursing, considering other options as well.

src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -0,0 +1,238 @@
...
+    case lit: ILit => throwException(s"Internal Error! Unexpected ILit: $lit")
+  }
+
+  // alt indicates to a WhenEnd whether we're closing an alt or just a regular when
+  // TODO we should probably have a different structure in the IR to close elses
+  private case class WhenFrame(when: fir.Conditionally, outer: Queue[fir.Statement], alt: Boolean)
+
+  // Whens markers are flat so scope must be inferred
+  // TODO refactor
+  def convert(cmds: Seq[Command], ctx: Component): fir.Statement = {
```
ducky64: There's quite a bit of code here, can you write a summary of the algorithm in comments?
This mostly seems to be a straightforward 1:1 map of Chisel to FIRRTL types except with some funny stuff for whens.

src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -0,0 +1,238 @@
...
+          rec(acc :+ mem, scope)(cmds.tail)
+        case e: DefMemPort[_] =>
+          val port = firrtl.CDefMPort(convert(e.sourceInfo), e.name, fir.UnknownType,
+            e.source.fullName(ctx), Seq(convert(e.index, ctx), convert(e.clock, ctx)), convert(e.dir))
+          rec(acc :+ port, scope)(cmds.tail)
+        case Connect(info, loc, exp) =>
+          val con = fir.Connect(convert(info), convert(loc, ctx), convert(exp, ctx))
+          rec(acc :+ con, scope)(cmds.tail)
+        case BulkConnect(info, loc, exp) =>
+          val con = fir.PartialConnect(convert(info), convert(loc, ctx), convert(exp, ctx))
```
ducky64: Are FIRRTL PartialConnect semantics equivalent to Chisel BulkConnect semantics? Or should this exception out because this should never be seen?

src/main/scala/chisel3/internal/firrtl/Converter.scala
jackkoenig: Yeah this definitely needs a summary, will add

src/main/scala/chisel3/internal/firrtl/Converter.scala
jackkoenig: FIRRTL PartialConnect semantics are closer to Chisel2 (and the current `Chisel._` `<>`) semantics actually. It's still used by the compatibility wrapper.

jackkoenig:
I guess I should write down some of my thinking on this and let y'all know where I'm going with this:

It appears that the current ANTLR Parser for FIRRTL is the main culprit in the high memory use for FIRRTL. There are two (obvious) possible solutions:

1. Reduce the parser memory use
1. Don't use the parser

I spent some time trying to improve the parser (see https://github.com/freechipsproject/firrtl/pull/824), but due to past experience and pain with the parser, I've been working on number 2. For this, I saw two main ways to do it:

1. Convert from Chisel to FIRRTL in memory and invoke FIRRTL in the same JVM
1. Use ProtoBuf (or other library) for serialization and deserialization

I originally was working on number 1 (see this PR), but it turns out to be a pretty complex game of wack-a-mole with static references to Chisel objects that prevent garbage collection (harming memory use substantially). To illustrate, using a large design, I get the following memory stats:

* Elaborated Chisel Objects: 16 GB
* FIRRTL Representation: 3.5 GB

When I invoke them in the same JVM, the elaborated Chisel graph sticks around the whole time. It's probably possible to fix them all, but due to this issue as well as the general notion that it is simpler on the build system to keep Chisel and FIRRTL split, I decided to refocus on number 2. I am in the process of providing a translator from FIRRTL to ProtoBuf. The plan is to just implement FIRRTL<->Protobuf and use this PR to translate Chisel to FIRRTL for serialization.

For the record, the Chisel object is still massive but I suspect there are easy gains there. Even the FIRRTL representation has plenty of room for improvement.

jackkoenig:
retest this please

src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  def convert(width: Width): fir.Width = width match {
+    case UnknownWidth() => fir.UnknownWidth
+    case KnownWidth(value) => fir.IntWidth(value)
+  }
+
+  def convert(bp: BinaryPoint): fir.Width = bp match {
+    case UnknownBinaryPoint => fir.UnknownWidth
+    case KnownBinaryPoint(value) => fir.IntWidth(value)
+  }
```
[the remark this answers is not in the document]
jackkoenig: Yeah that was my thinking, a later refactor can clean a lot of stuff related to this and the Emitter

src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -0,0 +1,238 @@
...
+      val uint = convert(ULit(unsigned, slit.width), ctx)
+      fir.DoPrim(firrtl.PrimOps.AsSInt, Seq(uint), Seq.empty, fir.UnknownType)
+    // TODO Simplify
+    case fplit @ FPLit(n, w, bp) =>
+      val unsigned = if (n < 0) (BigInt(1) << fplit.width.get) + n else n
+      val uint = convert(ULit(unsigned, fplit.width), ctx)
+      val lit = bp.asInstanceOf[KnownBinaryPoint].value
+      fir.DoPrim(firrtl.PrimOps.AsFixedPoint, Seq(uint), Seq(lit), fir.UnknownType)
+    case lit: ILit => throwException(s"Internal Error! Unexpected ILit: $lit")
+  }
```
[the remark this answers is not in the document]
jackkoenig: Yeah I agree the density is pretty tough. I ended up deciding to have all of the cases be multi-line. The matches were so different in length that alignment on `=>` wasn't working very well

jackkoenig:
I've refactored the style of the convert method for Commands to Statements (since that is by far the densest part). I split out conversion for Commands that map 1:1 to Statements from the when scope handling logic which is much more complicated. I also added clarifying comments to describe the basic process and noted things that I feel are non-obvious (or inconsistent). I also added `Driver.dumpProto` (which is why this is currently failing, it will pass once https://github.com/freechipsproject/firrtl/pull/832 is merged).

I have also posted performance numbers in https://github.com/freechipsproject/firrtl/pull/832. I backed out of using FIRRTL serialize because it is like 4x slower than the Chisel Emitter (as noted in https://github.com/freechipsproject/firrtl/issues/838). However, the converter is still used by `Driver.execute` and `Driver.dumpProto`. I think speeding up FIRRTL serialization so that the Chisel Emitter can be removed is desirable but beyond the scope of this PR

jackkoenig:
This is ready to go

PR #838, opened 2018-06-22
Add instance inline API

### LINKED ISSUE #602: Chisel API for Inlining Module Instances
Firrtl has the transformation and annotation, just need to add a Chisel API for it.

seldridge:
This adds a circuit/module/instance inlining and flattening API. This adds three new traits:
  - `InlineInstance` that will inline any Module/Instance it's mixed into
  - `FlattenInstance` that will flatten any Module/Instance it's mixed into

This provides Chisel-level access to existing FIRRTL annotations (`InlineAnnotation` and `FlattenAnnotation`) that map to existing FIRRTL passes/transforms (`InlineInstances` and `Flatten`).

**Related issue**: 
Resolves #602

**Release Notes**

- Add instance inlining and flattening API

### DIFF
--- a/src/test/scala/chiselTests/InlineSpec.scala
+++ b/src/test/scala/chiselTests/InlineSpec.scala
@@ -1,0 +1,57 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import chisel3.util.experimental.InlineInstance
+import chisel3.internal.firrtl.Circuit
+import firrtl.FirrtlExecutionSuccess
+import firrtl.passes.InlineAnnotation
+import firrtl.annotations.Annotation
+import firrtl.analyses.InstanceGraph
+import firrtl.{ir => fir}
+import firrtl.WDefInstance
+import firrtl.Mappers._
+import org.scalatest.{FreeSpec, Matchers}
+
+class InlineSpec extends FreeSpec with ChiselRunners with Matchers {
+
+  trait Internals { this: Module =>
+    val io = IO(new Bundle{ val a = Input(Bool()) })
+  }
+  class Sub extends Module with Internals
+  trait HasSub { this: Module with Internals =>
+    val sub = Module(new Sub)
+    sub.io.a := io.a
+  }
+
+  class Foo extends Module with Internals with InlineInstance with HasSub
+  class Bar extends Module with Internals with HasSub
+  class Baz extends Module with Internals with HasSub
+  class Qux extends Module with Internals with HasSub
+
+  def collectInstances(c: fir.Circuit, top: Option[String] = None): Seq[String] = new InstanceGraph(c)
+    .fullHierarchy.values.flatten.toSeq
+    .map( v => (top.getOrElse(v.head.name) +: v.tail.map(_.name)).mkString(".") )
+
+  "Module Inlining" - {
+    class Top extends Module with Internals {
+      val x = Module(new Foo)
+      val y = Module(new Bar with InlineInstance)
+      val z = Module(new Bar)
+      Seq(x, y, z).map(_.io.a := io.a)
+    }
+    "should compile to low FIRRTL" - {
+      Driver.execute(Array("-X", "low", "--target-dir", "test_run_dir"), () => new Top) match {
+        case ChiselExecutionSuccess(Some(chiselCircuit), _, Some(firrtlResult: FirrtlExecutionSuccess)) =>
+          "emitting TWO InlineAnnotation at the CHIRRTL level" in {
+            chiselCircuit.annotations.map(_.toFirrtl).collect{ case a: InlineAnnotation => a }.size should be (2)
+          }
+          "low FIRRTL should contain only instance z" in {
+            val instances = collectInstances(firrtlResult.circuitState.circuit, Some("Top")).toSet
+            Set("Top", "Top.x$sub", "Top.y$sub", "Top.z", "Top.z.sub") should be (instances)
+          }
+      }
+    }
+  }
+}
--- a/src/test/scala/chiselTests/InlineSpec.scala
+++ b/src/test/scala/chiselTests/InlineSpec.scala
@@ -3,11 +3,12 @@
 package chiselTests
 
 import chisel3._
-import chisel3.util.experimental.InlineInstance
+import chisel3.util.experimental.{InlineInstance, FlattenInstance}
 import chisel3.internal.firrtl.Circuit
 import firrtl.FirrtlExecutionSuccess
 import firrtl.passes.InlineAnnotation
 import firrtl.annotations.Annotation
+import firrtl.transforms.FlattenAnnotation
 import firrtl.analyses.InstanceGraph
 import firrtl.{ir => fir}
 import firrtl.WDefInstance
@@ -50,6 +51,25 @@ class InlineSpec extends FreeSpec with ChiselRunners with Matchers {
           "low FIRRTL should contain only instance z" in {
             val instances = collectInstances(firrtlResult.circuitState.circuit, Some("Top")).toSet
             Set("Top", "Top.x$sub", "Top.y$sub", "Top.z", "Top.z.sub") should be (instances)
+          }
+      }
+    }
+  }
+
+  "Module Flattening" - {
+    class Top extends Module with Internals {
+      val x = Module(new Qux with FlattenInstance)
+      x.io.a := io.a
+    }
+    "should compile to low FIRRTL" - {
+      Driver.execute(Array("-X", "low", "--target-dir", "test_run_dir"), () => new Top) match {
+        case ChiselExecutionSuccess(Some(chiselCircuit), chirrtl, Some(firrtlResult: FirrtlExecutionSuccess)) =>
+          "emitting ONE FlattenAnnotation at the CHIRRTL level" in {
+            chiselCircuit.annotations.map(_.toFirrtl).collect{ case a: FlattenAnnotation => a }.size should be (1)
+          }
+          "low FIRRTL should contain instance x only" in {
+            val instances = collectInstances(firrtlResult.circuitState.circuit, Some("Top")).toSet
+            Set("Top", "Top.x") should be (instances)
           }
       }
     }
--- a/src/main/scala/chisel3/util/experimental/Inline.scala
+++ b/src/main/scala/chisel3/util/experimental/Inline.scala
@@ -1,0 +1,50 @@
+// See LICENSE for license details.
+
+package chisel3.util.experimental
+
+import chisel3._
+import chisel3.internal.InstanceId
+import chisel3.experimental.{BaseModule, ChiselAnnotation, RunFirrtlTransform}
+import firrtl.Transform
+import firrtl.passes.{InlineAnnotation, InlineInstances}
+import firrtl.transforms.{NoDedupAnnotation, FlattenAnnotation, Flatten}
+import firrtl.annotations.{CircuitName, ModuleName, ComponentName, Annotation}
+
+/** Inlines an instance of a module
+  *
+  * @example {{{
+  * trait Internals { this: Module =>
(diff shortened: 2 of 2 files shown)


src/main/scala/chisel3/util/Inline.scala
```diff
@@ -0,0 +1,96 @@
...
+  *  class Top extends Module {
+  *    val io = IO(new Bundle{ val a = Input(Bool()) })
+  *    val x = Module(new Foo)                     // x will be inlined
+  *    val y = Module(new Bar with InlineInstance) // y will also be inlined
+  *    val z = Module(new Bar)                     // z will not be inlined
+  *    Seq(x, y, z).map(_.io.a := io.a)
+  *  }
+  * }}}
+  */
+trait InlineInstance { self: BaseModule =>
```
jackkoenig: I think the trait is useful to inline all instances of a given Module, do you think we could also have a way to inline a single instance?

jackkoenig: Oh, I see you do it in the example! Nice :)

src/main/scala/chisel3/util/Inline.scala
```diff
@@ -0,0 +1,96 @@
...
+    *   Seq(x, y, z).map(_.io.a := io.a)
+    *   /* Only the insance x will be inlined */
+    *   inlineComponent(x)
+    * }
+    * }}}
+    */
+  def inlineComponent(m: BaseModule): Unit = chisel3.experimental.annotate(
+    new ChiselAnnotation with RunFirrtlTransform {
+      def toFirrtl = InlineAnnotation(ComponentName(m.instanceName, self.toNamed))
+      def transformClass = classOf[InlineInstances] }
```
jackkoenig: What is the utility of this alternative way of specifying that an instance should be inlined? I guess you could make a submodule in code you don't control or don't want to modify inline an instance.

src/test/scala/chiselTests/InlineSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+class InlineSpec extends FreeSpec with ChiselRunners with Matchers {
+
+  trait Internals { this: Module =>
+    val io = IO(new Bundle{ val a = Input(Bool()) })
+  }
+
+  class Foo extends Module with Internals with InlineHelpers with InlineInstance
+  class Bar extends Module with Internals with InlineHelpers
+
+  def collectInstances(c: fir.Circuit): Array[WDefInstance] = {
```
jackkoenig: Why Array? It's not all that important but usually we just use `Seq`

src/test/scala/chiselTests/InlineSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+        case i: WDefInstance => insts += i
+          s
+        case _ =>
+          s
+      }
+      m map onStmt
+      m
+    }
+    c map onModule
+    insts.toArray.map{ case WDefInstance(_, n, m, _) => WDefInstance(fir.NoInfo, n, m, fir.UnknownType) }
```
jackkoenig: Alternatively:
```scala
insts.toArray.map(inst => inst.copy(info = fir.NoInfo, tpe = fir.UnknownType))
```

src/main/scala/chisel3/util/Inline.scala
```diff
@@ -0,0 +1,96 @@
...
+    *   val io = IO(new Bundle{ val a = Input(Bool()) })
+    *   val Seq(x, y, z) = Seq.fill(3)(Module(new Bar))
+    *   Seq(x, y, z).map(_.io.a := io.a)
+    *   /* All modules in the entire circuit will be inlined. Here, that is x, y,
+    *    * and z. */
+    *   inlineCircuit()
+    * }
+    * }}}
+    */
+  def inlineCircuit(): Unit = chisel3.experimental.annotate(
```
jackkoenig: This one feels kind of dangerous. I dunno how I feel about an arbitrary point in the Module hierarchy being able to instigate inlining *all* modules. I would maybe make `InlineCircuit` as `trait` and have it error unless it's mixed in to the top module. Alternatively, I think `Flatten` in FIRRTL inlines everything below a given module? Maybe make a `FlattenHierarchy` trait that can be mixed in to flatten everything below and thus Inlining the whole circuit can be accomplished by mixing that in at the top?

seldridge: I personally don't like it either... This PR only includes that because the FIRRTL inlining annotations support a `CircuitName` that will flatten the whole thing. I was struggling with a good API for that at the Chisel level (and will usually aim to expose whatever FIRRTL API exists at the Chisel level). I like the approach of requiring that this be mixed into the top module if that is the desired effect. Or: that's much better than "spooky inlining at a distance..."

This does bring up a point, which did not occur to me, that a Chisel flattening API should be a part of the inlining API (since they're both frontends for `InlineInstances`). Are you in agreement that that would round out this PR?

jackkoenig: Let me check that I'm understanding correctly but I think I agree.

By Chisel flattening API do you mean some API in Chisel (eg. `trait FlattenHierarchy`) to flatten everything below the given module? And in adding that, we would remove `def inlineCircuit()`?

seldridge: (Sorry, dropped off on this...)

Yes, I mean an API for flattening everything below a given module (an annotator that emits a `firrtl.transforms.FlattenAnnotation` which will run `firrtl.transforms.Flatten` should be all that's needed). I'm fine with removing the `inlineCircuit` API as it seems at worst dangerous (somebody uses this in their library) and at best confusing (SO posts asking, "Why is my circuit inlined?"). The initial motivation for `inlineCircuit` was only to fully expose the entire FIRRTL API of `firrtl.passes.Inline`, but it's far better to expose a usable sane Chisel API.

src/main/scala/chisel3/util/Inline.scala
seldridge: Had to think about this as I didn't have good justification originally. If a library writer chooses to only expose a wrapped `Module` via a factory then this would be the only way to inline that instance. Consider:
```scala
object BarFactory {
  def apply(): Bar = Module(new Bar)
}
class Top extends Module with InlineHelpers {
  val io = IO(new Bundle{ val a = Input(Bool()) })
  val Seq(x, y, z) = Seq.fill(3)(BarFactory())
  Seq(x, y, z).map(_.io.a := io.a)
  inlineComponent(x)
}
```
There's nothing that I'm aware of where this type of API is the only one that is exposed (something like `chisel3.util.Queue` is generally used only via factory, but the underlying class is public). Nonetheless, there may be some motivation for only exposing a factory and causing that.

Do you think this is worth including or is the case here too pathological?

src/test/scala/chiselTests/InlineSpec.scala
seldridge: Was due to the internal use of a mutable `ArrayBuffer`, but that gets converted automatically. I'll switch the internal to a `ListBuffer` as that's all that needed and return a `Seq`. Consistency is good...

src/main/scala/chisel3/util/Inline.scala
edwardcwang: As a user I think it would be good to be able to inline a module created in a 3rd-party library without having to modify that library (promotes re-use instead of fork/copy-paste).

src/test/scala/chiselTests/InlineSpec.scala
seldridge: Much cleaner, thanks! I can even shave off the `inst => inst.copy` --> `_.copy` ...

seldridge:
@jackkoenig: this should be good for another look. I added a trait and a method for accessing `firrtl.transforms.Flatten` as it's natural to include this here (since flattening is a wrapper around `firrtl.passes.InlineInstances`) and it seems to have been implicitly needed from your comments about circuit inlining. Related to circuit inlining, I removed the spooky `inlineCircuit` method.

src/main/scala/chisel3/util/Inline.scala
jackkoenig: Sorry I missed this 10 days ago.

I agree that we would like to be able to inline modules constructed in factories methods, but in practice this won't work for that either. As in the case of `Queue`, factory methods for constructing modules typically don't return the actual module reference itself, rather they return an interface (eg. see [object Queue](https://github.com/freechipsproject/chisel3/blob/98177d3647e51efe869de1031872adb6406fa34d/src/main/scala/chisel3/util/Decoupled.scala#L276)).

I think we need to come up with another way to support that in the general case. One idea I (or @aswaterman, I can't remember who proposed it) had was to provide a mechanism similar to `withClock` where you could scope things below to be inlined. There could be something like `inlineInstances { }` or the like. Unfortunately, this would require some form of support from Chisel, either natively or some API to get references to the Modules instantiated within a block.

In any case, I think we should remove this API and just have the mixin.

src/main/scala/chisel3/util/Inline.scala
```diff
@@ -0,0 +1,152 @@
...
+    *   //   - Top.y.bar
+    * class Top extends Module with Internals with InlineHelpers {
+    *   val x = Module(new Foo)
+    *   val y = Module(new Foo)
+    *   Seq(x, y).map(_.io.a := io.a)
+    *   flattenComponent(x)
+    * }
+    * }}}
+    */
+  def flattenComponent(m: BaseModule): Unit = chisel3.experimental.annotate(
```
jackkoenig: I don't love this API. It's not the worst thing in the world (it's similar to dontTouch) so I don't think it's worth blocking over but similarly to prefering some sort of `inlineInstances { }`-like API maybe there could be a `flattenSubModules { }` or something, of course it would require support in Chisel core

jackkoenig: If we keep it, I think we should make it return the argument itself and probably rename it to `flattenHierarchy` or something which IMO is more clear about what it's doing

src/main/scala/chisel3/util/Inline.scala
seldridge: Yeah, `Queue` was only intending to be illustrative of how a user might write a library (and hopefully pass back the module as opposed to the IO). I like the idea of `inlineInstances { }`, but yes, that's likely out of scope here.

As API deprecation is such a hot topic, I think it's best to err with new APIs on the side of caution. I'll remove this.

src/main/scala/chisel3/util/Inline.scala
seldridge: I'll remove this as well. If a user wants this behavior, they can write their own annotator. Otherwise, we should be absolutely certain that the API we expose publicly is the one we want to support. Keeping this back in favor of a future _better_ API makes more sense than pushing this out and then deprecating or supporting both.

jackkoenig:
retest this please

seldridge:
No clue what's going on with Jenkins. Is it applying more stringent scalastyle checks than are in the public XML? Since I was taking a look at this anyway to make sure it was passing tests/scalastyle, I went ahead and rebased it into two logical commits that both build.

One question: should we move this into `chisel3.util.experimental` like `BoringUtils`?

seldridge:
Sounds good to me. Taking care of the rebase now.

ucbjrl:
Jenkins checkstyle is configured to complain with 40 or more style violations.

image

image

seldridge:
@ucbjrl, thanks! That clarifies things!

jackkoenig:
@seldridge Do you want to rebase this (again)? Looks like another PR sneaked in ahead

seldridge:
@jackkoenig: rebased and building. Thanks for the ping. It would be nice if GitHub had a "rebase and then create a merge commit" option...

jackkoenig:
Yeah for real, or if CI could rerun a PR against the new master automatically so that it would be fully safe to just do a merge

PR #842, opened 2018-06-30
Catch returns from within when blocks and provide an error message

### LINKED ISSUE #841: Return in When statement causes exciting bugs and error messages
**Type of issue**: bug report/feature request

**Other information**
 

Code sample:
```scala
def safeAccess(): UInt = {
  when(cacheUsable) {
    return reqCache.data
  }
  return io.req.data.bits
}
```
Currently, this fails during elaboration, with the error message "Internal Error! When depth is != 0, this should not be possible".
This can be fixed with the below, under "Expected behaviour".

There are two possibilities: Either this is the wrong way to do it, and the correct way is:
```scala
def safeAccess(): UInt = {
  val temp = Wire(UInt())
  when(cacheUsable) {
    temp := reqCache.data
  } .otherwise {
    temp := io.data.bits
  }
  temp
}
```
or this should be allowed. (Is a Mux the correct option? This should compile to the same as one)
If this is the wrong way (which I think is likely), then it would be helpful to have a better error message and/or guidance that this is the wrong way to do it.

a6

Easier use of functions for convenience.

jackkoenig:
Resolves #841 

It turns out that non-local returns are an exception so we can catch this 💯 

**Related issue**:  #841 

**Type of change**: bug fix

**Impact**: no functional change (better error message)

**Release Notes**

Returns within when scopes used to give opaque error messages. Now they are caught earlier (with a stack trace) and explained in the error message.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -49,7 +49,7 @@ object Module {
     val module: T = bc  // bc is actually evaluated here
 
     if (Builder.whenDepth != 0) {
-      throwException("Internal Error! \"When\" scope depth is != 0, this should have been caught!")
+      throwException("Internal Error! when() scope depth is != 0, this should have been caught!")
     }
     if (Builder.readyForModuleConstr) {
       throwException("Error: attempted to instantiate a Module, but nothing happened. " +
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -49,7 +49,7 @@ object Module {
     val module: T = bc  // bc is actually evaluated here
 
     if (Builder.whenDepth != 0) {
-      throwException("Internal Error! When depth is != 0, this should not be possible")
+      throwException("Internal Error! \"When\" scope depth is != 0, this should have been caught!")
     }
     if (Builder.readyForModuleConstr) {
       throwException("Error: attempted to instantiate a Module, but nothing happened. " +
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -49,7 +49,7 @@ object Module {
     val module: T = bc  // bc is actually evaluated here
 
     if (Builder.whenDepth != 0) {
-      throwException("Internal Error! When depth is != 0, this should not be possible")
+      throwException("Internal Error! \"When\" scope depth is != 0, this should have been caught!")
```
edwardcwang: If `"When"` is referring to the construct (unless I misunderstood it), then I would recommend `when() scope depth is != 0; this should have been caught!`.

chiselFrontend/src/main/scala/chisel3/core/When.scala
```diff
@@ -71,7 +71,14 @@ final class WhenContext(sourceInfo: SourceInfo, cond: Option[() => Bool], block:
   if (firrtlDepth > 0) { pushCommand(AltBegin(sourceInfo)) }
   cond.foreach( c => pushCommand(WhenBegin(sourceInfo, c().ref)) )
   Builder.whenDepth += 1
-  block
+  try {
+    block
+  } catch {
+    case ret: scala.runtime.NonLocalReturnControl[_] =>
+      throwException("Cannot exit from a \"when\" block with a \"return\"!" +
+        " Perhaps you meant to use Mux or a Wire as a return value?"
```
edwardcwang: :+1:

chiselFrontend/src/main/scala/chisel3/core/Module.scala
jackkoenig: Done

jackkoenig:
@ducky64 Yeah I was pretty stoked that we could give a stack trace

PR #846, opened 2018-07-06
Undeprecate log2Up and log2Down

jackkoenig:
They should not be deprecated until zero-width wires actually work. See https://github.com/freechipsproject/firrtl/pull/530.

Obviously the better solution is to make zero-width wires fully work but it's been a year and it hasn't happened. Lots of people are using chisel3 now and are pretty annoyed by the deprecation warnings when they really can't switch to `log2Ceil` and `log2Floor` yet.

**Related issue**: 

People have reported problems with this https://github.com/freechipsproject/chisel3/issues/652 and more have complained to me directly.

**Release Notes**

Undo the deprecation of log2Up and log2Down until zero-width wires work.

### DIFF
--- a/src/main/scala/chisel3/util/Math.scala
+++ b/src/main/scala/chisel3/util/Math.scala
@@ -10,8 +10,9 @@ import chisel3.internal.chiselRuntimeDeprecated
 
 /** Compute the log2 rounded up with min value of 1 */
 object log2Up {
-  @chiselRuntimeDeprecated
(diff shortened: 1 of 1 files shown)


ducky64:
Also maybe include a link to a tracking issue for zero width wires, so we can understand why these were undeprecated in the future, when we want to re-deprecate them?

jackkoenig:
Yeah I think that is sensible, I'll file a meta tracking issue and add a link to it to the comments added here

Issue #844, opened 2018-07-06
Cannot use Seq in Bundle construction?


edwardcwang:
It seems like Chisel just doesn't like it when we use Seq to construct Bundles. Two test cases (which may or may not be caused by the same code paths):

1) Using Seq to construct Input(UInt()):

```scala
class MyTest extends Bundle {
    val el = Seq(1, 4, 9).map(i => Input(UInt(i.W)))
    // val el = Seq(Input(UInt(1.W)), Input(UInt(4.W)), Input(UInt(9.W))) // equivalent
    override def cloneType = (new MyTest).asInstanceOf[this.type]
}

class MyTestMod extends Module {
  val io = IO(new Bundle {
    val test = new MyTest()
    val out = Output(UInt(12.W))
  })
  
  io.out := io.test.el(0) + io.test.el(1) + io.test.el(2)
}
```

Results in Chisel not being able to figure out the direction of `MyTest`:

```
chisel3.core.Binding$MixedDirectionAggregateException: Aggregate 'MyTestMod' can't have elements that are both directioned and undirectioned: Vector((chisel3.core.UInt@e,Output), (MyTest@6,Unspecified))
```

2) Using Seq to construct UInt() and then specifying the direction in the bundle above it (e.g. `Input(new MyTest())`):
```scala
class MyTest extends Bundle {
    val el = Seq(1, 4, 9).map(i => UInt(i.W))
    override def cloneType = (new MyTest).asInstanceOf[this.type]
}

class MyTestMod extends Module {
  val io = IO(new Bundle {
    val test = Input(new MyTest())
    val out = Output(UInt(12.W))
  })
  
  io.out := io.test.el(0) + io.test.el(1) + io.test.el(2)
}
```

Results in Chisel misclassifying the elements of `el` as bare Chisel types as opposed to hardware.
```
chisel3.core.Binding$ExpectedHardwareException: bits operated on 'chisel3.core.UInt@b' must be hardware, not a bare Chisel type
```

**Type of issue**: bug report

**Impact**: no functional change

**Development Phase**: request

**What is the current behavior?** See above

**What is the expected behavior?** The examples should construct hardware and elaborate properly as expected, without any error messages, or Chisel should give the user a clear error message if using Seq to construct Bundles is unsupported.

**Please tell us about your environment:** Chisel 3.1

**What is the use case for changing the behavior?** The error message is confusing to users (directions were specified from a user's perspective) and it is not immediately clear why this is causes a problem to begin with.

[...]

Issue #849, opened 2018-07-08
Vec literals


edwardcwang:
Now that a new literal framework is in place after #820, would it make sense to create a mechanism (possibly by co-opting `VecInit` or by creating a `VecLit` constructor) to create Vec literals? The current closest equivalent, `VecInit,` still creates a wire and assigns the literals to it the old way we worked around lack of bundle literals.

**Type of issue**: feature request

**Impact**: no functional change | API addition (no impact on existing code)

**Development Phase**: request | proposal

ducky64:
I think unifying literal constructor and init is something that needs more thought.

VecInit is a useful shorthand for packing elements into a Vec, but can't be used in Bundle literal constructors. Bundle literal constructors also can't take non-literal types (like VecInit results), but can be unpacked into Scala-land data (used for testers2).

I think there are use case arguments for both, though what the API looks like should be up for debate. `Vec(3, UInt(32.W)).Lit(..., ..., ...)`? Should there also be a Bundle equivalent of init, as a shorthand for packing Bundles, like `(new MyBundle(...)).Init(a=..., b=...)`? Can we do something about unifying the syntax, since the equivalent VecInit would look more like `Vec(3, UInt(32.W)).Init(..., ...)`? But VecInit can also determine the Vec type from its arguments, which Bundle Lit (or speculatively, init) can't do.

ducky64:
Writing up discussion from chisel-dev two weeks back:

Vector literals could have two APIs, one analogous to `Seq(...)` and another analogous to `.toSeq()`
```scala
Vec.Lit(0.U, 1.U, 2.U, 3.U, …)

val w = Wire(Vec(n, UInt(3.W)))
w := Vec.Lit(0.U, 1.U, 2.U, 3.U, …)

(0 until n).map(_.U).toVecLit
```

These might need to support some way to specify a base type, eg if you wanted a Vec of SInts. One proposal is, analogous to how we specify widths:
```scala
(0 until n).map(_.U).toVecLit(SInt(3.W))
```

But the `Vec.Lit` case is trickier, since there's less of an analogy. Perhaps analogous to Bundle literals, create the "prototype" object and then create a literal from it:
```scala
Vec(6, SInt(3.W)).Lit(0.U, 1.U, 2.U, 3.U, …)
```
though this overspecifies the Vec since the length is specified both explicitly in the `Vec(...)` prototype constructor, and then again implicitly in the `.Lit(...)` argument list.

chick:
[...]

PR #850, opened 2018-07-08
Heterogeneous Vec (MixedVec)

edwardcwang:
As discussed with @jackkoenig, this implements the action items which fell out of our discussion in #844.

This PR creates a new `HeterogeneousVec` type inspired by rocket-chip's `HeterogeneousBag` or `HetVec` from dsptools. This is a dynamically indexable type like `Vec` but supports different underlying types - e.g. `myVec(0)` could return a `UInt(8.W)` while `myVec(1)` could return `UInt(16.W)`, which is not supported by `Vec`, and as shown in #844 we cannot use the Seq alternative in Bundle construction as we can in Module construction.

Additionally, as discussed in #844, this PR also adds an (optionally-overridable) warning when public `Seq[Data]` members are present so that users can get a better error message than the current one. The ability to have non-Chisel `Data` `Seq`s are unaffected. (along with unit test demonstrating this)

As for the name, HVec is ambiguous (can be confused for Haskell HList or in VLSI can be confused for *hierarchical*) as noted in #844. `HetVec` is another option as noted in #844 as well but seems strange as a user (what is "Het"?). Another option that seems reasonable and short is `MixedVec`, which seems preferable.

For now the compromise is to make the class name `HeterogeneousVec` and add shorter aliases for `HetVec` and `MixedVec`. In addition, rocket-chip has been using the longer name `HeterogeneousBag` without much issue.

**Type of change**: bug report | feature request

**Release Notes**
- Add a new dynamically indexable sequence like Vec that supports different underlying widths/types.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -479,6 +479,21 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
   def toPrintable: Printable = toPrintableHelper(elements.toList)
 }
 
+/**
+  * Mix-in for Bundles that have arbitrary Seqs of Chisel types that aren't
+  * involved in hardware construction.
+  *
+  * Used to avoid raising an error/exception when a Seq is a public member of the
+  * bundle.
+  * This is useful if we those public Seq fields in the Bundle are unrelated to
+  * hardware construction.
+  */
+trait IgnoreSeqInBundle {
+  this: Bundle =>
+
+  override def ignoreSeq: Boolean = true
+}
+
 class AutoClonetypeException(message: String) extends ChiselException(message, null)
 
 /** Base class for data types defined as a bundle of other data types.
@@ -555,8 +570,8 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
                 case d: Data => throwException("Public Seq members cannot be used to define Bundle elements " +
                   s"(found public Seq member '${m.getName}'). " +
                   "Either use a Vec() if all elements are of the same type, or HeterogeneousVec() if the elements " +
-                  "are of different types. If this Seq member is not intended to construct RTL, override ignoreSeq " +
-                  "to be true.")
+                  "are of different types. If this Seq member is not intended to construct RTL, mix in the trait " +
+                  "IgnoreSeqInBundle.")
                 case _ => // don't care about non-Data Seq
               }
               case _ => // not a Seq
@@ -568,8 +583,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
   }
 
   /**
-    * Override this to be true to avoid raising an error/exception when a Seq is a public member of the bundle.
-    * This is useful if you have public Seq fields in the Bundle that are unrelated to hardware construction.
+    * Overridden by [[IgnoreSeqInBundle]] to allow arbitrary Seqs of Chisel elements.
     */
   def ignoreSeq: Boolean = false
 
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -539,17 +539,39 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     val nameMap = LinkedHashMap[String, Data]()
     val seen = HashSet[Data]()
     for (m <- getPublicFields(classOf[Bundle])) {
-      getBundleField(m) foreach { d =>
-        if (nameMap contains m.getName) {
-          require(nameMap(m.getName) eq d)
-        } else if (!seen(d)) {
-          nameMap(m.getName) = d
-          seen += d
-        }
+      getBundleField(m) match {
+        case Some(d: Data) =>
+          if (nameMap contains m.getName) {
+            require(nameMap(m.getName) eq d)
+          } else if (!seen(d)) {
+            nameMap(m.getName) = d
+            seen += d
+          }
+        case None =>
+          if (!ignoreSeq) {
+            m.invoke(this) match {
+              case s: scala.collection.Seq[Any] if s.nonEmpty => s.head match {
+                // Ignore empty Seq()
+                case d: Data => throwException("Public Seq members cannot be used to define Bundle elements " +
+                  s"(found public Seq member '${m.getName}'). " +
+                  "Either use a Vec() if all elements are of the same type, or HeterogeneousVec() if the elements " +
+                  "are of different types. If this Seq member is not intended to construct RTL, override ignoreSeq " +
+                  "to be true.")
+                case _ => // don't care about non-Data Seq
+              }
+              case _ => // not a Seq
+            }
+          }
       }
     }
     ListMap(nameMap.toSeq sortWith { case ((an, a), (bn, b)) => (a._id > b._id) || ((a eq b) && (an > bn)) }: _*)
   }
+
+  /**
+    * Override this to be true to avoid raising an error/exception when a Seq is a public member of the bundle.
+    * This is useful if you have public Seq fields in the Bundle that are unrelated to hardware construction.
+    */
+  def ignoreSeq: Boolean = false
 
   /** Returns a field's contained user-defined Bundle element if it appears to
     * be one, otherwise returns None.
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -569,7 +569,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
                 // Ignore empty Seq()
                 case d: Data => throwException("Public Seq members cannot be used to define Bundle elements " +
                   s"(found public Seq member '${m.getName}'). " +
-                  "Either use a Vec() if all elements are of the same type, or HeterogeneousVec() if the elements " +
+                  "Either use a Vec if all elements are of the same type, or MixedVec if the elements " +
                   "are of different types. If this Seq member is not intended to construct RTL, mix in the trait " +
                   "IgnoreSeqInBundle.")
                 case _ => // don't care about non-Data Seq
--- a/src/main/scala/chisel3/util/MixedVec.scala
+++ b/src/main/scala/chisel3/util/MixedVec.scala
@@ -77,18 +77,21 @@ final class MixedVec[T <: Data](private val eltsIn: Seq[T]) extends Record with 
 
   /**
     * Dynamically (via a mux) retrieve the element at the given index.
+    * This is implemented via a mux with the width of the widest element in this MixedVec.
+    * For example, a MixedVec of type Seq(UInt(4.W), UInt(8.W)) will create an 8-bit mux for this operation.
+    * Note: it is up to the user to process the resultant UInt (e.g. unflatten, etc).
     *
-    * @param index Index with which to retrieve.
-    * @return Retrieved index.
+    * @param index Index to retrieve. If the index is out of range, it will return the first element.
+    * @return Retrieved index as a UInt with the width of the widest element.
     */
-  def apply(index: UInt): T = {
+  def apply(index: UInt): UInt = {
     requireIsHardware(index, "index must be hardware")
 
     if (length < 1) {
       throw new IndexOutOfBoundsException("Collection is empty")
     }
 
-    MuxLookup(index, elts.head, elts.zipWithIndex.map { case (el, el_index) => el_index.U -> el })
+    MuxLookup(index, elts.head.asUInt, elts.zipWithIndex.map { case (el, el_index) => el_index.U -> el.asUInt })
   }
 
   /** Strong bulk connect, assigning elements in this MixedVec from elements in a Seq.
--- a/src/main/scala/chisel3/util/MixedVec.scala
+++ b/src/main/scala/chisel3/util/MixedVec.scala
@@ -8,7 +8,7 @@ import chisel3.internal.naming.chiselName
 
 import scala.collection.immutable.ListMap
 
-object MixedVecWireInit {
+object MixedVecInit {
   /**
     * Construct a new wire with the given bound values.
     * This is analogous to [[chisel3.core.VecInit]].
--- a/src/main/scala/chisel3/util/MixedVec.scala
+++ b/src/main/scala/chisel3/util/MixedVec.scala
@@ -1,0 +1,129 @@
+// See LICENSE for license details.
+
+package chisel3.util
+
+import chisel3._
+import chisel3.core.{Data, requireIsChiselType, requireIsHardware}
+import chisel3.internal.naming.chiselName
+
+import scala.collection.immutable.ListMap
+
+object MixedVecWireInit {
+  /**
+    * Construct a new wire with the given bound values.
+    * This is analogous to [[chisel3.core.VecInit]].
+    * @param vals Values to create a MixedVec with and assign
+    * @return MixedVec with given values assigned
+    */
+  def apply[T <: Data](vals: Seq[T]): MixedVec[T] = {
+    // Create a wire of this type.
+    val hetVecWire = Wire(MixedVec(vals.map(_.cloneTypeFull)))
+    // Assign the given vals to this new wire.
+    for ((a, b) <- hetVecWire.zip(vals)) {
+      a := b
+    }
+    hetVecWire
+  }
+}
+
+object MixedVec {
+  /**
+    * Create a MixedVec from that holds the given types.
+    * @param eltsIn Element types. Must be Chisel types.
+    * @return MixedVec with the given types.
+    */
+  def apply[T <: Data](eltsIn: Seq[T]): MixedVec[T] = new MixedVec(eltsIn)
+
+  /**
+    * Create a MixedVec from the type of the given Vec.
+    * For example, given a Vec(2, UInt(8.W)), this creates MixedVec(Seq.fill(2){UInt(8.W)}).
+    * @param vec Vec to use as template
+    * @return MixedVec analogous to the given Vec.
+    */
+  def apply[T <: Data](vec: Vec[T]): MixedVec[T] = {
+    MixedVec(Seq.fill(vec.length)(vec.sample_element))
+  }
+}
+
+/**
+  * A hardware array of elements that can hold values of different types/widths,
+  * unlike Vec which can only hold elements of the same type/width.
+  *
+  * @param eltsIn Element types. Must be Chisel types.
+  *
+  * @example {{{
+  * val v = Wire(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(32.W))))
+  * v(0) := 100.U(8.W)
+  * v(1) := 10000.U(16.W)
+  * v(2) := 101.U(32.W)
+  * }}}
+  */
+@chiselName
+final class MixedVec[T <: Data](private val eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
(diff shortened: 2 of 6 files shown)


src/main/scala/chisel3/util/HeterogeneousVec.scala
```diff
@@ -0,0 +1,121 @@
...
+  *
+  * @param eltsIn ...
+  *
+  * @example {{{
+  * val arb = Module(new Arbiter(UInt(), 2))
+  * arb.io.in(0) <> producer0.io.out
+  * arb.io.in(1) <> producer1.io.out
+  * consumer.io.in <> arb.io.out
+  * }}}
+  */
```
ucbjrl: I'm having trouble reconciling the previous comment with the following code.

edwardcwang: My bad - just wrote the actual scaladoc for the code.

src/main/scala/chisel3/util/util.scala
```diff
@@ -11,5 +11,6 @@ package object util {
   type ValidIO[+T <: Data] = chisel3.util.Valid[T]
   val ValidIO = chisel3.util.Valid
   val DecoupledIO = chisel3.util.Decoupled
-
+  val MixedVec = chisel3.util.HeterogeneousVec
+  val HetVec = chisel3.util.HeterogeneousVec
```
jackkoenig: I think we should settle on shorter name rather than just providing multiple

edwardcwang: Sure. I'm personally partial to `MixedVec`. I agree with you that `HVec` is overloaded and can be confusing, and `HetVec` still requires me as a user to stop and think about what is a "Het".

shunshou:
I think MixedVec is pretty clear :o

src/main/scala/chisel3/util/HeterogeneousVec.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /** Strong bulk connect, assigning elements in this HeterogeneousVec from elements in a Vec.
+    *
+    * @note the lengths of this and that must match
+    */
+  def :=(that: Vec[T]): Unit = {
+    require(this.length == that.length)
+    for ((a, b) <- this zip that)
+      a := b
+  }
```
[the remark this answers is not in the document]
edwardcwang: Good catch; will fix.

src/test/scala/chiselTests/HeterogeneousVecSpec.scala
```diff
@@ -0,0 +1,254 @@
...
+  property("It should be possible to bulk connect a HeterogeneousVec and a Seq") {
+    assertTesterPasses{ new HetVecConnectWithSeqTester }
+  }
+
+  property("HeterogeneousVecs of a single 1 bit element should compile and work") {
+    assertTesterPasses { new HetVecOneBitTester }
+  }
+
+  property("Connecting a HeterogeneousVec and something of different size should report a ChiselException") {
+    a[IllegalArgumentException] should be thrownBy {
```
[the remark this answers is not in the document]
edwardcwang: Will fix.

src/main/scala/chisel3/util/util.scala
jackkoenig: `MixedVec` sounds good to me, should we discuss in the developers meeting tomorrow?

jackkoenig: Works for me, I think I would just replace all `HeterogeneousVec` with `MixedVec`

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -539,18 +539,40 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+      getBundleField(m) match {
+        case Some(d: Data) =>
+          if (nameMap contains m.getName) {
+            require(nameMap(m.getName) eq d)
+          } else if (!seen(d)) {
+            nameMap(m.getName) = d
+            seen += d
+          }
+        case None =>
+          if (!ignoreSeq) {
```
ducky64: I understand why this check is here, but is there a cleaner way to implement the bypass than setting a global flag (which seems a bit unclean).

edwardcwang: I'm happy to take suggestions, if you have any. This is actually a method that a user can override in their bundle and is read-only, so I don't think this can cause any concurrency issues?

jackkoenig: It’s not a global flag right? It’s a per-Bundle flag which is fairly contained.

src/main/scala/chisel3/util/HeterogeneousVec.scala
```diff
@@ -0,0 +1,112 @@
...
+  */
+@chiselName
+case class HeterogeneousVec[T <: Data](private val eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
+  // We want to create HeterogeneousVec only with Chisel types.
+  if (compileOptions.declaredTypeMustBeUnbound) {
+    eltsIn.foreach(e => requireIsChiselType(e))
+  }
+
+  // Clone the inputs so that we have our own references.
+  val elts: Seq[T] = eltsIn.map(_.cloneTypeFull)
```
jackkoenig: We need to make elts an `IndexedSeq`

src/main/scala/chisel3/util/HeterogeneousVec.scala
```diff
@@ -0,0 +1,112 @@
...
+  *
+  * @example {{{
+  * val v = Wire(HeterogeneousVec(Seq(UInt(8.W), UInt(16.W), UInt(32.W))))
+  * v(0) := 100.U(8.W)
+  * v(1) := 10000.U(16.W)
+  * v(2) := 101.U(32.W)
+  * }}}
+  */
+@chiselName
+case class HeterogeneousVec[T <: Data](private val eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
```
jackkoenig: This should not be a `case class` but probably be final.

`case class` has lots of implications about equality and immutability that really don't fit with Chisel types

src/main/scala/chisel3/util/HeterogeneousVec.scala
edwardcwang: Will do.

jackkoenig: We also might consider making `elts` private? Since `MixedVec` extends `IndexedSeq` it can be used like one so I"m not sure there's any reason to expose the underlying representation as a public API

src/main/scala/chisel3/util/HeterogeneousVec.scala
jackkoenig: Generally allowing a class to be extended has a cost if we want to make any changes in the future. I think start with it as `final` and then if people ask for it to be make extendible we can evaluate it at a later time.

src/main/scala/chisel3/util/HeterogeneousVec.scala
edwardcwang: Will do.

src/main/scala/chisel3/util/HeterogeneousVec.scala
edwardcwang: Will do.

edwardcwang:
@jackkoenig Renaming, removal of aliases, and comments all addressed.

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,112 @@
...
+    * @return Retrieved index.
+    */
+  def apply(index: UInt): T = {
+    requireIsHardware(index, "index must be hardware")
+
+    if (length < 1) {
+      throw new IndexOutOfBoundsException("Collection is empty")
+    }
+
+    MuxLookup(index, elts.head, elts.zipWithIndex.map { case (el, el_index) => el_index.U -> el })
```
jackkoenig: This actually won't work, you can't create a mux with heterogeneous types. This is a big problem but it's not really clear how you can possibly handle dynamically indexing heterogeneous types...

edwardcwang: Hmm, would it be possible for you to elaborate on that? I wrote a test for how I would use it from a user's perspective (MixedVecDynamicIndexTester) and it seems to behave/work as expected.

jackkoenig: The issue with the test is that it only checks dynamic indexing when all of the elements are `UInts`, if you try dynamically indexing a `MixedVec` composed of a `UInt` and some other `Data` like `SInt` or `Bundle`, you will get an elaboration time exception

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,112 @@
...
+  *
+  * @example {{{
+  * val v = Wire(HeterogeneousVec(Seq(UInt(8.W), UInt(16.W), UInt(32.W))))
+  * v(0) := 100.U(8.W)
+  * v(1) := 10000.U(16.W)
+  * v(2) := 101.U(32.W)
+  * }}}
+  */
+@chiselName
+final case class MixedVec[T <: Data](private val eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
```
jackkoenig: I still think it should be a just a `class`, not a `case class`. `case class` *implies* things about equality that are not true about this type

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+  * A hardware array of elements that can hold values of different types/widths,
+  * unlike Vec which can only hold elements of the same type/width.
+  *
+  * @param eltsIn Element types. Must be Chisel types.
+  *
+  * @example {{{
+  * val v = Wire(HeterogeneousVec(Seq(UInt(8.W), UInt(16.W), UInt(32.W))))
```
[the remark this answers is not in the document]
edwardcwang: Done

src/main/scala/chisel3/util/MixedVec.scala
edwardcwang: Okay, will do.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ducky64: Sorry, global was bad terminology, I meant throwing flags in the body of the instance as override vals. I don't think this is something we've done elsewhere, and while it's mitigated by clear time-of-use documentation (which is great!), it still feels odd. I don't have any solutions right now, but it would be worth brainstorming alternatives at a future meeting.

src/main/scala/chisel3/util/MixedVec.scala
edwardcwang: Thanks for the catch Jack. I thought about it some more, and came up with a solution that seems to work for heterogeneous elements as long as the user is careful about it, and added a unit test to demonstrate/test it.

Since Chisel's mux already creates a result of the maximum width (e.g. muxing `UInt(3.W)` and `UInt(8.W)` -> `UInt(8.W)`), so we just return a UInt of the maximum width and let the designer handle packing/unpacking/etc. I think this interpretation makes the most sense in light of dynamic indexing as a hardware construction operation.

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
jackkoenig: Yeah I agree that overriding a function isn't ideal. What if we made it a trait you could mixin instead?

edwardcwang: Done.

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,122 @@
...
+    */
+  def apply(index: UInt): UInt = {
+    requireIsHardware(index, "index must be hardware")
+
+    if (length < 1) {
+      throw new IndexOutOfBoundsException("Collection is empty")
+    }
+
+    MuxLookup(index, elts.head.asUInt, elts.zipWithIndex.map { case (el, el_index) => el_index.U -> el.asUInt })
+  }
```
jackkoenig: We discussed this in the meeting, but let's see if we can make this return `T` and dynamically pick the right concrete type with `UInt` being the default for full Heterogeneous types. We should then have a test for a MixedVec of all UInts, all SInts, all (the same) Bundle, and then a mix. That or we can remove this, merge the PR, and do that later

edwardcwang: Hmm, would you have any suggestions on how to do that? I thought about pattern matching to handle the special cases but we can't match on type parameters (i.e. `T match { ... }` isn't valid).

Originally I tried returning T at one point and the problem occurs, as you mentioned in a previous comment, when heterogeneous types get mixed. It would be okay if I could detect if `T` isn't all UInt/SInt/same bundle, but I don't see an obvious way to do that.

jackkoenig: You can do it with ClassTags: https://www.scala-lang.org/api/2.12.3/scala/reflect/ClassTag.html

I'm not 100% on how to do it but that's the basic approach.

I think it might be worth removing the dynamic index for now and merging MixedVec without it since it's useful for IOs and dynamic indexing can be shimmed with `VecInit(myMixedVec)(idx)` (I *think*).

edwardcwang: One concern here is that if the MixedVec is composed of 2 different bundles, then their common supertype would be `Bundle`, and I think we get a `ClassCastException` if we try to cast a generic a generic UInt to it.

Perhaps we could keep both a UInt-returning version as well as a T-returning version (possibly Option[T]-returning) with different names and documentation? It seems unlikely that we can cover all cases by returning just T, and in those cases it might still be useful to have a mux function, I think.

ducky64: I'm also struggling to come up with a good and clean solution - perhaps that indicates a deeper problem? Returning a concrete supertype (like Bundle) that carries data / bits would require a major overhaul of behavior, and returning UInt seems special-case and unclean. It might could also be unintuitive ("wait, how am I getting a UInt when this vec of full of Bundle types?!")

I'm trying to think of what the user wants when dynamically indexing such a vec, would using a regular vec with tagged unions (which don't exist yet but are a requested feature) be a cleaner solution?

Anyways, I'm also leaning towards removing this until we can figure out a clean API. It's easier to add features than to remove them, especially if there are alternative (though potentially clunkier) pathways available currently.

edwardcwang: Alright, done.

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,103 @@
...
+
+package chisel3.util
+
+import chisel3._
+import chisel3.core.{Data, requireIsChiselType, requireIsHardware}
+import chisel3.internal.naming.chiselName
+
+import scala.collection.immutable.ListMap
+
+object MixedVecWireInit {
```
ducky64: This seems like a lot of words. Why not MixedVecInit? Note that VecInit also internally creates a wire.

Also, thoughts on a varargs version vs. the Seq version? In the comments for VecInit, it appears that we would have preferred the varargs version for consistency with Scala.

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+object MixedVec {
+  /**
+    * Create a MixedVec from that holds the given types.
+    * @param eltsIn Element types. Must be Chisel types.
+    * @return MixedVec with the given types.
+    */
+  def apply[T <: Data](eltsIn: Seq[T]): MixedVec[T] = new MixedVec(eltsIn)
```
ducky64: ditto as above, should this be varargs?

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,103 @@
...
+  * val v = Wire(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(32.W))))
+  * v(0) := 100.U(8.W)
+  * v(1) := 10000.U(16.W)
+  * v(2) := 101.U(32.W)
+  * }}}
+  */
+@chiselName
+final class MixedVec[T <: Data](private val eltsIn: Seq[T]) extends Record with collection.IndexedSeq[T] {
+  // We want to create MixedVec only with Chisel types.
+  if (compileOptions.declaredTypeMustBeUnbound) {
```
ducky64: Doesn't need to happen for this PR, but I'm in favor of getting rid of compile options and just having a catchall compatibility flag. At one point we might have thought about letting people pick and choose individual options, but I don't think that's the philosophy currently, and I'd have all new features be strict, since there's no old code to support.

src/main/scala/chisel3/util/MixedVec.scala
```diff
@@ -0,0 +1,103 @@
...
+    * @param index Index with which to retrieve.
+    * @return Retrieved index.
+    */
+  def apply(index: Int): T = elts(index)
+
+  /** Strong bulk connect, assigning elements in this MixedVec from elements in a Seq.
+    *
+    * @note the lengths of this and that must match
+    */
+  def :=(that: Seq[T]): Unit = {
```
ducky64: This also seems a bit special case-y, though I think the right thing to do is to match behavior with Vec for consistency. I don't think this is supported in Vecs yet (though there have been requests?), and how to do it right and generally is a discussion we should have.

src/main/scala/chisel3/util/MixedVec.scala
edwardcwang: 1) Mainly to make room for the introduction of Vec (and associated MixedVec) literals (#849) if it ever happens.

2) I wouldn't be opposed to a varargs version in addition to a Seq version, though I would prefer it if the Seq version were not completely removed, since sometimes we do want to programmatically create a Seq to use as the type and not have to worry about `:_*` etc.

ducky64: 1. I think that whatever we do, we at least need to keep Vec and MixedVec consistent, and ideally would also keep them consistent with their closest Scala analogues. I'm not sure how we want to do Vec literals in the future, though we might want to keep that consistent with Bundle literals (so Vec(...).Lit, or even Vec.Lit). Or maybe VecLit. But the alternative solution here would be to deprecate VecInit and have it be VecWireInit (or just VecWire?), but we'd end up waiting at least one deprecation cycle (if ever) before we could introduce a different meaning  for VecInit. So I highly doubt that's a practical pathway.
2. The argument for a varargs version is that it's more consistent with the Scala collections library, where you do Seq(elems*), or elems.toSeq. So if we didn't already have VecInit(Seq), a consistent solution might be mySeq.toVec (or mySeq.toVecWire). The idea behind consistency is so that programmers can move between Scala and Chisel without having to think about a different set of interfaces.

edwardcwang: 1) I'm happy to rename to `MixedVecInit` and will do so tomorrow.

2) So in conclusion can we keep the Seq version as well, since we don't have `.toVec` or things like that?

ducky64: I think it's worth a discussion about whether we want .toVec (or similar), since I think there can be issues with a constructor that takes both a varargs and a Seq.

edwardcwang: I agree -  since (I think) `.toVec` would be a further API addition, perhaps we can merge this PR first and propose/discuss `.toVec` in a future issue/PR?

edwardcwang:
TODO from meeting: Edward to add varargs API as well

edwardcwang:
Varargs API added and tested; will merge this PR as soon as the status checks pass.

PR #868, opened 2018-07-31
Cleanup implicit conversions

jackkoenig:
Remove public vals from implicit classes that can surprisingly be
invoked yet do nothing. Many implicit classes were Value Classes which
require the argument to be a public val but the avoidance of runtime
objects compared to the rest of Chisel runtime is minimal at best.

**Related issue**: NA

**Type of change**: bug fix

**Impact**: API modification

This PR is not source-code compatible since someone could be accidentally invoking one of these "methods". We could instead do this via a deprecation schedule.

**Release Notes**

Removed implicitly added identity methods `.data` and `.target` from `Data` and `DecoupledIO`, and removed `.x` from `String`, `Int`, `Long`, `BigInt`, and `Boolean`

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/package.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/package.scala
@@ -23,7 +23,7 @@ package chisel3 {
     * interpreting calls that have a non-Width parameter as a chained apply, otherwise things like
     * `0.asUInt(16)` (instead of `16.W`) compile without error and produce undesired results.
     */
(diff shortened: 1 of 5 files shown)


jackkoenig:
So they do have to be vals if it's a Value Class (the ones that extended `AnyVal`), but the only purpose of Value Classes is to provide an implicit class without runtime boxing. I thought about changing those vals to be `self` or `_self` or something but since such runtime boxing is a drop in the bucket of the overfall performance of Chisel, I thought it cleaner just to convert them to standard implicit classes and removes the vals.

PR #869, opened 2018-07-31
Add infinite-width bit extract operator

albert-magyar:
**Related issue**: #857 
**Development Phase**: hypothetical proposal

@jackkoenig @azidar @ducky64 This is an example of the hypothetical "infinite extract" operator. Not that I'm saying this is the right solution, but it is _one_ easy solution.

Note: `i` is a terrible name that I just made up as an example.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -143,6 +143,15 @@ sealed abstract class Bits(width: Width)
 
   final def do_apply(x: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = apply(BigInt(x))
 
+  /** Returns the specified bit of an infinitely-padded copy of this wire
+    * as a [[Bool]], statically addressed.
+    *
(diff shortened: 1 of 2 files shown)


ducky64:
Can we do something more compositional rather than introducing a special-case extract operation? I'd prefer something like `(5.U.pad)(4, 2)`, though we should think about that an unbounded pad would do (or whether it's even useful) in other cases.

azidar:
I'm not sure what the final word on this was, but I personally am in favor of not breaking backwards compatibility for this, and just adding a safe extract operator. I don't like `i` as the API though....

ducky64:
So the issue is that the backwards compatible solution means inconsistent behavior depending on whether your bit extraction is handled in Chisel frontend on a literal (in which case it implicitly pads the number to infinity) or in FIRRTL (where it would error out). And I'd argue that the former case is probably more of a bug than anything else, since it was a result of trying to special case literals in the chisel frontend but failing to maintain consistent behavior with FIRRTL.

We could also introduce a runtime deprecation phase, where the inconsistency will continue to work but emit a runtime deprecation warning, before removing it outright.

PR #875, opened 2018-08-21
[RFC] [do-not-merge] Refactor Bits into Separate Files

seldridge:
`Bits.scala` is currently 1200 lines and a pain to read (it's 1700 with the API documentation of #873!). The initial motivation for this was the _good_ approach of sealing everything in the `Bits` hierarchy (as bad things happen if you start extending `Bits`). However, the need to ram all of the implementations of all sealed children of `Bits` into one file isn't necessary and it contributes to the bloat of `Bits.scala`.

This proposes breaking up the sealed traits in `Bits.scala` into separate files with private traits containing the macro and method implementations of these traits. This would create the following files:
  - `Element.scala`
  - `Bits.scala`
  - `Num.scala`
  - `UInt.scala`
  - `SInt.scala`
  - `FixedPoint.scala`
  - `Analog.scala`

This PR currently shows an example of how this can be done for `UInt` by breaking it up into:
  - `private [core] UIntMacros`
  - `private [core] UIntMethods`

It may also make sense to breakup the whole hierarchy going back to `Data` into macro and method traits.

There is no effect on ScalaDoc by doing this. Methods provided by private traits mixed into public classes will still show up.

As this will have major conflicts with #873, I'll wait to act on any full implementation of this.

**Related issue**: 

**Release Notes**


chick:
This seems like a good idea. Another reason to wait, a little is we are in the middle of adding yet another bunch of line in order to support the new Interval type.

PR #876, opened 2018-08-22
Minor tweaks to the style guide

edwardcwang:
Some minor tweaks/copyedits to the style guide (does not affect functionality).

### DIFF
--- a/doc/style.md
+++ b/doc/style.md
@@ -48,12 +48,11 @@ import chisel3._
 ## Imports
 Avoid wildcard ( ._ ) imports, with the exception of chisel3._
 All other imports must call out used methods.
-import chisel3._ must be first and separated from remaining imports with extra
-blank line.
+`import chisel3._` must be first, and separated from remaining imports with an extra blank line.
 
 **Reason:** This makes it clear where methods are defined.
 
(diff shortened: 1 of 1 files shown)


doc/style.md
```diff
@@ -48,12 +48,11 @@ import chisel3._
...
 Avoid wildcard ( ._ ) imports, with the exception of chisel3._
 All other imports must call out used methods.
-import chisel3._ must be first and separated from remaining imports with extra
-blank line.
+`import chisel3._` must be first, and separated from remaining imports with an extra blank line.
 
 **Reason:** This makes it clear where methods are defined.
 
-Remaining imports must be listed alphabetically.
+Any remaining imports must be listed alphabetically.
```
chick: How do you feel about this one? I think it is generally ignored, that there is a case for grouping internal vs external imports separately and it begs the question of whether the sort should be case insensitive. It is also not checked AFAIK by the style checker.

edwardcwang: I don't have a strong opinion about this one - if there's a decent consensus to remove it, I'd be fine with that. As you point out, there are a number of subtleties and not enforced anyhow, and there might be a case for cutting out extraneous red tape.

Issue #885, opened 2018-09-04
[RFC] Strong, Annotated Enums


hngenc:
In response to deafening demand from billions of Chisel users worldwide, I am proposing a new enum API that addresses the concerns raised in #373. It aims to deliver a new strongly-typed enum class that is automatically annotated for the convenience of FIRRTL transforms and waveform viewers.

## Motivation
Enums in Chisel suffer from the following deficiencies:
- They are weakly-typed, and completely interchangeable with any `UInt`.
- They do not preserve their names when translated to FIRRTL, so HDL backends and waveform viewers cannot make any use of them.
- The current API only allows enums to be given sequential values. Users cannot map their enums to custom values (such as for one-hot states).
- The current API requires users to count their enum values by hand. This is annoying, especially when adding new enum values.

## Existing Solutions
Here, I list existing enum implementations in Scala and Chisel to see if any satisfy our needs. Unfortunately, Scala 2's lack of an enum keyword make most of these solutions feel awkward.

### Chisel
First, of course, there is Chisel's current enum API:
```scala
val sIdle :: s1 :: s2 :: Nil = Enum(3)
```
The weaknesses of this approach have already been discussed above, but it does get points for being concise and readable.

### Scala Enumeration Class
A simple example can be seen below:
```scala
object MyEnums extends Enumeration {
  val sIdle, s1, s2 = Value
}
```
If the user wishes, they can also map specific enums to custom integers:
```scala
object MyEnums extends Enumeration {
  val sIdle, s1 = Value
  s3 = Value(10)
}
```
This API addresses some of our concerns. It is also easy to iterate over the enum values after they are created using the `MyEnums.Values` set. Unfortunately, it has the following drawbacks that make it unsuitable for use in Chisel:
- Scala's enums can be mapped to integers and strings, but not to `UInt`s or other Chisel datatypes.
- Scala's enums are weakly-typed. All enumerations, even if they are in different objects, share the same type.

### Scala Case Objects
An example of this enum implementation can be seen in the [Chisel JTAG FSM](https://github.com/ucb-art/chisel-jtag/blob/master/src/main/scala/JtagStateMachine.scala). A simplified example is shown below:
```scala
sealed abstract class MyEnum(val id: Int) {
  def U: UInt = id.U
}

object MyEnum {
  val all: Set[MyEnum] = Set(sIdle, s1, s2)
}

case object sIdle extends MyEnum(0)
case object s1 extends MyEnum(1)
[...]

PR #892, opened 2018-09-18
Strong enums

hngenc:
This is an implementation of the strongly-typed enums proposed in #885. It satisfies all the conditions laid out there, along with these extra features:
- UInts can be cast into enums
- The user can retrieve a list of all valid enum values through the `StrongEnum.all` function

### Caveats
The enums do have two primary caveats that should be considered:
- An enum class must be "static" in the Java sense. In other words, you can't define an enum class inside of another class, although you can define it globally, or in an object, as in this example:
```scala
object FSM { // The companion object to a module named FSM
    class State extends EnumType
    object State extends StrongEnum[EnumType] {
        val s1, s2 = Value
    }
}
```
- Casting non-literal UInts to enums is an inherently dangerous procedure, because Chisel has no way of knowing whether or not the UInt is a valid enum value. When this happens, we emit warnings, but @jackkoenig has suggested completely removing this feature unless users demand its inclusion. Needless to say, casting _literal_ UInts to enums is safe, and provides compile-time guarantees.

### API Modifications
- In the `master` branch, `Element` inherits from `Data` and override's `Data`'s `def width` with a `val width` in the constructor. I removed this constructor parameter from `Element`, so that `EnumType` could inherit from `Element` without having to specify the width directly in the constructor (since `EnumType`'s width isn't actually known until all `EnumType`s have been instantiated). Now, other subclasses of `Element`, like `Bits` and `Clock`, must explicitly override `Data`'s `width` with their own implementations (which I have already taken care of).
- Numerous functions, like `switch`, `is` and `bindLitArg` require parameters of type `Bits` in the `master` branch. However, since `EnumType` must also work with these functions, and since `EnumType` isn't a subclass of `Bits`, I modified these functions so that they instead take in parameters of type `Element`.

Neither of these API modifications should break any existing user code. All the old tests still pass.

**Related issue**: #885 

**Impact**: API modification

**Release Notes**
- Added support for strongly-typed, self-annotating enums

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -20,6 +20,10 @@ import chisel3.internal.firrtl.PrimOp._
   * @define coll element
   */
 abstract class Element extends Data {
+  private[chisel3] final def allElements: Seq[Element] = Seq(this)
+  def widthKnown: Boolean = width.known
+  def name: String = getRef.name
+
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection) {
     binding = target
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
@@ -30,9 +34,32 @@ abstract class Element extends Data {
     }
   }
 
-  private[chisel3] final def allElements: Seq[Element] = Seq(this)
-  def widthKnown: Boolean = width.known
-  def name: String = getRef.name
+  private[core] override def topBindingOpt: Option[TopBinding] = super.topBindingOpt match {
+    // Translate Bundle lit bindings to Element lit bindings
+    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
+      case Some(litArg) => Some(ElementLitBinding(litArg))
+      case _ => Some(DontCareBinding())
+    }
+    case topBindingOpt => topBindingOpt
+  }
+
+  private[core] def litArgOption: Option[LitArg] = topBindingOpt match {
+    case Some(ElementLitBinding(litArg)) => Some(litArg)
+    case _ => None
+  }
+
+  override def litOption: Option[BigInt] = litArgOption.map(_.num)
+  private[core] def litIsForcedWidth: Option[Boolean] = litArgOption.map(_.forcedWidth)
+
+  // provide bits-specific literal handling functionality here
+  override private[chisel3] def ref: Arg = topBindingOpt match {
+    case Some(ElementLitBinding(litArg)) => litArg
+    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
+      case Some(litArg) => litArg
+      case _ => throwException(s"internal error: DontCare should be caught before getting ref")
+    }
+    case _ => super.ref
+  }
 
   private[core] def legacyConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
     // If the source is a DontCare, generate a DefInvalid for the sink,
@@ -78,33 +105,6 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
   private[core] def cloneTypeWidth(width: Width): this.type
 
   def cloneType: this.type = cloneTypeWidth(width)
-
-  private[core] override def topBindingOpt: Option[TopBinding] = super.topBindingOpt match {
-    // Translate Bundle lit bindings to Element lit bindings
-    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
-      case Some(litArg) => Some(ElementLitBinding(litArg))
-      case _ => Some(DontCareBinding())
-    }
-    case topBindingOpt => topBindingOpt
-  }
-
-  private[core] def litArgOption: Option[LitArg] = topBindingOpt match {
-    case Some(ElementLitBinding(litArg)) => Some(litArg)
-    case _ => None
-  }
-
-  override def litOption: Option[BigInt] = litArgOption.map(_.num)
-  private[core] def litIsForcedWidth: Option[Boolean] = litArgOption.map(_.forcedWidth)
-
-  // provide bits-specific literal handling functionality here
-  override private[chisel3] def ref: Arg = topBindingOpt match {
-    case Some(ElementLitBinding(litArg)) => litArg
-    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
-      case Some(litArg) => litArg
-      case _ => throwException(s"internal error: DontCare should be caught before getting ref")
-    }
-    case _ => super.ref
-  }
 
   /** Tail operator
     *
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -17,7 +17,7 @@ import chisel3.internal.firrtl.PrimOp._
 /** Element is a leaf data type: it cannot contain other Data objects. Example
   * uses are for representing primitive data types, like integers and bits.
   */
-abstract class Element(private[chisel3] val width: Width) extends Data {
+abstract class Element extends Data {
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection) {
     binding = target
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
@@ -61,8 +61,8 @@ private[chisel3] sealed trait ToBoolable extends Element {
   * bitwise operations.
   */
 //scalastyle:off number.of.methods
-sealed abstract class Bits(width: Width)
-    extends Element(width) with ToBoolable {
+sealed abstract class Bits(private[chisel3] val width: Width)
+    extends Element with ToBoolable {
   // TODO: perhaps make this concrete?
   // Arguments for: self-checking code (can't do arithmetic on bits)
   // Arguments against: generates down to a FIRRTL UInt anyways
@@ -1130,7 +1130,7 @@ object FixedPoint {
   *
   * @note This API is experimental and subject to change
   */
-final class Analog private (width: Width) extends Element(width) {
+final class Analog private (private[chisel3] val width: Width) extends Element {
   require(width.known, "Since Analog is only for use in BlackBoxes, width must be known")
 
   private[core] override def typeEquivalent(that: Data): Boolean =
--- a/chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
@@ -35,33 +35,6 @@ import EnumAnnotations._
 abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolean = true) extends Element {
   override def cloneType: this.type = factory().asInstanceOf[this.type]
 
-  private[core] override def topBindingOpt: Option[TopBinding] = super.topBindingOpt match {
-    // Translate Bundle lit bindings to Element lit bindings
-    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
-      case Some(litArg) => Some(ElementLitBinding(litArg))
-      case _ => Some(DontCareBinding())
-    }
-    case topBindingOpt => topBindingOpt
-  }
-
-  private[core] def litArgOption: Option[LitArg] = topBindingOpt match {
-    case Some(ElementLitBinding(litArg)) => Some(litArg)
-    case _ => None
-  }
-
-  override def litOption: Option[BigInt] = litArgOption.map(_.num)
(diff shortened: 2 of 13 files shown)


edwardcwang:
With regards to "Casting a non-literal UInt to an enum" --> what if this casting operation returned a tuple of the enum and a valid bit?

i.e. as follows:

```scala
val myUInt: UInt = ...
val (asEnum, valid) = MyEnumClass.fromUInt(myUInt)
```

Then if `myUInt` was not a valid Enum value, then the `valid` signal is now, and `asEnum` is `DontCare` or other undefined behavior.

hngenc:
@edwardcwang The problem is that we would need to insert circuitry that constantly monitors the value of `asEnum` to make sure that its value is valid. This seems wasteful to me, especially in real hardware. SystemVerilog has an `assert property` feature that would fix this problem, but that would only be useful in simulation and not in real hardware.

edwardcwang:
Well, if we don't use the `valid` signal, most synthesis tools will prune it away. And in fact even if we generate the signal in Chisel but don't do anything else with it, I think FIRRTL's optimization passes will trim it away. If we were really desperate for it not to show up in Chisel, we could even do something like:

```scala
val asEnum = MyEnumClass.fromUIntRaw(myUInt)
// or
val (asEnum, unused) = MyEnumClass.fromUInt(myUInt, raw=true)
```

hngenc:
@edwardcwang Hmm, that makes sense. I'll look into implementing this feature. It should be pretty easy.

albert-magyar:
Why not just add an assertion instead of a valid signal? That makes it easier to use and automates the simulation-only checking.

aswaterman:
It might legitimately not be valid on all clock cycles, so you'd have to gate the assertion.

albert-magyar:
True!

hngenc:
OK, I've added an `isValid` boolean signal to all enums. I've also renamed the function that casts from non-literals to `castFromNonLit` and added the following restriction:
- Non-literals being cast to enums cannot be wider than the enum itself

I suppose if users want, they can `assert(enum.isValid)`.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,339 @@
...
+  def Value: T = {
+    val result = newEnum()
+    enum_instances.append(result)
+    enum_values.append(id)
+
+    width = (1 max id.bitLength).W
+    id += 1
+
+    // Instantiate all the enums when Value is called for the last time
+    if (enum_instances.length == enum_names.length && isTopLevelConstructor) {
```
jackkoenig: It would be DRY-er to use `finishedInstantiation` here

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,339 @@
...
+  protected def checkEmptyConstructorExists(): Unit = {
+    try {
+      implicitly[ClassTag[T]].runtimeClass.getDeclaredConstructor()
+      getClass.getDeclaredConstructor()
+    } catch {
+      case ex: NoSuchMethodException => throw NoEmptyConstructorException(s"$enumTypeName does not have a no-args constructor. Did you declare it inside a class?")
+    }
+  }
+
+  private val isTopLevelConstructor: Boolean = {
```
jackkoenig: This is a clever way to do this, I'm guessing there isn't a way to do this with reflection?

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,339 @@
...
+      case Some(litArg) => Some(ElementLitBinding(litArg))
+      case _ => Some(DontCareBinding())
+    }
+    case topBindingOpt => topBindingOpt
+  }
+
+  private[core] def litArgOption: Option[LitArg] = topBindingOpt match {
+    case Some(ElementLitBinding(litArg)) => Some(litArg)
+    case _ => None
+  }
```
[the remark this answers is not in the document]
hngenc: Sure thing. I actually just copied that from somewhere else in the codebase :)

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
hngenc: Not one that I can find, unfortunately. I also considered using global variables, but I'm honestly not sure whether that's more or less elegant.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,339 @@
...
+
+  private val enum_names = getEnumNames
+  private val enum_values = mutable.ArrayBuffer.empty[BigInt]
+  private val enum_instances = mutable.ArrayBuffer.empty[T]
+
+  private def getEnumNames(implicit ct: ClassTag[T]): Seq[String] = {
+    val mirror = runtimeMirror(this.getClass.getClassLoader)
+
+    // We use Java reflection to get all the enum fields, and then we use Scala reflection to sort them in declaration
+    // order. TODO: Use only Scala reflection here
```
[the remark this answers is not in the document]
ducky64: Yes, there were cases where Scala reflection just pukes all over the place. I think this had to do with inner classes and fun like that, so we're still stuck with looking at id order, which is kind of hacky and nasty.
We use Scala reflection in autoclonetype, but that's fine because that's best effort - it's more or less accepted to tell the user that they need to overload cloneType (though we might not have if we were doing clean-slate design).

ducky64:
Expanding the use of annotations is interesting, though probably warrants discussion. The current place where annotations are used is only in chiselName, which doesn't affect the generated logic. There is also discussion to use annotations for templating the `do_*` functions, clone module (advanced user feature), and bundle literals (which potentially could see wide adoption, though the annotation is optional - and one downside is that people might define a bundle without using the annotation, or that the literal constructors might seem magic to those who don't read the docs)

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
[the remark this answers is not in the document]
hngenc: I think I'll leave it like this for now, just to preserve the symmetry with Bits.litArgOption

hngenc:
Sorry for doing this after my code had already been reviewed, but I've changed the API again, one last time:

```scala
object State extends ChiselEnum {
    val sIdle, s1, s2 = Value
}
```

This is just as concise as Scala Enumerations, but preserves type-safety. Additionally, with this API, enums do not have to be static.

Macro annotations were fun, but IntelliJ can't handle them, so I got rid of them.

ccelio:
Thanks for keeping IntelliJ in mind!

hngenc:
The new enums work by creating a different `ChiselEnum.Type` for each enum you declare.

Here's how they can be used (it's nearly the same as the old version):
```scala
object State extends ChiselEnum {
    val sIdle, s1, s2 = Value
}

val state = RegInit(sIdle)

// or, alternatively:
val state = Reg(State())
state := sIdle

// or, alternatively:
val state = Reg(State.Type())
state := sIdle
```

There is one downside to this implementation, which is shared by Scala enumerations:
```scala
def foo(e: MyEnum.Type) = ???
def foo(e: OtherEnum.Type) = ??? // This causes a compiler error
```
Basically, you can't overload functions with different types of enums as parameters.

However, it's important to note that type-safety _is_ preserved in the following sense:
```scala
def foo(e: MyEnum.Type) = ???

foo (MyEnum.s1)
foo (OtherEnum.sOther) // This will cause a compiler error
```
So during compilation, Scala will prevent users from passing in the wrong types of enums to functions.

Personally, I don't think there are going to be many use cases where users want to overload functions that take enums as parameters, so I think the reduced verbosity of this new style is worth it.

jackkoenig:
This looks awesome, great work!

@chick can you re-review?

Should we start out with this in experimental @ducky64?

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -19,7 +19,7 @@ import chisel3.internal.firrtl.PrimOp._
   *
   * @define coll element
   */
-abstract class Element(private[chisel3] val width: Width) extends Data {
+abstract class Element extends Data {
```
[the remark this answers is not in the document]
hngenc: No, I hadn't known they use `Element`. I've taken a glance through their codebase, and I doubt my changes will affect them, but I'll make sure.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,290 @@
...
+    if (id.litValue() < this.id)
+      throwException(s"Enums must be strictly increasing: $enumTypeName")
+
+    this.id = id.litValue()
+    do_Value(names)
+  }
+
+  def apply(): Type = new Type
+
+  def apply(n: UInt)(implicit sourceInfo: SourceInfo, connectionCompileOptions: CompileOptions): Type = {
```
ducky64: Might be worth having a discussion on apply vs. fromBits. Especially since fromBits gives you a warning anyways.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,290 @@
...
+abstract class EnumFactory {
+  class Type extends EnumType(this)
+  object Type {
+    def apply(): Type = EnumFactory.this.apply()
+  }
+
+  var id: BigInt = 0
+  var width: Width = 0.W
+
+  val enum_names = mutable.ArrayBuffer.empty[String]
```
ducky64: Are the contents of `enum_names`, `enum_values`, and `enum_instances` related (eg by index - as would be suggested by the zip in `globalAnnotation`)? If so, consider a data structure that preserves the association, like a Map of some type.

src/test/scala/cookbook/FSM.scala
```diff
@@ -7,16 +7,21 @@ import chisel3.util._
...
   val io = IO(new Bundle {
     val in = Input(Bool())
     val out = Output(Bool())
   })
 
-  val sNone :: sOne1 :: sTwo1s :: Nil = Enum(3)
+  object State extends ChiselEnum {
+    val sNone, sOne1, sTwo1s = Value
+  }
+  import State._
```
ducky64: Is this recommended practice? I'd lean towards having users be explicit about namespacing, eg using `State.sNone`

hngenc: I was gonna leave this completely up to users, and whether they want to keep their namespaces apart.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,290 @@
...
+  private[core] override def topBindingOpt: Option[TopBinding] = super.topBindingOpt match {
+    // Translate Bundle lit bindings to Element lit bindings
+    case Some(BundleLitBinding(litMap)) => litMap.get(this) match {
+      case Some(litArg) => Some(ElementLitBinding(litArg))
+      case _ => Some(DontCareBinding())
+    }
+    case topBindingOpt => topBindingOpt
+  }
+
+  private[core] def litArgOption: Option[LitArg] = topBindingOpt match {
```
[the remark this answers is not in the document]
hngenc: The main reason I didn't want to inherit from Bits was because Bits defines a lot of operations like shifting and bit-selecting, which don't really make sense for enums in my opinion

src/test/scala/cookbook/FSM.scala
ducky64: Ultimately, it's up to the users, but we should set good examples in test code.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -0,0 +1,290 @@
...
+    override def toFirrtl: Annotation = EnumDefAnnotation(enumTypeName, definition)
+  }
+}
+import EnumAnnotations._
+
+
+abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolean = true) extends Element {
+  override def cloneType: this.type = factory().asInstanceOf[this.type]
+
+  private[core] override def topBindingOpt: Option[TopBinding] = super.topBindingOpt match {
```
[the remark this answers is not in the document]
hngenc: I moved `topBindingOpt`, `litOption`, `litIsForcedWidth`, and `ref` into Element

chick:
@hngenc can you update the branch to up to date with master and this can be merged.

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
hngenc: As per our discussion, I've moved `fromBits` into `apply`. Enums will be experimental anyway, so we've got time to refine this.

PR #901, opened 2018-10-03
Add DataMirror.modulePorts

ducky64:
Adds `modulePorts(module: BaseModule): Map[String, Data]` to `experimental.DataMirror`. This provides the necessary interfaces to split out Testers2 development (#866) into its own repo.

Points which may or may not warrant discussion:
- This returns things as an unordered Map. Are there ever cases where we want to preserve ordering? Alternatives would be either a ListMap or a Seq.
- BlackBox has its `io` contents flattened out in the return here, consistent with the generated component. Note that using `io` directly on a BlackBox can be illegal in some cases, because the top-level object actually does not get emitted (as an artifact of dropping the `io_` prefix in the pre-ExtModule/MultiIOModule dats). The only alternative would be to error out if a BlackBox is passed in.

**Related issue**: N/A

**Release Notes**
Add chisel3.experimental.DataMirror.modulePorts(module: BaseModule): Map[String, Data]

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/BlackBox.scala
@@ -140,7 +140,8 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     require(!_closed, "Can't generate module more than once")
     _closed = true
 
-    val namedPorts = io.elements.toSeq
+    val namedPorts = io.elements.toSeq.reverse  // ListMaps are stored in reverse order
+
     // setRef is not called on the actual io.
     // There is a risk of user improperly attempting to connect directly with io
     // Long term solution will be to define BlackBox IO differently as part of
--- a/chiselFrontend/src/main/scala/chisel3/core/Data.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Data.scala
@@ -86,6 +86,7 @@ object DataMirror {
     requireIsHardware(target, "node requested directionality on")
     target.direction
   }
+  def modulePorts(target: BaseModule): Map[String, Data] = target.getChiselPorts
 
   // Internal reflection-style APIs, subject to change and removal whenever.
   object internal {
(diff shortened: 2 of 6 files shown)


ducky64:
Also, strictly speaking, `DataMirror` might not be the best place to put this, this might be more like `Driver` or `DriverUtils`, since the main purpose is to enable interaction with the compiled artifacts like simulators, instead of elaboration-time reflection.

edwardcwang:
> This returns things as an unordered Map. Are there ever cases where we want to preserve ordering? Alternatives would be either a ListMap or a Seq.

How does Chisel preserve port ordering in ports/IOs of a module right now? e.g. the order of `io_z` vs `io_a` is preserved in the emitted FIRRTL.

```scala
val io = IO(new Bundle {
  val z = Output(Bool())
  val a = Output(Bool())
})
```

vs.

```scala
val io = IO(new Bundle {
  val a = Output(Bool())
  val z = Output(Bool())
})
```

ducky64:
`Module._ports` is an ArrayBuffer, and `Bundle.elements` is a ListMap. So in theory the ordering is available, but I don't remember off the top of my head if the FIRRTL emitter takes it into account. Especially since ListMap stores in reversed order.

edwardcwang:
I mean (correct me if something has changed) the FIRRTL emitter and FIRRTL's Verilog emitter both seem to be respecting order in at least Bundles right now, no?

edwardcwang:
We could provide the map as an additional function e.g. `.modulePortsMap` if we wanted to?

ducky64:
Ok, I'm going to change the API to a Seq return, and since this is experimental we can change the API to VectorMap/SeqMap in the future. Or move this into like DriverUtils. Both of which I'll write into the comments so we don't forget about it.

ducky64:
And with a Seq API, if users want a map they can always do `.toMap`. I'm generally not in favor of presenting redundant APIs vs. a composition approach in calling code.

ducky64:
Changed the interface to be Seq[(String, Data)]. Added TODOs for possibly moving from DataMirror to Driver or something, and for changing the return type to SeqMap when it becomes available. Reverses the BlackBox ports order, to make them consistent with the declared order, since ListMaps store elements in reverse.

davidbiancolin:
This looks good. I should note, in MIDAS we call UserModule.getPorts() to get a firrtl port list, from which we reconstruct the module's IO.  There's a comment above that member :

```// For debuggers/testers, TODO: refactor out into proper public API  ```

+1 For order preserving.

PR #902, opened 2018-10-03
Modify ReadyValidIO noenq to set the data payload to DontCare.

stevenmburns:
**Related issue**: None

**Impact**: no functional change (probably)

**Release Notes**
Modified ReadyValidIO (also used in DecoupledIO) to set the data payload (.bits) to don't care when a no enqueue action (noenq) is called. This will clean up the frequently used idiom:
````scala
c.noenq
when( cond) {
  c.enq( data)
}
````
This currently needs to be written as:
````scala
c.noenq
c.bits := DontCare
when( cond) {
  c.enq( data)
}
````
It is possible that `noenq` was used like this in some existing code:
````scala
c.bits := 47.U
c.noenq
when( cond) {
  c.valid := true.B
}
````
If this was the case, then the PR will change the behavior of the circuit.
The comment for `noenq` is:
````
    /** Indicate no enqueue occurs. Valid is set to false, and bits are
      * connected to an uninitialized wire
      */
````
so it wasn't supposed to be used this way.

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -36,7 +36,7 @@ object ReadyValidIO {
   implicit class AddMethodsToReadyValid[T<:Data](target: ReadyValidIO[T]) {
     def fire(): Bool = target.ready && target.valid
 
-    /** push dat onto the output bits of this interface to let the consumer know it has happened.
+    /** Push dat onto the output bits of this interface to let the consumer know it has happened.
       * @param dat the values to assign to bits.
       * @return    dat.
       */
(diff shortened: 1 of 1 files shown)


ucbjrl:
This is the classic trade-off of "intent" vs. "ease of use".

Without the explicit `DontCare` connection, FIRRTL will issue a warning if `bits` isn't connected. There could be complicated logic with nested `when`s where one of the conditions leaves `bits` unassigned and this will be correctly flagged as an error by FIRRTL.

With the `DontCare` connection, this error (a missing connection to `bits`) will no longer be diagnosed by FIRRTL.

albert-magyar:
I think that connecting a `DontCare` for a default `noenq` actually does more explicitly align with programmer intent. 

Also, the problem with not flagging missing connections is universal for all "default assignment" patterns; a default `enq` at the top would cause the same problem

PR #903, opened 2018-10-05
add HasBlackBoxPath to BlackBoxUtils.scala

### LINKED ISSUE #963: HasBlackBoxResource doesn't work with absolute paths
**Type of issue**: bug report | feature request 

Do something like:
```
class MyBlackBox extends BlackBox with HasBlackBoxResource {
 ...
 setResource("/absolute/path/to/resource.v")

}
```
You'll get 
```
Caused by: java.io.FileNotFoundException: Resource '/absolute/path/to/resource.v'
```
even though the file exists.

  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**
java.io.FileNotFoundException

The file should get copied in as if it were a relative path.

Makes HasBlackBoxResource more useful

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

I'm working on a fix

albertchen-sifive:
- add `HasBlackBoxPath` that acts like `HasBlackBoxResource` except that it works with files outside of the resources directory.
- deprecate and rename `setResource` to `addResource` because multiple calls will result in multiple resources being copied to the target directory.

**Metadata**:
  - Fixes #963


ucbjrl:
ok to test

jackkoenig:
LGTM, should we also changed `setInline` to `addInline`? @chick thoughts on the API changes?

src/main/scala/chisel3/util/BlackBoxUtils.scala
```diff
@@ -4,12 +4,15 @@ package chisel3.util
...
+import firrtl.transforms.{BlackBoxPathAnno, BlackBoxResourceAnno, BlackBoxInlineAnno, BlackBoxSourceHelper}
 
 trait HasBlackBoxResource extends BlackBox {
   self: BlackBox =>
 
-  def setResource(blackBoxResource: String): Unit = {
+  @deprecated("Use addResource instead", "3.2")
+  def setResource(blackBoxResource: String): Unit = addResource(blackBoxResource)
+
+  def addResource(blackBoxResource: String): Unit = {
```
chick: I think it would be nice to add some scaladoc here, in particular, providing an example
of what a resource path looks like.

src/main/scala/chisel3/util/BlackBoxUtils.scala
```diff
@@ -29,3 +32,15 @@ trait HasBlackBoxInline extends BlackBox {
     chisel3.experimental.annotate(anno)
   }
 }
+
+trait HasBlackBoxPath extends BlackBox {
+  self: BlackBox =>
+
+  def addPath(blackBoxPath: String): Unit = {
```
chick: ScalaDoc here too. Probably worth mentioning that the working directory is
not likely to be the same as the target directory

PR #908, opened 2018-10-13
Add test for behavior of bundles with cloneType.

grebe:
This test currently gives an inscrutable NoSuchElementException. It should be better!

This test currently fails. It should always fail. However, it currently fails with a totally inscrutable error message. I'm not sure what the right error message is, but we should make it better and make the test expect it.

**Related issue**: 

**Development Phase**: bug report via test

**Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -72,6 +72,8 @@ trait InstanceId {
 
 }
 
+class GetReferenceException(message: String) extends Exception(message)
+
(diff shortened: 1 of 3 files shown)


chick:
Can you add me as collaborator, I'd like to push a proposed improvement for the
behavior described here

chick:
@ducky64 can you review this. There might be a better way, but this seemed like
the simplest way to provide a hint as to what is going on in the situation Paul found.

edwardcwang:
@ducky64 to create an issue to check for referential equality of source and dest bundle fields in autoclonetype

PR #912, opened 2018-10-25
Make BaseModule.name lazy

### LINKED ISSUE #911: Lazy evaluation of BaseModule.name
**Type of issue**: bug report

**Other information**
 

I'm building a wrapper for a non-Chisel top design that has to align with a specific Verilog interface. I have this split up into two classes, one that implements the interface in Chisel (`Foo`) and another that wraps this with Verilog and does the connections (`Wrapper`). This all works fine, but I'd like to have the name of the wrapper be a function of the child that I'm wrapping, e.g., if I wrap `Bar extends Foo`, the name should be `BarWrapper` not the generic `Wrapper`. Currently, the name of a module isn't lazy and overriding `desiredName` to depend on something that will be defined during elaboration doesn't work.

The following MWE fails:
```scala
class Wrapper(gen: => Module) extends Module {
  val io = IO(new Bundle{})
  val child = Module(gen)
  override lazy val desiredName = s"${child.desiredName}Wrapper"
}

Driver.elaborate(() => new Wrapper(new Foo)).name should be ("FooWrapper")
```

This currently barfs with a `NullPointerException` because `BaseModule.name` is eager.

This should work. `BaseModule.name` should be lazy. This is a one-line fix + a test case.

current freechipsproject/master: 600405254c20c14fb3389aa4758ec27dffe992d0

Enable this use case to work.

seldridge:
Fixes #911 

This makes `BaseModule.name` lazy (previously eager) so that a Module's `desiredName` can be a function of some property associated with a sub-instance.

**Related issue**: 

**Release Notes**

- Make BaseModule.name lazy to enable better name parameterization, e.g., a module name can now be a function of a submodule name

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Module.scala
@@ -180,7 +180,13 @@ abstract class BaseModule extends HasId {
   def desiredName = this.getClass.getName.split('.').last
 
   /** Legalized name of this module. */
-  final lazy val name = Builder.globalNamespace.name(desiredName)
+  final lazy val name = try {
+    Builder.globalNamespace.name(desiredName)
+  } catch {
+    case e: NullPointerException => throwException(
+      s"Error: desiredName of ${this.getClass.getName} is null. Did you evaluate 'name' before all values needed by desiredName were available?", e)
+    case t: Throwable => throw t
+  }
 
   /** Returns a FIRRTL ModuleName that references this object
     * @note Should not be called until circuit elaboration is complete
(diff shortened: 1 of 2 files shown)


seldridge:
Hmm... I ran this locally and it passed. Poking around...

Good point, @jackkoenig. Let me get a test and include that.

jackkoenig:
We really need to get the chisel3 Travis/CircleCI testing running

seldridge:
@jackkoenig: Pushed a commit that updates this with your message. This catches the case shown in the test. I've requested you re-review this and make sure that this catches what you were thinking.

The output of a new, excepting testing case is now:
```
[info]   chisel3.internal.ChiselException: Error: desiredName of chiselTests.NullModuleWrapper is null. Did you evaluate 'name' before all values needed by desiredName were available?
```

seldridge:
Ugh... Chisel master is broken with FIRRTL master for `StrongEnumSpec` and `LoadMemoryFromFileSpec`... Looking into it.

seldridge:
retest this please

PR #913, opened 2018-10-26
Add BigInt / Int to Bool conversion (0.B, 1.B)

ducky64:
I think there are cases where it's cleaner to use `0.B` and `1.B` instead of `false.B` and `true.B`, especially as testers is going to use chisel literals.

(the use case is in testing interface code, where signal levels are usually specified as 0 and 1 instead of false and true)

**Development Phase**: proposal, implementation

**Release Notes**


ducky64:
Yeah. It would be nice if there was a way to make that type-safe, but there's a trade-off between syntax and error type in this implementation. We can discuss at the meeting today.

Macros might be able to detect something like `2.B` at compile time, though I don't think it's possible to do that in general with variables (though I also can't think of when you would want to do an Int-to-Bool conversion on a variable).

aswaterman:
No objections here. I don’t think it’s especially problematic that 2.B is a runtime error.

seldridge:
retest this please

PR #915, opened 2018-10-29
Overload connect for StrongEnum annotations

seldridge:
Possible fix for Strong Enum annotations problems. Currently, the annotations generated by `StrongEnum` are invalid. 

I tried to figure out a way to handle this by fixing the `bind` overload, but I couldn't seem to find how to differentiate `bind` in the following two cases:
```scala
// case 1
val expected_state = VecInit(sNone, sNone) // for some enum sNone
// case 2
val y: UInt = ???
val x = expected_state(y) // <- this access
```

Currently, this produces annotations of the following form for the former:
```json
  {
    "class":"chisel3.core.EnumAnnotations$EnumComponentAnnotation",
    "target":"StrongEnumFSMTester.StrongEnumFSMTester.expected_state[0]",
    "enumTypeName":"chiselTests.StrongEnumFSM$State"
  },
  {
    "class":"chisel3.core.EnumAnnotations$EnumComponentAnnotation",
    "target":"StrongEnumFSMTester.StrongEnumFSMTester.expected_state[1]",
    "enumTypeName":"chiselTests.StrongEnumFSM$State"
  }
```

And for the latter:
```json
  {
    "class":"chisel3.core.EnumAnnotations$EnumComponentAnnotation",
    "target":"StrongEnumFSMTester.StrongEnumFSMTester.expected_state[value]",
    "enumTypeName":"chiselTests.StrongEnumFSM$State"
  }
```

The latter shouldn't be emitted as this is pointing to an invalid target. This PR changes it so that only the former is emitted.

Questions:
  - Does this need to also override `bulkConnect`?

**Related issue**: freechipsproject/firrtl#922

**Impact**: API modification

**Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
@@ -102,18 +102,11 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
     lit.bindLitArg(this)
   }
 
-  override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
-    super.bind(target, parentDirection)
-
-    // If we try to annotate something that is bound to a literal, we get a FIRRTL annotation exception.
-    // To workaround that, we only annotate enums that are not bound to literals.
-    if (selfAnnotating && litOption.isEmpty) {
-      annotateEnum()
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


seldridge:
@hngenc: Thanks for taking at this PR and for digging into the details of the problem underlying this. Yeah, connect overriding breaks down there. I'll close this in favor of a forthcoming better approach...

PR #916, opened 2018-10-29
Turn off strong enum annotations

hngenc:
Enum annotations don't work properly with dynamically indexed values in `Vec`s. This PR temporarily turns them off so that an elegant solution can be found.

**Related issue**: freechipsproject/firrtl#922, #915

**Type of change**: temporary work-around

**Impact**: API modification (for annotations)

**Release Notes**

### DIFF
--- a/src/test/scala/chiselTests/StrongEnum.scala
+++ b/src/test/scala/chiselTests/StrongEnum.scala
@@ -376,6 +376,7 @@ class StrongEnumAnnotationSpec extends FreeSpec with Matchers {
   import chisel3.experimental.EnumAnnotations._
   import firrtl.annotations.ComponentName
 
+  /*
   "Test that strong enums annotate themselves appropriately" in {
 
     def test() = {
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/StrongEnum.scala
```diff
@@ -376,6 +376,7 @@ class StrongEnumAnnotationSpec extends FreeSpec with Matchers {
   import chisel3.experimental.EnumAnnotations._
   import firrtl.annotations.ComponentName
 
+  /*
   "Test that strong enums annotate themselves appropriately" in {
```
seldridge: [nit] ScalaTests's `ignore` is super useful for this type of thing and generally better than commenting out tests:
```scala
ignore should "Test that strong enums annotate themselves appropriately" in {
// ...
}
```

That way, the test is just skipped, but still shows up when you run tests and counts in the "ignored" bucket. Some more info from [ScalaTest](http://www.scalatest.org/user_guide/tagging_your_tests).

PR #930, opened 2018-11-08
Add Chisel logo to repo and README

albert-magyar:
@seldridge @ucbjrl , this should make it easier for people to find.


edwardcwang:
Can we have less spacing around the logo, or put it under the header?

albert-magyar:
I switched to a cropped SVG, which will fix that space @edwardcwang. I also reverted the README changes; this way, I can add them back by referencing a raw Github content link directly to the vector version.

seldridge:
Looks good. Anecdotally, I like the rebase/amend to get it down to one commit. However, at least with Travis, the jobs started in the rebase don't get killed. Jenkins may be doing the same thing here.

Also, what happened to CircleCI? @ucbjrl, I don't think it was turned off?

albert-magyar:
Yeah, the real problem is that I kept pushing to my branch after I opened the PR to check the README formatting :(

ucbjrl:
I think the issue is this is coming from a repo outside of freechipsproject. I'm not sure what needs to happen to get CircleCI to do its thing. I suspect this is a CircleCI bug.

ucbjrl:
It looks like there's a setting in the CircleCI project which needs to be enabled to allow builds for pull requests from forks. I've enabled it for freechipsproject. Let's see if it requires another push to trigger.

ucbjrl:
I tried redelivering the notification from GitHub but that doesn't seem effective. @albert-magyar could you push a gratuitous change to see if that will trigger CircleCI?

albert-magyar:
Yeah, it did trigger it @ucbjrl

PR #936, opened 2018-11-13
Fix enum annotations

hngenc:
**Related issue**:  freechipsproject/firrtl#922

**Type of change**: bug fix

**Impact**: no functional change (turns enum annotations back on)

**Release Notes**

Turns enum annotations back on and adds a new EnumVecAnnotation that annotates vecs of enums, rather than annotating every element of a vec individually. This resolves the bug where enums that were dynamically indexed from a vec would be annotated, causing a FIRRTL error.

For the special cases where we have multidimensional vecs or vecs of bundles which have internal enum fields, the EnumVecAnnotation has a parameter called `fields: Seq[Seq[String]]` which describes the paths from the outermost vec to the innermost enum fields. For example, for the vec of bundles below:
```scala
class Bund extends Bundle {
    val enum_field = MyEnum()
    val inner_bundle = new Bundle {
        val inner_enum_field = MyEnum()
    }
}

val vec = Wire(3, new Bund())
```
The `fields` parameter would be `Seq(Seq("enum_field"), Seq("inner_bundle", "inner_enum_field"))`


chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -15,32 +14,41 @@ import firrtl.annotations._
 
 
 object EnumAnnotations {
-  case class EnumComponentAnnotation(target: Named, enumTypeName: String) extends SingleTargetAnnotation[Named] {
+  case class EnumComponentAnnotation(target: Named, typeName: String) extends SingleTargetAnnotation[Named] {
```
chick: Can you add a comment explaining what this, and the other annotations, are for?
The chisel ones can just refer to the firrtl ones

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -102,25 +106,60 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
     lit.bindLitArg(this)
   }
 
-  override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
+  override private[chisel3] def bind(target: Binding, parentDirection: SpecifiedDirection = SpecifiedDirection.Unspecified) = {
```
chick: Nit: Bind has lost it's return type here

src/test/scala/chiselTests/StrongEnum.scala
```diff
@@ -367,62 +367,198 @@ class StrongEnumSpec extends ChiselFlatSpec {
...
+  io.other := OtherEnum.otherEnum
+  simple := e100
+  bund.init()
+  vec_of_bundles.foreach(_.init())
+
+  // All variables starting with "ignored" must not be annotated
+  val cycle = RegInit(0.U)
+  cycle := cycle + 1.U
+
+  val ignore = EnumExample()
```
chick: What is this ignoring thing? Can you add a little more comment about what is being tested/demonstrated here

hngenc: I just got rid of the _ignore_ tests. On further thought, I don't think they're needed anyway.

edwardcwang:
@hngenc Are there are any other issues blocking this? If so, could you please reference them here?

hngenc:
@edwardcwang It should be done with my latest push. I've also added more documentation on how the annotations should be interpreted.

src/test/scala/chiselTests/StrongEnum.scala
```diff
@@ -135,7 +135,7 @@ class StrongEnumFSM extends Module {
     is (sOne1) {
       when (io.in) {
         state := sTwo1s
-      } .otherwise {
+      }.otherwise {
```
edwardcwang: typo?

hngenc: The old one was a typo. I personally think it looks better without the space.

edwardcwang: Fine with me though I believe the typical style has been to treat `otherwise` like `else` since Chisel is a DSL, so preferring

```scala
when (a) {
  // ...
} .otherwise {
  // ...
}
```

in analogy with

```scala
if (foo) {
  // ...
} else {
  // ...
}
```

Maybe the others have some thoughts on this as well.

hngenc: Oh, I wasn't aware of that. I've actually been using the `}.otherwise {` style everywhere else. I can just change them though, so that the style is consistent

hngenc: It turns out there was only one `.otherwise`, which I've changed to have a space.

PR #941, opened 2018-11-16
Bump sbt to 1.2.6; update dependencies

ucbjrl:

**Release Notes**
Bump sbt version to 1.2.6
Update dependency versions

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -94,8 +94,9 @@ lazy val chiselSettings = Seq (
   name := "chisel3",
 
   libraryDependencies ++= Seq(
-    "org.scalatest" %% "scalatest" % "3.0.1" % "test",
-    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
+    "junit" % "junit" % "4.12" % "test",
+    "org.scalatest" %% "scalatest" % "3.0.5" % "test",
+    "org.scalacheck" %% "scalacheck" % "1.14.0" % "test",
     "com.github.scopt" %% "scopt" % "3.7.0"
   ),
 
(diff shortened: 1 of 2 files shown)


ucbjrl:
Why the explicit reference to junit?

I was running into problems with transitive dependencies, specifically:
Symbol 'term org.junit' is missing from the classpath.
I suspect this is due to interaction between the latest version of sbt and org.scala-sbt.ivy. It may be because junit is special - it's pure Java and independent of the Scala compiler. Forcing a more recent version of junit seemed to resolve those problems.

jackkoenig:
@ucbjrl can you add a comment about that? That is a bit weird...

ucbjrl:
The:
`Symbol 'term org.junit' is missing from the classpath`
occurred when running tests under Scala 2.11.12.
```bash
sbt test
```
and
```bash
sbt "++2.12.7" test
```
both work correctly. But
```bash
sbt +test
```
and
```bash
sbt "++2.11.12" test
```
both fail with 
```
[error] /nscratch/ucbjrl/firrtl/src/test/scala/firrtlTests/FirrtlSpec.scala:193:18: Symbol 'term org.junit' is missing from the classpath.
[error] This symbol is required by 'class org.scalatest.junit.JUnitRunner'.
[error] Make sure that term junit is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'JUnitRunner.class' was compiled against an incompatible version of org.
[error]   implicit class TestingFunctionsOnCircuit(val circuit: Circuit) extends AnyVal {
[error]                  ^
[warn] one warning found
[error] one error found
[error] (Test / compileIncremental) Compilation failed
```
when compiling for 2.11.12. Adding a specific dependency on junit in build.sbt seems to alleviate the issue.

NOTE:  mill has a similar problem, so it may be a ammonite issue.

jackkoenig:
Would you mind adding a short explanation as a comment in the build.sbt?

Issue #618, opened 2017-05-25
"Clone Module" API Idea and Request for Comment


jackkoenig:
As designs grow larger and larger, we are starting to hit fundamental performance problems in both Firrtl *and* Chisel. There are many techniques to deal with this, and one useful one I think we should consider is allowing users to explicitly "clone" Modules. This would actually partially address both #472 and https://github.com/ucb-bar/firrtl/issues/538.

Cloning would essentially create another Firrtl instance of an already elaborated Module. There are some subtle issues to think about with annotations (Module.component annotations would probably be fine but hierarchical ones would need duplication). 

I propose something like:
```scala
val inst0 = Module(new MyModule)
val inst1 = inst0.cloneModule
// or
val inst2 = Module.cloneInst(inst0)
// or 
val inst3 = Module.duplicate(inst0)
```

This does present some serious problems for systems like rocket-chips's LazyModule that utilize their own data structures to emit information about the configuration and the like. However, as a strictly opt-in feature such libraries could better leverage annotations or provide their own APIs that call ours to take advantage.

Thoughts? Questions? Concerns?

chick:
Do you have some estimate of the gain here?
Your example was of an unparameterized Module, is that relevant or just a simpler example.
It seems to me that requiring users to do this is not that desirable, it seems a bit hard to explain, users would likely not employ the technique at first and then when they hit the wall would have to go back and fix a bunch of existing code.
Could Module.apply could be some soft of factory that (optionally) recognized creating of previous modules and did this under the hood.

aswaterman:
This will most usefully be employed at a fairly macro level (e.g., an N-core design would be constructed once and cloned N-1 times), where it will achieve a nearly N-fold gain.

jackkoenig:
The gain is a substantial performance increase for large designs and a guarantee of deduplication (since the module was never duplicated).

An unparameterized Module is a simpler case, it is parameterized Modules where accidentally not dedupable Modules become a bigger problem. This is really only useful for large designs, and I cannot imagine many large designs that are not parameterized.

This is intended to be an advanced feature since it's strictly a performance optimization, and users are **not** required to use it. Chisel works just fine as is, it's just that large designs--which are almost always hierarchical and instantiate Modules that *must* be deduplicated multiple times--suffer from really bad compile times.

[...]

PR #943, opened 2018-11-16
RFC: Add Rocket Chip-style clonemodule as CloneAndGetIO to experimental

albert-magyar:
**Related issue**: #618 
**Type of change**:  feature request
**Development Phase**: proposal &&  implementation

**EXAMPLE USAGE**

```
val q1 = Module(new Queue(...))
val q2_io = CloneAndGetIO(q1)("io").asInstanceOf[q1.io.type]
q2_io.enq <> q1.io.deq
```

**Release Notes**:
This is a port of Rocket Chip's `CloneModule` functionality (which returns a `Record`) as `chisel3.experimental.CloneAndGetIO`. This simplifies the implementation drastically, as it can live inside chisel3. The idea here is that there may be some desire to have this live alongside a full `CloneModule` implementation that returns a `Module`, as the API for using the `Record`-returning version is much simpler for basic use cases.

**Pros**: it makes the simple use cases simple. Also, I am almost positive that anything returning something more module-like would have also have large, non-overlapping caveats associated with its use, which incentivizes having both variants.
**Cons**: it might be confusing to have both `CloneAndGetIO` and another version of `CloneModule` that returns a `Module`. Also, more API features means more to support ad infinitum.

### DIFF
--- a/src/main/scala/chisel3/package.scala
+++ b/src/main/scala/chisel3/package.scala
@@ -436,6 +436,28 @@ package object chisel3 {    // scalastyle:ignore package.object.name
 
     val IO = chisel3.core.IO
 
+    // Rocket Chip-style clonemodule
+
+    /** A record containing the results of CloneModuleAsRecord
+      * The apply method is retrieves the element with the supplied name.
+      */
+    type ClonePorts = chisel3.core.BaseModule.ClonePorts
+
+    object CloneModuleAsRecord {
+      /** Clones an existing module and returns a record of all its top-level ports.
+        * Each element of the record is named with a string matching the
+        * corresponding port's name and shares the port's type.
+        * @example {{{
+        * val q1 = Module(new Queue(UInt(32.W), 2))
+        * val q2_io = CloneModuleAsRecord(q1)("io").asInstanceOf[q1.io.type]
+        * q2_io.enq <> q1.io.deq
+        * }}}
+        */
+      def apply(proto: BaseModule)(implicit sourceInfo: chisel3.internal.sourceinfo.SourceInfo, compileOptions: chisel3.core.CompileOptions): ClonePorts = {
+        chisel3.core.BaseModule.cloneIORecord(proto)
+      }
+    }
+
     // Implicit conversions for BlackBox Parameters
     implicit def fromIntToIntParam(x: Int): IntParam = IntParam(BigInt(x))
     implicit def fromLongToIntParam(x: Long): IntParam = IntParam(BigInt(x))
--- a/src/test/scala/chiselTests/CloneModuleSpec.scala
+++ b/src/test/scala/chiselTests/CloneModuleSpec.scala
@@ -1,0 +1,91 @@
+// See LICENSE for license details.
+
+package chiselTests
+
+import chisel3._
+import chisel3.util.{Queue, EnqIO, DeqIO, QueueIO, log2Ceil}
+import chisel3.experimental.{CloneModuleAsRecord, MultiIOModule, withClockAndReset}
+import chisel3.testers.BasicTester
+import org.scalatest._
+import org.scalatest.prop._
+
+class MultiIOQueue[T <: Data](gen: T, val entries: Int) extends MultiIOModule {
+  val clk = IO(Input(Clock()))
+  val rst = IO(Input(Bool()))
+  val enq = IO(Flipped(EnqIO(gen)))
+  val deq = IO(Flipped(DeqIO(gen)))
+  val count = IO(Output(UInt(log2Ceil(entries + 1).W)))
+  val impl = withClockAndReset(clk, rst) { Module(new Queue(gen, entries)) }
+  impl.io.enq <> enq
+  deq <> impl.io.deq
+  count := impl.io.count
+}
+
+class QueueClone(multiIO: Boolean = false) extends Module {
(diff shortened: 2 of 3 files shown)


azidar:
@albert-magyar I agree with @edwardcwang , can we see some tests demonstrating the new API?

aswaterman:
@terpstra FYI

src/main/scala/chisel3/package.scala
```diff
@@ -436,6 +436,28 @@ package object chisel3 {    // scalastyle:ignore package.object.name
...
+      */
+    type ClonePorts = chisel3.core.BaseModule.ClonePorts
+
+    object CloneAndGetIO {
+      /** Clones an existing module and returns a record of all its top-level ports.
+        * Each element of the record is named with a string matching the
+        * corresponding port's name and shares the port's type.
+        * @example {{{
+        * val q1 = Module(new Queue(UInt(32.W), 2))
+        * val q2_io = CloneAndGetIO(q1)("io").asInstanceOf[q1.io.type]
```
edwardcwang: Will this still work post-structural types?

src/test/scala/chiselTests/CloneModuleSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+import chisel3.experimental.CloneAndGetIO
+import chisel3.testers.BasicTester
+import org.scalatest._
+import org.scalatest.prop._
+
+class QueueClone extends Module {
+  val io = IO(new QueueIO(UInt(32.W), 2))
+  val q1 = Module(new Queue(UInt(32.W), 2))
+  val q2_io = CloneAndGetIO(q1)
+  q2_io.elements.foreach { case (name, port) => println(name) }
```
edwardcwang: is this `println()` used anywhere?

src/test/scala/chiselTests/CloneModuleSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+  q1.io.enq <> io.enq
+  val q2_io_bundle = q2_io("io").asInstanceOf[q1.io.type]
```
edwardcwang: How does this work with MultiIOModule?

edwardcwang:
Thanks Albert for the tests/examples. Just added some additional questions.

src/main/scala/chisel3/package.scala
albert-magyar: I'm not sure what part of the code you're referring to here.

src/test/scala/chiselTests/CloneModuleSpec.scala
albert-magyar: Silly me

src/test/scala/chiselTests/CloneModuleSpec.scala
albert-magyar: It works with MultiIOModule. I added a test.

src/main/scala/chisel3/package.scala
edwardcwang: Will `q2_io` still have the right type or will it just fall back to `Bundle` post-structural types?

albert-magyar: As far as I can tell, I do not use structural types anywhere here, so this PR should be totally orthogonal to that issue. If someone writes a module that has an anonymous bundle that is accessed with inferred structural types and support for that feature is removed, it is true that it will not be possible to clone that module. However, that is because it will not be possible to even *compile* a circuit using that module.

albert-magyar:
@stevenmburns (sorry I should have tagged you earlier), it would be interesting to see if this would work for your use case.

stevenmburns:
This looks good to me. I can use it to do what I want this way:
````
package pg

import chisel3._
import chisel3.experimental.{CloneAndGetIO}

class And extends Module {
  val io = IO(new Bundle {
    val a = Input( Bool())
    val b = Input( Bool())
    val z = Output( Bool())
  })

  io.z := io.a & io.b
}

class AndVec extends Module {
  val io = IO(new Bundle {
    val a = Input( Vec(8, Bool()))
    val b = Input( Vec(8, Bool()))
    val z = Output( Vec(8, Bool()))
  })

  val template = Module( new And())

  for { ((a,b,z),idx) <- ((io.a,io.b,io.z).zipped).toList.zipWithIndex} {
     val instance_io = if ( idx == 0) {
       template.suggestName( s"And_${idx}")
       template.io 
     } else {
       val instance = CloneAndGetIO( template)
       instance.suggestName( s"And_${idx}")
       instance("io").asInstanceOf[template.io.type]
     }
     instance_io.a := a
     instance_io.b := b
     z := instance_io.z
  }

}

object Driver extends App {
  chisel3.Driver.execute(args, () => new AndVec())
}
````
It is a little bit of a pain to have to do the first instance a little differently than the rest. I tried to do all eight using CloneModuleAndGetIO, but then `template` was dangling (not connected) and produced a firrtl error. 

Also, the name is probably not correct. Maybe `CloneModuleAndGetRecord` or `CloneModuleAsRecord`.

albert-magyar:
@stevenmburns, I actually am out today for the development meeting, but I talked to @jackkoenig, and he’ll review after I finalize the PR today. It will be in this weekend.

albert-magyar:
Resolves #618

PR #945, opened 2018-11-20
Make Vec cloneType keep directions of elements

### LINKED ISSUE #893: Multi-dimensional Vec has weird behavior with bidirectional elements
See https://stackoverflow.com/questions/52415589 for original report

**Type of issue**: bug report

**Impact**:  unknown

**Other information**
 

The below examples are taken from the StackOverflow post above

```scala
// This works
  val io = IO(new Bundle {
    val in = Vec(3, Flipped(Decoupled(UInt(3.W))))
    val out = Vec(3, Decoupled(UInt(3.W)))
  })
  io.in.foreach(_.ready := false.B) // Only here for consistency
  io.out <> io.in
}

// This does not
  val io = IO(new Bundle {
    val sel = Input(UInt(4.W))
    val in = Vec(10, Vec(3, Flipped(Decoupled(UInt(3.W)))))
    val out = Vec(3, Decoupled(UInt(3.W)))
  })
  io.in.foreach(_.foreach(_.ready := false.B))
  io.out <> io.in(io.sel)
}
```

It appears that the nested vec is flipping the directions of the leaf elements. I tried adding another level of nested but it didn't "re-flip", it just matched the directions of the 2 levels of flipping. Note that the emitted FIRRTL directions were correct, there is a mismatch with the Chisel directions vs. the [correct] emitted FIRRTL ones.

Note you can work around this by hoisting the `Flipped` to outside of the `Vec(`, but I think it should work in both cases.

The directions of bidirectional Vec elements should be correct

jackkoenig:
Fixes #893

The added tests test way more than just this fix, I just decided to test several (working) similar things with them

**Related issue**: 

**Type of change**: bug fix

**Release Notes**

Fix bug where Chisel might lose the direction of elements of Vecs

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -240,7 +240,7 @@ sealed class Vec[T <: Data] private[core] (gen: => T, val length: Int)
   def apply(idx: Int): T = self(idx)
 
   override def cloneType: this.type = {
-    new Vec(gen.cloneType, length).asInstanceOf[this.type]
+    new Vec(gen.cloneTypeFull, length).asInstanceOf[this.type]
   }
 
   override def getElements: Seq[Data] =
--- a/src/test/scala/chiselTests/Direction.scala
+++ b/src/test/scala/chiselTests/Direction.scala
@@ -2,11 +2,9 @@
 
 package chiselTests
 
-import chisel3._
 import org.scalatest._
-import org.scalatest.matchers._
-import org.scalatest.prop._
-import chisel3.testers.BasicTester
+import chisel3._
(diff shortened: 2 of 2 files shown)


jackkoenig:
@ducky64, I did a quick scan through and `cloneTypeFull` appears to be used throughout the code base for obvious reasons. One issue with `cloneType` preserving binding data is that it used to have to be implemented in custom `Bundle`s and still has to be implemented in custom `Record`s. Since binding is an internal API, that would be a bit weird. Perhaps worth revisiting though

ducky64:
Right, I think rocket-chip still relied on the binding-clearing behavior of cloneType, which was why we couldn't make everything fully consistent. But it's basically an internal-only API (from a chisel3 point of view) anyways.

PR #962, opened 2018-12-12
Added support to extract bit in reverse order

bhimsenpadalkar:

**Impact**: API modification

**Release Notes**


seldridge:
ok to test

ucbjrl:
retest this please

bhimsenpadalkar:
Makes perfect sense.. Thanks !!

jackkoenig:
I agree with @seldridge, I think an explicit reverse followed by extraction makes the most sense.

I would like to add that there is already a reverse utility: [`chisel3.util.Reverse`](https://github.com/freechipsproject/chisel3/blob/83f2a65a3ab1d21258883c0b113406ef9900a57f/src/main/scala/chisel3/util/Bitwise.scala#L92)

PR #966, opened 2018-12-21
Remove GhpagesPlugin.

maloneymr:
I don't know if this GhpagesPlugin is important to anyone, but I'm proposing removing it from firrtl and Chisel3 to eliminate some friction it is causing in another project.

If this plugin is in fact useful to anyway, I would like to make it configurable so that it may be disabled where it is not needed.

**Related issue**: https://github.com/freechipsproject/firrtl/pull/979

**Release Notes**
Removed sbt dependency on GhpagesPlugin.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,8 +2,6 @@
 
 enablePlugins(SiteScaladocPlugin)
 
-enablePlugins(GhpagesPlugin)
-
 def scalacOptionsVersion(scalaVersion: String): Seq[String] = {
   Seq() ++ {
     // If we're building with Scala > 2.11, enable the compile option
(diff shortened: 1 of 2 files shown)


ucbjrl:
ok to test

build.sbt
```diff
@@ -35,7 +33,6 @@ val defaultVersions = Map("firrtl" -> "1.2-SNAPSHOT")
 lazy val commonSettings = Seq (
   organization := "edu.berkeley.cs",
   version := "3.2-SNAPSHOT",
-  git.remoteRepo := "git@github.com:freechipsproject/chisel3.git",
```
ucbjrl: Are you sure this is part of the `GhpagesPlugin`? I think we want to keep this

ucbjrl: It does seem this depends on `GhpagesPlugin`. I'll look into reviving the old way to set the SCM repository.

maloneymr:
> Shouldn't we also remove:
> 
>     addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.2")
> from `project/plugins.sbt`?
> 
> I'm concerned this may end up causing future issues, either through neglect or confusion.



Yes. I had hand-copied the patch from my other repository for this PR, and I missed this change. Thank you. I updated the PR with this change.

PR #977, opened 2019-01-08
Upstream intervals 1

chick:
## Part of introduction of Interval Type
Depends on freechipsproject/firrtl#870

## Enhancement and API Addition
This provides the support necessary to create references to and operations on IntervalTypes.
It provides a new way of specify wire properties as IntervalRanges. IntervalRanges are defined via a range interpolator, that can define a minimum value, maximum value, and binary point. The minimum and maximum values may be open or closed in the sense of an arithmetic interval.

## Release Notes

A long time coming, Intervals allows chisel users to specify not just the width but instead the smallest and largest value that the wire will carry.  Intermediate wires will can be left undetermined and the Firrtl compiler will infer the appropriate widths. Intervals will also support binary points so they support all the functionality of UInt, SInt and FixedPoint.


chick:
@azidar The test IntervalSpec has failing tests that raise some of the remaining issues/questions with firrtl. Take a look when you get a chance. I'd be happy to go over it with you

chick:
This will be superseded by a forthcoming PR

Issue #361, opened 2016-11-10
chisel3 exception handling


ucbjrl:
We should have a top-level exception handler in chisel3. It should report accumulated errors and warnings before exiting.

aswaterman:
To the extent that we can continue executing past errors, so that we can report as many errors as possible without exiting, I agree.  But in the unresumable case, the stack trace is helpful, so the top-level exception handler should still print it.

ducky64:
So once an exception is thrown and it reaches the top level handler, can you actually resume from where the exception was thrown?
(under the assumption that the exception is non-fatal to the current pass, and where the possibility for more similar exceptions exist)

ducky64:
Resolution: place a "top-level" exception handler in (or around) Driver.elaborate, which just prints out the error (with a stack track) along with other non-fatal errors.

ucbjrl:
To do this correctly will require significant restructuring of the code in order to avoid reporting the same error multiple times. I'm submitting a PR that demonstrates a simple (naive) solution, but marking it DO NOT MERGE. We should discuss this at a development meeting, but for the moment I'm bumping the milestone.

jackkoenig:
This is resolved via Stage/Phase in Chisel 3.3.

PR #980, opened 2019-01-11
Exception handling

ucbjrl:
This is a preliminary attempt at converting chisel runtime exceptions into simple non-terminating errors. It uses a common method: `Builder.exception()` to report exceptions and fatal errors. The single argument version generates (and throws) an exception. The two-argument version generates an error based on the first argument and returns the second argument. The fact that there are errors will ultimately cause the elaboration to fail, but not before other potential errors have been reported.

Care is need to structure the code such that:
 - cascading errors aren't generated
 - the same error isn't reported multiple times

The first issue can be solved be returning something suitable at the point the error is detected. The second is more problematic. The chisel2 approach was to effectively turn the accumulated errors into a set and suppress attempts to enter the same error (text and source line number) multiple times.

**Related issue**: #361

**Impact**: API addition (internal API to get error messages accumulated during build)

**Release Notes**


src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -76,15 +77,24 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils {
...
       elaborate {
         new Module {
           val io = IO(new Bundle {
             val b = new BadSeqBundle
           })
         }
       }
-    }).getMessage should include("Public Seq members cannot be used to define Bundle elements")
+    }
+    // Combine exception and (possible) error messages.
```
edwardcwang: Can this part be made less boilerplate-y? From the perspective of a test author this looks like an increase in complexity.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -299,23 +299,58 @@ private[chisel3] object Builder {
...
+    error(e.getMessage)
+    result
+  }
+  /** Record an exception as an error, and throw it.
+    *
+    * @param e exception
+    * @return This method does not return.
+    */
+  @throws(classOf[Exception])
+  def exception(e: Exception): Nothing = {
```
edwardcwang: Should this maybe be `fatalException`? Is it a good idea to have the presence of an argument determine the difference between whether or not the error is recorded right away?

Also, should Chisel devs generally use the no-throw version of the throw version by default? I'd imagine you'd want to use the former as possible since otherwise it means that compilation terminates on the first exception.

src/test/scala/chiselTests/BundleSpec.scala
ucbjrl: Yes, I believe we can add some special `SubString extends String` types and the corresponding `areEqual` definition to ChiselTesters. I'll look into this.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
ucbjrl: The basic idea is to make it easy to switch between **fatal** and **non-fatal** errors, assuming the rest of the code is in place to deal with the **non-fatal** condition. That may be an invalid assumption since in principle, additional code is required to make **non-fatal** work correctly.

I suspect Chisel devs will start with the **fatal** version of the error (as it is simpler to implement) before investing the additional time and effort into providing the additional support for the **non-fatal** condition.

This would be simplified if we had a general guide for implementing error diagnostics in Chisel, i.e. something like:
```scala
def chiselMethodWithErrorChecking(args: Data*): Data = {
  def validateArgs: Option[Seq[Error]]  = {
    val errors: ArrayBuffer[Error]
    // Each validate will update errors and return true or false depending on whether or not
    //  the remaining validation tests should be run.
    if (validateTest1 && validateTest2 && (validateTest3 || validateTest4)) {
      validateTestn
    }
    if (errors.isEmpty) {
      None
    } else {
      Some(errors.toSeq)
    }
  }
  validateArgs: match {
    case None =>
      someResult(args)
    case Some(errors: Seq[Error]) =>
      errors foreach Builder.exception(_.error, {} )
      ErrorData
  }
}
```
where `ErrorData` would be something like `DontCare` but would work for LHS (connect operations) as well.

ucbjrl:
Currently, errors (and warning) are ephemeral - they're gone once a circuit is elaborated. Should we convert them into annotations? Presumably we wouldn't even run the FIRRTL compiler if there were chisel errors but this would be a way to make warnings persist and make them available to FIRRTL passes.

seldridge:
Annotations make sense to me. That could be delayed until Stage/Phase lands.

After Stage/Phase, the FIRRTL compiler expects to see at least one `FirrtlCircuitAnnotation`. If it doesn't see that, it will error out. However, I did write this in such a way that it will compile multiple FIRRTL circuits in parallel. This brings up a more complicated question of what if the Chisel stage produces multiple FIRRTL circuits, but one of those failed with an error?

One idea would be to have errors be fatal and introduce annotations that can convert errors to warning and allow execution to continue for as long as it can. Same thing for promoting warnings to errors.

jackkoenig:
Stale with too many conflicts, still a good idea though that we really ought to do.

PR #981, opened 2019-01-11
Chisel Stage

seldridge:
**Related issue**: 

This PR provides a reimplementation of the Chisel Driver as a `Stage`. The Chisel Driver is converted to a compatibility wrapper around the FIRRTL Stage that freechipsproject/firrtl#920 provides. This PR is split up into separate commits that shows the order in which the porting was done and a rough guide for how this can be done for other repositories:

1. Add expected annotations
2. Write the CLI
3. Add the `xOptions` class (which is not a case class)
4. Write an options view to construct your `xOptions`
5. Add methods to existing `ExecutionOptions` to reconstruct an `AnnotationSeq`
6. Write all new phases to reimplement Driver functionality
7. Write any necessary compatibility layer and/or re-use existing compatibility layers
8. Write the new `Stage`
9. Replace old Driver logic with new phases

Anecdotally, this process is non-linear. I did this by incrementally porting the old Driver in pieces and moving whatever functionality I could into it when I had it. Git's `--squash` and `--fixup` help a lot for keeping the history sane when you do rebase. Additionally, writing tests as you go is tedious, but likely necessary.

__Most of the porting difficulty is (1) deciding, without constraints, what you want the new stage to do and (2) writing the compatibility layer that interfaces this with the old Driver.__ 

There are several temptations here that I think should be avoided:

- Opting to have the old Driver call an old, already ported, sub-Driver (e.g., have `chisel3.Driver` call `firrtl.Driver`). This causes the old Driver and the new Stage to use different code paths. While not insurmountable, this would seem to introduce the potential for subtle bugs between the two versions.
- For the same reasons as above, no functionality should differ between the old Driver and the new stage excepting compatibility layer phases and a view to an execution result.

Most of difficulty in not yielding to these temptations is dealing with situations where the old Driver does not do things in a linear, pipe-like order. E.g., the old Chisel Driver would do elaboration and then continually refer back to the elaborated Chisel circuit. The stage doesn't need to do this as it's moving directly towards a CHIRRTL circuit or to run the FIRRTL compiler. However, the compatibility wrapper has to reconstruct what happened. This PR gets around this, kludgily, by viewing deleted annotations. This isn't really advised, but for a compatibility layer it's passable.

Depends on:
  - https://github.com/freechipsproject/firrtl/pull/920.

**Release Notes**

- Adds ChiselStage, a the Chisel Driver as a firrtl.options.Stage
- The original Chisel Driver is converted to a compatibility layer around ChiselStage

### DIFF
--- a/src/main/scala/chisel3/ChiselExecutionOptions.scala
+++ b/src/main/scala/chisel3/ChiselExecutionOptions.scala
@@ -2,7 +2,9 @@
 
 package chisel3
 
-import firrtl.{ExecutionOptionsManager, ComposableOptions}
+import chisel3.stage.{NoRunFirrtlCompilerAnnotation, PrintFullStackTraceAnnotation}
+
+import firrtl.{AnnotationSeq, ExecutionOptionsManager, ComposableOptions}
 
 //TODO: provide support for running firrtl as separate process, could alternatively be controlled by external driver
 //TODO: provide option for not saving chirrtl file, instead calling firrtl with in memory chirrtl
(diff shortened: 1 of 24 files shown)


seldridge:
@ucbjrl: I actually don't think I can change the compare branch for this. See image below. I'm going to re-PR under freechipsproject.

2019-01-25-173936_912x93_scrot

ucbjrl:
You should have been able to create new branches - chisel-stage - in the
free chips projects, then change "master" to "chisel-stage" for your PRs
and merge them into chisel-stage.

PR #985, opened 2019-01-16
Define Data .toString

### LINKED ISSUE #953: Chisel type prints need improvement
Printing Chisel types gives fun and useless strings like `UInt@4fd5e7a`. But the FIRRTL emitter can make coherent types, though that functionality isn't directly accessible from Data subtypes. What would be nice is for Data subtypes to override toString with:

- as a chisel type: print the FIRRTL type, or something close to it, eg `UInt<8>`
- as a bound object: also indicate the binding and parent module, ideally also add basic source locator information, eg `UInt<8>(Reg in MyModule @ MyModule.scala:104)`
- as a literal: indicate the type and value, eg `UInt<8>(233)`

Bikeshedding on the precise representation is welcome.

**Type of issue**: feature request, other enhancement

**Development Phase**: request, proposal

ducky64:
toString now defined on Data subtypes, and handles literals properly.

See test cases for expected behavior.

Meant to support testers2, which uses literals as its interface. So you can directly print the result of a peek, instead of unpacking it through litValue and friends. Also gives type information for printf debugging when writing RTL.

Discussion points:
- Naming for Bundles and ChiselEnums are inconsistent, as ChiselEnum uses a qualified path name where Bundles are just the class simpleName. Perhaps we should standardize on one? @hngenc 
- @hngenc comments on the changes to StrongEnum.scala are welcome.
- I'm not sure why Bundle has its className hardcoded to Bundle. I removed that to get it to default to the class name in Record. @jackkoenig appears to be the last person associated with that line of code, thoughts?
- The FixedPoint parameter print style is consistent with FIRRTL `<width><<binaryPoint>>` but not how I've defined Vec `<n, type>`. Note that FIRRTL defines Vec as `type[n]`. @azidar thoughts if we want things to print one particular way or the other?
- Some potential for deduplication with Emitter.scala.
- We currently don't have source locators for binding operations, but it could be added in the future.

**Related issue**: Resolves #953

**Release Notes**
toString now defined on Data subtypes, and handles literals properly.


edwardcwang:
General questions: is this meant to be the "standard definition/format" for printing Chisel objects? Is testers2 going to parse these strings?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -543,8 +559,6 @@ class AutoClonetypeException(message: String) extends ChiselException(message)
   * }}}
   */
 abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
-  override def className = "Bundle"
-
```
jackkoenig: `className` is used by `Printable` for hardware pretty printing. I think just using the default Aggregate `.getClass.getSimpleName` is better for obvious reasons. I can't recall my original thinking, but I think it had to do with making this pretty printing look like printing a Map in Scala (since that's sort of what Records/Bundles are), but I think giving the name of the Bundle class is generally better.

My only question is what about anonymous or inner bundles? You'll get some pretty weird names but perhaps that's still desirable.

ducky64: Since the intention of this is for human debugging, I think providing better naming on a best-effort basis makes sense. So anonymous Bundles will probably give some nasty name. Inner bundles should work fine, it's one of the test cases (though not intentionally).

Random thought: maybe we could regex match anonymous bundle names and replace it with a more friendly AnonymousBundle? Might be more trouble than it's worth though, so I'd prefer to wait for user feedback first.

jackkoenig: Yeah I think that's sensible, at least `MyModule$anon$1` tells you more than just `Bundle`!

src/test/scala/chiselTests/DataPrint.scala
```diff
@@ -0,0 +1,81 @@
...
+      Clock().toString should be ("Clock")
+      FixedPoint(5.W, 3.BP).toString should be ("FixedPoint<5><<3>>")
+      Vec(3, UInt(2.W)).toString should be ("Vec<3, UInt<2>>")
+      EnumTest.Type().toString should be ("chiselTests.DataPrintSpec$EnumTest")
+      (new BundleTest).toString should be ("BundleTest")
+    } }
+  }
+
+  class BoundDataModule extends MultiIOModule {  // not in the test to avoid anon naming suffixes
+    Wire(UInt()).toString should be("UInt(Wire in DataPrintSpec$BoundDataModule)")
```
jackkoenig: Do you think it might be worth having a source locator for Data? Let people know where the Wire was constructed?

ducky64: Possibly. This is new enough that I don't know what the use case for this will be (other than printing literals), so I'd like to wait and see first.

There would also need to be infrastructure built for it, mainly IO needs to have a source locator attached to it, and all Bindings needs a copy of the source locator (currently is only part of the FIRRTL command). Adding source locators for literals would be more difficult because of the possibility for a chained apply, so we might have to build more source info transform macros.

ducky64:
Another interesting idea would be to print IO names where possible, though naming data isn't available until after the Module finishes elaboration. Perhaps something like
`Clock(IO in unelaborated MyModule)`
and
`Clock(IO clock in MyModule)`

chiselFrontend/src/main/scala/chisel3/core/StrongEnum.scala
```diff
@@ -33,6 +33,17 @@ import EnumAnnotations._
 
 
 abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolean = false) extends Element {
+  override def toString: String = {
+    val bindingString = litOption match {
+      case Some(value) => factory.nameOfValue(value) match {
+        case Some(name) => s"($value=$name)"
+        case None => s"($value)"
```
hngenc: Would it be better to put something like `($value=INVALID)` here?

hngenc:
I made a small comment, but overall, I think the changes to `ChiselEnum` are just fine.

@ducky64 If you're talking about changing `enumTypeName` to the simple name, then I'm fine with that.

ducky64:
- Added fallback Bundle naming so anonymous Bundles are named `Bundle`, otherwise they had an empty string

PR #987, opened 2019-01-17
Fix scaladoc for UInt.unary_!

aswaterman:
It performs the operation (x === 0.U), just like in C.   The scaladoc incorrectly described it as performing the operation !x(0).  (Obviously, these are equivalent for Bool, but not for UInt in general).

**Related issue**: 

**Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Bits.scala
@@ -852,7 +852,7 @@ sealed class UInt private[core] (width: Width) extends Bits(width) with Num[UInt
 
   /** Unary not
     *
-    * @return a hardware [[Bool]] asserted if the least significant bit of this $coll is zero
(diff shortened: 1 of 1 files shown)


seldridge:
My bad... Width inference really does work. Warning: there may be more errors in this Scaladoc. These were all updated at once and the size of `Bits.scala` makes my eyes glaze over whenever I open it.

aswaterman:
Ack.  I'll do a quick sanity check on the others in that file.

aswaterman:
The width behavior of UInt << UInt is slightly wrong, and the description of UInt.asSInt is correct but a little vague.  Fixing.

aswaterman:
Looks like unary_- has the wrong width description, too - these are truncating but described as expanding.  Fixing.

aswaterman:
@seldridge please re-review

ucbjrl:
... and set some timeouts for the Jenkins jobs.

aswaterman:
Interesante.

PR #994, opened 2019-01-21
Unify internal (chisel3.core) and external (chisel3 / chisel3.experimental) Module class names

### LINKED ISSUE #899: Unify Module names in chisel3.core._ and chisel3.experiemental._

**Other information**
 

It's confusing when we always have to think: is UserModule the same as RawModule? Is ImplicitModule the same as MultiIOModule?

We can change internal (chisel3.core._) names as we wish).

ducky64:
UserModule -> RawModule (in chisel3.experimental)
ImplicitModule -> MultiIOModule (in chisel3.experimental)

I can't rename `LegacyModule` to `Module`, since it conflicts with `object Module` in Module.scala. It appears companion objects need to be in the same source file as the class, whereas we've separated the Module-with-RTL (UserModule.scala) and the base abstract Module (Module.scala, which includes things like `ExtModule`).

Discussion on resolving that is welcome, but the options I see (none of them great) are:
- Leave as-is (chisel3.core.LegacyModule = chisel3.Module)
- Move `object Module` to UserModule.scala
- Move `Module` / `LegacyModule` to Module.scala
- Merge UserModule.scala and Module.scala (one giant file w/ BaseModule, ExtModule, and all RTL modules)

---

**Related issue**: Resolves #899

**Release Notes**
N/A

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
@@ -14,7 +14,7 @@ import chisel3.internal.sourceinfo.UnlocatableSourceInfo
 
 /** Abstract base class for Modules that contain Chisel RTL.
   */
-abstract class UserModule(implicit moduleCompileOptions: CompileOptions)
+abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
     extends BaseModule {
   //
   // RTL construction internals
@@ -58,7 +58,7 @@ abstract class UserModule(implicit moduleCompileOptions: CompileOptions)
     require(!_closed, "Can't generate module more than once")
     _closed = true
 
-    val names = nameIds(classOf[UserModule])
+    val names = nameIds(classOf[RawModule])
 
     // Ports get first naming priority, since they are part of a Module's IO spec
     namePorts(names)
@@ -117,8 +117,8 @@ abstract class UserModule(implicit moduleCompileOptions: CompileOptions)
   *
   * @note Module instantiations must be wrapped in a Module() call.
   */
-abstract class ImplicitModule(implicit moduleCompileOptions: CompileOptions)
-    extends UserModule {
+abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
+    extends RawModule {
   // Implicit clock and reset pins
   val clock: Clock = IO(Input(Clock()))
   val reset: Reset = IO(Input(Bool()))
@@ -143,7 +143,7 @@ abstract class ImplicitModule(implicit moduleCompileOptions: CompileOptions)
   * in a withClock/withReset/withClockAndReset block, or directly hook up clock or reset IO pins.
   */
 abstract class LegacyModule(implicit moduleCompileOptions: CompileOptions)
-    extends ImplicitModule {
+    extends MultiIOModule {
   // These are to be phased out
   protected var override_clock: Option[Clock] = None
   protected var override_reset: Option[Bool] = None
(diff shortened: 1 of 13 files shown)


chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
```diff
@@ -14,7 +14,7 @@ import chisel3.internal.sourceinfo.UnlocatableSourceInfo
 
 /** Abstract base class for Modules that contain Chisel RTL.
```
edwardcwang:
```suggestion
/** Abstract base class for Modules that contain Chisel RTL.
This abstract base class is a user-defined module which does not include implicit clock and reset and supports multiple IO() declarations.
```

chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
```diff
@@ -117,8 +117,8 @@ abstract class UserModule(implicit moduleCompileOptions: CompileOptions)
   *
```
edwardcwang:
```suggestion
  * This abstract base class includes an implicit clock and reset.
  *
```

src/main/scala/chisel3/package.scala
```diff
@@ -430,8 +430,8 @@ package object chisel3 {    // scalastyle:ignore package.object.name
     val dontTouch = chisel3.core.dontTouch
 
     type BaseModule = chisel3.core.BaseModule
-    type MultiIOModule = chisel3.core.ImplicitModule
-    type RawModule = chisel3.core.UserModule
+    type MultiIOModule = chisel3.core.MultiIOModule
+    type RawModule = chisel3.core.RawModule
```
seldridge: This is the biggest of nits, but I used to confuse myself everytime I looked at this because these aren't sorted topologically by inheritance. Consider putting `RawModule` before `MultiIOModule`.

ucbjrl:
This breaks rocketchip:
```
[error] .../src/main/scala/diplomacy/CloneModule.scala:11:8: object UserModule is not a member of package chisel3.core
[error] import chisel3.core.UserModule
[error]        ^
[error] .../src/main/scala/diplomacy/CloneModule.scala:37:26: not found: type UserModule
[error]     val method = classOf[UserModule].getDeclaredMethod("_commands")
[error]                          ^
 ...
[error] two errors found
[error] (Compile / compileIncremental) Compilation failed```

edwardcwang:
Oops. @ducky64  (Though why was rocket-chip importing from `chisel3.core`?)

jackkoenig:
Despite our best efforts to tell people to not, people end up importing from `chisel3.core` because it's a public API. I do wonder if we should just get rid of the whole `chisel3._` vs. `chisel3.core._`, it seems to cause more confusion than anything.

We can fix this particular issue by adding deprecated aliases for the old names to a package object for `chisel3.core`.

jackkoenig:
I think there's kind of an impedance mismatch in how we've been using aliases vs. how they're intended to be used. In Scala, Predef brings certain useful things into scope automatically for convenience (eg. [`Map`](https://github.com/scala/scala/blob/9e264483d04f48513f3a82e91365667f01a2f5f3/src/library/scala/Predef.scala#L161), [`println`](https://github.com/scala/scala/blob/9e264483d04f48513f3a82e91365667f01a2f5f3/src/library/scala/Predef.scala#L419). But fundamentally, these are merely *aliases*--it's perfectly okay to import or use these from their actual definitions: `scala.collections.Map` and `scala.Console.println`.

Our argument that people should only get things from `chisel3`, `chisel3.experimental`, and `chisel3.util` (although those aren't aliases), is not really in the spirit of what Scala does. `chisel3.UInt` is an alias for `chisel3.core.UInt`, and that fact is abundantly clear in the ScalaDoc. Rather than finding better ways to try to stop getting people to import things from `chisel3.core._`, I think we should reconsider the premise of trying to stop people from importing things from `chisel3.core`.

PR #995, opened 2019-01-22
Support DontCare in Mux and cloneSupertype

### LINKED ISSUE #848: Can't use Mux with DontCare while equivalent when() statement is fine
**Type of issue**: bug report

Consider the following two functions below, `muxLess` and `whenLess`. As a user, they appear to implement the same logic and I would expect them to both function. However, the `whenLess` case elaborates and works as expected while `muxLess` flames out with an error:

```
java.lang.AssertionError: can't create Mux with heterogeneous types class chisel3.core.UInt and class chisel3.core.DontCare$
  chisel3.core.cloneSupertype$$anonfun$1.apply(Data.scala:132)
  chisel3.core.cloneSupertype$$anonfun$1.apply(Data.scala:112)
[...]
  scala.collection.AbstractTraversable.reduce(Traversable.scala:104)
  chisel3.core.cloneSupertype$.apply(Data.scala:112)
  chisel3.core.Mux$.do_apply(Mux.scala:32)
[...]
```

```scala
// doesn't work
def muxLess(in: UInt, outWire: UInt): Unit = {
    outWire := Mux(in < 48.U, in, DontCare)
}

// works fine
def whenLess(in: UInt, outWire: UInt): Unit = {
    when(in < 48.U) {
        outWire := in
    } .otherwise {
        outWire := DontCare
    }
}

class MyMyModule extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(12.W))
    val out = Output(UInt(12.W))
  })
  
  whenLess(io.in, io.out) // works
  muxLess(io.in, io.out) // doesn't work
}
```

**Please tell us about your environment:** Chisel 3.1

**What is the use case for changing the behavior?** Seems like an arbitrary restriction to a user that we can't use DontCare in a Mux() but can use it in a when() statement to accomplish the exact same thing.

ducky64:
Adds general support for DontCare to cloneSupertype, and add special case handling in Mux for the chisel style (DontCare object) to firrtl style (invalidate wire) conversion. Not sure how to actually test the DontCare case, but this tests the other case.

Note that it will still be difficult to use DontCare in a Mux, because Scala doesn't understand that DontCare should be a bottom type of Data, so it will infer the output type of Mux to be the common supertype of DontCare (which extends Element) and your Data, which usually means you'll get a Element or Data out, instead of something more specific eg UInt.

This also kind of feels like a game of whackamole, since DontCare support may be subtly broken in other contexts, though I guess it also makes no sense to do something like `1.U + DontCare`

---

**Related issue**: resolves #848

**Release Notes**
Support DontCare in Muxes

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Data.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Data.scala
@@ -113,8 +113,11 @@ private[core] object cloneSupertype {
                                                           compileOptions: CompileOptions): T = {
     require(!elts.isEmpty, s"can't create $createdType with no inputs")
 
-    if (elts.head.isInstanceOf[Bits]) {
-      val model: T = elts reduce { (elt1: T, elt2: T) => ((elt1, elt2) match {
+    val filteredElts = elts.filter(_ != DontCare)
+    require(!filteredElts.isEmpty, s"can't create $createdType with only DontCare inputs")
+
+    if (filteredElts.head.isInstanceOf[Bits]) {
+      val model: T = filteredElts reduce { (elt1: T, elt2: T) => ((elt1, elt2) match {
         case (elt1: Bool, elt2: Bool) => elt1
         case (elt1: Bool, elt2: UInt) => elt2  // TODO: what happens with zero width UInts?
         case (elt1: UInt, elt2: Bool) => elt1  // TODO: what happens with zero width UInts?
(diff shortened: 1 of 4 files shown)


edwardcwang:
From a user-facing perspective this looks fine to me. Is there a way to get Scala to understand that DontCare is a bottom type of Data, or is that not worth it / it introduces unwanted side effects / etc?

ducky64:
The implementation of Nothing as a bottom type may be compiler magic, see this [StackOverflow thread](https://softwareengineering.stackexchange.com/questions/195793/how-is-nothing-a-subtype-of-every-other-type-in-scala). I'm going to merge this as-is and we can improve later if we figure out how.

edwardcwang:
Sounds good.

PR #998, opened 2019-01-22
Import aliases for chisel3.core

ducky64:
@jackkoenig does this fix rocket-chip?

**Related issue**: Fixes bug in #994

**Impact**: compatibility for the foolish

**Release Notes**
Don't import from chisel3.core. It's not part of the public API. You'll have a bad day when we muck with the internals, because you'll need to update your code. Unless that's your thing, or you're doing some kind of bleeding edge stuff and are OK with breakages. In which case, carry on.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/package.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/package.scala
@@ -124,10 +124,10 @@ package chisel3 {
 
     // These provide temporary compatibility for those who foolishly imported from chisel3.core
     @deprecated("Avoid importing from chisel3.core, these are not public APIs and may change at any time. " +
-      " Use chisel3.RawModule instead.", "since the beginning of time")
+      " Use chisel3.experimental.RawModule instead.", "since the beginning of time")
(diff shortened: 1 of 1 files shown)


chiselFrontend/src/main/scala/chisel3/core/package.scala
```diff
@@ -121,5 +121,13 @@ package chisel3 {
     implicit class fromIntToBinaryPoint(int: Int) {
       def BP: BinaryPoint = BinaryPoint(int)  // scalastyle:ignore method.name
     }
+
+    // These provide temporary compatibility for those who foolishly imported from chisel3.core
+    @deprecated("Avoid importing from chisel3.core, these are not public APIs and may change at any time. " +
+      " Use chisel3.RawModule instead.", "since the beginning of time")
```
edwardcwang: Is RawModule out of experimental yet?

ducky64: Nope and fixed.

ucbjrl:
This seems to fix rocket-chip.

PR #999, opened 2019-01-23
Reg enable valid construction

Sawwave:

**Release Notes**

This adds some shorthand RegEnable factory functions for updating registers from Valids or ReadyValidIOs with less boilerplate. Each of the 4 added functions use the previous 2 RegEnable() function overloads, and update to the given value based on the .valid signal for Valid values, or ready and valid for ReadyValidIO values.


aswaterman:
For the first case, there's already a construct that does this, called `Pipe`: https://github.com/freechipsproject/chisel3/blob/e564fff139768e6a232d118ec7155ecf080f1bc4/src/main/scala/chisel3/util/Valid.scala#L56 -- you can just write `Pipe(myValidIO)` to get a registered `Valid[T]`, or `Pipe(myValidIO).bits` to get the payload out, like your construct.

I'm reluctant about the others.  It's a departure from the RegEnable API to have a return type that's different from the input type (i.e., `Valid[T]` vs. `T`).  The second and fourth seem very domain specific, as it's rather unusual to reset the payload of a Valid/Decoupled.

The third one might be a candidate for adding to `Pipe`  than `RegEnable`, i.e., 

    def apply[T <: Data](enq: Valid[T], latency: Int = 1)(implicit compileOptions: CompileOptions): ReadyValidIO[T] = {
      apply(enq.fire(), enq.bits, latency)(compileOptions)
    }

but even that one is pretty debatable, because converting from ReadyValidIO to ValidIO is arguably better done in an explicit fashion, e.g., `Pipe(myReadyValidIO.fire(), myReadyValidIO.bits)`.

Sawwave:
Ah, excellent points, and with that explanation I have to agree with you. I've noticed that the documentation on Pipe only mentions delaying by a specific number of cycles, not with a valid signal, but after a closer look at the function arguments it seems obvious.

My personal preference would be to include a ReadyValidIO interface into Pipe, but I understand why it may be preferable to abandon with the others.

aswaterman:
Cool, I think it's reasonable to add that overload to Pipe, but would be good to hear what others think.

jackkoenig:
I hope others weigh in as well. Looking at (3), my main concern is that I'm not sure what people would assume. When I see an interface taking `ReadyValidIO`, I generally assume it is now responsible for providing backpressure (ie. it should drive `.ready`). In this case, the proposed API merely uses `.fire()` but doesn't drive ready which could be counterintuitive to some people.

I am also curious what other people think though; I do see how it expresses something useful.

ucbjrl:
ok to test

albert-magyar:
Closing this, since the consensus seems to be to rely on `Pipe` on `Valid[T]`.

PR #1010, opened 2019-02-06
Emit fewer _T variables

azidar:
**Related issue**: 

**Release Notes**

We can pretty up the emitted FIRRTL from Chisel3 if temporary nodes that are read once are inlined. We should only perform this inlining if the info source of the temporary declaration is the same as where it is read from.

This PR is just the transform implementation, it requires tests and integration into chisel3. I think its best to wait until after #1004 is merged, as this transform will become a phase in the Chisel3 compilation.

### DIFF
--- a/src/main/scala/chisel3/transform/InlineTemporaries.scala
+++ b/src/main/scala/chisel3/transform/InlineTemporaries.scala
@@ -1,0 +1,94 @@
+package chisel3.transform
+
+import firrtl._
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.annotations.ModuleTarget
+
+import scala.collection._
+
+class InlineTemporaries extends Transform {
+  override def inputForm: CircuitForm = ChirrtlForm
+  override def outputForm: CircuitForm = ChirrtlForm
+
+  override protected def execute(state: CircuitState): CircuitState = {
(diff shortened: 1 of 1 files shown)


edwardcwang:
Out of curiosity, is there a particular reason this transform couldn't go into FIRRTL? Is it meant to be Chisel-specific or maybe we need to iterate on it quickly on the Chisel side?

chick:
I would like to have this transform in treadle and diagrammer, although they typically take input from Chisel they are not dependent on Chisel and they are designed to started with raw firrtl of any level and thus this transform would be useful for them to have access.

ucbjrl:
retest this please

seldridge:
Does the Chisel naming plugin supersede this PR?

jackkoenig:
This is a bit different, it's prettier because there are fewer statements. The plugin notably doesn't help with when conditions so for:
```scala
when (a > 123.U) {
  foo := bar
}
```
We still emit:
```
node _T_1 = gt(a, UInt(123))
when _T_1 : 
  foo <= bar
```
This PR makes it prettier by emitting
```
when gt(a, UInt(123)) : 
  foo <= bar
```
But it doesn't help with the stability issue

jackkoenig:
SFC has been removed.

Issue #805, opened 2018-04-13
[RFC] Bundle Literals


ducky64:
(supersedes #418 with a more detailed proposal of syntax and thoughts on implementation concerns, supersedes #694 with an alternative, I think cleaner, implementation path)

## Motivation
I don't think much needs to be said here. Main use cases would be in testing (for concise and simple testvector specification) and in initializing Bundles with RegInit and WireInit. Maybe also ROMs, but there's a whole other issue for that.

Note: it's possible to construct Bundle-literal-like objects by casting from a UInt with asTypeOf, but specifying anything but all zeros requires knowledge of how Bundles are bit-packed, which is not great. This also requires a Builder context, making it incompatible with testers.

## Proposed syntax
Given a concrete instance of a Bundle type, an autogenerated (by macro annotation) Lit function would return a new literal of that type.

For example, if I have
```scala
@bundle
class MyBundle(val width: Int) extends Bundle {
  val a = Bool()
  val b = UInt(width.W)
}
```

I would be able to do:

```scala
myBundleIo := (new MyBundle(8)).Lit(a=true.B, b=255.U)
```

### Edge cases
It is not necessary to be in a Chisel Builder context to construct a Bundle literal (barring some logic construction being in the Bundle definition, which we have no control over). This allows its use in testers.

Using keyword arguments allows the partial specification of literals. Current proposal is for everything not specified to be treated as `DontCare`.

Bundle literals will take a literal interpretation of `DontCare`, that is, `DontCare` means `DontCare` (and might show up as X-like in simulation) instead of `DoesntExist` (where the previous value on the wire, if any, would not be overwritten).

### Open questions
Should keyword arguments be used? On the plus side, it allows a partial specification, however, it is incompatible with def macros, which are used for source locators. Note that currently, literals do not provide source locators, but this could prevent compatibility with def macros in the future. (current proposal will be to use keyword arguments)

Because Bundles can take parameters, the current proposal requires a literal to be constructed given a Bundle instance (so `Lit` would be a method on an instance, instead of on a companion object). Does it make sense to have a shorthand version in the companion object when no parameters are required? For example, if I defined:
```scala
@bundle
class MyBundle extends Bundle {
  val a = Bool()
  val b = UInt(8.W)
}
```
should `MyBundle.Lit(a=true.B, b=255.U)` be allowed?

### Dealing with LitArg
[...]

PR #1011, opened 2019-02-08
Add support for asynchronous reset

jackkoenig:
Add new AsyncReset type that extends trait Reset. The "reset-type" of a
register is set by the type of the in scope Reset:

  val asyncReset: AsyncReset = IO(Input(AsyncReset()))
  val syncReset: Bool = IO(Input(Bool()))

  val asyncReg = withReset(asyncReset) { RegInit(0.U) }
  val syncReg = withReset(syncReset) { RegInit(0.U) }

AsyncReset can be cast to and from Bool. Whereas synchronous reset is
equivalent to a mux in front of a flip-flop and thus can be driven by
logic, asynchronous reset requires that the reset value is a constant.

Fixes https://github.com/freechipsproject/chisel3/issues/343

This relies on ~freechipsproject/firrtl#1011~ and freechipsproject/firrtl#1068

**Related issue**: 

**Release Notes**

Add support for asynchronous reset


src/test/scala/chiselTests/AsyncResetSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+  } .elsewhen (count >= 3.U) {
+    assert(reg === 5.U)
+  }
+
+  when (done) {
+    stop()
+  }
+}
+
+class AsyncResetQueueTester extends BasicTester {
```
ducky64: Is this a common use for a Queue? Because it's not a clock crossing queue, right, it's just a async-reset queue?

jackkoenig: Queues aren't special here, this is just illustrating that for modules written the right way, you can use them as either synchronous or asynchronous reset just by the context within which they are used. This is in contrast to Verilog where you have to rewrite all of the always blocks to accomplish the same goal.

src/test/scala/chiselTests/AsyncResetSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+  val asyncResetNext = RegInit(false.B)
+  asyncResetNext := count === 0.U
+  val asyncReset = asyncResetNext.asAsyncReset
+
+  val reg = withClockAndReset(slowClk, asyncReset) {
+    RegInit(123.U(8.W))
+  }
+  reg := 5.U // Normal connection
+
+  when (count > 0.U && count < 3.U) {
```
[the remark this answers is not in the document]
jackkoenig: Yeah, good point, the Queue test does this much better (I wrote it after writing this one so incidentally did a better job 😛)

chiselFrontend/src/main/scala/chisel3/core/Reg.scala
```diff
@@ -153,12 +153,15 @@ object RegInit {
       requireIsChiselType(t, "reg type")
     }
     val reg = t.cloneTypeFull
-    val clock = Builder.forcedClock.ref
-    val reset = Builder.forcedReset.ref
+    val clock = Builder.forcedClock
+    val reset = Builder.forcedReset
+    if (reset.isInstanceOf[AsyncReset] && init.litOption.isEmpty) {
+      Builder.error(s"Register $this with async reset has non-literal reset value")
```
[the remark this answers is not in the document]
edwardcwang: This has nothing to do with clock domains - it's just that async reset's default value is usually reset to a fixed constant. Logic synthesis tools can use SR flops to have non-constant async reset values but there are some implementation issues associated with that (e.g. glitches). Jack can substantiate

chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
```diff
@@ -124,7 +124,11 @@ abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
     extends RawModule {
   // Implicit clock and reset pins
   val clock: Clock = IO(Input(Clock()))
-  val reset: Reset = IO(Input(Bool()))
+  val reset: Reset = {
+    // Default if no outer reset is synchronous reset
+    val gen = Builder.outerReset.map(_.cloneType).getOrElse(Bool())
```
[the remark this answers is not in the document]
jackkoenig: The old way of assigning clock and reset will work with async reset, but only if the implicit reset is AsyncReset. I grant that this example could trip someone up, and it's probably something people will try since you'll implicitly have a Bool as reset when you have a Module as your top module. I don't really know what else we can do though without sacrificing what little type safety we're able to have 🤷‍♂️

chiselFrontend/src/main/scala/chisel3/core/Reg.scala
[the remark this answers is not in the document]
jackkoenig: I would add that we definitely need checks for legal clock domains (and after this PR, legal async reset deassert), but that's something we'll have to do in FIRRTL using https://github.com/freechipsproject/firrtl/pull/937

src/test/scala/chiselTests/AsyncResetSpec.scala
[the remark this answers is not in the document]
jackkoenig: I've updated this test to check better

chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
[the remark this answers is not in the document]
jackkoenig: I added a test that illustrates this error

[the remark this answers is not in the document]
ducky64: I'm still a bit uncomfortable with the structure of this, where a Module's IO signature (class of reset field) is dependent implicitly on the enclosing Module's reset type. Also, is this getting you any more type safety? As far as the type checker is concerned `val reset` is of type Reset, and only runtime type information can determine whether that's a AsyncReset or (synchronous) Bool.

Also, the potential for reassignment of reset makes me feel that determining whether a reset is sync or async is a concern that's downstream of Chisel, since (generally speaking) that would need access to the hardware graph that's not available in Chisel. As implemented this handles a special case (though probably the most common one) where reset is not reassigned.

And if sync and async reset versions of a Module are used simultaneously, this would generate two different Modules even in HiFIRRTL? Should that be the case, and is that intuitive for debugging?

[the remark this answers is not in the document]
jackkoenig: Including Chisel's "type system" as a form of type safety, this proposal is no less "type safe" than most things in Chisel (see connections, etc.), while still providing nice flexibility. Would you rather than we have to have both a SyncResetQueue and an AsyncResetQueue and you have to statically pick the right one? Similarly to how Chisel doesn't make you worry about clock and reset if you're in the common case of only having one of both, in my opinion it's really nice to let the context within which a Module is used specify the type of the implicit reset (as long as it supports both).

> Also, the potential for reassignment of reset makes me feel that determining whether a reset is sync or async is a concern that's downstream of Chisel

I'm not sure I understand what you're proposing by determining sync vs. async reset downstream of Chisel. Async reset and synchronous reset are not functionally equivalent. Designers need the ability to specify the type of the reset, but should also be empowered to write reset-agnostic *generators*.

> And if sync and async reset versions of a Module are used simultaneously, this would generate two different Modules even in HiFIRRTL? Should that be the case, and is that intuitive for debugging?

Similarly to the above, a module using sync reset and a module using async reset are functionally different and should be treated as such in FIRRTL.

My proposal in this PR is to make it possible to express a Module such that it works with either (as in the case of the Queue instantiation example), or may only work with synchronous reset (as caught by the error checking for reset values being non-literals, and leaving catching safe reset treatment for a FIRRTL transformation akin to checking for safe clock crossings). The idea is that you can use the Scala type system if you want, by making your resets be of type `Bool`, `AsyncReset`, or `Reset` (for both). For convenience and backwards compatibility reasons, I thought it prudent to make `Module` work with either. Many modules are written to work with both so this is fairly convenient. And for those that won't work in this way, it will be caught by the checks mentioned above.

One thing that I *do* think is currently missing from this PR is an easy and explicit way to constraint the type of the implicit reset. Currently, the best way to do so would be:
```scala
require(Module.reset.isInstanceOf[AsyncReset])
// or
require(Module.reset.isInstanceOf[Bool])
```
I would appreciate suggestions for improvement if anyone has ideas! Some other possibilities:
```scala
require(Module.resetIsAsync)
require(Module.resetIsSync)
// or
require(Module.resetIsBool)
```
Perhaps a mixin for `Module` like:
```scala
class MyModule extends Module with ImplicitAsyncReset { ... }
```
I generally like the idea of a mixin but it doesn't work utility functions and whatnot that may require a specific type of reset.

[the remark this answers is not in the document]
ducky64: I think the type safety of Reset (general reset, that could be either sync or async), AsyncReset and Bool (synchronous reset) makes sense. But what about making the Reset type instantiable so that you could create a Module with an IO of type Reset that indicates that the Module could be generated as either sync or async reset? This seems to better encode user intent, and could also allow better linking of Modules in FIRRTL since whether the Module is sync or async is encoded in FIRRTL and you can delay the actual reset resolution to link time? And inter-module linking is really kind of in the FIRRTL domain (with Bindings being an exception, though the motivation is largely usability instead of separation of responsibilities).

I also do like the mixin approach, perhaps the mixin could hook into the Module reset IO generation logic and force it to a subtype of Reset? It would be very tight coupling between ImplicitAsyncReset/ImplicitSyncReset and Module, though the exposed API is nice.

jackkoenig:
Summary from Chisel Dev meeting 15/02/2019:
* Make abstract type `Reset` instantiable in Chisel
* Add abstract type `Reset` to FIRRTL
* Reset propagation will occur in FIRRTL
* Ensure that Chisel API to specify Reset-type in Chisel is concise and clear

Per that last point, I think the Chisel API on a Module-level is great:
```scala
class MyModule extends RawModule {
  val reset = IO(Input(AsyncReset()))
  ...
}
// elsewhere
val mod = Module(new MyModule)
mod.reset := nonAsyncReset
```
If `nonAsyncReset` is known to be a `Bool` by Chisel, it can error right away. If the runtime type of `nonAsyncReset` is just `Reset`, then FIRRTL will error.

The case I'm a little less clear on is how you do something like this in a function. The best I can come up with:

```scala
def myFunc(x: UInt): UInt = {
  // I need to specify that the reset *must* be async, but Module.reset may have type Reset
  val myAsyncReg = withReset(WireInit(AsyncReset(), Module.reset)) {
    RegInit(123.U)
  }
  myAsyncReg := x
  myAsyncReg
}
```

Note that it might feel obvious that something like this would work, but it will not:
```scala
val myAsyncReset = withReset(Module.reset.asAsyncReset) { RegInit(123.U) }
```
.as* casts are reinterpretations, as such `myBool.asAsyncReset` works just fine.

On the bright side, the future "check async resets are properly used" (similar to a check for legal clock crossings), will catch misuse here, so maybe this just isn't a problem

chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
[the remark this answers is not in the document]
jackkoenig: See below for plan of record

chiselFrontend/src/main/scala/chisel3/core/Reg.scala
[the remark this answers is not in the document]
ErikFanderson: Hey all! I have been messing around with the async-reset branch and have noticed that the litOption method (called in above diff) has been left unimplemented in the Aggregate class. This causes issues when trying to initialize a register of Vecs. I can come up with a local workaround for now, but I thought I would bring it to your attention in case you were unaware. I see the TODO is there so just tell me to shut up if this is already on your radar. All in all, very cool stuff!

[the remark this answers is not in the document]
edwardcwang: Yep, that's what #805 and #849 are meant to implement :)

johnsbrew:
@jackkoenig any updates on this PR?
I'am looking forward to deploying Preset integration in chisel-tester but it first requires a proper integration of AsyncReset in chisel.
May I be of any help?

jackkoenig:
@johnsbrew I apologies for the delay (and then my delay in responding). I am working on it, you can see my WIP (https://github.com/freechipsproject/firrtl/tree/infer-reset). Reset inference has definitely been trickier than I thought, but I think I'm relatively close, I'll let you know when it seems to work 🙂

chiselFrontend/src/main/scala/chisel3/core/Bits.scala
```diff
@@ -1207,7 +1207,102 @@ trait SIntFactory {
...
+
+  /** @group SourceInfoTransformMacro */
+  def do_toBool(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = do_asBool
+}
+
+object AsyncReset {
+  def apply(): AsyncReset = new AsyncReset
+}
+
+// TODO: Document this.
```
ducky64: TODO that needs doing

chiselFrontend/src/main/scala/chisel3/core/Module.scala
```diff
@@ -43,6 +43,8 @@ object Module extends SourceInfoDoc {
     // Save then clear clock and reset to prevent leaking scope, must be set again in the Module
     val clockAndReset: Option[ClockAndReset] = Builder.currentClockAndReset
     Builder.currentClockAndReset = None
+    // Record the outer reset for determining type of implicit reset in MultiIOModules
+    Builder.outerReset = clockAndReset.map(_.reset)
```
ducky64: Is this used anywhere?

chiselFrontend/src/main/scala/chisel3/core/Reg.scala
ducky64: If the Reset type is meant to be instantiable as both a Async or Sync reset, should this be a whitelist check (that the current reset of explicitly of a Bool type) instead?
Cons if it errors here: it's an over-eager error, you might never instantiate a non-literal reset value with an AsyncReset
Pros: you avoid a potential minefield of flakey code that uses non-literal reset values with Reset where the test cases don't try instantiating the module under AsyncReset

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -75,6 +75,7 @@ trait ChiselRunners extends Assertions with BackendCompilationUtilities {
     val manager = new ExecutionOptionsManager("compile") with HasFirrtlOptions
                                                          with HasChiselExecutionOptions {
       commonOptions = CommonOptions(targetDirName = testDir.toString)
+      chiselOptions = ChiselExecutionOptions(printFullStackTrace = true)
```
ducky64: Leftover debugging config?

jackkoenig:
So there's a remaining issue with interoperability with compatibility mode.

Here are the two options I see:

1. Make reset in chisel3.Module Reset, keep as Bool() in Chisel.Module
    * Every single time a chisel3 Module instantiates a Chisel Module, reset connection will fail
2. Make reset in Chisel Module a Reset()
    * Problem in connections
        * Caught in Firrtl
    * Problem in Mux
        * Caught in Chisel

(1) is the current issue, it seems like it'd be the best at first blush but I think every single instantiation of a Chisel Module in chisel3 code breaks
(2) We can fix the Mux part fairly easily in Chisel I think, the connections issue is harder but maybe also okay. Both solutions feel a little haphazard but fair more constrained then the problems that fall out of (1)

sequencer:
Is `AsyncReset` work in the chisel3 without legacy support now?
I'm going to try to use this branch directly.

edwardcwang:
@jackkoenig I just noticed that a piece of code that used to work with `Bool` fails in `VerilogEmitter` when replaced with `AsyncReset`:

```scala
import chisel3._
import chisel3.experimental.RawModule

class Hello extends RawModule {
  val clock = IO(Input(Clock()))
  val reset = IO(Input(Reset()))
  val in = IO(Input(UInt(16.W)))
  val out = IO(Output(UInt(16.W)))
  withClockAndReset(clock, reset) {
    out := RegNext(in, 12345.U)
  }
}

class HelloWorld extends RawModule {
  val clock = IO(Input(Clock()))
  
  val mux = IO(Input(Bool()))
  val in = IO(Input(UInt(16.W)))
  val out = IO(Output(UInt(16.W)))

  // reset2 fails with AsyncReset, works with Bool()
  //val reset = IO(Input(Bool()))
  val reset = IO(Input(AsyncReset()))

  // Neither of the two variants below work if reset is AsyncReset
  // Both variants work if reset is Bool
  val reset2 = IO(chiselTypeOf(reset))
  //val reset2 = IO(Input(chiselTypeOf(reset)))

  val m = Module(new Hello)
  m.clock := clock
  m.reset := Mux(mux, reset, reset2)
  m.in := in
  out := m.out
}

chisel3.Driver.execute(Array[String](), () => new HelloWorld)
```

Error message:

```
firrtl.FIRRTLException: Internal Error! unrecognized cast: WRef(reset,AsyncResetType,PortKind,MALE)
Please file an issue at https://github.com/ucb-bar/firrtl/issues
  firrtl.Utils$.error(Utils.scala:437)
  firrtl.Utils$.throwInternalError(Utils.scala:178)
  firrtl.VerilogEmitter.cast$1(Emitter.scala:229)
  firrtl.VerilogEmitter.emit(Emitter.scala:235)
  firrtl.VerilogEmitter.$anonfun$emit$1(Emitter.scala:257)
  firrtl.VerilogEmitter.$anonfun$emit$1$adapted(Emitter.scala:257)
  scala.collection.immutable.List.foreach(List.scala:389)
  firrtl.VerilogEmitter.emit(Emitter.scala:257)
  firrtl.VerilogEmitter.$anonfun$emit$1(Emitter.scala:257)
  firrtl.VerilogEmitter.$anonfun$emit$1$adapted(Emitter.scala:257)
  scala.collection.immutable.List.foreach(List.scala:389)
  firrtl.VerilogEmitter.emit(Emitter.scala:257)
  firrtl.VerilogEmitter.emit(Emitter.scala:222)
  firrtl.VerilogEmitter$VerilogRender.$anonfun$emit_streams$5(Emitter.scala:834)
  firrtl.VerilogEmitter$VerilogRender.$anonfun$emit_streams$5$adapted(Emitter.scala:834)
  scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:59)
  scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:52)
  scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
  firrtl.VerilogEmitter$VerilogRender.emit_streams(Emitter.scala:834)
  firrtl.VerilogEmitter$VerilogRender.emit_verilog(Emitter.scala:918)
  firrtl.VerilogEmitter.$anonfun$emit$3(Emitter.scala:969)
  scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:59)
  scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:52)
  scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
  firrtl.VerilogEmitter.emit(Emitter.scala:963)
  firrtl.VerilogEmitter.$anonfun$execute$3(Emitter.scala:978)
  scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
  scala.collection.immutable.List.foreach(List.scala:389)
  scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
  scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
  scala.collection.immutable.List.flatMap(List.scala:352)
  firrtl.VerilogEmitter.execute(Emitter.scala:975)
```

jackkoenig:
@edwardcwang thanks for the heads up! I'll investigate

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -1205,7 +1205,102 @@ trait SIntFactoryBase {
...
+
+  /** @group SourceInfoTransformMacro */
+  def do_asAsyncReset(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): AsyncReset
+}
+
+object Reset {
+  def apply(): Reset = new ResetType
+}
+
+/** "Abstract" Reset Type that can be inferred to be either [[AsyncReset]] or [[Bool]]
```
ducky64: Perhaps clarify that abstract is in the sense that it is inferred in FIRRTL to the FIRRTL-concrete AsyncReset and Bool types?

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -1205,7 +1205,102 @@ trait SIntFactoryBase {
...
+
+  /** @group SourceInfoTransformMacro */
+  def do_toBool(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = do_asBool
+}
+
+object AsyncReset {
+  def apply(): AsyncReset = new AsyncReset
+}
+
+// TODO: Document this.
```
ducky64: TODO?

chiselFrontend/src/main/scala/chisel3/Reg.scala
```diff
@@ -178,12 +178,15 @@ object RegInit {
       requireIsChiselType(t, "reg type")
     }
     val reg = t.cloneTypeFull
-    val clock = Builder.forcedClock.ref
-    val reset = Builder.forcedReset.ref
+    val clock = Builder.forcedClock
+    val reset = Builder.forcedReset
+    if (reset.isInstanceOf[AsyncReset] && init.litOption.isEmpty) {
```
ducky64: I assume this an opportunistic check that won't detect non-literal-resets for AsyncResets inferred in FIRRTL? If so, might be worth noting this limitation in the comments.

src/test/scala/chiselTests/ResetSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+  reg := reg + 1.U
+  out := reg
+}
+
+
+class ResetSpec extends ChiselFlatSpec {
+
+  behavior of "Reset"
+
+  it should "allow writing modules that are reset agnostic" in {
```
ducky64: Is it also worth testing that this work with the implicit clock and reset in Module?

chiselFrontend/src/main/scala/chisel3/core/Reg.scala
jackkoenig: Check is removed from Chisel into FIRRTL, checked after reset inference and supports async reset registers reset to aggregates.

chiselFrontend/src/main/scala/chisel3/Reg.scala
jackkoenig: Removed in favor of more sophisticated check in FIRRTL

src/test/scala/chiselTests/ResetSpec.scala
jackkoenig: All tests using a `Module` are testing this, in particular, `AsyncResetQueueTester` ensures that `Queue` works with async reset.

chiselFrontend/src/main/scala/chisel3/Bits.scala
jackkoenig: done

chiselFrontend/src/main/scala/chisel3/Bits.scala
jackkoenig: Agreed, done

jackkoenig:
Done, obviously still fails until https://github.com/freechipsproject/firrtl/pull/1068 is merged

timsliu:
Hi there!
I've read through this thread and am trying to add asyncReset to a LazyModule. I've tried adding:

override val reset = IO(Input(AsyncReset()))

to my LazyModuleImp definition but run into a FIRRTL error. Is there documentation somewhere on how to add an async reset?

Thanks!

jackkoenig:
There's is a PR of documentation for the website that needs a little cleanup but has useful information: https://github.com/freechipsproject/www.chisel-lang.org/pull/74

timsliu:
Great, thanks for the help!

timsliu:
The reset documentation was really helpful! I looked around and saw that ```RequireAsyncReset``` is in the chisel3 package. But when I try to compile the example:

```scala
class MyAlwaysAsyncResetModule extends Module with RequireAsyncReset {
  val myAsyncResetReg = RegInit(false.B) // reset is of type AsyncReset
}
```
I get a type not found error for requireAsyncReset. I think I'm on the most recent version of Chisel - is there some other package that needs  to be imported?

Thanks!

seldridge:
The `RequireAsyncReset` is a Chisel 3.3 features (and Chisel3 3.3 hasn't been released, yet). @ucbjrl is working super hard to get a release candidate out before the weekend. 

In the meantime, if you really want this feature, this may be in a snapshot release on Sonatype (oss.sonatype.org appears to be down right now...). You can also publish FIRRTL and Chisel3 locally and use that. However, I wouldn't advise either of these approaches unless you're intending to do development of a Chisel3 library.

Sorry about that.

timsliu:
Hey seldridge,

Thanks for the follow up! Is there an alternative way to implement an async reset that doesn't require Chisel 3.3? I'm not wedded to using the ```RequireAsyncReset``` feature - really any method that works will work. 

Thanks for all the hard work, and hope you're staying safe!

Tim

mwachs5:
yes, if you have a higher level wrapping module you can do something like:

```
withReset(reset.asAsyncReset) {
 val foo = Module(new MyModuleIWantToBeAsyncReset())
}
```

timsliu:
Thanks for the quick reply!

We're actually looking for a top level way of making our entire design asynchronously reset instead of just a single module. Is there a top level parameter that can make every reset asynchronous rather than changing individual modules?

Thanks!

Tim

jackkoenig:
In `3.2.x`, your best bet is to make your top-level module a `RawModule`, eg:
```scala
@chisel3.experimental.chiselName
class MyTopModule extends RawModule {
  val clock = IO(Input(Clock()))
  val reset = IO(Input(AsyncReset()))
  // rest of ports
  withClockAndReset(clock, reset) {
    // module instantiations and logic here
    // vals in here will get named by chiselName
  }
}
```

mwachs5:
@timsliu i had the exact same question ;p

timsliu:
Hi everyone,

Thanks for all the quick support!

I'm using Chipyard 1.2. I tried using a RawModule but the problem I have is the Top already extends another class, and so I can't make it also extend RawModule.  I've tried wrapping where my Top is built with a RawModule:

```scala
@chisel3.experimental.chiselName
class MyTopModule extends RawModule {
  val clock = IO(Input(Clock()))
  val reset = IO(Input(AsyncReset()))
  // rest of ports
  withClockAndReset(clock, reset) {
    // module instantiations and logic here
    Module(LazyModule(new Top()(p)).suggestName("top").module)
  }
}
```

and then using this wrapper when I instantiate Top:

```scala
case object BuildTop extends Field[Parameters => Any]((p: Parameters) => Module(new MyTopModule(p)))
```
But it leads to a FIRRTL error 
```
Type mismatch in 'top.reset <= reset'
top.reset: UInt<1>
reset: AsyncReset
```
I'm not sure how to change the type of ```top.reset``` to also be AsyncReset.

@mwachs5 I've also tried your suggestion and built my Top like so:

```scala
case object BuildTop extends Field[Parameters => Any] ((p: Parameters) => {
  withReset(reset.asAsyncReset){
    Module(LazyModule(new Top()(p)).suggestName("top").module)
  }
}
```

But here ```reset``` is not found, probably b/c I'm not doing this inside a module.

Thanks!

Tim

timsliu:
Hey again!

Just wanted to check back in and see where Chisel 3.3 is and if ```RequireAsyncReset``` is available. If not, I'll give the snapshot release a try.

Thanks!

Tim

seldridge:
Chisel 3.3.0-RC1 was published on Friday, so this is now available (and you can verify by querying the latest API documentation):

- https://www.chisel-lang.org/api/latest/chisel3/RequireSyncReset.html
- https://www.chisel-lang.org/api/latest/chisel3/RequireAsyncReset.html

timsliu:
Great, I'll check it out. Thanks for all the hard work on this!!

PR #1017, opened 2019-02-16
Add Scaladoc for chisel3.util.TransitName

seldridge:
Adds some verbose documentation from a test case that I wrote. This could likely be converted to an actual test of TransitName.

This also removes the unneeded "import everything in chisel3". It's not needed here.

**Release Notes**

- Update documentation for util.TransitName

### DIFF
--- a/src/main/scala/chisel3/util/TransitName.scala
+++ b/src/main/scala/chisel3/util/TransitName.scala
@@ -2,25 +2,61 @@
 
 package chisel3.util
 
-import chisel3._
-import internal.HasId
+import chisel3.internal.HasId
 
-/**
- * The purpose of TransitName is to allow a library to 'move' a name
- * call to a more appropriate place.
- * For example, a library factory function may create a module and return
- * the io. The only user-exposed field is that given IO, which can't use
- * any name supplied by the user. This can add a hook so that the supplied
- * name then names the Module.
- * See Queue companion object for working example
- */
+/** The purpose of `TransitName` is to improve the naming of some object created in a different scope by "transiting"
+  * the name from the outer scope to the inner scope.
+  *
+  * Consider the example below. This shows three ways of instantiating `MyModule` and returning the IO. Normally, the
+  * instance will be named `MyModule`. However, it would be better if the instance was named using the name of the `val`
+  * that user provides for the returned IO. `TransitName` can then be used to "transit" the name ''from'' the IO ''to''
+  * the module:
+  *
+  * {{{
+  * /* Assign the IO of a new MyModule instance to value "foo". The instance will be named "MyModule". */
+  * val foo = Module(new MyModule).io
+  *
+  * /* Assign the IO of a new MyModule instance to value "bar". The instance will be named "bar". */
+  * val bar = {
+  *   val x = Module(new MyModule)
+  *   TransitName(x.io, x) // TransitName returns the first argument
+  * }
+  *
+  * /* Assign the IO of a new MyModule instance to value "baz". The instance will be named "baz_generated". */
+  * val baz = {
+  *   val x = Module(new MyModule)
+  *   TransitName.withSuffix("_generated")(x.io, x) // TransitName returns the first argument
+  * }
+  * }}}
+  *
+  * `TransitName` helps library writers following the [[https://en.wikipedia.org/wiki/Factory_method_pattern Factory
+  * Method Pattern]] where modules may be instantiated inside an enclosing scope. For an example of this, see how the
+  * [[Queue$ Queue]] factory uses `TransitName` in
+  * [[https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/util/Decoupled.scala
+  * Decoupled.scala]] factory.
+  */
 object TransitName {
+
+  /** Transit a name from one type to another
+    * @param from the thing with a "good" name
+    * @param to the thing that will receive the "good" name
+    * @return the `from` parameter
+    */
   def apply[T<:HasId](from: T, to: HasId): T = {
     from.addPostnameHook((given_name: String) => {to.suggestName(given_name)})
     from
   }
+
+
+  /** Transit a name from one type to another ''and add a suffix''
+    * @param suffix the suffix to append
+    * @param from the thing with a "good" name
+    * @param to the thing that will receive the "good" name
+    * @return the `from` parameter
+    */
   def withSuffix[T<:HasId](suffix: String)(from: T, to: HasId): T = {
     from.addPostnameHook((given_name: String) => {to.suggestName(given_name + suffix)})
     from
   }
+
 }
--- a/src/test/scala/chiselTests/TransitNameSpec.scala
+++ b/src/test/scala/chiselTests/TransitNameSpec.scala
@@ -1,0 +1,55 @@
+// See LICENSE for license details.
+package chiselTests
+
+import org.scalatest.{FlatSpec, Matchers}
+
+import chisel3._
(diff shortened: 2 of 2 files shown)


edwardcwang:
@seldridge This seems nice/complete enough that we could just make it a full test. Thoughts?

seldridge:
@edwardcwang: probably a good idea to just add this as it isn't tested...

edwardcwang:
Let's add this as a TransitName unit test then?

src/main/scala/chisel3/util/TransitName.scala
```diff
@@ -2,25 +2,129 @@
...
+  *
+  * Consider the example below. This contains a simple module called `Not` that implements the functionality of an
+  * [[https://en.wikipedia.org/wiki/Inverter_(logic_gate) inverter]]. For convenience, this has a companion object that
+  * implements the [[https://en.wikipedia.org/wiki/Factory_method_pattern Factory Method Pattern]] to create instances
+  * of the module `Not`. However, the factory only returns the IO of the instantiated module.
+  *
+  * By using [[TransitName]] you can "transit" the name ''from'' the IO (that the user specifies with a `val`) ''to''
+  * the instance. This results in better names in the resulting FIRRTL or Verilog.
+  *
+  * {{{
```
edwardcwang: Should this example be duplicated or could we just point people to the testcase here?

seldridge:
Added the test. I'd prefer to leave the documentation as is, but am open to cutting it down.

src/main/scala/chisel3/util/TransitName.scala
seldridge: Let me simplify this somewhat... Adding the do not merge tag for now.

seldridge: Simplifying this to only refer to a portion of the testcase.

src/main/scala/chisel3/util/TransitName.scala
```diff
@@ -2,25 +2,57 @@
...
+  *
+  * [[TransitName]] help library writers following the [[https://en.wikipedia.org/wiki/Factory_method_pattern Factory
+  * Method Pattern]] where modules may be instantiated inside an enclosing scope. For an example of this, see the use of
+  * [[TransitName]] in the [[Queue$ Queue]] factory.
+  */
 object TransitName {
+
+  /** Transit a name from one type to another
+    * @param from the thing with a "good" name
+    * @param to the thing to get the "good" name
```
edwardcwang:
```suggestion
    * @param to the thing that will receive the "good" name
```

src/main/scala/chisel3/util/TransitName.scala
```diff
@@ -2,25 +2,57 @@
...
   def apply[T<:HasId](from: T, to: HasId): T = {
     from.addPostnameHook((given_name: String) => {to.suggestName(given_name)})
     from
   }
+
+
+  /** Transit a name from one type to another ''and add a suffix''
+    * @param suffix the suffix to append
+    * @param from the thing with a "good" name
+    * @param to the thing to get the "good" name
```
edwardcwang:
```suggestion
    * @param to the thing that will receive the "good" name
```

src/main/scala/chisel3/util/TransitName.scala
```diff
@@ -2,25 +2,57 @@
...
+  * [[TransitName]] help library writers following the [[https://en.wikipedia.org/wiki/Factory_method_pattern Factory
+  * Method Pattern]] where modules may be instantiated inside an enclosing scope. For an example of this, see the use of
+  * [[TransitName]] in the [[Queue$ Queue]] factory.
+  */
 object TransitName {
+
+  /** Transit a name from one type to another
+    * @param from the thing with a "good" name
+    * @param to the thing to get the "good" name
+    */
```
edwardcwang: Missing `@return`? (ditto below)

src/test/scala/chiselTests/TransitNameSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+/** A top-level module that instantiates three copies of "MyModule" */
+class Top extends RawModule {
+
+  /* The instance name will be named "MyModule" */
+  val foo = Module(new MyModule).io
+
+  /* The instance will be named "bar" */
+  val bar = {
+    val x = Module(new MyModule)
+    TransitName(x.io, x)
```
edwardcwang: This testcase implicitly relies on the return value of TransitName? (I had to think twice for a bit before I realized that `bar` became `x.io`.

src/test/scala/chiselTests/TransitNameSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import org.scalatest.{FlatSpec, Matchers}
+
+import chisel3._
+import chisel3.experimental.RawModule
+import chisel3.util.TransitName
+
+import firrtl.FirrtlExecutionSuccess
+
+/** A NOT gate that inverts its input. */
```
edwardcwang: Out of date comment?

edwardcwang:
I do think the new testcase is more succinct in general modulo a comment.

src/main/scala/chisel3/util/TransitName.scala
seldridge: Fixed in rebase.

src/main/scala/chisel3/util/TransitName.scala
seldridge: Fixed in rebase.

src/main/scala/chisel3/util/TransitName.scala
seldridge: Fixed in rebase. (Good catch... this should help clarify some confusion about how to use this.)

src/test/scala/chiselTests/TransitNameSpec.scala
seldridge: Yeah, `TransitName` is confusing in that it returns the `from` argument. I've clarified this with your above suggestion about adding documentation of the return type. Also, there are now better comments in the testcase.

src/test/scala/chiselTests/TransitNameSpec.scala
seldridge: Fixed in rebase.

src/test/scala/chiselTests/TransitNameSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  /** A top-level module that instantiates three copies of MyModule */
+  class Top extends RawModule {
+
+    /* Assign the IO of a new MyModule instance to value "foo". The instance will be named "MyModule". */
+    val foo = Module(new MyModule).io
+
+    /* Assign the IO of a new MyModule instance to value "bar". The instance will be named "bar". */
+    val bar = {
+      val x = Module(new MyModule)
+      TransitName(x.io, x)
```
edwardcwang: Not to nitpick too much on this:  would it be clearer to write it this way for new users so they don't need to guess at what TransitName returns?
```scala
    val bar = {
      val x = Module(new MyModule)
      TransitName(x.io, x)
      x.io
    }
```

seldridge: All good. Probably. `TransitName` [has been brought up on SO as confusing before](https://stackoverflow.com/questions/49863120/what-is-the-purpose-of-transitname-in-chisel3). Nonetheless, it's one usage (in the `Queue` factory) does it like this. I'm more in favor of the alternative approach so that users are less likely to be confused if they read other people's code.

Issue #411, opened 2016-12-14
ScalaDoc Improvements to Utils


jackkoenig:
The chisel3 ScalaDoc needs some cleaning up.

chisel3.util (Note these are done by file in src/main/scala/chisel3/util)
- [x] Arbiter through CircuitMath - @ducky64 
- [ ] CircuitMath through ImplicitConversions - @donggyukim 
- [ ] LFSR through OneHot - @davidbiancolin 
- [ ] Reg through util - @jackkoenig

chisel3.core
- [x] Vec - @chick 
- [x] Bundle - @chick 

Recall the discussion in #398 that examples belong on the class if it exists, the object otherwise. On apply methods and the like, just document what the different parameters are. For inspiration, look at [Scala List](http://www.scala-lang.org/api/current/scala/collection/immutable/List.html).

Feel free to suggest more items to add.

ducky64:
*poke*

ducky64:
Skimmed through the contents of utils, and everything looks documented. Once the final fixes from #1015 are merged we can close this,

PR #1019, opened 2019-02-16
ScalaDocs improvement for utils Math, MixedVec

ducky64:
Adds recommended usage and examples to the Math utils (log2* and isPow2).
Slight improvement to documentation for MixedVec, primarily deduping method ScalaDoc into the object / class ScalaDoc, and clarifying wires / types.

**Related issue**: #411 

**Release Notes**
Scaladocs for Math (log2*, isPow2) and MixedVec.

### DIFF
--- a/src/main/scala/chisel3/util/Math.scala
+++ b/src/main/scala/chisel3/util/Math.scala
@@ -72,6 +72,8 @@ object log2Down {
 
 /** Compute the log2 of a Scala integer, rounded down.
   *
+  * Can be useful in computing the next-smallest power of two.
+  *
   * @example {{{
   * log2Floor(1) = 0
   * log2Floor(2) = 1
--- a/src/main/scala/chisel3/util/Math.scala
+++ b/src/main/scala/chisel3/util/Math.scala
@@ -8,7 +8,21 @@ package chisel3.util
 import chisel3._
 import chisel3.internal.chiselRuntimeDeprecated
 
-/** Compute the log2 rounded up with min value of 1 */
+/** Compute the log2 of a Scala integer, rounded up, with min value of 1.
+  * Useful for getting the number of bits needed to represent some number of states (in - 1),
+  * To get the number of bits needed to represent some number n, use log2Up(n + 1).
+  * with the minimum value preventing the creation of currently-unsupported zero-width wires.
+  *
+  * Note: prefer to use log2Ceil when in is known to be > 1 (where log2Ceil(in) > 0).
+  * This will be deprecated when zero-width wires is supported.
+  *
+  * @example {{{
+  * log2Up(1) = 1
+  * log2Up(2) = 1
+  * log2Up(3) = 2
+  * log2Up(4) = 2
+  * }}}
+  */
 object log2Up {
   // Do not deprecate until zero-width wires fully work:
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/Math.scala
```diff
@@ -8,7 +8,21 @@ package chisel3.util
...
+/** Compute the log2 of a Scala integer, rounded up, with min value of 1.
+  * Useful for getting the number of bits needed to represent some number of states (in - 1),
+  * To get the number of bits needed to represent some number n, use log2Up(n + 1).
+  * with the minimum value preventing the creation of currently-unsupported zero-width wires.
+  *
+  * Note: prefer to use log2Ceil when in is known to be > 1 (where log2Ceil(in) > 0).
+  * This will be deprecated when zero-width wires is supported.
+  *
+  * @example {{{
+  * log2Up(1) = 1
```
edwardcwang: Should the example be executable? Maybe something like `assert(log2Up(1) == 1)`

src/main/scala/chisel3/util/Math.scala
```diff
@@ -17,7 +31,20 @@ object log2Up {
...
-/** Compute the log2 rounded up */
+/** Compute the log2 of a Scala integer, rounded up.
+  * Useful for getting the number of bits needed to represent some number of states (in - 1).
+  * To get the number of bits needed to represent some number n, use log2Ceil(n + 1).
+  *
+  * Note: can return zero, and should not be used in cases where it may generate unsupported
+  * zero-width wires.
+  *
+  * @example {{{
+  * log2Up(1) = 0
```
edwardcwang: Isn't this `object log2Ceil`?

ducky64:
I was also supposed to write docs for Lookup, but I think we should have a discussion on whether to keep those in util, since I'm not sure if they're used, and they seem like very special-purpose constructs that might be better done by composing other functionality like PriorityMux and Vec.

src/main/scala/chisel3/util/Math.scala
ducky64: Fixed... derp

src/main/scala/chisel3/util/Math.scala
ducky64: Do we have stylistic conventions on that? That's probably good to figure out.

ducky64:
Didn't we discuss putting docs in objects vs. methods during the beginning of the meeting (if that was changed later, then I probably missed it)? I think the reasoning was so that you don't have to expand the function defs when you navigate to the object page.

seldridge:
Yeah, put it into the objective and not the method if you have big examples. Take a look at how it comes out in the scaladoc.

ducky64:
Is there anything blocking review approval for this PR?

src/main/scala/chisel3/util/Math.scala
edwardcwang: poke?

ducky64: Fixed. Didn't realize this was a blocking issue. I opted to put the return value in comments because the assert is not part of the example.

ducky64:
Poke?

PR #1021, opened 2019-02-17
Util doc lsfr

chick:
Update documentation for LSFR16

- Moved bulk of comments to object.
- Added an example
- Added functional test
- example based on section of test

### DIFF
--- a/src/main/scala/chisel3/util/LFSR.scala
+++ b/src/main/scala/chisel3/util/LFSR.scala
@@ -8,10 +8,10 @@ package chisel3.util
 import chisel3._
 import chisel3.internal.naming.chiselName  // can't use chisel3_ version because of compile order
 
-/** LSFR16 generates a 16-bit linear feedback shift register, returning the register contents.
-  * May be useful for generating a pseudo-random sequence.
+/** LFSR16 generates a 16-bit linear feedback shift register, returning the register contents.
+  * This is useful for generating a pseudo-random sequence.
   *
-  * The example, taken from the unit tests.
+  * The example below, taken from the unit tests, creates two 4-sided dice using `LFSR16` primitives:
   * @example {{{
   *   val bins = Reg(Vec(8, UInt(32.W)))
   *
--- a/src/main/scala/chisel3/util/LFSR.scala
+++ b/src/main/scala/chisel3/util/LFSR.scala
@@ -7,13 +7,28 @@ package chisel3.util
 
 import chisel3._
 import chisel3.internal.naming.chiselName  // can't use chisel3_ version because of compile order
-//import chisel3.core.ExplicitCompileOptions.Strict
 
+/** LSFR16 generates a 16-bit linear feedback shift register, returning the register contents.
+  * May be useful for generating a pseudo-random sequence.
+  *
+  * The example, taken from the unit tests.
+  * @example {{{
+  *   val bins = Reg(Vec(8, UInt(32.W)))
+  *
+  *   // Create two 4 sided dice and roll them each cycle.
+  *   // Use tap points on each LFSR so values are more independent
+  *   val die0 = Cat(Seq.tabulate(2) { i => LFSR16()(i) })
+  *   val die1 = Cat(Seq.tabulate(2) { i => LFSR16()(i + 2) })
+  *
+  *   val rollValue = die0 +& die1  // Note +& is critical because sum will need an extra bit.
+  *
+  *   bins(rollValue) := bins(rollValue) + 1.U
+  *
+  * }}}
+  */
 // scalastyle:off magic.number
 object LFSR16 {
   /** Generates a 16-bit linear feedback shift register, returning the register contents.
-    * May be useful for generating a pseudorandom sequence.
-    *
     * @param increment optional control to gate when the LFSR updates.
     */
   @chiselName
--- a/src/test/scala/chiselTests/LFSR16.scala
+++ b/src/test/scala/chiselTests/LFSR16.scala
@@ -13,14 +13,14 @@ import chisel3.util._
   */
 //noinspection TypeAnnotation
 //scalastyle:off magic.number
-class LSFRTester extends BasicTester {
+class LFSRTester extends BasicTester {
   val bins = Reg(Vec(8, UInt(32.W)))
 
   // Use tap points on each LFSR so values are more independent
   val die0 = Cat(Seq.tabulate(2) { i => LFSR16()(i) })
   val die1 = Cat(Seq.tabulate(2) { i => LFSR16()(i + 2) })
 
-  val (trial, done) = Counter(true.B, 100000)
+  val (trial, done) = Counter(true.B, 10000)
 
   val rollValue = die0 +& die1  // Note +& is critical because sum will need an extra bit.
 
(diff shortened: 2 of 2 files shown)


src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,45 @@ import chisel3._
...
+//noinspection TypeAnnotation
+//scalastyle:off magic.number
+class LSFRTester extends BasicTester {
+  val bins = Reg(Vec(8, UInt(32.W)))
+
+  // Use tap points on each LFSR so values are more independent
+  val die0 = Cat(Seq.tabulate(2) { i => LFSR16()(i) })
+  val die1 = Cat(Seq.tabulate(2) { i => LFSR16()(i + 2) })
+
+  val (trial, done) = Counter(true.B, 100000)
```
edwardcwang: How long does this test take to run?

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,45 @@ import chisel3._
...
-  })
-  val res = RegInit(1.U(16.W))
-  when (io.inc) {
-    val nxt_res = Cat(res(0)^res(2)^res(3)^res(5), res(15,1))
-    res := nxt_res
+/**
+  * This test creates two 4 sided dice.
+  * each cycle it adds them together and adds a count to the bin corresponding to that value
+  * The asserts check that the bins show the correct distribution.
+  */
```
seldridge: A possibly better test to consider would be to check for maximum periodicity. I.e., start it at 1 and see how many cycles it takes to get back to 1. That breaks down for bigger LFSRs, but as this is only 16-bit it's tractable.

chick: I can do this if you think it would be better. I kind of like the distribution test as it gives a bit more of the flavor of how one might use the block.

src/test/scala/chiselTests/LFSR16.scala
chick: This takes no time to run. The run-time is negligible compared to the verilator compile time.
I could shorten if you think it's important but I've tried it at even 10 times more and it still takes
no time.

src/test/scala/chiselTests/LFSR16.scala
seldridge: Maybe both? I think the example-motivated (craps odds calculating) approach is great and pays dividends long-term. From personal experience, it's super easy to screw up LFSR taps, though.

If you want a tiny test, how about:
```scala
class LFSR16MaxPeriod extends BasicTester {

  val lfsr16 = LFSR16()
  val started = RegNext(true.B, false.B)

  val (_, wrap) = Counter(started, math.pow(2.0, 16).toInt - 1)
  when (lfsr16 === 1.U && started) {
    chisel3.assert(wrap)
    stop()
  }

}
```

chick: Added your test too

src/test/scala/chiselTests/LFSR16.scala
chick: Shortened it by factor of 10

src/main/scala/chisel3/util/LFSR.scala
```diff
@@ -7,13 +7,28 @@ package chisel3.util
 
 import chisel3._
 import chisel3.internal.naming.chiselName  // can't use chisel3_ version because of compile order
-//import chisel3.core.ExplicitCompileOptions.Strict
 
+/** LSFR16 generates a 16-bit linear feedback shift register, returning the register contents.
+  * May be useful for generating a pseudo-random sequence.
```
seldridge:
```suggestion
  * This may be useful for generating a pseudo-random sequence.
```

src/main/scala/chisel3/util/LFSR.scala
```diff
@@ -7,13 +7,28 @@ package chisel3.util
...
+  *
+  * The example, taken from the unit tests.
```
seldridge:
```suggestion
  * The example below, taken from the unit tests, creates two 4-sided dice using `LFSR16` primitives:
```

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,61 @@ import chisel3._
...
-    val inc = Input(Bool())
-    val out = Output(UInt(16.W))
-  })
-  val res = RegInit(1.U(16.W))
-  when (io.inc) {
-    val nxt_res = Cat(res(0)^res(2)^res(3)^res(5), res(15,1))
-    res := nxt_res
+/**
+  * This test creates two 4 sided dice.
+  * each cycle it adds them together and adds a count to the bin corresponding to that value
```
seldridge:
```suggestion
  * Each cycle it adds them together and adds a count to the bin corresponding to that value.
```

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,61 @@ import chisel3._
...
+  * The asserts check that the bins show the correct distribution.
+  */
+//noinspection TypeAnnotation
```
seldridge: nit: IntelliJ specific comment

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,61 @@ import chisel3._
...
+//scalastyle:off magic.number
+class LFSRTester extends BasicTester {
```
seldridge: This is a cool test. I like it. :game_die: :slot_machine:

src/test/scala/chiselTests/LFSR16.scala
chick: Done

src/test/scala/chiselTests/LFSR16.scala
chick: Done

src/main/scala/chisel3/util/LFSR.scala
chick: Done

src/main/scala/chisel3/util/LFSR.scala
chick: Done

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,60 @@ import chisel3._
...
-    val inc = Input(Bool())
-    val out = Output(UInt(16.W))
-  })
-  val res = RegInit(1.U(16.W))
-  when (io.inc) {
-    val nxt_res = Cat(res(0)^res(2)^res(3)^res(5), res(15,1))
-    res := nxt_res
+/**
+  * This test creates two 4 sided dice.
+  * Each cycle it adds them together and adds a count to the bin corresponding to that value
```
seldridge:
```suggestion
  * Each cycle it adds them together and adds a count to the bin corresponding to that value.
```

src/test/scala/chiselTests/LFSR16.scala
```diff
@@ -6,43 +6,60 @@ import chisel3._
...
+
+  val (trial, done) = Counter(true.B, 10000)
+
+  val rollValue = die0 +& die1  // Note +& is critical because sum will need an extra bit.
+
+  bins(rollValue) := bins(rollValue) + 1.U
+
+  when(done) {
+    printf("bins: %d %d %d %d %d %d %d %d\n",
+      bins(0), bins(1), bins(2), bins(3), bins(4), bins(5), bins(6), bins(7))
```
jackkoenig: I would note that this is more or less what printable is for, you could just do:
```scala
printf(p"bins: $bins\n")
```
I'll do something like:
```
bins: Vec(125, 187, ...)
```

chick:
retest this please

PR #1028, opened 2019-02-25
Docs for ListLookup

ducky64:
Document the beast that is ListLookup. It's something that seems to make a lot more sense with an example than can be described in words or implementation.

Additionally gives background for where this came from (chisel2, probably without much critical evaluation) and where this could go (deprecation).

**Related issue**: #411 

**Release Notes**
Scaladoc for ListLookup / Lookup.

### DIFF
--- a/src/main/scala/chisel3/util/Lookup.scala
+++ b/src/main/scala/chisel3/util/Lookup.scala
@@ -13,7 +13,7 @@ import chisel3._
   *
   * @param addr common select for cases, shared (same) across all list elements
   * @param default default value for each list element, should the address not match any case
-  * @param mapping list of cases, where each entry consists of a BitPat (compared against addr) and
+  * @param mapping list of cases, where each entry consists of a [[chisel3.util.BitPat BitPath]] (compared against addr) and
   *                a list of elements (same length as default) that is the output value for that
   *                element (will have the same index in the output).
   *
@@ -37,7 +37,7 @@ object ListLookup {
 }
 
 /** Muxes between cases based on whether an address matches any pattern for a case.
-  * Similar to MuxLookup, but uses BitPat for address comparison.
+  * Similar to [[chisel3.util.MuxLookup MuxLookup]], but uses [[chisel3.util.BitPat BitPat]] for address comparison.
   *
   * @note This appears to be an odd, specialized operator that we haven't seen used much, and seems
   *       to be a holdover from chisel2. This may be deprecated and removed, usage is not
@@ -45,7 +45,7 @@ object ListLookup {
   *
   * @param addr address to select between cases
   * @param default default value should the address not match any case
-  * @mapping list of cases, where each entry consists of a BitPat (compared against addr) and the
+  * @param mapping list of cases, where each entry consists of a [[chisel3.util.BitPat BitPat]] (compared against addr) and the
   *          output value if the BitPat matches
   */
 object Lookup {
--- a/src/main/scala/chisel3/util/Lookup.scala
+++ b/src/main/scala/chisel3/util/Lookup.scala
@@ -23,7 +23,7 @@ import chisel3._
   *     Array(BitPat(2.U) -> List(20.U, 21.U, 22.U),  // this "row" hardware-selected based off address 2.U
   *           BitPat(3.U) -> List(30.U, 31.U, 32.U))
   * ) // hardware-evaluates to List(20.U, 21.U, 22.U)
-  * // Note: if given address 0.U, the above would hardware evaluate to List(00.U, 01.U, 02.U)
+  * // Note: if given address 0.U, the above would hardware evaluate to List(10.U, 11.U, 12.U)
   * }}}
   */
 object ListLookup {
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/util/Lookup.scala
```diff
@@ -13,6 +36,18 @@ object ListLookup {
...
+/** Muxes between cases based on whether an address matches any pattern for a case.
+  * Similar to MuxLookup, but uses BitPat for address comparison.
+  *
+  * @note This appears to be an odd, specialized operator that we haven't seen used much, and seems
+  *       to be a holdover from chisel2. This may be deprecated and removed, usage is not
+  *       recommended.
+  *
+  * @param addr address to select between cases
+  * @param default default value should the address not match any case
+  * @mapping list of cases, where each entry consists of a BitPat (compared against addr) and the
```
seldridge:
```suggestion
  * @param mapping list of cases, where each entry consists of a BitPat (compared against addr) and the
```

src/main/scala/chisel3/util/Lookup.scala
```diff
@@ -13,6 +36,18 @@ object ListLookup {
   }
 }
 
+/** Muxes between cases based on whether an address matches any pattern for a case.
+  * Similar to MuxLookup, but uses BitPat for address comparison.
```
seldridge:
```suggestion
  * Similar to [[chisel3.util.MuxLookup MuxLookup]], but uses [[chisel3.util.BitPat BitPat]] for address comparison.
```

src/main/scala/chisel3/util/Lookup.scala
```diff
@@ -4,6 +4,29 @@ package chisel3.util
...
+/** For each element in a list, muxes (looks up) between cases (one per list element) based on a
+  * common address.
+  *
+  * @note This appears to be an odd, specialized operator that we haven't seen used much, and seems
+  *       to be a holdover from chisel2. This may be deprecated and removed, usage is not
+  *       recommended.
+  *
+  * @param addr common select for cases, shared (same) across all list elements
+  * @param default default value for each list element, should the address not match any case
+  * @param mapping list of cases, where each entry consists of a BitPat (compared against addr) and
```
seldridge:
```suggestion
  * @param mapping list of cases, where each entry consists of a [[chisel3.util.BitPat BitPath]] (compared against addr) and
```

src/main/scala/chisel3/util/Lookup.scala
```diff
@@ -4,6 +4,29 @@ package chisel3.util
...
+  *                element (will have the same index in the output).
+  *
+  * @example {{{
+  *   ListLookup(1.U,  // address for comparison
+  *       List(      00.U, 01.U, 02.U)   // default "row" if none of the following cases match
+  *       List(1.U, (10.U, 11.U, 12.U),  // this "row" hardware-selected based off address 1.U
+  *            2.U, (20.U, 21.U, 22.U)
+  *       )
+  *   )  // hardware-evaluates to List(10.U, 11.U, 12.U)
+  *   // Note: if given address 0.U, the above would hardware evaluate to List(00.U, 01.U, 02.U)
```
edwardcwang: Not sure the types in this example are fully correct. I tried the following and it appears to work.

(Fun fact, `00.U` appears to be illegal)

```scala
ListLookup(2.U,  // address for comparison
                              List(10.U, 11.U, 12.U),   // default "row" if none of the following cases match
         Array(BitPat(2.U) -> List(20.U, 21.U, 22.U),  // this "row" hardware-selected based off address 2.U
               BitPat(3.U) -> List(30.U, 31.U, 32.U)
         )
     )
// hardware-evaluates to List(20.U, 21.U, 22.U)
```

ducky64: Fixed to your example. I guess that's what I get for not testing my code...

src/main/scala/chisel3/util/Lookup.scala
```diff
@@ -4,6 +4,28 @@ package chisel3.util
...
+  *                a list of elements (same length as default) that is the output value for that
+  *                element (will have the same index in the output).
+  *
+  * @example {{{
+  * ListLookup(2.U,  // address for comparison
+  *                          List(10.U, 11.U, 12.U),   // default "row" if none of the following cases match
+  *     Array(BitPat(2.U) -> List(20.U, 21.U, 22.U),  // this "row" hardware-selected based off address 2.U
+  *           BitPat(3.U) -> List(30.U, 31.U, 32.U))
+  * ) // hardware-evaluates to List(20.U, 21.U, 22.U)
+  * // Note: if given address 0.U, the above would hardware evaluate to List(00.U, 01.U, 02.U)
```
edwardcwang:
```suggestion
  * // Note: if given address 0.U, the above would hardware evaluate to List(10.U, 11.U, 12.U)
```

ducky64: derrrrrp

PR #1032, opened 2019-03-13
Update recommended verilator version to 4.006

ucbjrl:

**Release Notes**
Update recommended Verilator version to 4.006 in README


ucbjrl:
Yes. Both Jenkins and CircleCI are using 4.006. We'd switch to 4.008 but there are multithreading issues that require an additional command line parameter.

PR #974, opened 2019-01-04
Aggregate coverage - aggregate tests but not publishing

ucbjrl:
**Related issue**: #772

**Type of change**: bug report

**Impact**: no functional change

**Development Phase**: implementation

**Release Notes**
Discover a working combination of aggregate usage to enable coverage of subprojects testing but publish a single Jar.
This should finally fix #772.

PR #1040, opened 2019-03-18
Aggregate coverage - aggregate tests but not publishing

### LINKED ISSUE #772: Running coverage under sbt 1.0 does not cover source in chiselFrontEnd

ucbjrl:
Discover a working combination of aggregate usage to enable coverage of subproject testing but publish a single Jar.
Use "scalastyle-test-config.xml" for scalastyle config in tests.
Enable `_` in method names and accept method names ending in `_=`.
Re-sync scalastyle-test-config.xml with scalastyle-config.xml
This should finally fix #772.

**Related issue**: #772, #974, #1037 

**Release Notes**
Support aggregation for the purposes of generating test coverage reports and Scala style checks.


jackkoenig:
Should we turn on https://codecov.io/ now? 🙂I can tell you we will not be happy with the result

PR #1047, opened 2019-03-22
Check field referential equality in autoclonetype

### LINKED ISSUE #1038: Check referential equality of source and destination bundle fields in autoclonetype

**Other information**
Cleaner implementation for an issue exposed in #908, where forgetting an autoclonetype can leave a cloned Bundle field aliased. Probably can copy the tests from #908.

Note that without the Field API #909 there isn't a good suggested resolution that doesn't involve restructuring code so that all Data subtype inputs as passed in as parameters (and automatically clonetyped).

ducky64:
Check field referential equality in autoclonetype, and check type equivalence in all cases.
We should work on getting #909 merged; the current recommended workaround is to pass Chisel types as Bundle arguments to prevent accidental capture.

**Related issue**: Fixes #1038, closes #908  (supersedes)

**Type of change**: bug fix

**Impact**: API modification (the check is stricter, but if it fails this check your code was broken anyways)

**Release Notes**


ducky64:
@grebe this supersedes #908, right?

chick:
@ducky64 grebe just left on vacation for two weeks

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -172,4 +172,15 @@ class AutoClonetypeSpec extends ChiselFlatSpec {
...
   }
+
+  "Aliased fields" should "be caught" in {
+    a [ChiselException] should be thrownBy {
+      elaborate { new Module {
+        val bundleFieldType = UInt(8.W)
+        val io = IO(Output(new Bundle {
+          val a = bundleFieldType
+        }))
+      } }
```
edwardcwang: Is the following also subject to issues related to aliased fields?

```scala
new Module {
        val io = IO(Output(new Bundle {
          val a = UInt(8.W)
        }))
      }
```

ducky64: Shouldn't, because it creates a new UInt on the constructor invocation, and auto-clone-type constructs a new Bundle each time.

edwardcwang: So the following would also have aliasing issues?
```scala
class AliasedBundle(foo: UInt) extends Bundle {
  val a = foo
  val b = foo
}
```

ducky64: Yes, but I don't think that's in the domain of auto-cloneType to check, and under current code it will probably cause a rebinding error anyways.

edwardcwang: What I don't understand is why autoclonetype can't check for `AliasedBundle` when it can check for the example in your test...?

ducky64: It's not a automatically generated clone type issue (as in, issue between the original bundle and cloned bundle), it's your Bundle definition is wrong (problem is solely within one Bundle object). And I think it's also caught by the rebinding checks. If you want I'll add a test case for that?

edwardcwang: Sounds good

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -741,10 +757,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
       clone match {
         case Some(clone) =>
           clone._outerInst = this._outerInst
-          if (!clone.typeEquivalent(this)) {
-            autoClonetypeError(s"automatically cloned $clone not type-equivalent to base." +
-            " Constructor argument values were not inferred, ensure constructor is deterministic.")
-          }
+          validateClone(clone, "Constructor argument values were not inferred, ensure constructor is deterministic.")
```
jackkoenig: Why not run this on creation of `elements`? That will always happen on or before cloning

ducky64:
Added a test for aliasing from an inadequate cloneType. Please review this PR only in the scope of aliasing between an original and a clone.

I'm working on aliasing within a single Bundle, but it appears that's an explicitly ignored case in the Bundle element detection logic. I'll PR that separately, because it redefines a semi-explicitly ignored case and probably merits more discussion.

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -183,4 +183,16 @@ class AutoClonetypeSpec extends ChiselFlatSpec {
       } }
     }
   }
+
+  "Aliased fields from inadequate autoclonetype" should "be caught" in {
+    a [ChiselException] should be thrownBy {
+      class BadBundle(val typeTuple: (Data, Int)) extends Bundle {
+        val a = typeTuple
```
edwardcwang: should this be `typeTuple._1`?

ducky64: lol derp, fixed

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
ducky64: @jackkoenig does #1050 address this concern?

chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
```diff
@@ -658,6 +658,22 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+        autoClonetypeError(s"Automatically cloned $clone not type-equivalent to base $this. " + equivDiagnostic)
+      }
+
+      for ((name, field) <- elements) {
+        if (clone.elements(name) eq field) {
+          autoClonetypeError(s"Automatically cloned $clone has field $name aliased with base $this." +
+            " In the future, this can be solved by wrapping the field in Field(...)," +
+            " see https://github.com/freechipsproject/chisel3/pull/909." +
+            " For now, ensure Chisel types used in the Bundle definition are passed through constructor arguments," +
+            " or wrapped in Input(...), Output(...), or Fliuped(...) if appropriate.")
```
jackkoenig: Slight typo, "Flipped"

PR #1050, opened 2019-03-22
Detect bundle aliasing

ducky64:
From #1047, there was the idea to check if fields aliasing within a Bundle would be problematic. I thought it would be caught by a rebinding error, but apparently it's not - because aliased fields are explicitly ignored in the Bundle field detection logic, the binding logic only finds one field, and binds it without a problem. However, all accesses still point to the aliased element (which IS bound), so you don't get a binding error on access either.

Because there was explicit ignore logic (which I'm not sure about the motivation behind, or how intentional it was) this could be more contentious than the autoclonetype referential equality PR, so it's being separated out for its own discussion.

This does not appear to affect chisel3 regressions, so if there was some intentional reason behind this, it's not a test case. I also traced through the git blame, and `seen` appears to have been introduced (also without explanation) in https://github.com/freechipsproject/chisel3/commit/6f3bc0551ec89e397d2972d7a664915c63e19a06

**Related issue**: #1047

**Type of change**: ???

**Impact**: API modification

**Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -12,6 +12,8 @@ import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo._
 import chisel3.SourceInfoDoc
 
+class AliasedAggregateFieldException(message: String) extends ChiselException(message)
+
 /** An abstract class for data types that solely consist of (are an aggregate
   * of) other Data objects.
   */
@@ -20,6 +22,10 @@ sealed abstract class Aggregate extends Data {
     binding = target
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
+    val duplicates = getElements.groupBy(identity).collect { case (x, elts) if elts.size > 1 => x }
+    if (!duplicates.isEmpty) {
+      throw new AliasedAggregateFieldException(s"Aggregate $this contains aliased fields $duplicates")
+    }
     for (child <- getElements) {
       child.bind(ChildBinding(this), resolvedDirection)
     }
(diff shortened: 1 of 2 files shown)


edwardcwang:
This generally seems sane unless there was a hidden reason for the old behaviour...

ducky64:
@aswaterman (since you wrote the commit that introduced `seen`) do you remember if we were intentionally allowing aliased Bundle fields, and if so, why?
@ucbjrl @chick do you know if any projects (eg rocket) relies on aliased Bundle fields?

aswaterman:
The code originally comes from chisel2:

    commit 26dbafa5ed19085d66c4cf6f88eaff4ae3d7e465
    Author: Huy Vo <huytbvo@berkeley.edu>
    Date:   Mon Feb 6 07:20:33 2012 -0800

        fix to bundle to handle duplicates when overriding a subclass of a Bundle

which provides a pretty clear indication of why we needed to add it when bringing up chisel3.  It is possible that other improvements to bundle-field detection have rendered this irrelevant since then.  I can push this PR through SiFive regression testing to see if it breaks anything on our side.

aswaterman:
(To be clear, it wasn't for allowing duplicates, it was for correct handling of bundle inheritance.)

aswaterman:
Looking through old commits, I'm thinking this was obviated by https://github.com/freechipsproject/chisel3/commit/5272db0bc2ab5a1653168906c636b9099a96348b

That commit filters out duplicates that have the same reflective name, which handles the inheritance issue without the side effect of filtering out explicit duplicates.  I'm pretty sure that's the reason this PR is OK now, but would have failed previously.

I will continue testing, though.

ducky64:
@aswaterman did this end up breaking other code?

Also still needs a review from someone.

aswaterman:
I didn’t see any failures, so I’m sticking to my hypothesis that filtering out duplicates with the same reflective name is the reason we included this code originally.

PR #1051, opened 2019-03-25
Allow naming annotation to work outside builder context

### LINKED ISSUE #1046: Better debug visibility / error messages for chiselName
It should only be used to annotate hardware generation code. Basically, annotated functions will crash if invoked outside a Builder context - because names can't be generated in those cases.

We can think about improving debug visibility, though. Possibilities:
- Improving the error message when a transformed method is invoked outside a Builder context, so that it's more obvious that it's caused by chiselName.
- Having it fail silently when invoked outside a Builder context.

_Originally posted by @ducky64 in https://github.com/freechipsproject/chisel3/pull/1045#issuecomment-475457792_

ducky64:
Exactly as the title says. If there is no Builder context upon entering a naming annotated function, all the naming calls become nops. It does check that the Builder context remained stable at function entry and exit.

**Related issue**: Fixes #1046

**Type of change**: bug fix

**Release Notes**

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -358,7 +358,7 @@ object DynamicNamingStack {
     }
   }
 
-  def popReturnContext[T <: Any](prefix_ref: T, until: internal.naming.NamingContextInterface): T = {
+  def popReturnContext[T <: Any](prefixRef: T, until: internal.naming.NamingContextInterface): T = {
     until match {
       case internal.naming.DummyNamer =>
         require(Builder.namingStackOption.isEmpty,
@@ -366,8 +366,8 @@ object DynamicNamingStack {
       case context: internal.naming.NamingContext =>
         require(Builder.namingStackOption.isDefined,
           "Builder context must remain stable throughout a chiselName-annotated function invocation")
-        Builder.namingStackOption.get.popContext(prefix_ref, context)
+        Builder.namingStackOption.get.popContext(prefixRef, context)
     }
-    prefix_ref
+    prefixRef
   }
 }
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -351,14 +351,14 @@ private[chisel3] object Builder {
   * objects.
   */
 object DynamicNamingStack {
-  def push_context(): internal.naming.NamingContextInterface = {
+  def pushContext(): internal.naming.NamingContextInterface = {
     Builder.namingStackOption match {
-      case Some(namingStack) => namingStack.push_context()
+      case Some(namingStack) => namingStack.pushContext()
       case None => internal.naming.DummyNamer
     }
   }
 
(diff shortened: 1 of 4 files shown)


ducky64:
@chick fixed to camelCase, also simplified some of the logic to push the AnyRef/AnyVal logic into the NamingContext instead of keeping parallel APIs all the way through. Does this address everything?

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -204,6 +203,8 @@ private[chisel3] object Builder {
     val dummy = core.DontCare
   }
 
+  def namingStackOption: Option[internal.naming.NamingStack]= dynamicContextVar.value.map(_.namingStack)
```
edwardcwang:
```suggestion
  def namingStackOption: Option[internal.naming.NamingStack] = dynamicContextVar.value.map(_.namingStack)
```

chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
```diff
@@ -51,22 +77,14 @@ class NamingContext {
...
-  def add_anonymous_descendant(descendant: NamingContext) {
-    anonymousDescendants += descendant
   }
 
-  /** Suggest a name (that will be propagated to FIRRTL) for an object, then returns the object
-    * itself (so this can be inserted transparently anywhere).
-    * Is a no-op (so safe) when applied on objects that aren't named, including non-Chisel data
-    * types.
-    */
   def name[T](obj: T, name: String): T = {
```
edwardcwang: Is there a reason scaladoc is being deleted?

chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
```diff
@@ -76,11 +94,7 @@ class NamingContext {
     obj
   }
 
-  /** Gives this context a naming prefix (which may be empty, "", for a top-level Module context)
-    * so that actual naming calls (HasId.suggestName) can happen.
-    * Recursively names descendants, for those whose return value have an associated name.
-    */
-  def name_prefix(prefix: String) {
+  def namePrefix(prefix: String): Unit = {
```
edwardcwang: ditto

src/test/scala/chiselTests/NamingAnnotationTest.scala
```diff
@@ -215,4 +220,8 @@ class NamingAnnotationSpec extends ChiselPropSpec {
   property("NonNamedFunction should elaborate") {
     elaborate { new NonNamedFunction }
   }
+
+  property("NonBuilderFunction should run outside a Builder context") {
```
edwardcwang: Very cool

chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
ducky64: Moved to the interface definition in the trait

chiselFrontend/src/main/scala/chisel3/internal/Namer.scala
ducky64: ditto

PR #1052, opened 2019-03-25
Implement connectFromBits in ChiselEnum

jackkoenig:
This is necessary to use ChiselEnum in aggregates where things are
casted using .asTypeOf

I also noticed that `ReinterpretCast` and `AsTypeOf` testers were the same so I deleted the duplicates.

We should also consider removing the warning when you cast a non-literal UInt to the enum type because this is called every time you cast, this is especially problematic when you put enums in aggregates. Alternatively, it could be turned on for the API `MyEnum(someUInt)`, but off whenever using `.asTypeOf`...

@hngenc if you could take a look that would be great

**Related issue**: 

**Type of change**: bug fix

**Release Notes**

Fix bugs in casting to ChiselEnums


hngenc:
This change looks good to me. I'm also fine with either turning off the warning completely, or turning it off only when writing `MyEnum(someUInt)`. I suppose a great many people will ignore warnings anyway.

In the future, it may also be nice to automatically add `asserts` to make sure that ChiselEnums aren't invalid.

PR #1053, opened 2019-03-26
Try to eliminate JVM hang due to static initialization deadlock

ucbjrl:

**Release Notes**
When running multiple tests in parallel, we occasionally experience JVM hangs which appear to be due to static initialization deadlock when `Builder` initialization collides with `chisel3.package` initialization of the `DontCare` value.

Let the initialization in `chisel3.package` take responsibility, and assume this will happen before any user code creates (and owns) the `DontCare` element.

See:
- http://ternarysearch.blogspot.com/2013/07/static-initialization-deadlock.html
- https://bugs.openjdk.java.net/browse/JDK-8037567
- https://stackoverflow.com/questions/28631656/runnable-thread-state-but-in-object-wait

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -200,7 +200,14 @@ private[chisel3] object Builder {
 
   // Initialize any singleton objects before user code inadvertently inherits them.
   private def initializeSingletons(): Unit = {
-    val dummy = core.DontCare
+    // This used to contain:
+    //    val dummy = core.DontCare
+    //  but this would occasionally produce hangs dues to static initialization deadlock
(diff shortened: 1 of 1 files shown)


ucbjrl:
My understanding is that the problem is (at least partially) due to thread-safe behavior under the hood, and the fact that threads may acquire locks **in a potentially cyclic order**. I'm not sure if adding   `synchronized` would eliminate that.

chick:
Could well be, you should go ahead and merge as is

jackkoenig:
Do we know why it was there in the first place?

ucbjrl:
Yes. I added it because `DontCare` was being created on first reference as a client/user node. I'm hoping the `val` assignment in `chisel3.package` is sufficient to prevent that. Maybe not.

jackkoenig:
ah yeah I see, alright, looks good to me!

PR #1060, opened 2019-04-10
Hide awkward DecoupledIO .bits attribute from user

johnsbrew:

**Release Notes**

Accessing the elements of a `DecoupledIO(Bundle)` is awkward from an hardware point of view because the ready/valid and the elements of the Bundle does not stand on the same level.
```scala
class MyModule(WIDTH: Int) extends Module {
    val flow = new Bundle {
      val data       = UInt(WIDTH.W)
      // ...
    }
    val io = IO(new Bundle {
      val in  = Flipped(Decoupled(flow))
      val out = Decoupled(flow)
      // ...
    })

    when(io.in.ready){
        io.out.valid     := io.in.valid
        io.out.bits.data := io.in.bits.data
    }
    io.in.ready := io.out.ready
}
```

Hardware engineers do not care about our syntactic sugars and implementation details, they want things to be straight accessible and not being hidden behind an irrelevant `.bits` attribute.
 
This PR enables the following access:
```scala
    when(io.in.ready){
        io.out.valid := io.in.valid
        io.out.data  := io.in.data
    }
```

Note0: renaming `flow` as `bits` in an attempt to bring some *(actually misleading)* consistency is not a solution since it will break as soon as you have 2 distinct DecoupledIOs.

Note1: The match is incomplete, I think it should raise a proper (not the java mismatch) error when  `this.bits` is not a `Bundle`.

Please share your thoughts =)

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -5,6 +5,8 @@
 
 package chisel3.util
 
+import scala.language.dynamics
+
 import chisel3._
 import chisel3.experimental.{DataMirror, Direction, requireIsChiselType}
 import chisel3.internal.naming._  // can't use chisel3_ version because of compile order
@@ -78,9 +80,13 @@ object ReadyValidIO {
   * of ready or valid.
   * @param gen the type of data to be wrapped in DecoupledIO
   */
-class DecoupledIO[+T <: Data](gen: T) extends ReadyValidIO[T](gen)
-{
+class DecoupledIO[+T <: Data](gen: T) extends ReadyValidIO[T](gen) with Dynamic {
   override def cloneType: this.type = new DecoupledIO(gen).asInstanceOf[this.type]
+  def selectDynamic(name: String) = {
+    this.bits match {
+      case b: Bundle => b.elements(name)
+    }
+  }
 }
 
 /** This factory adds a decoupled handshaking protocol to a data bundle. */


johnsbrew:
FYI The test failed for the following reason:
```bash
[info] .target on DecoupledIO
[info] - should not work *** FAILED ***
[info]   Expected a type error, but got none for code: Decoupled(UInt(8.W)).target (ImplicitConversionsSpec.scala:40)
```

Which goes with my question about handling the **not** `Bundle` case.

aswaterman:
Flattening things in this way is going to create confusion when there are fields in the payload with the name 'valid' or 'ready'.  You might never do that, but it does happen, e.g., when making a `Decoupled[Valid[_]]`.

I also have concerns about reducing the amount of static type safety we provide.  This change will move errors to elaboration time that are currently caught at compile time.  When writing generators, the errors might not even manifest in all elaborations.

I would recommend that, if you want to do this sort of thing, you introduce custom types in your designs, rather than changing the ones in chisel3.util.

ducky64:
I think there's also something to be said about consistency - since it's special case syntax that only works on Decoupled, a beginner (or anyone who isn't aware that this is a special case) might wonder why the same thing doesn't work across some different Bundle. Especially if one is looking at the methods overview generated by the ScalaDoc, there would be a bits, ready, and valid field and nothing else. And to understand why this works given a list of member methods, one now needs to understand the Scala Dynamic system on top of all the other Scala stuff needed to use Chisel effectively.

I do think you raise a valid point, but I these are not the trade-offs I would make. It may be worth thinking about how else boilerplate can be minimized within the constraints of the features Scala provides, and whether trade-offs (if any) can be acceptable.

johnsbrew:
Hi @edwardcwang @ducky64 @aswaterman and thank you for your feedbacks :)
I was not actually expecting this proposal to be merged but I rather wanted trigger a debate about the idea of flattening things at some point.
I fully understand your points and I agree that this would probably add even more confusion.

> I would recommend that, if you want to do this sort of thing, you introduce custom types in your designs, rather than changing the ones in chisel3.util.

Yes, definitely.

I just want to bring 2 other points to the discussion.
First from the chisel3 cheatsheet (https://chisel.eecs.berkeley.edu/doc/chisel-cheatsheet3.pdf)
> `DecoupledIO` is a `Bundle` with a ready-valid interface
>

I would tend to say this formulation is misleading: if it is a `Bundle` its elements should be accessible without having to explicitly refer to `.bits.xxx`.
It is indeed actually a `bundle` of 3 elements: `ready`, `valid` and `bits` but from the conception point of view it is a *(`Bundle` of)* `Data` with a ready/valid interface. Then one is free to put a bundle as `Data`.

Last but not least (I know it is not possible to change it by now but I wanted to raise the topic as well) why `bits`? 
In our hardware vocabulary *(which may not be shared far outside of the place I worked)* `bits` would tend to refer to a number of  `bit`, pretty much like a `width`.
Indeed the verilog function `$bits(XXX)` returns the width of `XXX`.

I am glad that a quick search on *ready/valid interface* pointed me straight on this document, a first-year lecture from berkeley: https://inst.eecs.berkeley.edu/~cs150/Documents/Interfaces.pdf
As in many (all?) other places a decoupled interface is a bundle of 3 elements
- ready
- valid
- data 

So why `bits` and not `value` or `val` or `data` ?

ccelio:
> So why bits and not value or val or data?

Fun fact, this artifact of Chisel DecoupledIOs actually harkens back to the primordial times when The Creators used to write Verilog, and the coding style involved writing decoupled interfaces with `val`, `rdy`, and `bits`.  

For example: https://github.com/aswaterman/trainwreck/blob/7b47bb55b53faba3e1ab44209517c5ea8accd861/vector/vuVMU-Ctrl-vec.v

```verilog
  input  [`VMCMD_SZ-1:0] vmu_vcmdq_bits,
  input                  vmu_vcmdq_val,
  output                 vmu_vcmdq_rdy,

  input  [`VMIMM_SZ-1:0] vmu_vbaseq_bits,
  input                  vmu_vbaseq_val,
  output                 vmu_vbaseq_rdy,
  etc.
```

As Andrew said, flattening is a complete no-go, since there are times where the `bits` payload is holding a token that may itself have a `valid` field.

johnsbrew:
> Fun fact, this artifact of Chisel DecoupledIOs actually harkens back to the primordial times when The Creators used to write Verilog, and the coding style involved writing decoupled interfaces with `val`, `rdy`, and `bits`.
> 
> ```verilog
>   input  [`VMCMD_SZ-1:0] vmu_vcmdq_bits,
>   input                  vmu_vcmdq_val,
>   output                 vmu_vcmdq_rdy,
> ```


IMHO `dat` would have made much more sense in this context ;)

jwright6323:
One could create an alias method:

```
def dat: T = this.bits
```

but I'd argue this is just as confusing to a new user, who might expect it to be a different hardware field than `bits` (whose name, however confusing, is so widely used now it could never be changed)

ducky64:
There seems to agreement that this implementation is not what we want, so I'm closing this PR.
Feel free to PR other ideas, or open an issue to continue the discussion of the broader underlying problem.

PR #1061, opened 2019-04-10
Add user-friendly DecoupledIn, DecoupledOut objects

johnsbrew:

**Release Notes**
- `DecoupledIO` being a producer by default is error-prone.
- `Flipped` construct should not to be required to be understood by average users
- `EnqIO` and `DeqIO` companion objects do not make sense in from a generic point of view 

I hence suggest the introduction of 2 companion objects: `DecoupledIn` and `DecoupledOut`

Currently we need the following syntax:
```scala
class MyModule(WIDTH: Int) extends Module {
    val flow = new Bundle {
      val data       = UInt(WIDTH.W)
      // ...
    }
    val io = IO(new Bundle {
      val in  = Flipped(Decoupled(flow))
      val out = Decoupled(flow)
      // ...
    })
   // ...
}
```

With this PR, hardware designers will have a clearer syntax :

```scala
class MyModule(WIDTH: Int) extends Module {
    val flow = new Bundle {
      val data       = UInt(WIDTH.W)
      // ...
    }
    val io = IO(new Bundle {
      val in  = DecoupledIn(flow)
      val out = DecoupledOut(flow)
      // ...
    })
   // ...
}
```

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -104,6 +104,20 @@ object Decoupled
   }
 }
 
+
+/** Decoupled Output from the point of view of a module - drives (outputs) valid and bits, inputs ready.
+  * @param gen The type of data to enqueue
+  */
+object DecoupledOut {
+  def apply[T<:Data](gen: T): DecoupledIO[T] = Decoupled(gen)
+}
+/** Decoupled Input from the point of view of a module - drives (outputs) ready, inputs valid and bits.
+  * @param gen The type of data to dequeue
+  */
+object DecoupledIn {
+  def apply[T<:Data](gen: T): DecoupledIO[T] = Flipped(Decoupled(gen))
+}
+
 /** A concrete subclass of ReadyValidIO that promises to not change
   * the value of 'bits' after a cycle where 'valid' is high and 'ready' is low.
   * Additionally, once 'valid' is raised it will never be lowered until after


edwardcwang:
I'm not intrinsically opposed to this though I'd like to note:

* DecoupledIO being a producer by default is well documented in the Scaladoc and (IIRC) the wiki
* Flipped construct should not to be required to be understood by average users -> `Flipped` is a generically useful construct. Why shouldn't it be understood by users?

jackkoenig:
Because `Flipped(Decoupled(...))` is so common, I think such utilities could be nice, and they have the added benefit of being a little more obvious what's going on at a glance.

I do agree with @edwardcwang though that people need to be aware of `Flipped` because it is necessary for any custom bidirectional Bundles.

Still, I'm always happy for less typing and more self-documenting code 🙂

johnsbrew:
> I'm not intrinsically opposed to this though I'd like to note:
> 
>     - DecoupledIO being a producer by default is well documented in the Scaladoc and (IIRC) the wiki
I agree to disagree, being a producer by default is not explained in most user documentations pieces
- not mentioned in the API: https://chisel.eecs.berkeley.edu/api/3.1.0/chisel3/util/DecoupledIO.html
- not mentioned in cheatsheet: https://chisel.eecs.berkeley.edu/doc/chisel-cheatsheet3.pdf
  
As the notion of *flipped* was mentioned (but not defined) in the cheatsheet, to discover the use of `Flipped()` I went to `chisel3/util/Decoupled.scala` and read the following:
```scala
/** Producer - drives (outputs) valid and bits, inputs ready.
  * @param gen The type of data to enqueue
  */
object EnqIO {
  def apply[T<:Data](gen: T): DecoupledIO[T] = Decoupled(gen)
}
/** Consumer - drives (outputs) ready, inputs valid and bits.
  * @param gen The type of data to dequeue
  */
object DeqIO {
  def apply[T<:Data](gen: T): DecoupledIO[T] = Flipped(Decoupled(gen))
}
``` 

- I just checked the wiki https://github.com/freechipsproject/chisel3/wiki/Interfaces-Bulk-Connections and it appears that you've just modified this page last month which is great.
I have to admit that wiki pages tended at some point to be incomplete/outdated so I haven't counted that much on them when I started to play with Chisel. After some further investigations it looks like their quality and relevance have dramatically improved lately thanks to your numerous contributions @edwardcwang! I will now spend time to RTFM more often 👍 

>     - `Flipped` is a generically useful construct. Why shouldn't it be understood by users?

My point is: I think one should not **NEED** to understand (or even be aware of) it to express basic hardware pieces with chisel.

> I do agree with @edwardcwang though that people need to be aware of Flipped because it is necessary for any custom bidirectional Bundles.

If so I think we are lacking some small pieces of documentation about it: API and cheatsheet.

ducky64:
We discussed this at the dev meeting today, and the results were:
- There's agreement that the idea of a DecoupledIn / DecoupledOut is good syntactic sugar
- We should add more documentation on Flipped and Bundle directionality - which apparently [has just been done](https://github.com/freechipsproject/chisel3/wiki/Bundles-and-Vecs#flipping-bundles)
- There is a broader question of when and where we should add these helper methods, eg should we have ValidIn and ValidOut? Having DecoupledIn / DecoupledOut as a special case doesn't seem like the most principled take, and could be confusing to users expecting similar constructors for other types.
- I'm also a bit concerned about style fragmentation, where there's multiple common ways of expressing something - with the practical effect being potential confusion and needing to learn both ways of doing something. Scala isn't great in this regard (so this isn't a top concern for us), but it would be good to have style guidelines of when to use what.

**The blocking issue** is figuring out an answer to the broader question of what (at least as far as Chisel3 standard library is concerned, and maybe also as guidelines for Chisel users) should get these helper methods.

aswaterman:
`Irrevocable` as well.

sdtwigg:
Why not have DecoupledIO.In and DecoupledIO.Out? It introduces fewer new
top-level identifiers. Also, then it is less surprising when DecoupledIO
still shows up in type errors, etc.

ValidIn and ValidOut seems redundant with Input(Valid()) and Output(Valid())

aswaterman:
I think @sdtwigg's suggestion has merit (though it should use the `Decoupled` companion object, as in `Decoupled.In` or `Decoupled.in`).  Likewise `Irrevocable`.

But I'd still recommend doing the same thing for `Valid` for consistency.

ucbjrl:
retest this please

johnsbrew:
I am not sure we reached a consensus here.
I would also agree that In and Out, despite short does not always make sense.
Actually it only makes sense at module interface.
I would tend to prefer something like `Decoupled.Producer` / `DecProducer` and `Decoupled.Consumer` / `DecConsumer` which make sense in all cases.

I finally made myself comfortable with the use of Flipped, so I am not going to argue strongly in favor of this addition.

ucbjrl:
ok to test

PR #1064, opened 2019-04-11
Fix wrong directionality for Vec(Flipped())

### LINKED ISSUE #1063: Weird binding/directionality issue with Vec(n, Flipped(...))
**Type of issue**: bug report

**Other information**: Chisel 3.2, revision 9f77bf54543724594711f0455ef91f51edfbd22e

Consider the following snippet of code:
```scala
import chisel3._
import chisel3.experimental.{DataMirror, RawModule}
    val a = Input(Bool())
    val b = Output(Bool())
}

class MyModule(val n: Int) extends RawModule {
  val proto = new MyBundle
  // def proto = new MyBundle <-- same problem

  val foo = IO(proto)
  val vecFoo = IO(Vec(n, proto))

  val bar = IO(Flipped(proto))
  val vecBar = IO(Vec(n, Flipped(proto)))

  foo <> DontCare
  vecFoo <> DontCare
  bar <> DontCare
  vecBar <> DontCare

  // directionOf prints the wrong direction deceptively
  println(DataMirror.directionOf(vecBar.head.a)) // returns Output
  println(DataMirror.directionOf(vecBar.head.b)) // returns Input
}
```

The `vecBar` line generates incorrect Verilog:

```verilog
module MyModule(
  input   foo_a,
  output  foo_b,
  input   vecFoo_0_a,
  output  vecFoo_0_b,
  input   vecFoo_1_a,
  output  vecFoo_1_b,

  // IO(Flipped()) works fine
  output  bar_a,
  input   bar_b,

  // But Vec(n, Flipped()) generates wrong directions
  input   vecBar_0_a, // <-- this should be output!
  output  vecBar_0_b, // <-- this should be input!
  input   vecBar_1_a,
  output  vecBar_1_b
);
  assign foo_b = 1'h0;
  assign vecFoo_0_b = 1'h0;
  assign vecFoo_1_b = 1'h0;
  assign bar_a = 1'h0;
  assign vecBar_0_b = 1'h0;
  assign vecBar_1_b = 1'h0;
endmodule
```

edwardcwang:
Create Chisel IR Port() in a way that Converter is happy with.
 Also add more extensive test suite for future-proofing.
 
Close #1063

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/RawModule.scala
@@ -85,7 +85,24 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
       id._onModuleClose
     }
 
-    val firrtlPorts = getModulePorts map {port => Port(port, port.specifiedDirection)}
+    val firrtlPorts = getModulePorts map { port: Data =>
+      // Special case Vec to make FIRRTL emit the direction of its
+      // element.
+      // Just taking the Vec's specifiedDirection is a bug in cases like
+      // Vec(Flipped()), since the Vec's specifiedDirection is
+      // Unspecified.
+      val direction = port match {
+        case v: Vec[_] => v.specifiedDirection match {
+          case SpecifiedDirection.Input => SpecifiedDirection.Input
+          case SpecifiedDirection.Output => SpecifiedDirection.Output
+          case SpecifiedDirection.Flip => SpecifiedDirection.flip(v.sample_element.specifiedDirection)
+          case SpecifiedDirection.Unspecified => v.sample_element.specifiedDirection
+        }
+        case _ => port.specifiedDirection
+      }
+
+      Port(port, direction)
+    }
     _firrtlPorts = Some(firrtlPorts)
 
     // Generate IO invalidation commands to initialize outputs as unused,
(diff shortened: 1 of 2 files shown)


edwardcwang:
Another option is to fix this in `chisel3.internal.firrtl.Port`'s apply method and make `Port` not a case class.

ducky64:
The fix seems structurally odd: what's special about the Vec case that makes it different from the Bundle case (which I don't see special-cased in this code path)? Can the two be unified? And is this bug only for top-level Vecs or should it also be recursive?

edwardcwang:
I'm not sure why the Bundle case works out of the box, and the recursive Vec case needs to be investigated.

What is the intended purpose of `SpecifiedDirection`? Is it considered an internal/experimental API?

The crux of the issue is that Chisel has the notion of separate specified directions for the "Vec itself" and the "Vec element", whereas when emitted to FIRRTL as a port, FIRRTL has only a single notion of direction for the whole port e.g. `input vecFlipped : { flip a : UInt<1>, b : UInt<1>}[1]`.

jackkoenig:
Bundles probably work out of the box because you can flip their children, eg.:
```
output bundlePort : { flip element : UInt<8> }   ; element is now an input
```
But for Vecs, you can't do that:
```
output vecPort : flip UInt<8>[8]  ; This is illegal
input vecPort : UInt<8>[8]        ; this is what you want
```

edwardcwang:
Thanks Chick. I will dig a bit into recursive tests. If I don't get anywhere, we can merge this and open a new PR to fix whatever falls out.

edwardcwang:
Resolution: merge this and open #1074

Issue #1041, opened 2019-03-18
withClock calls withClockAndReset which causes error in RawModule with no reset


chick:
## Bug report
`withReset` and `withClock` both internally use withClockAndReset, which can cause
errors when the container is a `RawModule` which does not have a clock or reset respectively.

## Impact
This should be reviewed, but I believe this has no API consequences other than to no longer cause an error in the case.

## Description
This was reported by @MK.Bin as [stackoverflow question 55209951](https://stackoverflow.com/questions/55209951)

The following code should be legal, but currently errors with **no implicit reset**
```scala
class Test extends RawModule {
  val clock = IO(Input(Clock()))
  val a = withClock(clock) {
    RegNext(~io.in)
  }
}
```

## Expected Behavior
withClock when used in a RawModule with no reset defined should be legal as long as nothing
with the `withClock` scope, references reset directly or indirectly (such as a `Reg` declaration that has an Init)

chick:
This has been fixed

PR #1065, opened 2019-04-11
RawModule with no reset should be able to use withClock method.

chick:
## Motivation
Fixes bug identified in Issue #1041 

## Description
A user defining defining a `Module` with a clock and no reset cannot use the 
`withClock(clock) { val r = RegNext(...) }`. Attempting to do so will generate an exception saying
there is no reset even though no reset is required.

## Details
- refactor ClockAndReset
  - now has `clockOpt: Option[Clock]` and `resetOpt: Option[Reset]` constructor params
  - convenience methods clock and reset tries to deref the option
  - ClockAndReset.empty is factory method for (None, None)
- In Builder
  - forcedClock does not check resetOpt now
  - forcedReset does not check clockOpt now
- withClock no longer looks at resetOpt
- withReset no longer looks at clockOpt
- Module starts with empty ClockAndReset

## Impact
Internal API change.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
@@ -9,7 +9,24 @@ import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.{SourceInfo}
 
-private[chisel3] final case class ClockAndReset(clock: Clock, reset: Reset)
+private[chisel3] final class ClockAndReset(val clockOpt: Option[Clock], val resetOpt: Option[Reset]) {
+  def clock: Clock = clockOpt.get
+  def reset: Reset = resetOpt.get
+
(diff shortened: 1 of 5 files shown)


chick:
@ducky64 good points, I have refactored to simply have two variables since they are orthogonal.

chiselFrontend/src/main/scala/chisel3/core/MultiClock.scala
```diff
@@ -48,7 +56,15 @@ object withReset {  // scalastyle:ignore object.name
     * @param block the block of code to run with new implicit Reset
     * @return the result of the block
     */
-  def apply[T](reset: Reset)(block: => T): T =
-    withClockAndReset(Module.clock, reset)(block)
+  def apply[T](reset: Reset)(block: => T): T = {
+    // Save parentScope
+    val parentReset = Builder.currentReset
+    Builder.currentReset = Builder.currentReset
```
jackkoenig: I believe this should be
```suggestion
    Builder.currentReset = Some(reset)
```

chick: Good catch @jackkoenig, thanks!

PR #1066, opened 2019-04-13
Avoid silently truncating BigInt to Int

aswaterman:
- Introduce internal helper `castToInt`, which issues an error when the input
BigInt can't be represented as Int.

- Use `castToInt` wherever we were using `toInt` in a potentially unsafe way.

Technically, this is an API modification, since previously users could have done such things as `UInt.apply(BigInt.pow(2, 32))`, which would have silently wrapped around and extracted bit 0.

**Impact**: API modification

**Release Notes**


aswaterman:
There are still some places we'd need to truncate to Int, e.g. since Vecs are backed by Scala Vectors.  And other CAD tools barf on very wide Verilog arrays anyway.  So, I don't think switching to BigInts internally would bear much fruit.

PR #1069, opened 2019-04-17
BitPat.pase bug fixed

jijingg:
a bug when use ‘_’ in BitPat string got error width 
`BitPat.pase("b11000000")  = (192,255,8)`
`BitPat.pase("b1100_0000") = (192,255,9)`  got 9bit width actually 8bit

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -32,7 +32,7 @@ object BitPat {
         bits = (bits << 1) + (if (d == '1') 1 else 0)
       }
     }
-    (bits, mask, x.length - 1)
+    (bits, mask, x.filter(_!='_').length - 1)
   }
 
   /** Creates a [[BitPat]] literal from a string.


jackkoenig:
ok to test

jackkoenig:
Looks like Jenkins doesn't listen to me 🙁

ucbjrl:
ok to test

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -32,7 +32,7 @@ object BitPat {
         bits = (bits << 1) + (if (d == '1') 1 else 0)
       }
     }
-    (bits, mask, x.length - 1)
+    (bits, mask, x.filter(_!='_').length - 1)
```
ducky64: I think this is structurally brittle, some it's duplicating the ignore-character logic with the main loop.

The current structure isn't great (lots of use of mutable state), but I think the most consistent way to write it would be to keep a mutable count variable that increments for each character, and use that in the return?

Alternative, you can rewrite everything in a more functional style, but that's probably a bit excessive.

edwardcwang:
Can we add `BitPat.parse("b1100_0000") = (192,255,8)` to the unit tests?

chick:
This has been superseded by PR #1283 which addresses the way the whitespace/underscore was handled and adds whitespace and underscores to the unittest of `BitPat.parse`

PR #1073, opened 2019-04-18
Grouping Chisel API

azidar:

**Release Notes**
At the request of @colinschmidt, added a chisel annotation API for triggering the `firrtl.transforms.GroupComponents` transformation.


edtechexplorers:
LGTM! Thanks!

colinschmidt:
LGTM! Thanks!

src/main/scala/chisel3/util/experimental/group.scala
```diff
@@ -0,0 +1,23 @@
+package chisel3.util.experimental
+
+import chisel3._
+import chisel3.core.{ChiselAnnotation, CompileOptions, RunFirrtlTransform, annotate, requireIsHardware}
+import firrtl.Transform
+import firrtl.transforms.{GroupAnnotation, GroupComponents}
+
+
+object group {
+  def apply[T <: Data](components: Seq[T], newModule: String, newInstance: String, outputSuffix: Option[String] = None, inputSuffix: Option[String] = None)(implicit compileOptions: CompileOptions): Unit = {
```
edwardcwang: Scaladoc would be appreciated :)

ducky64:
Would it make sense for this to be a scoped API? Though I guess the problem there is that the items in the scope aren't 'visible' outside, depending on if they need to be referred to from outside or if it's more modular and encapsulated.

colinschmidt:
The scoping probably won't work for most of my use cases. If I were able to cleanly put a scope around the items I wanted grouped I wouldn't need this pass at all.

jwright6323:
cc @harrisonliew

Is this still in the works?

azidar:
I think it can be backported to 1.2.x?

PR #1076, opened 2019-04-20
Change size of memories from Int to BigInt

jackkoenig:
Relies on https://github.com/freechipsproject/firrtl/pull/1075

**Related issue**: Fixes https://github.com/freechipsproject/chisel3/issues/856

**Release Notes**
Support memories larger than 2^(31)-1 (not supported by most simulators)

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -89,9 +89,9 @@ trait VecFactory extends SourceInfoDoc {
   }
 
   /** Truncate an index to implement modulo-power-of-2 addressing. */
-  private[core] def truncateIndex(idx: UInt, n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = { // scalastyle:ignore line.size.limit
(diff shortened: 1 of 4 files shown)


aswaterman:
It looks like this can break some code, both because it relies on an implicit conversion from BigInt to Int, and because `length` is a public field whose type has changed.

I'd suggest fixing the first set of problems by adding new `apply` methods that accept `Int` (both for the companion-object factories and for the class methods used for indexing the structure).

The dependence on `length` is fundamentally backwards incompatible (it breaks code like `for (i <- 0 until mem.length)`), but I guess we have to suck it up.

ducky64:
For reference, looking through [Bits.scala](https://github.com/freechipsproject/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Bits.scala) indicates that many (if not all) of the methods that take BigInt as an argument also have Int counterparts, so having an Int version of `apply` here would be consistent with the rest of the codebase.

jackkoenig:
Yeah y’all are right of course, will do

jackkoenig:
I added back `Int` versions of the `apply` methods. It *appears* unnecessary to have them for the `do_apply`s as well, although if y'all think I should add those too to be safe I will.

jackkoenig:
retest this please

jackkoenig:
retest this please

PR #1077, opened 2019-04-20
Aspect-Oriented Programming for Chisel

azidar:
**Type of change**: New Feature Implementation

**Development Phase**: Ready to merge, please review!

**Motivation**

The idea behind this work is to increase the direct reuse of Chisel hardware libraries. A common and unfortunate pattern we see in Chisel land is the forking of Chisel library repositories, e.g. the rocket-chip forks required by every Berkeley tape-out, as well as FireSim's rocket-chip use (while not a fork, requires a custom build system).

The fundamental problem is that these projects require adding additional instrumentation, annotations, or minor modifications to a Chisel3 library that are both design- and backend-specific. One example is annotating signals for Midas transformations; this information does not belong in rocketchip (because it is backend-dependent), nor does it belong in Midas (because it is design-specific). What then ends up happening is that FireSim (the project that applies Midas to rocket-chip) must fork rocket-chip to add Midas annotations.

This example is a form of a problem known in software-land as dependency injection, and has many potential solutions, one of which is [aspect-oriented programming](https://www.cs.ubc.ca/~gregor/papers/kiczales-ECOOP1997-AOP.pdf). While its now generally regarded as an anti-pattern in software, I'm hypothesizing that not only does this paradigm make sense in the hardware-generation context, its already present in the form of tcl scripts for CAD tools and is being exacerbated by the power of Chisel parameterization.

This PR contains modifications to Chisel that enable aspect-oriented dependency injection. It will be experimental (and thus no claims for preserving backwards-compatibility), but we want to see whether this feature can indeed address the previously described reuse problem.

**Next Steps**

For some example use-cases demonstrating the power of Chisel Aspects, check out the following tests in the aop branch of riscv-mini:
- Logging and histogramming [example](https://github.com/ucb-bar/riscv-mini/blob/aop/src/test/scala/myminiTests/MyTests.scala)
- Redundancy [example](https://github.com/ucb-bar/riscv-mini/blob/aop/src/test/scala/myminiTests/RedudancyTests.scala)

For some concrete implementations of Chisel Aspects, check out the following. Please note that all of these aspects ***do not require any modifications to riscv-mini's source code!!!***
- Injecting arbitrary Chisel code (included in this PR)
- Histogramming signal values during test execution, then printing the histogram bins results at the end - [HistogrammingAspect](https://github.com/ucb-bar/riscv-mini/blob/aop/aoplib/src/main/scala/aoplib/histogram)
- Creating triple-redundant registers - [RedundancyAspect](https://github.com/ucb-bar/riscv-mini/blob/aop/aoplib/src/main/scala/aoplib/redundancy/RedundancyAspect.scala)
- Injecting stuck-at faults for signals - [StuckFaultAspect](https://github.com/ucb-bar/riscv-mini/blob/aop/aoplib/src/main/scala/aoplib/redundancy/StuckFaultAspect.scala)

**Outline of Software Architecture**

`Aspect[T]` --- `T` is type of the top-level module
- Overview:
  - Represents an aspect of a top-level design. Resolving an aspect means executing its `toAnnotation` function prior to executing FIRRTL. These generated FIRRTL annotations determine the behavior of the aspect. Aspect ordering is relative only to how the annotations are ordered/consumed by FIRRTL transforms (e.g. every aspect is lazily applied to the circuit whenever the FIRRTL complier runs a transform that consumes an aspect-generated annotation). This is similar to how annotations work in Chisel - the order a signal is annotated does not matter; what matters is when the annotation is consumed.
- Unimplemented Members:
  - `def toAnnotation(dut: DUT): Annotation`
- Aspect Reuse
  - Can reuse class, but it is tied to the top-level module. If you want to reuse for a submodule, you can call a sub-module's aspect within the top-level-module's aspect.

**TODO:**
- [X] Update to master
- [x] Finish ScalaDoc for Select functions
- [x] Add tests of Select functions


chiselFrontend/src/main/scala/chisel3/core/ModuleAspect.scala
```diff
@@ -0,0 +1,23 @@
+package chisel3.core
```
chick: File Header wrong

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -0,0 +1,418 @@
...
+  /** Used to represent a connection or bulk connection
+    *
+    * Additionally contains the sequence of when predicates seen when the connection is declared
+    *
+    * @param preds
+    * @param loc
+    * @param exp
+    * @param isBulk
+    */
+  case class PredicatedConnect(preds: Seq[Predicate], loc: HasId, exp: HasId, isBulk: Boolean) extends Serializeable {
```
azidar: make data

src/main/scala/chisel3/aop/injecting/InjectingTransform.scala
```diff
@@ -0,0 +1,46 @@
...
+      case other => Seq(other)
+    }
+
+    // Append all statements to end of corresponding modules
+    val newModules = state.circuit.modules.map { m: ir.DefModule =>
+      m match {
+        case m: ir.Module if addStmtMap.contains(m.name) =>
+          m.copy(body = ir.Block(m.body +: addStmtMap(m.name)))
+        case m: _root_.firrtl.ir.ExtModule if addStmtMap.contains(m.name) =>
+          ir.Module(m.info, m.name, m.ports, ir.Block(addStmtMap(m.name)))
```
seldridge: Should this case be allowed? This seems to allow you to override the defname of an `ExtModule` using aspects. However, this name override is being derived from an `ir.Statement`. This seems odd... Is there an intended use case for this right now?

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -0,0 +1,423 @@
...
+      case l: LitArg => l.num.intValue().toString
+    }
+  }
+
+  // Collects when predicates as it searches through a module, then applying processCommand to non-when related commands
+  private def searchWhens(module: BaseModule, processCommand: (Command, Seq[Predicate]) => Unit) = {
+    check(module)
+    module._component.get.asInstanceOf[DefModule].commands.foldLeft((Seq.empty[Predicate], Option.empty[Predicate])) {
+      (blah, cmd) =>
+        //println(s"On cmd: $cmd with $blah")
```
seldridge: nit: remove

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -0,0 +1,423 @@
...
+  /** Collects all components selected by collector within module and all children modules it instantiates
+    *   directly or indirectly
+    * Accepts a collector function, rather than a collector partial function (see [[collectDeep]])
+    * @param module Module to collect components, as well as all children module it directly and indirectly instantiates
+    * @param collector Collector function to pick, given a module, which components to collect
+    * @param tag Required for generics to work, should ignore this
+    * @tparam T Type of the component that will be collected
+    * @return
+    */
+  def getDeep[T](module: BaseModule)(collector: BaseModule => Seq[T])(implicit tag: TypeTag[T]): Seq[T] = {
```
seldridge: Do you need the `TypeTag` here? Same thing for `getDeep` with the partial function parameter.

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -0,0 +1,423 @@
...
+    }
+  }
+
+  /** Selects a kind of arithmetic or logical operator directly instantiated within given module
+    * The kind of operators are contained in [[chisel3.internal.firrtl.PrimOp]]
+    * @param opKind the kind of operator, e.g. "mux", "add", or "bits"
+    * @param module
+    * @return
+    */
+  def ops(opKind: String)(module: BaseModule): Seq[Data] = {
```
seldridge: Should this use a `String` parameter or `chisel3.internal.firrtl.PrimOp`? I expect your motivation here is to not rely on `chisel3.internal.firrtl`. I think that's fine, just wanted to clarify.

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -0,0 +1,423 @@
...
+    val sensitivitySignals = getIntermediateAndLeafs(signal).toSet
+    val predicatedConnects = mutable.ArrayBuffer[PredicatedConnect]()
+    val isPort = module._component.get.asInstanceOf[DefModule].ports.flatMap{ p => getIntermediateAndLeafs(p.id) }.contains(signal)
+    var prePredicates: Seq[Predicate] = Nil
+    var seenDef = isPort
+    searchWhens(module, (cmd: Command, preds) => {
+      cmd match {
+        case cmd: Definition if cmd.id.isInstanceOf[Data] =>
+          val x = getIntermediateAndLeafs(cmd.id.asInstanceOf[Data])
+          //println(s"Does ${cmd.name} with $x contain $signal? ${x.contains(signal)}")
```
seldridge: nit: remove

src/main/scala/chisel3/stage/phases/MaybeAspectStage.scala
```diff
@@ -0,0 +1,18 @@
+// See LICENSE for license details.
+
+package chisel3.stage.phases
+
+import chisel3.aop.Aspect
+import firrtl.AnnotationSeq
+import firrtl.options.Phase
+
+/** Run [[firrtl.stage.FirrtlStage]] if a [[chisel3.stage.NoRunFirrtlCompilerAnnotation]] is not present.
```
seldridge: Wrong Scaladoc.

chiselFrontend/src/main/scala/chisel3/Printable.scala
```diff
@@ -157,22 +161,40 @@ object FirrtlFormat {
   }
 }
 /** Format bits as Decimal */
-case class Decimal(bits: Bits) extends FirrtlFormat('d')
+case class Decimal(bits: Bits) extends FirrtlFormat('d') {
+  override def length: Int = bits.widthOption match {
+    case Some(w) => Math.ceil(Math.log10(Math.pow(2, w) + 1)).toInt
+    case None => 0
+  }
+}
```
seldridge: Would you mind factoring out these changes from this PR? Or: are these critical to the functioning of this PR?

chiselFrontend/src/main/scala/chisel3/aop/Aspect.scala
```diff
@@ -0,0 +1,45 @@
...
+import firrtl.{AnnotationSeq, RenameMap, Transform}
+
+import scala.reflect.runtime.universe.TypeTag
+
+/** Represents an aspect of a Chisel module, by specifying
+  *  what behavior should be done to instance, via the FIRRTL Annotation Mechanism
+  * @param dutTag Needed to prevent type-erasure of the top-level module type
+  * @tparam T Type of top-level module
+  */
+abstract class Aspect[T <: RawModule](implicit dutTag: TypeTag[T]) extends Annotation with Unserializable {
```
seldridge: Is the `TypeTag` necessary here?

chiselFrontend/src/main/scala/chisel3/Mem.scala
```diff
@@ -41,7 +41,7 @@ object Mem {
     do_apply(BigInt(size), t)(sourceInfo, compileOptions)
 }
 
-sealed abstract class MemBase[T <: Data](t: T, val length: BigInt) extends HasId with NamedComponent with SourceInfoDoc {
+sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends HasId with NamedComponent with SourceInfoDoc {
```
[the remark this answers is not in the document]
azidar: Yes, but also for the `Select` stuff.

chiselFrontend/src/main/scala/chisel3/ModuleAspect.scala
```diff
@@ -0,0 +1,26 @@
+// See LICENSE for license details.
+
+package chisel3
+
+import chisel3.internal.Builder
+import chisel3.experimental.RawModule
+
+/** Used by Chisel Aspects to inject Chisel code into modules, after they have been elaborated.
+  * This is an internal API - don't use!
```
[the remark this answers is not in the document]
azidar: It can! Will do.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -180,6 +208,9 @@ private[chisel3] class DynamicContext() {
   val components = ArrayBuffer[Component]()
   val annotations = ArrayBuffer[ChiselAnnotation]()
   var currentModule: Option[BaseModule] = None
+  // Contains a mapping from a elaborated module to their aspect
+  // Set by [[ModuleAspect]]
+  val aspectModule: mutable.HashMap[BaseModule, BaseModule] = mutable.HashMap.empty[BaseModule, BaseModule]
```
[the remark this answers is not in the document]
azidar: So you can't have multiple aspects. This would only happen if you were aspecting a ModuleAspect, which you can't do given its an internal API. Basically this is only for the InjectingAspect, not Aspects generally.

chiselFrontend/src/main/scala/chisel3/Annotation.scala
```diff
@@ -26,10 +26,22 @@ trait ChiselAnnotation {
...
+
+  override def transformClasses: Seq[Class[_ <: Transform]] = Seq(transformClass)
 }
 
+/** Mixin for [[ChiselAnnotation]] that instantiates multiple associated FIRRTL Transforms when this
+  * Annotation is present during a run of [[chisel3.Driver.execute]]. Automatic Transform
+  * instantiation is *not* supported when the Circuit and Annotations are serialized before invoking
+  * FIRRTL.
+  */
+trait RunFirrtlTransforms extends ChiselAnnotation {
```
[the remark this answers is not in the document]
azidar: Correct - this PR doesn't need it anymore, but there still could be cases where an annotation wants to trigger more than one annotations. Perhaps we should instead deprecate `RunFirrtlTransform` and `toFIRRTL`, and instead require implementing a `toAnnotationSeq` method?

src/main/scala/chisel3/stage/AspectStage.scala
```diff
@@ -0,0 +1,38 @@
...
+import firrtl.AnnotationSeq
+import firrtl.options.{Shell, Stage}
+
+import scala.collection.mutable
+
+/** Stage associated with all Aspects
+  *
+  * Consumes the [[chisel3.stage.DesignAnnotation]] and converts every [[Aspect]] into their annotations prior to executing FIRRTL
+  */
+class AspectStage extends Stage {
```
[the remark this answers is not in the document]
azidar: I think you're right, I'll convert it to a Phase.

src/main/scala/chisel3/aop/injecting/InjectingAspect.scala
```diff
@@ -0,0 +1,63 @@
...
+      val modules = Aspect.getFirrtl(chiselIR.copy(components = comps)).modules.flatMap {
+        case m: firrtl.ir.Module if m.name == module.name =>
+          stmts += m.body
+          Nil
+        case other =>
+          Seq(other)
+      }
+
+      InjectStatement(ModuleTarget(circuit, module.name), ir.Block(stmts), modules, annotations)
+    }.toList
```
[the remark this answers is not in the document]
azidar: Nope. Would you rather I change it?

src/main/scala/chisel3/aop/injecting/InjectingAspect.scala
```diff
@@ -0,0 +1,63 @@
...
+  */
+case class InjectingAspect[T <: RawModule,
+                           M <: RawModule](selectRoots: T => Iterable[M],
+                                           injection: M => Unit
+                                          )(implicit tTag: TypeTag[T]) extends Aspect[T] {
+  final def toAnnotation(top: T): AnnotationSeq = {
+    toAnnotation(selectRoots(top), injection, top.name)
+  }
+
+  final def toAnnotation(modules: Iterable[M], inject: M => Unit, circuit: String): AnnotationSeq = {
```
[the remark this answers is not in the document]
azidar: Good catch, I imagine this was from an old refactoring.

src/main/scala/chisel3/aop/injecting/InjectingTransform.scala
azidar: So the name is from the `InjectStatement` which is an annotation, and the module name inside it refers to the module it is injecting the statements to. I had decided its ok to inject statements into an `ExtModule` by turning it into a normal `Module`. Do you think that is ok?

src/main/scala/chisel3/aop/Select.scala
azidar: Yeah you're right, its unnecessary.

src/main/scala/chisel3/aop/Select.scala
azidar: Yeah, I want to keep the `internal.firrtl` stuff as hidden away as possible.

chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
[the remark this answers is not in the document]
seldridge: Okay, that sounds good!

chiselFrontend/src/main/scala/chisel3/Printable.scala
azidar: No, they aren't necessary. They are only useful for one of my aspect libraries. I can factor it out.

chiselFrontend/src/main/scala/chisel3/aop/Aspect.scala
azidar: Nope, it isn't, I'll remove it. The only thing is its usually pretty important for Aspect descendants to have a typetag.

chiselFrontend/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
seldridge: I think for the purposes of keeping this PR sharp, would you mind removing this?

I am in agreement that there either needs to be an API like this for running multiple transforms, or this needs to be deprecated in favor of a method that returns more than one annotation.

I'm partial towards `toAnnotationSeq`.

src/main/scala/chisel3/aop/injecting/InjectingAspect.scala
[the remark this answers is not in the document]
seldridge: Would you mind? I think sticking with `Seq`, unless we really want `List`, is preferable considering everything in the code base is `Seq`.

src/main/scala/chisel3/aop/injecting/InjectingTransform.scala
seldridge: Ah, it's rewriting to an `ir.Module`. I missed that.

I think there should be some documentation about this, but I think this behavior makes sense. An `ExtModule` is just something with a to-be-defined implementation. That could be defined in Verilog, but it could be defined by an aspect. Nice thinking ahead here.

chiselFrontend/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
azidar: Ok I'll just back out `RunFirrtlTransforms` and leave the deprecation to another PR.

azidar:
I believe I've addressed all your comments! Some of the TypeTags were unnecessary - the only thing is they may be necessary for subclasses of Aspect. I think, however, that's something we can talk about in a tutorial or people can pick-up from the examples that exist.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -46,8 +46,9 @@ case class ChiselGeneratorAnnotation(gen: () => RawModule) extends NoTargetAnnot
 
   /** Run elaboration on the Chisel module generator function stored by this [[firrtl.annotations.Annotation]]
     */
-  def elaborate: ChiselCircuitAnnotation = try {
-    ChiselCircuitAnnotation(Builder.build(Module(gen())))
+  def elaborate: AnnotationSeq  = try {
```
ucbjrl: This breaks testers2 which expects to be able to reference a circuit:
```
[info] Compiling 29 Scala sources to .../testers2/target/scala-2.11/classes ...
[error] /vm/home/jenkins/workspace/chisel3-pl_2/testers2/src/main/scala/chisel3/tester/backends/treadle/TreadleExecutive.scala:30:49: value circuit is not a member of firrtl.AnnotationSeq
[error]     val circuit = generatorAnnotation.elaborate.circuit
[error]                                                 ^
```

seldridge: Since we're still pre-3.2, is it safe to fix this on the testers2 side? Something like:
```scala
val circuit = generatorAnnotation.elaborate._1.circuit
```

PR #1079, opened 2019-04-22
Make asTypeOf work for bundles with zero-width fields.

### LINKED ISSUE #1075: asTypeOf broken for bundles with zero-width fields
**Type of issue**: bug report

The following dies with builder errors about invalid bit range (-1, 0) and (0, 1)

```
import chisel3._
class B extends Bundle {
  val b = UInt(1.W)
  val c = UInt(0.W)
}
class A extends experimental.MultiIOModule {
  val in = IO(Input(UInt(10.W)))
  val out = IO(Output(new B))
  out := in.asTypeOf(new B)
}
Driver.execute(Array[String](), () => new A)
```

grebe:
Closes #1075.

**Related issue**: #1075 

**Type of change**: bug fix

**Impact**: API modification

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/Aggregate.scala
@@ -70,8 +70,16 @@ sealed abstract class Aggregate extends Data {
     var i = 0
     val bits = WireDefault(UInt(this.width), that)  // handles width padding
     for (x <- flatten) {
-      x.connectFromBits(bits(i + x.getWidth - 1, i))
-      i += x.getWidth
+      val fieldWidth = x.getWidth
+      if (fieldWidth > 0) {
+        x.connectFromBits(bits(i + fieldWidth - 1, i))
+        i += fieldWidth
+      } else {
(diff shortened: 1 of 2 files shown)


grebe:
Looks like the recent firrtl mem size switch from Int to BigInt is biting treadle tests on millenium for whatever reason. I think it's spurious.

ducky64:
More generally what else needs to be done to better support zero width wires?

Also, another implementation path, cleaner here and potentially applicable elsewhere, may be a bit selection operation that takes a start index and size, so the zero-width case doesn't need to be special-cased.

aswaterman:
A feature like Verilog's `foo[start +: length]` has been requested elsewhere and would be nice to have.  Could always rewrite this PR once such a thing exists.

grebe:
Agreed on (size, length) making it cleaner to implement this. I think this bugfix is worth pushing through without it, but it would certainly be nice to have that feature. I just made an issue.

aswaterman:
Yeah, I don't think it should block this PR

jackkoenig:
retest this please

PR #1088, opened 2019-05-06
PRNG and LFSR Utils

seldridge:
Adds PRNG and LFSR utilities.

- [x] Add LFSR type hierarchy (Galois, Fibonacci)
- [x] Allow for compile-time tap generation
- [x] Add maximum period Fibonaccis (the taps are the same for Fibonacci and Galois)
- [x] Add a lot of know maximum perdio taps: `(2 to 768) ++ Seq(1024, 2048, 4096)`
- [x] Make seed optional and reset in LFSRs based on XOR/XNOR usage

**Related Issues**:
- #818 

**Release Notes**

- Add chisel3.util.random package with Galois and Fibonacci LFSRs
- Add maximal period LFSR generation and maximal period taps
- Deprecate chisel3.util.LFSR16 in favor of chisel3.util.random.LFSR(16)


seldridge:
(I thought this was a draft PR when I submitted so this is slightly half-cocked currently. Oops.)

I need to double check that the taps are the same on LFSR16, but we should be able to deprecate that.

Not hard. Scaladoc is definitely needed...

seldridge:
Going through this, some questions arose:

1. Is there a reason to favor Galois or Fibonacci LFSRs? The original `LFSR16` is a Fibonacci while this generator provides Galois. There are other approaches that involve SRAMs that would avoid explicit data movement if this used a Fibonacci implementation.
2. What range of LFSR widths should we support? This provides 2--32 and 64, 128, 256, 512, 1024, 2048, 4096. The latter are likely not useful in a real world situation?
3. How should large LFSRs be tested? @chick uses a distribution based approach. This starts to break down for very large widths, e.g., > 1024.
4. `chisel3.testers.TesterDriver.execute` turns on VCD generations by default. When using `assertTesterPasses` this results in some very large VCDs getting generated for no real reason.

src/main/scala/chisel3/util/random/Lfsr.scala
```diff
@@ -0,0 +1,137 @@
...
+  * //  - 11: 0b1001
+  * //  - 12: 0b1000
+  * //  - 13: 0b0100
+  * //  - 14: 0b0010
+  * //  - 15: 0b0001
+  * }}}
+  * @param width LFSR width
+  * @param seed an initial seed (this cannot be zero)
+  */
+class Lfsr(width: Int, seed: BigInt = 1) extends Module {
```
edwardcwang: Does it ever make sense to have a LFSR with a hardware programmable seed?

edwardcwang: Also minor nitpick, is `chisel3.util.random.Lfsr` preferred or `chisel3.util.random.LFSR`?

src/main/scala/chisel3/util/random/Lfsr.scala
```diff
@@ -0,0 +1,137 @@
...
+  require(seed > 0, "Seed cannot be zero")
+  require(Lfsr.tapsGalois.contains(width), s"No LFSR taps stored for requested size '$width'")
```
edwardcwang: Is it the case that each sized tap contains only one polynomial (i.e. does it make sense for the user to specify a custom sequence of taps)?

Of course the common case should be easy/fast as it is now.

src/main/scala/chisel3/util/random/Lfsr.scala
johnsbrew: Yes it does, you might want to update your seed on a regular basis to increase the security of some applications :)

seldridge: Yes, it is needed. Fortunately, support for it is already there. :heavy_check_mark: There's a `Valid[UInt]` interface that'll let you do that. This should probably be shown in the examples... Can update it.

src/main/scala/chisel3/util/random/Lfsr.scala
seldridge: I think there are two use cases: 
1. The current `LFSR16` case of a user wanting a decent pseudo random source. As far as LFSRs go, my understanding is that if they're maximal period then it's good enough. We then provide (hopefully) solutions for all the widths a user cares about. The cited [table](http://courses.cse.tamu.edu/walker/csce680/lfsr_table.pdf) for `2 to 786` which we could scrape and populate here.
2. A user is constructing specific LFSRs for some reason. In that situation, it may make sense to think about an API for constructing general Galois or Fibonacci LFSRs.

I think this PR covers (1) relatively thoroughly and can perhaps be directly improved with more known maximal period LFSR taps.

An API for (2) probably makes sense as well. One thing here is that I think you can only cleanly express a Fibonacci LFSR as a nice polynomial. I.e., `x^4 = x^0 + x^2` is all you need to know for a Fibonacci LFSR and it's correct. All the other `x^3 = x^4, x^2 = x^3, ...` are implicit. For Galois LFSRs you need a sequence of equations to correctly express it. Both are concisely handled as `Set[Int]` passed to a `GaloisLfsr` or `FibonacciLfsr` constructor.

All that said... these are perhaps most cleanly expressed as matrices and it may make sense to just use that under the hood. [The ECC ideas I'm playing around with currently do something like that](https://github.com/freechipsproject/chisel3/pull/1087/files#diff-06af32e6f783e29d75541b98a04471b4R9).

edwardcwang:
@seldridge Thanks for adding the ScalaDoc and deprecating the old LFSR!

Re: Galois or Fibonacci LFSRs, I don't have a strong preference on this directly, though are the polynomials the same? Less data movement sounds appealing on paper

Re: testing, I agree that large LFSRs can't easily be exhaustively tested by distribution for reasons as you mention. Maybe for those we can have a software LFSR model that gets used in parallel to test those cases specifically?

seldridge:
I was trying to dig into this yesterday. Polynomials are different between Galois and Fibonacci. My understanding is that they can produce the exact same cycle-by-cycle behavior in one bit of output (with a temporal offset), but their internal state traversal will be different. From a hardware implementation, they're basically the same if you assume shift registers. Galois may be slightly better timing-wise as they have the same fan-out and less gate delay. However, you can, I think, implement these with SRAMs and possibly make things more efficient.

In terms of code generation, Fibonacci can be more concisely expressed in FIRRTL as the predominant operation is a shift of a `UInt`. Without subword assignment, a Glaois would seem to either work on `Vec[Bool]` or require expanding via `UInt.asBools`. This may have implications for backend tools and/or simulation efficiency.

src/main/scala/chisel3/util/random/Lfsr.scala
seldridge: I think this should be `LFSR`. I don't know if we have a formal convention on this, but most things that are acronyms are uppercase, e.g., `ValidIO`. For consistency this is changed in the rebase.

seldridge: The example is updated to show this for the class. The run-time programmable seed is only available in the class and not the companion object.

jwright6323:
I've only skimmed this PR and it seems like most of these features are there, but it might be helpful for me to list them out anyway. We use fibonacci LFSRs/PRBSs for serial link bit-error measurements [here](https://github.com/ucb-bar/chisel-awl/blob/f7f671a97b203443cac235d0c9a93287ce7374bf/src/main/scala/BertDebug.scala#L121-L169), and some of the features I'd need to replace this with a core chisel LFSR would be:
- parametrizable bit outputs (= num state transitions) per cycle
- a boolean enable signal or state to control whether the LFSR state advances on a given cycle
- run-time programmable seed by parallel load
- run-time programmable seed by looking at a reference word (that may or may not be longer than the LFSR depth). If the output bits do not include the current state of the LFSR, this can be trivial case of the parallel load case for `word_bits >= lfsr_bits`, but requires access to the LFSR state for `word_bits < lfsr_bits`. If the output bits do include the current state of the LFSR then you need to pre-calculate the next state seed based on the reference word.
- We would probably want easy control over the feedback polynomial. Looking at polynomials listed here, I believe my needs are covered, but I'm sure the override would come in handy.

src/main/scala/chisel3/util/random/Lfsr.scala
seldridge: In the rebase this has been updated to allow a user to specify custom taps. If the user just wants a maximal period LFSR there are classes/objects to help with that.

aswaterman:
Here’s a feature request... I’d like the ability to minimally seed an LFSR, i.e. reset only the LSB and leave the remaining bits up to chance. This is both useful for testing using RTL sim (better coverage over more runs) and for reducing reset fanout.

It’s fine if that’s not directly supported in this API, as long as enough of the internals are exposed so that I can do that myself.

seldridge:
Lots of updates to this and I think it's almost ready for a review. Galois and Fibonacci are now both supported (and using XOR or XNOR for the reduction). Maximum period taps are the same for all, though internal state transitions will be different.

In terms of @jwright6323 requirements:
- [ ] (To be updated in rebase) Support for compile-time configuration to increment by a parameterized number of cycles
- [x] A `PRNGIO.increment: Bool` line can be used to control whether the PRNG advances
- [x] The seed (internal PRNG state) can be initialized at compile-time and set at run time using `PRNGIO.seed: Valid[UInt]`
- [ ] (To be update in rebase) Compile-time parameter to either (1) load the state from the seed or (2) load the state from the next state as determined by the seed
- [x] The feedback polynomial/taps can be set at compile-time. If a user wants to just use a maximal period polynomial/taps, then this includes known values for widths `(2 to 768) ++ Seq(1024, 2048, 4096)`.

@aswarterman: I'd have to think about how to support this. I haven't been able to get the desired behavior for that just yet.

An alternative here would be to have an option to NOT specify a seed and then provide a method that will kick the LFSR out of its bad state (all zeros for XOR or all ones for XNOR) if you get unlucky or if your substrate defaults the initialization to something bad (e.g., I think FPGAs will default to zero reset?).

Is there a particularly reason why you want partial initialization of state as opposed to uninitialized state?

aswaterman:
Resetting (XNOR) or setting (XOR) only one bit is sufficient to avoid the bad state, but is cheaper than detecting the bad state.

I don’t want to hold up the PR for this. As long as the components (namely the taps list) are in the public API, it’s possible to do this externally.

seldridge:
Makes sense @aswaterman. Should be doable to add.

src/main/scala/chisel3/util/LFSR.scala
```diff
@@ -27,17 +28,19 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   * }}}
   */
 // scalastyle:off magic.number
+@deprecated("LFSR16 is deprecated in favor of the parameterized chisel3.util.random.LFSR", "3.3")
 object LFSR16 {
   /** Generates a 16-bit linear feedback shift register, returning the register contents.
     * @param increment optional control to gate when the LFSR updates.
     */
+  @deprecated("Use chisel3.util.randome.LFSR(16) for a 16-bit LFSR", "3.3")
```
jwright6323:
```suggestion
  @deprecated("Use chisel3.util.random.LFSR(16) for a 16-bit LFSR", "3.3")
```

seldridge: Fixed in the rebase. Thanks!

seldridge:
Fixed up and ready for review.

- Seed is now optional (`Option[BigInt]`) for PRNGs. If an LFSR (which subclasses PRNG) has an optional seed, then the choice of reduction operation (XOR vs. XNOR) will set/reset the LSB to get the LFSR out of a locked up state. h/t @aswaterman 
- Known maximal period taps are changed from `Map[Int, Set[Int]]` to `Map[Int, Seq[Set[Int]]]` to capture the multiple known good taps that may be available. All taps are then added from  http://courses.cse.tamu.edu/walker/csce680/lfsr_table.pdf. This pushes against the JVM method limit, so the initialization is split into two methods.

seldridge:
@edwardcwang: all LFSR tests (`testOnly chiselTests.LFSRSpec chiselTests.util.random.*`) take 2m57s on my laptop generating 136MB of test data (VCDs). If I patch `TesterDriver` to disable VCD generation time drops minimally to 2m39s.

Total testing time (all tests / `test`) takes 13m on CircleCI. If I throw them at a 6-core server they take 4m and generate 1.3GB of data.

PR #1092, opened 2019-05-09
LFSR/PRNG Asynchronous Safety, Use Vec[Bool] to store internal state

seldridge:
Changes the internal state of PRNG to use Vec[Bool] instead of UInt.
This fixes an @aswaterman identified future problem with asynchronous
reset.

A register with an asynchronous reset can only be reset to a literal.
Previously, an LFSR would store state as a UInt. If it was not
parameterized with a seed it should have its least significant bit
reset to something to avoid locking up. It's ideal to not reset the
full UInt (better test coverage, decreased reset fanout). However,
it's difficult to only reset one bit of a UInt. Conversely, it's
trivial to reset one bit of a Vec[Bool]. This also moves PRNG/LFSR
closer to a canonical representation of their internal state, i.e.,
it's natural to think of generalizing Vec[Bool] to arbitrary finite
fields (Vec[A <: Field]) whereas UInt is tightly coupled to GF2.

The specific change for async reset can be seen here:
  - https://github.com/freechipsproject/chisel3/pull/1092/files#diff-b971a8f4b4adb579f0d85c8060d2fc7eL58

Minor updates:

- Updates/fixes to some scaladoc
- Add assertion to period test to make sure LFSR is changing

Signed-off-by: Schuyler Eldridge <schuyler.eldridge@ibm.com>

**Related issue**: 

**Impact**: API modification

**Release Notes**

- Use Vec[Bool] for internal LFSR/PRNG state

### DIFF
--- a/src/main/scala/chisel3/util/random/FibonacciLFSR.scala
+++ b/src/main/scala/chisel3/util/random/FibonacciLFSR.scala
@@ -47,7 +47,7 @@ class FibonacciLFSR(
   step: Int = 1,
   updateSeed: Boolean = false) extends PRNG(width, seed, step, updateSeed) with LFSR {
 
-  def delta(s: UInt): UInt = s(width-2,0) ## taps.map{ case i => s(i - 1) }.reduce(reduction)
+  def delta(s: Seq[Bool]): Seq[Bool] = taps.map{ case i => s(i - 1) }.reduce(reduction) +: s.dropRight(1)
 
 }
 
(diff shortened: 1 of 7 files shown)


seldridge:
Thanks for taking a look @chickm. I'll wait for somebody else to jump in before merging. Yeah, I think UInt was wrong... things are easier and more natural to express with Vec[Bool]. The generated FIRRTL/Verilog is more structural with Vec[Bool], but is subjectively easier to read, too.

aswaterman:
Depending on the implementation of the async reset stuff, the `when (reset.toBool)` construct won't be valid.  If not, we'll have to make `state` a `Seq[Bool]` where the first element is a `RegInit` and the subsequent ones are `Reg`s.  But I think we can cross that bridge when we get there, since `state` is now private.

seldridge:
I'll wait for this to rebase and merge it, then. Thanks for the explanation @aswaterman.

PR #1093, opened 2019-05-10
Fix LFSR regression

aswaterman:
This fixes LFSR16, which after #1088 ignored its `increment` argument and began incrementing unconditionally.

**Related issue**: 

**Release Notes**


seldridge:
Thanks for catching my regression, @aswaterman...

aswaterman:
Yeah, no problem, it was a quick fix once I realized my I$ couldn’t refill
because the replacement way kept changing under its feet :)

PR #1099, opened 2019-05-21
Minor Scaladoc update

ucbjrl:
- add/move package descriptive text
 - fix FixedPoint link (it's chisel3.experimental.FixedPoint)

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/core/package.scala
+++ b/chiselFrontend/src/main/scala/chisel3/core/package.scala
@@ -2,12 +2,12 @@
 
 package chisel3
 
+/**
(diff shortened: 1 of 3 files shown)


chiselFrontend/src/main/scala/chisel3/core/package.scala
```diff
@@ -2,12 +2,12 @@
 
 package chisel3
 
+/**
+ * These definitions exist to deal with those clients that relied on chisel3.core
```
aswaterman: Missing a period at the end of this sentence (I realize it was like this before and you just moved the text :))

PR #1111, opened 2019-06-20
New Website

seldridge:
First draft of a new website and wiki replacement.

To test this, run `sbt docs/makeMicrosite`. Then `jekyll serve` in `docs/target/site/` and navigate to `localhost:4000`.

This has the following major features:

- The main body of the website is pulled from the `README.md`. This can be extended to include other markdown (e.g., `CONTRIBUTORS.md`, etc.)
- Unidoc API is generated automatically and included
- The contents of the wiki are repurposed as a Users Guide "doc"
- Includes "Chisel Tool" favicons

This uses [`sbt-microsites`](https://47deg.github.io/sbt-microsites/) which is a pretty lean approach to website generation. This includes native support for [`tut`](http://tpolecat.github.io/tut/) which allows you to run Scala snippets inside of code blocks. If you want a relatively complex example of what you can do with this, see the [`cats` website](https://typelevel.org/cats/).

Things to investigate (though I don't think these should block this):

- Use `tut` for all code blocks
- Automatic generation of multiple API versions
- Automatic inclusion or linking to FIRRTL documentation
- Heavy refactor of the imported wiki

Notes:

- `tut` doesn't seem to handle structural types... :cry: 

TODO:

- [ ] Where/how to host?
- [ ] Update acknowledgements
- [ ] Bikeshed
- [ ] Add to CI

**Release Notes**

- New website and documentation packaged inside chisel3


edwardcwang:
@seldridge Looks pretty cool. Quick question - is it possible to still keep the wiki around (perhaps in some other form) that allows quick and easy editing, and we then periodically sync it with the website (maybe for releases? - then we can have user guides for each version of Chisel just like they do for Python)

seldridge:
@edwardcwang: Good question.

I think the wiki currently suffers from two opposing problems: staleness and being ahead of the release. Usually, the wiki is updated and then never touched causing staleness. When the wiki is updated, it's based on master (or, in a hyperbolic example, on [unmerged features](https://github.com/freechipsproject/chisel3/wiki/Experimental---Interval-Type)). It seems to me that we wind up with documentation that is then decoupled from any release and I'm worried that that's providing a disservice to the user base.

I definitely agree that we need something like what they do with Python in terms of versioned websites, documentation, etc. There's a very old issue on [`sbt-microsites`](https://github.com/47deg/sbt-microsites/issues/116) related to versioning.

Nonetheless, I think all this is solvable if we move to a more standard git release workflow (release/dev/hotfix branches). We have one version of the website for each release. The main version of the website points at the latest release. We then maintain a CI-built version of the website that points at the development branch. This then effectively replaces the wiki. There's more overhead in terms of editing the website vs. the wiki, but we get the enormous benefit of having the code examples tested via `tut`.

I'm open to ideas, though.

edwardcwang:
As one of the contributors to the wiki, I perceive that since we don't have a whole documentation team dedicated to this, it would be good to have a scheme where the wiki exists in some form (doesn't have to be the current one) that's easy/low-effort to update coupled with some automation that pulls it and then tests it (via `tut` for example). I agree ideally we'd have teams of people maintain separate documentation for each release, though I'm not sure we quite fit that model.

I've been trying to add mentions of version numbers as necessary in my edits (e.g. this feature is 3.2+, etc).

Also nice PR number (`1111`)

ucbjrl:
This is fantastic @seldridge, but I'd prefer to see this in the [chisel-doc](https://github.com/ucb-bar/chisel-doc) repository (which should be renamed to `chisel-web`). I realize that there's a better chance of keeping it in sync with the code if it's in the same repository, but that didn't appear to work so well when the existing web site was a part of the main code repository. Additionally, the web site has to support multiple versions of the code and shouldn't be intimately connected to a single version.

If we do want to have the code directly available to the web site, I'd prefer to see the [chisel3](https://github.com/freechipsproject/chisel3) repository contained as a submodule (or several submodules - one for each major release) of the `chisel-web` repository.

seldridge:
@ucbjrl: By supporting multiple versions, do you think that there should be one version of the website for each version of Chisel or one version of the API documentation for each version of Chisel?

I'm tending to think that we only need the latter (which is what the current state of the world is), as opposed to the former.

ucbjrl:
I vote for one version of the web site, with support for multiple versions (ScalaDoc) of the APIs. In additional places where we do describe aspects of Chisel, we should note those areas where the code has evolved over time (with pointers to the appropriate ScalaDoc). Ideally, the ScalaDoc will become the ground truth for anything involving Chisel code.

ducky64:
@ucbjrl chisel-doc is not open and contains a bunch of legacy docs that aren't used anymore. If this is the website going forward and it integrates well with the main codebase, I think it would make sense for it to be in this repo. Though it would be worth looking at how other projects structure their website code.

@edwardcwang if the website content remains in a similar format like MarkDown, does that address much of the ease-of-use benefits of the wiki? GitHub has pretty decent builtin support for MarkDown editing, and this limits fragmentation of documentation. There might be a bit of a higher psychological barrier for submitting a PR for code as opposed to a wiki edit, though I don't think there are that many people editing the wiki who aren't devs?

seldridge:
Quickly looking at @ducky64's question, it looks like cats, scalaz, zio, dotty, and spark all have documentation in the mainline repo. Dotty maintains a separate `dotty-website` repo that gets it's gh-pages branch pushed to by mainline. The scala docs website (https://docs.scala-lang.org/) is not embedded in scala proper, however.

If you want to check out more examples, top trending Scala repos may be a good place to look: https://github.com/trending/scala?since=monthly

(Interestingly, the top trending repo has basically no documentation... :grimacing::rocket::grimacing:)

If this is any judge, the bandwagon approach is to embed.

edwardcwang:
Part of the ease-of-use of the wiki is the ability to edit quickly (it's not like anyone can edit, you already need to be a dev) and get instant gratification :) I think I would edit the wiki 5-10x less if that were not the case.

ucbjrl:
I would still prefer this to be a separate repository. The web site should refer to all the chisel-related repositories and not belong to just one of them. Additionally, I think teams who include chisel3 as a submodule in their repository (like rocket-chip), will be surprised to find it includes a web site.

I'm fine with setting up a brand new repository for this purpose (forget about chisel-doc).

seldridge:
@ucbjrl: Hmmm... could that indicate that we're going about it wrong with the BigN projects split up? Was there ever a mono-repoing discussion with broken up publishing? The BigN are all pretty inter-dependent and can't really operate standalone. It would be cool if we weren't as dependent on a bunch of `publishLocal`-ing to run tests outside CI.

I think the critical thing is that all API documentation *should* be published, but I think this can be setup per-project and properly linked with `-doc-external-doc` so that they link to each other. Doing full subproject unidoc would need a mono-repo or a doc repo.

If the website is expected to build API documentation for major/minor versions only, then I think we can make a standalone doc repo work. The drawback is the loss of a tight feedback loop of changes breaking documentation. I.e., it seems like an integrated one would keep the iteration cycle short. I wasn't around for the last integrated website, however.

While we need to keep our prime customer happy, Rocket's already grabbing the full unpublished test-suite. I expect they'd be fine with grabbing some additional markdown (that could be used to build a website). Probably per-PR snapshots would be the best solution for them.

ucbjrl:
I think chisel3 (the code), and chisel3 (the web site) are two distinct objects They have different goals, different constraints, and different development pressures. People developing the web site shouldn't require an intimate knowledge of chisel to make positive contributions.

Putting them in the same repository means they share branches, tags, and commits. If we want to correct a typo in a web page after releasing (and tagging) chisel 3.2.1, do we wait for chisel 3.2.2? How do we tag the repository with 3.2.1 plus web fix? How do we ensure that 3.2.1 plus web fix doesn't contain any changes to the chisel 3.2.1 code? Do we need two branches for every release - one for changes to chisel3-code, one for changes to chisel3-web? What if a bug in sbt-microsites, requires us to update to a new version? We'll need to update the chisel3/project/plugins.sbt. Does this entail bumping the version of chisel3? Any project that includes chisel3 as a subproject will similarly have to bump its .../project/plugins.sbt to use that version of chisel even though none of the chisel3-code has changed.

I think we need to define stable APIs that allow us to decouple the various components of the Chisel/FIRRTL ecosystem to allow independent development of their internal implementations. I don't think a mono-repo is the correct approach, although I admit I might be trying to swim upstream on this one - Google supposedly uses a mono-repo for their development; how they successfully manage it is a mystery to me.

seldridge:
@ucbjrl: Perhaps the website focus is distracting. I think I'm coming at this from the website being documentation (or a shim around documentation) whereas perhaps you're thinking of this as a more disjoint entity?

The problem of a typo is an interesting one... What do you think we should do for an egregious Scaladoc typo, e.g., an example block with `clas Foo extendz module` :grimacing: ? Similarly, what should we do about a scaladoc/unidoc bug where we have to bump?

Minor release (or a hotfix + minor release) would seem to make sense to me.

chick:
@seldridge I'm having enough trouble getting jekyll to run I assume I must have missed something. Should jekyll be installed somewhere by default?

seldridge:
@chick: I assume you want to test this out locally? I installed Jekyll as a gem (gem install --user-local jekyll) and then look in ~/.gem/. If you're asking about server/hosting install, it's likely easiest to publish to GitHub pages and it will deal with Jekyll.

seldridge:
Following previous discussion on not keeping this inside the Chisel repository, the content of this PR has been updated and migrated to: https://github.com/freechipsproject/www.chisel-lang.org/pulls.

PR #1126, opened 2019-07-19
Standard code formatting

chick:
We have talked about the need for consistent code formatting for a while now. Consistent formatting
should  make our code easier to read. Automating the formatting ensures that the formatting is consistent no matter who works on the code. This is more important as we have new contributors with different personal styles.

**Resources**:
[https://scalameta.org/scalafmt/](https://scalameta.org/scalafmt/)

**Type of change**: Code formatting

**Impact**: No functional change.
Visually different code

**Development Phase**: Proposal
I would encourage interested parties to play with the settings in the file. I personally like the more
open style but it comes in some cost in total number of lines

If adopted we should make a decision about doing a pass over every file or perhaps at first 
reformatting files as they are changed in other PRs

**Release Notes**
Chisel3 is moving to a standard automated code formatting style

### DIFF
--- a/.scalafmt.config
+++ b/.scalafmt.config
@@ -1,18 +1,164 @@
-style = defaultWithAlign
-
+version = "2.0.0"
+maxColumn = 100
+docstrings = ScalaDoc
+optIn.configStyleArguments = true
+optIn.breaksInsideChains = false
+optIn.breakChainOnFirstMethodDot = true
+optIn.selfAnnotationNewline = true
+optIn.annotationNewlines = true
+optIn.blankLineBeforeDocstring = false
+binPack.unsafeCallSite = false
+binPack.unsafeDefnSite = false
+binPack.parentConstructors = false
+binPack.literalArgumentLists = true
+binPack.literalsMinArgCount = 5
+binPack.literalsInclude = [
+  ".*"
+]
+binPack.literalsExclude = [
+  String
+  "Term.Name"
+]
+continuationIndent.callSite = 2
+continuationIndent.defnSite = 4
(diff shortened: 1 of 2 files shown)


ducky64:
Seems like a good idea. Where'd your .scalafmt.config come from? It seems to have a mix of defaults (so redundant lines?) and custom overrides (like `continuationIndent.defnSite = 2` which I disagree with).

Looking through the example code change, a major discussion point will probably be whether we want the vertical alignment (which is somewhat enabled as default).

There also seem to be a few cases where it does things inconsistently - for example, in a block of match-case statements, some of the cases have newlines and some don't.

chick:
@ducky64 I have now started with the .scalafmt.config fromt the scalafmt configuration webpage. It should make discovery or fiddling around with it a bit easier. The only think I changed was the columns to be 100. There do appear to be some inconsistencies, but I would guess there are similar inconsistencies in most developers hand formatted code

seldridge:
Are there any thoughts on how to enforce this going forward? 

For IntelliJ users, things should "just work" as long as it's configured correctly. As a catch-all, does a pre-commit hook make sense? 

A CI check would work, too, though that should ideally be one of the first things that runs to keep the feedback fast.

.scalafmt.config
```diff
@@ -0,0 +1,164 @@
+version = "2.0.0"
+maxColumn = 100
```
seldridge:
```suggestion
maxColumn = 120
```

chick:
Will create new PR for this later when PR queue is smaller

Issue #1125, opened 2019-07-17
Type after shift in IntelliJ


schoeberl:
Maybe this is just a bug in IntelliJ, but let's see if someone has an idea.

**Type of issue**: bug report

Following code
```Scala
  val a = 1.U << 1
  val b = 3.U
  println(a, b)

  // IntelliJ editor complains on 'a': Type mismatch, expected UInt, actual: Bits
  when(b >= a) {
    io.data := 1.U
  }
```

displays an error in the IntelliJ editor (a is marked in the when expression).  The editor thinks that the shift operation in the first line results in Bits and not a UInt.

However, the code compiles correctly with IntelliJ and also with sbt from the terminal.

What, in the Chisel type system, is confusing the IntelliJ editor here?

**Please tell us about your environment:**
-version 3.1.8
- IntelliJ latest on Mac (just updated)

kammoh:
*UInt* is a subtype of the class *BIts*, where `<<` is defined with a return type of Bits. So IntelliJ is somewhat right about that! As I understand it, what makes this actually work at compile time is the fact that `<<` is that defined as a macro, and with the proper macro expansion the correct `do_<<` implementation with the correct return type is selected (so maybe it's also fair to attribute this problem to the realtime macro expansions in IntelliJ). IIRC the same issue could arise with `>>` and `~`, for UInts,  FixedPoint, and Bool. 

This is probably just a small nuisance (or again IntelliJ could do better with the macro expansions), and can be resolved by adding a (redundant) `asUInt` at the outcome of these operators. 
But as a side-note, I actually do find the type hierarchy in Chisel (and its implementation) a little over-convoluted and deserving of some attention/revisit. At the same time, I can’t really think of an easy fix for this particular issue. 
Type classes are typically used in these situations and UInt (and SInt) already extends the type class `Num[T]`. Maybe we can argue that shift operators are in fact arithmetic operations and can go in `Num[T]` and return `T` (which resolves this issue). But the problem is that the bit extraction operator in Bits (definitely not arithmetic) is implemented using `>>`.

kammoh:
One possible workaround is to add an implicit conversion method in scope (possibly in your project's package object). Note that implicit conversion methods are generally frowned upon, but there shouldn't be much risk involved in using this in the Chisel context.

```scala
import scala.language.implicitConversions
implicit def bitsToUInt(b: Bits): UInt = b.asUInt
```

schoeberl:
[...]

Issue #957, opened 2018-12-08
[RFC] Bits as full Types


schoeberl:
Bits in Chisel are currently second-class citizens. This is different from Chisel 2, where more functions with Bits are available. Therefore, the reduced functionality of Bits creates a compatibility issue with Chisel 2.

Besides that, Bits as type has a merit on its own (see below).

**Type of issue**: feature request

**Impact**: API addition (no impact on existing Chisel 3 code), but improves Chisel 2 compatibility

**Development Phase**: proposal

Not all data types in hardware are signed or unsigned integers. For example, a word in memory may be an integer, an instruction, or a floating point values. Or something else. Representing this word as UInt is not the correct type. The same is true for a register in a processor, the data field in a bus, and more.

For stronger typing, it makes sense to have a complete Bits type. Bits shall be a collection of bits without an arithmetic meaning. Allowed operations are logical functions (&, |, ^, ~), reduction, and comparison (===, =/=). There is no place for addition or < comparison. Bits can be cast to an integer type.

One question is, how bit vectors with different widths are handled. Either automatic expansion as for UInt or having the length as part of the Bits type and disallow operations with different lengths. The later would be a stronger type restriction. If easy to implement, the later is preferable.

When looking at VHDL as an example language, there are three types: std_logic_vectore, which has no arithmetic operations, unsigned, and signed. Bits would be the equivalent of std_logic_vector.

Bits can be lowered to UInt in FIRRTL and then Verilog without losing the additional type safety in Chisel 3.

ducky64:
This was discussed at the dev meeting today, though as @schoeberl was not present, we didn't make much progress or finalize any resolutions.

Concerns mainly revolved around the details, especially what happens when implicit width expansion is needed. Compatibility with current chisel3 code is also a concern, since the `Bits(...)` companion object constructor (perhaps counterintuitively) returns type UInt, and having it return Bits may break code.

azidar:
Make a PR, have a deprecation schedule for the arithmetic operators on Bits.

kammoh:
I'm experimenting on promoting `Bits` as a full concrete type in #1133 along with other clean-ups of the type system. Everything seems to be in place now, while seems not to be breaking the currently deployed code-base (at least the ones I was I able to verify). 
We can possibly add a factory object for the concrete `Bits` type with a new name and deprecate the current `Bits` factory object producing `UInt`.
[...]

PR #1133, opened 2019-07-25
Basic hardware type-hierarchy cleanup

kammoh:
This is an effort to cleanup Chisel's hardware type hierarchy. This PR addresses:
- Wrong return type of polymorphic operators, causing annoyances such as #1125 with negative impact on the user experience.
- Breaking up the implementations in `Bits.scala`
    - moving unsealed classes/traits (`Num` and `Element`) to designated files
    - introducing a private common trait for `Num` types derived from `Bits` called `NumBits`
    - adding DRYness while retaining compatibility

**Notes**
- May require subsequent clean-up and proof-reading of scaladocs and code comments.

- [x] `rocket-chip:master` compiles and passes basic asm tests
- [x] `dsptools:master` all tests pass

**Related issue**:  #1125

**Impact**: API modification
- `asFixedPoint` is now only defined for `Bits` vs throwing a elaboration-time exception for `Bits`. Not a functional change and also `FixedPoint` is still experimental.

**Release Notes**


ucbjrl:
ok to test

kammoh:
Updated the description to reflect the current state. 
Ready for review.

kammoh:
It should be good for review now!

kammoh:
I made a slightly API-incompatible change. Trying `Bits` as concrete. IIRC `Bits` used to be concrete at some point in the past (maybe chisel2?). I was not able to trace why and when it was made abstract, but I think there's a chance that with what we're doing here maybe we can restore the notion of Bits as the "non-arithmetic hardware bit" type. This would be something similar to many traditional HDLs and I think helps with catching errors as well as code readability. In VHDL for example `std_ulogic_vector` is just a bit vector with different use/operations than the sub-types `signed` and `unsigned`.

kammoh:
Comments in the source code

kammoh:
I realized `Bits` has always been abstract form the initial commit by @jackbackrack.
@ducky64 any thoughts on this? You had left the following TODO in `Bits.scala` back in 2015:
```
  // TODO: perhaps make this concrete?
  // Arguments for: self-checking code (can't do arithmetic on bits)
  // Arguments against: generates down to a FIRRTL UInt anyways
```

ducky64:
@kammoh Bits used to be a concrete type in Chisel2 (iirc, main difference was it didn't support arithmetic), but was made abstract for reason I don't fully remember. There has been people interested in making it concrete again (#957 would be the place to discuss), and I believe there's agreement in the dev group that this is a good idea, but it needs someone to lead the effort and produce an implementation.

There are some design decisions that need to be made, such as what is the exact scope of Bits (though a strawman proposal or PR would be a starting point), and how do you deal with Bits returning UInt (perhaps a deprecation schedule).

kammoh:
If it's ok I'm going to break up this PR into smaller chunks.

seldridge:
I just want to note that this adds `+&`/`+%` to `Num` which makes sense to me. I have a situation where I would like to make something parametric in `A <: Data with Num[A]`. However, it was unexpected that `+&` is currently not a member of `A`. Summarily, the DRYness has tangible benefits for writing generators.

kammoh:
> I just want to note that this adds `+&`/`+%` to `Num` 
@seldridge I maintained this approach in #1168, but I guess any user library extending `Num[T]` now needs to implement those methods.

PR #1136, opened 2019-07-28
Make Queue.irrevocable work properly in chisel3

### LINKED ISSUE #1134: Queue.irrevocable
I write the following codes in Chisel2. And it generates verilog fine.
`val io  = new Bundle  {
        val in  = Flipped(Decoupled(UInt(w.W)))
    }
val x = Queue.irrevocable(io.in, 4)
`

But when I try to convert this to chisel3, it reports "must be a Chisel type, not hardware".

I find that the problem is due to [https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/util/Decoupled.scala](url).

`private val genType = (DataMirror.internal.isSynthesizable(gen), chisel3.internal.Builder.currentModule) match {
    case (true, Some(module: MultiIOModule))
        if !module.compileOptions.declaredTypeMustBeUnbound => chiselTypeOf(gen)
    case _ => gen
  }`
If I change "!module.compileOptions.declaredTypeMustBeUnbound => chiselTypeOf(gen)" to “module.compileOptions.declaredTypeMustBeUnbound => chiselTypeOf(gen)”, I can get verilog without error.

edwardcwang:
Close #1134

**Type of change**: bugfix

**Release Notes**: Fix unexpected error from attempting to use Queue.irrevocable in chisel3.

### DIFF
--- a/src/test/scala/chiselTests/QueueSpec.scala
+++ b/src/test/scala/chiselTests/QueueSpec.scala
@@ -269,6 +269,13 @@ class QueueSpec extends ChiselPropSpec {
         }
       }
     }
+  }
 
+  property("Queue.irrevocable should elaborate") {
+    class IrrevocableQueue extends MultiIOModule {
+      val in = Wire(Decoupled(Bool()))
+      val iQueue = Queue.irrevocable(in, 1)
+    }
+    (new chisel3.stage.phases.Elaborate).transform(Seq(chisel3.stage.ChiselGeneratorAnnotation(() => new IrrevocableQueue)))
   }
 }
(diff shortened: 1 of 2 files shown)


edwardcwang:
This is ready for review.

src/test/scala/chiselTests/QueueSpec.scala
```diff
@@ -233,4 +233,35 @@ class QueueSpec extends ChiselPropSpec {
       }
     }
   }
+
+  property("Queue.irrevocable should elaborate") {
```
seldridge: Since this test is really just doing elaboration and there's no reason to check the functionality of an irrevocable `Queue`, how about stripping this down to a minimal test that elaboration works?

Something like the following is how I would do it in the Stage/Phase world:

```scala
  import chisel3.stage.phases.Elaborate
  import chisel3.stage.ChiselGeneratorAnnotation
  import chisel3.experimental.MultiIOModule

  property("Queue.irrevocable should elaborate") {
    class IrrevocableQueue extends MultiIOModule {
      val in = Wire(Decoupled(Bool()))
      val iQueue = Queue.irrevocable(in, 1)
    }
    (new Elaborate).transform(Seq(ChiselGeneratorAnnotation( () => new IrrevocableQueue )))
  }
```

edwardcwang:
@seldridge Fixed after a long delay ;)

PR #1138, opened 2019-07-30
Core deprecation "since" should be "3.2" not "3.3"

seldridge:
I was looking at trying to index things that were deprecated and noticed that the core deprecation messages (introduced by #1078) should be "3.2" and not "3.3". My understanding is that the second argument of `@deprecated` is the version since which things were deprecated, not the version when things will disappear. I think this was your intention @ucbjrl?

**Related issue**: 


ucbjrl:
Good catch @seldridge. I can remember dithering about this, and at the time, it seemed 3.2 would appear before code containing this PR. Since that's no longer true, I heartily approve this change.

PR #1140, opened 2019-07-31
Circular-shift (rotate) operations for UInt

### LINKED ISSUE #1135: Circular shift operator
Circular shift or bit rotation (https://en.wikipedia.org/wiki/Circular_shift) is common bit-wise operation in many applications including cryptography. VHDL offers operators `ror` and `rol` as well as more polished versions as functions `rotate_left` and `rotate_right` for signed and unsigned types. 
While the user can implement it using bit slicing and concatenation, it could become a little tricky and prone to mistakes. It would be nice to have operators/methods for this in Chisel.

A possible implementation for static rotation is:
```scala
  implicit class RotatableUInt(u: UInt) {
    def <<<(x: Int): UInt =
      if (x == 0)
        u
      else if (x > 0) {
        val w = u.getWidth
        u(w - x - 1, 0) ## u(w - 1, w - x)
      }
      else
        this >>> (-x)

    def >>>(x: Int): UInt =
      if (x > 0) {
        val w = u.getWidth
        u(x - 1, 0) ## u(w - 1, x)
      }
      else
        this <<< (-x)
```

Gotchas:
- Verilog and C don't seem to have operators for this operation. Verilog is using `>>>` operator as sign extension shift.

**Development Phase**:  proposal

**Other information**
 

kammoh:
Adds *static* circular shift (rotation) operations `rotateLeft` and `rotateRight` for `UInt`.

**Related issue**: 
Closes #1135

**Release Notes**


ucbjrl:
ok to test

chick (dismissed):
The logic for the rotate left and right seem correct to me. 
I think the tests need to be beefed up a bit because ConstProp is
basically turning them into if(true) tests
For example: `    assertTesterPasses(new RotateRight)`
Is being turned into the following verilog
```
module RotateRight(
  input   clock,
  input   reset
);
  wire  _T_91; // @[UIntOps.scala 118:7]
  assign _T_91 = reset == 1'h0; // @[UIntOps.scala 118:7]
  always @(posedge clock) begin
    `ifndef SYNTHESIS
    `ifdef STOP_COND
      if (`STOP_COND) begin
    `endif
        if (_T_91) begin
          $finish; // @[UIntOps.scala 118:7]
        end
    `ifdef STOP_COND
      end
    `endif
    `endif // SYNTHESIS
  end
endmodule
```
Here's an example of a test that actually creates some meaningful verilog
```scala
/** rotating a 10 bit word left by n should be equivalent to
  * rotating it by 10 - n to the left
  */
class MatchedRotateLeftAndRight extends BasicTester {
  val shiftAmount1 = RegInit(0.U(4.W))
  val shiftAmount2 = RegInit(10.U(4.W))
  shiftAmount1 := shiftAmount1 + 1.U
  shiftAmount2 := shiftAmount2 - 1.U

  val value = RegInit(BigInt("1000000001", 2).U(10.W))

  val out1 = value.rotateLeft(shiftAmount1)
  val out2 = value.rotateRight(shiftAmount2)

  assert(out1 === out2)
  when(shiftAmount1 === 10.U) {
    stop()
  }
}
```
Alternatively you can create a module and make a tester that drives an instance of it like the primop tests
The test

ekiwi-sifive:
We want to get this in, but need to review to ensure that the width semantics are consistent.

ekiwi-sifive:
The widths seem good, i.e., the result width should always be the same as the width of the input.

ekiwi-sifive:
I agree with @chick that tests for dynamic shifts are missing.

nindanaoto:
Hello to all,

Is this PR frozen? I hope these features will be integrated as chisel3's standard libraries.
I'm currently trying to use rocket chip's implementation. I think they will also be happy with this PR.
https://github.com/chipsalliance/rocket-chip/blob/4276f17f989b99e18e0376494587fe00cd09079f/src/main/scala/util/package.scala#L167-L187

edwardcwang:
@nindanaoto Thank you for your interest in Chisel! Would you be interested in helping us with the missing unit test to get this merged?

kammoh:
Really sorry for forgetting about this PR and even missing the ping :(

I think I'd made _some_ progress with the tests and should have them locally. I can clean up and push by tomorrow. I'm also OK if someone else wants to take over from hear.

nindanaoto:
@kammoh 
I'm happy with your quick reply!
Since this PR seems to be nearly able to merge, if you can do the rest of the works, it would be the best. 

@edwardcwang 
If no one will do the job, I'll do this because I'm one of the people who want to use this functionality. But I think I should wait a while for @kammoh's update.

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -855,6 +855,49 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
...
+  /**
+   * Circular shift to the left
+   * @param that number of bits to rotate
+   * @return UInt of same width rotated left n bits
+   */
+  final def rotateLeft(that: Int): UInt =  macro SourceInfoWhiteboxTransform.thatArg
+
+  def do_rotateLeft(n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = width match {
+    case KnownWidth(w) if n >= w => do_rotateLeft(n % w)
+    case _ if (n == 0) => this
```
aswaterman, line 867: I think you need to make this `if (n == 0 || n == 1)`, or else the code for the n=1 case might either break (in presence of zero-width wires) or produce the wrong width of 2, not sure which.

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -855,6 +855,49 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
...
+  /**
+   * Circular shift to the right
+   * @param that number of bits to rotate
+   * @return UInt of same width rotated right n bits
+   */
+  final def rotateRight(that: Int): UInt =  macro SourceInfoWhiteboxTransform.thatArg
+
+  def do_rotateRight(n: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = width match {
+    case KnownWidth(w) if n >= w => do_rotateRight(n % w)
+    case _ if (n <= 0) => do_rotateLeft(-n)
```
aswaterman, line 881: I think you should make this `n <= 1` so that the special n=1 case I mentioned in the previous comment gets invoked.

chiselFrontend/src/main/scala/chisel3/Bits.scala
kammoh: Do you mean `if (n == 0 || w == 0)`?

aswaterman: Sorry,  I meant `n == 0 || w == 0`, yes.

chiselFrontend/src/main/scala/chisel3/Bits.scala
aswaterman: Sorry, `n <= 0 || w <= 1`

sequencer:
I was thinking upstream [Barrel Shifter](https://github.com/sequencer/arithemic/blob/shifter/arithmetic/src/utils/shifter/barrel.scala) to provide a radix-configurable dynamic shifter, maybe this may help as well?

chiselFrontend/src/main/scala/chisel3/Bits.scala
kammoh: Yes a great catch!

kammoh:
> I was thinking upstream [Barrel Shifter](https://github.com/sequencer/arithemic/blob/shifter/arithmetic/src/utils/shifter/barrel.scala) to provide a radix-configurable dynamic shifter, maybe this may help as well?

I think this is also a barrel-shifter, but at bit-level. From your experience with the `Vec` implementation, do have any thoughts on improving this PR? Do you happen to have a test for that implementation? (I wasn't able to find one on the repo)

sequencer:
Yes, this is a barrel implementation.
https://github.com/chipsalliance/chisel3/blob/dc2484acfc52af4a4d4e7aa0506b874399520d35/core/src/main/scala/chisel3/Bits.scala#L649-L652

> do have any thoughts on improving this PR?

The reason I implementation my barrel rotation is because:
1. I need a `Vec` implementation.
2. I'm using using [shiftGranularity]( https://github.com/sequencer/arithemic/blob/514d73ecce493d70bc9ac8811f0d2cc5b8ca8047/arithmetic/src/utils/shifter/barrel.scala#L13-L15) to provide a fine control to shifter PPA.(synthesis tools(DC/Genus) always being stupid here.)

> Do you happen to have a test for that implementation?

That's my bad, I'll increase the priority of this branch here.

kammoh:
I added two simple (working) tests, one based on suggestion from @chick and another based on rotating `"b001"` and comparing against known good values. I'll also try to see if I can come up with better tests.
Also check for known widths when width is 0 or 1 should work. Of course only matters if the width is known at elaboration time.

chiselFrontend/src/main/scala/chisel3/Bits.scala
kammoh: Should be fixed now.

kammoh:
Do you think the tests are good enough for a merge? @aswaterman @ekiwi @chick
Also what are your thought on the `dontTouch` I added to `UIntOps.io`? Other tests (as well as the rotate ops) were also being optimized in the generated Verilog.

aswaterman (approved):
Yeah, I'm fine with it now.

chick (approved):
This looks good to me. Tests seem sufficient

PR #1146, opened 2019-08-06
Added Chisel documentation generation.

john-a-sifive:
Added chisel2rst.py and Sphinx domain.


edwardcwang:
Thank you so much for your contribution to Chisel @john-a-sifive ! At a high level, would it be possible for you to describe how this works (e.g. how to run it, what are the requirements, etc) and how it relates to ScalaDoc (e.g. does this replace ScalaDoc? does it create parallel documentation?)

seldridge:
We discussed this in the dev meeting, but we'll go ahead and close this. It may be reasonable to handle the overarching goal of this work with an in-tree Scala plugin. This work could also be split out into another repository if needed.

PR #1147, opened 2019-08-06
Avoid when(reset) construct in LFSR

aswaterman:
Muxes and resets are only isomorphic with synchronous reset.  Use a reset
instead of a conditional to make this async-reset-safe.

Unfortunately, the cleanest fix for this involves resetting some of the bits to DontCare, which is not currently optimized by Firrtl (https://github.com/freechipsproject/firrtl/issues/1149).  Merging this as-is will therefore cause a performance regression.

### DIFF
--- a/src/main/scala/chisel3/util/random/LFSR.scala
+++ b/src/main/scala/chisel3/util/random/LFSR.scala
@@ -47,19 +47,24 @@ trait LFSR extends PRNG {
     */
   def reduction: LFSRReduce
 
-  seed match {
-    case Some(s) =>
-      reduction match {
+  override protected def resetValue: Vec[Bool] = seed match {
+    case Some(s) => reduction match {
         case XOR  => require(s != 0, "Seed cannot be zero")
         case XNOR => require(s != BigInt(2).pow(width) - 1, "Seed cannot be all ones (max value)")
       }
-    case None =>
-      reduction match {
-        case XOR  => when (reset.toBool) { state(0) := 1.U }
-        case XNOR => when (reset.toBool) { state(0) := 0.U }
-      }
-  }
+      super.resetValue
 
+    case None => {
+      val res = WireDefault(Vec(width, Bool()), DontCare)
+
+      reduction match {
+        case XOR  => res(0) := true.B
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/random/PRNG.scala
```diff
@@ -33,15 +33,19 @@ abstract class PRNG(val width: Int, val seed: Option[BigInt], step: Int = 1, upd
 
   val io: PRNGIO = IO(new PRNGIO(width))
 
+  /** Allow implementations to override the reset value, e.g., if some bits
+    * should be don't-cares. */
```
seldridge: Mega-nit... Scaladoc guidelines on multi-line comments are to add an extra `*/`. Alternatively, I think this is short enough to fit on one line and stay under the 120 char limit:
```suggestion
    * should be don't-cares. 
    */
```

src/main/scala/chisel3/util/random/LFSR.scala
```diff
@@ -47,19 +47,24 @@ trait LFSR extends PRNG {
...
+      super.resetValue
+
+    case None => {
+      val res = WireDefault(Vec(width, Bool()), DontCare)
+
       reduction match {
-        case XOR  => when (reset.toBool) { state(0) := 1.U }
-        case XNOR => when (reset.toBool) { state(0) := 0.U }
+        case XOR  => res(0) := true.B
+        case XNOR => res(0) := false.B
```
seldridge: Use of `true.B`/`false.B` is cleaner than my original `1.U`/`0.U`. Thanks.

aswaterman: Haha, I wasn't intentionally changing that; it's a result of deleting the old code and rewriting it from scratch.

src/main/scala/chisel3/util/random/PRNG.scala
seldridge: Fixed in rebase.

aswaterman:
Thanks for fixing my busted comment.

PR #1150, opened 2019-08-09
Add sourceInfo to Data.connect exception messages

kammoh:
**Related issue**: 

**Release Notes**

Currently if a connect fails in Chisel (e.g. due to type incompatibility) an exception is thrown, but the `sourceInfo` is not printed. This could make troubleshooting of such problems frustrating, especially in larger designs. This PR adds the `@[filename.scala line:col]` to the exception messages thrown by Data.connect.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Data.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Data.scala
@@ -381,14 +381,15 @@ abstract class Data extends HasId with NamedComponent with SourceInfoDoc { // sc
       requireIsHardware(this, "data to be connected")
       requireIsHardware(that, "data to be connected")
       this.topBinding match {
-        case _: ReadOnlyBinding => throwException(s"Cannot reassign to read-only $this")
+        case _: ReadOnlyBinding =>
+          throwException(s"${sourceInfo.makeMessage(identity)} Cannot reassign to read-only $this")
         case _ =>  // fine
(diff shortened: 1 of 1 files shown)


ucbjrl:
ok to test

ducky64:
Would something like stack trace pruning (or better, highlighting / bolding the line that calls into Chisel - though not sure if that's possible) make sense here? I think the idea behind this makes a lot of sense (emphasize the likely problematic line), but making it easier to read through the stack trace might be a more general solution (since you sometimes need the context further up the stack trace).

seldridge:
@ducky64: That sounds cleaner to me. There's a couple of gates on that:
1. Stack trace pruning is currently selectively used, e.g., `Driver.execute` has it, but `Driver.elaborate` does not. 
2. Stack trace pruning is implemented: as catching an exception, `println`'ing a trimmed stack trace, swallowing the exception, and returning something (pre-Stage/Phase this was `None`, post-Stage/Phase this is the original `AnnotationSeq`). This works pre-Stage/Phase where elaboration is a top-level method, but post-Stage/Phase, the elaboration happens in an elaboration Phase which is not at the top level of the call stack. This is precipitated as behavior like https://github.com/freechipsproject/chisel3/issues/1159.

I have some proposals for how to handle (2) with a `StageError`, but there may be a better approach.

PR #1153, opened 2019-08-12
Added length to Printable

azidar:
Added a length value to Printable to enable better aligning of prints.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Printable.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Printable.scala
@@ -56,6 +56,7 @@ sealed abstract class Printable {
   final def +(that: Printable): Printables = Printables(List(this, that))
   /** Allow for appending Strings to Printables */
   final def +(that: String): Printables = Printables(List(this, PString(that)))
+  def length: Int
 }
 object Printable {
   /** Pack standard printf fmt, args* style into Printable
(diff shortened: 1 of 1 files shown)


chiselFrontend/src/main/scala/chisel3/Printable.scala
```diff
@@ -157,22 +161,40 @@ object FirrtlFormat {
...
+}
 /** Put innermost name (eg. field of bundle) */
 case class Name(data: Data) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) = (data.ref.name, List.empty)
+  def length = data.ref.name.length
 }
 /** Put full name within parent namespace (eg. bundleName.field) */
 case class FullName(data: Data) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) = (data.ref.fullName(ctx), List.empty)
+  def length = sys.error("Cannot calculate langth of FullName")
```
chick: Typo in length (langth)

chiselFrontend/src/main/scala/chisel3/Printable.scala
```diff
@@ -157,22 +161,40 @@ object FirrtlFormat {
   }
 }
 /** Format bits as Decimal */
-case class Decimal(bits: Bits) extends FirrtlFormat('d')
+case class Decimal(bits: Bits) extends FirrtlFormat('d') {
+  override def length: Int = bits.widthOption match {
+    case Some(w) => Math.ceil(Math.log10(Math.pow(2, w) + 1)).toInt
+    case None => 0
```
aswaterman: Should the `None` case be an error, like it is for `FullName`?  Or alternatively should `length`'s type be `Option[Int]`?

azidar: Perhaps. The reason I made it zero was because it seems like an estimate of the length (e.g. all the other widths in an aggregate) would potentially be preferable to a None. I don't have a strong preference, though.

aswaterman: I see.  If it's only an estimate, I don't have any objections, but it might be worth indicating that fact in the scaladoc for `length`.

PR #1154, opened 2019-08-12
ExtModule's lacked support built in support for providing

chick:
**Inconsistent API between ExtModule and BlackBox**
ExtModule's lacked support built in support for providing the verilog source. This changes creates traits that can be used with ExtModule to provide the support currently found in BlackBox

- Add support for ExtModule helpers
  - HasExtModuleResource to use addResource
  - HasExtModuleInline to use setInline
  - HasExtModulePath to use addPath
- Add tests of the above support.
  - Note: These tests use Stage instead of Driver
- Added ScalaDoc for HasBlackBoxInline#setInline

**Type of change**: Make ExtModule consistent with BlackBox

**Release Notes**
ExtModule now has support for 3 methods of providing the underlying verilog implementations: in-line string, scala/java resource (best for use in libraries), and via a file path name.


chick:
@seldridge Thanks, I like the idea of trailing commas, but I've got to find some mechanism for warning me when I commit with them. BTW, I haven't seen the suggested change in a code comment before. Is that some new feature?

src/test/scala/chiselTests/ExtModuleImpl.scala
```diff
@@ -0,0 +1,141 @@
...
+  "ExtModule can have verilator source implementation" - {
+
+    "Implementations can be contained in-line" in {
+      val targetDir = "test_run_dir/extmodule-inline"
+
+      val annotations = Seq(
+        TargetDirAnnotation(targetDir),
+        ChiselGeneratorAnnotation(() => new UsesExtModuleAddViaInline)
+      )
+      val newAnnotations = (new ChiselStage).run(annotations)
```
seldridge, line 105: It would be better to use the `transform` or `execute` methods instead of `run`. The latter is what the user is supposed to implement, while the former two are running all the pre- and post-run steps. 

Ditto for other usages below.

seldridge (approved):
Thanks for adding this missing API, @chick!

One comment, but feel free to merge when ready.

src/test/scala/chiselTests/ExtModuleImpl.scala
chick: Changed to use `.transform`

jackkoenig:
We forgot to backport this which is causing the tests in https://github.com/freechipsproject/chisel3/pull/1650 to fail on 3.4.x

jackkoenig:
@Mergifyio backport 3.4.x

PR #1162, opened 2019-08-20
Move dontTouch, RawModule, and MultiIOModule out of experimental

ucbjrl:

**Impact**: API modification

**Release Notes**
The object `dontTouch` and the classes `RawModule` and `MultiIOModule` are promoted from experimental, implying their respective APIs are stable.

**NOTE**: There will be corresponding pull requests for the rest of the BIG10 that are impacted by these changes. This PR is required for the 3.2.0-RC1 release.


ducky64:
What's up with LegacyModule being in experimental? Did we ever expose that as an API - I thought it was more an internal thing to support old-style Module?

I also think there's an argument for deciding whether we want to make Module be able to take multiple top-level IOs ("merge" MultiIOModule into Module) in this change - since this determines whether we add MultiIOModule (as a separate construct) in the non-experimental API.

ucbjrl:
LegacyModule moved from core to experimental with pr #1078:
 * Move almost all chisel3.core definitions to chisel3 or chisel3.experimental
Should I move it to internal (chiselFrontend/src/main/scala/chisel3/internal)?

It sounds like we're not quite ready for the 3.2 release.

PR #1168, opened 2019-08-31
Bits.scala cleanup part 2

### LINKED ISSUE #1125: Type after shift in IntelliJ

kammoh:
**Related issue**: #1133

**Impact**: no functional change | API modification

**Release Notes**

- Refactor common methods of `UInt`, `SInt`, and `FixedPoint` to `NumBits[T]` trait (type class pattern)
- Resolves  #1125 by overwriting `Bits` methods in derived types. Having a DRY implementation requires removal of shift and negation operators from `Bits` abstract type which breaks compatibility.
- While ost user RTL should be fine but could break code extending `Num[T]` (libraries, `dsptools`?) as it introduces expanding/non-expanding addition and subtraction operations which need to be overwritten, but could help to enforce better consistency regarding the use of default (non-expanding) `+` and `-` operators.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Bits.scala
@@ -8,8 +8,7 @@ import chisel3.experimental.FixedPoint
 import chisel3.internal._
 import chisel3.internal.Builder.pushOp
 import chisel3.internal.firrtl._
-import chisel3.internal.sourceinfo.{SourceInfo, SourceInfoTransform, SourceInfoWhiteboxTransform,
-  UIntTransform}
+import chisel3.internal.sourceinfo.{SourceInfo, SourceInfoTransform, SourceInfoWhiteboxTransform, UIntTransform}
 import chisel3.internal.firrtl.PrimOp._
 
 // scalastyle:off method.name line.size.limit file.size.limit
@@ -55,7 +54,7 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
   // Only used for in a few cases, hopefully to be removed
   private[chisel3] def cloneTypeWidth(width: Width): this.type
 
-  def cloneType: this.type = cloneTypeWidth(width)
+  override def cloneType: this.type = cloneTypeWidth(width)
 
   /** Tail operator
     *
@@ -135,7 +134,7 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
 
   /** @group SourceInfoTransformMacro */
   final def do_apply(x: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = {
-    val theBits = this >> x
+    val theBits = this.asUInt >> x
     theBits(0)
   }
 
@@ -209,31 +208,14 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
     pushOp(DefPrim(sourceInfo, Bool(), op, this.ref))
   }
 
-  /** Pad operator
+  /** Static left shift operator
     *
-    * @param that the width to pad to
-    * @return this @coll zero padded up to width `that`. If `that` is less than the width of the original component,
-    * this method returns the original component.
-    * @note For [[SInt]]s only, this will do sign extension.
+    * @param that an amount to shift by
+    * @return this $coll with `that` many zeros concatenated to its least significant end
+    * $sumWidthInt
     * @group Bitwise
     */
-  final def pad(that: Int): this.type = macro SourceInfoTransform.thatArg
-
-  /** @group SourceInfoTransformMacro */
-  def do_pad(that: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): this.type = this.width match {
-    case KnownWidth(w) if w >= that => this
-    case _ => binop(sourceInfo, cloneTypeWidth(this.width max Width(that)), PadOp, that)
-  }
-
-  /** Bitwise inversion operator
-    *
-    * @return this $coll with each bit inverted
-    * @group Bitwise
-    */
-  final def unary_~ (): Bits = macro SourceInfoWhiteboxTransform.noArg
-
-  /** @group SourceInfoTransformMacro */
-  def do_unary_~ (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bits
+  def << (that: BigInt): Bits = macro SourceInfoWhiteboxTransform.thatArg
 
   /** Static left shift operator
     *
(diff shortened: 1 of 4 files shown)


kammoh:
> Can one of the admins verify this patch?

ping?

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -55,7 +54,7 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
   // Only used for in a few cases, hopefully to be removed
   private[chisel3] def cloneTypeWidth(width: Width): this.type
 
-  def cloneType: this.type = cloneTypeWidth(width)
+  override def cloneType: this.type = cloneTypeWidth(width)
```
seldridge: I like the use of `override` when implementing abstract members. :+1:

chiselFrontend/src/main/scala/chisel3/NumBits.scala
```diff
@@ -0,0 +1,140 @@
...
+  final def unary_-% (): T = macro SourceInfoTransform.noArg
+
+  override def do_< (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, LessOp, that)
+  override def do_> (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, GreaterOp, that)
+  override def do_<= (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, LessEqOp, that)
+  override def do_>= (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, GreaterEqOp, that)
```
seldridge: Move these operators into `Num`.

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -408,30 +380,67 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
...
-  override def do_+ (that: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = this +% that
-  override def do_- (that: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = this -% that
+  /** Static left shift operator
+   *
+   * @param that an amount to shift by
+   * @return this $coll with `that` many zeros concatenated to its least significant end
+   * $sumWidthInt
+   * @group Bitwise
+   */
+  override def << (that: BigInt): UInt = macro SourceInfoWhiteboxTransform.thatArg
```
seldridge: Can this be converted to a `SourceInfoTransform` blackbox macro now since we don't need to refine the return type? 

Whereas before this was defined in `Bits`, this is now implementing a typeclass.

Same comment for all the other methods implementing `NumBits`.

chiselFrontend/src/main/scala/chisel3/Num.scala
```diff
@@ -28,10 +28,35 @@ import chisel3.internal.sourceinfo.{SourceInfo, SourceInfoTransform}
...
+  final def +% (that: T): T = macro SourceInfoTransform.thatArg
+
+  /** @group SourceInfoTransformMacro */
+  def do_+& (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T
+  /** @group SourceInfoTransformMacro */
+  def do_+% (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T
+  /** @group SourceInfoTransformMacro */
+  def do_-& (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T
+  /** @group SourceInfoTransformMacro */
+  def do_-% (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T
```
seldridge: These methods were noticeably missing. Thanks for adding them.

ucbjrl:
ok to test

chiselFrontend/src/main/scala/chisel3/Bits.scala
```diff
@@ -135,7 +134,7 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
 
   /** @group SourceInfoTransformMacro */
   final def do_apply(x: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = {
-    val theBits = this >> x
+    val theBits = this.asUInt >> x
```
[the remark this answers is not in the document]
kammoh: Really sorry the late response @seldridge.
With this refactor, shift operators will only be defined for types extending `Bitwise` and *not* for `Bits` (which is kinda weird, but at the same time the only solution I can think of to address #1125)

chiselFrontend/src/main/scala/chisel3/NumBits.scala
kammoh: It won't be possible to move the implementations to `Num` as it does not require the sub-classes to extend `Bits`, and the `compop` is defined in `Bits`. I was also under the impression that `Num` is not supposed to have specific implementation as it's part of the user API. Is there currently consensus on what `Num` actually is?

chiselFrontend/src/main/scala/chisel3/Bits.scala
kammoh: Turns out we can eliminate *all* uses of `SourceInfoWhiteboxTransform` in the codebase. Should we remove it all-together or comment it out?

seldridge: Orthogonally: I would prefer to never comment stuff out (just remove it). I think that `SourceInfoWhiteboxTransform` can then be removed, but I would like to check with @ducky64 and @jackkoenig.

seldridge:
@kammoh: The history on this got weird. Can you rebase this onto `master`? I think the following should work:

```bash
git rebase origin/master
git push --force
```

chiselFrontend/src/main/scala/chisel3/NumBits.scala
seldridge: Unfortunately, there's no good consensus. It's something that we need to figure out... If you have ideas, it would be reasonable to propose something.

chiselFrontend/src/main/scala/chisel3/Bitwise.scala
```diff
@@ -0,0 +1,140 @@
...
+  final def unary_-% (): T = macro SourceInfoTransform.noArg
+
+  override def do_< (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, LessOp, that)
+  override def do_> (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, GreaterOp, that)
+  override def do_<= (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, LessEqOp, that)
+  override def do_>= (that: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool =
+    compop(sourceInfo, GreaterEqOp, that)
```
seldridge: Thanks for your comment about this before. I see the issue with `compop` being defined in `Bits`. I'm trying to reason about this, but I think that's wrong.

It would seem more natural if this PR could be reduced to only add `Bitwise[T <: Data]` (a bitwise analog to `Num[T <: Data]`). However, `pad` requires `binop`...

Let me bring this up at the developer meeting today. I think there may be a prerequisite PR that sorts out where `binop`, etc. should live first before this goes through.

chiselFrontend/src/main/scala/chisel3/Bitwise.scala
```diff
@@ -0,0 +1,140 @@
...
+package chisel3
+
+import scala.language.experimental.macros
+import chisel3.internal.firrtl.PrimOp.{GreaterEqOp, GreaterOp, LessEqOp, LessOp, PadOp}
+import chisel3.internal.firrtl.{KnownWidth, Width}
+import chisel3.internal.sourceinfo.{SourceInfo, SourceInfoTransform}
+
+// scalastyle:off method.name
+
+trait Bitwise[T <: Bits] extends Num[T] {
```
edwardcwang: Could use some Scaladoc here?

sequencer:
Hi @kammoh , any update on this? :p

PR #1175, opened 2019-09-12
Internal BoringUtils.bore Bug Fix

### LINKED ISSUE #1212: Failed to bore submodule signal to current module
**Type of issue**: bug report

```scala
class Constant extends Module {
  val io = IO(new Bundle{})
  val x = Wire(UInt(6.W))
  x := 42.U
}
class Top extends Module {
  val io = IO(new Bundle{})
  val constant = Module(new Constant)
  val y = Wire(UInt(6.W))
  BoringUtils.bore(y, Seq(constant.x))
}
```

```
  chisel3.internal.ChiselException: signalName/pathName should be called after circuit elaboration
    chisel3.internal.throwException$.apply(Error.scala:42)
    chisel3.internal.HasId.instanceName(Builder.scala:130)
    chisel3.internal.HasId.instanceName$(Builder.scala:124)
    chisel3.Data.instanceName(Data.scala:257)
    chisel3.util.experimental.BoringUtils$.genName$lzycompute$1(BoringUtils.scala:184)
    chisel3.util.experimental.BoringUtils$.genName$1(BoringUtils.scala:184)
    chisel3.util.experimental.BoringUtils$.$anonfun$bore$1(BoringUtils.scala:185)
    chisel3.util.experimental.BoringUtils$.$anonfun$bore$1$adapted(BoringUtils.scala:185)
    scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
    scala.collection.immutable.List.foreach(List.scala:392)
    scala.collection.TraversableLike.map(TraversableLike.scala:238)
    scala.collection.TraversableLike.map$(TraversableLike.scala:231)
    scala.collection.immutable.List.map(List.scala:298)
    chisel3.util.experimental.BoringUtils$.bore(BoringUtils.scala:185)
```

Bore success.

    - version: `3.2.0`

seldridge:
This fixes a bug introduced in the original `BoringUtils.bore` implementation. This was using `instanceName` which meant that it could only be used for closed modules. The user-facing impact of this is an exception if you use `BoringUtils.bore` for signals in the module in which you do the bore (because that module is not closed). This adds one small test for this case.

This PR will introduce worse Verilog naming for boring signals using `BoringUtils.bore`.

Fixes #1212.

**Type of change**: bug report | other enhancement

**Release Notes**

- Bugfix for BoringUtils.bore with signals in the same module

### DIFF
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -181,8 +181,14 @@ object BoringUtils {
     * component
     */
   def bore(source: Data, sinks: Seq[Data]): String = {
-    lazy val genName = addSource(source, source.instanceName, true, true)
+    val boringName = try {
+      source.instanceName
+    } catch {
+      case _: Exception => "bore"
+    }
+    val genName = addSource(source, boringName, true, true)
     sinks.map(addSink(_, genName, true, true))
     genName
   }
+
 }
--- a/src/test/scala/chiselTests/BoringUtilsSpec.scala
+++ b/src/test/scala/chiselTests/BoringUtilsSpec.scala
@@ -106,4 +106,21 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
       .getMessage should startWith ("Unable to determine source mapping for sink")
   }
 
+  class InternalBore extends RawModule {
+    val in = IO(Input(Bool()))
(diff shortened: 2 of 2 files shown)


seldridge:
From dev meeting discussion, as this is a weird corner case that globally worsens naming, see about using the internal `isClosed` to check if we can use a "good" name or a "bad" name.

seldridge:
Updated to opt for catching the exception here.

PR #1181, opened 2019-09-14
Updated to sbt 1.3.0

tampler:
compilation fails for sbt 1.2.7
Switching to 1.3.0 solved the problem


azidar:
We should bump all chisel project sbt's to 1.3 as well.

jackkoenig:
To clarify, it's not that chisel3 doesn't work with sbt 1.2.7, but we might as well move stuff to sbt 1.3.0 anyway.

I'm suspecting the failure @tampler ran into is due to sbt grabbing SNAPSHOTs from the web instead of locally which has long been the behavior of ivy dependency resolution in SBT. Since 1.3.0 uses coursier by default, it likely has different behavior.

PR #1183, opened 2019-09-18
Add read-under-write parameter to SyncReadMem

albert-magyar:

**Release Notes**:
An optional new `ruw` (“read-under-write”) parameter has been added to the `SyncReadMem` factory method. It can take one of three values, describing the value that is read when a read access and a write access occur to the same address in the same cycle.
* `SyncReadMem.ReadFirst` specifies that the old value is read out
* `SyncReadMem.WriteFirst` specifies that the new value is read out
* `SyncReadMem.Undefined` leaves the behavior unspecified

These three modes are implemented using the `read-under-write` parameter of FIRRTL memories.


edwardcwang:
Does this explicitly generate logic to handle these cases, or does it assume that the underlying SRAM will do it for the designer?

albert-magyar:
This is just the Chisel half, so everything translates to a FIRRTL memory. Specifying, say, `SyncReadMem.ReadFirst` just causes the emitted `DefMemory` to have its `read-under-write` parameter set to `"old"`.

In current FIRRTL, this parameter is basically ignored when targeting Verilog, and all sequential memories are translated to a microarchitectural pattern that results in write-first behavior in `VerilogMemDelays`. The [companion FIRRTL PR](freechipsproject/firrt#1183) fixes this and actually emits logic that implements the read-first behavior. Of course, this is just for "native" sequential memories that actually get processed by `VerilogMemDelays` -- memories that get replaced with SRAMs rely on the behavior of the underlying SRAM implementation / Verilog model.

edwardcwang:
I see; thanks for clarifying. Perhaps this would be good to document for users so they don't get surprising behavior with ASIC SRAMs.

albert-magyar:
I am going to add some more scaladoc for the parameter I added, but this PR is not meant to go in until after the FIRRTL PR, so I'd still like feedback on things like parameter and value names.

As far as documentation for user expectation for ASIC SRAMs, those SRAMs are exclusively being generated by the `repl-seq-mem` machinery in FIRRTL, which is basically entirely distinct from the companion FIRRTL PR to this PR. I think some more documentation on that process / conf files could very well be useful.

The FIRRTL PR is more-or-less a "bugfix" (though I'm using that label somewhat loosely), since it brings the semantics of the emitted Verilog in line with the spec in what has previously been an unused corner case.

chiselFrontend/src/main/scala/chisel3/Mem.scala
```diff
@@ -142,34 +145,48 @@ sealed class Mem[T <: Data] private (t: T, length: BigInt) extends MemBase(t, le
...
     *
     * @param size number of elements in the memory
     * @param t data type of memory element
     */
   def apply[T <: Data](size: Int, t: T): SyncReadMem[T] = macro MemTransform.apply[T]
+  def apply[T <: Data](size: Int, t: T, ruw: ReadUnderWrite): SyncReadMem[T] = macro MemTransform.apply_ruw[T]
 
   /** @group SourceInfoTransformMacro */
-  def do_apply[T <: Data](size: BigInt, t: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): SyncReadMem[T] = {
+  def do_apply[T <: Data](size: BigInt, t: T, ruw: ReadUnderWrite = Undefined)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): SyncReadMem[T] = {
```
ekiwi: :+1:

PR #1194, opened 2019-09-27
Strip Object and Outer Class from desiredName

### LINKED ISSUE #1166: Default desiredName and Inner Classes / Objects
When playing around with Chisel on the REPL or in an environment that uses the REPL (e.g., `tut` is being used for the website and does this), you get some pretty terrible default module names.

The underlying reason for this is because the REPL is running commands inside classes, so your Chisel modules are actually inner classes. The result is readable names like `readiwiwiwiwFoo`:

```
scala> class Foo extends RawModule
defined class Foo

scala> Driver.emitVerilog(new Foo)
[info] [0.005] Elaborating design...
[info] [0.050] Done elaborating.
Total FIRRTL Compile Time: 850.7 ms
res1: String =
"module readiwiwiwiwFoo(
);
  initial begin end
endmodule
"

```

This problem similarly exists for testing as classes declared inside a scalatest class are also inner classes.

One way to fix this is to change the default `desiredName` for modules to strip both package names and inner class/object mangling via `$`. This aligns with the behavior of `getSimpleName` (though `getSimpleName` has a bug in JDK 8 that prevents it from working on classes inside objects).

Proposed change:
```scala
    def desiredName: String = this.getClass.getName.split("\\.|\\$").last
```

This would potentially result in worse naming for heavy users of inner classes or people defining classes inside objects. E.g., a pattern like the following results in worse naming via mangles:

```scala
trait Factory { def gen: RawModule }
object CNN extends Factory {
  class Net extends RawModule
  def gen = Module(new Net)
}
object MLP extends Factory {
  class Net extends RawModule
  def gen = Module(new Net)
}
class Top extends RawModule {
  val cnn = CNN.gen
  val mlp = MLP.gen
}
```

```verilog
circuit Top :
  module Net :

  module Net_1 :

  module Top :
    inst cnn of Net @[<pastie> 21:19]
    inst mlp of Net_1 @[<pastie> 25:19]
```

On `master`, `Net` would be named `CNNNet` and `Net_1` would be named `MLPNet`. If you run this on the REPL, prepend `readiwiwiwiw`.

**Other information**
 

Module names will include enclosing class/object information. This isn't so bad for standalone use, but on the REPL or in environments that use the REPL, this can result in a universal long prefix.

Modules are named "what they should be". (What is that?)

- 5bccb888fd3bd129b00e09f946bf820c17f7cc7f

Improved module naming.

seldridge:
This implements the fix proposed in #1166. 

In effect, this strips package names, object names, and outer classes from inner classes when computing the `desiredName`. This mirrors the behavior of what I *think* `desiredName` is supposed to be, i.e., behaving like `getSimpleName`.

An effect of this is that naming behaves much better in the REPL.

**Related issue**: Fixes #1166

**Impact**: API modification

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Module.scala
@@ -221,7 +221,7 @@ package experimental {
     /** Desired name of this module. Override this to give this module a custom, perhaps parametric,
       * name.
       */
-    def desiredName: String = this.getClass.getName.split('.').last
+    def desiredName: String = this.getClass.getName.split("\\.|\\$").last
 
     /** Legalized name of this module. */
     final lazy val name = try {
--- a/src/test/scala/chiselTests/DataPrint.scala
+++ b/src/test/scala/chiselTests/DataPrint.scala
@@ -33,14 +33,14 @@ class DataPrintSpec extends ChiselFlatSpec with Matchers {
   }
 
   class BoundDataModule extends MultiIOModule {  // not in the test to avoid anon naming suffixes
-    Wire(UInt()).toString should be("UInt(Wire in DataPrintSpec$BoundDataModule)")
-    Reg(SInt()).toString should be("SInt(Reg in DataPrintSpec$BoundDataModule)")
+    Wire(UInt()).toString should be("UInt(Wire in BoundDataModule)")
+    Reg(SInt()).toString should be("SInt(Reg in BoundDataModule)")
     val io = IO(Output(Bool()))  // needs a name so elaboration doesn't fail
-    io.toString should be("Bool(IO in unelaborated DataPrintSpec$BoundDataModule)")
+    io.toString should be("Bool(IO in unelaborated BoundDataModule)")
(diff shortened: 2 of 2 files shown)


seldridge:
It's an API change only if we count changes to generated FIRRTL (in any form) as an API change. I think that could go either way.

jackkoenig:
I would definitely count changed Verilog module names as API changes since that's a pretty important API

seldridge:
I agree with you about the API change, but I'd argue that counting on Verilog names is wrong due to problems with composition (the name can change depending on the namespace even with `desiredName` overridden). 

We probably need some way of failing to elaborate if a desired name isn't successfully set as I don't think there's a mechanism to guarantee a name.

seldridge:
@jackkoenig: What's the merge policy for this? Would this need to run through additional regressions given the fact that it's changing Verilog name generation?

Also, @jackkoenig: this should enable us to use custom `mdoc` tags on the website to generate FIRRTL and Verilog without resulting in terrible `readiwiwiwiwiwFoo` module names.

PR #1196, opened 2019-09-28
Fix direction of dynamic index in complex Vec

### LINKED ISSUE #1192: Unexpected behavior when composing Flipped and Vec
**Type of issue**: bug report

**Impact**: unknown

```scala
val in = Flipped(Vec(inChannels, Decoupled(Bool()) ))
```
has different behavior than:
```scala
val in = Vec(inChannels, Flipped(Decoupled(Bool())) ))
```
When doing something like this:
```
for(i <- 0 until inChannels) { io.in(i).ready := false.B }
...
io.in(io.chan).ready := true.B
```
The second type definition and second assignment produces an exception during elaboration:
```
[error] chisel3.internal.ChiselException: Connection between sink (Bool(IO in unelaborated MyModule)) and source (Bool(true)) failed @: Sink is unwriteable by current module.
```
Changing to the first type clears the exception.

The two forms should be equivalent.

chisel: 3.2.0-RC2
scala: 2.12.8

jackkoenig:
Dynamically indexing a Vec of Flipped bidirectional Bundles would get
the wrong directions on the elements of the Bundles

Fixes #1192

See https://github.com/freechipsproject/chisel3/issues/1192#issuecomment-536018644 for insight into the issue. Basically, we were copying the direction from the `sample_element` to the `Vec` itself as an optimization, but if you have a Flipped Bundle as the element this could result in double flipping. Instead, we want to maintain a different direction for the Vec itself so that we can differentiate between complex flippedness.

**Related issue**: #1192

**Type of change**: bug fix

**Release Notes**

Fix bug where dynamically indexing a Vec of Flipped bidirectional Bundles would get the wrong directions on the elements of the Bundles

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Aggregate.scala
@@ -156,7 +156,9 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
       child.bind(ChildBinding(this), resolvedDirection)
     }
 
-    direction = sample_element.direction
+    // Since all children are the same, we can just use the sample_element rather than all children
+    // .get is safe because None means mixed directions, we only pass 1 so that's not possible
+    direction = ActualDirection.fromChildren(Set(sample_element.direction), resolvedDirection).get
   }
 
   // Note: the constructor takes a gen() function instead of a Seq to enforce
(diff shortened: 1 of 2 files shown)


jackkoenig:
Since this is trivially workaroundable (put Flipped outside the Vec rather than on the element), this can wait until 3.2.1. If we end up with a 3.2.0-RC3 we might as well include it, but no reason to make one just for this.

PR #1201, opened 2019-10-01
Don't use MuxLookup default for full mapping

seldridge:
If you use `MuxLookup` with a full mapping, i.e., you fully enumerate all possible values for a key, you still wind up with unused hardware being generated in the case of a `Mux` failure case that is never used and which FIRRTL does *not* optimize away.

This PR modifies `MuxLookup` to examine if the user has provided a full mapping and to *not* emit the additional failure case for the initial `Mux`. In effect, the `default` is immediately discarded and the first mapping becomes the default.

This generally improves Verilog generation for users that are using `MuxLookup` like a Verilog case statement.

This would benefit from tests, but there weren't any tests already written. :sob: I can write them (read: I have not fully tested this). I just figured I would drop the PR first.

### Example

For the following Chisel:

```scala
import chisel3._
import chisel3.util.MuxLookup
import chisel3.stage.{ChiselGeneratorAnnotation, ChiselStage}

class Qux extends RawModule {
  val in = IO(Input(Vec(2, UInt(2.W))))
  val sel = IO(Input(UInt(1.W)))
  val out = IO(Output(UInt(2.W)))

  val mapping = Seq( 0.U -> in(0),
                     1.U -> in(1) )

  out := MuxLookup(sel, DontCare, mapping)
}

(new ChiselStage).execute(Array("-X", "verilog"), Seq(ChiselGeneratorAnnotation(() => new Qux)))
```

Without this PR:

```verilog
module Qux(
  input  [1:0] in_0,
  input  [1:0] in_1,
  input        sel,
  output [1:0] out
);
  wire [1:0] _T_1; // @[Mux.scala 68:16]
  wire  _T_3; // @[Mux.scala 68:19]
  assign _T_1 = sel ? in_1 : 2'h0; // @[Mux.scala 68:16]
  assign _T_3 = 1'h0 == sel; // @[Mux.scala 68:19]
  assign out = _T_3 ? in_0 : _T_1; // @[<pastie> 29:7]
endmodule
```

With this PR:

```verilog
module Qux(
  input  [1:0] in_0,
  input  [1:0] in_1,
  input        sel,
  output [1:0] out
);
  assign out = sel ? in_1 : in_0; // @[<console> 23:7]
endmodule
```

**Related issue**: 

**Impact**: API modification (in terms of generating different Verilog)

**Release Notes**

- Improved `MuxLookup` code generation

### DIFF
--- a/src/main/scala/chisel3/util/Mux.scala
+++ b/src/main/scala/chisel3/util/Mux.scala
@@ -63,10 +63,21 @@ object MuxLookup {
     * @return the value found or the default if not
     */
   def apply[S <: UInt, T <: Data] (key: S, default: T, mapping: Seq[(S, T)]): T = {
-    var res = default
-    for ((k, v) <- mapping.reverse)
-      res = Mux(k === key, v, res)
-    res
+    /* If the mapping is defined for all possible values of the key, then don't use the default value */
+    val (defaultx, mappingx) = key.widthOption match {
+      case Some(width) =>
+        val keySetSize = BigInt(1) << width
+        val keyMask = keySetSize - 1
+        val distinctLitKeys = mapping.flatMap(_._1.litOption).map(_ & keyMask).distinct
+        if (distinctLitKeys.size == keySetSize) {
+          (mapping.head._2, mapping.tail)
+        } else {
+          (default, mapping)
+        }
+      case None => (default, mapping)
+    }
+
+    mappingx.foldLeft(defaultx){ case (d, (k, v)) => Mux(k === key, v, d) }
   }
 }
 
--- a/src/test/scala/chiselTests/MuxSpec.scala
+++ b/src/test/scala/chiselTests/MuxSpec.scala
@@ -3,6 +3,7 @@
 package chiselTests
 
 import chisel3._
+import chisel3.util.{MuxLookup, log2Ceil}
 import chisel3.testers.BasicTester
 
 class MuxTester extends BasicTester {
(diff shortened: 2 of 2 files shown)


src/main/scala/chisel3/util/Mux.scala
```diff
@@ -63,10 +63,20 @@ object MuxLookup {
...
     */
   def apply[S <: UInt, T <: Data] (key: S, default: T, mapping: Seq[(S, T)]): T = {
-    var res = default
-    for ((k, v) <- mapping.reverse)
-      res = Mux(k === key, v, res)
-    res
+    /* If the mapping is defined for all possible values of the key, then don't use the default value */
+    val (defaultx, mappingx) = try {
+      val k = mapping.map(_._1)
+      if ((math.pow(2, key.getWidth) == k.size) && (k.distinct.size == k.size)) {
```
jackkoenig: Due to Data equality being defined as referential equality, using `distinct.size` is not actually correct. There can also be non-literals.

As discussed in dev, we should instead check if all of the keys are literals and mutually exclusive (modulo select width*)

*I'm actually not sure that's correct, I believe when you compare larger things to smaller things it will sign/zero extend, so anything greater than the max value for the select can never match? We should test this.

albert-magyar: Let's say your key is 2-bit, and your lookup keys are `0.U`, `1.U`, `2.U` and `4096.U`. If you don't mask the lookup case keys before checking if their lit values are mutually exclusive, you see that there are four values, and conclude that it is an exhaustive case. However, since there is nothing that matches for `key === 3.U`, it is not actually an exhaustive case.

The false "phantom distinct value" is from `4096.U`, which would be correctly flagged as not distinct if the mapping keys' lit values are masked first.

albert-magyar:
@jackkoenig I would review this, but I wrote some of the code, so you should take a look. The literal exhaustiveness check handles the bitwidth corner case oddities, too.

src/main/scala/chisel3/util/Mux.scala
```diff
@@ -63,10 +63,22 @@ object MuxLookup {
...
+      val keyMask = keySetSize - 1
+      val distinctLitKeys = mapping.flatMap(_._1.litOption).map(_ & keyMask).distinct
+      if (distinctLitKeys.size == keySetSize) {
+        (mapping.head._2, mapping.tail)
+      } else {
+        (default, mapping)
+      }
+    } catch {
+      // This catches the case where the width of the key is unknown, i.e., key.getWidth fails
+      case _: Exception => (default, mapping)
```
jackkoenig: Rather than catching an Exception, should we just check if the width of the key is known?

jackkoenig: Otherwise, we should match on the exact Exception so that we don't mask other issues

seldridge: Unfortunately, [`getWidth` uses `throwException`](https://github.com/freechipsproject/chisel3/blob/92d88ffbe1663b4aa917c6c8b66a4fd697282410/chiselFrontend/src/main/scala/chisel3/Data.scala#L510) and that is inexact. I can just use `getOption` to avoid the problem entirely.

seldridge: Updated in the rebase.

PR #1205, opened 2019-10-07
Use masked litOption comparisons to detect exhaustive MuxLookup

albert-magyar:
@seldridge, this shows what I was talking about with the `0.U` vs `4096.U` and the false "full case" detection.

Fortunately, we don't have to deal with sometimes-negative `SInt` mapping keys :)


albert-magyar:
I meant to open a PR on `full-MuxLookup`, but somehow I did some back touchpad gesture and messed that up. Fixed the base now.

PR #1209, opened 2019-10-14
Enable @chiselName on non-module classes

johnsbrew:

**Release Notes**
Enable `@chiselName` on non-module classes

Hardware val declared inside the scope of a standard scala class (ie that do not extend any kind of `Module`) can now benefit from the `@chiselName` macro and get hw-user-friendly names thanks to the existing fancy macro work of  @ducky64 & @jackkoenig =)


edwardcwang:
Does this affect #1046?

johnsbrew:
@edwardcwang Well, I don't know... it does not break the existing tests but I am not sure they are fully exhaustive ?

ucbjrl:
ok to test

johnsbrew:
Any chance to get a review on this one ? @azidar ?

jackkoenig:
@ducky64 could you take a look at this? Looks very useful

coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
```diff
@@ -118,12 +118,14 @@ class NamingTransforms(val c: Context) {
...
     val transformedBody = (new ModuleTransformer(contextVar)).transformTrees(stats)
-
+    // Note: passing "this" to popReturnContext is mandatory for propagation through non-module classes
     q"""
     val $contextVar = $globalNamingStack.pushContext()
     ..$transformedBody
-    $contextVar.namePrefix("")
-    $globalNamingStack.popReturnContext((), $contextVar)
+    if($globalNamingStack.length == 1){
+      $contextVar.namePrefix("")
```
ducky64: This made sense when Module was top-level, since in all cases the first naming prefix you get would correlate directly with something in the Module. In this case, you could have a non-Module class as the top-level, and the first naming prefix would not necessarily correlate with anything in the Module. Is this behavior we want? Or do we want a check somewhere that the naming top-level is the Module?

coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
```diff
@@ -118,12 +118,14 @@ class NamingTransforms(val c: Context) {
   def transformModuleBody(stats: List[c.Tree]): Tree = {
```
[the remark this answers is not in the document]
johnsbrew: Done. I didn't see any comments to be updated in `Namer.scala` though since the case of Module is described as a special behaviour. 
Closing context on a non-module is not an issue.  
The only change to `Namer.scala` I would suggest is to rename `namePrefix` as `closeContext` as it sounds a bit clearer to me.

coreMacros/src/main/scala/chisel3/internal/sourceinfo/NamingAnnotations.scala
johnsbrew: As far as I understand the macro, when top level is not a module what does the transform is harmless and does not reflect in anyway to the emitted names so it is not an issue.

jackkoenig:
It's not that uncommon Chisel style to have non-Module classes for generating hardware. For example, in this repo: https://github.com/freechipsproject/chisel3/blob/410f03b9122978e43db938d7774b451f2b9111d0/src/main/scala/chisel3/util/Counter.scala#L27

johnsbrew:
I can hear your point of view @ducky64, but I agree with @jackkoenig  I definitely think that being able to build hardware generator objects that are neither fully-featured modules nor functions is interesting.  
Let's merge it then ? I do not have the right to do it.

azidar:
@Mergifyio backport 3.2.x

PR #1213, opened 2019-10-21
Deprecate Driver methods in favor of ChiselStage

seldridge:
Deprecate existing `Driver` methods that have exact `ChiselStage` equivalents.

**Type of change**: documentation | other enhancement

**Release Notes**

- Deprecate old Driver methods with ChiselStage equivalents

### DIFF
--- a/src/main/scala/chisel3/Driver.scala
+++ b/src/main/scala/chisel3/Driver.scala
@@ -82,6 +82,7 @@ case class ChiselExecutionSuccess(
   */
 case class ChiselExecutionFailure(message: String) extends ChiselExecutionResult
 
+@deprecated("Please switch to chisel3.stage.ChiselStage. Driver will be removed in 3.4.", "3.2.4")
 object Driver extends BackendCompilationUtilities {
 
   /**
@@ -90,6 +91,7 @@ object Driver extends BackendCompilationUtilities {
     * @param gen A function that creates a Module hierarchy.
     * @return The resulting Chisel IR in the form of a Circuit. (TODO: Should be FIRRTL IR)
     */
+  @deprecated("Use ChiselStage.elaborate or use a ChiselStage class. This will be removed in 3.4.", "3.2.4")
   def elaborate[T <: RawModule](gen: () => T): Circuit = internal.Builder.build(Module(gen()))._1
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/Driver.scala
```diff
@@ -120,6 +121,7 @@ object Driver extends BackendCompilationUtilities {
     * @param gen A function that creates a Module hierarchy.
     * @return A String containing the design in Verilog.
     */
+  @deprecated("Use chisel3.stage.ChiselStage with '-X verilog -E verilog' arguments", "3.2.1")
   def emitVerilog[T <: RawModule](gen: => T): String = {
```
edwardcwang: Can't we keep this around? This is useful for debuggging.

seldridge: I thought about this after I PR'd it. I intended to only deprecate things for which equivalents exist. Let me come up with a solution here.

seldridge: @edwardcwang: I'll file an issue about this and have something in before 3.3. There are a couple of slight complications here that are hard to untangle and I'd prefer to not do it in this PR.

seldridge: To clarify: the issue here is https://github.com/freechipsproject/firrtl/issues/1261. To get Verilog at the Chisel level, you need to have an `EmittedAnnotation`. Currently, FIRRTL has a dedicated phase, `WriteEmitted` that writes this annotation to a file. Hence, to currently get the Verilog you need to rely on examining a deleted annotation. The proposal with freechipsproject/firrtl#1277 is to keep the annotations intact and make side effects (IO operations, here) separate. The `WriteEmitted` is removed and annotations that define how they should be serialized to files are automatically serialized when the stage finishes. 

In that world, an `emitVerilog` method is trivially examining a non-deleted annotation (which is how I think it should work).

seldridge:
Rebased and updated to include helper methods in `ChiselStage`. This relies on examining deleted annotations, but we can migrate to something better once the "how to serialize" ideas are finished and merged.

src/main/scala/chisel3/Driver.scala
seldridge: This is now possible directly with a `ChiselStage`. See:

```scala
val chirrtlString: String = (new ChiselStage).emitChirrtl(new Foo)
val firrtlString: String  = (new ChiselStage).emitFirrtl(new Foo)
val verilogString: String = (new ChiselStage).emitVerilog(new Foo)
```

PR #1216, opened 2019-10-23
Make ChiselStage targets not private

colinschmidt:

**Development Phase**:implementation

**Release Notes**

This enables users to use the nice run method of `ChiselStage` with their own set of phases.


jackkoenig:
We could consider making such things `protected` instead of `private`. I don't see any reason in this case but perhaps it's a better default.

seldridge:
Good point, @jackkoenig. That would likely be an even better approach...

PR #1224, opened 2019-11-02
Improve naming of anonymous/class-in-function Modules

### LINKED ISSUE #1223: Anonymous Modules get bad names
**Type of issue**: bug report

**Impact**: unknown (is this a modification? I'd argue no, but it changes module names)

**Other information**
 

Given some way of constructing anonymous Modules:

```scala
import chisel3._

abstract class CoolModule(x: Int) extends Module {
  val io = IO(new Bundle {
    val in = Input(UInt(8.W))
    val out = Output(UInt(8.W))
  })
  io.out := io.in + x.U
}

  val io = IO(new Bundle {
    val in = Input(UInt(8.W))
    val out = Output(UInt(8.W))
  })
  val inst = Module(new CoolModule(3) {})
  inst.io.in := io.in
  io.out := inst.io.out
}

object Jack extends App {
  chisel3.Driver.execute(args, () => new MyModule)
}
```

The name we get for `CoolModule` is:
```
  module _3 : 
    input clock : Clock
    input reset : Reset
    output io : {flip in : UInt<8>, out : UInt<8>}
    
    node _T = add(io.in, UInt<2>("h03")) @[Test.scala 9:19]
    node _T_1 = tail(_T, 1) @[Test.scala 9:19]
    io.out <= _T_1 @[Test.scala 9:10]
```

It would be ideal if Chisel could name it based on the abstract class

Prior to #1194, you would get:
```
  module MyModuleanon3 : 
    input clock : Clock
    input reset : Reset
    output io : {flip in : UInt<8>, out : UInt<8>}
    
    node _T = add(io.in, UInt<2>("h03")) @[Test.scala 9:19]
    node _T_1 = tail(_T, 1) @[Test.scala 9:19]
    io.out <= _T_1 @[Test.scala 9:10]
```

That's not ideal either, but better than `_3`

Better module names

seldridge:
This provides a fix for #1223 where modules that are either anonymous or defined in function bodies now produce non-purely-numeric names.

This is an amalgamation of a fix proposed by @aswaterman (which actually is necessary to fix classes in modules) and the use of an `_Anon` prefix for anonymous modules.

This includes tests and more documentation for `desiredName`.

**Related issue**: Fixes #1223 

**Type of change**: bug report | documentation

**Impact**: No functional change

**Release Notes**

- Better naming of anonymous modules and modules defined in functions

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Module.scala
@@ -218,10 +218,32 @@ package experimental {
     //
     // Chisel Internals
     //
-    /** Desired name of this module. Override this to give this module a custom, perhaps parametric,
-      * name.
+
+    /** The desired name of this module (which will be used in generated FIRRTL IR or Verilog).
+      *
+      * The name of a module approximates the behavior of the Java Reflection [[`getSimpleName` method
+      * https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getSimpleName--]] with some modifications:
+      *
+      * - Anonymous modules will get an `"_Anon"` tag
+      * - Modules defined in functions will use their class name and not a numeric name
+      *
+      * @note If you want a custom or parametric name, override this method.
       */
-    def desiredName: String = this.getClass.getName.split("\\.|\\$").last
+    def desiredName: String = {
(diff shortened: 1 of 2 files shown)


azidar:
@mergify backport 3.2.x

PR #1225, opened 2019-11-05
Support literals cast to aggregates as async reset reg init values

jackkoenig:
Accomplished by changing the code gen for casting literals to
aggregates. Rather than connecting the literal to a wire that is then
bit selected from, just bit select from the literal which saves the
creation of an intermediate wire and matches FIRRTL's semantics for
legal async reset initial values.

Turned out to be a very clean implementation, most work (as usual) in the testing.

**Related issue**: https://github.com/freechipsproject/firrtl/pull/1201

**Type of change**: bug fix

**Impact**: API addition? (only in that it fixes a bug)

**Release Notes**
Fix bug where literals cast to aggregates could not be used as async reset register initial values.

### DIFF
--- a/src/test/scala/chiselTests/AsyncResetSpec.scala
+++ b/src/test/scala/chiselTests/AsyncResetSpec.scala
@@ -169,4 +169,34 @@ class AsyncResetSpec extends ChiselFlatSpec {
     assertTesterPasses(new AsyncResetQueueTester)
   }
 
+  it should "allow literals cast to Aggregates as reset values" in {
+    class MyBundle extends Bundle {
+      val x = UInt(16.W)
+      val y = UInt(16.W)
+    }
+    assertTesterPasses(new BasicTester {
+      val reg = withReset(reset.asAsyncReset) {
+        RegNext(0xbad0cad0L.U.asTypeOf(new MyBundle), 0xdeadbeefL.U.asTypeOf(new MyBundle))
+      }
+      val (count, done) = Counter(true.B, 4)
+      when (count === 0.U) {
+        chisel3.assert(reg.asUInt === 0xdeadbeefL.U)
+      } .otherwise {
+        chisel3.assert(reg.asUInt === 0xbad0cad0L.U)
+      }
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/AsyncResetSpec.scala
```diff
@@ -169,4 +169,34 @@ class AsyncResetSpec extends ChiselFlatSpec {
...
+      }
+      val (count, done) = Counter(true.B, 4)
+      when (count === 0.U) {
+        chisel3.assert(reg.asUInt === 0xdeadbeefL.U)
+      } .otherwise {
+        chisel3.assert(reg.asUInt === 0xbad0cad0L.U)
+      }
+      when (done) { stop() }
+    })
+    assertTesterPasses(new BasicTester {
```
ducky64: Suggestion: make this its own test case.

johnsbrew:
@jackkoenig regarding your comment on https://github.com/freechipsproject/firrtl/pull/1201 I guess you do not expect to fix the issues raised in https://github.com/freechipsproject/chisel3/pull/1208 test-cases now but rather wait for an in-depth work about proper constant propagation ?
As far as I understand the current PR does not address the idea of 
```scala
      val x = WireInit(123.U + 456.U)
      withReset(reset.asAsyncReset)(RegInit(x))
```
seen as a valid reset value ?

jackkoenig:
@johnsbrew, you are correct that this is not intended to address test cases like `123.U + 456.U` as a reset value. I'll continue discussion of the more general issue on https://github.com/freechipsproject/firrtl/pull/1201.

src/test/scala/chiselTests/AsyncResetSpec.scala
jackkoenig: 👍

jackkoenig:
Failing CircleCI will be fixed by https://github.com/freechipsproject/chisel3/pull/1227

azidar:
@Mergifyio backport 3.2.x

PR #1210, opened 2019-10-16
Interval Data Type Support for Chisel

chick:
All tests pass
- Renamed Range to IntervalRange to avoid name collision with scala Range
- Changed RangeTransform macro to Return an IntervalRange
  - Improved error messages on missing comma or decimal
  - Added notational support for binary point
  - Some formatting cleanup also
- SIntFactory
  - Change to use IntervalRange API
- UIntFactory
  - UInt from range has custom width computation
    - It does not need to deal with lowerbound extending bit requirements
    - Code to handle special case of range"[0,0]" to have a width of 1
- IR.scala
  - Removed Bound and other constraint code that was duplicating firrtl stuff
  - Added new RangeType
  - Added IntervalRange class and object
- RangeSpec
  - modified just a bit to handle notational differences
    - previous range interpolator returned tuple now returns IntervalRange
- Add IntervalType to emitter
- Added IntervalSpec with many tests
- Added ScalaIntervalSimulatorSpec which tests golden model for Interval
- Added ScalaIntervalSimulator which is a golden model for Interval
  - This gold may not have been polished to a high sheen
- Add IntervalLit cases to Converter
- Add Interval PrimOps to IR
  - asInterval, wrap, squz, clip, setp, decp, incp
  - Add IntervalLit class to IR
- Add Interval to MonoConnect
- Add Interval Type to Bits (in experimental package)
  - add conversions to Interval from other types
- Add Interval clone stuff to Data
- Add Literal creation helpers to chisel3 package
  - these may move to experimental if I can figure that out

**Related issue**: 

**Type of change**: bug report | feature request | documentation | other enhancement

**Impact**: no functional change | API addition (no impact on existing code) | API modification

**Development Phase**: proposal |  implementation

**Release Notes**

PR #1226, opened 2019-11-05
Revert "Interval Data Type Support for Chisel (#1210)"

ucbjrl:
This reverts commit 7b93b0f8c48e39cc9730cf9f91340cf733dadafe.


jackkoenig:
It's failing due to the pulled FIRRTL SNAPSHOT: `sbt.librarymanagement.ResolveException: Error downloading edu.berkeley.cs:firrtl_2.12:1.2-20191030-SNAPSHOT`. I'm thinking release branches should build from release branches, so `3.2.x` should probably build FIRRTL from source and use that, alternatively it could depend on `1.2.+`, thoughts?

ucbjrl:
Typically (in the past), I do all the tests by building from source locally (after deleting the edu.berkeley.cs ivy cache). I don't submit the dependency changes as PRs - I merge them directly into the Z.Y.x branch. They don't get CI testing. The dependencies (pretty much have to) reflect post-publishing.

I think reproducible builds are important so I'm hesitant to support wild cards in dependencies.

If we do want to have CI testing on release updates (which would be a good thing), I think it has to use build from source and publishLocal.

jackkoenig:
I think we definitely want CI on release branches because I want to automate bugfix backporting to make bugfix releases as easy to do as possible. I agree with you about reproducibility, so does that mean we should test against FIRRTL `1.2.0` and bump that number each time we release a new FIRRTL?

Per Jenkins, is that configuration in the repo or is it defined elsewhere? Can you fix the chisel3 CI for `3.2.x` or is there some way I can help?

ucbjrl:
It's defined in the Jenkins job, which does allow overriding dependency versions. I can manually push this through with the correct dependencies, but we should use structured comments to define them.
(i.e., Requires firrtl <pr ddd>|<branch abc>)

jackkoenig:
Ideally there's no need for manual intervention at all if we want to automate backporting, can we make Jenkins treat release branches like 3.2.x differently than master?

ucbjrl:
Yes. We can add branch-specific Jenkins job code.

PR #1228, opened 2019-11-05
Adds a link to the Chris Celio style guide

chick:

**Release Notes**

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ These simulation-based verification tools are available for Chisel:
 - [**Wiki**](https://github.com/freechipsproject/chisel3/wiki), which contains various feature-specific tutorials and frequently-asked questions.
 - [**ScalaDoc**](https://www.chisel-lang.org/api/latest/chisel3/index.html), a listing, description, and examples of the functionality exposed by Chisel
 - [**Gitter**](https://gitter.im/freechipsproject/chisel3), where you can ask questions or discuss anything Chisel
+- [**Style Guide**](https://github.com/ccelio/chisel-style-guide), A basic guide to chisel style conventions
 - [**Website**](https://www.chisel-lang.org)
 
 If you are migrating from Chisel2, see [the migration guide](https://www.chisel-lang.org/chisel3/chisel3-vs-chisel2.html).


README.md
```diff
@@ -105,6 +105,7 @@ These simulation-based verification tools are available for Chisel:
 - [**Wiki**](https://github.com/freechipsproject/chisel3/wiki), which contains various feature-specific tutorials and frequently-asked questions.
 - [**ScalaDoc**](https://www.chisel-lang.org/api/latest/chisel3/index.html), a listing, description, and examples of the functionality exposed by Chisel
 - [**Gitter**](https://gitter.im/freechipsproject/chisel3), where you can ask questions or discuss anything Chisel
+- [**Style Guide**](https://github.com/ccelio/chisel-style-guide), A basic guide to chisel style conventions
```
edwardcwang: Is this meant to be an official style guide? Should it be prefixed?

This might be nitpicking; the style guide recommends looking at rocket as an example for beginners to read....

chick: Baby steps.

I think further discussion and work here would be well spent. I wonder if we should fork his repo, but his is the top hit if one search chisel-style guide.

I had planned to variable naming section. That's a simple discussion I'm sure.

ekiwi:
>     * What other style suggestions do people have?

I think a lot of the variable naming comes down to "programmers" vs "hardware designers", i.e., do you name things like people would in Scala or as it is common in Verilog/VHDL?

Afaik Scala code normally does not use `snake_case` for example.

chick:
- This is just a question of whether to link to Chris's document
- His document is top search result on google. 
- There are really not any good alternatives
- Chris is willing to accept PR's that improve it, or we could fork his and maintain our own
- People are asking for style guide on SO and elsewhere.
- I think this is worth merging now while noting that more work should be done

ducky64:
I'd prefer not to advocate a style guide that doesn't yet have consensus across the community, just so we can tick a box saying "we have this". I think releasing (and more importantly, expecting people to follow) a style guide that is going to see significant flux right after is not a great idea, since the whole point of a style guide is stability and consistency.

Like, this is a starting point to discuss on, and we can prioritize that discussion, but I think this is something that needs to be done as right as we can.

Also, wrt this being the top result on Google: the chisel2 cheatsheet also used to be the top result on Google, well into the chisel3 era. If this document truly reflects the currently dominant style (if one even exists) and best practices, then sure, but I'm not sure if that's the case.

ccelio:
> > ```
> > * What other style suggestions do people have?
> > ```
> 
> 
> I think a lot of the variable naming comes down to "programmers" vs "hardware designers", i.e., do you name things like people would in Scala or as it is common in Verilog/VHDL?
> 
> 
> Afaik Scala code normally does not use `snake_case` for example.

Chisel is for describing hardware designs; although the host language style guide can provide some guidance, users shouldn't be afraid to write code in a manner that improves understandability. 

As such, though perhaps it betrays my C/C++ and Verilog origins, I have found camelCase does a poor job of providing visual distinctions between prefixes and suffixes which are commonly required for describing hardware.

In particular, this shows up in hardware's need to describe temporal behaviors within the code: across pipeline stages, the differing update semantics of Reg vs Wire, multi-cycle interface contracts of some I/Os, and more. 

If you are lucky, you can write your RTL in a functional manner that lines up closely with Scala style. But for other modules, a more structural style with explicit pipelining lends itself closer to more C/Verilog-like naming conventions. Skimming through rocket-chip, I've found it changes style as needed depending upon how best to describe the Module in question.

jwright6323:
I've gone back and forth on snake vs camel throughout my Chisel career and see merits to both. To add to @ccelio's `snake_case` pros, having `snake_case` for hardware values and `camelCase` for software values/functions can aid in differentiating between the two. OTOH, one relatively big pro for `camelCase` is that it is much easier to correlate emitted bundle hierarchies with their source scala if you can reliably use `_` as a hierarchy delimiter (e.g. `someReally_longNamed_bundleWith_lotsOf_hierarchy` vs `some_really_long_named_bundle_with_lots_of_hierarchy`).

seldridge:
Dev meeting conclusion: we'd prefer to just point to the Scala style guide. If anyone feels differently we can re-open.

PR #1234, opened 2019-11-07
Recursively suggests names to supported Map

colin4124:

**Release Notes**

Support declare base data types by using `Map[String, Data]` . Suggested Name as key, the `Data` type as value. It allows automatic generation by reading some configure files to get name strings.

```scala
class Mux2 extends RawModule {
  val io = Map(
    "sel" -> IO(Input(UInt(1.W))),
    "in0" -> IO(Input(UInt(1.W))),
    "in1" -> IO(Input(UInt(1.W))),
    "out" -> IO(Output(UInt(1.W))),
  )
  val inner = Map(
    "sel_in0" -> (io("sel" )& io("in1")),
    "sel_in1" -> (~io("sel") & io("in0")),
  )
  io("out" ):= inner("sel_in0") | inner("sel_in1")
}
```
```verilog
module Mux2(
  input   sel,
  input   in0,
  input   in1,
  output  out
);
  wire  sel_in0; // @[Mux.scala 12:29]
  wire  _T; // @[Mux.scala 13:19]
  wire  sel_in1; // @[Mux.scala 13:30]
  assign sel_in0 = sel & in1; // @[Mux.scala 12:29]
  assign _T = ~ sel; // @[Mux.scala 13:19]
  assign sel_in1 = _T & in0; // @[Mux.scala 13:30]
  assign out = sel_in0 | sel_in1; // @[Mux.scala 15:13]
endmodule
```

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/Builder.scala
@@ -371,11 +371,14 @@ private[chisel3] object Builder {
   /** Recursively suggests names to supported "container" classes
     * Arbitrary nestings of supported classes are allowed so long as the
     * innermost element is of type HasId
-    * (Note: Map is Iterable[Tuple2[_,_]] and thus excluded)
     */
   def nameRecursively(prefix: String, nameMe: Any, namer: (HasId, String) => Unit): Unit = nameMe match {
     case (id: HasId) => namer(id, prefix)
     case Some(elt) => nameRecursively(prefix, elt, namer)
+    case m: Map[_, _] =>
+      m foreach { case (k, v) =>
+        nameRecursively(s"${k}", v, namer)
+      }
     case (iter: Iterable[_]) if iter.hasDefiniteSize =>
       for ((elt, i) <- iter.zipWithIndex) {
         nameRecursively(s"${prefix}_${i}", elt, namer)


ucbjrl:
ok to test

jackkoenig:
After a little further thought, I take back that this is "setName-like". It still prefixes the names (as it should) so it's not quite fulfilling that role. You can also achieve this goal with a Record like the `CustomBundle` in the tests (https://github.com/freechipsproject/chisel3/blob/2c449c5d6e23dcbb60e8c64cab6b6f4ba6ae313f/src/test/scala/chiselTests/RecordSpec.scala#L15). Perhaps we should move that to `chisel3.util`(and name it something like `MapBundle`?) so that it can be more easily used for this purpose.

I'm actually against this PR with that in mind because you can't bulk connect to a `Map`, nor we can we clone it so this is strictly less powerful than just using a Record.

colin4124:
What **non-determinism** of `Map` means?

  As in my opinion, non-determinism of `Map` is that its elements is out-of-order.

 It makes sense that `IO` ports declaration is in-order. Take below as example:

```scala
val io = Map(
    "sel" -> IO(Input(UInt(1.W))),
    "in0" -> IO(Input(UInt(1.W))),
    "in1" -> IO(Input(UInt(1.W))),
    "out" -> IO(Output(UInt(1.W))),
  )
```
Before creating a **out-of-order** `Map`, it executes arguments **in-order**.  so that `IO()` would add it to `Module`'s `_ports` in-order. 

But it can't add prefix `io` automatic. it must add prefix "io_sel"  explicitly. In the situation that  programming automatic generation by reading some configure files to get name strings at the same time adding prefix is not a matter.

If using `CustomBundle`, `requireIsChiselType(elt)` does not allow `("sel",  IO(Input(UInt(1.W)))` due to `IO()` is hardware type. It only allow `val io = IO(new CustomBundle(...))`, but it adds a prefix `io_`, what if I don't want prefix?

Why can't bulk connect to a Map? What the difference between `Map` and other collection such as `List`? Connecting between `Data` and `Data`, It does not matter wherever the `Data` could be got directly, or from collection. As long as I could get the `Data` from collection, it doesn't matter I got it by index from in-order one or by key from out-of-order one.

PR #1235, opened 2019-11-08
Fix ExtModule port's suggestedName not work

colin4124:

**Release Notes**

`ExtModule` port's name should take `suggestdName` first as `RawModule` do


ucbjrl:
ok to test

colin4124:
The original `trait NamedModuleTester extends MultiIOModule ` didn't support `ExtModule`, so I change to `trait NamedModuleTesterBase` due to it doesn't depend on `MultiIOModule`. 

Changing `trait NamedModuleTester extends MultiIOModule `  to `trait NamedModuleTester extends MultiIOModule with  NamedModuleTesterBase`  does not affect other code.

src/test/scala/chiselTests/experimental/ProgrammaticPortsSpec.scala
```diff
@@ -39,6 +40,23 @@ class ProgrammaticPortsSpec extends ChiselFlatSpec {
...
+        expectName(clock.suggestName("clk"), "clk")
+        expectName(reset.suggestName("rst"), "rst")
+        expectName(wdata.suggestName("foo"), "foo")
+        expectName(rdata.suggestName("bar"), "bar")
+      })
+    }
+    elaborate { module }
+    assert(module.ext_module.getNameFailures() == Nil)
+  }
+
```
jackkoenig: This is kind of in conflict with the `LegacyModule` test below:
```scala

  "LegacyModule" should "ignore suggestName on ports" in {
    doTest(new Module with NamedModuleTester {
      val io = IO(new Bundle {
        val foo = Output(UInt(8.W))
      })
      expectName(io.suggestName("cheese"), "io")
      expectName(clock.suggestName("tart"), "clock")
      expectName(reset.suggestName("teser"), "reset")
    })
  }
```
I'm not really sure what we should do here, but ports have been something we've been a bit loathe to allow arbitrary naming of. I'm not saying we shouldn't do it though, just that we have avoided it in the past. The only `suggestName` on ports that's currently supported I think is when they aren't a public val in the class (see the "PrivatePort" test above).

jackkoenig:
Also thank you for writing the test, that was a pretty clever way to do it :)

azidar:
We think this might have been fixed, needs to be tested.

src/test/scala/chiselTests/experimental/ProgrammaticPortsSpec.scala
colin4124: Sorry I reply so late, I think that **BlackBox** act as **LegacyModule** which restricted by `val io`, **ExtModule** act as **RawModule** which would be flexible. So **ExtModule** can rename IO doesn't conflict with **LegacyModule**.

colin4124:
@azidar Do you want this to be fixed? If so, I can renew this PR for the latest version of the code base.

PR #1244, opened 2019-11-22
Create .mergify.yml

azidar:

**Release Notes**
Adds mergify to facilitate backporting and merging PRs.


.mergify.yml
```diff
@@ -0,0 +1,72 @@
...
+      backport:
+        branches:
+          - 3.2.x
+      label:
+        add: [Backported]
+
+  - name: label Mergify backport PR
+    conditions:
+      - base=3.2.x
+      - body~=This is an automated backport of pull request \#\d+ done by Mergify.io
```
seldridge:
```suggestion
      - body~=This is an automated backport of pull request \#\d+ performed in the hall of the robot king
```

jackkoenig: 😂

.mergify.yml
```diff
@@ -0,0 +1,72 @@
+pull_request_rules:
+  - name: remove outdated reviews
+    conditions:
+      - base=master
+    actions:
+      dismiss_reviews:
```
jackkoenig: I see we also have this on firrtl, but what does this rule do? A naive reading would be that it always just dismisses reviews on master but it clearly doesn't do that...

azidar: Looks like it enables mergify to take the last approved review; a stale review isn't considered if there is another approving one. https://doc.mergify.io/examples.html?highlight=dismiss#id11

PR #1252, opened 2019-11-27
Fix bidirectional Wire with Analog

jackkoenig:
For direction purposes, Analogs were treated differently in IOs vs. in Wires. This difference was added in https://github.com/freechipsproject/chisel3/pull/617, specifically commit https://github.com/freechipsproject/chisel3/commit/f4a1c40fa147924087065bf7024038cc465b95ae, but I am unable to ascertain why. @ducky64 as a reviewer in case he has insight.

The motivating reason for fixing this is that `BundleBridge` in rocket-chip materializes the tunneled Bundle via a Wire in the destination module. Currently, you cannot have BundleBridges with any mix of Analog and non-Analog components. This fixes that issue.

**Related issue**: 

**Type of change**: bug fix

**Release Notes**
Support bidirectional Aggregate Wires with Analog

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/experimental/Analog.scala
+++ b/chiselFrontend/src/main/scala/chisel3/experimental/Analog.scala
@@ -57,10 +57,8 @@ final class Analog private (private[chisel3] val width: Width) extends Element {
       case SampleElementBinding(parent) => parent.topBinding
     }
 
-    // Analog counts as different directions based on binding context
     targetTopBinding match {
-      case WireBinding(_) => direction = ActualDirection.Unspecified  // internal wire
-      case PortBinding(_) => direction = ActualDirection.Bidirectional(ActualDirection.Default)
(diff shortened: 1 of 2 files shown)


ducky64:
I don't specifically remember the rationale for that in #617, I'd guess it might have something to do with that internal wires shouldn't be directioned? That being said, in #617 it doesn't appear that Bundles have similar behavior, and at a glance I think keeping it consistent with the other data types makes sense?

jackkoenig:
> I don't specifically remember the rationale for that in #617, I'd guess it might have something to do with that internal wires shouldn't be directioned? That being said, in #617 it doesn't appear that Bundles have similar behavior, and at a glance I think keeping it consistent with the other data types makes sense?

Yeah I agree. I'm not sure what the rationale was either but I think making the behavior consistent makes the most sense.

jackkoenig:
@Mergifyio backport 3.2.x

PR #1274, opened 2019-12-11
 Bug fixes to support code for Interval

chick:
- Change getPossibleValues of Interval to return a NumericRange formerly Seq materialized all values
- Fixed computation in getHighestPossibleValue, erroneously was using lower intead of upper

**Type of change**: bug fix

**Release Notes**
Fixed bug that miscalculated the highest possible value of an IntervalRange

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/chiselFrontend/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -9,6 +9,7 @@ import chisel3.experimental._
 import _root_.firrtl.{ir => firrtlir}
 import _root_.firrtl.PrimOps
 
+import scala.collection.immutable.NumericRange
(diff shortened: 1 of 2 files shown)


seldridge:
Can this be targeted at 3.2.x?

chick:
Good question. It's fairly obscure and a problem that would mostly affect testing of code using interval. But it's a pretty simple fix

chick:
Depends on [Chisel PR #1274](https://github.com/freechipsproject/chisel3/pull/1274)

ucbjrl:
I don't think we have intervals support in `3.2.x`. It was purposefully removed and postponed to `3.3.0`. I'm not sure if that has any impact on backporting this fix.

seldridge:
@ucbjrl: Ah, got it. No intervals support means this would backport terribly to 3.2.x. This is targeted at the right milestone, 3.3.0.

PR #1275, opened 2019-12-12
Fixed problem creating Interval literals with full ranges

chick:
**Type of change**: bug fix

- boundary testing was not taking binary point into account correctly
- add tests to show where things work and where they are supposed to fail

**Release Notes**
A bug that affected could cause exceptions when creating intervals with range specifiers was fixed, e.g. `val iLit = 6.0.I(range"[0,6]")` would fail even though 6 is legal in that range.


jackkoenig:
Since this is a bugfix, should it be backported to 3.2.x?

ucbjrl:
If it needs (or is based on) intervals, it shouldn't be backported. Otherwise, I'd say yes, but let's get @chick's input.

jackkoenig:
Oh right, intervals are not in 3.2, nevermind!

PR #1279, opened 2019-12-17
Run chisel test suite with Treadle

chick:
This PR introduces a dependency on Treadle for backend testing.
Looking for comments on implementation and concept.

**Type of change**: Testing enhancement

**Impact**: no functional change to chisel
This could enable running unit tests to go much faster, probably about 10x.
This also helps make sure Treadle has a high fidelity to verilator.

**Release Notes**
Chisel test suite tests can now be run using Treadle  (manually).
This PR presents a use case for the Issue on [global settings](https://github.com/ucb-bar/chisel-testers2/issues/101)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ test_run_dir
 *~
 \#*\#
 .\#*
+TestBackendName
--- a/build.sbt
+++ b/build.sbt
@@ -192,3 +192,33 @@ lazy val chisel = (project in file(".")).
 addCommandAlias("com", "all compile")
 addCommandAlias("lint", "; compile:scalafix --check ; test:scalafix --check")
 addCommandAlias("fix", "all compile:scalafix test:scalafix")
+
+// Tasks for running tests with different backends
+//
+//
+
+lazy val setTreadleBackend = taskKey[Int]("Sets the test harnes backend to treadle")
+lazy val setVerilatorBackend = taskKey[Int]("Sets the test harness backend to verilator")
+lazy val setDefaultBackend = taskKey[Int]("Sets the test harness backend to default")
+
+import java.io.PrintWriter
+
+setTreadleBackend := {
+  val p = new PrintWriter(new File("TestBackendName"))
+  p.println(s"treadle")
+  p.close()
+  0
+}
+
+setVerilatorBackend := {
+  val p = new PrintWriter(new File("TestBackendName"))
+  p.println(s"verilator")
+  p.close()
+  0
+}
+
+setDefaultBackend := {
+  val f = new File("TestBackendName")
+  if (f.exists()) { f.delete() }
+  0
+}
--- a/build.sbt
+++ b/build.sbt
@@ -28,7 +28,10 @@ def javacOptionsVersion(scalaVersion: String): Seq[String] = {
   }
 }
 
-val defaultVersions = Map("firrtl" -> "1.4-SNAPSHOT")
+val defaultVersions = Map(
+  "firrtl" -> "1.4-SNAPSHOT",
+  "treadle" -> "1.3-SNAPSHOT"
+)
 
 lazy val commonSettings = Seq (
   resolvers ++= Seq(
@@ -50,7 +53,7 @@ lazy val commonSettings = Seq (
   //  this has to be a Task setting.
   //  Fortunately, allDependencies is a Task Setting, so we can modify that.
   allDependencies := {
-    allDependencies.value ++ Seq("firrtl").collect {
+    allDependencies.value ++ Seq("firrtl", "treadle").collect {
       // If we have an unmanaged jar file on the classpath, assume we're to use that,
       case dep: String if !(unmanagedClasspath in Compile).value.toString.contains(s"$dep.jar") =>
         //  otherwise let sbt fetch the appropriate version.
--- a/src/main/scala/chisel3/testers/TesterDriver.scala
+++ b/src/main/scala/chisel3/testers/TesterDriver.scala
@@ -2,23 +2,69 @@
 
 package chisel3.testers
 
-import chisel3._
 import java.io._
 
-import chisel3.aop.Aspect
+import chisel3._
 import chisel3.experimental.RunFirrtlTransform
 import chisel3.stage.phases.AspectPhase
-import chisel3.stage.{ChiselCircuitAnnotation, ChiselStage, DesignAnnotation}
-import firrtl.{Driver => _, _}
+import chisel3.stage.{ChiselCircuitAnnotation, ChiselStage, DesignAnnotation, NoRunFirrtlCompilerAnnotation}
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.options.Unserializable
 import firrtl.transforms.BlackBoxSourceHelper.writeResourceToDirectory
+import firrtl.{Driver => _, _}
+import treadle.stage.TreadleTesterPhase
+import treadle.{CallResetAtStartupAnnotation, TreadleTesterAnnotation, WriteVcdAnnotation}
 
+//scalastyle:off magic.number method.length
 object TesterDriver extends BackendCompilationUtilities {
+  var MaxTreadleCycles = 10000L
+
+  trait Backend extends NoTargetAnnotation with Unserializable
+  case object VerilatorBackend extends Backend
+  case object TreadleBackend extends Backend
+  case object NoBackend extends Backend
+
+//  val defaultBackends: Seq[Backend] = Seq(TreadleBackend)
+  val defaultBackends: Seq[Backend] = Seq(VerilatorBackend)
+//  val defaultBackends: Seq[Backend] = Seq(TreadleBackend, VerilatorBackend)
+
+  /** Use this to force a test to be run only with backends that are not Treadle
+    */
+  def verilatorOnly: AnnotationSeq = {
+    defaultBackends.map {
+      case TreadleBackend => NoBackend
+      case other => other
+    }
+  }
 
   /** For use with modules that should successfully be elaborated by the
     * frontend, and which can be turned into executables with assertions. */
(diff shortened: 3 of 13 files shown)


jackkoenig:
I'd love to use treadle for testing, I think it'll vastly speed up how long the Chisel CI tests take. One ask I have though is that it's relatively easy to make it use Verilator-only. Just for sanity checking we should probably run all the tests using Verilator as part of a nightly regression or something. For normal CI though, I'd love to switch to treadle so that they run faster. on FIRRTL CI, the firrtl tests take 5-7 min while the Chisel tests take 28 minutes.

azidar:
Nightly run, do verilator. For general CI tests, run treadle. Thus, we should make it easy to switch backends.

jackkoenig:
I think this needs rebasing, I doubt you're trying to PR 109 commits

ducky64:
In general, I think that using a persistent-state mechanism (such as the filesystem) basically ephemeral communication is kind of a red flag:
- Most importantly: because state is persistent, you have to be really thorough about initializing and clearing state, or you might have dependencies between runs causing non-intuitive behavior.
- Changes to the filesystem may also be detected by other tools - for example, how would this interact with git? Should we also be concerned with the edge case of files being modified during test runs (admittedly a somewhat unlikely case, and assuming the file is an intermediate only)?
- The mechanism isn't immediately obvious (you basically would have to locate the particular piece of documentation), especially if you were trying to debug behavior. Potentially could be mitigated with detailed logging / printing, so it's clear where the system is looking in deciding the backend - kind of a form of on-demand documentation.
- Probably subjective, but it feels hacky.

One question that may be worth asking, is who is the target audience?
- General RTL engineers, for any Chisel project: maybe we should emphasize mechanisms that are analogous to what RTL tools today do - for example, environment variables? Command line flags? If we're fine with a dependency on ScalaTest, it provides its own test environment map, that you can read during the test case and define on the sbt shell. I think that last one (possibly with a generalization / interface to multiple test frameworks) is the cleanest solution, and uses an "expected" interface.
- Chisel developers / CI server, not exposed as a "public" interface for Chisel-based projects to use: can definitely be a bit more rough around the edges. Environment variables might be fine, though still raises the problem of persistent state. The file might be acceptable - it's still a potential gotcha, but affects much less people. Or maybe something as simple as changing the code that returns the default backend - where the usage is super hacky, but architecturally eliminates the risk of unintuitive behavior from most users?

chick:
@jackkoenig I rebased again, not sure how I ended up with that many commits
@ducky64 
Using the file system has been red-flagged, but I believe it to be the best compromise
- Environment variables are equally problematic as files, and are harder to use in an IDE
- The file system is relatively clean, I have provided sbt tasks for setting and clearing the file
- The file used has been added to .gitignore so that it is is unlikely to be accidentally passed up to github
- Tests will not run if the file is in a messed up state.
- The control file could conceivably be modified during test runs but I don't see it as that likely

The main target audience is developers working on chisel3
- During coding-debugging loop this mechanism provides a lightweight comprehensible way to switch to a faster running test suite

The secondary target is having variations in testing strategy during different CI runs.
- The sbt tasks created provide a standard mechanism for controlling this.

The file based approach is not immediately obvious.
- I will be providing website documentation on why and how one would use this
- There should be more documentation for internal developers and that would be a natural place to document this.
- I don't like the idea of changing the code to get different backends, that seems way more hacky than this approach.

I have been trying to get this basic goal of treadle running the chisel3 tests for over a year, in all that time no more elegant control mechanism has been proposed. The other major benefit of getting this in is it strengthens treadle testing and fidelity to verilator. As we are pushing treadle as the default backends for our unit testing frameworks this is a pretty important goal

ducky64:
So thinking about it more fundamentally, here's how I see the trade-offs:
- Under the current PR'd code: sbt file is the user-facing interface, files are used as an intermediate. Pros: reasonable interface. Cons: hacky (and potentially leaky - though in uncommon circumstances) implementation.
- Changing code directly. Pros: very clean, lightweight implementation, basically no side effects. Cons: hacky interface.

**But crucially, to change the backend, in both cases** the user needs to modify a version-controlled file (.sbt file in the first option, .scala file in the second option) and make sure to not commit the changes. Yes, changing the .sbt file is a bit cleaner and nicer, but I don't see that as so different from changing the source file. However, the difference is that in option 1, there needs to be additional (potentially leaky) plumbing and complexity, whereas that isn't the case in option 2. Both are hacky, but for an internal developer tool, I'd argue option 2 is the least-worst solution.

I'll also put in my 2c in that I don't think documentation alone is sufficient, nor should it be the first line of defense. Getting started guides are great, but beyond that, documentation needs to be discoverable (one typically does not read through a long manual or crawl websites for fun, it's probably more search for one specific task that I want to get done - or worse, search for symptoms of some weird bug and hope there's like a StackOverflow link). Even better is if documentation isn't needed in the first place, either because the functionality is "obvious", or we're re-using structures (precedent) that are common or expected in the field.

I think there's also a solution space between options 1 and 2:
- Is it possible to get SBT data directly into Scala? For example, a quick search turned up [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo), will that do what you need? It would also eliminate the raw file IO functions (to some degree, implementation details) currently in build.sbt, replacing them with a more structured version.
- If you want to avoid changing version-controlled files completely, there's the option is just having a configuration file as the interface, instead of the configuration file as an intermediate from SBT (which kind of seems like a worst-of-both-worlds solution). There's at least precedent in that some logging frameworks are configured from a separate .xml file - though I think this is kind of cumbersome. I don't know if there's a general underlying configuration framework that can be re-used.

I definitely agree that it's valuable to test with treadle (possibly / probably even having treadle be the default backend), but I'm not convinced this is the right solution.

If you wanted to widen the solution space, it might also be worth considering whether the common use case will be to switch backends for all tests, as opposed to having a default backend for most tests (that test chisel functionality), and then a few tests that do sanity checks on specific backends. Under this philosophy, where it's very rare for developers to change the backend for all tests, option 2 (from way above) makes a lot more sense. It still might be worth having CI run all tests against all backends, but adding the complexity to the CI setup might be less worse than adding complexity in chisel base.

chick:
@ducky64 said

> But crucially, to change the backend, in both cases the user needs to modify a version-controlled file (.sbt file in the first option, .scala file in the second option) and make sure to not commit the changes.

This is NOT the case. There is no requirement that the build.sbt be changed in order to switch backends. My changes to sbt provides new tasks that can be run from CI scripts that can set the backend using the external file.

I'll also put in my 2c in that I don't think documentation alone is sufficient, nor should it be the first line of defense. Getting started guides are great, but beyond that, documentation needs to be discoverable (one typically does not read through a long manual or crawl websites for fun, it's probably more search for one specific task that I want to get done - or worse, search for symptoms of some weird bug and hope there's like a StackOverflow link). Even better is if documentation isn't needed in the first place, either because the functionality is "obvious", or we're re-using structures (precedent) that are common or expected in the field.

I think there's also a solution space between options 1 and 2:

> Is it possible to get SBT data directly into Scala? For example, a quick search turned up sbt-buildinfo
We already use BuildInfo for versioning in Firrtl, it's an interesting tool  but it does not change the fact that the build.sbt file must be changed to change the backend. Given that I don't see that it provides any better solution to the problem.

Discovery of this feature is a small problem but it is not really that different from the discovery issues of, for example, how to turn on debug of the firrtl passes. Only a small pool of people are at this level of development and they ask questions. Tips like these are often just word-of-mouth.

ducky64:
So if we wanted to draw bright lines between interface and implementation, both the project .sbt and the file it reads are considered user-facing interfaces - essentially, two ways to do the same thing (possibly depending on use case - which I assume would be documented somewhere)? I'm definitely not a fan of this philosophy - yes, Scala does do it, but I think there's a lot of drawbacks unless there's strong driving use cases and guidance on which to use when.

As for documentation / discovery: really a question of trade-offs. I'd argue that if a elegant architectural fix is reasonably implement-able and provides gains in usability, then it's worth examining. Is there a reason something a solution like sbt-buildinfo won't work?

In general, I think it's worth thinking clearly about who's going to be using this, what the interface cleanliness should be, the trade-off of implementation cleanliness vs. dev time, and separation of concerns. From your reply, it looks like the driver being those functions in the .sbt is just CI hooks, which might be used just a handful of times, ever, only in the Chisel CI setup. Subjectively to me, I'd argue that those functions just add complexity to the build.sbt file (think from the standpoint of a new user - perhaps reading to chisel build.sbt in integrating to another project) - yes, can be alleviated with surrounding documentation, but maybe it's better to just move in to the CI config?

To keep progress moving, it might be prudent to split this into two PRs: one that adds the non-contentious internals (treadle .sbt dependencies, treadle runner, global default backend, excluding file reading) that could be merged sooner, while splitting the more contentious interface parts into another PR?

chick:
@ducky64 You have come up with the same solution that I did last night. I will split these PR's and we can discuss the issues separately.
I will note in passing the following.
- This features is for internal developers of Chisel3 and Continuous Integration systems.
- For the latter in particular the ability to switch backends should not require editing of source controlled files.
- I have experimented with BuildInfo, and talked to Jim about it and short of using it to read a file or environment variable, it requires editing the build.sbt (or mill equivalent)

seldridge:
FWIW, I tried to do this with environment variables and `sbt` build targets. With `sbt` you can set a variable with `envVars in Test := Map("BACKEND" -> "verilator")`. However, I couldn't sort out how to have two different testing build targets that would both inherit whatever `Test` does, but set different environment variables. That would theoretically enable my ideal API for this: `sbt treadle:test` or `sbt verilator:test` without filesystem state. However, my `sbt` skills are just way too weak and I couldn't sort out how to do it.

chick:
Replacing this with PR #1483 which just creates the Treadle backend mechanism
And a PR to be numbered later to provide a switch for this feature

PR #1277, opened 2019-12-17
Band aid until litOption is implemented for Aggregates.

ucbjrl:
This is just a band aid until an Aggregate `isLit()` method (for which work has begun) is implemented.

**Related issue**: #1256, ucb-bar/dsptools#184

**Type of change**: bug report

**Impact**: API modification

**Development Phase**: implementation

**Release Notes**
Prevent `isLit()` on Aggregates failing with a `scala.NotImplementedError: an implementation is missing` error.

PR #1280, opened 2019-12-18
Provide an implementation of litOption() for BundleLits

ucbjrl:
**Related issue**: #1277, ucb-bar/dsptools#184

**Release Notes**
Provide an implementation of `litOption` for Aggregates (Bundles).


jackkoenig:
Probably want to rebase/cherry-pick on master since that was squashed-and-merged

ucbjrl:
We should add some tests to verify this works with Vec literals as well.

ucbjrl:
Add a check for `width == 0`.

src/test/scala/chiselTests/BundleLiteralSpec.scala
```diff
@@ -28,7 +28,8 @@ class BundleLiteralSpec extends ChiselFlatSpec {
       chisel3.assert(outsideBundleLit.a === 42.U)
       chisel3.assert(outsideBundleLit.b === true.B)
       chisel3.assert(outsideBundleLit.c === MyEnum.sB)
-
+      chisel3.assert(outsideBundleLit.isLit())
+      chisel3.assert(outsideBundleLit.litValue().U === outsideBundleLit.asUInt())
```
ducky64: Is there a literal packing test anywhere? Do we want to test that too, by equivalence with a "magic number" in the test case, instead of solely equivalence with the .asUInt behavior?

ducky64:
Updated PR:
- merged in latest
- got rid of the try block, in favor of using Option throughout the foldLeft
- added packing tests, including with signed and fixed-point types

It seems to work.

But one more general concerning point: in the LongBundle literal specifier, if I don't have explicit widths for the components (not in the most-significant position, at least), it seems .asUInt will return a different value in simulation. So it seems the structural width of the bundle is being ignored for the widths of literal elements in the simulation bundle packing, which seems wrong?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -43,7 +43,29 @@ sealed abstract class Aggregate extends Data {
...
+        Some((accumulator << width) + masked)
+      case (None, _) => None
+      case (_, None) => None
+    }
+    topBindingOpt match {
+      case Some(BundleLitBinding(_)) =>
+        getElements
+          .reverse
+          .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAdd)
+      case _ =>  None
```
sequencer, line 66: maybe 1 space before `None` will make scalafmt happy?
just nitpicking :)

src/test/scala/chiselTests/BundleLiteralSpec.scala
```diff
@@ -22,14 +23,39 @@ class BundleLiteralSpec extends ChiselFlatSpec with Utils {
...
+  }
+
+  "bundle literals" should "pack" in {
+    assertTesterPasses{ new BasicTester {
+      val bundleLit = (new MyBundle).Lit(_.a -> 42.U, _.b -> false.B, _.c -> MyEnum.sB)
+      bundleLit.litOption should equal (Some(169))  // packed as 42 (8-bit), false=0 (1-bit), sB=1 (1-bit)
+      chisel3.assert(bundleLit.asUInt() === bundleLit.litOption.get.U)  // sanity-check consistency with runtime
+
+      val longBundleLit = (new LongBundle).Lit(
+        _.a -> 0xDEADDEADBEEFL.U, _.b -> -0x0BEEF00DL.S(32.W), _.c -> 4.5.F(16.W, 4.BP))
```
sequencer, line 39: maybe bracket for `(-0x0BEEF00DL).S`? I don't know does `-` have priority or `.S` does, but Intellij blames for this.

src/test/scala/chiselTests/BundleLiteralSpec.scala
sequencer: I think `LiteralExtractorSpec` provides `Lit` test for `Element` type, and tests for `Lit`.
But in my understanding, `Lit` and `LitOption` are two different thing:
`Lit` will return a `Option[BigInt]`, while `LitOption` will return `T` itself.
As for this case:
I think the result of `outsideBundleLit.litValue().U === outsideBundleLit.asUInt()` is only a necessary condition to this test.

If we wanna fully test this, I think the we should test the behavior of firrtl behavior of `chisel3.internal.firrtl.PrimOp.AsUIntOp` is equal to the `(accumulator << width) + masked)` behavior. At my first glance to `firrtl.PrimOps.AsUInt`, this behavior is correct. I think this is good enough here.

ducky64:
Fixed stylistic issues mentioned in review comments.

Not sure why GitHub isn't allowing a response to the last comment but:
- I'm not sure I completely understand what you mean. LiteralExtractorSpec does test litValue and litOption, but doesn't have tests for bundles. I think an argument could be made for the bundle litOption tests to go in LiteralExtractorSpec, but I like centralizing all the bundle literal operations (including extraction) in a bundle literal test file.
- I don't understand what you with Lit vs. LitOption. In the Data base class, litValue just delegates to litOption, so testing either makes sense?
- For `outsideBundleLit.litValue().U === outsideBundleLit.asUInt()`, this should test FIRRTL behavior since it goes through to compilation and simulation. It's a blackbox test of FIRRTL since it only checks the overall simulation result, but I think that makes sense for this test.

PR #1283, opened 2019-12-18
BitPat supports whitespace and underscores

chick:
The whitespace and _ are presumably for human readability.

**Related issue**: 
This is a fix for chisel3 PR #1069

**Type of change**: bug fix

The BitPat.parse factory though did take these into account when 
reporting the bit pattern size in `BitPat.parse`.

This fixes that adds whitespace and underscores to the unit tests

**Release Notes**
Fixes `BitPat.parse` returned value for the size of the bit pattern, which was not
adjusted for the presence of _ or whitespace (legal formatting characters)

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -14,7 +14,7 @@ object BitPat {
     * @return bits the literal value, with don't cares being 0
     * @return mask the mask bits, with don't cares being 0 and cares being 1
     * @return width the number of bits in the literal, including values and
-    * don't cares.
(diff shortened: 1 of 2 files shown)


jackkoenig:
I think spaces and maybe tabs are fine, but do we really want all whitespace supported? I guess I don't have any reason to be strongly opposed but like, newlines? vertical tabs?

chick:
@jackkoenig I don't have strong opinions and I really don't think it will come up. But the comments (which could be changed) don't distinguish kinds of whitespace, so to restrict would be a teeny tiny API change.

Issue #1238, opened 2019-11-17
Add asBool to Clock


stevobailey:
**Type of issue**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: request

**What is the current behavior?**
To change a Clock to a Bool, I have to do myClock.asUInt.asBool.

**What is the expected behavior?**
Simply call asBool: myClock.asBool.

chick:
This has been done in PR #1285

PR #1285, opened 2019-12-19
Add method asBool to Clock.

chick:
**Related issue**: 
Addresses Chisel issue #1238

**Impact**: Minor API addition (no impact on existing code)

**Release Notes**
There is now a short cut method for turning a clock to a bool
`clock.asBool`. Previously had to do `clock.asUInt.asBool`

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Clock.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Clock.scala
@@ -2,7 +2,8 @@
 
 package chisel3
 
-import chisel3.internal.Builder.{pushOp}
+import scala.language.experimental.macros
+import chisel3.internal.Builder.pushOp
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo._
 import chisel3.internal.firrtl.PrimOp.AsUIntOp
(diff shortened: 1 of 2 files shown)


chiselFrontend/src/main/scala/chisel3/Clock.scala
```diff
@@ -30,9 +31,13 @@ sealed class Clock(private[chisel3] val width: Width = Width(1)) extends Element
...
 
+  /** Returns the contents of this wire as a [[scala.collection.Seq]] of [[Bool]]. */
+  final def asBool(): Bool = macro SourceInfoTransform.noArg
+  def do_asBool(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = this.asUInt().asBool()
+
   override def do_asUInt(implicit sourceInfo: SourceInfo, connectCompileOptions: CompileOptions): UInt = pushOp(DefPrim(sourceInfo, UInt(this.width), AsUIntOp, ref)) // scalastyle:ignore line.size.limit
   private[chisel3] override def connectFromBits(that: Bits)(implicit sourceInfo: SourceInfo,
       compileOptions: CompileOptions): Unit = {
-    this := that.asBool.asClock
+    this := that.asBool().asClock
```
jackkoenig: Why are you adding parentheses here?

chick: Accidental, removed them

src/test/scala/chiselTests/Clock.scala
```diff
@@ -18,6 +19,8 @@ class WithClockAndNoReset extends RawModule {
   val a = withClock(clock2) {
     RegNext(in)
   }
+
+  in.asUInt()
```
ucbjrl: What is the purpose of this cast?

chick: None, remove it. Thanks for catching

chiselFrontend/src/main/scala/chisel3/Clock.scala
```diff
@@ -30,6 +31,10 @@ sealed class Clock(private[chisel3] val width: Width = Width(1)) extends Element
   /** Not really supported */
   def toPrintable: Printable = PString("CLOCK")
 
+  /** Returns the contents of this wire as a [[scala.collection.Seq]] of [[Bool]]. */
```
ucbjrl: I'm confused. `asBool` returns a `Seq`? Shouldn't the method be `asBools`?

aswaterman: I think the method is as intended, but the comment is wrong

ucbjrl: I've updated the comment. If others approve, so do I.

PR #1286, opened 2019-12-19
Fix add other module's port

colin4124:

## Testcase

```scala
class Mux2 extends RawModule {
  def foo = IO(Input(UInt(1.W))).suggestName("foo")
  val sel = IO(Input(UInt(1.W)))
  val in0 = IO(Input(UInt(1.W)))
  val in1 = IO(Input(UInt(1.W)))
  val out = IO(Output(UInt(1.W)))
  out := (sel & in1) | (~sel & in0)
}

class Mux4 extends RawModule {
  val in0 = IO(Input(UInt(1.W)))
  val in1 = IO(Input(UInt(1.W)))
  val sel = IO(Input(UInt(1.W)))
  val out = IO(Output(UInt(1.W)))

  val m0 = Module(new Mux2)
  m0.sel := sel
  m0.in0 := in0
  m0.in1 := in1

  m0.foo

  out := m0.out
}
```

`Mux2`'s port `foo` is called in module `Mux4` and then it would add the port `foo` to module `Mux4`.

```firrtl
circuit Mux4 : 
  module Mux2 : 
    input sel : UInt<1>
    input in0 : UInt<1>
    input in1 : UInt<1>
    output out : UInt<1>
    ...
  module Mux4 : 
    input in0 : UInt<1>
    input in1 : UInt<1>
    input sel : UInt<1>
    output out : UInt<1>
    input foo : UInt<1>
    ...
```

Fix this problem and will throw exception:

```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: Mux4 can't add other module Mux2's port
```

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Module.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Module.scala
@@ -100,8 +100,9 @@ package experimental {
       * requested (so that all calls to ports will return the same information).
       * Internal API.
       */
-    def apply[T<:Data](iodef: T): T = {
+    def apply[T<:Data](ctx: BaseModule, iodef: T): T = {
       val module = Module.currentModule.get // Impossible to fail
+      require(ctx == module, s"${module.name} can't add other module ${ctx.name}'s port")
       require(!module.isClosed, "Can't add more ports after module close")
       requireIsChiselType(iodef, "io type")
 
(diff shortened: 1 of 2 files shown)


azidar:
I'm not sure this is actually desired behavior - @ducky64 what do you think?

azidar:
After a discussion in the dev meeting:

- Shouldn't do something to IO that's different from RegInit
- module.IO does the same thing as experimental.IO under the hood
- There are use cases for describing hookup logic in a child that is called from the parent

Conclusion: Existing behavior is ok (not a bug). Closing this issue - feel free to reopen if needed.

ducky64:
To clarify a bit: your test case is an example of where current behavior is leading to unintuitive behavior, but we weren't able to come up with a fix that doesn't break code compatibility (add explicit "context" Module argument to global `experimental.IO`) and maintain symmetry with `Reg` / `Wire` / etc (force `Module.IO` to check against 'this', while (global) `experimentl.IO` and `Reg` doesn't).

PR #1301, opened 2020-01-17
Add test for fixed-width addition for FixedPoint.

grebe:
This is a failing test! It shouldn't fail, though.

The way that fixed-width addition works relies on having a defined binary point. This is an unfortunate limitation, and should be fixed.

**Related issue**: 

**Development Phase**: test demonstrating bug

**Release Notes**

### DIFF
--- a/src/test/scala/chiselTests/FixedPointSpec.scala
+++ b/src/test/scala/chiselTests/FixedPointSpec.scala
@@ -131,6 +131,23 @@ class FixedPointLitExtractTester extends BasicTester {
   stop()
 }
 
+class FixedPointInferenceTester extends BasicTester {
+  val start = WireInit(1.0.F(10.W, 5.BP))
+
+  val blankWire = Wire(FixedPoint())
+
+  blankWire := start // should be inferred in FIRRTL backend
+
+  val sumGrow = blankWire +& blankWire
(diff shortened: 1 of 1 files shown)


chick:
@azidar Any theories as to how to fix this. The +% is adding tail(1) to the chirrtl and the converting back to fixedpoint with asFixedPoint. This fails because it asFixedPoint requires
a known binary point. This might be possible to fix somehow in firrtl passes but currently this fails in chisel well before it gets to firrtl. @grebe

chick:
@grebe Can you update this branch with master, it's complicating my testing

chick:
@azidar I'm wondering if this could be fixed in the width inference pass. When the width and binary point of a unknown get decided could it look down the logic flow to detect this situation and add the binary point to the following asFixedPoint

chick:
I've tried some experiments where I convert `fixpoint1 +% fixpoint1` to 
`fixedpoint1.asInterval +% fixedpoint2.asInterval`, but there doesn't seem to be a way to 
create chirrtl `asInterval` with unknown bounds or binary points.
@azidar I think you mentioned using shift somehow, but I don't know that would look like

jackkoenig:
FixedPoint has been removed.

PR #1303, opened 2020-01-17
Bugfix: Select.instances now works with blackboxes

azidar:

**Release Notes**
Previously, `Select.instances` would fail on a cast error if called on a module that was a black box. Now, it just returns `Nil`, because a black box by default has no instances. This enables other functions to rely on more consistent behavior, such as `Select.collectDeep`.

### DIFF
--- a/src/test/scala/chiselTests/aop/SelectSpec.scala
+++ b/src/test/scala/chiselTests/aop/SelectSpec.scala
@@ -7,7 +7,9 @@ import chiselTests.ChiselFlatSpec
 import chisel3._
 import chisel3.aop.Select.{PredicatedConnect, When, WhenNot}
 import chisel3.aop.{Aspect, Select}
-import firrtl.{AnnotationSeq}
+import chisel3.experimental.ExtModule
+import chisel3.stage.{ChiselGeneratorAnnotation, DesignAnnotation}
+import firrtl.AnnotationSeq
 
 import scala.reflect.runtime.universe.TypeTag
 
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/aop/SelectSpec.scala
```diff
@@ -139,5 +141,17 @@ class SelectSpec extends ChiselFlatSpec {
...
   }
 
+  "Blackboxes" should "be supported in Select.instances" in {
+    class BB extends ExtModule { }
+    class Top extends RawModule {
+      val bb = Module(new BB)
+    }
+    val top = ChiselGeneratorAnnotation(() => {
+      new Top()
+    }).elaborate(1).asInstanceOf[DesignAnnotation[Top]].design
```
seldridge: This `elaborate(1)` threw me for a loop... Maybe `elaborate.head`? (That's a nit.)

azidar: Its not head, its the second in the list.. `elaborate.tail.head`?

seldridge: Ugh... not thinking (or thinking in tuples). `.tail.head` isn't amazing either. Pattern match (`case _ :: a :: _ => `) might be the most readable. Not worth it, though. I'd just merge it.

azidar:
@chick yeah I don't think it could possibly effect existing code, unless they were catching the exception, which I think would be fine to break.

PR #1310, opened 2020-01-24
Fix syntax error in Pipe example

antonblanchard:


jackkoenig:
@antonblanchard can you sign the CLA?

seldridge:
Friendly ping @antonblanchard.

nullobject:
I think we could go one step further with the fix, the name should be `pipe`:

```scala
 val pipe = Module(new Pipe(UInt(8.W), 4))
```

src/main/scala/chisel3/util/Valid.scala
```diff
@@ -152,7 +152,7 @@ object Pipe {
   * As an example, to construct a 4-stage pipe of 8-bit [[UInt]]s and connect it to a producer and consumer, you can use
   * the following:
   * {{{
-  *   val foo = Module(new Pipe(UInt(8.W)), 4)
+  *   val foo = Module(new Pipe(UInt(8.W), 4))
```
nullobject, line 155: I think the value should actually be named `pipe`:

```suggestion
  *   val pipe = Module(new Pipe(UInt(8.W), 4))
```

jackkoenig: 👍

PR #1315, opened 2020-01-28
Emit FIRRTL andr, orr for Bits.{andR, orR}

seldridge:
Change the emission strategy for `Bits` methods `andR` and `orR` to emit FIRRTL bitwise reduce operations `andr` and `orr`. The CHIRRTL emission strategy for these two operations now matches the emission of `xorR` (which already used `xorr`).

Add two tests that assert the correct behavior of these operations in `BitwiseOpsSpec`.

This should fix one of the Verilog emission problems shown in freechipsproject/firrtl#1338 of:

```scala
class Foo(n: Int) extends RawModule {
  val in = IO(Input(UInt(n.W)))
  val out = IO(Output(Vec(3, Bool())))
  out(0) := in.andR
  out(1) := in.orR
  out(2) := in.xorR
}
```

This will now emit the following CHIRRTL for `Foo(4)`:

```
circuit Foo : 
  module Foo : 
    input in : UInt<4>
    output out : UInt<1>[3]
    
    node _T = andr(in)
    out[0] <= _T
    node _T_1 = orr(in)
    out[1] <= _T_1
    node _T_2 = xorr(in)
    out[2] <= _T_2
```

And the resulting Verilog:

```verilog
module Foo(
  input  [3:0] in,
  output       out_0,
  output       out_1,
  output       out_2
);
  assign out_0 = &in;
  assign out_1 = |in;
  assign out_2 = ^in;
endmodule
```

For the case of a zero-width wire, you get the reductions that @albert-magyar introduced in https://github.com/freechipsproject/firrtl/pull/1362.

```verilog
module Foo(
  output  out_0,
  output  out_1,
  output  out_2
);
  assign out_0 = 1'h1;
  assign out_1 = |1'h0;
  assign out_2 = ^1'h0;
endmodule

```

**Related issue**: 
  - https://github.com/freechipsproject/firrtl/issues/1338
  - Depends on https://github.com/freechipsproject/firrtl/pull/1362

**Impact**: API modification (to emitted CHIRRTL)

**Release Notes**

- Change andR, orR emission to use FIRRTL andr, orr

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Bits.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Bits.scala
@@ -581,13 +581,9 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
   final def xorR(): Bool = macro SourceInfoTransform.noArg
 
   /** @group SourceInfoTransformMacro */
-  def do_orR(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = this =/= 0.U
+  def do_orR(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = redop(sourceInfo, OrReduceOp)
   /** @group SourceInfoTransformMacro */
-  def do_andR(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = width match {
-    // Generate a simpler expression if the width is known
-    case KnownWidth(w) => this === ((BigInt(1) << w) - 1).U
-    case UnknownWidth() =>  ~this === 0.U
-  }
+  def do_andR(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = redop(sourceInfo, AndReduceOp)
   /** @group SourceInfoTransformMacro */
   def do_xorR(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = redop(sourceInfo, XorReduceOp)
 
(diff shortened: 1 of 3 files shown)


seldridge:
(Added this above...) This PR does change how `andR` behaves for a zero-width `UInt`. 

Previously, this would generate:

```
outAnd = 1'h1;
```

This PR changes this emission to:

```
outAnd = &1'h0;
```

aswaterman:
The old behavior (andR on a 0-bit UInt returns 1) is correct, for the same reason that `Seq().forall(f)` returns true.  Otherwise, invariants like `(a ## b).andR === a.andR && b.andR` don't hold up.

seldridge:
While we can special case this in Chisel, this should be fixed in FIRRTL first. That Chisel is generating the following FIRRTL which results in the Verilog that is incorrect as @aswaterman identifies.

```
circuit Foo_0 : 
  module Foo_0 : 
    input in : UInt<0>
    output outAnd : UInt<1>

    node _T = andr(in)
    outAnd <= _T
```

chick:
@seldridge can we address @aswaterman concern? Requires a firrtl change?

seldridge:
With the @albert-magyar fixes in https://github.com/freechipsproject/firrtl/pull/1362, this is now doing the right thing in terms of Verilog emission. The language up top has been updated accordingly. I'm going to go ahead and merge as there is no effective change to the content of this PR.

Issue #1311, opened 2020-01-24
RegNext Widths


seldridge:
The asymmetry of how `Reg`, `RegNext`, and `RegInit` set their widths can lead to unexpected behavior. 

h/t @antonblanchard

**Type of issue**: bug report | feature request | other enhancement

**Impact**: API modification

**Development Phase**: request

**Other information**
 

There's an old PR referencing and documenting some of this behavior: https://github.com/freechipsproject/chisel3/pull/455

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

Consider the following example. While `foo`, `bar`, and `baz` have almost identical construction, `foo` has an unset width, `bar` was a width of 8, and `baz` is a bundle with an internal signal of width `8`.

```scala
package examples

import chisel3._
import chisel3.stage.{ChiselStage, ChiselGeneratorAnnotation}
import chisel3.internal.sourceinfo.UnlocatableSourceInfo

object NoInfo {
  implicit val noInfo = UnlocatableSourceInfo
}

import NoInfo._

class DelayedBundle extends Bundle {
  val in = UInt(8.W)
}

class Delayed extends MultiIOModule {
  val in = IO(Input(UInt(8.W)))
  val bundle = IO(Input(new DelayedBundle))

  /** The width of foo is unset */
  val foo = RegNext(in)

  /** The width of bar is set to 8.W */
  val bar = Reg(chiselTypeOf(in))
  bar := in
  
  /** The width of baz is set to 8.W */
  val baz = RegNext(bundle)

  Seq(foo, bar, baz).foreach(a => println(a.widthOption))

}

object Delayed extends App {
  (new ChiselStage).execute(Array.empty, Seq(ChiselGeneratorAnnotation(() => new Delayed)))
}
```

This results in the following FIRRTL:

```
circuit Delayed : 
  module Delayed : 
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<8>
    input bundle : {in : UInt<8>}
    
    reg foo : UInt, clock
    foo <= in
    reg bar : UInt<8>, clock
    bar <= in
    reg baz : {in : UInt<8>}, clock
    baz.in <= bundle.in
```

[...]

PR #1318, opened 2020-02-05
Add Scaladoc about RegNext Unset/Inferred Widths

seldridge:
Updates Scaladoc for RegNext to make it more clear about its behavior:

RegNext

**Related issue**: #1311 

**Release Notes**

- Add Scaladoc for RegNext about unset/inferred widths

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Reg.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Reg.scala
@@ -48,11 +48,33 @@ object Reg {
 
 }
 
+/** Utility for constructing one-cycle delayed versions of signals
+  *
+  * ''The width of a `RegNext` is not set based on the `next` or `init` connections'' for [[Element]] types. In the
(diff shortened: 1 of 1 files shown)


chiselFrontend/src/main/scala/chisel3/Reg.scala
```diff
@@ -48,11 +48,33 @@ object Reg {
...
+  *   val x = UInt(4.W)
+  * }
+  * val foo = Wire(new MyBundle)     // the width of foo.x is 4
+  * val bar = RegNext(foo)           // the width of bar.x is 4
+  * }}}
+  */
 object RegNext {
-  /** Returns a register with the specified next and no reset initialization.
-    *
-    * Essentially a 1-cycle delayed version of the input signal.
```
ucbjrl: The 1-cycle delay is probably obvious to HW developers. Should its mention be retained for the benefit of those coming from a pure SW perspective?

PR #1320, opened 2020-02-06
Mux1H: don't special-case (sel: UInt, in: UInt)

johningalls-sifive:
**Related issue**: https://github.com/freechipsproject/chisel3/issues/1314

**Impact**: API modification

**Release Notes**
if instantiating a `Mux1H(sel: UInt, in: UInt)` when `in.getWidth`==1, do the same behavior as instantiating a `Mux1H(sel.asBools, in.asBools)`: ignore the select and connect the input directly to the output

### DIFF
--- a/src/main/scala/chisel3/util/Mux.scala
+++ b/src/main/scala/chisel3/util/Mux.scala
@@ -19,7 +19,7 @@ import chisel3._
   * ))
   * }}}
   *
-  * @note results undefined if multiple select signals are simultaneously high
+  * @note results unspecified unless exactly one select signal is high
   */
 object Mux1H {
   def apply[T <: Data](sel: Seq[Bool], in: Seq[T]): T =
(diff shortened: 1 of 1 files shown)


aswaterman:
Bringing this Mux1H API in line with the others seems like a good thing to do, but it raises some subtle issues:
- This new version calls `asBools`, so it'll bail out if the width isn't known at elaboration time, whereas the previous version would just work.
- Even though this brings the API in line with the others, and even though both the old and new behaviors are within the bounds of a one-hot API, it is a change that some users could theoretically notice.  I guess we need a release note or something.
- The original `@note` is incomplete, as it fails to mention the zero-hot case.  It should say "results unspecified unless exactly one select signal is high".  Then the new note you added isn't necessary.

It is possible to get rid of the dependence on `asBools` by doing something esoteric like

    val isLessThanTwoBitsWide = (sel + 1.U) === (sel - 1.U)
    Mux(isLessThanTwoBitsWide, in, (sel & in).orR)

where the first statement is a way to check whether `sel` is one bit wide, but to postpone the check til Firrtl time.  (There's got to be a better way of doing that...)

albert-magyar:
@aswaterman, I think the idea for your example is that the `Mux` will be optimized away via const-prop after the select case is coerced to a constant?

If so, I imagine that implicitly depends on the idea that `(sel + 1.U)` and `(sel - 1.U)` will be strength-reduced to matching expressions for 1-bit `sel`. This currently does not happen in FIRRTL but seems reasonable to add.

It also in turn depends on FIRRTL const-propping `a == a`, which I had assumed might already be the case but is actually not currently done. I just PRed those matching-argument degeneracies to const prop in freechipsproject/firrtl#1361.

aswaterman:
Yeah, we would obviously want some support from Firrtl. Synthesis will get
rid of this for sure, but the Verilog is still unreadable.

johningalls-sifive:
> I definitely don’t want to change the behavior of the 1-wide case for the other ones. If you want the and-or thing even in the 1-wide case, it’s a one-liner in Scala, or you can make your own library method for it rather than using the Chisel one.

will do

https://github.com/freechipsproject/chisel3/issues/1314#issuecomment-583031365

PR #1323, opened 2020-02-06
Printf: Add support for tabs, and give helpful error messages

mwachs5:
**Related issue**: 
https://github.com/freechipsproject/chisel3/issues/1321

I added support for \t in printf. I also made the `require` statements give more information if you do something wrong which is how I figured out the tab was the problem.

**Impact**: API modification

**Release Notes**

Printables printf now supports \t characters.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Printf.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Printf.scala
@@ -26,7 +26,7 @@ object printf { // scalastyle:ignore object.name
         "\\n"
       } else if (x == '\t') {
         "\\t"
-      }else {
+      } else {
         require(x.toInt >= 32, s"char ${x} to Int ${x.toInt} must be >= 32") // TODO \xNN once FIRRTL issue #59 is resolved
         x
       }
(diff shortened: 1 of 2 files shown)


chiselFrontend/src/main/scala/chisel3/Printf.scala
```diff
@@ -19,13 +19,15 @@ object printf { // scalastyle:ignore object.name
...
+      require(x.toInt >= 0, s"char ${x} to Int ${x.toInt} must be >= 0")
       if (x == '"' || x == '\\') {
         s"\\${x}"
       } else if (x == '\n') {
         "\\n"
-      } else {
-        require(x.toInt >= 32) // TODO \xNN once FIRRTL issue #59 is resolved
+      } else if (x == '\t') {
+        "\\t"
+      }else {
```
jackkoenig:
```suggestion
      } else {
```

jackkoenig:
Oh I didn't realize this was against `3.2.x`, we usually PR against master and then backport. That's okay, I'll forward port this.

mwachs5:
oh, are the tests passing now? cuz it failed before...

jackkoenig:
Yeah, you fixed the failing test

PR #1329, opened 2020-02-11
Clone child elements lazily in Vec

jackkoenig:
Credit to @aswaterman for this idea while we walked across Bordeaux at ORConf. We can lazily clone the elements of `Vecs` to improve performance. The key observation is that most clones are thrown away, so why do we bother creating more than just the `sample_element` for a `Vec` unless necessary?

This simple change has an absolutely **massive** performance impact on SiFive designs:

| Design | Runtime - master | Runtime - this branch |
| -- | -- | -- |
| Small | 108 seconds | 39 seconds |
| Large | 8 minutes | 3 minutes |

If anyone isn't convinced, try running the following:
```scala
object MyBundle {
  private var count = 0L
  def inc(): Unit = this.synchronized {
    count += 1
    println(s"Cloned $count times")
  }
}

  MyBundle.inc()
  val foo = UInt(8.W)
}

  val io = IO(new Bundle {
    val in = Input(Vec(4, Vec(4, Vec(4, new MyBundle))))
  })
}

object MyMain extends App {
  chisel3.Driver.elaborate(() => new MyModule)
}
```

On master, you'll see:
```
[info] Cloned 4862 times
```
On this branch:
```
[info] Cloned 682 times
```

**Note that 682 times is kind of absurd.** I think there is so much more room for improvement, but this simple change is worth it on its own.

**Related issue**: 

**Release Notes**
Improve performance by lazily cloning Vec elements

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/Aggregate.scala
+++ b/chiselFrontend/src/main/scala/chisel3/Aggregate.scala
@@ -165,9 +165,12 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
   // Note: the constructor takes a gen() function instead of a Seq to enforce
   // that all elements must be the same and because it makes FIRRTL generation
   // simpler.
-  private val self: Seq[T] = Vector.fill(length)(gen)
-  for ((elt, i) <- self.zipWithIndex)
-    elt.setRef(this, i)
+  private lazy val self: Seq[T] = {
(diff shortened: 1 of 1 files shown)


aswaterman:
Thanks again for prototyping this idea! It has been a while since we saw that magnitude of perf improvement.

davidbiancolin:
See, Bordeaux does make good wine.

aswaterman:
lol

PR #1337, opened 2020-02-12
Added Overprovision Widths transform and tests

azidar:

**Release Notes**
This API is useful for physical designers who are partitioning a design and want to "freeze" a sub-hierarchy's interface, including the port widths. If a future run causes different width inference, or different parameters cause different widths to be generated, this enables a designer to overprovision a port width so future changes are less likely to break the interface.

### DIFF
--- a/src/main/scala/chisel3/util/experimental/OverprovisionWidths.scala
+++ b/src/main/scala/chisel3/util/experimental/OverprovisionWidths.scala
@@ -1,0 +1,289 @@
+// See LICENSE for license details.
+
+package chisel3.util.experimental
+
+import chisel3.Bits
+import chisel3.experimental.{ChiselAnnotation, RunFirrtlTransform, annotate}
+import chisel3.internal.firrtl.Width
+import firrtl.passes.{InferTypes, PadWidths, Pass}
+import firrtl.annotations.{Annotation, CircuitTarget, ModuleTarget, ReferenceTarget, SingleTargetAnnotation}
+import firrtl.{ResolvedAnnotationPaths, Transform}
+
+import scala.collection.mutable
+
+
+/** Use to overprovision the width of a module port
+  *
+  * Regardless of the width inferred, this utility will increase the port width to the desired width
+  * Any references to the port are properly trimmed to not modify circuit behavior
+  * Any assignments to the port are properly padded to the overprovisioned width
+  *
+  * If the port is deleted (e.g. by dead code elimination), nothing is overprovisioned
+  * If you do not want your port deleted, use [[chisel3.dontTouch]] to mark the signal
+  *
+  * If the same signal is overprovisioned multiple times, the largest width wins
+  *
+  * This generates errors on the following cases:
+  * - overprovision width is smaller than the existing width
+  * - port is not a UInt or SInt
+  * - port is not an IO
+  */
+object overprovision {
+  def apply(port: Bits, width: Width): Unit = {
+    require(width.known, s"You can only overprovision to a known width!")
+    annotate(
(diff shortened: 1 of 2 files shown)


mwachs5:
👍

src/main/scala/chisel3/util/experimental/OverprovisionWidths.scala
```diff
@@ -0,0 +1,289 @@
...
+        targetWidthMap(target) = math.max(width, targetWidthMap.getOrElse(target, 0))
+        Nil
+      case other => Seq(other)
+    }
+
+    val circuitTarget = CircuitTarget(state.circuit.main)
+    val allErrors = mutable.ListBuffer[OverprovisionedWidthException]()
+
+    // Overprovision all ports of all modules, collect errors
+    val overProvisionedModules = state.circuit.modules.map {
```
albert-magyar:
```suggestion
    val overprovisionedModules = state.circuit.modules.map {
```

src/main/scala/chisel3/util/experimental/OverprovisionWidths.scala
```diff
@@ -0,0 +1,289 @@
...
+        newM
+      }.getOrElse(m)
+    }
+
+    // Build new map for trimming
+    val modulePorts = moduleToTargetWidthMap.map{ case (mname, targetMap) =>
+      mname -> targetMap.map { case (ref: ReferenceTarget, i: Int) => ref.ref -> i}.toMap
+    }.toMap
+
+    val trimmedModules = overProvisionedModules map { m => trimModule(m, modulePorts) }
```
albert-magyar:
```suggestion
    val trimmedModules = overprovisionedModules map { m => trimModule(m, modulePorts) }
```

src/main/scala/chisel3/util/experimental/OverprovisionWidths.scala
```diff
@@ -0,0 +1,289 @@
...
+    * @param module module to trim
+    * @param modulePorts map of (moduleName -> (map of (portName -> overprovisioned width)))
+    * @return module with trimmed expressions
+    */
+  def trimModule(module: DefModule, modulePorts: Map[String, Map[String, Int]]): DefModule = {
+    val instancePorts = buildInstancePortMap(module, modulePorts)
+    val ports = modulePorts.get(module.name)
+    def onStatement(s: Statement): Statement = s map onStatement map trimExpressions(ports.get, instancePorts)
+    if(ports.nonEmpty) module map onStatement else module
+  }
```
jackkoenig: If we don't want to run `InferTypes`, we could fix the `Expressions` in `trimExpressions` (ie. giving them the correct type). Because they are being trimmed anyway, this should be relatively easy since ultimately the type is the same.

Good way to test that this is correct is run `CheckTypes` after the transform in the tests. I think that means doing something more sophisticated than just `compile` though.

jackkoenig: Oh and we also need to update the type of the instance `WRefs`, to be more concrete:
* Inside a module with an overprovisioned port, fix the type of all `WRef(<overprovisioned port name>, ...)`
* In modules that *instantiate* modules with overprovisioned ports fix 3 types:
    * Fix type of the instance itself: `WDefInstance(<instance name>, ...)` (the `BundleType`)
    * Fix type of  `WRef` pointing to the instance (same type as `WDefInstance`)
    * Fix type of `WSubfield` pointing to the port

azidar: Because this works as is on the *.3 release, do you still think I should update it? Or are we ok as is?

PR #1358, opened 2020-02-24
CompileOptions: allow the top module reset to be assumed as Async

mwachs5:
This PR adds an additional compile option which overrides the current behavior of always assuming that if you have `inferResets` option set, but top level module doesn't actually specify its reset type, then it assumes `Bool`. This lets you override that behavior.

**Related issue**:  https://github.com/freechipsproject/chisel3/issues/1355
Also related to https://github.com/freechipsproject/firrtl/issues/1406

**Release Notes**

Additional field `topAsyncReset: Bool = false` added to `CompileOptions` which allows you to override the reset type for the top module.

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
+++ b/chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
@@ -22,7 +22,7 @@ trait CompileOptions {
   val explicitInvalidate: Boolean
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
-  // Should the top module reset, if inferred, be an Async instead of a Bool?
+  // Should the top module reset be an Async instead of a Bool
   val topAsyncReset: Boolean
 }
 
@@ -55,7 +55,7 @@ object ExplicitCompileOptions {
     val explicitInvalidate: Boolean,
     // Should the reset type of Module be a Bool or a Reset
     val inferModuleReset: Boolean,
-    // Should the top module reset, if inferred, be a Async instead of Bool?
+    // Should the top module reset be a Async instead of Bool
     val topAsyncReset: Boolean
   ) extends CompileOptions
 
--- a/chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
+++ b/chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
@@ -22,6 +22,8 @@ trait CompileOptions {
   val explicitInvalidate: Boolean
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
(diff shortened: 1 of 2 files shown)


jackkoenig:
Please make PRs against `master`, we use Mergify to backport to the release branches.

jackkoenig:
Also this fails the binary compatibility check, so unfortunately, this probably isn't back-portable.

chiselFrontend/src/main/scala/chisel3/RawModule.scala
```diff
@@ -148,7 +148,14 @@ abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
   val reset: Reset = {
     // Top module and compatibility mode use Bool for reset
     val inferReset = _parent.isDefined && moduleCompileOptions.inferModuleReset
-    IO(Input(if (inferReset) Reset() else Bool()))
+    val topAsyncReset = !_parent.isDefined && moduleCompileOptions.topAsyncReset
```
jackkoenig:
```suggestion
    val topAsyncReset = _parent.isEmpty && moduleCompileOptions.topAsyncReset
```

chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
```diff
@@ -52,7 +54,9 @@ object ExplicitCompileOptions {
     // Require an explicit DontCare assignment to generate a firrtl DefInvalid
     val explicitInvalidate: Boolean,
     // Should the reset type of Module be a Bool or a Reset
-    val inferModuleReset: Boolean
+    val inferModuleReset: Boolean,
+    // Should the top module reset, if inferred, be a Async instead of Bool?
+    val topAsyncReset: Boolean
```
jackkoenig: I would not mention the "if inferred". Whether inferred or not, the current behavior is that the top-reset is `Bool`. This is kind of an orthogonal parameter.

mwachs5: but it is not my intenntion to override if someone actually has a module with a Bool reset (which might not even make sense now that i think about it... not sure how to specify that in Chisel)

mwachs5:
"Please make PRs against master, we use Mergify to backport to the release branches." -> Done

chiselFrontend/src/main/scala/chisel3/CompileOptions.scala
jackkoenig: For this case, since it's only the top module, it is **only** the case that the top is a `Bool`, so I would argue that the current behavior is that this parameter already exists (implicitly), it's just always hardcoded to `false`.

mwachs5:
I believe #1361 would render this unncessary

mwachs5:
closing in favor of #1361

PR #1359, opened 2020-02-25
Cleanup aspects

azidar:
**Type of change**:  feature request | other enhancement

**Impact**: API modification

**Release Notes**

Improves the Aspect API with the following changes:
- Remove type-tags (it never needed them to begin with)
- Added AspectLibrary to enable parameterless-aspects from the commandline (either case objects or case classes with no arguments work correctly)
- Added abstract class InjectorAspect (and InspectorAspect) which enables creating custom parameterless injecting aspects


src/main/scala/chisel3/aop/AspectLibrary.scala
```diff
@@ -0,0 +1,47 @@
...
+
+package chisel3.aop
+
+import chisel3.RawModule
+import firrtl.options.{OptionsException, RegisteredLibrary, ShellOption}
+
+/** Enables adding aspects to a design from the commandline, e.g.
+  *  sbt> runMain chisel3.stage.ChiselMain --module <module> --with-aspect <aspect>
+  */
+case class AspectLibrary() extends RegisteredLibrary  {
```
seldridge: Any reason why this is a `case class` as opposed to a straight `class`?

src/main/scala/chisel3/aop/AspectLibrary.scala
```diff
@@ -0,0 +1,47 @@
...
+  val name = "AspectLibrary"
+
+  import scala.reflect.runtime.universe._
+
+  private def isObject[T](x: T)(implicit tag: TypeTag[T]): Boolean = PartialFunction.cond(tag.tpe) {
+    case SingleType(_, _) => true
+  }
+
+  def apply(aspectName: String): Aspect[RawModule] = {
```
seldridge: Any reason to not make this `private`?

src/main/scala/chisel3/aop/AspectLibrary.scala
azidar: No, but I don't think it should be extended. Perhaps make it `final` ?

src/main/scala/chisel3/aop/AspectLibrary.scala
azidar:
```suggestion
  private def apply(aspectName: String): Aspect[RawModule] = {
```

src/main/scala/chisel3/aop/AspectLibrary.scala
azidar:
```suggestion
final class AspectLibrary() extends RegisteredLibrary  {
```

PR #1360, opened 2020-02-25
Make implicit clock and reset final vals

jackkoenig:
Overriding will always result in a `NullPointerException` due to https://github.com/freechipsproject/chisel3/blob/aac6e175645960b65eace70d16d0e4df523e0327/chiselFrontend/src/main/scala/chisel3/RawModule.scala#L155 which 

This is definitely not source compatible, so I don't think we should backport it. Although if someone is doing this it cannot be a Module they ever actually construct because it's `NullPointerException` 100% of the time.

**Related issue**: 

**Type of change**: bug fix

**Impact**: API modification

**Release Notes**
Make implicit clock and reset final vals fixing buggy hole in API

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/RawModule.scala
+++ b/chiselFrontend/src/main/scala/chisel3/RawModule.scala
@@ -144,8 +144,8 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
 abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
     extends RawModule {
   // Implicit clock and reset pins
-  val clock: Clock = IO(Input(Clock()))
-  val reset: Reset = {
+  final val clock: Clock = IO(Input(Clock()))
+  final val reset: Reset = {
     // Top module and compatibility mode use Bool for reset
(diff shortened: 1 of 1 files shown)


jackkoenig:
@freechipsproject/chisel-reviewers any opinions on this? Fixes a hole in the API

sdtwigg:
Nit: it should still work if you use an early initializer so I don't think it is quite 100%

jackkoenig:
> Nit: it should still work if you use an early initializer so I don't think it is quite 100%

Hey @sdtwigg! I honestly had forgotten about that feature. It does provide a potentially better way to express https://github.com/freechipsproject/chisel3/pull/1361, although perhaps esoteric...

albert-magyar:
Deprecated in 2.14, too

jackkoenig:
I believe `2.14` has been dropped, but they are removed from dotty, replaced by trait parameters which may provide a better mechanism anyway: https://dotty.epfl.ch/docs/reference/dropped-features/early-initializers.html

I think it's worth reconsidering this and #1361 for longer term, but for the purposes of `3.3` I think this is best.

PR #1361, opened 2020-02-25
Provide API to set concrete type of implicit reset

jackkoenig:
Introduces mutually-exclusive traits RequireAsyncReset and
RequireSyncReset to set the type of the implicit reset in
MultiIOModules. The Scala-type remains Reset, but the Chisel
elaboration-time checks apply.

This implements my proposal from the Chisel Dev Meeting today (24 Feb 2020). In the notes this is the "subtractive" approach because `MultiIOModule` has an implicit reset of type `Reset` and these traits allow us to constrain the runtime type. There is an alternative "additive" approach that I still plan to prototype but I do think this is pretty slick and relatively intuitive.

**Related issue**: 

**Type of change**: feature

**Development Phase**: proposal &  implementation

**Release Notes**
Add RequireAsyncReset and RequireSyncReset traits for setting type of implicit reset

### DIFF
--- a/chiselFrontend/src/main/scala/chisel3/RawModule.scala
+++ b/chiselFrontend/src/main/scala/chisel3/RawModule.scala
@@ -134,6 +134,14 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
   }
 }
 
+trait RequireAsyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: AsyncReset = AsyncReset()
+}
+
+trait RequireSyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: Bool = Bool()
+}
+
 /** Abstract base class for Modules, which behave much like Verilog modules.
   * These may contain both logic and state which are written in the Module
   * body (constructor).
@@ -145,10 +153,12 @@ abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
     extends RawModule {
   // Implicit clock and reset pins
   val clock: Clock = IO(Input(Clock()))
-  val reset: Reset = {
+  val reset: Reset = IO(Input(mkReset))
+
+  private[chisel3] def mkReset: Reset = {
     // Top module and compatibility mode use Bool for reset
     val inferReset = _parent.isDefined && moduleCompileOptions.inferModuleReset
-    IO(Input(if (inferReset) Reset() else Bool()))
+    if (inferReset) Reset() else Bool()
   }
 
   // Setup ClockAndReset
--- a/src/test/scala/chiselTests/ResetSpec.scala
+++ b/src/test/scala/chiselTests/ResetSpec.scala
@@ -68,4 +68,28 @@ class ResetSpec extends ChiselFlatSpec {
     })
     async should include ("always @(posedge clk or posedge rst)")
   }
+
+  behavior of "Users"
+
+  they should "be able to force implicit reset to be synchronous" in {
+    val fir = generateFirrtl(new MultiIOModule with RequireSyncReset {
+      reset shouldBe a [Bool]
+    })
+    fir should include ("input reset : UInt<1>")
+  }
+
+  they should "be able to force implicit reset to be asynchronous" in {
+    val fir = generateFirrtl(new MultiIOModule with RequireAsyncReset {
+      reset shouldBe an [AsyncReset]
+    })
+    fir should include ("input reset : AsyncReset")
+  }
+
+  "Chisel" should "error if sync and async modules are nested" in {
+    a [ChiselException] shouldBe thrownBy {
+      elaborate(new MultiIOModule with RequireAsyncReset {
+        val mod = Module(new MultiIOModule with RequireSyncReset)
+      })
+    }
+  }
 }


chiselFrontend/src/main/scala/chisel3/RawModule.scala
```diff
@@ -145,10 +153,12 @@ abstract class MultiIOModule(implicit moduleCompileOptions: CompileOptions)
...
   // Implicit clock and reset pins
   val clock: Clock = IO(Input(Clock()))
-  val reset: Reset = {
+  val reset: Reset = IO(Input(mkReset))
+
+  private[chisel3] def mkReset: Reset = {
     // Top module and compatibility mode use Bool for reset
     val inferReset = _parent.isDefined && moduleCompileOptions.inferModuleReset
-    IO(Input(if (inferReset) Reset() else Bool()))
+    if (inferReset) Reset() else Bool()
```
mwachs5: but this still does not allow the top module to have an AsyncReset?

mwachs5: oh NVM I see, this is just the default, and then by using the mixin you override the default. Nice.

chiselFrontend/src/main/scala/chisel3/RawModule.scala
```diff
@@ -134,6 +134,14 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
...
 }
 
+trait RequireAsyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: AsyncReset = AsyncReset()
+}
+
+trait RequireSyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: Bool = Bool()
+}
+
```
mwachs5: Is there no possibility for something like:

```scala
trait RequireSpecificResetType extends MultiIOModule {
  def resetTypeGen: () => Reset
  override private[chisel3] def mkReset: Reset = resetTypeGen()
}
```

This would later let someone do something like:

```scala
val MySpecifiedResetTypeModule(useAsyncReset: Bool) extends Module with RequireSpecificResetType {
  override def resetTypeGen() = if (useAsyncReset) AsyncReset() else Bool()
  ...
}
```

chiselFrontend/src/main/scala/chisel3/RawModule.scala
```diff
@@ -134,6 +134,14 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
   }
 }
 
+trait RequireAsyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: AsyncReset = AsyncReset()
```
seldridge: Should this be `final`?

Or: what should happen for `new Foo with RequireAsyncReset with RequireSyncReset`?

chiselFrontend/src/main/scala/chisel3/RawModule.scala
```diff
@@ -134,6 +134,14 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
...
+}
+
+trait RequireSyncReset extends MultiIOModule {
+  override private[chisel3] def mkReset: Bool = Bool()
```
seldridge: Should this be `final`?

chiselFrontend/src/main/scala/chisel3/RawModule.scala
jackkoenig: At the moment, you'll get this error:
```
error] /Users/jack/work/chisel3-control-reset-type/src/main/scala/Jack.scala:7:7: overriding method mkReset in trait RequireAsyncReset of type => chisel3.AsyncReset;
[error]  method mkReset in trait RequireSyncReset of type => chisel3.Bool has incompatible type
[error] class MyModule extends Module with MyTrait with RequireAsyncReset with RequireSyncReset {
[error]       ^
```
If I make them both `final`, you get
```
[error] /Users/jack/work/chisel3-control-reset-type/src/main/scala/Jack.scala:7:7: overriding method mkReset in trait RequireAsyncReset of type => chisel3.AsyncReset;
[error]  method mkReset in trait RequireSyncReset of type => chisel3.Bool cannot override final member
[error] class MyModule extends Module with MyTrait with RequireAsyncReset with RequireSyncReset {
[error]       ^
```
I would argue the formal (ie. without `final`) is actually marginally better

jackkoenig:
@seldridge I don't know if you saw @mwachs5's comments, but she (essentially) requested the ability to override `mkReset`, then one could write something like this (her example):
```scala
val MySpecifiedResetTypeModule(useAsyncReset: Bool) extends Module with RequireSpecificResetType {
  override def resetTypeGen() = if (useAsyncReset) AsyncReset() else Bool()
  ...
}
```
Details and names aside, there is a case to be had that `mkReset` or an equivalent should be `protected` and thus implementable by the user if they so choose. I don't love it because it's a little unintuitive that you can't just `override val reset` instead, but we can just document why this is.

chiselFrontend/src/main/scala/chisel3/RawModule.scala
seldridge: Ah, I see. You get mutual exclusion through the type system. I think that's fine.

chiselFrontend/src/main/scala/chisel3/RawModule.scala
seldridge: Would a companion object be an alternative that would work for you, @mwachs5?

```scala
object Foo { 
  def apply(useAsyncReset: Boolean) = useAsyncReset match {
    case true  => new Foo with RequireAsyncReset
    case false => new Foo with RequireSyncReset
  }
}
```

Barring that being insufficient, I think it's reasonable (as @jackkoenig suggests) to open up `mkReset` as `protected`.

mwachs5: Hm, that could work, will try it out

mwachs5: So trying this out it's still a little clunky. Specifically rocket-chip generation flow needs a top level Module and a Config class. So if i want my top level module fixed but the Config class to have information about the reset type for the top level, this is hard to express with a companion object or a fixed reset scheme. I guess i can change the top level module itself but this is rather unnatural.

mwachs5: Yes I really think this is rather limiting. Can I request to have the `mkReset` be protected?

jackkoenig:
Adding DO NOT MERGE while we discuss @mwachs5's use case

mwachs5:
> Adding DO NOT MERGE while we discuss @mwachs5's use case

Basically, my higher level use case is "allow a module to set its own destiny and decide what reset type it should have, at runtime, without relying on the instantiator". Keeping in mind that modules may be top level or may be instantiated in numerous contexts. relying on "the instantiator" requires making sure that every instantiator is "doing the right thing" to decide on the reset type, whereas just letting the module decide for itself is a much more intuitive interface from a user's perspective.

As a concrete example, imagine a component which would like to be fully AsyncReset or SyncReset given a certain runtime parameter. That component can be instantiated in multiple testharnesses / prototyping platforms / larger SoC contexts, many of which the designer of the component won't even be aware of. The current API requires that every caller now can't directly instantiate the Module, they need to be told/aware of the parameter which decides the component's reset scheme and make sure they use it. 

If it was under the component's control, there would be a runtime error if the instantiator drives the reset type incorrectly, but that may be better than the silent error of changing the component unintentionally by driving its reset with the wrong type, 

There is the clunkier issue of requiring even in correct usage every callsite for the module to consider this parameter/remembering to use the companion object. This gets more complicated with inheritance -- what if we want the parameter to be considered in the component's abstract base class that gets widely used? Now we have to fan out companion objects for each of those subclasses as well, because they have to concretely call the right superclass / `with Require...Reset` method. Where we could have just done the right thing in the shared base class in a single line.

seldridge:
This may warrant a larger discussion. 

I think there's two separable issues here:

1. There's no way to set the top module reset
1. There's a use case for setting a specific non-top module's reset

I'm fine with this PR as a fix for (1), but I'm not sure if (2) has to be fixed here. I think the broader issue of (2) is that the module hierarchy for all children of `RawModule` currently includes an abstract reset member IO. It would seem to me that setting the reset type should be something done via Scala's type system and not via passing parameters that override methods or trait mix-ins.

I.e., I think the following construction is the most logical:

```scala
class Foo extends RawModule {
  val clock: Clock      = IO(Input(Clock))
  val reset: AsyncReset = IO(Input(AsyncReset))
  
  withClockAndReset(clock, reset) {
    // ...
  }
}
```

mwachs5:
I agree that there are two issues at play here, and I am OK if this PR only addresses the first one. So my comments below are only about the first one. 

This PR as is doesn't really address the first use case, because the current implementation requires me to instantiate different top modules, not actually allow me to specify the reset type of my top module.

Our use case is to add Async Reset functionality to an existing code base where users are expecting to instantiate a variety of modules as the top level module. Modules which used to be the top level and were sync reset, we would now like to be (optionally) async reset. With this PR as is, *any* module we used to have as a top level in our chisel generation, we now have to create two classes for it -- the sync reset and async reset one. Or we have to change the entire top level Generator flow to actually have the companion object for the top level module solely to add the right `with Require...Reset` trait.

mwachs5:
To the second point... you *seem* to be assuming that `Bool` reset is more "normal" than `AsyncReset`. If someone wants to enforce `AsyncReset` they have to turn their module into a `RawModule` (wheras they can turn off `CompileOptions.inferResets` to get Bool resets I think...?) Again, retrofitting this onto an existing code base is difficult... things that inherit from the RawModule now have to put everything inside `withClockAndReset` blocks.

I hypothesize that now that you have Async reset support in Chisel3, people may start considering the AsyncReset as more "normal".

seldridge:
It's more that I think there are assumptions in the existing module hierarchy that make describing different reset types difficult. I'm then questioning if `RawModule` is a "better" abstraction for all modules of all three reset types.

mwachs5:
Anyway, I guess I am fine with merging this PR as-is as it does add necessary functionality. Again I don't think forcing any module (including all its mixins, subchildren, etc) that wants to be async reset to be rewritten as a `RawModule` is realistic, so this PR solves the problem except at the very top level.

jackkoenig:
Any thoughts from other @freechipsproject/chisel-lang-developers?

jackkoenig:
I think this is a nice thing for `3.3` if other devs agree

jackkoenig:
Alright I'm gonna merge this

johnsbrew:
Hi there @jackkoenig @seldridge 
I've seen that @mwachs5 's request of having mkReset set to `protected` instead of `private` has finally not been integrated.
Is there any reason for it?

My current use-case is the following:
- I need all modules in my lib to default to AsyncReset (to be able to be annotated as Preset as one would guess) when they are emitted as top-level module (and only as top level)
- However I don't want any error when integrated into a hierarchy of Synchronous Resets because the whole point of all this mess with implicit resets is to produce reset-agnostic designs ... 
- I also need that this option not to affect tester flow behaviour (remain synchronous there for compatibility with firrtl-backends such as treadle)

This cannot be done by mixing in `RequireAsyncReset` as it would error on SyncReset.
For now I am using an ugly wrapper with a simple withResetAndClock and iteration over ports for propagating IOs.
And as the wrapper needs then to be removed, I do even more ugly file operations at the end: overwrite the emited file while triming the wrapper module to get what I need.

I am currently thinking about leveraging annotations coupled with tiny custom Firrtl transform to do the job much more gracefully. 
I keep you updated on that topic and I wanted to share my general thoughts about chisel integration but as it became quite long I decided to go for another issue #1445!

jackkoenig:
Hi @johnsbrew, you're absolutely right that we need the ability to set the concrete type of the reset via Scala parameters. I'm not a fan of using inheritance to accomplish this goal (ie. making `mkReset` `protected`), especially because that would only work for the implicit reset of `Module` and `MultiIOModule`. The good news is that I've been thinking about this and I think we can accomplish it with 2 minor changes to Chisel in a way that is generic to all components of type `Reset`:
1. Make it legal to drive `Bool` and `AsyncReset` with `Reset`. This has been legal in FIRRTL since https://github.com/freechipsproject/firrtl/pull/1396 and it is a bug that I forgot to make it legal in Chisel
2. No longer default the top-level implicit `reset` to `Bool`. This is done for backwards compatibility reasons, but because the behavior of unification-style reset inference _defaults to Bool if there are no concrete constraints_, the behavior is maintained even if the type is now inferable

Now, you're probably wondering: "How do these two changes fulfill my use case?" It would make it possible to constrain the type of implicit reset merely by using it to drive a wire (even if unused) of a concrete reset type, eg.
```scala
// Constrain type of reset:
val constraint = if (useAsyncReset) AsyncReset() else Bool()
WireDefault(constraint, reset) // alternatively: Wire(constraint) := reset
```
What I really like about this is that unlike overriding `mkReset` it works for any arbitrary component of type `Reset`:
```scala
val myWireReset = Wire(Reset())
val constraint = if (useAsyncReset) AsyncReset() else Bool()
WireDefault(constraint, myWireReset)
```

Issue #1362, opened 2020-02-26
Current Implementation of Counter makes it impossible to create a 32bit or larger counter


claford-v-lawrence:
**Type of issue**: feature request 

**Impact**: API modification

**Development Phase**:  proposal

**Other information**
 
I submitted a PR

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
val x = Module(new Counter(4294967296L))
val x = Module(new Counter(4294967296))
these doesn't work

**What is the current behavior?**
Counter does not work with integers other than Int

**What is the expected behavior?**
Counter should work with long etc

**Please tell us about your environment:**

**What is the use case for changing the behavior?**
enabling all Counters to be able to handle large ints and longs without breaking backward compatibility

aswaterman:
This seems desirable, but would be nice to figure out how to do it without breaking API compatibility (w.r.t. `Counter.n`).

albert-magyar:
I imagine the easiest non-API breaking way to do it would be with a rename of the actual internal BigInt counter. A type parameter and the `Numeric` type class would be cool, but that is going to end up getting exposed some way or another.

Hilariously, this may be binary-compatible.
```scala
 class Counter(val max: BigInt) {
   def this(intMax: Int) {
     this(BigInt(intMax)) // or just rely on implicits
   }

   def n: Int = {
     val truncated = max.toInt
     val msg = "Viewing Counter.n as Int caused an overflow. Use Counter.max!"
     assert(BigInt(truncated) == max, msg)
     truncated
   }
  ...
  // change all remaining refs to 'n' to use 'max' in inc
```

albert-magyar:
@claford-v-lawrence, I think it would be good to re-open your PR. Maybe you could incorporate some adapters to smooth the API change?

claford-v-lawrence:
Sure. I'll keep digging if there is a better way to handle the changes.

albert-magyar:
Great! If it's decided that it's necessary to keep around `n` as an `Int` in the short term, I think something more complete version of the sketch above could work, which refactors the counter around using a `BigInt` but preserves the API by keeping `n` around as a `def` of type `Int`.

This pushes some errors (calling `.n` on a counter that would not fit in an `Int`) to runtime, though.

claford-v-lawrence:
Thanks for the help. I'm using some other work around to avoid this problem :)

PR #1363, opened 2020-02-26
make the Counter util use BigInt to create 32bit or larger counters

claford-v-lawrence:
**Related issue**: #1362

**Release Notes**


aswaterman:
It seems impossible to do this without breaking source compatibility because of the type of `Counter.n`, but the source incompatibilities could be minimized by defining an alternate constructor and alternate companion-object apply method that accept `Int`.

claford-v-lawrence:
> It seems impossible to do this without breaking source compatibility because of the type of `Counter.n`, but the source incompatibilities could be minimized by defining an alternate constructor and alternate companion-object apply method that accept `Int`.

I'm afraid I don't follow, could you give me an example as to when the Counter.n would cause an incompatibility issue? I'm running this version in my project so it would be very kind of you to tell me when it does work

aswaterman:
I mean that existing code that access the `n` field of a counter and expects it to be of type `Int` will break.

claford-v-lawrence:
Sure, but I was thinking that unless you're explicitly type checking `Counter.n` is of type Int, in which case it will break. Otherwise the arithmetics doesn't break, right?

aswaterman:
Arithmetic itself is OK, but passing it to a function checks the type, e.g.

```
scala> def foo(x: Int) = x
foo: (x: Int)Int

scala> foo(BigInt(1))
<console>:13: error: type mismatch;
 found   : scala.math.BigInt
 required: Int
       foo(BigInt(1))
```

claford-v-lawrence:
Sure, that makes sense. I wasn't really thinking about that. Thanks a lot for your help.

seldridge:
Just to clarify, I don't think there's necessarily a need for you to be closed (and I don't think Andrew was advocating for that), just that the PR may want to explore what Andrew is talking about to minimize backwards incompatible breakages.

APIs can and do break, just there's an effort to try to be disciplined about it. The 3.3 merge window is basically closed, but we can aim to get something like this, with any breakages, in for

aswaterman:
Yeah, what @seldridge said.  `Counter.n` seems like a tenable API breakage, kind of like what we did for `Mem` a while back.  Primarily, I was trying to point out that we can obviate the API breakages for the constructor and companion object.

claford-v-lawrence:
@seldridge @aswaterman @albert-magyar I reopened the PR, let's think of a better way to handle the API change. I only closed it because I think it requires a significant amount of change to the existing API that I didn't really think through, but maybe we could lessen the impact.

aswaterman:
I guess we should wait for others to opine on the acceptability of changing `n`'s type to `BigInt`.

azidar:
I think we can use @albert-magyar 's suggestion on the accompanying issue to keep it backwards source compatible. If so, this can be 3.3.x.

claford-v-lawrence:
Here is a quick thought.
Still use the companion object for counter, but have another class for BigCounter
The file would look like this
```scala
class Counter(val n: Int)
class BigCounter(val n: BigInt)
object Counter {
def apply(n: Int): Counter = Counter(n)
def apply(n: BigInt): BigCounter = BigCounter(n)
}
```
This should be easy to implement ( albeit a little dirty ) and keep backward compatibility.

PR #1367, opened 2020-03-03
Java API Documents Linking

seldridge:
Three things:

1. This appends to the build.sbt apiMappings so that linking against Java API docs (for the appropriate version of Java) works. I reuse the same build.sbt from Scala upstream.
1. This shows an example of this working to link to a Java exception
1. This shows an example of this working to link to Scala main API docs (which already worked before this PR)

Example:

chisel-scaladoc

This then links to:

chisel-javadoc

**Related issue**: None

**Release Notes**

- Scaladoc now can link against Javadoc


seldridge:
Err... Slightly better way to do this with a plugin... Will rebase.

PR #1377, opened 2020-03-13
Remove toNamed (and friends) deprecation.

ucbjrl:
**Related issue**: freechipsproject/firrtl#1449

**Release Notes**
Remove the deprecation of `Named` until we have time to sort out the issues with de-duplication of `Target`s


ucbjrl:
As with the firrtl version of this PR, I've left the old deprecation warnings as comments, but bumped the affected versions.

chiselFrontend/src/main/scala/chisel3/Module.scala
```diff
@@ -258,7 +258,7 @@ package experimental {
       *
       * @note Should not be called until circuit elaboration is complete
       */
-    @deprecated("toNamed API is deprecated -- use toTarget instead", "3.2")
+    //@deprecated(toNamed API is deprecated -- use toTarget instead", "3.3")
```
jackkoenig: Did you mean to delete the leading double quotes?

ucbjrl: Nope. Faulty regex. Thanks @jackkoenig.

PR #931, opened 2018-11-09
Trim Stack Trace

albertchen-sifive:
**Related issue**: 
https://github.com/freechipsproject/chisel3/issues/845
This is what the trimmed stack trace looks like (using the example from the issue). I only trimmed the top and bottom of the stack. Not sure about how the middle part should be filtered.
```
[info] [0.003] Elaborating design...
[error] chisel3.core.Binding$ExpectedChiselTypeException: reg type 'chisel3.core.UInt@3400978d' must be a Chisel type, not hardware
[error]         ...
[error]         at NeedChiselType.$anonfun$delays$1(ChiselExceptionTest.scala:8)
[error]         at scala.collection.generic.GenTraversableFactory.fill(GenTraversableFactory.scala:90)
[error]         at NeedChiselType.<init>(ChiselExceptionTest.scala:8)
[error]         at NeedChiselTypeTop$.$anonfun$code$1(ChiselExceptionTest.scala:17)
[error]         at chisel3.core.Module$.do_apply(Module.scala:51)
[error]         at chisel3.Driver$.$anonfun$elaborate$1(Driver.scala:94)
[error]         at chisel3.internal.Builder$.$anonfun$build$2(Builder.scala:317)
[error]         at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
[error]         at chisel3.internal.Builder$.$anonfun$build$1(Builder.scala:315)
[error]         at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
[error]         at chisel3.internal.Builder$.build(Builder.scala:315)
[error]         at chisel3.Driver$.elaborate(Driver.scala:94)
[error]         at chisel3.Driver$.execute(Driver.scala:185)
[error]         at chisel3.Driver$.execute(Driver.scala:264)
[error]         at NeedChiselTypeTop$.delayedEndpoint$NeedChiselTypeTop$1(ChiselExceptionTest.scala:17)
[error]         at NeedChiselTypeTop$delayedInit$body.apply(ChiselExceptionTest.scala:16)
[error]         at scala.Function0.apply$mcV$sp(Function0.scala:34)
[error]         at scala.Function0.apply$mcV$sp$(Function0.scala:34)
[error]         at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
[error]         at scala.App.$anonfun$main$1$adapted(App.scala:76)
[error]         at scala.collection.immutable.List.foreach(List.scala:389)
[...]

PR #1382, opened 2020-03-21
Set StageError cause in ChiselStage

seldridge:
Change `ChiselStage` to set the cause of a `StageError` due to a `ChiselException` to be the `ChiselException`. This should fix issues where libraries that want to inspect the stack trace can do so. Previously, this would do stack trace trimming and drop a no-cause `StageError`.

**Related issue**: https://github.com/freechipsproject/chisel3/issues/1114#issuecomment-601971550


jackkoenig:
Looks good to me as well. One thing I was wondering, instead of `println`ing the trimmed stack trace ourselves, what would happen if we override some of the methods in `Throwable`? https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html

Perhaps then we can use normal `printStackTrace()` for both and it will make it easier for custom Stages to compose with this

ucbjrl:
Unfortunately, this by itself, isn't sufficient. `Driver.execute` catches `StageError` exceptions and returns an error:
```
    val annosx = try {
      phases.foldLeft(annos)( (a, p) => p.transform(a) )
    } catch {
      /* ChiselStage and FirrtlStage can throw StageError. Since Driver is not a StageMain, it cannot catch these. While
       * Driver is deprecated and removed in 3.2.1+, the Driver catches all errors.
       */
      case e: StageError => annos
    }

    view[ChiselExecutionResult](annosx)
```
which eventually gets converted into another zero information exception in `chisel3.iotesters.setupTreadleBackend`:
```
    chisel3.Driver.execute(optionsManager, dutGen) match {
      case ChiselExecutionSuccess(Some(circuit), _, Some(firrtlExecutionResult)) =>
        ...
      case _ =>
        throw new Exception("Problem with compilation")
    }
```

jackkoenig:
@ucbjrl I think we should reconsider how we compose things and update those places to not catch Errors. As described in the [Java documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html): "An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch."

ucbjrl:
Perhaps it's time to revisit #361?

seldridge:
Good catch, @ucbjrl.

I expect the `Driver` shouldn't be catching anything. In 3.1.5 it didn't... This fails some `DriverSpec` tests, though (checking what's going on).

Hot take: the sooner we can remove `Driver`, the sooner we can move to a unified exception model...

seldridge:
Actually, @ucbjrl: I think what's happening is the correct behavior of stack trace trimming:

1. Elaboration fails and `ChiselStage` will print a verbose trimmed stack trace on stdout
2. A `StageError` gets thrown
3. The `Driver` converts this to a `ChiselExecutionFailure`
4. `iotesters` converts this to an information-less exception

Even though (4) happens, the information is *still provided to the user*. This is, as far as I can tell, how #931 was intended to work. 

The contract of `StageError` is that the thrower is providing all information to the user to debug the problem (here, a trimmed stack trace on stdout). Hence, `StageError` may legally provide zero information on what is going on because the thrower has already provided that information. This is likely an anti-pattern.

@jackkoenig has suggested that stack trace trimming should instead directly manipulate the stack trace. This may be a better approach. Perhaps we should implement something like this instead?

ucbjrl:
Right, the problem is the failure isn't available programmatically (without some contortions). Chisel is a library and may not always be the top application. One could argue that we shouldn't be printing errors without some negotiation with/advice from  the enclosing application.

jackkoenig:
Printing generally isn't great, imagine if there's bug in `MIDAS` Chisel code that isn't supposed to be exposed to the user, they can catch and report a better user error, but Chisel is still going to print the stack trace as if the Chisel code itself is user code.

I think there's general agreement about finding a more composable solution, the real question is if we should do this for `3.3` or perhaps do this for `3.4`.

seldridge:
Hot take :two:: We're late enough on 3.3, let's just fix it. (I volunteer...)

PR #1383, opened 2020-03-23
Add NoChiselNamePrefix to ignore instances in @chiselName

jackkoenig:
Add trait chisel3.experimental.NoChiselNamePrefix which causes
@chiselName to skip naming of the instance effectively preventing it
from prefixing any vals inside the instance. It can be applied to
classes such that all instances of that class have this property, or to
individual instances (via creating an anonymous class inline).

Also add basic ScalaDoc for NoChiselNamePrefix and chiselName.

Resolves https://github.com/freechipsproject/chisel3/issues/1371

This may not be the best way to solve the issue, but it works well enough and it's an experimental API such that if we find a better solution we can change it.

**Related issue**: https://github.com/freechipsproject/chisel3/issues/1371

**Release Notes**

Add trait `chisel3.experimental.NoChiselNamePrefix` for excluding instances of `@chiselName` annotated classes from naming. This prevents the `val` name of the instance from prefixing the names components *inside* the object.


chiselFrontend/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -71,7 +71,62 @@ package object experimental {  // scalastyle:ignore object.name
...
+    *   val io = IO(new Bundle {
+    *     val in = Input(UInt(8.W))
+    *     val out = Output(UInt(8.W))
+    *   })
+    *   def createReg(): Unit = {
+    *     // @chiselName allows Chisel to name this Reg
+    *     val myReg = RegInit(io.in)
+    *     io.out := myReg
+    *   }
+    * }
```
mwachs5: shouldn't this example actually call "createReg()" somewhere?

jackkoenig: Indeed we should, good catch. This is why we prefer to do examples using `mdoc` in the website documentation, but I figured there should be at least one example in the ScalaDoc

PR #1384, opened 2020-03-23
No more compile internal

jackkoenig:
We're close to `3.3.0`, seems like a good time to do this.

Eliminate SBT compile internal stuff. This causes problems for users upstream (see https://github.com/scalacenter/bloop/issues/1138) and since the entire purpose is a bit boutique and counter to standard Scala practice, we might as well eliminate it.

Since the subprojects `coreMacros` and `chiselFrontend` will be published separately with this change, I also renamed them to more standard names:

* `coreMacros` -> `macros`, published as `chisel3-macros`
* `chiselFrontend` -> `core`, published as `chisel3-core`

**Related issue**: 

**Impact**: API modification

Kind of an API modification, it changes the way the project is structured and published. It's possible users are doing source dependencies and reaching in to the subprojects, but this would be unwise and `3.3.0` is a reasonable time to change it.

**Release Notes**
Publish chisel3 as subprojects: `chisel3-macros`, `chisel3-core`, and `chisel3`


jackkoenig:
> I do suggest a minor name change for the sub-projects to match the actual name they will be published with.

@ucbjrl I take it you mean changing the names from like `Chisel3 Macros` to `chisel3-macros`? Sure happy to do that.

ucbjrl:
Yes.

PR #1400, opened 2020-04-04
Change BundleLiteral to RecordLiteral

sequencer:
**Related issue**: 

**Release Notes**

Since RocketChip use `Record` for `AutoBunlde` and diplomatic bundles like `TLBundle`, if we need to test RocketChip with native chisel tester, we should allow `Lit` to `Record`. 
This PR changes all `BundleLiterals` to `RecordLiterals`, which enables `Lit` method for `Record`, and deprecate original `BundleLiterals`, it won't affect existing code, and enalbe peek/poke for Diplomatic Bundle in chiseltester.

### DIFF
--- a/core/src/main/scala/chisel3/experimental/package.scala
+++ b/core/src/main/scala/chisel3/experimental/package.scala
@@ -130,7 +130,7 @@ package object experimental {  // scalastyle:ignore object.name
   trait NoChiselNamePrefix
 
   object BundleLiterals {
-    implicit class AddBundleLiteralConstructor[T <: Bundle](x: T) {
+    implicit class AddBundleLiteralConstructor[T <: Record](x: T) {
       //scalastyle:off method.name
       def Lit(elems: (T => (Data, Data))*): T = {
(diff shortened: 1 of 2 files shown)


sequencer:
Yes, my intention is to make a minimal patch for this feature.
If we mainline this feature, I think the feature should be implemented at https://github.com/freechipsproject/chisel3/blob/aa419c7190b42be1e7df78d5cb7c1ca4a8bc85b6/core/src/main/scala/chisel3/Aggregate.scala#L46
But if I change code here, code compatibility could not keep unchanged, and we may need to deprecate `RecordLiterals.AddRecordLiteralConstructor.Lit` to a common used API `litValue`.
Not sure if my PR is going to refactor these.

sequencer:
Yes my intention is to PR a minimal patch to this feature.
I'm not sure if I we can remove `RecordLiterals`/`BundleLiterals`, and directly use `implicit def` for this feature, since it contains a breaking modification to codes.

sequencer:
I think this is a minimal patch now:)

PR #1406, opened 2020-04-08
Add/use findCause Exception Cause Search Utility

seldridge:
This adds a utility, `chiselTest.Util.findCause[A <: Throwable]`, that can be used to look for a polymorphic exception anywhere in the stack trace returning an `Option[A]` if found and `None` if not.

This is intended to get out ahead of some future changes to testing infrastructure to move off of `Driver` and onto `ChiselStage` where exceptions may wind up wrapped in a `StageError`.

Two tests are migrated to use this (as these fail when migrating).

**Related issue**: None

**Type of change**: feature request | other enhancement

**Release Notes**

None.

### DIFF
--- a/src/test/scala/chiselTests/ChiselSpec.scala
+++ b/src/test/scala/chiselTests/ChiselSpec.scala
@@ -11,6 +11,7 @@ import firrtl.{AnnotationSeq, CommonOptions, ExecutionOptionsManager, FirrtlExec
 import firrtl.util.BackendCompilationUtilities
 import java.io.ByteArrayOutputStream
 import java.security.Permission
+import scala.reflect.ClassTag
 
 /** Common utility functions for Chisel unit tests. */
 trait ChiselRunners extends Assertions with BackendCompilationUtilities {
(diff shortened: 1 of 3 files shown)


chick:
- Shouldn't this be on the main side of things so it can be used by others
  - And if it is on the main side maybe it should be in firrtl to be accessible everywhere.

seldridge:
@chick: Yes. Good catch. I didn't realize that `firrtl.testutils.Utils` now exists. That can be moved over (and I think that makes `chiselTests.Utils` obsolete?). 

Are we publishing that? Can I then remove the copied over code in the `chiselTests.Utils` object?

chick:
I don't think the firrtl.testutils.Utils is published as it is under the sr/test (right @ucbjrl ?)
But I think it should be move to src/main.
I'm not sure what that means for this PR. Are you trying to get this into RC-2?

ucbjrl:
I believe @chick is correct. We don't currently publish tests.

seldridge:
No, this doesn't need to go in for RC-2. This is useful for 3.4 changes that start removing `Driver`.

chick:
I've created a [Firrtl PR 1503](https://github.com/freechipsproject/firrtl/pull/1503) to firrtl that exposes the firrtl test Utils (they are indeed identical to the chisel3) ones. You may want to move this PR on top of that

azidar:
Add to firrtl test package that we publish? Otherwise, its good.

PR #1408, opened 2020-04-10
Make Counter emit valid FIRRTL

jackkoenig:
Remove var from object Counter.apply, using a Wire instead. Also improve
some ScalaDoc and the class Counter require message.

**EDIT**: At @seldridge's suggestion, here is a more minimal fix that doesn't mess with the API of class `Counter.inc()`

~~The one thing I want to highlight:~~
```diff
-// This if-guard maintains older behavior for wrap when n == 1
-val wrap = if (n > 1) c.inc(cond) else cond
```
~~As the comment says, if we did just `c.inc(cond)`, it changes the behavior for `n == 1` to return `true.B` rather than `cond`.~~

**Related issue**: 

**Type of change**: bug fix

**Release Notes**
* Fix bug where `chisel3.util.Counter` emitted illegal FIRRTL


src/main/scala/chisel3/util/Counter.scala
```diff
@@ -62,13 +65,8 @@ object Counter
   @chiselName
   def apply(cond: Bool, n: Int): (UInt, Bool) = {
     val c = new Counter(n)
-    // Note this use of var is generally frowned upon! Don't use this as an example.
-    // This is done because we wanted the hardware generated by c.inc() to be wrapped
-    // in a when statement, but still needed to refer to the final value returned by
-    // c.inc() so we could return cond && wrap. Unless you really, really know what
-    // you are doing, IGNORE THIS CODE!!!!!
-    var wrap: Bool = null
```
aswaterman, line 70: mea culpa, mea culpa, mea maxima culpa

jackkoenig: To be fair, there are 2 other issues:
* https://github.com/freechipsproject/chisel3/issues/1409
* https://github.com/freechipsproject/firrtl/issues/1505

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -25,16 +25,17 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   */
 @chiselName
 class Counter(val n: Int) {
-  require(n >= 0)
+  require(n >= 0, "Counter must count to value greater than 0!")
```
colinschmidt, line 28: This text doesn't match the require. The require says "greater than or equal to", but the text just says "greater than"

jackkoenig: Good point, I misread the require, will fix

jackkoenig:
This fix is also backportable now

PR #1405, opened 2020-04-07
Switch to HowToSerialize for Emission

seldridge:
This removes the `chisel3.stage.phases.Emitter` and moves its emission logic into the `ChiselCircuitAnnotation`. This uses the new `HowToSerialize` mix-in provided by https://github.com/freechipsproject/firrtl/pull/1277. 

A consequence of this is that there is no longer a need for generating `DeletedAnnotation`s and for introspection of `DeletedAnnotation`s to determine an execution result or a view of some sequence of annotations.

**Related issue**: Depends on https://github.com/freechipsproject/firrtl/pull/1277

**Type of change**: other enhancement

**Impact**: API modification

**Development Phase**: implementation

**Release Notes**

PR #1412, opened 2020-04-14
Have phases.Emitter name phases.Convert as a dependent

davidbiancolin:
This must be prefaced with i'm not at all sure this is the correct solution...

That said, under the current ordering constraints, the phase manager might elect to schedule the convert phase before the emitter phase, which removes the chisel circuit annotation before it can be processed by the emitter. Really i'm using dependent here more in "make-sure-i-run-before" sense, i'm not sure that's the right idea.

The underlying problem here is that certain phases need to run at some point during the lifetime of specific annotations and often it's not clear what that lifetime is, short of reading each phase. If an annotation is going to be used by multiple consumer passes, some of which might be outside of chisel's control, it may make sense to dedicate an entire pass to the removal of certain annotations if we choose to remove them at all (ie. RemoveChiselCircuitAnnotationPhase) -- after all, another possible fix here is simply not to delete the annotation (and potentially mark it so it is not serialized). 

**Release Notes**

Have the emitter Phase name the Convert phase as a dependent.

### DIFF
--- a/src/main/scala/chisel3/stage/phases/Emitter.scala
+++ b/src/main/scala/chisel3/stage/phases/Emitter.scala
@@ -30,6 +30,8 @@ class Emitter extends Phase {
          Dependency[AddImplicitOutputAnnotationFile],
          Dependency[MaybeAspectPhase] )
 
+  override val dependents = Seq(Dependency[Convert])
+
   override def invalidates(phase: Phase): Boolean = phase match {
     case _: Elaborate => true
     case _ => false


seldridge:
This should be fine as a temporary fix for 3.2.x. All this should be fixed up after the "How To Serialize" series ([firrtl#1277](https://github.com/freechipsproject/firrtl/pull/1277) and [chisel#1405](https://github.com/freechipsproject/chisel3/pull/1405)) is merged. Currently this is slated as 3.4/1.4, so a 3.2.x fix makes sense.

#1405 makes the change where `Convert` will no longer delete the annotation. That annotation is then serialized to a file automatically when the `Stage` finishes.

ucbjrl:
@davidbiancolin, the CLA bot doesn't think you've signed the CLA.

PR #1414, opened 2020-04-14
Add tests for async reset regs of non-UInt types

jackkoenig:
Add some tests that would've caught an embarrassing bug

Will fail until https://github.com/freechipsproject/firrtl/pull/1523 is merged

Marked for backport, will fail because of Intervals, I'll fixup manually.

**Related issue**: 

**Type of change**: add tests

**Release Notes**

### DIFF
--- a/src/test/scala/chiselTests/AsyncResetSpec.scala
+++ b/src/test/scala/chiselTests/AsyncResetSpec.scala
@@ -191,6 +191,54 @@ class AsyncResetSpec extends ChiselFlatSpec {
     assertTesterPasses(new AsyncResetQueueTester)
   }
 
+  it should "support SInt regs" in {
+    assertTesterPasses(new BasicTester {
+      // Also check that it traces through wires
+      val initValue = Wire(SInt())
+      val reg = withReset(reset.asAsyncReset)(RegNext(initValue, 27.S))
+      initValue := -43.S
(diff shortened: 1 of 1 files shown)


src/test/scala/chiselTests/AsyncResetSpec.scala
```diff
@@ -191,6 +191,54 @@ class AsyncResetSpec extends ChiselFlatSpec {
     assertTesterPasses(new AsyncResetQueueTester)
   }
 
+  it should "support SInt regs" in {
+    assertTesterPasses(new BasicTester {
+      // Also check that it traces through wires
+      val initValue = Wire(SInt())
+      val reg = withReset(reset.asAsyncReset)(RegNext(initValue, 27.S))
```
edwardcwang, line 198: How does this specifically test asnyc reset? Is `reset` still sync reset?

jackkoenig: These tests aren't specifically testing that it's actually async reset, they assume that is tested by other tests (which it is, see `AsyncResetTester` and others above). They are just checking that non-UInt types are supported since they were tripping FIRRTL previously due to Chisel emitting SInt literals as `asSInt(UInt(...))` (and similar for the others).

PR #1417, opened 2020-04-16
Revert "Make uselessly public fields in utils private"

azidar:
At the bequest of @jackkoenig, this reverts commit c279860c36a73984cd1b7b0ac6c213e8b44a7143.

**Release Notes**

To properly name private and protected fields in `chisel3.util.Arbiter` and `chisel3.util.Decoupled` with the `@chiselName` macro annotation, these fields were made public, and thus retain a better name.

### DIFF
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -33,8 +33,8 @@ private object ArbiterCtrl {
 }
 
 abstract class LockingArbiterLike[T <: Data](gen: T, n: Int, count: Int, needsLock: Option[T => Bool]) extends Module {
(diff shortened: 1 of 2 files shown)


jackkoenig:
Should we backport this?

ucbjrl:
Re: backporting - is there any chance previously private names may now collide with user/client definitions? I would hope not, but I vaguely remember that accessibility (private vs public) did interfere with inheritance uniqueness, despite the hope that it wouldn't.

jackkoenig:
> Re: backporting - is there any chance previously private names may now collide with user/client definitions? 

Good point, yeah this would fail the binary incompatibility check so we shouldn't backport. 3.3 will be out soon so we'll just tell people to upgrade to get the fix.

PR #1418, opened 2020-04-24
Ensure that Elaborate only runs once

seldridge:
Fixes a bug where Chisel elaboration would run twice. Specifically this PR is fixing two problems:

1. The `Convert` phase deletes the `ChiselCircuitAnnotation` and therefore must invalidate `Elaborate`. While `Emitter` must not invalidate `Elaborate` because it preserves the `ChiselCircuitAnnotation`. Previously this was reversed.
2. This makes `ChiselStage` solve a smaller sub-problem that does not include `Elaborate`. This results in a better ordering that does not cause `Elaborate` to run twice.

Note: there appears to be a performance bug in the `DependencyManager` with a test case available in https://github.com/freechipsproject/firrtl/pull/1538. 

**Type of change**: bug fix

**Release Notes**

- Fix bug where elaboration runs twice


jackkoenig:
Replaced by https://github.com/chipsalliance/chisel/pull/1426

PR #1424, opened 2020-05-04
Add SourceInfo to chisel3.util gadgets

kammoh:
**Related issue**: 

**Release Notes**

Adding `SourceInfo` to `chisel3.util` utilities so that the generated Verilog/FIRRTL can be easily traced to user's design.

### DIFF
--- a/src/main/scala/chisel3/compatibility.scala
+++ b/src/main/scala/chisel3/compatibility.scala
@@ -4,6 +4,7 @@
   *  while moving to the more standard package naming convention `chisel3` (lowercase c).
   */
 import chisel3._    // required for implicit conversions.
+import chisel3.internal.sourceinfo.SourceInfo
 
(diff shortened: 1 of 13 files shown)


jackkoenig:
This is very useful, thanks @kammoh!

> The methods which return a UInt will need to use macros to add the source info/compile options.

Not just the UInts, but the Seqs and generic ones too.

Also I strongly suspect that such changes are not binary compatible which is unfortunate but is just a motivating reason to make the 1.4 release sooner rather than later!

jackkoenig:
Actually do we want to propagate the compileOptions? That makes the `chisel3.util`s preserve `import Chisel` semantics in some cases but that's not the intent. I think we just want the implicit for the source locator, and this probably necessitates new macros (that are very similar to the existing ones).

PR #1433, opened 2020-05-12
Enhance chiselName for anonymous Modules and Companion Objects

azidar:
**Related issue**: 
Related to: https://github.com/freechipsproject/chisel3/issues/646

**Type of change**: feature requestt

**Release Notes**

Enables `@chiselName` to work on (1) on companion objects, which basically just applies the annotation to all functions in the object, and (2) on `@chiselName lazy val module = new LazyModuleImpl(this) {...}`


colinschmidt:
It would be cool if it could work on anonymous classes like this: https://github.com/ucb-bar/hwacha/blob/master/src/main/scala/sequencer-lane.scala#L189

PR #1434, opened 2020-05-12
Longer path info

azidar:
**Related issue**: 

**Type of change**: bug report | feature request | documentation | other enhancement

**Impact**: no functional change | API addition (no impact on existing code) | API modification

**Development Phase**: proposal |  implementation

**Release Notes**

### DIFF
--- a/core/src/main/scala/chisel3/internal/SourceInfo.scala
+++ b/core/src/main/scala/chisel3/internal/SourceInfo.scala
@@ -52,7 +52,14 @@ object SourceInfoMacro {
   def generate_source_info(c: Context): c.Tree = {
     import c.universe._
     val p = c.enclosingPosition
-    q"_root_.chisel3.internal.sourceinfo.SourceLine(${p.source.file.name}, ${p.line}, ${p.column})"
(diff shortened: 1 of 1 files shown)


mwachs5:
does this give you the full(er)  file path instead of just the Foo.scala?

seldridge:
@mwachs5: My read of this is that if `/scala/` exists in source locator, you'll get everything after that. So instead of `Foo.scala`, the info would be `packag/subpackage/Foo.scala`.

If `/scala/` doesn't exist, then you'll just get the full path.

mwachs5:
hmm, but I guess if you have some things in rocket-chip/src/main/scala/utils/Foo.scala and (hypothetical) chisel3/src/main/scala/utils/Foo.scala, both would just say "utils/Foo.scala"

seldridge:
Good point. I think I'm in favor of making the full path available to enable things like underlining the problematic code---something like https://github.com/freechipsproject/firrtl/issues/1598.

jackkoenig:
Fixed by https://github.com/chipsalliance/chisel/pull/2791

PR #1446, opened 2020-05-20
make run function more composable

debs-sifive:
Could I get this onto `3.3.x` so I can use it in rocket chip, please?

**Related issue**: https://github.com/chipsalliance/rocket-chip/issues/2467

**Release Notes**

This allows for `ChiselStage`'s `run` function to be called externally. For example, from `RocketStage`.


seldridge:
I've been thinking that `Stage.run` should likely be a protected method and not exposed publicly. Hence, I'd prefer a solution that avoids further expanding the `ChiselStage` API. However, I also want to get the underlying issue fixed quickly in Rocket Chip.

Towards this, I have an implementation of an alternative solution in https://github.com/chipsalliance/rocket-chip/pull/2481. Would that be an acceptable solution?

debs-sifive:
@seldridge your solution makes sense to me!

PR #1448, opened 2020-05-27
Improved Chisel Naming via Compiler Plugins + Prefixing

azidar:
Old text below, new description is still being worked on.

This PR replaces the functionality of `@chiselName` with two distinct, semantically distinct features: (1) a compiler plugin to find default names, (2) a user-available prefixing object.

(1) A compiler plugin which, if included in your `build.sbt` file, will search your Chisel source code for any place it finds `val <name> = <value>` where `<value>` extends Chisel's `HasId` trait, and replaces it with `val <name> = <value>.pluginName("<name>")`. This provides a default seed for the name, which is a lower priority than (and can be overwritten by) the existing `suggestName` API.

(2) A `prefix` object which can be used to append a prefix to any generated component created within its scope. For example, the wire built in `builder` will be named `foo_bar`:

```scala
def builder() = {
  val bar = Wire(3.U)
}
prefix("foo") {
  builder()
}
```

The prefix object can take either a `String` or a `Data` as its first argument to indicate the prefix.

Rocket-Chip Results (without source code changes):
 - Just auto-naming: reduced number of Chisel source locations which generated anonymous registers from 95 to 19, and anonymous wires from 19220 to 16073 (16% decrease in bad names).
- Auto-naming + prefixing of temporaries: further reduction of anonymous wires from 16073 to 10880.
- Auto-naming + auto-prefixing + added prefixing to one file (Monitor.scala) reduces 10880 -> 1395 anonymous wires.

Open questions:
 - How should we best include the compiler plugin? We can publish it to maven and have it included in a user's build.sbt, or (as you can see from the changes to chisel's build.sbt) you can depend on its source files directly if (like rocket-chip) you build chisel directly.

**Related issue**: #327

**Type of change**: feature request & documentation

**Impact**: API addition (no impact on existing code) (although its MAYBE a API modification)

**Release Notes**

TODO: Write this once all details are finalized.


davidbiancolin:
What are the justifications for making the plugin optional (aside from the fact plugins naturally want to be so)?  Would it not make more sense to mandate it in newer chisel versions so that output verilog is more likely to be consistent across different elaboration environments?

azidar:
I don't think we need it to be optional, and its probably best if it's mandatory. I'm not sure the best way to do that besides a dynamic check of sorts? I don't know whether Scala has a built-in functionality to require a user to have a certain plugin, or to find the plugin for them. I'll look into it.

jackkoenig:
>  I'm not sure the best way to do that besides a dynamic check of sorts? I don't know whether Scala has a built-in functionality to require a user to have a certain plugin, or to find the plugin for them. I'll look into it.

I don't know of the mechanism, but use of macros would error if you didn't have macros paradise which used to be implemented as a plugin. I suspect there's some mechanism there. If not, I think the Builder checking at the beginning of Chisel is reasonable.

jackkoenig:
In any case, I am of the opinion that the plugin needs to be optional in 3.3.x (if we can backport this), but we should make it mandatory in 3.4.0.

plugin/build.sbt
```diff
@@ -0,0 +1,10 @@
+name := "plugin"
+
+version := "0.1.1-SNAPSHOT"
+
+scalaVersion := "2.12.8"
+
+libraryDependencies ++= Seq(
+  "org.scala-lang" % "scala-compiler" % "2.12.8" withSources()
+)
+
```
jackkoenig, line 10: Why `2.12.8`, the rest of this repo is on `2.12.11`? Also do you know about the publishing rules for plugins? I believe they are binary incompatible at a Scala minor version level, ie. I think we need to publish this for `2.12.0-11`.

azidar:
After some initial searching, I found that there isn't a way to automatically get downstream users to have the plugin without them changing something about their build. We can either explicitly have them call [addCompilerPlugin](https://stackoverflow.com/questions/43757369/sbt-plugin-how-to-add-compiler-plugin-as-a-dependency-that-is-not-propagated-do?rq=1) or we can write an sbt [plugin](https://stackoverflow.com/questions/23485426/sbt-compiler-plugin-as-transitive-dependency).

plugin/build.sbt
azidar: Yup, I think i fixed it now.

plugin/src/main/scala/plugin/package.scala
```diff
@@ -0,0 +1,49 @@
+// See LICENSE for license details.
+
+import java.io.FileWriter
+
+package object plugin {
```
ducky64, line 5: Is this mainly a collection of debugging utils? This seems kind of generic, almost something I'd expect that we should be able to pull in from the Scala compiler code? (but I may also be wrong)

src/test/scala/chiselTests/naming/PrefixSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  property("Using String as prefix") {
+    class Test extends MultiIOModule {
+      def builder() = {
+        val wire1 = Wire(UInt(3.W))
+        val wire2 = Wire(UInt(3.W))
+        wire2
+      }
+      val x1 = prefix("x1") { builder() }
```
ducky64, line 37: Is this test just a special case of the one above, except the val name is the prefix name? Or alternatively put, an explicit version of the below? It feels kind of odd?

src/test/scala/chiselTests/naming/ValNameSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+
+package chiselTests.naming
+
+import chisel3._
+import chisel3.aop.Select
+import chisel3.experimental.{ValName, prefix}
+import chiselTests.ChiselPropSpec
+
+class ValNameSpec extends ChiselPropSpec {
+  property("Nested prefixes should work with plugin") {
```
ducky64, line 11: How does this differ from the test in PrefixSpec?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -811,7 +811,16 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
     }
 
+    val isAnonFunc = ".*\\$\\$anonfun\\$\\d+$".r
+    def getNonFuncClass(clz: Class[_]): Option[Class[_]] = {
+      clz.getName match {
+        case isAnonFunc() => getNonFuncClass(clz.getEnclosingClass)
+        case _ => Some(clz)
+      }
+    }
+
```
jackkoenig, line 821: I don't love regex, but at the same time, there's reflection all over this function so I doubt it really matters performance-wise. We could also put the 2.11 specific code in `src/main/scala-2.11` and implement the same method in `src/main/scala-2.12.` but just have it return `Some(clz)`. I don't know that that's necessary though.

jackkoenig:
The full commit history of this PR is kind of a mess. I can do some intense rebasing, or we can squash it all away and have that be the "rebase". Any opinions?

jackkoenig:
I've made some modifications to @azidar's original work, mainly:
* Audit and enhance the plugin publishing
  * It skips publishing in 2.11, and cross publishes for all minor versions of 2.12 that it can (compiler plugins must be published against minor versions)
* Add checking for recursive types so that we don't infinitely recurse in the plugin

107350qin:
why not naming for vec connect as so on?

PR #1449, opened 2020-05-29
Add scalafix linter framework

albert-magyar:
**Type of change:** enhancement

### DIFF
--- a/scalafix/rules/src/main/scala/fix/ChiselLinter.scala
+++ b/scalafix/rules/src/main/scala/fix/ChiselLinter.scala
@@ -1,0 +1,18 @@
+package fix
+
+import scalafix.v1._
+import scala.meta._
+
+case class AmbiguousLiteralExtract(pos: Position) extends Diagnostic {
+  override def position: Position = pos
+  override def message: String =
+    s"Passing an Int to .U is usually a mistake: it does not set the width but does a bit extract."
+}
+
+class ChiselLinter extends SemanticRule("ChiselLinter") {
+
+  override def fix(implicit doc: SemanticDocument): Patch = {
+    ???
+  }
+
+}
--- a/scalafix/README.md
+++ b/scalafix/README.md
@@ -1,0 +1,7 @@
+# Scalafix rules for chisel-linter
+
+To develop rule:
+```
+sbt ~tests/test
+# edit rules/src/main/scala/fix/ChiselLinter.scala
+```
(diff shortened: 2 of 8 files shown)


albert-magyar:
@jackkoenig I added the lit pattern (the rule for which is conseratively sensitive) as an example pattern-matching case to a consolidated `ChiselLinter` rule.

albert-magyar:
> I think you should add a suggestion for how to make it more clear in these cases if it’s what you really meant. Like what if I really did mean ==, or to do a bit wise index?

That is a great point. The current idea for immediately doing an extract on an anonymous literal value is to suggest `123.U()(4)` instead of `123.U(4)`.

For `==` on hardware types, I guess suggesting `x.equals(y)` would be a possibility?

albert-magyar:
@mwachs5 I added some basic suggestions. I agree that if we introduce simple linter, there should be well-defined workarounds.

scalafix/rules/src/main/scala/fix/ChiselLinter.scala
```diff
@@ -0,0 +1,41 @@
+package fix
+
+import scalafix.v1._
+import scala.meta._
+
+case class AmbiguousLiteralExtract(conv: String, valLit: Lit.Int, bitLit: Lit.Int) extends Diagnostic {
+  override def position: Position = valLit.pos
+  override def message: String =
+    s"Passing an integer to .${conv} does not set width but does a bit extract. If you really want this bit extract, use ${valLit}.${conv}()(${bitLit})"
```
albert-magyar, line 9: This suggestion depends on adding parens to the definitions of `.U` et al. I know @jackkoenig was considering doing that, since it makes it easier to protect against the immediate-bit-extract construct.

jackkoenig:
> That is a great point. The current idea for immediately doing an extract on an anonymous literal value is to suggest 123.U()(4) instead of 123.U(4).

As you @albert-magyar noted in the dev meeting today, the style to do the bit extract requires changes in https://github.com/freechipsproject/chisel3/tree/deprecate-literal-bit-extract. Note that I think if we do that change, we might as well make the bad style deprecated so the lint rule wouldn't be needed once that it removed.

Alternatively, without that change, to do the bit extract you need to write `123.U.apply(4)`

jackkoenig:
Or alternatively, we try to make the lint rule accept `(123.U)(4)`, but this is tricky because ScalaFix can't distinguish [easily] them because the AST does not keep that information around. We would have to use the source locators and manually inspect the text ourselves which seems error prone.

albert-magyar:
> Alternatively, without that change, to do the bit extract you need to write `123.U.apply(4)`

I actually like this -- I think it's even clearer than requiring the parens. I updated the suggestion.

PR #1452, opened 2020-05-29
make parameters for util modules public

albertchen-sifive:
**Related issue**: 
https://github.com/chipsalliance/rocket-chip/pull/2452#discussion_r425963100 The parameters of modules like `Queue` are useful for generating better names, but it is not possible to generate a fully unique name because parameters like `pipe` and `flow` are private.

**Release Notes**

make parameters of modules in chisel3.util public


seldridge:
Would it make sense to override the desired name in these modules directly based on these parameters?

jackkoenig:
> Would it make sense to override the desired name in these modules directly based on these parameters?

That's been my thinking. It also suggests we should provide some standard, structural serialization of arbitrary `Data` types (not hardware values). A "prettyPrinting" would be nice, and would work fine for small things, eg. `UInt(8.W)` -> `U8`, but for deep Bundles wouldn't be ideal. You could allow Bundles to override it in a way similar to overriding `desiredName` in modules, but then there's risk users could cause collisions. Perhaps worth it to get a name like `TLBundle_paramA_paramB_paramC` instead of `7fc4e38`

src/main/scala/chisel3/util/Arbiter.scala
```diff
@@ -101,7 +101,7 @@ class LockingArbiter[T <: Data](gen: T, n: Int, count: Int, needsLock: Option[T
   * }}}
   */
 @chiselName
-class RRArbiter[T <: Data](gen:T, n: Int) extends LockingRRArbiter[T](gen, n, 1)
+class RRArbiter[T <: Data](gen:T, val n: Int) extends LockingRRArbiter[T](gen, n, 1)
```
mwachs5, line 104:
```suggestion
class RRArbiter[T <: Data](val gen: T, val n: Int) extends LockingRRArbiter[T](gen, n, 1)
```

src/main/scala/chisel3/util/Arbiter.scala
```diff
@@ -117,7 +117,7 @@ class RRArbiter[T <: Data](gen:T, n: Int) extends LockingRRArbiter[T](gen, n, 1)
   * }}}
   */
 @chiselName
-class Arbiter[T <: Data](gen: T, n: Int) extends Module {
+class Arbiter[T <: Data](gen: T, val n: Int) extends Module {
```
mwachs5, line 120:
```suggestion
class Arbiter[T <: Data](val gen: T, val n: Int) extends Module {
```

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -187,8 +187,8 @@ class QueueIO[T <: Data](private val gen: T, val entries: Int) extends Bundle
 @chiselName
 class Queue[T <: Data](gen: T,
```
mwachs5, line 188:
```suggestion
class Queue[T <: Data](val gen: T,
```

src/main/scala/chisel3/util/Valid.scala
```diff
@@ -168,7 +168,7 @@ object Pipe {
   * @see [[Queue]] and the [[Queue$ Queue factory]] for actual queues
   * @see The [[ShiftRegister$ ShiftRegister factory]] to generate a pipe without a [[Valid]] interface
   */
-class Pipe[T <: Data](gen: T, latency: Int = 1)(implicit compileOptions: CompileOptions) extends Module {
+class Pipe[T <: Data](gen: T, val latency: Int = 1)(implicit compileOptions: CompileOptions) extends Module {
```
mwachs5, line 171:
```suggestion
class Pipe[T <: Data](val gen: T, val latency: Int = 1)(implicit compileOptions: CompileOptions) extends Module {
```

seldridge (approved):
lgtm!

PR #1455, opened 2020-05-30
Queue-FlipFlop-Output

scott-kim-7:
**Related issue**: 

**Release Notes**
Chisel Queue class is used for various purposes
(data buffering, cutting critical path, register slicing)

But sometimes it is not enough because it's output is MUX output.
Deep Queue could make data-critical-path.
This PR proposes to make the Queue's output flip-flop without function changing and it has the same operation timing as it used to have.
https://app.lucidchart.com/documents/edit/691dc934-c928-4f8a-8153-91eb09e6571f/0_0

### DIFF
--- a/src/main/scala/chisel3/util/Decoupled.scala
+++ b/src/main/scala/chisel3/util/Decoupled.scala
@@ -206,54 +206,84 @@ class Queue[T <: Data](gen: T,
 
   val io = IO(new QueueIO(genType, entries))
 
-  val ram = Mem(entries, genType)
-  val enq_ptr = Counter(entries)
-  val deq_ptr = Counter(entries)
-  val maybe_full = RegInit(false.B)
+  private val withFlipFlopOutput = (entries >= 2) && (!pipe) && (!flow)
+  private val bodyEntries = if( withFlipFlopOutput ) (entries-1) else (entries)
 
-  val ptr_match = enq_ptr.value === deq_ptr.value
-  val empty = ptr_match && !maybe_full
-  val full = ptr_match && maybe_full
-  val do_enq = WireDefault(io.enq.fire())
-  val do_deq = WireDefault(io.deq.fire())
+  private val bodyIO = Wire(new QueueIO(genType, bodyEntries))
 
-  when (do_enq) {
-    ram(enq_ptr.value) := io.enq.bits
+  private val ram = Mem(bodyEntries, genType)
+  private val enq_ptr = Counter(bodyEntries)
+  private val deq_ptr = Counter(bodyEntries)
+  private val bodyMaybe_full = RegInit(false.B)
+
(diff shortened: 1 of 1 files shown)


aswaterman:
Unconditionally changing the timing of Queue is going to cause new timing problems to manifest in other people's code.  Note there's now a Mux on the input to ff_bits.  For the common case of a 2-entry queue, the total delay from enq.bits to deq.bits is the same, but is now on the input side.

You could alternatively just transform your own code in the problematic places: replacing `Queue(n, r)` with `Queue(1, Queue(n-1, r, flow=true), pipe=true)` does what you want.

We could conceivably add a new argument or method to `object Queue` that emits the flow-queue-followed-by-pipe-queue for convenience.

scott-kim-7:
Basically you're right.
Queue(1, Queue(n-1, r, flow=true), pipe=true) is what I wanted.
Thank you.

This PR's intention was to improve legacy code's performance up without modifying every repository. And hopefully, it can help third parties to get fewer critical-paths from Queue.
That's why I modified the Queue directly, not as a utility class.
(At first, I made it optional and the default disabled. But I realized it is not appropriate to my intention.)

This PR removes Nx1 output Mux by putting a 2x1 input Mux.
As you mentioned, in the 2-depth case there is nothing better.
But in the other case, we can get more chances to reduce critical-paths.

So I'd like to propose supporting Queue(1, Queue(n-1, r, flow=true), pipe=true) in object Queue.apply with the following condition.
- private val withFlipFlopOutput = (entries >= **3**) && (!pipe) && (!flow)

or
- private val withFlipFlopOutput = (entries >= **3**) && (!pipe) && (!flow) **&& (!muxOutput) // additional argument muxOutput, default=false**

What do you think?

albert-magyar:
> This PR's intention was to improve legacy code's performance up without modifying every repository. And hopefully, it can help third parties to get fewer critical-paths from Queue.
> That's why I modified the Queue directly, not as a utility class.
> (At first, I made it optional and the default disabled. But I realized it is not appropriate to my intention.)

We generally try to avoid changing the emitted microarchitectures for existing code Chisel library functions if they were previously stable and functionally correct. One of the benefits of Chisel is that it lets you combine fine-grained control with productive programming, which does lead to a desire to have extremely predictable behavior in things like `Queue`. While getting QoR improvements "for free" via library changes might help some designs, it's hard to weigh that against what effectively amounts to designer intent, since there is always a chance that some design had the input-side valid on its critical path.

It does sound like the composed queue solution is a reasonable way to get the "efficient data to output path" behavior for new designs without too much code.

If you have a small queue that is usually empty, `ShiftQueue` from Rocket Chip is another way to optimize timing on the output side at all costs.

scott-kim-7:
It sounds like Chisel is basically putting productivity before optimization.
I agree with that.
Thank you for your review and opinions.

PR #1474, opened 2020-06-10
Builder: use LazyLogging.logger.warn to print elaboration message

ekiwi-sifive:
This is supposed to make it possible to disable the "Elaborating design..." output by setting the log level to error.

As @jackkoenig suggested in https://github.com/freechipsproject/firrtl/issues/1680#issuecomment-642186707 this uses `logger.warn` instead of `logger.info`.

I am not entirely sure how exactly `LazyLogging` determines the log level...
@jackkoenig or @seldridge am I using the logger correctly?

fixes https://github.com/freechipsproject/chisel3/issues/1473

### DIFF
--- a/core/src/main/scala/chisel3/internal/Error.scala
+++ b/core/src/main/scala/chisel3/internal/Error.scala
@@ -101,10 +101,6 @@ private[chisel3] class ErrorLog {
   def warning(m: => String): Unit =
     errors += new Warning(m, getUserLineNumber)
 
-  /** Emit an informational message */
-  def info(m: String): Unit =
-    println(new Info("[%2.3f] %s".format(elapsedTime/1e3, m), None))  // scalastyle:ignore regex
-
   /** Log a deprecation warning message */
   def deprecated(m: => String, location: Option[String]): Unit = {
     val sourceLoc = location match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -101,10 +101,6 @@ private[chisel3] class ErrorLog {
   def warning(m: => String): Unit =
     errors += new Warning(m, getUserLineNumber)
 
-  /** Emit an informational message */
-  def info(m: String): Unit =
-    println(new Info("[%2.3f] %s".format(elapsedTime/1e3, m), None))  // scalastyle:ignore regex
-
```
seldridge, line 107: This should only be deprecated and not outright removed. This should be removed in 3.5, but will have been deprecated since 3.2.2.

ekiwi-sifive: Isn't this an internal object and thus not part of the public API?

ekiwi-sifive:
> Can you add a test in `ChiselMainSpec` that checks that this is still printed out?

Yes. I will add that.

core/src/main/scala/chisel3/internal/Error.scala
seldridge: Ah, good catch. Yes, this should be fine.

Note that I think this will fail backwards compatibility checks (similar to https://github.com/freechipsproject/firrtl/pull/1642) as package private is public Java. That could either be (1) not backported, (2) waived, or (3) straight removed in 3.4.

sequencer:
Since I cannot directly push to ekiwi-sifive/chisel3, so I reopened this PR.
Closed #1670

PR #1475, opened 2020-06-15
Remove Deprecated Driver

seldridge:
Remove `chisel3.Driver` and organize tests for FIRRTL 1.4 / Chisel 3.4.

- `ChiselStage` methods are used instead of those in the `BackendCompilationUtilities` trait, e.g., for elaboration
- Any custom transforms in tests are converted to use the Dependency API
- A new utility, `extractCause[A <: Throwable : ClassTag]`, is added that rethrows a nested `A` if it exists
- The `extractCause` utility is used to convert tests that expected an exception, but now get a `StageError`.
- All usages of `chisel3.Driver` are removed
- Compatibility phases for the `Driver` are removed
- Some lingering `Driver` infrastructure (e.g., `ChiselExecutionResult`) is deprecated

**Related issue**: 

  - Closes #1406 with a better utility, `extractCause`

**Impact**: API modification

**Release Notes**

- Remove deprecated chisel3.Driver

**tl;dr**:

It's simple... We, uh, kill the Dri-Ver


seldridge:
So, it's not so simple...

For this to be merged, both testers and testers2 need to migrate to Stage/Phase and likely go full a full deprecation cycle.

I'm converting this to a draft and factoring out the pieces that can merged independently of fully removing the driver.

seldridge:
Fixed via #1730. Amazing work @sequencer!

PR #1480, opened 2020-06-18
Broken auto-clonetype on Scala 2.11.12

azidar:
This PR just contains a simple test case illustrating an example where autoclonetype is broken. The weird thing is it breaks on Scala 2.11.12 but works on Scala 2.12.11.

Edit: Looks like this is the problem: https://www.scala-lang.org/news/2.12.0/#:~:text=In%20Scala%202.11%2C%20the%20body,method%20in%20the%20enclosing%20class.

In Scala 2.11, anonymous functions were compiled to their own classes, while in Scala 2.12, they are directly compiled into the enclosing classes. This meant that checking the enclosing parent in 2.12 would work, but in 2.11 they wouldn't. This fix just looks for the first enclosing class which is not an anonymous function.

**Related issue**: 

**Release Notes**


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -811,7 +811,16 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
       }
     }
 
+    val isAnonFunc = ".*\\$\\$anonfun\\$\\d+$".r
+    def getNonFuncClass(clz: Class[_]): Option[Class[_]] = {
```
ducky64, line 815: Can you add the rationale as a comments here, specifically including that this is a workaround for an older version of Scala? The explanation in the issue description should be fine.

azidar, line 815:
```suggestion
    // In Scala 2.11, anonymous functions were compiled to their own classes, while in Scala 2.12,
    // they are directly compiled into the enclosing classes. This meant that checking the enclosing
    // parent in 2.12 would work, but in 2.11 they wouldn't. This fix just looks for the first enclosing class
    // which is not an anonymous function.
    def getNonFuncClass(clz: Class[_]): Option[Class[_]] = {
```

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -196,4 +196,26 @@ class AutoClonetypeSpec extends ChiselFlatSpec with Utils {
...
   }
+
+  "Wrapped IO construction without parent reference" should "not fail for autoclonetype" in {
+    class TestModule extends MultiIOModule {
+      def thunk[T](f: => T): T = f
+      val works = thunk(IO(new Bundle {
+        val x = Output(UInt(3.W))
+      }))
+    }
+    elaborate { new TestModule }
```
azidar, line 207:
```suggestion
    ChiselStage.elaborate { new TestModule }
```

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -196,4 +196,26 @@ class AutoClonetypeSpec extends ChiselFlatSpec with Utils {
...
+  "Wrapped IO construction with parent references" should "not fail for autoclonetype" in {
+    class TestModule(blah: Int) extends MultiIOModule {
+      // Note that this currently fails only if f: =>T on Scala 2.11.12
+      // This works successfully with 2.12.11
+      def thunk[T](f: => T): T = f
+      val broken = thunk(IO(new Bundle {
+        val x = Output(UInt(blah.W))
+      }))
+    }
+    elaborate { new TestModule(3) }
```
azidar, line 219:
```suggestion
    ChiselStage.elaborate { new TestModule(3) }
```

PR #1484, opened 2020-06-19
Rename defaultVersions to chiselDependencies and define ChiselIvy class.

ucbjrl:
This allows more code duplication between sbt and mill build files and simplifies the extraction of version information by build management tools.


ucbjrl:
The main issue here is this introduces an sbt project dependency, which will force any sbt projects that contain chisel3 as a subproject, to incorporate the `project/ChiselLibrary.scala` in their top level project.

PR #1489, opened 2020-06-22
Fix Mergify Backport labeling for 3.2.x

jackkoenig:
Brings it inline with the config used in FIRRTL

**Related issue**: 

**Release Notes**

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -57,7 +57,6 @@ pull_request_rules:
 
   - name: label Mergify backport PR
     conditions:
-      - base=3.3.x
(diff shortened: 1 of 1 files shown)


albert-magyar:
I.e., why not make two separate rules (one with 3.2.x as base, one with 3.3.x)

jackkoenig:
Because if we add `3.4.x` this applies to everything

jackkoenig:
I think matching the body of `This is an automated backport of pull request \#\d+ done by Mergify` is sufficient to tell if something is a backport 🙂

Issue #1507, opened 2020-07-07
Interval literal creation is inconsistent 


chick:
**Type of issue**: bug report |  documentation

**Impact**:  Possible API modification

**Other information**
Not the discrepancy with the following Interval literals. Decimal values are encode using the .I arguments
while BigInt and Int values are interpreted according to the .I arguments. I think this is likely to create confusion
though the cat may already be out of the bag on this one.

import chisel3._

| Statement  | Value |
| ------------- | ------------- |
| val b = BigInt(-2)       | b: scala.math.BigInt = -2  |
| val e = BigDecimal(-2.0) | e: scala.math.BigDecimal = -2.0  |
| val d = -2.0             | d: Double = -2.0  |
| val l = -2               | l: Int = -2  |
|                          |   |
| val ib = b.I(2.BP)       | ib: chisel3.experimental.Interval = Interval<2>(-2)  |
| val ie = e.I(2.BP)       | ie: chisel3.experimental.Interval = Interval<4>(-8)  |
| val id = d.I(2.BP)       | id: chisel3.experimental.Interval = Interval<4>(-8)  |
| val il = l.I(2.BP)       | il: chisel3.experimental.Interval = Interval<2>(-2)  |
|                          |   |
| ib.litToDouble           | res0: Double = -0.5  |
| ie.litToDouble           | res1: Double = -2.0  |
| id.litToDouble           | res2: Double = -2.0  |
| il.litToDouble           | res3: Double = -0.5  |
|                          |   |
| ib.litValue()            | res4: BigInt = -2  |
| ie.litValue()            | res5: BigInt = -8  |
| id.litValue()            | res6: BigInt = -8  |
| il.litValue()            | res7: BigInt = -2  |

**What is the current behavior?**
When creating Interval literals the behavior is different for BigInts, Longs, and Ints, than it is for Double and BigDecimal.
In the former case the bits of the number are interpreted according to the paramters of the cap-I conversion while the latter the number is the supplied 

**What is the expected behavior?**
Seems to me that the behavior should be the same as Double and BigDecimal

**Please tell us about your environment:**
Chisel3 3.3

chick:
Check behavior of FixedPoint with respect to above.

chick:
Consider looking at scalafix to provide way conversion of existing code

chick:
Addressed by PR #1509

PR #1509, opened 2020-07-10
Fixes inconsistencies declaring FixedPoint and Interval literals

chick:
Now all follow a what you see is what you get pattern.
Formerly Longs and BigInts would be interpreted as a bit pattern. Now they are handled as the literal value.
Also FixedPoints could not be created from a BigInt, now they can.
Creating literals using the range"range-string" interpolator has been made more robust.
Tests have been added to show the new behavior

| Statement  | Original Hardware Value |  PR  Hardware Value | 
| ------------- | ------------- |  ------------- |
| BigInt(-2).I(2.BP)       | -0.50  |  -2.0  |
| -2L.I(2.BP)       | -0.50  |  -2.0  |
| BigDecimal(-2.0).I(2.BP)  | -2.0  | -2.0  | 
| -2.0.I(2.BP)  | -2.0  | -2.0  | 

**Type of change**:  enhancement

**Impact**: Minor API modification
Prior to this fix. Declaring a literal for `Interval` like `3.0.I(2.BP)` would give you a literal value of 3.00. But 
declaring `3.I(2.BP)` would create a literal with the value 0.75. This is inconsistent and counter intuitive. `FixedPoint`
literals shared the same inconsistency

With these changes `FixedPoint` and `Interval` literal declaration follow the convention of least surprise.

**Release Notes**
Creation of Interval and FixedPoint literals is consistent across the different number types BigDecimal, BigInt, Double, Long, Int. Now

### DIFF
--- a/core/src/main/scala/chisel3/Bits.scala
+++ b/core/src/main/scala/chisel3/Bits.scala
@@ -1613,6 +1613,16 @@ package experimental {
         }
       }
 
+      implicit class fromBigIntToLiteral(bigInt: BigInt) {
+        def F(binaryPoint: BinaryPoint): FixedPoint = {
+          FixedPoint.fromBigDecimal(BigDecimal(bigInt), Width(), binaryPoint)
+        }
+
+        def F(width: Width, binaryPoint: BinaryPoint): FixedPoint = {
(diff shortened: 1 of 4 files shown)


chick:
Address Issue #1507

chick:
Fixed Point going away

PR #1515, opened 2020-07-15
Allow a counter to be instantiated using a Scala range

nullobject:
This PR adds the ability for a counter to be instantiated using a Scala range.

The original API has been preserved, but the counter now works internally with a range. Using a range to configure the counter allows us to do interesting things.

Counting upwards:

```scala
val (value, wrap) = Counter(0 to 3)
// 0, 1, 2, 3, 0, 1...
```

Counting downwards:

```scala
val (value, wrap) = Counter(3 to 0)
// 3, 2, 1, 0, 3, 2...
```

Counting upwards by twos:

```scala
val (value, wrap) = Counter(0 to 9 by 2)
// 0, 2, 4, 6, 8, 0, 2...
```

It also allows an optional condition attribute (similar to the existing API):

```scala
val enable = true.B
val (value, wrap) = Counter(0 to 3, enable)
// 0, 1, 2, 3, 0, 1...
```

I'm fairly new to Chisel, but I think that having the condition as the second parameter feels more consistent with the rest of the Chisel API.

I'm happy to scale this PR back if you think it's changing too much at once, but I'm keen to get some feedback on the overall goal. Thoughts?

**Release Notes**
Allow a counter to be instantiated using a Scala range

### DIFF
--- a/src/main/scala/chisel3/util/Counter.scala
+++ b/src/main/scala/chisel3/util/Counter.scala
@@ -22,16 +22,17 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   */
 @chiselName
 class Counter private (r: Range) {
+  require(r.length > 0, s"Counter range cannot be empty, got: $r")
   require(r.start >= 0 && r.end >= 0, s"Counter range must be positive, got: $r")
 
-  private val delta = math.abs(r.step)
-  private val width = math.max(log2Up(r.last + 1), log2Up(r.head + 1))
+  private lazy val delta = math.abs(r.step)
+  private lazy val width = math.max(log2Up(r.last + 1), log2Up(r.head + 1))
 
   /** Creates a counter with the specified number of steps.
     *
     * @param n number of steps before the counter resets
     */
-  def this(n: Int) { this(0 until n) }
+  def this(n: Int) { this(0 until math.max(1, n)) }
 
   /** The current value of the counter. */
   val value = if (r.length > 1) RegInit(r.head.U(width.W)) else r.head.U
--- a/src/main/scala/chisel3/util/Counter.scala
+++ b/src/main/scala/chisel3/util/Counter.scala
@@ -24,8 +24,8 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
 class Counter private (r: Range) {
   require(r.start >= 0 && r.end >= 0, s"Counter range must be positive, got: $r")
 
-  private val delta = math.abs(r.step).U
-  private val width = math.max(log2Up(r.head), log2Up(r.last)) + 1
+  private val delta = math.abs(r.step)
+  private val width = math.max(log2Up(r.last + 1), log2Up(r.head + 1))
 
   /** Creates a counter with the specified number of steps.
     *
@@ -34,7 +34,7 @@ class Counter private (r: Range) {
   def this(n: Int) { this(0 until n) }
 
   /** The current value of the counter. */
-  val value = if (r.length > 1) RegInit(r.start.U(width.W)) else r.start.U
+  val value = if (r.length > 1) RegInit(r.head.U(width.W)) else r.head.U
 
   /** The range of the counter values. */
   def range: Range = r
@@ -48,16 +48,19 @@ class Counter private (r: Range) {
     if (r.length > 1) {
       val wrap = value === r.last.U
 
-      when (wrap) {
-        value := r.start.U
-      } otherwise {
-        if (r.step > 0) {
-          // Increasing range
-          value := value + delta
-        } else {
-          // Decreasing range
-          value := value - delta
-        }
+      if (r.step > 0) {
+        // Increasing range
+        value := value + delta.U
+      } else {
+        // Decreasing range
+        value := value - delta.U
+      }
+
+      // We only need to explicitly wrap counters that don't start at zero, or
+      // end on a power of two. Otherwise we just let the counter overflow
+      // naturally to avoid wasting an extra mux.
+      if (!(r.head == 0 && isPow2(r.last + delta))) {
+        when(wrap) { value := r.head.U }
       }
 
       wrap
--- a/src/main/scala/chisel3/util/Counter.scala
+++ b/src/main/scala/chisel3/util/Counter.scala
@@ -19,27 +19,47 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   *     ...
   *   }
   * }}}
-  *
-  * @param n number of counts before the counter resets (or one more than the
-  * maximum output value of the counter), need not be a power of two
   */
 @chiselName
-class Counter(val n: Int) {
-  require(n >= 0, s"Counter value must be nonnegative, got: $n")
-  val value = if (n > 1) RegInit(0.U(log2Ceil(n).W)) else 0.U
+class Counter private (r: Range) {
+  require(r.start >= 0 && r.end >= 0, s"Counter range must be positive, got: $r")
 
-  /** Increment the counter
+  private val delta = math.abs(r.step).U
+  private val width = math.max(log2Up(r.start), log2Up(r.end)) + 1
+
+  /** Creates a counter with the specified number of steps.
+    *
+    * @param n number of steps before the counter resets
+    */
+  def this(n: Int) { this(0 until n) }
+
+  /** The current value of the counter. */
+  val value = if (r.length > 1) RegInit(r.start.U(width.W)) else r.start.U
+
+  /** The range of the counter values. */
+  def range: Range = r
+
+  /** Increments the counter by a step.
     *
     * @note The incremented value is registered and will be visible on the next clock cycle
-    * @return whether the counter will wrap to zero on the next cycle
+    * @return whether the counter will wrap on the next clock cycle
     */
   def inc(): Bool = {
-    if (n > 1) {
-      val wrap = value === (n-1).U
-      value := value + 1.U
-      if (!isPow2(n)) {
-        when (wrap) { value := 0.U }
+    if (r.length > 1) {
+      val wrap = value === r.last.U
+
+      when (wrap) {
+        value := r.start.U
+      } otherwise {
+        if (r.step > 0) {
+          // Increasing range
+          value := value + delta
+        } else {
+          // Decreasing range
+          value := value - delta
+        }
       }
+
       wrap
     } else {
       true.B
@@ -65,6 +85,21 @@ object Counter
     val c = new Counter(n)
     val wrap = WireInit(false.B)
     when (cond) { wrap := c.inc() }
+    (c.value, wrap)
+  }
+
+  /** Creates a counter that steps through a specified range of values.
+    *
+    * @param r the range of counter values
+    * @param enable controls whether the counter increments this cycle
+    * @return tuple of the counter value and whether the counter will wrap (the value is at
+    * maximum and the condition is true).
+    */
+  @chiselName
+  def apply(r: Range, enable: Bool = true.B): (UInt, Bool) = {
+    val c = new Counter(r)
+    val wrap = WireInit(false.B)
+    when (enable) { wrap := c.inc() }
     (c.value, wrap)
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/scala/chisel3/util/Counter.scala
```diff
@@ -24,22 +24,45 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
     *
     * @note The incremented value is registered and will be visible on the next clock cycle
-    * @return whether the counter will wrap to zero on the next cycle
+    * @return whether the counter will wrap on the next clock cycle
     */
   def inc(): Bool = {
-    if (n > 1) {
-      val wrap = value === (n-1).U
-      value := value + 1.U
-      if (!isPow2(n)) {
```
nullobject, line 40: I'm not sure that this `isPow2` check can still be used, as a range allows stepping by numbers other than one.

jackkoenig: I think the new implementation here more than covers that case.

My only [mild] concern is that power-of-2 size Counters that start at 0 may be marginally more efficient with the old specialized logic taking advantage of rollover. There is a slight difference in the emitted Verilog:

Before:
```verilog
  reg [3:0] count; // @[Counter.scala 29:33]
  wire [3:0] _T_2 = count + 4'h1; // @[Counter.scala 39:22]
  always @(posedge clock) begin
    if (reset) begin
      count <= 4'h0;
    end else if (io_en) begin
      count <= _T_2;  // Hopefully we will be inlining this sooner rather than latter
    end
```

After:
```verilog
  reg [4:0] count; // @[Counter.scala 37:40]
  wire  _T = count == 5'hf; // @[Counter.scala 49:24]
  wire [4:0] _T_2 = count + 5'h1; // @[Counter.scala 56:26]
  always @(posedge clock) begin
    if (reset) begin
      count <= 5'h0;
    end else if (io_en) begin
      if (_T) begin
        count <= 5'h0;
      end else begin
        count <= _T_2;
      end
    end
  end
```

I think we should eliminate that extra `if`. It may be that synthesis tools can figure this out, but even if they can it'd be ideal to keep the Verilog for that common case the same. While I don't like the style, it might be simplest just to have special logic for that case.

Also the width is too big.

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,47 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
   */
 @chiselName
-class Counter(val n: Int) {
-  require(n >= 0, s"Counter value must be nonnegative, got: $n")
-  val value = if (n > 1) RegInit(0.U(log2Ceil(n).W)) else 0.U
+class Counter private (r: Range) {
+  require(r.start >= 0 && r.end >= 0, s"Counter range must be positive, got: $r")
+
+  private val delta = math.abs(r.step).U
+  private val width = math.max(log2Up(r.start), log2Up(r.end)) + 1
```
jackkoenig, line 28:
```suggestion
  private val width = math.max(log2Up(r.head), log2Up(r.last)) + 1
```
I think using the actual first and last (rather than specified `start` and `end`) actually brings the step and inclusivity into the calculation. Here's a Scastie I used to decide what to use:
https://scastie.scala-lang.org/LsvLdMkoQBeTcXUoZh4brA

aswaterman:
Yosys, at least, doesn't optimize the mux.  Keep the special case.

src/main/scala/chisel3/util/Counter.scala
aswaterman: Yosys, at least, doesn't optimize the mux.  Keep the special case.

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,47 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
-    if (n > 1) {
-      val wrap = value === (n-1).U
-      value := value + 1.U
-      if (!isPow2(n)) {
-        when (wrap) { value := 0.U }
+    if (r.length > 1) {
+      val wrap = value === r.last.U
+
+      when (wrap) {
+        value := r.start.U
```
aswaterman, line 52: I think the special case can be applied by sticking an `if (!(isPow2(r.end) && isPow2(r.step)))` here.

src/main/scala/chisel3/util/Counter.scala
nullobject: Thanks, that is much better :+1:

src/main/scala/chisel3/util/Counter.scala
nullobject: I have re-added the optimization to let the counter overflow naturally when it is possible.

src/main/scala/chisel3/util/Counter.scala
nullobject: Thanks. I ended up using something very similar to that conditional, but had to also check that the range starts at zero.

nullobject:
@jackkoenig @aswaterman Please re-review when you get a chance.

I'm happy to keep tweaking this until you guys are happy :smile_cat:

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,50 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
+        value := value + delta.U
+      } else {
+        // Decreasing range
+        value := value - delta.U
       }
+
+      // We only need to explicitly wrap counters that don't start at zero, or
+      // end on a power of two. Otherwise we just let the counter overflow
+      // naturally to avoid wasting an extra mux.
+      if (!(r.head == 0 && isPow2(r.last + delta))) {
```
aswaterman, line 62: Thanks for addressing my concern.  I agree this works to cover all the old cases, as well as some new ones.  I do think there's a superset of this set that doesn't need to be reset, though: `isPow2((r.end - r.start).abs)`.  I don't think it matters that the initial value is 0, just that the total span is a power of 2.  Check my logic, though.

I'll let @jackkoenig review the rest.

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,50 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   *     ...
   *   }
   * }}}
-  *
-  * @param n number of counts before the counter resets (or one more than the
-  * maximum output value of the counter), need not be a power of two
```
jackkoenig, line 24: It would be nice to have ScalaDoc with the new range API

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,50 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
   */
 @chiselName
-class Counter(val n: Int) {
-  require(n >= 0, s"Counter value must be nonnegative, got: $n")
-  val value = if (n > 1) RegInit(0.U(log2Ceil(n).W)) else 0.U
+class Counter private (r: Range) {
```
jackkoenig, line 24: Related to https://github.com/freechipsproject/chisel3/pull/1363, better would probably be `NumericRange[BigInt]` (while having factory methods for both kinds of ranges). That being said, because this is private it should be easy enough to change so I think it's best to merge this PR and leave the `BigInt` form for later 🙂

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -67,4 +90,19 @@ object Counter
...
+    * @param cond condition that controls whether the counter increments this cycle
+    * @return tuple of the counter value and whether the counter will wrap (the value is at
+    * maximum and the condition is true).
+    */
+  @chiselName
+  def apply(r: Range, cond: Bool = true.B): (UInt, Bool) = {
+    val c = new Counter(r)
+    val wrap = WireInit(false.B)
+    when (cond) { wrap := c.inc() }
+    (c.value, wrap)
```
jackkoenig, line 106: This is a minor API question, but the range API is a little inconsistent with the old API, if you want an enable its:
```scala
Counter(0 until 16, en)
// vs.
Counter(en, 16)
```
I won't contest that maybe it's the existing API that's awkward, but we probably ought to be consistent here

jackkoenig:
@chick or something with admin rights, can you check why the CI isn't reporting?

src/main/scala/chisel3/util/Counter.scala
nullobject: Absolutely, I want to do a follow-up PR to update all the ScalaDoc for the Counter API.

I originally had those changes in here, but removed them to keep this PR more focused.

src/main/scala/chisel3/util/Counter.scala
nullobject: Cool, that makes sense :+1:

src/main/scala/chisel3/util/Counter.scala
nullobject: @jackkoenig Thanks for your thoughts :thinking: 

Unfortunately, flipping the order of the params would make having default values for the `enable` param more awkward. IMHO, since you can't construct a counter without a range value (i.e. what good is a counter with an `enable` and no range?), it makes sense that this is the first parameter.

My vote would be to deprecate the old API. That way users could have:

```scala
Counter(10)
Counter(10, en)
Counter(0 to 10)
Counter(0 to 10, en)
```

I'm happy to open a new PR to discuss?

jackkoenig:
@nullobject the test failure may be a real one, the tests that failed use `Counter`

nullobject:
> the test failure may be a real one, the tests that failed use Counter

@jackkoenig Fixed. It was due to a counter being instantiated with `Counter(0)` in the shift register test. This was creating an empty range.

I have update the legacy constructor to always create a non-empty range, and added an assertion to ensure that counters can never be instantiated with an empty range.

I don't think supporting empty ranges makes much sense, because how can you ever count zero elements?

src/main/scala/chisel3/util/Counter.scala
```diff
@@ -19,27 +19,51 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
...
 @chiselName
-class Counter(val n: Int) {
-  require(n >= 0, s"Counter value must be nonnegative, got: $n")
-  val value = if (n > 1) RegInit(0.U(log2Ceil(n).W)) else 0.U
+class Counter private (r: Range) {
+  require(r.length > 0, s"Counter range cannot be empty, got: $r")
+  require(r.start >= 0 && r.end >= 0, s"Counter range must be positive, got: $r")
+
+  private lazy val delta = math.abs(r.step)
+  private lazy val width = math.max(log2Up(r.last + 1), log2Up(r.head + 1))
```
aswaterman, line 29: Not a problem, but just FYI: `abs` and `max` are defined on the various integral types, so you can write e.g. `r.step.abs` or `foo max bar`.

aswaterman:
I think updating the legacy constructor to support `Counter(0)` is fine.  I think it makes sense that the range constructor should require a nonempty range since it represents a half-open interval.

src/main/scala/chisel3/util/Counter.scala
nullobject: Thanks :smile:

I'm still learning how to write idomatic Scala.

jackkoenig:
@nullobject do you care how this is merged? Do you want it with a merge commit or is a squash-and-merge fine?

Also sorry for the delay, somehow I missed the last few emails about this PR.

nullobject:
> @nullobject do you care how this is merged? Do you want it with a merge commit or is a squash-and-merge fine?
> 
> Also sorry for the delay, somehow I missed the last few emails about this PR.

@jackkoenig Not really, whatever you guys normally do.

Happy to squash it all into one commit if you prefer?

jackkoenig:
@nullobject since we require PRs to be up-to-date to merge, we usually usually let the trust bot @Mergify.io queue them up, merging master, then squash-and-merging. I'll get that started now. Thanks!

nullobject:
@jackkoenig Fantastic. Thanks for persisting with my PR.

I'm happy to have made my first contribution to Chisel :raised_hands:

jackkoenig:
@nullobject Thank you so much! It was a pleasure and an excellent contribution!

PR #1519, opened 2020-07-20
Remove scalastyle configurations

albert-magyar:
See freechipsproject/firrtl#1767, we will be moving to scalafmt and will do one big "auto-cleanup" right before 1.4.


albert-magyar:
> This looks fine.
> I am curious as to why you removed the //scalastyle: from just 22 of the files.

Good point, I mistakenly only applied the scripts to `src`! Fixed now, thanks.

PR #1527, opened 2020-07-29
Allow a counter to be manually reset

nullobject:
- [x] Depends on #1515.

This PR adds a `reset` parameter to the Counter API, that allows a counter to be reset to its initial value.

It is sometimes useful to be able to reset a counter arbitrarily. By default the `reset` parameter is `false`, which means the proposed change is compatible with the existing API.

For example

```scala
val counterReset = WireInit(false.B)
val (value, wrap) = Counter(0 to 3, reset = counterReset)

// Trigger a reset
counterReset := true.B
```

**Release Notes**
Allow counters to be reset

### DIFF
--- a/src/test/scala/chiselTests/Counter.scala
+++ b/src/test/scala/chiselTests/Counter.scala
@@ -27,6 +27,19 @@ class EnableTester(seed: Int) extends BasicTester {
   }
 }
 
+class ResetTester(n: Int) extends BasicTester {
+  val triggerReset = WireInit(false.B)
+  val wasReset = RegNext(triggerReset)
+  val (value, _) = Counter(0 until 8, reset = triggerReset)
+
+  triggerReset := value === (n-1).U
+
+  when(wasReset) {
+    assert(value === 0.U)
+    stop()
+  }
+}
+
 class WrapTester(max: Int) extends BasicTester {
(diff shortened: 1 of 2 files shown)


seldridge:
With #1515 merged, would you be able to rebase this onto `master` @nullobject?

And thanks for this. It's something that I've written one-off workarounds on multiple occasions for.

nullobject:
@seldridge Sure thing, I will get it out of draft status and mergable :+1:

nullobject:
> With #1515 merged, would you be able to rebase this onto master @nullobject?

@seldridge It has been rebased.

src/test/scala/chiselTests/Counter.scala
```diff
@@ -57,6 +70,10 @@ class CounterSpec extends ChiselPropSpec {
     forAll(safeUInts) { (seed: Int) => whenever(seed >= 0) { assertTesterPasses{ new EnableTester(seed) } } }
   }
 
+  property("Counter can be reset") {
+    forAll(smallPosInts) { (seed: Int) => assertTesterPasses{ new ResetTester(seed) } }
+  }
+
```
seldridge, line 76: This test is probably overkill and it seems to repeat (at least locally, for me) seeds. Out of scope for this PR, though. I like having this PR consistent with the other tests here.

PR #1546, opened 2020-08-11
Restore Counter.n API

jackkoenig:
Includes special case support for Counter(0) which has identical
behavior to Counter(1) except for the value of n.

Restore API accidentally deleted in https://github.com/freechipsproject/chisel3/pull/1515. I'm a bit unsure about the `oldN` which is added only to support the case of `Counter(0)`. It's private so not a big deal to add or remove, but it shows that the API when passed `0` is funky.

Also, `Counter.n` makes sense to me for ranges starting at 0 with step 1, but it's unclear what it *should* mean for other cases so I made it throw an exception in those cases.

**Related issue**: 

**Impact**: API ~modification~ restoration

**Release Notes**

### DIFF
--- a/src/main/scala/chisel3/util/Counter.scala
+++ b/src/main/scala/chisel3/util/Counter.scala
@@ -19,20 +19,42 @@ import chisel3.internal.naming.chiselName  // can't use chisel3_ version because
   *     ...
   *   }
   * }}}
+  * @example {{{
+  *   // Using Scala Range API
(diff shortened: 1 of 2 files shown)


jackkoenig:
> The n=0 case is certainly debatable, but it makes sense to err on the side of backwards compatibility.

I'm open to changing it, but will go ahead and merge this in the meantime.

PR #1551, opened 2020-08-13
Deprecate support for Scala 2.11

jackkoenig:
Leverages chiselRuntimeDeprecated infrastructure. As such it is not
currently suppressible.

Related to https://github.com/freechipsproject/firrtl/pull/1842. This doesn't have suppression because nothing in Chisel elaboration does. I looked at adding suppression via annotations like in the FIRRTL PR, but some invocations of the Builder don't have annotations to propagate (like aspects). Also nothing else is suppressible so I figured this needs further thought.

**Related issue**: 

**Impact**: API modification

**Release Notes**

Deprecate support for Scala 2.11

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -180,6 +180,12 @@ lazy val macros = (project in file("macros")).
 
 lazy val core = (project in file("core")).
   settings(commonSettings: _*).
+  enablePlugins(BuildInfoPlugin).
+  settings(
+    buildInfoPackage := "chisel3",
+    buildInfoUsePackageAsPath := true,
+    buildInfoKeys := Seq[BuildInfoKey](buildInfoPackage, version, scalaVersion, sbtVersion)
+  ).
   settings(publishSettings: _*).
   settings(
     name := "chisel3-core",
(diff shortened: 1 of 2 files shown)


jackkoenig:
To be dry, this gets the link from https://github.com/freechipsproject/firrtl/pull/1842 so will fail CI until that's merged.

ducky64:
It might also be a good idea to write the worst case scenario in the message itself, before the user needs to click through the link, to motivate them to click through to the link. For example - compile time failures (though that also means they'll never see this message)? Runtime failures (the message could be used to debug "weird" runtime failures)? Generated circuit is subtly wrong?

jackkoenig:
> It might also be a good idea to write the worst case scenario in the message itself, before the user needs to click through the link, to motivate them to click through to the link. For example - compile time failures (though that also means they'll never see this message)? Runtime failures (the message could be used to debug "weird" runtime failures)? Generated circuit is subtly wrong?

Yeah good suggestion. It's really just "You can't upgrade to Chisel 3.5", maybe not that scary but hopefully catches someone's eye 🙂

jackkoenig:
I changed the message to:
```
Chisel 3.4 is the last version that will support Scala 2.11. Please upgrade to Scala 2.12. See https://www.chisel-lang.org/chisel3/upgrading-from-scala-2-11.html
```
FIRRTL dependency is merged and link is live so marking `Please Merge`.

PR #1554, opened 2020-08-14
Move Treadle dependency to the tests

jackkoenig:
Use inheritance to make TesterDriver Backend API extensible, then define
a TreadleBackend in the test project
Now, Treadle is only a dependency of the test configuration. When doing this manual modification

### Thoughts / Issues:
* Treadle is only a dependency of the test configuration now
* It doesn't run out of the box (requires modifying `ChiselSpec.scala`), but it is similar to the case before
* ~~Tests that require Verilator no longer work when you do the modification to use Treadle.~~
    * ~~This needs to be fixed but I would argue is less important than getting this in to remove an unnecessary dependency.~~
    * This is fixed
* While FIRRTL can be built from source, this does not allow building Treadle from source
    * I don't think the old code worked anyway--2 fat jars as unmanaged dependencies would have conflicting classes in them
    * I intend to improve the rocket-chip building from source stuff separately anyway which involves killing the firrtl fat jar
* The backend API is kind of awkward, but it's basically the same as before + the required `execute` method.

**Related issue**: 

**Impact**: API modification

**Release Notes**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,8 +29,7 @@ def javacOptionsVersion(scalaVersion: String): Seq[String] = {
 }
 
 val defaultVersions = Seq(
-  "edu.berkeley.cs" %% "firrtl" % "1.4-SNAPSHOT",
-  "edu.berkeley.cs" %% "treadle" % "1.3-SNAPSHOT"
+  "edu.berkeley.cs" %% "firrtl" % "1.4-SNAPSHOT"
 )
 
 lazy val commonSettings = Seq (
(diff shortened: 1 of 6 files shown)


jackkoenig:
@chick I hope we can get this merged before release. I'm working on fixing the `VerilatorOnly` tests when running with `Treadle`

jackkoenig:
I think this is good to go. I manually inspected the `.pom` when doing `publishLocal` and Treadle is only a dependency of the test configuration.

PR #1556, opened 2020-08-14
fix build release problem

chick:
Fixes a problem where build.sc was not in agreement with build.sbt
Broke release scripts

Build release issue

**Release Notes**


ucbjrl:
This wasn't @jackkoenig's fault. There is insufficient documentation on release process/script assumptions.

PR #1558, opened 2020-08-18
Remove use of PreservesAll, cleanup dependencies

seldridge:
Remove usages of the deprecated trait PreservesAll and use an explicit
false invalidates. Additionally, all phases are converted to be more
canonical in there specification of dependencies by:

  1. Overriding all default dependency implementations
  2. Using def instead of val

**Related issue**: None

**Release Notes**

None


jackkoenig:
Is this something that should be in 3.4.0?

seldridge:
Yeah, this would be good to sneak in. No functional change from RC-1, but a binary incompatibility.

PR #1566, opened 2020-08-25
Add ChiselPhase, Stop writing files in ChiselStage$ methods, Expand ChiselStage$ helpers

seldridge:
Adds a new `private[chisel3]` phase called `ChiselPhase` which does the work of the one-off phase manager inside the existing `ChiselStage`. This is then used inside `ChiselStage`. `ChiselStage` companion object methods, `elaborate` and `convert`, are then switched to use `ChsielStage` so that they stop writing files.

This adds new APIs to the `ChiselStage` object that write no files:

- `emitChirrtl(gen: => RawModule): String`
- `emitFirrtl(gen: => RawModule): String`
- `emitVerilog(gen: => RawModule): String`
- `emitSystemVerilog(gen: => RawModule): String`

Tests are added, thanks to Java security manager magic, to assert that these methods do, in fact, write no files. Existing tests that use the `ChiselStage` class are modified to use the object if they are only checking a string output (and have no need to generate files).

### Commentary

The new methods which run the FIRRTL compiler have to not use `FirrtlStage` (as that will write files), but this patch series shows how to use a `PhaseManager` to piece together a custom solution out of existing `Phases`. I think that this hints at an overall better API than the omnibus `Stage` which is opinionated in the fixed set of `Phases` that it wants to run before and after your code.

tl;dr: I think a flat structure is better than nesting of stages and it's worthwhile to think about how to refactor things to work this way.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

None.

None.

#### Desired Merge Strategy

#### Release Notes

Add additional helper methods to ChiselStage object: emitChirrtl, emitFirrtl, emitVerilog, emitSystemVerilog. These new methods, as well as existing ChiselStage object methods, elaborate and convert, will no longer write files.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/stage/ChiselStage.scala
+++ b/src/main/scala/chisel3/stage/ChiselStage.scala
@@ -22,17 +22,7 @@ class ChiselStage extends Stage {
 
   val shell: Shell = new Shell("chisel") with ChiselCli with FirrtlCli
 
-  val targets: Seq[Dependency[Phase]] =
-    Seq( Dependency[chisel3.stage.phases.Checks],
-         Dependency[chisel3.stage.phases.AddImplicitOutputFile],
-         Dependency[chisel3.stage.phases.AddImplicitOutputAnnotationFile],
-         Dependency[chisel3.stage.phases.MaybeAspectPhase],
-         Dependency[chisel3.stage.phases.Convert],
-         Dependency[chisel3.stage.phases.MaybeFirrtlStage] )
-
-  final lazy val phaseManager = new PhaseManager(targets) {
-    override val wrappers = Seq( (a: Phase) => DeletedWrapper(a) )
-  }
+  final lazy val phaseManager = new ChiselPhase
 
   def run(annotations: AnnotationSeq): AnnotationSeq = try {
     phaseManager.transform(annotations)
--- a/src/main/scala/chisel3/stage/ChiselStage.scala
+++ b/src/main/scala/chisel3/stage/ChiselStage.scala
@@ -2,10 +2,15 @@
 
 package chisel3.stage
 
-import firrtl.{ir => fir, AnnotationSeq, EmittedFirrtlCircuitAnnotation, EmittedVerilogCircuitAnnotation}
+import firrtl.{
+  ir => fir,
+  AnnotationSeq,
+  EmittedFirrtlCircuitAnnotation,
+  EmittedVerilogCircuitAnnotation,
+  HighFirrtlEmitter}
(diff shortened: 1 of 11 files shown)


jackkoenig:
> This removes the ChiselStage.targets API

This is a backwards compatible change from 3.3.2, why are we removing it?

src/main/scala/chisel3/stage/ChiselStage.scala
```diff
@@ -22,17 +22,7 @@ class ChiselStage extends Stage {
...
-         Dependency[chisel3.stage.phases.AddImplicitOutputFile],
-         Dependency[chisel3.stage.phases.AddImplicitOutputAnnotationFile],
-         Dependency[chisel3.stage.phases.MaybeAspectPhase],
-         Dependency[chisel3.stage.phases.Convert],
-         Dependency[chisel3.stage.phases.MaybeFirrtlStage] )
-
-  final lazy val phaseManager = new PhaseManager(targets) {
-    override val wrappers = Seq( (a: Phase) => DeletedWrapper(a) )
-  }
+  final lazy val phaseManager = new ChiselPhase
```
jackkoenig, line 25: I don't think we should remove this API without first deprecating. It is used by rocket-chip after all: https://github.com/chipsalliance/rocket-chip/blob/d71c5a674275676b9c870a20cb0a42cd039a4115/src/main/scala/system/RocketChipStageGenerator.scala#L15

I dunno how many other users there might be, but I'd say preserve it by keeping `phaseManager` the same and then implement targets as `def targets: Seq[Dependency[Phase]] = (new ChiselPhase).targets`, then it can still be overridden to affect `phaseManager`, feel free to deprecate it though, instead suggesting composition I assume?

seldridge: Good catch. Updated in the rebase. There's annoyingly no good Scala 2.11 compatible way to deprecate this. `@deprecatedOverriding` is what we want, but it's a 2.12 API addition. (And just using `@deprecated` produces no warnings if you override it.)

That said, I feel like there are some warts with the `Stage` API that can be better handled with a flat structure. So, I wouldn't sweat this and would rather put more thought into how to better express this flat structure (which would likely be a larger update to the `Stage` API.)

seldridge:
@jackkoenig: One minor update to this. I changed tests that were using the class methods to using the object methods to prevent unnecessary file writing. Still okay to merge?

jackkoenig:
Sorry for the delay, this looks great!

PR #1570, opened 2020-08-26
Update README.md

schoeberl:
A quick fix for the developer confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ These are the base data types for defining circuit components:
 
 ![Image](https://raw.githubusercontent.com/freechipsproject/chisel3/master/doc/images/type_hierarchy.svg?sanitize=true)
 
-## Developer Documentation
+## Chisel Library Developer Documentation
 This section describes how to get started developing Chisel itself, including how to test your version locally against other projects that pull in Chisel using [sbt's managed dependencies](https://www.scala-sbt.org/1.x/docs/Library-Dependencies.html).
 
 ### Compiling and Testing Chisel
--- a/README.md
+++ b/README.md
@@ -138,8 +138,8 @@ These are the base data types for defining circuit components:
 
 ![Image](https://raw.githubusercontent.com/freechipsproject/chisel3/master/doc/images/type_hierarchy.svg?sanitize=true)
 
-## Chisel Library Developer Documentation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,7 +138,7 @@ These are the base data types for defining circuit components:
 
 ![Image](https://raw.githubusercontent.com/freechipsproject/chisel3/master/doc/images/type_hierarchy.svg?sanitize=true)
 
-## Developer Documentation
+## Chisel Library Developer Documentation
```
edwardcwang, line 141: Just curious: what does "Chisel Library" refer to here?

schoeberl: What is Chisel? It is a Scala library, right? But I am open to a better name. Just Chisel Developer Documentation is not enough. If I am a Java developer I develop Java programs and not the Java compiler or the JVM. If I am a Chisel developer than I am developing hardware in Chisel, not the Chisel language/library. So maybe: "Chisel Language Developer Documentation"?

edwardcwang: From a software perspective I thought "developer doc" was fine since it connoted the opposite of "user doc".

I almost thought "Chisel Library Developer Documentation" was "how to develop a library e.g. rocketchip that uses Chisel"

schoeberl:
But developer documentation is very misleading, as the example with Java developer from above. There was a very long email thread by one Chisel "user" who misunderstood how to install/use Chisel. Again, Chisel "users" are developers. When you add code to Chisel you are a Scala developer, although you are not developing the Scala compiler, right?

edwardcwang:
I interpret a Chisel user as a Scala user, not a Scala developer since they are not making changes to the Scala compiler or language.

edwardcwang:
Though to be fair a Chisel user also "develops in Scala" so I do see the ambiguity.

README.md
jackkoenig: I get the same impression as @edwardcwang, Chisel is technically a library but I find most new users don't really grasp that for a while. I've noticed that a lot of projects use the word "Contributor", so what about: "Contributor Documentation"?

The next line mentions "This section describes how to get started developing Chisel itself", but maybe we should slightly rephrase that and bold it.

schoeberl:
+1 for Contributor, that should be clear

PR #1583, opened 2020-09-09
Fix load memory from file to work with binary

### LINKED ISSUE #1582: loadMemoryFromFile always results in $readmemh even if the user specifies the Binary option
**Type of issue**: bug report

Currently there is a bug in loadMemoryFromFile. The hexOrBinary parameter is not passed into ChiselLoadMemoryAnnotation. This results in the generation of $readmemh instead of $readmemb, even if MemoryLoadFileType.Binary is specified by the user.

The solution is to simply pass in the hexOrBinary parameter to ChiselLoadMemoryAnnotation. Like so:

```scala
/** Annotate a memory such that it can be initialized using a file
    * @param memory the memory
    * @param filename the file used for initialization
    * @param hexOrBinary whether the file uses a hex or binary number representation
    */
  def apply[T <: Data](
    memory: MemBase[T],
    fileName: String,
    hexOrBinary: MemoryLoadFileType.FileType = MemoryLoadFileType.Hex
  ): Unit = {
    annotate(ChiselLoadMemoryAnnotation(memory, fileName, hexOrBinary))
  }
}
```

https://github.com/freechipsproject/chisel3/blob/05ac57c160c58268c571b74d6f688b87ff4312b0/src/main/scala/chisel3/util/experimental/LoadMemoryTransform.scala#L115

HappyQuark:
Passed in hexOrBinary parameter to ChiselLoadMemoryAnnotation

Test included

Fixes #1582

#### API Impact
No change in API

Allows $readmemb to be generated when the user specifies MemoryLoadFileType.Binary.

#### Desired Merge Strategy

Squash

#### Release Notes
Fixed bug in loadMemoryFromFile that caused the hexOrBinary parameter to be ignored.

### DIFF
--- a/src/main/scala/chisel3/util/experimental/LoadMemoryTransform.scala
+++ b/src/main/scala/chisel3/util/experimental/LoadMemoryTransform.scala
@@ -112,7 +112,7 @@ object loadMemoryFromFile {
     fileName: String,
     hexOrBinary: MemoryLoadFileType.FileType = MemoryLoadFileType.Hex
   ): Unit = {
-    annotate(ChiselLoadMemoryAnnotation(memory, fileName))
+    annotate(ChiselLoadMemoryAnnotation(memory, fileName, hexOrBinary))
   }
 }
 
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM, thanks!

PR #1614, opened 2020-10-12
When prefixing with a data, eagly get local name

### LINKED ISSUE #1606: Quadratic naming with recursive functions
When using an intermediate `Wire` (or `Reg`, any `Data`) in a recursive function, the new naming behavior results in a quadratic naming blowup:

```scala
def func(bools: Seq[Bool]): Bool = {
  if (bools.isEmpty) true.B
  else {
    val w = Wire(Bool())
    w := bools.head && func(bools.tail)
    w
  }
}
// Call it in a Module
val x = func(io.in)
```

This gives:
```
    wire x : UInt<1>
    wire x_x_w : UInt<1>
    wire x_x_x_x_w_w : UInt<1>
    wire x_x_x_x_w_x_x_x_x_w_w_w : UInt<1>
    node _x_x_x_x_w_x_x_x_x_w_w_x_x_x_x_w_x_x_x_x_w_w_w_T = and(io.in[3], UInt<1>("h1"))
    x_x_x_x_w_x_x_x_x_w_w_w <= _x_x_x_x_w_x_x_x_x_w_w_x_x_x_x_w_x_x_x_x_w_w_w_T
    node _x_x_x_x_w_x_x_x_x_w_w_T = and(io.in[2], x_x_x_x_w_x_x_x_x_w_w_w)
    x_x_x_x_w_w <= _x_x_x_x_w_x_x_x_x_w_w_T
    node _x_x_x_x_w_T = and(io.in[1], x_x_x_x_w_w)
    x_x_w <= _x_x_x_x_w_T
    node _x_x_T = and(io.in[0], x_x_w)
    x <= _x_x_T
```

Executable example: https://scastie.scala-lang.org/O0xhmu4eRgqdXmUb2CfqbA

While this is very funny, it's also an important bug to fix. Internally I'm seeing names well over 2000 characters long and since the Verilog spec allows tools to only support names up to 1024 characters, we're already hitting problems in various tools.

**Type of issue**: bug report

jackkoenig:
Fixes #1606

Previously, the Data itself would be put on the prefix stack and its
full name would be used as the prefix for subsequent names. This meant
that prefixes would grow quadratically as the prefix is present both on
the Data pushed to the stack, and on the stack itself. This is fixed by
just using the "local" name of the Data being pushed on the stack.

A related issue is deferring the name resolution. This led to unintuitive
behavior when the name of a Data used as a prefix is overridden later
(usually when the Data is a return value). The overriding name would
show up in prefixes twice instead of once. It is much more intuitive to
grab the name at the moment of the connection creating the prefix rather
than deferring to later.

TODO
- [ ] ~Fix linear lookup of aggregate children - proposal is to force the local names of aggregate children, this could violate weird assumptions so need to be careful~ Later
- [ ] ~Switch implementation of prefix stack to List[String] - this will improve memory and performance a decent amount~ Later

### Contributor Checklist

#### Type of Improvement

   - bug fix  

#### API Impact

No impact on API (changing types of methods on a package private object will require MiMa waivers but isn't visible to the user).

Improves names from recursive functions where prefixing occurs due to a connection quite a bit

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix quadratic naming in recursive functions (Fixes #1606)  

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -149,21 +149,6 @@ private[chisel3] trait HasId extends InstanceId {
     * @return the name, if it can be computed
     */
   def computeName(defaultPrefix: Option[String], defaultSeed: Option[String]): Option[String] = {
-    // Recursively builds a name if referenced fields of an aggregate type
-    def buildAggName(id: HasId): Option[String] = {
-      def recArg(node: Arg): Option[String] = node match {
-        case Slot(imm, name) => recArg(imm).map(_ + "_" + name)
-        case Index(imm, ILit(num)) => recArg(imm).map(_ + "_" + num)
-        case Index(imm, n: LitArg) => recArg(imm).map(_ + "_" + n.num)
-        case Index(imm, _: Node) => recArg(imm)
-        case Node(id) => recArg(id.getOptionRef.get)
-        case Ref(name) => Some(name)
-        case ModuleIO(mod, name) if _parent.contains(mod) => Some(name)
-        case ModuleIO(mod, name) => recArg(mod.getRef).map(_ + "_" + name)
-      }
-      id.getOptionRef.flatMap(recArg)
-    }
-
     /** Computes a name of this signal, given the seed and prefix
       * @param seed
       * @param prefix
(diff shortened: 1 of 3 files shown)


jackkoenig:
This is Draft for now because I want to try it out and see how well it works despite the linear lookup for names of aggregate children. I have an idea to fix it, but as mentioned it might violate bad assumptions so need to be careful.

I also want to change the prefix stack to be a `List[String]` since we no longer need to put `Data` on it, but maybe that should be a separate PR, thoughts @azidar?

azidar:
I think separate PRs is good. Let's merge this ASAP.

azidar (approved):

jackkoenig:
Alright, do you think I should do https://github.com/freechipsproject/chisel3/pull/1614/files#diff-61c46b849f372e39e0c29dc7945edcbbR374 or if performance looks okay do that in follow on as well?

azidar:
Let's do a follow-on PR.

jackkoenig:
@Mergifyio refresh

PR #1616, opened 2020-10-13
Use Data refs for name prefixing with aggregate elements

jackkoenig:
Vecs set the refs of their elements upon construction of those elements.
In the past, Records haven't set their elements refs until module close,
but it can be done sooner. Doing it upon binding means that refs will at
least be available for Records used in hardware elements. Since only
bound Data can be connected to anyway, Aggregate elements being
connected to will always have a ref which we can then use for creating
naming prefixes.

No release notes because this is a speed up to code that shouldn't be released without this (https://github.com/freechipsproject/chisel3/pull/1614)

### Contributor Checklist

#### Type of Improvement

 - performance improvement    

#### API Impact

No impact

No impact

#### Desired Merge Strategy

   - Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -479,9 +479,21 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId with Source
   * RTL writers should use [[Bundle]].  See [[Record#elements]] for an example.
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
+
+  // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
+  // of hardware created when connecting to one of these elements
+  private def setElementRefs(): Unit = {
+    // Since elements is a map, it is impossible for two elements to have the same
+    // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
+    // which can cause collisions
+    val _namespace = Namespace.empty
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk=true)) }
+  }
+
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
     try {
       super.bind(target, parentDirection)
+      setElementRefs()
     } catch {  // nasty compatibility mode shim, where anything flies
       case e: MixedDirectionAggregateException if !compileOptions.dontAssumeDirectionality =>
         val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
@@ -633,13 +645,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
     case _ => false
   }
 
-  // NOTE: This sets up dependent references, it can be done before closing the Module
   private[chisel3] override def _onModuleClose: Unit = {
-    // Since Bundle names this via reflection, it is impossible for two elements to have the same
-    // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
-    // which can cause collisions
-    val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk=true)) }
+    // This is usually done during binding, but these must still be set for unbound Records
+    if (this.binding.isEmpty) {
+      setElementRefs()
+    }
   }
 
   private[chisel3] final def allElements: Seq[Element] = elements.toIndexedSeq.flatMap(_._2.allElements)
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -480,9 +480,9 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId with Source
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
-  // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
-  // of hardware created when connecting to one of these elements
-  private def setElementRefs(): Unit = {
+  // This is almost always set during binding, with the exception of Mem types which have no binding
+  // but are used in emission of FIRRTL
+  private[chisel3] def setElementRefs(): Unit = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
@@ -643,13 +643,6 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
         that.elements.contains(name) &&
         (that.elements(name) typeEquivalent model)}
     case _ => false
-  }
-
-  private[chisel3] override def _onModuleClose: Unit = {
-    // This is usually done during binding, but these must still be set for unbound Records
-    if (this.binding.isEmpty) {
-      setElementRefs()
-    }
   }
 
   private[chisel3] final def allElements: Seq[Element] = elements.toIndexedSeq.flatMap(_._2.allElements)
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -480,9 +480,9 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId with Source
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
-  // This is almost always set during binding, with the exception of Mem types which have no binding
-  // but are used in emission of FIRRTL
-  private[chisel3] def setElementRefs(): Unit = {
+  // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
+  // of hardware created when connecting to one of these elements
+  private def setElementRefs(): Unit = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
@@ -643,6 +643,13 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
         that.elements.contains(name) &&
         (that.elements(name) typeEquivalent model)}
     case _ => false
+  }
+
+  private[chisel3] override def _onModuleClose: Unit = {
+    // This is usually done during binding, but these must still be set for unbound Records
+    if (this.binding.isEmpty) {
+      setElementRefs()
+    }
   }
 
   private[chisel3] final def allElements: Seq[Element] = elements.toIndexedSeq.flatMap(_._2.allElements)
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -165,7 +165,8 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     // Long term solution will be to define BlackBox IO differently as part of
     //   it not descending from the (current) Module
     for ((name, port) <- namedPorts) {
-      port.setRef(ModuleIO(this, _namespace.name(name)))
+      // We have to force override the _ref because it was set during IO binding
+      port.setRef(ModuleIO(this, _namespace.name(name)), force = true)
     }
 
     // We need to call forceName and onModuleClose on all of the sub-elements
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -74,6 +74,13 @@ package experimental {
         port.setRef(ModuleIO(this, _namespace.name(names(port))))
       }
 
+      // All suggestions are in, force names to every node.
+      // While BlackBoxes are not supposed to have an implementation, we still need to call
+      // _onModuleClose on all nodes (for example, Aggregates use it for recursive naming).
+      for (id <- getIds) {
+        id._onModuleClose
+      }
+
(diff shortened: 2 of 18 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -633,13 +645,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
...
   }
 
-  // NOTE: This sets up dependent references, it can be done before closing the Module
   private[chisel3] override def _onModuleClose: Unit = {
-    // Since Bundle names this via reflection, it is impossible for two elements to have the same
-    // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
-    // which can cause collisions
-    val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk=true)) }
+    // This is usually done during binding, but these must still be set for unbound Records
```
ducky64, line 649: Is there an example of when this could be the case? If possible, it'd be nice for setElementRefs to only be called from one place?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -180,7 +180,7 @@ private[chisel3] trait HasId extends InstanceId {
  *
     * @return the current calculation of a name, if it exists
     */
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
```
ducky64, line 183: Is this intentionally part of this PR, to expose this as a public-facing API?
If so, why?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -371,13 +371,10 @@ private[chisel3] object Builder {
...
-      def getSubName(field: Data, parent: Data): Option[String] = parent match {
-        case vec: Vec[_] =>
-          val idx = vec.indexOf(field)
-          if (idx >= 0) Some(idx.toString) else None
-        case rec: Record => rec.elements.collectFirst { case (name, d) if d == field => name }
-        case _ => Builder.exception(s"Shouldn't see non-Aggregate $parent as parent in ChildBinding!")
+      def getSubName(field: Data): Option[String] = field.getOptionRef.flatMap {
+        case Slot(_, field) => Some(field) // Record
+        case Index(_, ILit(n)) => Some(n.toString) // Vec
+        case _ => None // Vec dynamic indexing
```
ducky64, line 377: Can this check be made stricter and fail noisily on unexpected cases?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -479,9 +479,21 @@ trait VecLike[T <: Data] extends collection.IndexedSeq[T] with HasId with Source
...
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
+
+  // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
+  // of hardware created when connecting to one of these elements
+  private def setElementRefs(): Unit = {
+    // Since elements is a map, it is impossible for two elements to have the same
+    // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
+    // which can cause collisions
+    val _namespace = Namespace.empty
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk=true)) }
```
ducky64, line 490: It looks like setRef doesn't have any protection against overwriting refs. Is there ever a case where overwriting a ref might be intentional, or does it make sense to add a guard - either here, or in setRef? Especially if there are now two call sites to setElementRefs (yes, in theory they should never both happen for any object - but defensive programming is generally a good practice)

ducky64 (commented):
Looks mostly fine, but some questions / suggestions for defensive programming.

azidar:
I don't remember offhand, but should we be checking that the Data passed to `prefix` is bound?

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: That was a mistake, will fix

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: This is a better naming path where failure leads to it just not supplying a prefix for names. Do you really think it should fail noisily here?

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I would love to have it only call from one place; the reason I factored the code out into a function was because AFAICT that's not possible. The basic problem is that you have to do it at some point, and since users provide `def elements` in `Record`, there's not really a hook to do it. We could do it as part of `lazy val elements` in `Bundles`, but that doesn't solve the `Record` problem.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I don't think we ever want to override it so I think adding a guard makes sense

jackkoenig:
> I don't remember offhand, but should we be checking that the Data passed to prefix is bound?

No, the error is and should be reported by the connection. So if a `Data` passed to prefix is unbound it will be `None` and thus no prefix will be added.

core/src/main/scala/chisel3/internal/Builder.scala
ducky64: I don't fully understand the ref system and lifecycle, but the overall principle I'm getting at is that we should have code bugs fail as noisily as possible. If the no-name path is expected and intended, then this is fine - though can it be made tighter than `_`?

jackkoenig: I'll add a tighter bound

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: Done

core/src/main/scala/chisel3/Aggregate.scala
ducky64: Are we doing something with unbound Data (or maybe I'm missing something really basic here)?

ducky64 (approved):
lgtm!

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I was thinking about `val io` in compatibility modules, but there is `bindIoInPlace`, so maybe we don't need to bother setting the refs on module close. The fact that `io` in compatibility modules may not yet be bound *does* mean it's important that the naming not fail for aggregates without refs, they'll just get slightly worse prefixes for `io.subfield :=` connections in compatibility modules. We can still get good names if we do ref sending in `lazy val elements` inside `Bundle`, but then we need to make sure we still set the refs on binding for non-Bundle `Records`.

jackkoenig:
Well @ducky64, you win.

Removing the default case in the naming caught a couple of issues, and asserting that ref isn't already set found 1 case where we need to override it (The fields of `val io` in `BlackBox` which needs to override the refs to ensure the `io` prefix does not show up in the FIRRTL). You can see the commit I just pushed which made all of these cases more explicit which is good.

In particular, the case of `BlackBox` ports makes it clear that the approach in https://github.com/freechipsproject/chisel3/pull/1624 of getting the `Record` field names from the `Record` itself will not work.

jackkoenig:
In fact, I'm going to take a stab at solving the issue in https://github.com/freechipsproject/chisel3/pull/1624#issuecomment-711415558. My guess is that the _only_ case of unbound `Record`s needing their refs set is Record types of Mems. If that's the case, we can either call `setElementRefs` there (instead of `onModuleClose`), or add a binding for it.

core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -165,15 +158,8 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
...
-    }
-
-    // We need to call forceName and onModuleClose on all of the sub-elements
-    // of the io bundle, but NOT on the io bundle itself.
-    // Doing so would cause the wrong names to be assigned, since their parent
-    // is now the module itself instead of the io bundle.
-    for (id <- getIds; if id ne _io) {
-      id._onModuleClose
+      // We have to force override the _ref because it was set during IO binding
+      port.setRef(ModuleIO(this, _namespace.name(name)), force = true)
```
jackkoenig, line 162: The diff here is really screwed up. All I did was delete the `for` loop calling `id._onModuleClose` (and it's associated comment).

jackkoenig:
@ducky64 in particular, I'd appreciate if you'd take a look at the 2 new commits. The last commit is related to https://github.com/freechipsproject/chisel3/pull/1624#issuecomment-711415558, I think we basically either:
* Call `setElementRefs` when building `Mems` and `SyncReadMems` (the current approach), OR
* Create a binding for the types of mems, similar to `SampleElementBinding` in for Vec sample elements

jackkoenig:
Another commit that highlighted a problem with the work to remove `_onModuleClose` (again, related to the goal in #1624). 
Basically, you can ask for `.instanceName` of non-hardware `Data`. These values are kind of wrong because they don't correspond to anything in the emitted FIRRTL, but it's behavior we have preserved (see https://github.com/freechipsproject/chisel3/pull/853) because it's used by DSPTools (even though it's ignored for non-hardware values). 

If we stop do `setElementRefs` in `_onModuleClose`, then non-hardware `Record` elements don't have names anymore (Checked by `InstanceNameSpec`). Providing names for non-hardware seems like a terrible reason to keep the whole `onModuleClose` business, so I preserved the behavior, but _only when using the compiler plugin_. I made the error message specify this, but I want other people's opinions about this change @ducky64 @azidar @albert-magyar @seldridge. 

Furthermore, I think we should deprecate `instanceName` or at least deprecate it being legal for non-hardware. Perhaps we could just deprecate `instanceName` and point to `.toTarget`, although that also "works" for non-hardware Data so it's a bit awkward either way.

jackkoenig:
The failing test in `test-chisel-2_11` (no compiler plugin available) really highlights the problems I'm discussing here:
```
[info] - should work with bundle elements *** FAILED ***
[info]   chisel3.internal.ChiselException: Data is not bound to hardware in the circuit so has no name. To emulate old naming behavior, use chisel3-plugin when compiling this code.
[info]   at chisel3.internal.throwException$.apply(Error.scala:85)
[info]   at chisel3.internal.HasId$$anonfun$instanceName$1.apply(Builder.scala:224)
[info]   at chisel3.internal.HasId$$anonfun$instanceName$1.apply(Builder.scala:221)
[info]   at scala.Option.getOrElse(Option.scala:121)
[info]   at chisel3.internal.HasId$class.instanceName(Builder.scala:221)
[info]   at chisel3.Data.instanceName(Data.scala:277)
[info]   at chisel3.internal.HasId$class.pathName(Builder.scala:233)
[info]   at chisel3.Data.pathName(Data.scala:277)
[info]   at chiselTests.InstanceNameSpec$$anonfun$6.apply(InstanceNameSpec.scala:46)
[info]   at chiselTests.InstanceNameSpec$$anonfun$6.apply(InstanceNameSpec.scala:45)
[info]   ...
```

jackkoenig:
#1624 (Making Data GC-able) has definitely bled into this PR, but I think it's okay. The problem common to both PRs is basically "What are refs for in Data and when should they be set?" There seem to be 3 cases:
1. Any Data that is hardware
2. Non-hardware Data that shows up as types in the FIRRTL (Memory types)
3. Any Data (even non-hardware) that the user might as for its name

(1) is the obvious primary use case. I have a workaround for (2) but could do it differently (see https://github.com/freechipsproject/chisel3/pull/1616#issuecomment-713240725 above).

(3) is the problematic one. I'm loathe to break APIs, but I think this is a terrible reason to not be able to GC any Data. Alternatives to what I have currently implemented (which breaks the API in 2.11):
* Return something like `"<non-hardware>"` as the `instanceName` for non-hardware `Data`
* Call `setElementRefs` when calculating `elements` in `Bundles`. This would maintain the API for `Bundles` but it would still be broken for `Records` could be combined with the previous idea.

azidar:
> Return something like "<non-hardware>" as the instanceName for non-hardware Data

I like this option, and then for the 3.5 release, error. We should seriously think about deprecating instanceName and improving .toTarget, but I think that is a major refactor.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I'd say this is answered pretty well in https://github.com/freechipsproject/chisel3/pull/1616#issuecomment-714011111. I've reverted the changes discussed in that comment and will leave those improvements for future work.

jackkoenig:
I've reverted the more aggressive changes and will give them another try as part of #1624. This is more-or-less the same as it was when @ducky64 originally approved, with some added safety and tests as well as a minor improvement to the plugin naming.

jackkoenig:
Also, please don't "Please Merge" this, I want to squash it myself to remove the unnecessary commit messages that were reverted, but I don't want to rebase because I want to keep the history on this PR.

azidar (approved):

ducky64:
> (3) is the problematic one. I'm loathe to break APIs, but I think this is a terrible reason to not be able to GC any Data. Alternatives to what I have currently implemented (which breaks the API in 2.11):
> 
> * Return something like `"<non-hardware>"` as the `instanceName` for non-hardware `Data`
> * Call `setElementRefs` when calculating `elements` in `Bundles`. This would maintain the API for `Bundles` but it would still be broken for `Records` could be combined with the previous idea.

Probably a bit late now, but I think that generally explicit error indicators are preferred over an empty string - so on an unexpected situation, there's at least some output for debugging.

jackkoenig:
> Probably a bit late now

Not too late! I backed off those changes and left this PR as a fairly modest improvement. I’ll keep this in mind in future work.

PR #1617, opened 2020-10-13
Change prefix stack to List[String]

jackkoenig:
This allows for sharing of the prefix stacks between Data, and removes a
boxing per prefix stack element that is no longer necessary.

My intuition is that this should reduce memory by quite a bit and thus improve performance, but I'd like to try it out on complex designs before we merge.

This should improve memory and performance for a few reasons:
1. `construction_prefix` for each `Data` is now free (it formerly copied from ArrayBuffer to List)
2. Structural sharing of common prefix stacks and intermediate prefix stacks
3. No more `Either` in prefix stack, removes an object allocation per stack element 

While we do have to copy the prefix stack when we `reverse` for naming, this is a copy on naming of `Data` rather than on construction. In theory, most `Data` should not be named, but due to the fact that every `HasId` adds itself to the `ids` of its parent module, every `Data` does currently get named. Once that latter issue is resolved, this will result in far fewer copies of the prefix stack. It is still better with this change because the copies can be GCed right away rather than sticking around being referred to by each `Data`. We can also consider memoizing the prefix stack reversals.

### Contributor Checklist

#### Type of Improvement

  - performance improvement  

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Reduce memory overhead from better naming improvements in `v3.4.0`

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -99,7 +99,7 @@ private[chisel3] trait HasId extends InstanceId {
   private val construction_prefix: Prefix = Builder.getPrefix()
 
   // Prefix when the latest [[suggestSeed]] or [[autoSeed]] is called
-  private var prefix_seed: Prefix = List.empty[Either[String, Data]]
+  private var prefix_seed: Prefix = Nil
 
   // Post-seed hooks called to carry the suggested seeds to other candidates as needed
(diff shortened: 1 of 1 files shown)


azidar (approved):
As is LGTM. I don't think you need a test because you aren't changing any behavior.

PR #1624, opened 2020-10-18
Make Data GC-able

jackkoenig:
As you can see in the diff of `Builder.scala`, currently, every single `HasId` (which `Data`, `BaseModule` and `MemBase` extend) is added to the `_ids` of its parent Module. Thus, no `HasId` can ever be garbage collected. This is obviously terrible for memory use, especially because we create a lot of unnecessary `Data` objects that are never used. By changing the behavior so that `BaseModule` and `MemBase` are still added, but `Data` is only added upon binding, now `Data` can be GC-ed.

This PR also adds a new test-suite `noPluginTests`. This is essentially a Scala project that just recompiles some of the normal Chisel tests without the Chisel compiler plugin and runs them. I found that the tests `InstanceNameSpec.scala` failed without the plugin, so I made a minimal project that just uses symlinks to build it (and necessary dependencies) without the plugin. (Note: `ChiselTestUtilitiesSpec` simply got split out of `ChiselSpec` because I didn't want those tests to also run in `noPluginTests`).

There are 2 changes related to making GC-able:
1. Add MemTypeBinding
2. Add `closeUnboundIds`

To explain these, I need to touch on some of the internals of Chisel
* Refs are used for determining the "names" of things, mainly for emitting FIRRTL
* Refs of Record elements are set upon binding or by _onModuleClose
* _onModuleClose is only called on Records added to the _ids of a Module
* Data are only added to _ids of Module on binding (new behavior in this PR)

(1) Mem types are technically not hardware, but are Data needed for emitting FIRRTL, so I added a pseudo-binding for them.
For (2), there are some legacy APIs still in use (see `InstanceNameSpec`) that need the refs to be set to work. `closeUnBoundIds` ensures that at least any reflectively namable Records, even unbound, have `_onModuleClose` called on them. It's a little weird but at least it's a fairly self-contained way to handle this. If we deprecate and remove these legacy APIs (not part of this PR), we can ultimately remove it.

### Validation

I validated this with a simple Module to create lots of useless `Data`:
```scala
class MyModule(n: Int, gen: () => Data) extends Module {
  val io = IO(new Bundle {
    val in  = Input(UInt(8.W))
    val out = Output(UInt(8.W))
  })

  for (i <- 0 until n) {
    for (j <- 0 until n) {
      for (k <- 0 until n) {
        for (l <- 0 until n) {
          gen()
        }
      }
    }
  }

  io.out := io.in
}
```

Using `openjdk version "1.8.0_265"`, I validated these numbers on Feb 9, 2021.
Comparing this PR (commit 0a0d7c6aac4326f2127d6d95efa5a4e10c81946c) vs. master (commit 2ed343e2305b7c22000f3f46fa81d73a369907eb)

| n | gen | min `-Xmx` master | min `-Xmx` this PR |
| --- | --- | --- | --- |
| 50 | `Bool()` |  1270M | 6M |
| 30 | `new MyBundle` | > 500M | 10M |
| 50 | `Vec(4, Bool())` | > 2G | 10M |

### Results in Practice

(these measurements are old, I'm rerunning)

Resident Set Size (RSS) is the system measured memory use which includes the virtual machine as measured by `/usr/bin/time -v`

| design | min `-Xmx` master | RSS master (MiB) | min `-Xmx` this PR | RSS this PR (MiB) |
| --- | --- | --- | --- | --- |
| medium | 400M | 1034 | 300M | 930 |
| large | 3600M | 4857 | 2600M | 3652 |
| very large | 17G | 19394 | 14G | 16400 |

In summary, a reduction of 10-25% in total memory use. It's actually smaller than I expected but I think there are other issues in Chisel and a couple in the SiFive generator that when also fixed, will result in much bigger reductions. Regardless this is a pretty big improvement on its own.

### Contributor Checklist

#### Type of Improvement

  - performance improvement           
  - code refactoring                  

#### API Impact

No impact

No impact

(TODO fix Record field name sanitation bug)

#### Desired Merge Strategy

 - Squash

#### Release Notes

Decrease memory use of Chisel by 10%+

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -201,6 +201,11 @@ lazy val chisel = (project in file(".")).
     )
   )
 
+lazy val noPluginTests = (project in file ("no-plugin-tests")).
+  dependsOn(chisel).
+  settings(commonSettings: _*).
+  settings(chiselSettings: _*)
+
 lazy val docs = project       // new documentation project
   .in(file("docs-target")) // important: it must not be docs/
   .dependsOn(chisel)
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -209,6 +209,8 @@ package experimental {
     */
   // TODO: seal this?
   abstract class BaseModule extends HasId {
+    _parent.foreach(_.addId(this))
+
     //
     // Builder Internals - this tracks which Module RTL construction belongs to.
     //
@@ -380,6 +382,17 @@ package experimental {
       }
 
       names
+    }
+
+    /** Invokes _onModuleClose on HasIds found via reflection but not bound to hardware
+      * (thus not part of _ids)
+      * This maintains old naming behavior for non-hardware Data
+      */
+    private[chisel3] def closeUnboundIds(names: HashMap[HasId, String]): Unit = {
+      val idLookup = _ids.toSet
+      for ((id, _) <- names if !idLookup(id)) {
+        id._onModuleClose
+      }
     }
 
     /** Compatibility function. Allows Chisel2 code which had ports without the IO wrapper to
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
         if: matrix.scala == '2.12.13'
         run: sbt ++${{ matrix.scala }} docs/mdoc
       - name: Test
-        run: sbt ++${{ matrix.scala }} test
+        run: sbt ++${{ matrix.scala }} test noPluginTests/test
       - name: Binary compatibility
         run: sbt ++${{ matrix.scala }} mimaReportBinaryIssues
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -20,7 +20,8 @@ class AliasedAggregateFieldException(message: String) extends ChiselException(me
   * of) other Data objects.
   */
 sealed abstract class Aggregate extends Data {
-  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection) {
+  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
+    _parent.foreach(_.addId(this))
     binding = target
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
@@ -166,7 +167,8 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
     case _ => false
   }
 
-  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection) {
+  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
+    _parent.foreach(_.addId(this))
     binding = target
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -81,6 +81,8 @@ package experimental {
         id._onModuleClose
       }
 
+      closeUnboundIds(names)
+
       val firrtlPorts = getModulePorts map {port => Port(port, port.specifiedDirection)}
       val component = DefBlackBox(this, name, firrtlPorts, SpecifiedDirection.Unspecified, params)
       _component = Some(component)
--- a/core/src/main/scala/chisel3/Data.scala
+++ b/core/src/main/scala/chisel3/Data.scala
@@ -339,13 +339,14 @@ abstract class Data extends HasId with NamedComponent with SourceInfoDoc {
   private[chisel3] final def isSynthesizable: Boolean = _binding.map {
     case ChildBinding(parent) => parent.isSynthesizable
     case _: TopBinding => true
-    case _: SampleElementBinding[_] => false
+    case (_: SampleElementBinding[_] | _: MemTypeBinding[_]) => false
   }.getOrElse(false)
 
   private[chisel3] def topBindingOpt: Option[TopBinding] = _binding.flatMap {
     case ChildBinding(parent) => parent.topBindingOpt
     case bindingVal: TopBinding => Some(bindingVal)
     case SampleElementBinding(parent) => parent.topBindingOpt
+    case _: MemTypeBinding[_] => None
   }
 
   private[chisel3] def topBinding: TopBinding = topBindingOpt.get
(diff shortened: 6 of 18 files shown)


jackkoenig:
We definitely need a test here because it looks like Chisel `v3.4.0` changed the behavior of unsanitary names:
```scala
  val foo = UInt(8.W)
  val `foo😎` = UInt(8.W)
  val `foo😎😎` = UInt(8.W)
}

  val in = IO(Input(new MyBundle))
  val out = IO(Output(new MyBundle))
  
  out := in
}
```

In `v3.3.2` (https://scastie.scala-lang.org/RKOm3m74SDqy7HlZ18PPCA)
```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : { foo : UInt<8>, foouD83DuDE0E : UInt<8>, foouD83DuDE0EuD83DuDE0E : UInt<8>}
    output out : { foo : UInt<8>, foouD83DuDE0E : UInt<8>, foouD83DuDE0EuD83DuDE0E : UInt<8>}
  
    out.foouD83DuDE0EuD83DuDE0E <= in.foouD83DuDE0EuD83DuDE0E
    out.foouD83DuDE0E <= in.foouD83DuDE0E
    out.foo <= in.foo
```

In `v3.4.0` (https://scastie.scala-lang.org/rEzBz9LmTe2uRrV3lgiC0Q)
```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : { foo : UInt<8>}
    output out : { foo : UInt<8>}
    out.foo <= in.foo @[main.scala 16:7]
```

jackkoenig:
Possible solutions:
* Keep this approach and sanitize on creation of FIRRTL types
* Require that the names of `elements` are sanitary (API change)
* Add binding for type of Mems (it's kind of similar to the SampleElementBinding, except the parent is a `MemBase` instead of a `Vec[_]`)
* Set refs for Mem types during mem construction (related to discussion in #1616)

jackkoenig:
This is ready to go and the PR description has been updated.

build.sbt
```diff
@@ -201,6 +201,11 @@ lazy val chisel = (project in file(".")).
     )
   )
 
+lazy val noPluginTests = (project in file ("no-plugin-tests")).
```
mwachs5, line 204: why was it necessary to add this `noPluginTests` as part of this PR?

core/src/main/scala/chisel3/Module.scala
```diff
@@ -382,6 +384,17 @@ package experimental {
       names
     }
 
+    /** Invokes _onModuleClose on HasIds found via reflection but not bound to hardware
```
mwachs5, line 387: what is an example of a HasIds that would be found via reflection but not bound to hardware?

build.sbt
jackkoenig: I noticed that all of the tests passed, but I had a hunch that things wouldn't work without it. Hunch turned out to be correct so I wanted to make sure the no plugin behavior was supported.

mwachs5:
"We definitely need a test here because it looks like Chisel v3.4.0 changed the behavior of unsanitary names:" -- I'm confused what this has to do with this PR. I mean, the described behavior does seem buggy with the `foo sunglasses` stuff, but is that still relevant to this PR? Has that issue been resolved elsewhere?

core/src/main/scala/chisel3/Module.scala
jackkoenig: It doesn't show up in the diff because all I did was make a symlink so that the same file is compiled with and without the plugin, but from [InstanceNameSpec](https://github.com/chipsalliance/chisel3/blob/2ed343e2305b7c22000f3f46fa81d73a369907eb/src/test/scala/chiselTests/InstanceNameSpec.scala#L9):
```scala
class InstanceNameModule extends Module {
  // ...
  val z = new Bundle {
    val foo = UInt(8.W)
  }
  // ...
}
```
It's unused, but after elaborating that module you can ask for `z.foo.pathName`.

mwachs5: ah yeah, we talked about that IRZoomL. Ok, so this is handling the case that we said was weird, but people are apparently doing it, and we don't have a great way of deprecating it

jackkoenig:
> "We definitely need a test here because it looks like Chisel v3.4.0 changed the behavior of unsanitary names:" -- I'm confused what this has to do with this PR. I mean, the described behavior does seem buggy with the foo sunglasses stuff, but is that still relevant to this PR? Has that issue been resolved elsewhere?

That issue is no longer relevant to this PR. A previous implementation involved changing how we get the names in the emitted FIRRTL from the Chisel objects which raised the issue of name sanitation. The issue is unfortunately not resolved, so I'll file an issue.

mwachs5 (approved):
I wouldn't consider my approval sufficient (or necessary), but thanks for answering my questions!

jackkoenig:
I can confirm that with a backport to `3.4.x`, this passes SiFive's internal regressions (and we have lots of funky, legacy code)

core/src/main/scala/chisel3/Module.scala
```diff
@@ -382,6 +384,17 @@ package experimental {
...
+      * (thus not part of _ids)
+      * This maintains old naming behavior for non-hardware Data
+      */
+    private[chisel3] def closeUnboundIds(names: HashMap[HasId, String]): Unit = {
+      val idLookup = _ids.toSet
+      for ((id, _) <- names if !idLookup(id)) {
+        id._onModuleClose
+      }
+    }
```
seldridge, line 396: It would seem like this is undoing the work of this PR by promoting `HasId`s which we don't care about to not be GC-able by unnecessarily linking them to their parent module. 

What's going on here? I expect I'm missing something...

seldridge (approved):
This looks fine and good job preserving the old weird behavior (which I think we should remove at some point).

One question about how `closeUnboundIds` isn't undoing the work of this PR.

core/src/main/scala/chisel3/Module.scala
jackkoenig: This doesn't actually undo the work of this PR for 2 reasons:

1. it doesn't actually make any `HasIds` not GC-able. [_onModuleClose](https://github.com/chipsalliance/chisel3/blob/2ed343e2305b7c22000f3f46fa81d73a369907eb/core/src/main/scala/chisel3/Aggregate.scala#L648) in Record sets the refs of the elements of an unbound Record. This just provides references from the elements of the `Record` to the Record itself, there's a circular dependency there but it doesn't involve the Module at all which is where the real non-GCable Data comes from.
2. Because these are the `names` that come from reflective naming, the Module has a reference to each of these `HasIds` *anyway*, thus they aren't GC-able.

A follow on question is why not just add them to `_ids` and let the `for (id <- getIds) {` logic above handle calling `_onModuleClose`? I thought about it, but my reasoning is that it's clearer Iif `_ids` only contains the `HasIds` we care about and instead this logic (which is maintaining and old, questionable API) is intentionally kept separate. That's my thinking at least.

PR #1630, opened 2020-10-21
Make `-e` option work with ChiselStage methods

### LINKED ISSUE #1629: One-file-per-module is incompatible with ChiselStage.emit*
The following fails and it shouldn't:

```scala
import chisel3._

class Foo extends Module {
  val io = IO(new Bundle {
    val in = Input(Bool())
    val out = Output(Bool())
  })

  io.out := ~io.in
}

println((new ChiselStage).emitVerilog(new Foo, Array("-e", "verilog")))
```

This does a `None.get` because the `-e` results in one-file-per-module and the match is narrowly looking for an `EmittedCircuitAnnotation`. 

Change the behavior here so that this works with `-e` by appending the individual emitted modules together.

**Type of issue**: bug report

**Other information**
 

https://gitter.im/freechipsproject/chisel3?at=5f90959ca7e77a0ff1602e6e

Compile the example above.

`None.get` exception.

No match error.

3.4.0 

This is an extremely confusing error for an end user...

seldridge:
Prevent a `None.get` where a user passes a `-e` (one-file-per-module) option to one of the `ChiselStage` methods. This was previously checking for the only a circuit, but throwing in `-e` would cause modules to be emitted. Append all the modules together and return this string.

Fixes #1629 

### Contributor Checklist

- [n/a] Did you add Scaladoc to every public function/method?
- [n/a] Did you add appropriate documentation in `docs/src`?

#### Type of Improvement

- bug fix

#### API Impact

None.

None.
#### Desired Merge Strategy

#### Release Notes

None.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/stage/ChiselStageSpec.scala
+++ b/src/test/scala/chiselTests/stage/ChiselStageSpec.scala
@@ -4,6 +4,7 @@ package chiselTests.stage
 
 import chisel3._
 import chisel3.stage.ChiselStage
+import chisel3.testers.TesterDriver.createTestDirectory
 
 import chiselTests.Utils
 
@@ -52,8 +53,9 @@ class ChiselStageSpec extends AnyFlatSpec with Matchers with Utils {
   }
 
   it should "return a flattened FIRRTL string with '-e high'" in {
+    val args = Array("-e", "high", "-td", createTestDirectory(this.getClass.getSimpleName).toString)
     (new ChiselStage)
-      .emitFirrtl(new Foo, Array("-e", "high", "-td", "test_run_dir/ChiselStageSpec")) should include ("module Bar")
+      .emitFirrtl(new Foo, args) should include ("module Bar")
   }
 
   behavior of "ChiselStage.emitVerilog"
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/stage/ChiselStageSpec.scala
```diff
@@ -40,7 +48,12 @@ class ChiselStageSpec extends AnyFlatSpec with Matchers with Utils {
   behavior of "ChiselStage.emitFirrtl"
 
   it should "return a High FIRRTL string" in {
-    ChiselStage.emitFirrtl(new Foo) should include ("mem bar")
+    ChiselStage.emitFirrtl(new Foo) should include ("mem memory")
+  }
+
+  it should "return a flattened FIRRTL string with '-e high'" in {
+    (new ChiselStage)
+      .emitFirrtl(new Foo, Array("-e", "high", "-td", "test_run_dir/ChiselStageSpec")) should include ("module Bar")
```
jackkoenig, line 56: I'd rather this not hardcode `test_run_dir`

seldridge: Switched to `createTestDirectory`. I've hardcoded this a lot in the past... 😬

jackkoenig: I'm watching you 🕵️

jackkoenig (approved):

PR #1634, opened 2020-10-23
Added Force Name API

azidar:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

Adds a new API - forcename.

This only changes generated Verilog names if the API is used. Otherwise, no changes.

#### Desired Merge Strategy

#### Release Notes

New 'forcename' API enables enforcing the naming of signals or instances, even if a FIRRTL transform renames them.

- [x] Did you review?


docs/src/cookbooks/naming.md
```diff
@@ -60,3 +60,36 @@ class ExampleNoPrefix extends MultiIOModule {
...
+
+class WrapperExample extends MultiIOModule {
+  val IN = IO(Input(UInt(3.W)))
+  val OUT = IO(Output(UInt(3.W)))
+  val inst = Module(new Wrapper)
+  inst.IN := IN
+  OUT := inst.OUT
+}
+class Wrapper extends MultiIOModule with InlineInstance {
+  val IN = IO(Input(UInt(3.W)))
```
chick, line 79: Quibble: All caps names are against (I think) standard style.

hcook (approved):
I approve of the API and believe this will address my use case.

chick (approved):
I think this looks fine (with one quibble about the mdocs)
Based on the commits in it perhaps it needs a rebase

docs/src/cookbooks/naming.md
jackkoenig: I agree with Chick here, would prefer lowerCamelCase (just all lower case in this case)

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+
+import scala.collection.mutable
+
+object forceName {
+
+  /** Force the name of this signal
+    * @param signal Signal to name
+    * @param name Name to force to
+    */
+  def apply(signal: => chisel3.Bits, name: String): Unit = {
```
jackkoenig, line 25: `Bits` or can we do `Element`?
```suggestion
  def apply[T <: chisel3.Element](signal: T, name: String): T = {
```
A few things in this suggestion:
* `Element` over `Bits` - Allow for marking `Clock`, `Reset`, and `Analog`.
* eager vs. by-name parameter - Because this doesn't evaluate the signal, bad things could happen if you had hardware in this. That being said, since we don't currently return it, it's less likely.
* Returning the `signal` - helpful in not having to have an extra Expression

Not saying all of these are needed, just think all 3 are worth considering.

jackkoenig: Sam suggestions for all APIs below

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+
+  // Errors if renaming to multiple targets
+  override def update(renames: RenameMap): Seq[Annotation] = {
+    renames.get(target) match {
+      case None => List(this)
+      case Some(newTargets) if newTargets.size > 1 =>
+        throw CustomTransformException(
+          new FirrtlUserException(
+            s"Cannot force the name of $target to $name because it is renamed to $newTargets." +
+              " Perhaps $target is not a ground type?"
```
jackkoenig, line 88: This doesn't happen that often, but what if an annotated module is duplicated? Perhaps there should be different annotations for Modules vs. Data.

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    val lookup: String => List[List[(Instance, OfModule)]] =
+      str => graph.findInstancesInHierarchy(str)
+        .view
+        .map(_.map(_.toTokens).toList)
+        .toList
+    allInstancePaths(lookup) _
+  }
+
+  /** Returns a function which returns all instance paths to a given IsModule
+    * Uses Lists because we want to recurse on inner List (head :: tail)-style
```
jackkoenig, line 122: ScalaDoc style, the first line should be a short summary, followed by a blank comment line, followed by longer description. This is just a very long description.

jackkoenig: Also this justification for the use of `List` doesn't really bear out in the uses below. `List` is good but I would drop the justification.

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+  /** Returns the [[IsModule]] which is referred to, or if a [[ReferenceTarget]], the enclosing [[IsModule]]
+    * @param a signal/instance/module
+    * @return referring IsModule
+    */
+  def referringIsModule(a: IsMember): IsModule = a match {
+    case b: ModuleTarget    => b
+    case b: InstanceTarget  => b.targetParent
+    case b: ReferenceTarget => b.copy(component = Nil).targetParent.asInstanceOf[IsModule]
+  }
+
```
jackkoenig, line 107: All of these utilities seem pretty generic, should we add them to FIRRTL? That shouldn't preclude using them here but seems like a more natural place. I think maybe making them private or package private so that we don't introduce a new public API while also introducing them to FIRRTL makes more sense.

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+            }
+            seen(path) = map
+          }
+        case _ =>
+      }
+      allNameMaps.map {
+        case (isModule, nameMap) => Target.referringModule(isModule).module -> nameMap
+      }
+    }
+    if (renames.nonEmpty) { Some(renames) } else None
```
jackkoenig, line 182:
```suggestion
    if (renames.nonEmpty) Some(renames) else None
```

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    * @return
+    */
+  def buildForceNameMap(state: CircuitState,
+                     igraph: => InstanceKeyGraph
+                    ): Option[Map[String, Map[String, String]]] = {
+    val forceNames = state.annotations.collect { case f: ForceNameAnnotation => f }
+    val badNames = mutable.HashSet[ForceNameAnnotation]()
+    val allNameMaps = forceNames.groupBy { case f => referringIsModule(f.target) }.mapValues { value =>
+      value.flatMap {
+        case f@ForceNameAnnotation(rt: ReferenceTarget, name) if rt.component.nonEmpty =>
```
jackkoenig, line 154:
```suggestion
        case f @ ForceNameAnnotation(rt: ReferenceTarget, name) if rt.component.nonEmpty =>
```

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    * @param state CircuitState to operate on
+    * @param igraph built instance key graph from state's circuit
+    * @return
+    */
+  def buildForceNameMap(state: CircuitState,
+                     igraph: => InstanceKeyGraph
+                    ): Option[Map[String, Map[String, String]]] = {
+    val forceNames = state.annotations.collect { case f: ForceNameAnnotation => f }
+    val badNames = mutable.HashSet[ForceNameAnnotation]()
+    val allNameMaps = forceNames.groupBy { case f => referringIsModule(f.target) }.mapValues { value =>
```
jackkoenig, line 152: `mapValues` returns a `MapView` which means it will rerun this closure each time you access each value. Since you're not even using `allNameMaps` as a `Map` below, I'd suggest casting to a `Seq` or something, but at the very least you need to force to an eager collection.

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    renameMap: RenameMap,
+    ct: CircuitTarget, igraph: InstanceKeyGraph
+  )(mod: DefModule): DefModule = {
+
+    val mt = ct.module(mod.name)
+    val instToOfModule = mutable.HashMap[String, String]()
+    val names = modToNames.getOrElse(mod.name, Map.empty[String, String])
+    // Need to find WRef referring to mems for prefixing
+    def onExpr(expr: Expression): Expression = expr match {
+      case wref @ WRef(n, _,_,_) if names.contains(n) =>
```
jackkoenig, line 226:
```suggestion
      case ref @ Reference(n, _,_,_) if names.contains(n) =>
```
FIRRTL 1.4!

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+        instToOfModule(inst.name) = inst.module
+        val newName = names(inst.name)
+        renameMap.record(mt.instOf(inst.name, inst.module), mt.instOf(newName, inst.module))
+        inst.copy(name = names(inst.name))
+      case inst: DefInstance =>
+        instToOfModule(inst.name) = inst.module
+        inst
+      case error @ (_: WDefInstanceConnector) =>
+        throw new Exception(s"Unexpected instance object $error")
+      // Prefix SeqMems because they result in Verilog modules
```
jackkoenig, line 248: As `WDefInstanceConnector` is kind of a FIRRTL implementation detail, is this necessary?

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    def onPort(port: Port): Port = {
+      if (names.contains(port.name)) {
+        renameMap.record(mt.ref(port.name), mt.ref(names(port.name)))
+        port.copy(name = names(port.name))
+      } else port
+    }
+
+    val childInstanceHasRename = igraph.getChildInstanceMap(OfModule(mod.name)).valuesIterator.collectFirst {
+        case o if modToNames.contains(o.value) => o
+      }.isDefined
```
jackkoenig, line 263:
```suggestion
    val childInstanceHasRename = igraph.getChildInstanceMap(OfModule(mod.name)).exists(o => modToNames.contains(o.value))
```
Was there a reason for the `valuesIterator`?

src/test/scala/chiselTests/experimental/ForceNames.scala
```diff
@@ -0,0 +1,108 @@
...
+import chisel3.stage.{ChiselGeneratorAnnotation, ChiselStage}
+import chisel3.util.experimental.{ForceNameAnnotation, ForceNamesTransform, InlineInstance, forceName}
+import firrtl.annotations.{Annotation, ReferenceTarget}
+import firrtl.options.{Dependency, TargetDirAnnotation}
+import firrtl.stage.RunFirrtlTransformAnnotation
+import logger.{LogLevel, LogLevelAnnotation}
+
+/** Object containing Modules used for testing */
+object ForceNamesHierarchy {
+  class WrapperExample extends MultiIOModule {
```
jackkoenig, line 17: Similar to Chick's comment above, would prefer names in line with Scala-style below.

src/main/scala/chisel3/util/experimental/ForceNames.scala
```diff
@@ -0,0 +1,290 @@
...
+    * @param renameMap Record renames
+    * @param ct Circuit target
+    * @param igraph InstanceKeyGraph of parent circuit
+    * @param mod Enclosing module of the names to force
+    * @return
+    */
+  private def forceNamesInModule(
+    modToNames: Map[String, Map[String, String]],
+    renameMap: RenameMap,
+    ct: CircuitTarget, igraph: InstanceKeyGraph
```
jackkoenig, line 218:
```suggestion
    ct: CircuitTarget,
    igraph: InstanceKeyGraph
```

jackkoenig (changes requested):
Overall looks good. Lots of nits and please rebase.

colinschmidt:
Is there a reason to not expect a new firttl `ForceRename` in the future that overrides this?
We've already done `suggestName`->`Rename`->`ForceName` so it seems like there is an inherent conflict between the designer wanting a certain name and transform writers wanting a different name.
I'm not opposed to this feature and would probably use it but just interested if there is some deeper problem of "who owns naming" here.

src/main/scala/chisel3/util/experimental/ForceNames.scala
azidar: So I casted it to a Seq, and then back to a Map because its the return value. Do you have a better suggestion?

src/main/scala/chisel3/util/experimental/ForceNames.scala
azidar: I now check if its a referencetarget.

azidar:
@colinschmidt This is the first time we've exposed a user-facing API which uses transforms to rename names. It is has the strongest semantics, meaning basically error if I can't do it. I'd see this as a core utility, so we wouldn't introduce a second way for users to rename things via a FIRRTL transform.

In general, it's best not to use this utility (it should be a last resort). However, in some cases you need backwards compatibility, and this enables more control/enforcement of names that was not supported by the existing naming mechanisms.

src/main/scala/chisel3/util/experimental/ForceNames.scala
jackkoenig: That's fine, or you can instead just to `.toMap` instead of `.toSeq` and keep it as a `Map` the whole time, I didn't realize it was the return value.

jackkoenig (approved):
LGTM!

PR #1635, opened 2020-10-23
Bugfix - module name collision for injecting aspect

azidar:
### Contributor Checklist

#### Type of Improvement

 - bug fix

#### API Impact

Adds private [chisel3] APIs to enable fixing module name collision behavior.

None.

#### Desired Merge Strategy

 - Squash: The PR will be squashed and merged (choose this if you have no preference.)

#### Release Notes

Bugfix: Instantiating modules in an InjectingAspect will no longer create modules with conflicting names in the original design.

- [ ] Did you review?


core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -627,7 +627,11 @@ private[chisel3] object Builder {
 
 
   def build[T <: RawModule](f: => T): (Circuit, T) = {
-    dynamicContextVar.withValue(Some(new DynamicContext())) {
+    build(f, new DynamicContext())
+  }
+
+  private [chisel3] def build[T <: RawModule](f: => T, dynamicContext: DynamicContext): (Circuit, T) = {
```
ducky64, line 633: Is this the right API, to expose all of dynamicContext, as opposed to perhaps something more targeted like preinitialized / reserved names?

core/src/main/scala/chisel3/Module.scala
```diff
@@ -251,7 +251,7 @@ package experimental {
 
     /** Legalized name of this module. */
     final lazy val name = try {
-      Builder.globalNamespace.name(desiredName)
+      if(this.isInstanceOf[ModuleAspect]) desiredName else Builder.globalNamespace.name(desiredName)
```
ducky64, line 254: A comment explaining why this is bypassing the normal naming process might be useful here?

src/main/scala/chisel3/aop/injecting/InjectingAspect.scala
```diff
@@ -42,17 +42,23 @@ abstract class InjectorAspect[T <: RawModule, M <: RawModule](
     injection: M => Unit
 ) extends Aspect[T] {
   final def toAnnotation(top: T): AnnotationSeq = {
-    toAnnotation(selectRoots(top), top.name)
+    val moduleNames = Select.collectDeep(top) { case i => i.name }.toSeq
+    toAnnotation(selectRoots(top), top.name, moduleNames)
   }
 
-  final def toAnnotation(modules: Iterable[M], circuit: String): AnnotationSeq = {
+  final def toAnnotation(modules: Iterable[M], circuit: String, moduleNames: Seq[String]): AnnotationSeq = {
```
ducky64, line 49: Scaladoc for this might be useful? Is there supposed to be some parallel structure between modules and moduleNames?

src/main/scala/chisel3/aop/injecting/InjectingAspect.scala
```diff
@@ -42,17 +42,23 @@ abstract class InjectorAspect[T <: RawModule, M <: RawModule](
...
+    val dynamicContext = new DynamicContext()
+    moduleNames.foreach { n =>
```
ducky64, line 51: A comment for what this is doing would be nice?

src/test/scala/chiselTests/aop/InjectionSpec.scala
```diff
@@ -3,42 +3,47 @@
...
+
+  class AspectTester(results: Seq[Int]) extends BasicTester {
+    val values = VecInit(results.map(_.U))
+    val counter = RegInit(0.U(results.length.W))
+    counter := counter + 1.U
+    when(counter >= values.length.U) {
+      stop()
+    }.otherwise {
+      when(reset.asBool() === false.B) {
+        printf("values(%d) = %d\n", counter, values(counter))
```
ducky64, line 38: prints as part of unit tests generally not recommended?

src/test/scala/chiselTests/aop/InjectionSpec.scala
```diff
@@ -67,6 +72,15 @@ class InjectionSpec extends ChiselFlatSpec {
     }
   )
 
+  val duplicateSubmoduleAspect = InjectingAspect(
+    {dut: SubmoduleManipulationTester => Seq(dut)},
+    {_: SubmoduleManipulationTester =>
+      val moduleSubmoduleA2 = Module(new SubmoduleA)
```
ducky64, line 78: Is this supposed to be the name conflict? If so, I don't see it?

ducky64 (changes requested):
I think more comments would be very helpful, it's not obvious what's going on and why just by looking at code...

core/src/main/scala/chisel3/internal/Builder.scala
azidar: So I made the function package private, which should give us the ability to change it in the future. However, I think it actually does make sense to expose the dynamic context, because I've often hit cases with the Instance API that I've been wanting ways to add context-state to a chisel execution.

jackkoenig:
@ducky64 can you re-review?

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: 👍

src/test/scala/chiselTests/aop/InjectionSpec.scala
ducky64: Oh I see, it's taking the module name instead of the val name?

ducky64 (approved):
lgtm, the docs definitely helped clarify behavior!

PR #1645, opened 2020-10-30
Fix bug where refs may not get set for Records

jackkoenig:
This requires a combination of things, but it happens to be a
combination used by Diplomacy in Rocket Chip. It must be a Record in
compatibility code with Vecs as fields and a mix of components with and
without set directions.

Bug introduced in https://github.com/freechipsproject/chisel3/pull/1616, not yet released so no need for release notes

h/t to @azidar for helping me debug this, I seriously thought I was losing my mind. I never use `try-catch` to recover from exceptions (only to report more precise errors) so I just wasn't expecting it to be skipping over a critical line of code and continuing. Might be better to replace this pattern with a more explicit `Try` but in any case the fix is simple.

### Contributor Checklist

#### Type of Improvement

   - bug fix

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -493,7 +493,6 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
     try {
       super.bind(target, parentDirection)
-      setElementRefs()
     } catch {  // nasty compatibility mode shim, where anything flies
       case e: MixedDirectionAggregateException if !compileOptions.dontAssumeDirectionality =>
         val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -503,6 +502,7 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
           case _ => ActualDirection.Bidirectional(ActualDirection.Default)
         }
     }
+    setElementRefs()
```
seldridge, line 505: It should be equivalent, but the above could be put in a `finally` block. (Not sure if that's clearer in any way or there's any real benefit to do ing so.)

seldridge (approved):
Nice find! Thanks for figuring this out and for the @azidar help.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I thought about it, IMO think `finally` implies that the `catch` block rethrows (which it doesn't), but I can see the argument.

PR #1653, opened 2020-11-03
Add boring through another module

azidar:
### Contributor Checklist

- [ ] Did you review?


docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+```scala mdoc
+import chisel3._
+class Constant extends MultiIOModule {
+  val x = Wire(UInt(6.W))
+  x := 42.U
+}
+class Expect extends MultiIOModule {
+  val y = Wire(UInt(6.W))
+  y := 0.U
+  // This assertion will fail unless we bore!
```
mwachs5, line 29:
```suggestion
  // This assertion will fail unless we find a way to drive it!
```

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+  chisel3.assert(y === 42.U, "y should be 42 in module Expect")
+}
+```
+
+We can then connect `x` to `y` using [[BoringUtils]] without modifiying the Chisel IO of `Constant`, `Expect`, or
```
mwachs5, line 34:
```suggestion
We can then drive `y` with `x` using [[BoringUtils]] without modifiying the Chisel IO of `Constant`, `Expect`, or
```

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+modules that may instantiate them. There are two approaches to do this:
+
+1. Hierarchical boring using [[BoringUtils.bore]]
+
+2. Non-hierarchical boring using [[BoringUtils.addSink]]/[[BoringUtils.addSource]]
+
+### Hierarchical Boring
+
+Hierarchical boring involves connecting one sink instance to another source instance in a parent module. Below,
```
mwachs5, line 43: I'm not quite sure what "instance" means in the above sentence. To me an instance is almost always an instance of a module. Do you mean a wire? Or do you mean a module instance?

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+module `Top` contains an instance of `Constant` and `Expect`. Using [[BoringUtils.bore]], we can connect
```
mwachs5, line 44:
```suggestion
module `Top` contains an instance of `Constant`, called `constant`,  and an instance of `Expect` called `expect`. Using [[BoringUtils.bore]], we can connect
```

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+In addition, you can specify additional modules to route the connection through:
+
+```scala mdoc
+class Dummy extends MultiIOModule { }
+class TopThroughModules extends MultiIOModule {
+  val constant = Module(new Constant)
+  val dummy1 = Module(new Dummy)
+  val dummy2 = Module(new Dummy)
+  val expect = Module(new Expect)
+  BoringUtils.bore(constant.x, Seq(expect.y), Seq(("first", dummy1), ("second", dummy2)))
```
mwachs5, line 65: what is the significance of the strings "first" and "second"?

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+}
+println(ChiselStage.emitVerilog(new TopThroughModules))
+```
+
+### Non-hierarchical Boring
+
+Non-hierarchical boring involves connections from sources to sinks that cannot see each other. Here, `x` is
+described as a source and given a name, `uniqueId`, and `y` is described as a sink with the same name. This is
+equivalent to the hierarchical boring example above, but requires no modifications to `Top`.
```
mwachs5, line 74:
```suggestion
Non-hierarchical boring involves connections from sources to sinks that cannot both be referenced
within the same module. In this example, we build up a source and sink pair called `uniqueId`.
`x` is described as a source and associated with the name `uniqueId`, and `y` is described as a sink with the same name. This is
equivalent to the hierarchical boring example above, but requires no modifications to `Top`.
```

Does it always have to be a pair, or can it be a broadcast / gather? What happens if you try a many-to-many or many-to-one or one-to-many?

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+}
+println(ChiselStage.emitVerilog(new Top))
+```
+
+### Additional Comments
+
+Both hierarchical and non-hierarchical boring emit FIRRTL annotations that describe sources and sinks. These are
+matched by a `name` key that indicates they should be wired together. Hierarchical boring safely generates this name
+automatically. Non-hierarchical boring unsafely relies on user input to generate this name. Use of non-hierarchical
+naming may result in naming conflicts that the user must handle.
```
mwachs5, line 104: same question ... must there be exactly one source and one sink for each named pair? this isn't obvious from the text as written

docs/src/cookbooks/boringutils.md
```diff
@@ -0,0 +1,107 @@
...
+
+The automatic generation of hierarchical names relies on a global, mutable namespace. This is currently persistent
```
mwachs5, line 106: what does "This is currently persistent across circuit elaborations" mean?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -45,8 +49,8 @@ class BoringUtilsException(message: String) extends Exception(message)
   *
   * ===Hierarchical Boring===
   *
-  * Hierarchcical boring involves connecting one sink instance to another source instance in a parent module. Below,
-  * module `Top` contains an instance of `Cosntant` and `Expect`. Using [[BoringUtils.bore]], we can connect
+  * Hierarchical boring involves connecting one sink instance to another source instance in a parent module. Below,
```
mwachs5, line 52:
```suggestion
  * Hierarchical boring involves driving a signal in a sink instance from a signal in a source instance, when both instances are in the same parent module. Below,
```

same question... what exactly is an instance? I might rephrase as above...

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -45,8 +49,8 @@ class BoringUtilsException(message: String) extends Exception(message)
...
+  * module `Top` contains an instance of `Constant` and `Expect`. Using [[BoringUtils.bore]], we can connect
   * `constant.x` to `expect.y`.
```
mwachs5, line 54:
```suggestion
  * module `Top` contains an instance of `Constant` and `Expect`. Using [[BoringUtils.bore]], we can drive
  * `expect.y` from `constant.x`.
```

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -191,4 +194,46 @@ object BoringUtils {
     genName
   }
 
+  /** Connect a source to one or more sinks, through a list of modules
+   * @param source a source component
+   * @param sinks one or more sink components
+   * @param through a list of wire names, the modules to instantiate them in,
```
mwachs5, line 200: ah, the "list of wire names" would be helpful to explain in the tutorial

azidar:
@jackkoenig can i have a review? i think we should merge as is if possible.

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -5,18 +5,41 @@ package chisel3.experimental
...
+
+  /** Conversion to FIRRTL AnnotationSeq */
+  protected def toAnnotationSeq: AnnotationSeq = Nil
+
+  private[chisel3] def convert: AnnotationSeq = {
+    toFirrtl match {
+      case EmptyAnnotation => toAnnotationSeq
+      case other => Seq(other)
+    }
+  }
```
jackkoenig, line 42: I'm down with the idea, just wondering about the migration path here. If we deprecate toFirrtl yet it's left as a virtual method, then every implementation of ChiselAnnotation will have a deprecation warning yet still be required to implement the method. We might consider either adding another trait and either leaving ChiselAnnotation alone or deprecating it. We could alternatively do an DependencyAPIMigration-style thing like in FIRRTL which implements the deprecated method for you.

PR #1660, opened 2020-11-06
Refine autonaming to have more intuitive behavior

jackkoenig:
Last name in an Expression wins, while the first Statement to name wins.
This is done via checking the _id of HasIds during autonaming and only
applying a name if the HasId was created in the scope of autonaming.
There is no change to `.autoSeed` or `.suggestName` behavior.

Behavior of chisel3-plugins from before this change is maintained. More precisely, assuming this PR is released with Chisel v3.4.1, if you use Chisel v3.4.1 with the v3.4.0 (or v3.4.0-RC3) plugin, you'll get the v3.4.0 behavior. 

The bulk of the work here was making `unapply` continue to work because it previously worked via overriding names. Now, a signal will only be named by `autoNameRecursively` or `autoNameRecursivelyProduct` calls that wrap the `Expression` within which the `Data` is created.

Fixes https://github.com/freechipsproject/chisel3/issues/1603

### Contributor Checklist

#### Type of Improvement

  - bug fix
  - new feature/API    

#### API Impact

This changes the naming behavior to what I've been calling "Outer Expression, First Statement" wins. For example:

```scala
  val width: Option[Int] = Some(8)
  val debug = width.map { w =>
    val foo = IO(Output(UInt(w.W))) // name: "debug_foo"
    val delay = RegNext(in) // name: "debug_delay"
    foo := delay
    foo
  } // name: "debug"
  val bar = debug // name: "bar"
```
Intuitively, the name of this port should be `debug`. It is useful that things built within the anonymous function should have `debug` as a prefix in addition to their val name (eg. `debug_delay`), but because `foo` is returned, the outer name should win (`debug`). It is also intuitive that `debug` wins over `bar` since `debug` is the name of the declaration rather than an alias.

Prior to this PR, the port would be named `debug_foo` and if it were a non-port, it would be named `bar`. This PR unifies the behavior of ports and non-ports into what is (in my opinion) more intuitive.

Change in naming algorithm

#### Desired Merge Strategy

   - Squash

#### Release Notes

Change naming priority to "Outer Expression, First Statement"

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -59,6 +59,7 @@ private[chisel3] class IdGen {
     counter += 1
     counter
   }
+  def value: Long = counter
 }
 
 /** Public API to access Node/Signal names.
@@ -119,7 +120,9 @@ private[chisel3] trait HasId extends InstanceId {
     * @param seed Seed for the name of this component
     * @return this object
     */
-  private [chisel3] def autoSeed(seed: String): this.type = {
+  private[chisel3] def autoSeed(seed: String): this.type = forceAutoSeed(seed)
+  // Bypass the overridden behavior of autoSeed in [[Data]], apply autoSeed even to ports
+  private[chisel3] def forceAutoSeed(seed: String): this.type = {
     auto_seed = Some(seed)
     for(hook <- auto_postseed_hooks) { hook(seed) }
     prefix_seed = Builder.getPrefix()
--- a/plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
+++ b/plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
@@ -39,7 +39,7 @@ class ChiselComponent(val global: Global) extends PluginComponent with TypingTra
   class MyTypingTransformer(unit: CompilationUnit)
     extends TypingTransformer(unit) {
 
-    private def shouldMatchGen(bases: Tree*): ValDef => Boolean = {
+    private def shouldMatchGen(bases: Tree*): Type => Boolean = {
       val cache = mutable.HashMap.empty[Type, Boolean]
       val baseTypes = bases.map(inferType)
 
@@ -72,8 +72,7 @@ class ChiselComponent(val global: Global) extends PluginComponent with TypingTra
       }
 
       // Return function so that it captures the cache
-      { p: ValDef =>
-        val q = inferType(p.tpt)
+      { q: Type =>
         cache.getOrElseUpdate(q, {
           // First check if a match, then check early exit, then recurse
           if(terminate(q)){
@@ -87,9 +86,10 @@ class ChiselComponent(val global: Global) extends PluginComponent with TypingTra
       }
     }
 
-    private val shouldMatchData      : ValDef => Boolean = shouldMatchGen(tq"chisel3.Data")
-    private val shouldMatchDataOrMem : ValDef => Boolean = shouldMatchGen(tq"chisel3.Data", tq"chisel3.MemBase[_]")
-    private val shouldMatchModule    : ValDef => Boolean = shouldMatchGen(tq"chisel3.experimental.BaseModule")
+
+    private val shouldMatchData      : Type => Boolean = shouldMatchGen(tq"chisel3.Data")
+    private val shouldMatchDataOrMem : Type => Boolean = shouldMatchGen(tq"chisel3.Data", tq"chisel3.MemBase[_]")
+    private val shouldMatchModule    : Type => Boolean = shouldMatchGen(tq"chisel3.experimental.BaseModule")
 
     // Given a type tree, infer the type and return it
     private def inferType(t: Tree): Type = localTyper.typed(t, nsc.Mode.TYPEmode).tpe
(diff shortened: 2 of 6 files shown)


mwachs5:
👍 for the algorithm/behavior as documented in the comments

jackkoenig:
Marked `DO NOT MERGE` because I need to write docs and I'm seeing a handful of cases of worse names that I wouldn't expect so I need to dig into it.

jackkoenig:
> I'm seeing a handful of cases of worse names that I wouldn't expect

In a randomish sampling, I found this behavior to be correct. In some cases a temporary name would get overridden to a better, non-temporary name; however, it could also result in overriding names in a bad way. For example:

```scala
  def extractBelow(from: UInt, idx: Int): UInt = {
    val x = from
    x(idx, 0)
  }
  
  val in  = IO(Input(UInt(16.W)))
  val out = IO(Output(UInt(16.W)))
  // If foo is a field of Example, reflective naming override hides the bug
  if (true) {
    val foo = Reg(UInt(16.W))
    foo := in
    out := extractBelow(foo, 15)
  }
}
```
With v3.4.0 gives:
```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<16>
    output out : UInt<16>
    reg out_x : UInt<16>, clock with :
      reset => (UInt<1>("h0"), out_x)
    out_x <= in
    node _out_T = bits(out_x, 15, 0)
    out <= _out_T
```
(Scastie: https://scastie.scala-lang.org/T6mdGw3LQOSNc2yEyUMy5w)

See how the `val x` in the utility function (with prefix `out` from the `:=`) overrides the name? In this case, it's clearly not desirable. That being said, in other cases, this change turns some good names into temporaries. For example, a slight modification on the above:

```scala
  def extractBelow(from: UInt, idx: Int): UInt = {
    val x = from
    x(idx, 0)
  }
  
  val in  = IO(Input(UInt(16.W)))
  val out = IO(Output(UInt(16.W)))
  // If foo is a field of Example, reflective naming override hides the bug
  if (true) {
    val foo = Reg(UInt(16.W))
    foo := in
    out := extractBelow(foo +& 1.U, 15) // <---- difference here
  }
}
```
With v3.4.0 gives:
```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<16>
    output out : UInt<16>
    reg foo : UInt<16>, clock with :
      reset => (UInt<1>("h0"), foo)
    foo <= in
    node out_x = add(foo, UInt<1>("h1"))
    node _out_T = bits(out_x, 15, 0)
    out <= _out_T
```
(Scastie: https://scastie.scala-lang.org/3FoJCFOmRUK5iYteYkkvjg)

In this case, it's nice that the sum gets the name `out_x`. With the change in this PR, instead you get:
```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<16>
    output out : UInt<16>
    reg foo : UInt<16>, clock with :
      reset => (UInt<1>("h0"), foo)
    foo <= in
    node _out_T = add(foo, UInt<1>("h1"))
    node _out_T_1 = bits(_out_T, 15, 0)
    out <= _out_T_1                                                                                                                                                                                                
```
The loss of `out_x` is a bit unfortunate, but one could argue that this is _more predicable_ since the Expression at the callsite does not have a name. If we prioritize predictability over "best possible names" then I think this is actually good.

jackkoenig:
@azidar I've done an initial updating of the docs, but I have not yet added a description of "Outer Expression, First Statement". I would appreciate review of the code and tests while I finish up the docs.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -119,7 +120,9 @@ private[chisel3] trait HasId extends InstanceId {
     * @param seed Seed for the name of this component
     * @return this object
     */
-  private [chisel3] def autoSeed(seed: String): this.type = {
+  private[chisel3] def autoSeed(seed: String): this.type = forceAutoSeed(seed)
+  // Bypass the overridden behavior of autoSeed in [[Data]], apply autoSeed even to ports
+  private[chisel3] def forceAutoSeed(seed: String): this.type = {
```
azidar, line 125: Is the plan to remove autoseed in the future?

plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -131,30 +175,47 @@ class ChiselComponent(val global: Global) extends PluginComponent with TypingTra
...
           treeCopy.ValDef(dd, mods, name, tpt, localTyper typed named)
         } else {
           // Otherwise, continue
           super.transform(tree)
         }
+      case dd @ ValDef(mods, name, tpt, rhs @ Match(_, _)) if okUnapply(dd) =>
+        val tpe = inferType(tpt)
+        val fieldsOfInterest: List[Boolean] = tpe.typeArgs.map(shouldMatchData)
+        // Only transform if at least one field is of interest
+        if (fieldsOfInterest.reduce(_ || _)) {
```
azidar, line 207: Is this never empty?

azidar (approved):

plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
jackkoenig: If none of the fields are `<: Data` then it will be empty, eg.
```scala
val (a, b) = ("hi", 3)
```

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: Possibly, a good question for the dev meeting

PR #1661, opened 2020-11-11
Ignore tests using System.setSecurityManager

jackkoenig:
The SecurityManager is global so is not thread-safe. This is the source
of flaky tests in FIRRTL CI.

### Contributor Checklist

None

#### Type of Improvement

 - bug fix (flaky tests)

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/ChiselSpec.scala
+++ b/src/test/scala/chiselTests/ChiselSpec.scala
@@ -281,6 +281,7 @@ trait Utils {
     * or doesn't try to write at all.
     */
   def catchWrites[T](thunk: => T): Either[String, T] = {
+    throw new Exception("Do not use, not thread-safe")
     try {
       System.setSecurityManager(new ExceptOnWrite())
       Right(thunk)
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -266,6 +266,7 @@ trait Utils {
     * @return either the output of the thunk (`Right[T]`) or an exit code (`Left[Int]`)
     */
   def catchStatus[T](thunk: => T): Either[Int, T] = {
+    throw new Exception("Do not use, not thread-safe")
```
seldridge: I would prefer to add documentation about this and monitor for usages of it in code review rather than to always throw an exception. 

Tangentially, I expect that `catchStatus` is actually safe to use even if it isn't thread safe. (I was wondering why this hasn't yet caused a problem [in spite of it being used for a long time in FIRRTL](https://github.com/freechipsproject/firrtl/blob/bbcee06d406c3755c10d41a71d69a69eb7d6f321/src/test/scala/firrtlTests/stage/FirrtlMainSpec.scala#L83).) FIRRTL seems to be serializing all tests that use it and you don't have anything else in FIRRTL calling `System.exit`.

seldridge (approved):

src/test/scala/chiselTests/ChiselSpec.scala
jackkoenig: Switched this PR to continue to use `catchStatus`

PR #1694, opened 2020-12-08
Add when.cond for getting the current when condition

jackkoenig:
This is useful for libraries to guard operations implemented via
annotations or BlackBoxes by the current when predicate

@albert-magyar, remember how we discussed this like 3 years ago?

One question I have is how far we should backport this. It's a feature addition so we could just do `3.4.x`, but it is a critical feature for verification libraries so it might be nice to go back to `3.2.x`.

The main complexity in the implementation is that `elsewhen` and `otherwise` are actually parallel with previous conditions that must not be true (rather than being nested as they are in FIRRTL). I originally had `altConds` as a single `Option[() => Bool]`, but the test caught the issue with `elsewhen`. Basically, `altConds` is a `List` to keep track of the condition of the starting `when` and prior `elsewhens` in the same chain so they can be inverted. Nesting is handled inside of `when.cond` when it accesses the `whenStack`.

I also used `foldRight` throughout the implementation so that conditions will flow in the order of the `whens` rather than being reversed.

### Contributor Checklist

#### Type of Improvement

  - new feature/API  

#### API Impact

This adds `when.cond` for determining the current when condition

No impact

#### Desired Merge Strategy

   - Squash

#### Release Notes

Add `when.cond` for determining the current when condition

- [ ] Did you review?


ducky64 (approved):
Looks fine. A bit more documentation on alt might be helpful, especially motivation (I assume it's so you can lazily evaluate the inverse, because you expect it will rarely need to happen?)

sequencer (commented):
I test this today, works great!
This is really useful for Verification and plugin dev.

jackkoenig:
Because this API is very much a "power user library writer" thing (I actually have struggled to provide useful-looking examples using it other than directly driving a blackbox or annotating the signal), @albert-magyar suggested we put it in a different package like `chisel3.advanced` or something. Any thoughts/opinions on that @ducky64 @azidar @sequencer?

ducky64:
Something similar did cross my mind: it's a bit of a reflection style API, which we've typically put (maybe a bit haphazardly) in DataMirror. I don't know if that's the right answer, though.

aswaterman:
I think putting features that are too important to exclude, but are dangerous or not useful for typical users, into something like chisel3.advanced is a sensible proposal.

core/src/main/scala/chisel3/When.scala
```diff
@@ -29,7 +28,30 @@ object when {
...
+    * }}}
+    *  */
+  def cond: Bool = {
+    implicit val compileOptions = ExplicitCompileOptions.Strict
+    implicit val sourceInfo = UnlocatableSourceInfo
+    val whens = Builder.whenStack
+    whens.foldRight(true.B) {
+      case (ctx, acc) => acc && ctx.localCond()
+    }
   }
```
jackkoenig, line 55: Can we get a source locator from the when conditions themselves?

jackkoenig:
We talked about this in dev yesterday, I'm swayed the most by @chick's argument that moving things from `experimental` has been painful so maybe things like this should be handled more with documentation than anything. Obviously `experimental` isn't a perfect comparison since `chisel3.advanced` is somewhere we'd probably not plan to move things out of, but if we did want to then it would be the same problem.

@seldridge asked if there's any prior art for "advanced" packages, and I don't think I know of any. Do other projects have packages for "advanced features" or do they just document things?

azidar (approved):

PR #1714, opened 2021-01-02
Make `toTarget` fail if called on a Literal (or would otherwise not serialize properly)

### LINKED ISSUE #1713: Literals should not be allowed to pass as Reference Targets
**Type of issue**: bug report 

**Impact**: API modification 

**Other information**
 

Chisel allows a literal value to be Serialized as a `ReferenceTarget` in an annotation. However, FIRRTL cannot deserialize that and it's unclear what the semantics of that are. There is no other indication that literals are supposed to be valid reference targets.

Scastie demonstrating the issue:
https://scastie.scala-lang.org/poLW9X5VRrm6AhlczNOQzg

code duplicated here:
```
import chisel3._
import chisel3.experimental.{
  annotate,
  ChiselAnnotation
}
import chisel3.stage.{
  ChiselGeneratorAnnotation,
  ChiselStage
}

import firrtl.annotations.{
  JsonProtocol,
  ReferenceTarget,
  SingleTargetAnnotation
}

case class RefTargetAnnotation(
  name: String,
  target: ReferenceTarget) extends SingleTargetAnnotation[ReferenceTarget] {
  
  override def duplicate(target: ReferenceTarget) = this
  
}


class Foo extends RawModule { self: Foo =>
  
  val a = 1.U
  val b = WireInit(a)
  
  annotate(new ChiselAnnotation {
      override def toFirrtl = RefTargetAnnotation("wireRef", b.toTarget)
    }
  )
  
  annotate(new ChiselAnnotation {
      override def toFirrtl = RefTargetAnnotation("literalRef", a.toTarget)
    }
  )
  
}

val jsonResult = JsonProtocol.serialize((new ChiselStage)
  .execute(Array("--no-run-firrtl"), Seq(ChiselGeneratorAnnotation(() => new Foo)))
  .filter {
    case r: RefTargetAnnotation => true
    case _                         => false
  })

println("JSON Result: ")
println(jsonResult)

// This seems to silently not complete
val deserializedAnnos = JsonProtocol.deserialize(jsonResult)
println("Deserialized Annos:")
println(deserializedAnnos)
```

Produces (before failing in the deserialize):

```
JSON Result: 
[
  {
    "class":"Playground$RefTargetAnnotation",
    "name":"wireRef",
    "target":"~Foo|Foo>b"
  },
  {
    "class":"Playground$RefTargetAnnotation",
    "name":"literalRef",
    "target":"~Foo|Foo>UInt<1>(\"h01\")"
  }
]
```

See above scastie/code example

Chisel happily allows creating of Annotations which take `ReferenceTarget`s as arguments and serialize a literal passed in. However, deserializing the same will result in a hard-to-understand failure message of "unknown error", because the JSON parser stack trace is being eaten.

1) Chisel should not let you pass off a literal as a ReferenceTarget if FIRRTL is not going to accept that

It's unclear to me whether this would "just work" if you didn't have to go through the serialization/deserialization step, then its just up to the annotation to deal with the fact that it got a literal instead of a reference target.

    - version: `3.4.1`
    - OS: `Scastie`
    - 

People write annotations that take ReferenceTargets then users pass literals into them, and things seem to generally work until firrtl can't deserialize them with a hard-to-understand error message.

Also, the `toNamed` did not allow literals. `toTarget` should behave similiarly:

https://scastie.scala-lang.org/5lnPB8z9QPigMCKzNdy0Iw

mwachs5:
### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

It breaks an existing API, in that you will now get an exception if you try to call `.toTarget` on a literal data whereas before it would proceed. But this is a bug.

I don't think this would change any generated verilog since this use case is not supported.

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference)

#### Release Notes

.toTarget now checks to make sure that the ReferenceTarget would serialize/deserialize properly, and throws an exception otherwise. This is similar to how toNamed checks for proper instance name. Fixes #1713 

- [x] Did you review?


src/test/scala/chiselTests/ReferenceTargetSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+  * @param value a simple documentation string
+  */
+case class ReferenceTargetAnnotation(target: ReferenceTarget, value: String) extends SingleTargetAnnotation[ReferenceTarget] {
+  def duplicate(n: ReferenceTarget): ReferenceTargetAnnotation = this.copy(target = n)
+}
+
+/** ChiselAnnotation that corresponds to the [[ReferenceTargetAnnotation]] FIRRTL annotation
+  *
+  *  @param target the Data which will be converted to a FIRRTL ReferenceTarget
+  *  @praram value the documentation string which will be copied directly to the FIRRTL annotation
```
mwachs5, line 34:
```suggestion
  *  @param value the documentation string which will be copied directly to the FIRRTL annotation
```

seldridge (commented):
The fix looks great! Thanks for taking care of this!

Would you be able to pare down the test to it's essentials? This is pretty verbose as is. I think it would be sufficient to simply check that `toTarget` throws the expected exception (which you need to massage a bit because `toTarget` needs to be called after the module is elaborated). I think you could use something like the following and check that it throws the correct exception and message:

https://scastie.scala-lang.org/eKyFTKxhTRu41Y0r15Lm1A

seldridge (approved):

PR #1730, opened 2021-01-14
Remove deprecations 

sequencer:
Depends on https://github.com/freechipsproject/chisel-testers/pull/309
### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous printlns/debugging code?
- [ ] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [ ] Did you state the API impact?
- [ ] Did you specify the code generation impact?
- [ ] Did you request a desired merge strategy?
- [ ] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- code cleanup

#### API Impact
Remove all deprecated API in before and include 3.2

#### Desired Merge Strategy

- Rebase: You will rebase the PR onto master and it will be merged with a merge commit.

#### Release Notes
Remove deprecations since and include 3.2

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (3.2.x, 3.3.x, 3.4.0, 3.5.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1731, opened 2021-01-14
Expose AnnotationSeq to Module.

### LINKED ISSUE #1729: Expose AnnotationSeq to Builder

Current situation is that after changing a small part of codes, whole RTL will be re-elaborated.
But for time consuming part like a very large decoder, which won't need the QMC re-computation if don't change the decoder table.
So I propose a simple idea that, we can expose `AnnotationSeq` to `Builder`, served as a standard interface to user, to cache decoder table during different elaborations.

sequencer:
Resolve #1729
depends on #1730
### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact

add API `chisel3.util.experimental.getAnnotations` to get `AnnotationSeq` in current phase.

None
#### Desired Merge Strategy

#### Release Notes
Expose AnnotationSeq to Module.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -14,6 +14,7 @@ import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.{InstTransform, SourceInfo, UnlocatableSourceInfo}
 import chisel3.experimental.BaseModule
 import _root_.firrtl.annotations.{IsModule, ModuleName, ModuleTarget}
+import _root_.firrtl.AnnotationSeq
 
 object Module extends SourceInfoDoc {
   /** A wrapper method that all Module instantiations must be wrapped in
@@ -84,6 +85,8 @@ object Module extends SourceInfoDoc {
   def reset: Reset = Builder.forcedReset
   /** Returns the current Module */
   def currentModule: Option[BaseModule] = Builder.currentModule
+  /** Returns the global Annotations */
+  def annotationSeq: AnnotationSeq = Builder.annotationSeq
 }
 
 /** Abstract base class for Modules, which behave much like Verilog modules.
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -85,8 +85,6 @@ object Module extends SourceInfoDoc {
   def reset: Reset = Builder.forcedReset
   /** Returns the current Module */
   def currentModule: Option[BaseModule] = Builder.currentModule
-  /** Returns the global Annotations */
-  def annotationSeq: AnnotationSeq = Builder.annotationSeq
 }
 
 /** Abstract base class for Modules, which behave much like Verilog modules.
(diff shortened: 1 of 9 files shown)


core/src/main/scala/chisel3/Module.scala
```diff
@@ -85,6 +86,8 @@ object Module extends SourceInfoDoc {
   def reset: Reset = Builder.forcedReset
   /** Returns the current Module */
   def currentModule: Option[BaseModule] = Builder.currentModule
+  /** Returns the global Annotations */
+  def annotationSeq: AnnotationSeq = Builder.annotationSeq
```
jackkoenig, line 90:
```suggestion
  /** Returns the global Annotations */
  def annotations: AnnotationSeq = Builder.annotationSeq
```

I'm also not sure if `Module` is the right place for this, but I'm not really sure where else would be right either...

sequencer: When I was working on this, I also get puzzled where to put this API.
Maybe directly under `chisel3.experimental`?
But I'm not sure is experimental thread safe for this API?

jackkoenig: Since these annotations are set at the beginning of elaboration (ie. they aren't modified), I don't think thread-safety is really a concern fortunately

src/main/scala/chisel3/util/experimental/getAnnotations.scala
```diff
@@ -0,0 +1,9 @@
+package chisel3.util.experimental
+
+import chisel3.internal.Builder
+import firrtl.AnnotationSeq
+
+object getAnnotations {
```
sequencer, line 6: @jackkoenig I move this API to `chisel3.util.experimental`. Maybe a good place here?

jackkoenig:
@sequencer can you rebase this on master?

sequencer:
> @sequencer can you rebase this on master?

Done :)

sequencer:
This PR depends on #1730, since I'm too lazy to fix the Driver compatibility Mode;p
So maybe we can get freechipsproject/chisel-testers#308 and freechipsproject/chisel-testers#309 in firstly, then #1730, and finally #1729 this one.
I think those 3 are good to merge now.

azidar:
@sequencer  can you rebase again?

sequencer:
Sorry I was on vacation recently. I’ll do it once I return home tomorrow.

sequencer:
@azidar ready to review now :)

azidar (approved):

core/src/main/scala/chisel3/aop/Aspect.scala
```diff
@@ -12,6 +12,7 @@ import firrtl.AnnotationSeq
   * @tparam T Type of top-level module
   */
 abstract class Aspect[T <: RawModule] extends Annotation with Unserializable with NoTargetAnnotation {
+  var annotations: AnnotationSeq = Seq()
```
azidar, line 15: Do you think we should make this private, and provide a different accessor?

jackkoenig: Yes

jackkoenig:
Marked as `3.5` due to dependency on #1730

jackkoenig (approved):

PR #1740, opened 2021-01-19
Parametrized Mem- & SyncReadMem-based implementation of the Queue class

milovanovic:
Previously `Queue` class was implemented exclusively using `Mem` which practically prevents the SRAM-based realization in ASICs. Here, a new parameter `useSyncReadMem` with default value `false` is added to specify the internal `Queue` implementation and to allow the use of `SyncReadMem` and hence enable Queues to be mapped to SRAM cells. Backward compatibility is preserved.

### Contributor Checklist

#### Type of Improvement

New feature in backend code generation.

#### API Impact

Full backward compatibility is preserved with default parameter value.

If `SyncReadMem` is used it can be specified in FIRRTL to implement it using SRAM macros instead of D flip-flops/registers.

#### Desired Merge Strategy

None.

#### Release Notes

Added parametrized `SyncReadMem`-based implementation of the `Queue` class.

- [ ] Did you review?


src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -197,7 +197,8 @@ class QueueIO[T <: Data](private val gen: T, val entries: Int) extends Bundle
 class Queue[T <: Data](val gen: T,
                        val entries: Int,
                        val pipe: Boolean = false,
-                       val flow: Boolean = false)
+                       val flow: Boolean = false,
+                       val useSyncReadMem: Boolean = false)
```
edwardcwang, line 201: Should update the scaladoc above

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -215,9 +216,12 @@ class Queue[T <: Data](val gen: T,
 
   val io = IO(new QueueIO(genType, entries))
 
-  val ram = Mem(entries, genType)
+  val ram = if (useSyncReadMem) SyncReadMem(entries, genType) else Mem(entries, genType)
+  
   val enq_ptr = Counter(entries)
   val deq_ptr = Counter(entries)
+  val deq_ptr_wire = WireDefault(deq_ptr.value)
```
mwachs5, line 223: should this be called `deq_ptr_value` since that's what it is?

mwachs5: also should this entire wire be an Option since it's only used if you have SyncReadMem?

src/test/scala/chiselTests/QueueSpec.scala
```diff
@@ -202,70 +202,70 @@ class QueueSpec extends ChiselPropSpec {
   implicit val noShrinkInt = Shrink[Int](_ => Stream.empty)
 
   property("Queue should have things pass through") {
-    forAll(vecSizes, safeUIntN(20), Gen.choose(0, 15)) { (depth, se, tap) =>
+    forAll(vecSizes, safeUIntN(20), Gen.choose(0, 15), Gen.choose(0, 1)) { (depth, se, tap, isSync) =>
```
mwachs5, line 205: couldn't this just chose `true`, `false` and then that is passed directly instead of `isSync == 1`?

src/test/scala/chiselTests/QueueSpec.scala
```diff
@@ -9,8 +9,8 @@ import chisel3.testers.BasicTester
 import chisel3.util._
```
mwachs5, line 9: Overall I don't see what the new tests added are actually testing, other than the code still works? Shouldn't you also be checking for the existence of a SeqMem vs a Mem when the knob is set differently?

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -227,20 +231,33 @@ class Queue[T <: Data](val gen: T,
...
   }
 
   io.deq.valid := !empty
   io.enq.ready := !full
-  io.deq.bits := ram(deq_ptr.value)
-
+  
+  if (useSyncReadMem) {
+    io.deq.bits := ram.read(deq_ptr_wire)
+  }
```
sequencer, line 256: If `SyncReadMem` is introduced here, what's the behavior when `deq_ptr_wire === enq_ptr.value && do_enq`?
If this won't happen, you can add an assert here to placate users.
If this is going to happen, I think `ReadUnderWrite` behavior should also be specified.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Sorry, completely forgot about that one. It's now updated.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Sure, you're right. The name is changed as per your suggestion and the entire wire is an Option now.

src/test/scala/chiselTests/QueueSpec.scala
milovanovic: Actually it doesn't work directly with `Gen.choose(true, false)` but it does with `Gen.oneOf(true, false)`. Thanks for the suggestion.

src/test/scala/chiselTests/QueueSpec.scala
```diff
@@ -202,70 +202,70 @@ class QueueSpec extends ChiselPropSpec {
...
       whenever(se._1 >= 1 && depth >= 1 && se._2.nonEmpty) {
         assertTesterPasses {
-          new CountIsCorrectTester(se._2, depth, se._1, tap)
+          new CountIsCorrectTester(se._2, depth, se._1, tap, isSync)
         }
       }
     }
   }
-
+
```
mwachs5, line 233: can you delete these whitespaces

src/test/scala/chiselTests/QueueSpec.scala
```diff
@@ -202,70 +202,70 @@ class QueueSpec extends ChiselPropSpec {
...
       whenever(se._1 >= 1 && depth >= 1 && se._2.nonEmpty) {
         assertTesterPasses {
-          new ThingsPassThroughTester(se._2, depth, se._1, tap)
+          new ThingsPassThroughTester(se._2, depth, se._1, tap, isSync)
         }
       }
     }
   }
-
+
```
mwachs5, line 213: human lint says delete these whitespaces

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -227,20 +232,34 @@ class Queue[T <: Data](val gen: T,
...
+    }
+    else {
+      ram(enq_ptr.value) := io.enq.bits
+    }
     enq_ptr.inc()
   }
   when (do_deq) {
     deq_ptr.inc()
+    if (useSyncReadMem)
+      deq_ptr_value.get := Mux(deq_ptr.value === (entries.U - 1.U), 0.U, deq_ptr.value + 1.U)
```
albert-magyar, line 246: As an unfortunate consequence of how the behavioral `read` method works (the read is enabled when the address is updated), this actually uses `do_deq` as the read enable.

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -227,20 +232,34 @@ class Queue[T <: Data](val gen: T,
...
     maybe_full := do_enq
   }
 
   io.deq.valid := !empty
   io.enq.ready := !full
-  io.deq.bits := ram(deq_ptr.value)
-
+  
+  if (useSyncReadMem) {
+    io.deq.bits := ram.read(deq_ptr_value.get)
```
albert-magyar, line 257:
```suggestion
    val deq_addr = WireDefault(deq_ptr_value.get)
    io.deq.bits := ram.read(deq_addr)
```

An unconditionally connected wire will make the read unconditional as an easy, non-optimal-but-correct alternative.

albert-magyar (changes requested):

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -227,20 +232,34 @@ class Queue[T <: Data](val gen: T,
   val do_deq = WireDefault(io.deq.fire())
 
   when (do_enq) {
-    ram(enq_ptr.value) := io.enq.bits
+    if (useSyncReadMem) {
```
albert-magyar, line 235: Is this elaboration conditional actually necessary? Both cases should do the same thing.

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -215,9 +217,12 @@ class Queue[T <: Data](val gen: T,
 
   val io = IO(new QueueIO(genType, entries))
 
-  val ram = Mem(entries, genType)
+  val ram = if (useSyncReadMem) SyncReadMem(entries, genType) else Mem(entries, genType)
```
albert-magyar, line 220:
```suggestion
  val ram = if (useSyncReadMem) SyncReadMem(entries, genType, SyncReadMem.WriteFirst) else Mem(entries, genType)
```

Agreeing with @sequencer here, it's probably good to specify `WriteFirst` since it's necessary for correctness. While the default undefined R/W-collision behavior maps to write-first in the default flow, this isn't explicitly specified.

src/test/scala/chiselTests/QueueSpec.scala
milovanovic: You are right. The tests simply verify that both Mem and SyncReadMem implementations work. I agree with you that it would be good to check if that particular memory type exists, but I'm really not sure how it should be properly done. If you have any examples or hints, please share, and it'll be included.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Sure, since it might happen, as per @albert-magyar's suggestion, the `WriteFirst` strategy has been adopted.

src/test/scala/chiselTests/QueueSpec.scala
milovanovic: Absolutely, sorry about introducing them in the first place.

src/test/scala/chiselTests/QueueSpec.scala
milovanovic: Some additional whitespaces in `Decoupled.scala` have been detected and destroyed too.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: I guess you wanted to say that in this case the `read` method with enable signal can be potentially used and so the code could be organized differently. However, it seems that the `val ram = if ...` assignment actually causes that specific `read` method with enable signal is not visible and hence can not be used due compile-time error.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Thanks for pointing this out and suggesting the fix. It's included in the current commit.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: No, obviously not necessary, but vast majority of the code that deals with `SyncReadMem` uses `read` and `write` methods for appropriate read and write processes. Right now this `if`/`else` construct is eliminated, as it was in the original code, but it can be easily reverted back if it turns out to be good to have it.

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Once again thanks for suggesting the change. It's included in the latest code.

src/main/scala/chisel3/util/Decoupled.scala
albert-magyar: Thanks, I see that it correctly uses an unconditional read. Sorry for the confusion, it seems like I've been looking at multiple versions incorrectly, and I got a little confused with the new use of option + WireDefault at the top level.

I wrote a new non-line-comment to be a little easier to follow.

albert-magyar:
Sorry that I've unnecessarily commented on a few stale versions, but I personally found the `read_ptr_value` `Option` use a little confusing.

Could it be simpler to just limit the visibility of the "next read pointer" to the scope where it's used, namely the memory access? This would let the top-level `read_ptr_value` be deleted and make it simple to see that it's an unconditional read.
```
  if (useSyncReadMem) {
    val deq_ptr_next = Mux(deq_ptr.value === (entries.U - 1.U), 0.U, deq_ptr.value + 1.U)
    val r_addr = WireDefault(Mux(do_deq, deq_ptr_next, deq_ptr.value))
    io.deq.bits := ram.read(r_addr)
  } else {
    io.deq.bits := ram(deq_ptr.value)
  }
```

src/main/scala/chisel3/util/Decoupled.scala
milovanovic: Another great advice to limit the scope! Your idea is incorporated in the latest commit. Thank you.

albert-magyar (approved):
LGTM. While it would be nice to check to see what actual mem implementation is used in the test, I think that there is probably no reasonably convenient way to do this, and it's pretty clear that the parameter picks the right memory.

albert-magyar:
@azidar is it okay if I change this to 3.5.0 due to the API changes?

PR #1745, opened 2021-01-21
Remove "val io" and rename MultiIOModule to Module

jackkoenig:
This allows us to remove `-Xsource:2.11` from `scalacOptions` in Chisel and dependent projects. Removing `-Xsource:2.11` is required for supporting Scala 2.13.

### Contributor Checklist

~This will be more appropriate as part of larger Chisel 3.5 documentation~ I take it back, updating docs...

#### Type of Improvement

  - code refactoring  
  - code cleanup

#### API Impact

* `-Xsource:2.11` is no longer required
* `val io` is removed so code that hasn't fixed the deprecation warning will not compile
* `MultiIOModule` has been renamed to `Module`, `MultiIOModule` is now a (deprecated) type alias
* `chisel3.internal.LegacyModule` is no longer the same type as `Module`. This could be problematic for anyone using it (although they shouldn't be, it's in package `chisel.internal`). I tried to think about a better way to do this, but got very messy very fast. `LegacyModule` is more-or-less now just a bridge to enable `Chisel.Module` with autowrapping.
* I also deprecated `Module.override_clock` and `Module.override_reset` since they said they were going to be removed
    * This is a little tricky because `Chisel.QueueCompatibility` extends `chisel3.Queue` which extends `chisel3.Module`
* Autowrapping of `io` in `Chisel.Module` still works, using reflection.
* `BlackBox` still needs to handle `io` specially despite the removal of the virtual method; it strips the `io` prefix. It now just looks up a port named "io" which technically means that a user could just `.suggestName("io")` some other `val`. I don't love this behavior, but I'd also rather not require reflection for `BlackBox` to work. 

This does have one tiny impact on FIRRTL generation (but not Verilog). For instantiations of `Chisel.Module`s, the `clock` and `reset` will be invalidated and then promptly connected to. It has no effect except maybe increased code size for emitted `.fir` and `.pb`.

#### Desired Merge Strategy

 - Rebase (please do **not** `Please Merge` this!)

#### Release Notes

* scalac option `-Xsource:2.11` is no longer required in Chisel projects
* `Module.io` and `BlackBox.io` virtual methods have been removed (See https://github.com/freechipsproject/chisel3/pull/1550 for more information)
* `MultiIOModule` has been renamed to `Module`, `MultiIOModule` is now a (deprecated) type alias
* `chisel3.internal.LegacyModule` is no longer the same type as `Module`. Instead, `Chisel.Module` is a now just a type alias for `chisel3.internal.LegacyModule`. **Users should not directly use anything in package `chisel3.internal`**
* `Module.override_clock` and `Module.override_reset` are deprecated. These were an awkward holdover from Chisel 2.

- [ ] Did you review?


mwachs5:
would it be reasonable to request a corresponding PR to https://www.chisel-lang.org/chisel3/docs/wiki-deprecated/modules.html? I guess in general what is the workflow for keeping the website up-to-date, or is this something to do as part of the 3.5 release process?

jackkoenig:
> would it be reasonable to request a corresponding PR to https://www.chisel-lang.org/chisel3/docs/wiki-deprecated/modules.html? I guess in general what is the workflow for keeping the website up-to-date, or is this something to do as part of the 3.5 release process?

I should update the docs as part of this PR, thanks for pointing it out.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
```diff
@@ -31,15 +31,19 @@ class AnalogBlackBox(index: Int) extends BlackBox(Map("index" -> index)) {
   val io = IO(new AnalogBlackBoxIO(1))
 }
 
+trait AnalogBlackBoxModuleIntf extends Module {
```
ducky64, line 34: What's this for?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -234,17 +208,15 @@ package internal {
...
     }
 
-    private[chisel3] override def initializeInParent(parentCompileOptions: CompileOptions): Unit = {
-      // Don't generate source info referencing parents inside a module, since this interferes with
-      // module de-duplication in FIRRTL emission.
-      implicit val sourceInfo = UnlocatableSourceInfo
-
-      if (!parentCompileOptions.explicitInvalidate) {
-        pushCommand(DefInvalid(sourceInfo, _io.ref))
+    override def _compatAutoWrapPorts(): Unit = {
```
ducky64, line 211: Does this change behavior, that LegacyModule does auto-wrapping instead of just Chisel.Module? Is this intended? And if so, what should the difference between the two be, or can they be mostly merged?

ducky64 (approved):
Looks mostly fine - but two questions in comments.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
jackkoenig: It's an example that was hitting the `val io` deprecation. [`val mods` in `AnalogDUT`](https://github.com/chipsalliance/chisel3/blob/616256c35cb7de8fcd97df56af1986b747abe54d/src/test/scala/chiselTests/AnalogIntegrationSpec.scala#L66) relies on `io` being a field of `Module` (interestingly, it's actually inferring a structural type, `Module { val io: AnalogBlackBoxIO }`. The deprecation warning obviously became an error so I had to fix it and this is the canonical way.

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: > Does this change behavior, that LegacyModule does auto-wrapping instead of just Chisel.Module? Is this intended?

Yes that's right, and it is intentional. I thought it better that the reflection logic live inside of `package chisel3` so that `lazy val _io` could be private rather than leak (since `package object Chisel` is not in package `chisel3` so we can't use package private). I'll agree it's a bit awkward if anyone is using `LegacyModule`.

The alternative I thought about was adding trait in `package chisel3.internal` that extends `LegacyModule` with this new functionality. I ended up folding it into `LegacyModule` because it felt like just extra levels of inheritance that didn't actually do anything useful, and as `LegacyModule` is already in package `chisel.internal`, you're not supposed to be using it. That being said people *could be* so we could consider deprecating inheritance from `LegacyModule`, adding the extra layer of hierarchy, and folding them together in the next version. I just decided to keep it simpler.

> And if so, what should the difference between the two be, or can they be mostly merged?

The difference is minimal, really just the addition of the constructors allowing for overriding `clock` and `reset` in `Chisel.Module`. They could be merged, although doing so inside of package `chisel.internal` and using a type alias would harm the ScalaDoc of `Chisel.Module`. I don't feel too strongly either way though.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
ducky64: Ah, I see. Might be nice to add a quick comment with the trait about why it exists, though.

core/src/main/scala/chisel3/RawModule.scala
ducky64: I think it might be cleaner to move all the code into LegacyModule (so Chisel.Module is just a type alias) so we don't imply that there should be some kind of difference. If the documentation doesn't resolve cleanly, we can always make Chisel.Module an empty extension of LegacyModule (and document the rationale in the docs)?

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
ekiwi: :+1: on adding the scaladoc

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: It turns out that `Chisel.Module` is already a type alias for `Chisel.CompatibilityModule` so there's no harm in folding it into `chisel3.internal.LegacyModule` 🤷, I'll go ahead and do that and update the ScalaDoc

jackkoenig: On the flip side, we exclude everything in package `chisel3.internal` from the ScalaDoc, so with `Chisel.Module` as a type alias for `chisel3.internal.LegacyModule`, there is no ScalaDoc :-/

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
jackkoenig: This is in `test` so not sure "ScalaDoc" is useful, but I will add a comment 🙂

core/src/main/scala/chisel3/core/package.scala
```diff
@@ -66,13 +66,13 @@ package object core {
   type RawModule = chisel3.RawModule
   @deprecated("Avoid importing from chisel3.core, these are not public APIs and may change at any time. " +
     "Use chisel3.MultiIOModule instead. This alias will be removed in 3.4.", "since the beginning of time")
```
edwardcwang, line 68: Do these docs need to be updated?

jackkoenig: Yes, but they are just deleted by https://github.com/chipsalliance/chisel3/pull/1730 so I'll leave it alone to not conflict.

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: Done, I've changed `Chisel.Module` to be a type alias for `chisel3.internal.LegacyModule`.

src/test/scala/chiselTests/AnalogIntegrationSpec.scala
jackkoenig: Done.

jackkoenig:
I originally had forgotten to delete `def io` from `BlackBox` so I went back and did that. It led to one consequence, pasted from `API Impact`:

`BlackBox` (which has special handling, stripping the `io` prefix), still needs to handle `io` specially despite the removal of the virtual method. It now just looks up a port named "io" which technically means that a user could just `.suggestName("io")` some other `val`. I don't love this behavior, but I'd also rather not require reflection for `BlackBox` to work.

jackkoenig:
I have updated the docs. I'll wait a while, but barring objection I'm going to merge this tonight.

jackkoenig:
For those keeping track at home, this was by far the hardest part about getting Scala 2.13 working (`-Xsource:2.11` doesn't exist any more). I hope to have a 2.13 PR up tomorrow.

ekiwi:
> I hope to have a 2.13 PR up tomorrow.

I am excited!

core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -134,24 +134,20 @@ package experimental {
...
 
   // Allow access to bindings from the compatibility package
   protected def _compatIoPortBound() = portsContains(_io)
 
   private[chisel3] override def generateComponent(): Component = {
     _compatAutoWrapPorts()  // pre-IO(...) compatibility hack
 
-    // Restrict IO to just io, clock, and reset
-    require(_io != null, "BlackBox must have io")
+  // Restrict IO to just io, clock, and reset
```
mwachs5, line 149:
```suggestion
    // Restrict IO to just io, clock, and reset
```

why did this indentation change?

core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -134,24 +134,20 @@ package experimental {
...
+    require(_io != null, "BlackBox must have a port named 'io' of type Record!")
     require(portsContains(_io), "BlackBox must have io wrapped in IO(...)")
     require(portsSize == 1, "BlackBox must only have io as IO")
```
mwachs5, line 152:
```suggestion
    require(portsSize == 1, "BlackBox must only have one IO, called `io`")
```

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -142,86 +143,83 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
...
+      .orElse(tryScalaReflect)
+      .map(_.autoSeed("io"))
+      .orElse {
+        // Fallback if reflection fails, user can wrap in IO(...)
+        self.findPort("io")
+          .collect { case r: Record => r }
+      }.getOrElse(throwException(
+        s"Compatibility mode Module '$this' must have a 'val io' Bundle. " +
+        "If there is such a field and you still see this error, autowrapping has failed (sorry!). " +
+        "Please wrap the Bundle declaration in IO(...)."
```
mwachs5, line 193: Since user would hit this error message when using `Chisel._`, should this message say something like "Please wrap the `io` Bundle declaration in `chisel3.IO`?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -142,86 +143,83 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions)
...
-      "For more information, see: https://github.com/freechipsproject/chisel3/pull/1550.",
-      "Chisel 3.4"
-    )
-    def io: Record
-
-    // Private accessor to reduce number of deprecation warnings
-    private[chisel3] def _io: Record = io
+  abstract class LegacyModule(implicit moduleCompileOptions: CompileOptions) extends Module {
+    // Provide a non-deprecated constructor
+    def this(override_clock: Option[Clock]=None, override_reset: Option[Bool]=None)
```
mwachs5, line 205:
```suggestion
    def this(override_clock: Option[Clock] = None, override_reset: Option[Bool]=None)
```

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -234,17 +232,32 @@ package internal {
...
+    }
+  }
 
-      clock := override_clock.getOrElse(Builder.forcedClock)
-      reset := override_reset.getOrElse(Builder.forcedReset)
+  import chisel3.experimental.Param
+
+  /** Legacy BlackBox class will reflectively autowrap val io
+    *
+    * '''Do not use this class in user code'''. Use whichever `Module` is imported by your wildcard
```
mwachs5, line 246: should this message say Use whichever `BlackBox` is imported...?

docs/src/wiki-deprecated/modules.md
```diff
@@ -86,24 +86,18 @@ new object.  We then wire them up to one another and to the ports of
...
-
-A `MultiIOModule` allows you to define as many different `IO` as needed
-and does not require you to implement an abstract member `io`.
-This can be useful when programmatically adding `IO` or adding `IO` via inheritance.
-An artifact of this is that Verilog generated from a `MultiIOModule` will
-*not* have the `io_` prefix. `MultiIOModule`s still have an implicit
-clock and reset like `Module`.
-
-<!-- TODO: Some example -->
+> Historical Note: Prior to Chisel 3.5, Modules were restricted to only
```
mwachs5, line 92: 👍

docs/src/wiki-deprecated/ports.md
```diff
@@ -30,21 +30,21 @@ provide powerful wiring constructs described later.
 
 (Chisel 3.2+)
 
-Chisel 3.2+ introduces an API `DataMirror.modulePorts` which can be used to inspect the IOs of any Chisel module, including MultiIOModules, RawModules, and BlackBoxes.
+Chisel 3.2+ introduces an API `DataMirror.modulePorts` which can be used to inspect the IOs of any Chisel module, including Modules, RawModules, and BlackBoxes.
```
mwachs5, line 33: I am a little confused by this sentence. does "any Chisel module" mean "Any Chisel.Module, chisel3.Module, and ..."? If not can you clarify.

mwachs5 (commented):
nice!  few minor comments on the user-facing messages

core/src/main/scala/chisel3/BlackBox.scala
jackkoenig: mistake, whoops

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: `IO` is actually a `protected` method on `BaseModule` so it's available: https://scastie.scala-lang.org/e3dJdyqrR42K38zeCy5SKw

I think we should transition to `chisel3.IO` (there's actually only `chisel3.experimental.IO` which should move) but that's a separate discussion.

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: Yep, copy-paste mistake

docs/src/wiki-deprecated/ports.md
jackkoenig: Yes, it means anything that inherits from `BaseModule`, I'll clarify

core/src/main/scala/chisel3/RawModule.scala
jackkoenig, line 246:
```suggestion
    * '''Do not use this class in user code'''. Use whichever `BlackBox` is imported by your wildcard
```

docs/src/wiki-deprecated/ports.md
```diff
@@ -30,21 +30,21 @@ provide powerful wiring constructs described later.
...
```
jackkoenig, line 34:
```suggestion
Chisel 3.2 introduced `DataMirror.modulePorts` which can be used to inspect the IOs of any Chisel module (this includes modules in both `import chisel3._` and `import Chisel._`, as well as BlackBoxes from each package).
```

mwachs5: but not RawModules?

jackkoenig: It does include `RawModules`, it's everything 😑 . What do you think the best way to phrase it is? I was trying to avoid just listing every module class

PR #1747, opened 2021-01-21
Refactor EnumAnnotations and EnumFactory

Soya-Onishi:
This is a nit fix. no logic is changed.

Rename `enumTypeName` into `typeName` to retain name consistency
against other enum annotations and
there is also `typeName` at scaladoc's `@param`.

Add return type at public method

Rename `enum_records` into `enumRecords` to retain name consistency

### Contributor Checklist

- [ ] ~~Did you add Scaladoc to every public function/method?~~
- [ ] ~~Did you add at least one test demonstrating the PR?~~
- [ ] ~~Did you delete any extraneous printlns/debugging code?~~
- [ ] ~~Did you add appropriate documentation in `docs/src`?~~
- [ ] ~~Did you add text to be included in the Release Notes for this change?~~

#### Type of Improvement

- code refactor
- code cleanup

#### API Impact

none

none

#### Desired Merge Strategy

Squash

- [x] Did you review?


core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -20,12 +20,12 @@ object EnumAnnotations {
...
     */
-  case class EnumComponentAnnotation(target: Named, enumTypeName: String) extends SingleTargetAnnotation[Named] {
+  case class EnumComponentAnnotation(target: Named, typeName: String) extends SingleTargetAnnotation[Named] {
     def duplicate(n: Named): EnumComponentAnnotation = this.copy(target = n)
   }
 
-  case class EnumComponentChiselAnnotation(target: InstanceId, enumTypeName: String) extends ChiselAnnotation {
-    def toFirrtl: EnumComponentAnnotation = EnumComponentAnnotation(target.toNamed, enumTypeName)
+  case class EnumComponentChiselAnnotation(target: InstanceId, typeName: String) extends ChiselAnnotation {
+    def toFirrtl: EnumComponentAnnotation = EnumComponentAnnotation(target.toNamed, typeName)
```
jackkoenig, line 28: I appreciate this change overall (especially the removal of snake_case 🙂 ), but I'm a bit less sure here. I agree this naming is better, but it's a source and binary incompatible change (the latter is less relevant since we break binary compatibility on major versions). There are a non-trivial number of things to do to ensure this is at least source compatible though and I'm just not sure it's worth it.

If you're interested I can let you know what to do, otherwise let's back this part of the change out and I'll merge the rest.

Soya-Onishi: I was careless about compatibility. 
I'll revert this change, for now.

Soya-Onishi: I have a question about scaladoc.
This change originally came from mismatching scaladoc's param name, so I want to change param name from typeName into enumTypeName.
However, in terms of compatibility, is this safe?

jackkoenig: > This change originally came from mismatching scaladoc's param name, so I want to change param name from typeName into enumTypeName.

Ah yes having those in sync is pretty important. There's no risk with changing the ScalaDoc, go ahead!

jackkoenig (approved):
LGTM Thanks!

PR #1756, opened 2021-01-27
Fix incorrect comment in ScalaDoc

jackkoenig:
### Contributor Checklist

#### Type of Improvement

 - documentation

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Fix incorrect example in ScalaDoc for width behavior of RegInit

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Reg.scala
+++ b/core/src/main/scala/chisel3/Reg.scala
@@ -127,7 +127,7 @@ object RegNext {
   * val x = Wire(UInt())
   * val y = Wire(UInt(8.W))
   * val r1 = RegInit(x) // width will be inferred
-  * val r2 = RegInit(y) // width is set to 8
+  * val r2 = RegInit(y) // width will be inferred
   * }}}
   *
   * 3. [[Aggregate]] initializer - width will be set to match the aggregate


chick (approved):
I tested it on scastie, seems right now

core/src/main/scala/chisel3/Reg.scala
```diff
@@ -127,7 +127,7 @@ object RegNext {
   * val x = Wire(UInt())
   * val y = Wire(UInt(8.W))
   * val r1 = RegInit(x) // width will be inferred
-  * val r2 = RegInit(y) // width is set to 8
+  * val r2 = RegInit(y) // width will be inferred
```
edwardcwang, line 130: Isn't it set to 8 eventually? What's the mechanism?

jackkoenig: inference in FIRRTL

edwardcwang: Can we add it as a comment? It may not be obvious that it is inferred, especially if the example code doesn't assign anything else to r2 that could change its width.

chick: Here's the [scastie demo](https://scastie.scala-lang.org/b159bXm4R2yCkchFAzjFRA). I am not sure of the mechanism, but it confirms, I believe, the correctness of Jack's change

jackkoenig: Read the whole ScalaDoc, that’s sort of the point of this whole example: https://www.chisel-lang.org/api/3.4.1/chisel3/RegInit$.html

It’s just that that particular comment is wrong and contradicts the correct info above.

PR #1793, opened 2021-02-26
VecLiterals First Pass

chick:
Introduces a requested feature `Vec` literals. Much like `Bundle` literals, this allows
developers to specify a `Vec` aggregate literal in a convenient (relatively) compact form
e.g.
```
  val vecLiteral = Vec(2, UInt(4.W)).Lit(0 -> 0xA.U, 1 -> 0xB.U)
```
Features
- Can be defined outside of module (enables use with chiseltest library)
- Can be sparsely specified (with the effect that top level `litOption` returns None)
- litOption behaves like to UInt in bit ordering
- works as `Reg` initializer
- Decent error exceptions for
  - duplicate keys are found
  - wrong data types employed
  - width of individual element exceeds the basic `Vec` width

### Contributor Checklist

#### Type of Improvement

New feature

#### API Impact

This is an API addition. the `Vec` type now has a `Lit` method creates a `Vec` literal.

This should not impact generated verilog

#### Desired Merge Strategy

Squash: The PR will be squashed and merged

#### Release Notes
Introduces Vec literals which are similar to Bundle Literals. Simple example:
```scala
  val vecLiteral = Vec(2, UInt(8.W)).Lit(0 -> 0xA.U, 1 -> 0xB.U)
```

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -174,9 +174,6 @@ trait VecFactory extends SourceInfoDoc {
 sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
     extends Aggregate with VecLike[T] {
 
-  //TODO: what to do when lit widths are different than gen's width
-  //TODO: fix ordering problems with lits
-
   override def toString: String = {
     val bindingString = topBindingOpt match {
       case Some(VecLitBinding(vecLitBinding)) =>
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -520,7 +520,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
       }
     } // don't convert to a Map yet to preserve duplicate keys
 
-    clone.bind(VecLitBinding(vecLitLinkedMap))
+    clone.bind(VecLitBinding(ListMap(vecLitLinkedMap.toSeq:_*)))
     clone
   }
 }
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -50,38 +50,31 @@ sealed abstract class Aggregate extends Data {
     */
   override def litOption: Option[BigInt] = {
     // Shift the accumulated value by our width and add in our component, masked by our width.
-    def shiftAddBundle(accumulator: Option[BigInt], elt: Data): Option[BigInt] = (accumulator, elt.litOption()) match {
-      case (Some(accumulator), Some(eltLit)) =>
-        val width = elt.width.get
-        val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
-        Some((accumulator << width) + masked)
-      case (None, _) => None
-      case (_, None) => None
-    }
-
-    // Shift the accumulated value by our width and add in our component, masked by width of Vec element.
-    def shiftAddVec(accumulator: Option[BigInt], data: Data, elt: LitArg): Option[BigInt] = {
-      (accumulator, elt) match {
-        case (Some(accumulator), eltLit) =>
-          val width = data.width.get
-          val masked = ((BigInt(1) << width) - 1) & eltLit.num  // also handles the negative case with two's complement
+    def shiftAdd(accumulator: Option[BigInt], elt: Data): Option[BigInt] = {
+      (accumulator, elt.litOption()) match {
+        case (Some(accumulator), Some(eltLit)) =>
+          val width = elt.width.get
+          val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
           Some((accumulator << width) + masked)
         case (None, _) => None
-        case (_, _) => None
+        case (_, None) => None
       }
     }
+
     topBindingOpt match {
       case Some(BundleLitBinding(_)) =>
         getElements
           .reverse
-          .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAddBundle)
+          .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAdd)
       case Some(VecLitBinding(elementsToValue)) =>
         if (elementsToValue.size != this.asInstanceOf[Vec[_]].length) {
           None
         } else {
-          elementsToValue.toList.reverse.foldLeft[Option[BigInt]](Some(BigInt(0))) { case (accumulator, (data, litArg)) =>
-            shiftAddVec(accumulator, data, litArg)
-          }
+          elementsToValue
+            .toList
+            .map(_._1)
+            .reverse
+            .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAdd)
         }
       case _ => None
     }
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -5,14 +5,13 @@ package chisel3
 import scala.collection.immutable.ListMap
 import scala.collection.mutable.{HashSet, LinkedHashMap}
 import scala.language.experimental.macros
-
-import chisel3.experimental.BaseModule
-import chisel3.experimental.BundleLiteralException
-import chisel3.experimental.EnumType
+import chisel3.experimental.{BaseModule, BundleLiteralException, EnumType, VecLiteralException}
 import chisel3.internal._
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo._
+
+import scala.collection.mutable
 
 class AliasedAggregateFieldException(message: String) extends ChiselException(message)
 
@@ -50,7 +49,7 @@ sealed abstract class Aggregate extends Data {
     */
   override def litOption: Option[BigInt] = {
     // Shift the accumulated value by our width and add in our component, masked by our width.
-    def shiftAdd(accumulator: Option[BigInt], elt: Data): Option[BigInt] = (accumulator, elt.litOption()) match {
+    def shiftAddBundle(accumulator: Option[BigInt], elt: Data): Option[BigInt] = (accumulator, elt.litOption()) match {
       case (Some(accumulator), Some(eltLit)) =>
         val width = elt.width.get
         val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
@@ -58,11 +57,31 @@ sealed abstract class Aggregate extends Data {
       case (None, _) => None
       case (_, None) => None
     }
+
+    // Shift the accumulated value by our width and add in our component, masked by width of Vec element.
+    def shiftAddVec(accumulator: Option[BigInt], data: Data, elt: LitArg): Option[BigInt] = {
+      (accumulator, elt) match {
+        case (Some(accumulator), eltLit) =>
+          val width = data.width.get
+          val masked = ((BigInt(1) << width) - 1) & eltLit.num  // also handles the negative case with two's complement
+          Some((accumulator << width) + masked)
+        case (None, _) => None
+        case (_, _) => None
+      }
+    }
     topBindingOpt match {
       case Some(BundleLitBinding(_)) =>
         getElements
           .reverse
-          .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAdd)
+          .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAddBundle)
+      case Some(VecLitBinding(elementsToValue)) =>
+        if (elementsToValue.size != this.asInstanceOf[Vec[_]].length) {
+          None
+        } else {
+          elementsToValue.toList.reverse.foldLeft[Option[BigInt]](Some(BigInt(0))) { case (accumulator, (data, litArg)) =>
+            shiftAddVec(accumulator, data, litArg)
+          }
+        }
       case _ => None
     }
   }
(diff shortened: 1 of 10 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+      case (x: Vec[_], y: Vec[_]) =>
+        (x.getElements zip y.getElements).map { case (xElt, yElt) =>
+          getMatchedFields(xElt, yElt)
+        }.fold(Seq(x -> y)) {
+          _ ++ _
+        }
+    }
+
+    def checkLiteralConstruction(): Unit = {
+      val dupKeys = elementInitializers.map { x => x._1 }.groupBy(x => x).flatMap { case (k, v) =>
```
ducky64, line 410: There are shorter ways to detect duplicates, comparing size against `.distinct` or `.toSet`
https://stackoverflow.com/questions/7691971/easiest-way-to-decide-if-list-contains-duplicates

More set operations can probably get the duplicated keys.

You don't get the count, but I'd think the index is the important aspect.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+          None
+        }
+      }
+      if (dupKeys.nonEmpty) {
+        throw new VecLiteralException(
+          s"VecLiteral: has duplicated indices ${dupKeys.map { case (k, n) => s"$k($n times)" }.mkString(",")}"
+        )
+      }
+
+      val outOfRangeIndices = elementInitializers.map(_._1).flatMap {
```
ducky64, line 423: consider `.collect`

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+        case _ => None
+      }
+      if (outOfRangeIndices.nonEmpty) {
+        throw new VecLiteralException(
+          s"VecLiteral: The following indices (${outOfRangeIndices.mkString(",")}) " +
+            s"are less than zero or greater or equal to than Vec length"
+        )
+      }
+
+      val badLits = elementInitializers.flatMap {
```
ducky64, line 434: `.map` to transform to widths, followed by `.collect` to match on widths?

Also, I don't think we do width checking elsewhere in chisel frontend? We don't need to special-case Vecs.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+        throw new VecLiteralException(
+          s"VecLiteral: Vec[${gen}] has the following incorrectly typed or sized initializers: " +
+            badLits.map { case (a, b) => s"$a -> $b" }.mkString(",")
+        )
+      }
+
+    }
+
+    requireIsChiselType(this, "vec literal constructor model")
+    checkLiteralConstruction()
```
ducky64, line 458: Does it make sense to just inline the checks here, perhaps with comments for visual delimiting?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+
+    val clone = cloneType
+    val cloneFields = getRecursiveFields(clone, "(vec root)").toMap
+
+    // Create the Vec literal binding from litargs of arguments
+    val vecLitLinkedMap = new mutable.LinkedHashMap[Data, LitArg]()
+    elementInitializers.sortBy { case (a, _) => a }.foreach { case (fieldIndex, value) =>
```
ducky64, line 465: `_._1` ?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+      val clonedField = clone(fieldIndex)
+
+      val valueBinding = value.topBindingOpt match {
+        case Some(litBinding: LitBinding) => litBinding
+        case _ => throw new VecLiteralException(s"field $fieldIndex specified with non-literal value $value")
+      }
+
+      value match { // Get the litArg(s) for this field
+        case vecField: Vec[_] =>
+          if (!(vecField typeEquivalent value)) {
```
ducky64, line 476: Will this ever happen? vecField is the same object as value, is it not?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            throw new VecLiteralException(s"field $fieldIndex $vecField specified with non-type-equivalent value $value")
+          }
+          // Copy the source VecLitBinding with vecFields (keys) remapped to the clone
+          val remap = getMatchedFields(value, clonedField).toMap
+          value.topBinding.asInstanceOf[VecLitBinding].litMap.map { case (valueField, valueValue) =>
+            vecLitLinkedMap(remap(valueField)) = valueValue
+          }
+        case bitField: Bits =>
+          val adjustedBitField = if (bitField.isInstanceOf[Bool]) {
```
ducky64, line 485: Why is this special case needed?

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -91,6 +91,14 @@ abstract class LitArg(val num: BigInt, widthArg: Width) extends Arg {
     elem
   }
 
+  /** Provides a mechanism that LitArgs can have their width adjusted
+    * to match other members of a VecLiteral
+    *
+    * @param newWidth the new width for this
+    * @return
+    */
+  def cloneWithWidth(newWidth: Width): this.type
```
ducky64, line 100: Is this needed? Is the root cause here that it's possible to specify a width in both the Data and the LitArg and those two can be out of sync? Do we need to do something special for VecLiteral that needs this new code?

src/test/scala/chiselTests/BundleLiteralSpec.scala
```diff
@@ -47,6 +46,24 @@ class BundleLiteralSpec extends ChiselFlatSpec with Utils {
     } }
   }
 
+  "bundle literals" should "not pack when sparsely specified" in {
```
ducky64, line 49: Should this have gone in its own PR? Doesn't look related to Vec literals?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            bitField
+          } else {
+            bitField.cloneTypeWidth(field.getWidth.W)
+          }
+
+          val notSameClass = !field.typeEquivalent(adjustedBitField)
+          if (notSameClass) { // TODO typeEquivalent is too strict because it checks width
+            throw new VecLiteralException(
+              s"VecLit: Literal specified at index $fieldIndex ($value) does not match Vec type $sample_element"
```
ducky64, line 494: Do we do these checks in the frontend elsewhere? If so, the code should be unified with that, and if not, I don't think it's a good idea to have mixed semantics

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -123,3 +125,5 @@ sealed trait LitBinding extends UnconstrainedBinding with ReadOnlyBinding
 case class ElementLitBinding(litArg: LitArg) extends LitBinding
 // Literal binding attached to the root of a Bundle, containing literal values of its children.
 case class BundleLitBinding(litMap: Map[Data, LitArg]) extends LitBinding
+// Literal binding attached to the root of a Vec, containing literal values of its children.
+case class VecLitBinding(litMap: mutable.LinkedHashMap[Data, LitArg]) extends LitBinding
```
ducky64, line 129: I don't think this should be mutable. It might make sense to use mutable to build the final data structure, but the stored data structure should be immutable.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+      val field = apply(fieldIndex)
+      val clonedField = clone(fieldIndex)
+
+      val valueBinding = value.topBindingOpt match {
+        case Some(litBinding: LitBinding) => litBinding
+        case _ => throw new VecLiteralException(s"field $fieldIndex specified with non-literal value $value")
+      }
+
+      value match { // Get the litArg(s) for this field
+        case vecField: Vec[_] =>
```
ducky64, line 475: Does it make sense to keep the values consistent (value vs. vecField)? Otherwise it seems to be confusing to refer to either, when they're both the same thing?

ducky64 (changes requested):
At a high level this seems sane, but a few comments.

Overall themes:
- I think there's code paths that can be shared and DRYed out with the bundle lits code
- It might make sense to keep the strictness / checks consistent with normal elements, instead of special casing strict checks in vec literal construction? Good alternatives might to be apply them everywhere universally (and perhaps the shared infrastructure can lead to cleaner code), or just remove them (especially if they're caught in FIRRTL). In general, I think consistency in behavior is best.
- Some style / cleanliness issues

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -51,19 +50,39 @@ sealed abstract class Aggregate extends Data {
...
       case (Some(accumulator), Some(eltLit)) =>
         val width = elt.width.get
         val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
         Some((accumulator << width) + masked)
       case (None, _) => None
       case (_, None) => None
     }
+
+    // Shift the accumulated value by our width and add in our component, masked by width of Vec element.
+    def shiftAddVec(accumulator: Option[BigInt], data: Data, elt: LitArg): Option[BigInt] = {
```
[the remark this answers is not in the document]
chick: Reduced this to a single ShiftAdd for both Vec and Bundle

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -155,9 +174,18 @@ trait VecFactory extends SourceInfoDoc {
   */
 sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
     extends Aggregate with VecLike[T] {
+
   override def toString: String = {
+    val bindingString = topBindingOpt match {
+      case Some(VecLitBinding(vecLitBinding)) =>
+        val contents = vecLitBinding.map { case (data, lit) =>
+          s"$data=$lit"
```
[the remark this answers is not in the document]
chick: Good catch, now it shows indices, e.g. `SInt<4>[2](0=SLit(1,<4>), 1=SLit(-2,<4>)`

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +346,190 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+    * (mew MyVec).Lit(
+    *   _.a -> 42.U,
+    *   _.b -> true.B
+    * )
+    * }}}
+    */
+  private[chisel3] def _makeLit(elementInitializers: (Int, Data)*): this.type = {
+
+    // Returns pairs of all fields, element-level and containers, in a Record and their path names
+    def getRecursiveFields(data: Data, path: String): Seq[(Data, String)] = data match {
```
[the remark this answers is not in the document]
chick: I think this is a good idea. But I would not be very comfortable trying this right now. There's pretty much as many differences as there are similarities. Going DRY might break bundle lits in some subtle way, where as keeping it separate should keep any logic flaws limited to Vecs

src/test/scala/chiselTests/BundleLiteralSpec.scala
chick: Removed

core/src/main/scala/chisel3/internal/Binding.scala
chick: Switched after generation to immutable.ListMap, important to maintain order

core/src/main/scala/chisel3/Aggregate.scala
chick: I'm not completely sure. In general I think errors should be reported as early as possible and at compile time over run time (or firrtl time) also when possible. I think the error messages can be more helpful with information that is possessed at earlier stages

chick:
Added some mdoc for vec lits, plagiarized from bundle lits

docs/src/appendix/experimental-features.md
```diff
@@ -79,7 +79,85 @@ class Example3 extends RawModule {
...
+    Vec(4, UInt(8.W)).Lit(0 -> 0xAB.U(8.W), 1 -> 0xCD.U(8.W), 2 -> 0xEF.U(8.W), 3 -> 0xFF.U(8.W))
+  )
+}
+```
+
+```scala mdoc:verilog
+chisel3.stage.ChiselStage.emitVerilog(new Example3)
+```
+
+Registers can be partially initialized from Vec literals. Only specified elements will be assigned.
```
mwachs5, line 127: does that mean that some registers will have reset values and some will not? Can you make that more clear in this sentence, whatever the answer is?

chick:
Closing this in favor of PR #1834

PR #1795, opened 2021-03-01
Adding ChiselEnum Documentation Entry

chrisbaldwin2:
Added documentation for the ChiselEnum type with verified examples

### Contributor Checklist

#### Type of Improvement

- documentation

#### API Impact

No API Impact

No Backend Code Generation Impact

#### Desired Merge Strategy

- Squash

#### Release Notes

Added documentation for the ChiselEnum object type.

- [ ] Did you review?


chrisbaldwin2:
Just adding some documentation for the ChiselEnum type

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,129 @@
+# ChiselEnum
+
+The ChiselEnum type is an easy way to get away from value encodings and reduce the chance of error when encoding muxes, opcodes, and functional unit operations. As opposed to regular Enums, ChiselEnums allow for function names to be explicitly passed through Bundles like IO.
```
mwachs5, line 3: Thanks for writing this!

what is a "regular Enum"? Do you mean `scala.Enum`, or something else?

Also, not sure what you mean by "allow for function names". What is a "function name"?

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,129 @@
...
+class AluMux1File(val dl_size: Int) extends Module {
+    val io = IO(new Bundle {
+        val aluMux1Sel =  Input( AluMux1Sel() )
+        val rs1Out     =  Input(Bits(dl_size.W))
+        val pcOut      =  Input(Bits(dl_size.W))
+        val aluMux1Out = Output(Bits(dl_size.W))
+    })
+
+    // Default value for aluMux1Out
+    io.aluMux1Out := 0.U
```
mwachs5, line 43: shouldn't the default be one of the `ChiselEnum` types?

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+```
+```scala mdoc:verilog
+import 
+ChiselStage.emitVerilog.emitVerilog(new AluMux1File(dl_size = 32) )
+```
+
+
+ChiselEnum also allows for the user to define variables by passing in the value shown below. Note that the value must be increasing.
```
mwachs5, line 61: What happens if the value isn't increasing? Is it a runtime error?

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,129 @@
...
+        "bge"  -> 5.U,
+        "bltu" -> 6.U,
+        "bgeu" -> 7.U
+    */
+}
+```
+
+## Testing
+
+When testing your modules, the `.Type` and `.litValue` attributes allow for the the objects to be passed as parameters and for the value to be converted to BigInt type. Note that BigInts cannot be casted to Int with `.asInstanceOf[Int]`, they use their own methods like `toInt`. Please review the [scala.math.BigInt](https://www.scala-lang.org/api/2.12.5/scala/math/BigInt.html) page for more details!
```
mwachs5, line 97: the example below is a nice example of `litValue`, can you show an example of `.Type` usage?

Also, does `sel.litValue == 0.U` also work? It seems more intuitive... actually it seems way more intuitive to actually use the Enum values in such a comparison. The below seems like a pattern that should be discouraged

docs/src/explanations/chisel-enum.md
chrisbaldwin2: By regular Enum I mean the Chisel enum. I can clarify that. 
By function name I mean like a opcode being passed into an ALU or CMP.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,129 @@
...
+    val ukn = Value(7.U)
+    /**
+        "sb" -> 0.U,
+        "sh" -> 1.U,
+        "sw" -> 2.U
+    */
+}
+```
+
+Additionally I don't believe that signed values are supported so if you want the value signed, you must cast the UInt with `.asSInt`.
```
mwachs5, line 125: Before this PR is merged let's get rid of the `I don't believe` and just make sure what is written is correct.

docs/src/explanations/chisel-enum.md
chrisbaldwin2: It can be a default value depends on whether logic is being simplified or error checked. Something like `DontCare` or `0.U` can be used for checking that unknown selects are used. I believe that this is a matter of implementation.

mwachs5: Oh, I misread the example, the type is on the select not the data.

docs/src/explanations/chisel-enum.md
chrisbaldwin2: An exception is thrown when Chisel emits Verilog.
> chisel3.internal.ChiselException: Exception thrown when elaborating ChiselGeneratorAnnotation

docs/src/explanations/chisel-enum.md
chrisbaldwin2: Correct!

docs/src/explanations/chisel-enum.md
chrisbaldwin2: .Type is used in line 100. In the testbench, the .Type function is used to pass the ChiselEnum as a parameter since we do not want the literal value!

chrisbaldwin2: `sel.litValue == 0.U` does not work since .litValue returns a BigInt and not a UInt. This block is pulled out of my testbench which tests that the ChiselEnum type is enumerating correctly and that all values match as expected.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,131 @@
...
+
+```scala 
+def expectedSel(sel: AluMux1Sel.Type): Boolean = sel match {
+  case AluMux1Sel.rs1out => (sel.litValue == 0)
+  case AluMux1Sel.pcout  => (sel.litValue == 1)
+  case _                 => false
+}
+```
+
+The ChiselEnum type also has methods `.all` and `.getWidth` where all returns all of the enum instances and getWidth returns the width of the hardware type.
```
mwachs5, line 109:
```suggestion
The ChiselEnum type also has methods `.all` and `.getWidth` where `all` returns all of the enum instances and `getWidth` returns the width of the hardware type.
```

chrisbaldwin2:
@jackkoenig do you know why this is failing to emitVerilog? I have the code building here https://scastie.scala-lang.org/4NTKq8i2Rh228cDRQzImbg but fails ci.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,131 @@
+# ChiselEnum
+
+The ChiselEnum type is an easy way to get away from value encodings and reduce the chance of error when encoding muxes, opcodes, and functional unit operations. As opposed to the Chisel type Enums, ChiselEnums allow for names to be explicitly passed through Bundles like IO.
```
mwachs5, line 3:
```suggestion
The ChiselEnum type can be used to reduce the chance of error when encoding mux selectors, opcodes, and functional unit operations. In contrast with`Chisel.Enum`, `ChiselEnum` can be included in `Bundle`s.
```

I still don't know what you mean by "names to be explicitly passed" or "through Bundles like IO". Even from an english language level, I don't know if "like IO" means "Bundles which are similar to IO" or "similar to the way IO can be passed through Bundles" Is the above correct?

chrisbaldwin2: Well I wanted to explicitly mention `IO` because it is the most prevalent use of `Bundle`s. When I was first trying to use `Enum` for my mux selects, I couldn't pass the `Enum` type through the `IO` of my modules. This is only possible with `ChiselEnum`s which define a `Data` type that can be passed through Bundles and IO. Since `IO` takes in a `Bundle` object, I was trying to make it clear that `ChiselEnum`s can be passed through `Bundle`s like those used in `IO` while `Enum` objects cannot

mwachs5: thanks for the explanation! What do you think of my suggested change above? (Feel free to use any/all/none of it)

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,131 @@
...
+// }
+```
+
+Here we see a mux using the AluMux1Sel to select between different inputs. 
+
+```scala mdoc
+import AluMux1Sel._
+
+class AluMux1File extends Module {
+    val io = IO(new Bundle {
```
mwachs5, line 37: i think this is the reason your thing is failing CI, there is a bug that shows up in `mdoc` + naming plugin +` IO( new Bundle {...})`

You can try workaround this by defining your bundle type outside of the IO call:

```
class ALUMuxIO extends Bundle {
...
}
...

val io = IO(new ALUMuxIO)
```

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,131 @@
...
+Below we see ChiselEnum being used as mux select for a RISC-V core. While wrapping the object in a package is not required, it is highly recommended as it allows for the type to be used in multiple files more easily. 
+
+```scala mdoc
+// package CPUTypes {
+    object AluMux1Sel extends ChiselEnum {
+        val rs1out, pcout = Value
+        /**
+            "rs1out" -> 0.U,
+            "pcout"  -> 1.U
+        */
```
mwachs5, line 26:
```suggestion
        val selectRS1, selectPC = Value
        /**
            "selectRS1" -> 0.U,
            "selectPC"  -> 1.U
        */
```

for the purposes of a clear example, can I suggest more generic names above?

docs/src/explanations/chisel-enum.md
chrisbaldwin2: It looks good. The only thing that I would change is that `Enum` is in the `Chisel.util` package, not in `Chisel`. I don't know if you were trying to reference the package or not.

docs/src/explanations/chisel-enum.md
chrisbaldwin2: Oh that's a pretty nasty issue. I separated the bundle class out and looks like it is passing the ci now. Thanks!

docs/src/explanations/chisel-enum.md
mwachs5: yeah I was trying to reference the package to be explicit what we meant by "Enum". I'll fix it to say util

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,133 @@
+# ChiselEnum
+
+The ChiselEnum type is an easy way to get away from value encodings and reduce the chance of error when encoding muxes, opcodes, and functional unit operations. As opposed to the Chisel type Enums, ChiselEnums allow for names to be explicitly passed through Bundles like IO. 
+
+## Importing the ChiselEnum module
+
+```scala mdoc
```
mwachs5, line 7: can you break these up or make some of these invisible/silent since `ChiselStage` is not relevant to to "Importing the Chisel Enum module" ?

mwachs5: I'm not sure the general pattern in these docs if things that are only relevant for the examples are hidden or if the code examples are all supposed to work right out of the box

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+```
+```scala mdoc:verilog
+ChiselStage.emitVerilog(new AluMux1File )
+```
+
+ChiselEnum also allows for the user to define variables by passing in the value shown below. Note that the value must be increasing or else 
+ > chisel3.internal.ChiselException: Exception thrown when elaborating ChiselGeneratorAnnotation
+is thrown during Verliog generation.
```
mwachs5, line 65:
```suggestion
is thrown during Verilog generation.
```

docs/src/explanations/chisel-enum.md
mwachs5: gotcha. I can't resolve this comment for some reason but you've answered my question

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,133 @@
...
+  case AluMux1Sel.selectPC  => (sel.litValue == 1)
+  case _                 => false
+}
+```
+
+The ChiselEnum type also has methods `.all` and `.getWidth` where `all` returns all of the enum instances and `getWidth` returns the width of the hardware type.
+
+## Workarounds
+
+As of 2/26/2021, the width of the values is always infered so to get around this, just add an extra state that forces the width that is desired.
```
mwachs5, line 115:
```suggestion
As of 2/26/2021, the width of the values is always inferred. To work around this, you can add an extra `Value` that forces the width that is desired. This is shown in the example below, where we add a field `ukn` to force the width to be 3 bits wide: 
```

docs/src/explanations/chisel-enum.md
chrisbaldwin2: In referencing https://raw.githubusercontent.com/chipsalliance/chisel3/master/docs/src/explanations/naming.md they also had ChiselStage shown

mwachs5: yeah, gotcha. Then let's follow that pattern a bit more ... I'll suggest a change

chrisbaldwin2: I also believe this helps people who want to copy what is on screen and get it to run locally on their machine.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,131 @@
...
+```scala mdoc
+import chisel3._
+import chisel3.util._
+import chisel3.stage.ChiselStage
+import chisel3.experimental.ChiselEnum
+
+```
+
+## Functionality and Examples
+
```
mwachs5, line 16:
```suggestion

## Functionality and Examples

```scala mdoc
// Imports used in the following examples
import chisel3._
import chisel3.util._
import chisel3.experimental.ChiselEnum
```

```

Comment (chrisbaldwin2) [reply docs/src/explanations/chisel-enum.md:16]: I had the import in separate ticks from the example. You removed the end ticks from "Importing the ChiselEnum module" but not the start ticks of the next block

Comment (mwachs5) [reply docs/src/explanations/chisel-enum.md:16]: yeah but i think that is an artifact of github's suggest changes API... anyway, consider this just as a suggestion

File: docs/src/explanations/chisel-enum.md [hunk 14]
```diff
@@ -0,0 +1,130 @@
...
+
+## Functionality and Examples
+
+```scala mdoc
+// Imports used in the following examples
+import chisel3._
+import chisel3.util._
+import chisel3.stage.ChiselStage
+import chisel3.experimental.ChiselEnum
```
mwachs5, line 13:
```suggestion
import chisel3.experimental.ChiselEnum
```

```

Comment (mwachs5) [reply docs/src/explanations/chisel-enum.md:13]: sigh. i swear i am suggesting the additional backticks!

File: docs/src/explanations/chisel-enum.md [hunk 15]
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+```
+```scala mdoc:verilog
+ChiselStage.emitVerilog(new AluMux1File )
+```
+
+ChiselEnum also allows for the user to define variables by passing in the value shown below. Note that the value must be increasing or else 
+ > chisel3.internal.ChiselException: Exception thrown when elaborating ChiselGeneratorAnnotation
+is thrown during Verliog generation.
```
mwachs5: something weird here in the formatting. You can actually make the `mdoc` just display the thrown error by doing the bad thing in the code

docs/src/explanations/chisel-enum.md
chrisbaldwin2: I pushed a new commit with backticks

docs/src/explanations/chisel-enum.md
chrisbaldwin2: I added space around the block quote for now. I'm worried that the ci will start going crazy if I intentionally throw errors in the mdoc

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+// package CPUTypes {
+    object AluMux1Sel extends ChiselEnum {
+        val selectRS1, selectPC = Value
+        /**
+            "selectRS1" -> 0.U,
+            "selectPC"  -> 1.U
+        */
+    }
+// }
+```
```
seldridge, line 28: To narrowly document `ChiselEnum`, it's sufficient to ignore suggestions of how a user should organize/package their code.

That said, if you want to show this as living under `CPUTypes`, I think that's reasonable. Because Scala seemingly treats objects and packages as basically the same, you can show what you want with:
```scala
object CPUTypes {
  object AluMux1Sel extends Chisel Enum { /** */ }
}
```

seldridge: Additionally, it would be good to make clear what the `"selectRS1" -> 0.U` portion is supposed to be doing and what that does. It would be fine to show this with two different variants of `AluMux1Sel`, e.g., `AluMux1Sel` and `AluMex1SelExplicitValues` or something.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+    val imm   = Value(0x13.U) // i imm   -> 001_0011
+    val auipc = Value(0x17.U) // u auipc -> 001_0111
+    val store = Value(0x23.U) // s store -> 010_0011
+    val reg   = Value(0x33.U) // r reg   -> 011_0011
+    val lui   = Value(0x37.U) // u lui   -> 011_0111
+    val br    = Value(0x63.U) // b br    -> 110_0011
+    val jalr  = Value(0x67.U) // i jalr  -> 110_0111
+    val jal   = Value(0x6F.U) // j jal   -> 110_1111
+}
+```
```
seldridge, line 80: Any code block should use the `scala mdoc` tag. This this to get type checked by the compiler and ensure that our documentation isn't heinously busted. 😄

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+
+The user can 'jump' to a value and continue incrementing by passing a start point then using a regular Value assignment. 
+
+```scala
```
seldridge, line 84:
```suggestion
```scala mdoc
```

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+        "bgeu" -> 7.U
+    */
+}
+```
+
+## Testing
+
+When testing your modules, the `.Type` and `.litValue` attributes allow for the the objects to be passed as parameters and for the value to be converted to BigInt type. Note that BigInts cannot be casted to Int with `.asInstanceOf[Int]`, they use their own methods like `toInt`. Please review the [scala.math.BigInt](https://www.scala-lang.org/api/2.12.5/scala/math/BigInt.html) page for more details!
+
+```scala
```
seldridge, line 104:
```suggestion
```scala mdoc
```

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+The ChiselEnum type also has methods `.all` and `.getWidth` where `all` returns all of the enum instances and `getWidth` returns the width of the hardware type.
+
+## Workarounds
+
+As of 2/26/2021, the width of the values is always inferred. To work around this, you can add an extra `Value` that forces the width that is desired. This is shown in the example below, where we add a field `ukn` to force the width to be 3 bits wide: 
+
+```scala
```
seldridge, line 118:
```suggestion
```scala mdoc
```

seldridge (commented):
Thanks for your first PR! 👏 I really appreciate you taking the time to write this up.

This is looking good and I'll leave it to @mwachs5 to continue handling the actual content review.

My only real concern is that everything needs to run through `mdoc` so that it's type checked. Use the `scala mdoc` tag instead of a bare `scala` tag. 

Two FYI comments:
- You can build this locally to prevent having to wait on CI with `sbt docs/mdoc` (or keep an `sbt` shell open and just run `docs/mdoc` in it). The output, post-`mdoc` processing shows up in `docs/generated`.
- There are a lot of useful `mdoc` modifiers documented [here](https://scalameta.org/mdoc/docs/modifiers.html). This includes things for code which is expected to error.

docs/src/explanations/chisel-enum.md
seldridge: For an expected exception, I think `scala mdoc:crash` is what you want. There's also `scala mdoc:fail`. Lots of documentation on this [here](https://scalameta.org/mdoc/docs/modifiers.html).

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+ChiselStage.emitVerilog(new AluMux1File )
+```
+
+ChiselEnum also allows for the user to define variables by passing in the value shown below. Note that the value must be increasing or else 
+
+ > chisel3.internal.ChiselException: Exception thrown when elaborating ChiselGeneratorAnnotation
+
+is thrown during Verilog generation.
+
+```scala
```
mwachs5, line 68:
```suggestion
```scala mdoc
```

docs/src/explanations/chisel-enum.md
chrisbaldwin2: Right so this the mapping comment comes from a misunderstanding I had with the value assignment/verification code in the example StrongEnum code. This is the actual code that I have in my file

```scala
object AluMux1Sel extends ChiselEnum {
        val rs1out, pcout = Value
        val correct_annotation_map = Map[String, UInt](elems =
            "rs1out" -> 0.U,
            "pcout"  -> 1.U
        )
    }
```

When I first saw this mapping function, I thought the mapping is what actually assigned the value BUT IT DOESN'T!!! For the documentation I removed the actual function because it is only for verification and not for assigning value.

docs/src/explanations/chisel-enum.md
```diff
@@ -0,0 +1,134 @@
...
+        "sw" -> 2.U
+    */
+}
+```
+
+Signed values are not supported so if you want the value signed, you must cast the UInt with `.asSInt`.
+
+## Additional Resources
+
+The ChiselEnum type is much more powerful than stated above. It allows for Sequence, Vec, and Bundle assignments as well as the `.next` operation to allow for stepping through sequential states and the `.isValid` for checking that the value is an enum type. The source code for the ChiselEnum can be found [here](https://github.com/chipsalliance/chisel3/blob/2a96767097264eade18ff26e1d8bce192383a190/core/src/main/scala/chisel3/StrongEnum.scala) in the class `EnumFactory`. An example file for many of the ChiselEnum operations can be found [here](https://github.com/chipsalliance/chisel3/blob/dd6871b8b3f2619178c2a333d9d6083805d99e16/src/test/scala/chiselTests/StrongEnum.scala).
```
mwachs5, line 134:
```suggestion
The ChiselEnum type is much more powerful than stated above. It allows for Sequence, Vec, and Bundle assignments, as well as a `.next` operation to allow for stepping through sequential states and an `.isValid` for checking that a hardware value is a valid `Value`. The source code for the ChiselEnum can be found [here](https://github.com/chipsalliance/chisel3/blob/2a96767097264eade18ff26e1d8bce192383a190/core/src/main/scala/chisel3/StrongEnum.scala) in the class `EnumFactory`. Examples of the ChiselEnum operations can be found [here](https://github.com/chipsalliance/chisel3/blob/dd6871b8b3f2619178c2a333d9d6083805d99e16/src/test/scala/chiselTests/StrongEnum.scala).
```

docs/src/explanations/chisel-enum.md
mwachs5: i think the current form of the doc is helpful, I found it clear what it was saying

mwachs5:
@seldridge, @jackkoenig general question about this... it seems that since `ChiselEnum` is in the `experimental` package, it doesn't show up on the API docs? What's the general policy on including things from the `experimental`package in the API docs / the website documentation? It seems strange that we would have a nice doc like this but not an API doc. If the policy is to not put experimental features into the API docs, maybe we should have a docs division of "experimental features" to make this clear

docs/src/explanations/chisel-enum.md
seldridge: Ah, I get it now. Maybe just extend the comment to say: "This is what the mapping will be" or something. Not blocking. I'm fine with it the way it is. :+1:

seldridge (approved):

mwachs5:
@chrisbaldwin2 can you please fill out the contributor checklist at the top? this is looking good!

seldridge:
@mwachs5 wrote:
> general question about this... it seems that since ChiselEnum is in the experimental package, it doesn't show up on the API docs (https://www.chisel-lang.org/api/3.4.2/?search=ChiselEnum)? 

I'm not sure what's going on here, but it's not omitted intentionally. FWIW, it _does_ exist, but it appears that Scaladoc search doesn't work for type definitions? I.e., it exists here: https://www.chisel-lang.org/api/latest/chisel3/experimental/index.html#ChiselEnum=chisel3.experimental.EnumFactory, but I can't search for that.

The organization of the `ChiselEnum` stuff always seemed weird to me and worth a refactor... Why is it `type ChiselEnum = EnumFactory`? 

@mwachs5 wrote:
> What's the general policy on including things from the experimental package in the API docs / the website documentation? It seems strange that we would have a nice doc like this but not an API doc. If the policy is to not put experimental features into the API docs, maybe we should have a docs division of "experimental features" to make this clear

I think everything should be documented unless it's a non-public API that we're trying to hide. (The only thing in that category is anything `private [chisel3]` which should already be stripped and the public `do_foo` methods which use some magic to put them in a low priority group.) Experimental stuff does appear to be documented, but it's `ChiselEnum` that search doesn't like.

Adding API docs sounds like a good feature request. :+1:

docs/src/explanations/chisel-enum.md
```diff
@@ -19,7 +19,7 @@ Below we see ChiselEnum being used as mux select for a RISC-V core. While wrappi
 // package CPUTypes {
     object AluMux1Sel extends ChiselEnum {
         val selectRS1, selectPC = Value
-        /**
+        /** How the values will be mapped
```
seldridge, line 22: :100:

chrisbaldwin2:
I updated the PR checklist information. I believe that this is ready to be merged! Thanks for your time and help! I hope that this documentation can lead to more people using this great feature of Chisel!!!

mwachs5:
I don't have write access, but I approve. Thanks for doing this!

seldridge:
Thanks again for the contribution @chrisbaldwin2. This will show up on the website after the Chisel 3.4.3 release goes out (not sure on ETA for that).

PR #1804, opened 2021-03-05
Autoclonetype2

### LINKED ISSUE #1803: Bad interaction between autoclonetype and the compiler plugin
I haven't been able to identify exactly what the issue is, but sometimes you can get a `NullPointerException` in autoclonetype when using the compiler plugin. The issue shows up in `v3.4.1` and `v3.4.2`.

Showing issue: https://scastie.scala-lang.org/urTiFHCRRB6mWWUFjKJ6xw
v3.4.2 no plugin no issue: https://scastie.scala-lang.org/Q9cb7NnETDq8p6B2tkWVaA
v3.4.1 plugin has issue: https://scastie.scala-lang.org/awinn0u7QeCr9GyMQkLHNw
v3.4.0 plugin no issue: https://scastie.scala-lang.org/nwuijlIuQDChB3zSjEi8cQ

If you hit `Download` on Scastie, you can reproduce locally. Notably, Scastie adds partial unification and kind projector plugins which I think are part of the problem. 

I believe the issue is also related to https://github.com/chipsalliance/chisel3/pull/1660 (released in `v3.4.1`) which made `autoNameRecursively` lazy.

I believe this issue will be fixed by my forthcoming compiler plugin autoclonetype

**Type of issue**: bug report

**Other information**
 

jackkoenig:
Fixes #1803 (by coincidence as far as I can tell)

I need to benchmark but this should make elaboration *much* faster for designs with lots of Bundles.

Note that the diff is larger than it really is, I split `ChiselPlugin.scala` into 2 files. I did not make any changes to the `ChiselComponent` (which is really the `NamingComponent`). The bulk of my changes are in `BundleComponent`. I originally named it `AutoCloneTypeComponent` but I realized that it's not too big of an extension on this work to make the plugin also implement `val elements` in Bundles as well as solve the "Field problem". I almost did the latter, but it has some potential consequences that make me think it should be done on a major version boundary and I'd like to include this in `v3.4.3`. If we also implement `val elements` in the plugin, then there is no more runtime reflection necessary in Chisel.

Note the 2 new soft links in `no-plugin-tests`, reflective autoclonetype will continue being tested with all of the old tests. Those tests also apply to the new implementation along with some new tests I added.

I did add an assertion to reflective autoclonetype with the purpose of checking whether the plugin-generated version is actually in use. The cost of this assertion should be basically free compared to reflection; furthermore, users should use the plugin anyway for a massive performance improvement (to be measured).

### Contributor Checklist

#### Type of Improvement

 - bug fix       
 - performance improvement (no more collecting stack traces for every created bundle!)
 - code refactoring  (the plugin is a bit opaque, it is much clearer than the reflective version)

#### API Impact

This expands the autoclonetype API to support Bundles with parameters that are not vals. It also can generate cloneTypes for anonymous classes with non-Bundle/Module outer classes.

No impact

#### Desired Merge Strategy

  - Rebase (I'd like to keep the 4 commits)

#### Release Notes

Generate `cloneType` methods for Bundles in the compiler plugin when provided `-P:chiselplugin:useBundlePlugin` as a `scalacOption`.This feature is disabled by default because it is a breaking change to implement `cloneType` for any non-final Bundle (a child class extending the given Bundle could rely on reflective autoclonetype but will now call the newly implemented one in the superclass instead). Users who intend to publish libraries should not enable the feature until updating to Chisel 3.5. Everyone else should use it beginning in Chisel 3.4.3.

- [ ] Did you review?


src/test/scala/chiselTests/AutoCloneTypeSpec.scala
```diff
@@ -216,6 +236,83 @@ class AutoClonetypeSpec extends ChiselFlatSpec with Utils {
...
+      }
+      elaborate { new MultiIOModule {
+        implicit val x = 8
+        val in = IO(Input(new MyBundle))
+        val out = IO(Output(new MyBundle))
+        out := in
+      }}
+    }
+
+    it should "support Bundles with multiple parameter lists" in {
```
mwachs5, line 280: what about a case that is the "normal" one of 2 parameter lists, first normal second implicit?

jackkoenig: I tried to avoid writing redundant tests and there's an existing test that covers that case (at least when both sets of parameter lists are `vals`)

jackkoenig:
For the record, I think the CI failure is unrelated. The test appears to be failing on master, it might be due to a change in FIRRTL. I'll debug soonish but may not be able to get to it till early next week.

plugin/src/main/scala-2.12/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -0,0 +1,124 @@
...
+        val withMethods = deriveClassDef(bundle) { t =>
+          deriveTemplate(t)(_ :+ cloneTypeImpl :+ usingPlugin)
+        }
+
+        super.transform(localTyper.typed(withMethods))
+
+      case _ => super.transform(tree)
+    }
+  }
+}
```
azidar, line 124: Newline.

plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -0,0 +1,214 @@
...
+              super.transform(tree)
+          }
+        } else {
+          super.transform(tree)
+        }
+      // Otherwise, continue
+      case _ => super.transform(tree)
+    }
+  }
+}
```
azidar, line 214: Newline.

azidar (approved):
LGTM, minor changes (add newlines to new files)

azidar:
Looks stellar! Great job, this is very exciting and will have enormous impact on runtimes of large designs. Very exciting!!!

jackkoenig:
New TODO. I'd still like to backport this to `3.4.x`, but I have realized (by lucky accident) that this PR currently is not binary compatible. It's not that it changes any APIs and it doesn't (or at least, shouldn't), but it turns out that implementing `cloneType` on a non-final Bundle is not binary compatible. This is due to the fact that any Bundles that extend your Bundle may have been relying on reflective autoclonetype are now calling your method which will result in a runtime error. Because this generates a bunch of `cloneTypes` (well `_cloneTypeImpl` but the exact same idea applies), this would be a problem for anyone extending any Bundles in `chisel3.util`, for example, [rocket-chip](https://github.com/chipsalliance/rocket-chip/blob/9699c48d1f80ff9424509bdd1ef4b1f9393b2d06/src/main/scala/util/ShiftQueue.scala#L17).

The solution that allows users to still benefit from this without having a binary incompatible change in 3.4.3 is to make this behavior optional via an argument to the plugin, off by default. This Chisel itself will *not* use the feature in this PR which can be backported, and then will use it on master looking toward 3.5.

mwachs5:
> This Chisel itself will not use the feature in this PR which can be backported, and then will use it on master looking toward 3.5.

The implication being that users of this chisel can turn the plugin on for *their* code, and still link against the 3.4.x Chisel compiled with this bit of the plugin turned *off*, right?

jackkoenig:
> The implication being that users of this chisel can turn the plugin on for their code, and still link against the 3.4.x Chisel compiled with this bit of the plugin turned off, right?

Exactly, Chisel itself will remain binary compatible, anyone who wants to write a binary compatible library should keep this feature disabled, but anyone who wants the performance benefit in their own code can turn it on.

src/test/scala/chiselTests/AutoCloneTypeSpec.scala
```diff
@@ -0,0 +1,332 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests
+
+import chisel3._
+import chisel3.testers.TestUtils
+
+class BundleWithIntArg(val i: Int) extends Bundle {
```
ducky64, line 8: Can these (or most, or some) of these be inlined / placed closer to the test in which they are used, so tests are a bit more readable? For example, this only seems to be used in one place.

ducky64 (commented):
I can't say I understand what all of this is doing, but it look fine.

Otherwise, a high-level description of the transform / plugin behavior, ideally in scaladoc, would be helpful as a guide for reading through the implementation.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -311,6 +311,8 @@ private[chisel3] class DynamicContext(val annotationSeq: AnnotationSeq) {
   val components = ArrayBuffer[Component]()
   val annotations = ArrayBuffer[ChiselAnnotation]()
   var currentModule: Option[BaseModule] = None
+  // This is only used for testing, it can be removed if the plugin becomes mandatory
+  var allowReflectiveAutoCloneType = true
```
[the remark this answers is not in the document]
jackkoenig: It's not expected to change, (and never in user code, the setter is package private to `chisel3`). This is used for testing the `plugin` itself, from the changes to the tests:
```scala
  // Allows us to check that the compiler plugin cloneType is actually working
  val usingPlugin: Boolean = (new Bundle { def check = _usingPlugin }).check
  def elaborateNoReflectiveAutoCloneType(f: => RawModule): Circuit = {
    ChiselStage.elaborate {
      chisel3.internal.Builder.allowReflectiveAutoCloneType = !usingPlugin
      f
    }
  }
```

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -798,42 +798,71 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+  }
+
+  override def cloneType: this.type = {
+    val clone = _cloneTypeImpl.asInstanceOf[this.type]
+    checkClone(clone)
+    clone
+  }
 
-  override def cloneType : this.type = {
+  /** Implementation of cloneType. This should _never_ be overridden or called in user-code
```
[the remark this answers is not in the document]
jackkoenig: > Perhaps clarify that this implementation is the reflective auto clone type?

Will do

> And might not be called at all if the compiler plugin is active?

The `note` sort of says that but I can make it more explicit

plugin/src/main/scala-2.12/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -0,0 +1,124 @@
...
+  val phaseName: String = "chiselbundlephase"
+  val runsAfter: List[String] = "typer" :: Nil
+  def newPhase(prev: Phase): Phase = new BundlePhase(prev)
+
+  private class BundlePhase(prev: Phase) extends StdPhase(prev) {
+    override def name: String = phaseName
+    def apply(unit: CompilationUnit): Unit = {
+      // This plugin doesn't work on Scala 2.11. Rather than complicate the sbt build flow,
+      // instead we just check the version and if its an early Scala version, the plugin does nothing
+      if (scala.util.Properties.versionNumberString.split('.')(1).toInt >= 12) {
```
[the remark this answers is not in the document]
jackkoenig: Yeah I will make it check that too, although none of this code will compile in Scala 3, it'll still be nice to make it more clear.

plugin/src/main/scala-2.12/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -0,0 +1,124 @@
...
+    // Cached because this is run on every argument to every Bundle
+    def isData(sym: Symbol): Boolean = isDataCache.getOrElseUpdate(sym.tpe, sym.tpe <:< dataTpe)
+
+    def cloneTypeFull(tree: Tree): Tree =
+      localTyper.typed(q"chisel3.experimental.DataMirror.internal.chiselTypeClone[${tree.tpe}]($tree)")
+
+    def isNullaryMethodNamed(name: String, defdef: DefDef): Boolean =
+      defdef.name.decodedName.toString == name && defdef.tparams.isEmpty && defdef.vparamss.isEmpty
+
+    def getParamsAndConstructor(body: List[Tree]): (Option[DefDef], Seq[Symbol]) = {
```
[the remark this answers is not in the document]
jackkoenig: I didn't actually modify this code in this PR, just split it out. I'd rather not touch it in this PR just to help with the Git history.

plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -0,0 +1,214 @@
...
+      if(scala.util.Properties.versionNumberString.split('.')(1).toInt >= 12) {
+        unit.body = new MyTypingTransformer(unit).transform(unit.body)
+      }
+    }
+  }
+
+  class MyTypingTransformer(unit: CompilationUnit)
+    extends TypingTransformer(unit) {
+
+    private def shouldMatchGen(bases: Tree*): Type => Boolean = {
```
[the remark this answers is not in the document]
jackkoenig: I didn't modify this code in this PR, I just split it out into its own file. I'd rather not touch it while moving it just to keep the Git history more clear.

src/test/scala/chiselTests/AutoCloneTypeSpec.scala
jackkoenig: These are the original tests for reflective autoclonetype. The tests I added start at line 256. They need to be left alone for the time being since they're still being used by the "noPluginTests."

jackkoenig: I'm going to undo the file rename to make the diff more clear.

jackkoenig:
Feedback resolved and new option added so that this feature is opt-in rather than opt-out (for Chisel `3.4.x`). It is not used by chisel3 itself yet, and the binary and source compatibility issue related to this is now tested.

I intend to merge this shortly and let it backport, and then create a follow on PR for Chisel 3.5 that uses the plugin, perhaps changing the option to opt-out rather than opt-in.

PR #1820, opened 2021-03-18
Don't toggle top.cpp clock and reset on same cycle

seldridge:
Change top.cpp to deassert reset one time unit before the clock
asserts.  This avoids a simultation race condition if Chisel and a
FIRRTL compiler emit Verilog where a synchronous reset is not inside
the always block even though the Verilog is formally equivalent.

Reset now deasserts on time unit 10 and the clock ticks on time unit
11.

Concretely, the previous formulation of `top.cpp` is sensitive to changes in the Verilog emission that cropped up when trying to use `llvm/circt` purely for the Low -> Verilog path. This would result in things like the following two formally equivalent circuits resulting in simulation mismatches.

The Scala FIRRTL Compiler produces something like:
```verilog
module A(input clock, reset);
  reg [1:0] count;
  wire [1:0] _wrap_value_T_1 = count + 2'h1;
  always @(posedge clock) begin
    if (reset)
      count <= 2'h0;
    else
      count <= _wrap_value_T_1;
  end
endmodule
```

The MLIR FIRRTL Compiler (`llvm/circt`) produces something like:
```verilog
module B(input clock, reset);
  reg  [1:0] count;
  wire [1:0] _T;
  assign _T = count;
  wire [2:0] _T_2 = {1'h0, _T} + 3'h1;
  wire [1:0] _T_6 = ~{2{reset}} & _T_2[1:0];
  always_ff @(posedge clock)
    count <= _T_6;
endmodule
```

Both Yosys and Formality said these were equivalent.

As far as I can tell, the problem is that the `_T_6 = ~{2{reset}} & _T_2[1:0]` and the `posedge clock` blocks can execute in any order. One order is "correct" and one order is "incorrect". By moving the reset before the clock, `top.cpp` is then invariant to changes in however the backend decides to generate Verilog.

### Contributor Checklist

- [x] Did you add ~Scaladoc~ _C++ inline comments_ to every public function/method?
- [n/a] Did you add at least one test demonstrating the PR?

#### Type of Improvement

- bug fix                        

#### API Impact

None.

None.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/resources/chisel3/top.cpp
+++ b/src/main/resources/chisel3/top.cpp
@@ -47,8 +47,15 @@ int main(int argc, char** argv) {
   cout << "Starting simulation!\n";
 
   while (!Verilated::gotFinish() && main_time < timeout) {
-    if (main_time > 10) {
-      top->reset = 0;   // Deassert reset
+    // Deassert reset on timestep 10.  This needs to occur before the clock
+    // asserts on timestep 11 because there is a single call to top->eval() in
+    // this loop.  Verilator evaluates sequential logic (always blocks) before
+    // combinational logic during top->eval().  Staggering the reset update is
+    // necessary to produce the same simulation behavior independent of whether
+    // or not the generated Verilog puts synchronous reset logic inside or
+    // outside its associated always block.
+    if (main_time == 10) {
+      top->reset = 0;
     }
     if ((main_time % 10) == 1) {
       top->clock = 1;       // Toggle clock
@@ -92,4 +99,3 @@ int main(int argc, char** argv) {
     if (tfp) tfp->close();
 #endif
 }
-


aswaterman (approved):

albert-magyar:
This isn't a Verilog issue per se (hence the formal equivalence), but is actually consistent with how Verilator `eval` is supposed to work. TLDR: there must be a call to `eval()` between any modifications to inputs that rely on continuous assignments and any setting of clock inputs to high. The typical testbench solution is to set inputs out-of-phase with the clock edges (as you've done here).


> When eval() is called Verilator looks for changes in clock signals and evaluates related sequential always blocks, such as computing always_ff @ (posedge...) outputs. Then Verilator evaluates combinatorial logic.
>
>Note combinatorial logic is not computed before sequential always blocks are computed (for speed reasons). Therefore it is best to set any non-clock inputs up with a separate eval() call before changing clocks.
>
>Alternatively, if all always_ff statements use only the posedge of clocks, or all inputs go directly to always_ff statements, as is typical, then you can change non-clock inputs on the negative edge of the input clock, which will be faster as there will be fewer eval() calls.

src/main/resources/chisel3/top.cpp
```diff
@@ -47,8 +47,11 @@ int main(int argc, char** argv) {
   cout << "Starting simulation!\n";
 
   while (!Verilated::gotFinish() && main_time < timeout) {
-    if (main_time > 10) {
-      top->reset = 0;   // Deassert reset
+    // Deassert reset on and after cycle 10.  This needs to occur before the
```
albert-magyar, line 50:
```suggestion
    // Deassert reset on and after timestep 10.  This needs to occur before the
```

src/main/resources/chisel3/top.cpp
```diff
@@ -47,8 +47,11 @@ int main(int argc, char** argv) {
...
+    // clock asserts, on cycle 11, to avoid a race condition if a synchronous
```
albert-magyar, line 51:
```suggestion
    // clock asserts, on timestep 11, to avoid a race condition if a synchronous
```

src/main/resources/chisel3/top.cpp
seldridge: Fixed in the rebase.

src/main/resources/chisel3/top.cpp
seldridge: Fixed in the rebase.

seldridge:
Amazing @albert-magyar! I was tearing my hair out with this...

I've updated the `top.cpp` comment to indicate that this is because Verilator works the way it does.

PR #1826, opened 2021-03-19
Make plugin autoclonetype always on

jackkoenig:
Follow on to https://github.com/chipsalliance/chisel3/pull/1804, make the compiler plugin autoclonetype be always on and thus used in chisel3 itself.

### Contributor Checklist

#### Type of Improvement

- performance improvement 
 - new feature/API  

#### API Impact

As described in the release notes in https://github.com/chipsalliance/chisel3/pull/1804 and reflected in the test change, this will cause issues for any Bundle that inherits from a Bundle in `chisel3.util` if the code is not also compiled using the compiler plugin. Thus, this change can only occur on the major version bump.

#### Desired Merge Strategy

- Squash

#### Release Notes

Always generate `cloneType` methods for Bundles using the compiler plugin. This causes a binary and source incompatible change for any Chisel code that does not use the compiler plugin but extends a `Bundle` in other code that does. All Chisel users should use the compiler plugin.

- [ ] Did you review?

### DIFF
--- a/plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
+++ b/plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
@@ -5,8 +5,9 @@ package chisel3.internal.plugin
 import scala.tools.nsc
 import nsc.Global
 import nsc.plugins.{Plugin, PluginComponent}
+import scala.reflect.internal.util.NoPosition
 
-private[plugin] case class ChiselPluginArguments(var useBundlePlugin: Boolean = false) {
+private[plugin] case class ChiselPluginArguments(var useBundlePlugin: Boolean = true) {
   def useBundlePluginOpt = "useBundlePlugin"
   def useBundlePluginFullOpt = s"-P:chiselplugin:$useBundlePluginOpt"
 }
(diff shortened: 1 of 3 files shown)


plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -24,7 +25,8 @@ class ChiselPlugin(val global: Global) extends Plugin {
   override def init(options: List[String], error: String => Unit): Boolean = {
     for (option <- options) {
       if (option == arguments.useBundlePluginOpt) {
-        arguments.useBundlePlugin = true
+        val msg = s"'${arguments.useBundlePluginFullOpt}' is now default behavior, you can stop using the scalacOption."
```
azidar, line 28: Can you write the actual scalacOption here in the message?

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -265,19 +265,18 @@ class AutoClonetypeSpec extends ChiselFlatSpec with Utils {
...
-      val io = IO(new InheritingBundle)
-      io.deq <> io.enq
-      io.count := 0.U
-      io.error := true.B
-    }
-    elaborate(new MyModule)
-  }
-
   // New tests from the plugin
   if (usingPlugin) {
```
azidar, line 269: Do we need this anymore?

azidar (changes requested):

plugin/src/main/scala-2.12/chisel3/internal/plugin/ChiselPlugin.scala
jackkoenig: That's what the string interpolator is doing!

src/test/scala/chiselTests/AutoClonetypeSpec.scala
jackkoenig: Yes, the `noPluginTests` compile and run this same file https://github.com/chipsalliance/chisel3/blob/03ec2156988ddef35272365fb86415511eafb4e2/build.sbt#L206

We should consider making the plugin mandatory but that's not the purview of this PR

azidar (approved):

PR #1834, opened 2021-03-24
Introduce VecLiterals

chick:
Add support for Vec literals
Similar to Bundle literals, this PR extends the API with two methods of creating Vec literals.

Inferred:
```
Vec.Lit(0x1.U, 0x2.U)
```
or explicit:
```
Vec(2, UInt(4.W)).Lit(0 -> 0x1.U, 1 -> 0x2.U)
```
Explicit form allows for partial, or sparse, literals, and provides a mechanism for creating a literal from an existing
Vec literals can be used as Register initializers

Inferred form will set the element width to largest element width
Error checks in include: elements specified twice, width larger than
explicit type

Extensive test suite.

Features
- Can be defined outside of module (enables use with chiseltest library)
- Can be sparsely specified (with the effect that top level `litOption` returns None)
- litOption behaves like to UInt in bit ordering
- works as `Reg` initializer
- Decent error exceptions for
  - duplicate keys are found
  - wrong data types employed
  - width of individual element exceeds the basic `Vec` width
  - Arbitrary nesting (consistent with type constraints ins allowed)
 
Implementation is reasonably DRY except in the main body of _makeLit between
Bundle and Vec literals, mostly to avoid risks of breaking existing Bundle Lit behavior
This can be dealt with in following release

### Contributor Checklist

#### Type of Improvement

New Feature/API

#### API Impact
Extends API with
```
Vec.Lit(0x1.U, 0x2.U)
Vec(2, UInt(4.W)).Lit(0 -> 0x1.U, 1 -> 0x2.U)
```

This should not impact generated verilog

#### Desired Merge Strategy

Squash: The PR will be squashed and merged

#### Release Notes
Introduces Vec literals which are similar to Bundle Literals. 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -2,17 +2,18 @@
 
 package chisel3
 
+import chisel3.experimental.VecLiterals.AddVecLiteralConstructor
+
 import scala.collection.immutable.ListMap
 import scala.collection.mutable.{HashSet, LinkedHashMap}
 import scala.language.experimental.macros
-
-import chisel3.experimental.BaseModule
-import chisel3.experimental.BundleLiteralException
-import chisel3.experimental.EnumType
+import chisel3.experimental.{BaseModule, BundleLiteralException, ChiselEnum, EnumType, VecLiteralException}
 import chisel3.internal._
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo._
+
+import scala.collection.mutable
 
 class AliasedAggregateFieldException(message: String) extends ChiselException(message)
 
@@ -51,16 +52,19 @@ sealed abstract class Aggregate extends Data {
     */
   override def litOption: Option[BigInt] = {
     // Shift the accumulated value by our width and add in our component, masked by our width.
-    def shiftAdd(accumulator: Option[BigInt], elt: Data): Option[BigInt] = (accumulator, elt.litOption()) match {
-      case (Some(accumulator), Some(eltLit)) =>
-        val width = elt.width.get
-        val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
-        Some((accumulator << width) + masked)
-      case (None, _) => None
-      case (_, None) => None
+    def shiftAdd(accumulator: Option[BigInt], elt: Data): Option[BigInt] = {
+      (accumulator, elt.litOption()) match {
+        case (Some(accumulator), Some(eltLit)) =>
+          val width = elt.width.get
+          val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
+          Some((accumulator << width) + masked)
+        case (None, _) => None
+        case (_, None) => None
+      }
     }
+
     topBindingOpt match {
-      case Some(BundleLitBinding(_)) =>
+      case Some(BundleLitBinding(_)) | Some(VecLitBinding(_))=>
         getElements
           .reverse
           .foldLeft[Option[BigInt]](Some(BigInt(0)))(shiftAdd)
@@ -73,6 +77,7 @@ sealed abstract class Aggregate extends Data {
   def getElements: Seq[Data]
 
   private[chisel3] def width: Width = getElements.map(_.width).foldLeft(0.W)(_ + _)
+
   private[chisel3] def legacyConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
     // If the source is a DontCare, generate a DefInvalid for the sink,
     //  otherwise, issue a Connect.
@@ -83,11 +88,50 @@ sealed abstract class Aggregate extends Data {
     }
   }
 
+  // Returns pairs of all fields, element-level and containers, in a Record and their path names
+  def getRecursiveFields(data: Data, path: String): Seq[(Data, String)] = data match {
+    case data: Record =>
+      data.elements.map { case (fieldName, fieldData) =>
+        getRecursiveFields(fieldData, s"$path.$fieldName")
+      }.fold(Seq(data -> path)) {
+        _ ++ _
+      }
+    case data: Vec[_] =>
+      data.getElements.zipWithIndex.map { case (fieldData, fieldIndex) =>
+        getRecursiveFields(fieldData, path = s"$path($fieldIndex)")
+      }.fold(Seq(data -> path)) {
+        _ ++ _
+      }
+    case data => Seq(data -> path) // we don't support or recurse into other Aggregate types here
+  }
+
+
+  // Returns pairs of corresponding fields between two Records of the same type
+  def getMatchedFields(x: Data, y: Data): Seq[(Data, Data)] = (x, y) match {
+    case (x: Element, y: Element) =>
+      require(x typeEquivalent y)
+      Seq(x -> y)
+    case (x: Record, y: Record) =>
+      (x.elements zip y.elements).map { case ((xName, xElt), (yName, yElt)) =>
+        require(xName == yName) // assume fields returned in same, deterministic order
+        getMatchedFields(xElt, yElt)
+      }.fold(Seq(x -> y)) {
+        _ ++ _
+      }
+    case (x: Vec[_], y: Vec[_]) =>
+      (x.getElements zip y.getElements).map { case (xElt, yElt) =>
+        getMatchedFields(xElt, yElt)
+      }.fold(Seq(x -> y)) {
+        _ ++ _
+      }
+  }
+
   override def do_asUInt(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt = {
     SeqUtils.do_asUInt(flatten.map(_.asUInt()))
   }
+
   private[chisel3] override def connectFromBits(that: Bits)(implicit sourceInfo: SourceInfo,
-      compileOptions: CompileOptions): Unit = {
+                                                            compileOptions: CompileOptions): Unit = {
     var i = 0
     val bits = if (that.isLit) that else WireDefault(UInt(this.width), that) // handles width padding
     for (x <- flatten) {
@@ -155,9 +199,18 @@ trait VecFactory extends SourceInfoDoc {
   */
 sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
     extends Aggregate with VecLike[T] {
+
   override def toString: String = {
+    val bindingString = topBindingOpt match {
+      case Some(VecLitBinding(vecLitBinding)) =>
+        val contents = vecLitBinding.zipWithIndex.map { case ((data, lit), index) =>
+          s"$index=$lit"
+        }.mkString(", ")
+        s"($contents"
+      case _ => bindingToString
+    }
     val elementType = sample_element.cloneType
-    s"$elementType[$length]$bindingToString"
+    s"$elementType[$length]$bindingString"
   }
 
   private[chisel3] override def typeEquivalent(that: Data): Boolean = that match {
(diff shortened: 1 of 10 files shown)


mwachs5:
For the sparse case used as a Reg initializer, do the other registers get reset to 0? Or have no reset value?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+    * Vec(2, UInt(8.W)).Lit(
+    *   1 -> 0x0A.U,
+    *   2 -> 0x0B.U
+    * )
+    * }}}
+    */
+  private[chisel3] def _makeLit(elementInitializers: (Int, Data)*): this.type = {
+
+    def checkLiteralConstruction(): Unit = {
+      val dupKeys = elementInitializers.map { x => x._1 }.groupBy(x => x).flatMap { case (k, v) =>
```
ducky64, line 392: https://stackoverflow.com/questions/24729544/how-to-find-duplicates-in-a-list
might be a cleaner way to get duplicates (without the count though - and while nice, it might not be necessary)

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+          None
+        }
+      }
+      if (dupKeys.nonEmpty) {
+        throw new VecLiteralException(
+          s"VecLiteral: has duplicated indices ${dupKeys.map { case (k, n) => s"$k($n times)" }.mkString(",")}"
+        )
+      }
+
+      val outOfRangeIndices = elementInitializers.map(_._1).flatMap {
```
ducky64, line 405: Are you looking for filter instead?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+      }
+      if (outOfRangeIndices.nonEmpty) {
+        throw new VecLiteralException(
+          s"VecLiteral: The following indices (${outOfRangeIndices.mkString(",")}) " +
+            s"are less than zero or greater or equal to than Vec length"
+        )
+      }
+
+      val badLits = elementInitializers.flatMap {
+        case (index, lit) if (!(gen.getClass == lit.getClass)) =>
```
ducky64, line 417: Are these checks consistent with other literal ops?
These checks also feel a bit ad-hoc, and maybe (especially the width-based check) should be left to firrtl which will have more information to do the check properly?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+      )
+
+      val valueBinding = value.topBindingOpt match {
+        case Some(litBinding: LitBinding) => litBinding
+        case _ => throw new VecLiteralException(s"field $fieldIndex specified with non-literal value $value")
+      }
+
+      field match { // Get the litArg(s) for this field
+        case bitField: Bits =>
+//          val adjustedBitField = if (bitField.isInstanceOf[Bool]) {
```
ducky64, line 464: dead code

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+    requireIsChiselType(this, "vec literal constructor model")
+    checkLiteralConstruction()
+
+    val clone = cloneType
+    val cloneFields = getRecursiveFields(clone, "(vec root)").toMap
+
+    // Create the Vec literal binding from litArgs of arguments
+    val vecLitLinkedMap = new mutable.LinkedHashMap[Data, LitArg]()
+    elementInitializers.sortBy { case (a, _) => a }.foreach { case (fieldIndex, value) =>
+      val field = clone.apply(fieldIndex)
```
ducky64, line 450: would it make sense to use cloneFields here? or getMatchedFields?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+//            bitField
+//          } else {
+//            bitField.cloneTypeWidth(field.getWidth.W)
+//          }
+//          val notSameClass = !field.typeEquivalent(adjustedBitField)
+
+//          val adjustedBitField = bitField
+
+          val notSameClass = !field.typeEquivalent(bitField)
+          if (notSameClass) { // TODO typeEquivalent is too strict because it checks width
```
ducky64, line 474: Do you just want to inline the typeEquivalent?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            throw new VecLiteralException(
+              s"VecLit: Literal specified at index $fieldIndex ($value) does not match Vec type $sample_element"
+            )
+          } else if (bitField.getWidth > field.getWidth) {
```
ducky64, line 478: Is this truly an else-if case, instead of being a separate if block?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            throw new VecLiteralException(
+              s"VecLit: Literal specified at index $fieldIndex ($value) is too wide for Vec type $sample_element"
+            )
+          }
+          val litArg = valueBinding match {
+            case ElementLitBinding(litArg) => litArg
```
ducky64, line 484: It also seems like we want Data.getElementBinding that encapsulates this logic (and does the BundleLitBinding / VecLitBinding looksup).

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            vecLitLinkedMap(remap(valueField)) = valueValue
+          }
+
+        case vecField: Vec[_] =>
+          if (!(vecField typeEquivalent value)) {
+            throw new VecLiteralException(s"field $fieldIndex $vecField specified with non-type-equivalent value $value")
+          }
+          // Copy the source VecLitBinding with vecFields (keys) remapped to the clone
+          val remap = getMatchedFields(value, vecField).toMap
+          value.topBinding.asInstanceOf[VecLitBinding].litMap.map { case (valueField, valueValue) =>
```
ducky64, line 510: inconsistent use (w/ record case) of `valueBinding` vs. `value.topBinding`

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            vecLitLinkedMap(remap(valueField)) = valueValue
+          }
+
+        case enumField: EnumType => {
+          if (!(enumField typeEquivalent value)) {
+            throw new VecLiteralException(s"field $fieldIndex $enumField specified with non-type-equivalent enum value $value")
+          }
+          val litArg = valueBinding match {
+            case ElementLitBinding(litArg) => litArg
```
ducky64, line 519: Does this not need to handle the Record / Vec lookup? (and if it does we should probably use / have getElementBinding or something).
Also, does this not give an non-exhaustive match warning?

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,32 @@ package object experimental {
...
+    */
+  object VecLiterals {
+    implicit class AddVecLiteralConstructor[T <: Vec[_]](x: T) {
+      /** Given a generator of a list tuples of the form [Int, Data]
+        * constructs a Vec literal, parallel concept to `BundleLiteral`
+        *
+        * @param elems generates a lit of `(Int, Data)` where the I
+        * @return
+        */
+      def Lit(elems: (Int, Data)*): T = {
```
ducky64, line 145: Does it make sense, for `Vec[T]`, to only allow types of `(Int, T)`?

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,32 @@ package object experimental {
...
+        x._makeLit(elems: _*)
+      }
+    }
+
+    implicit class AddObjectLiteralConstructor[T <: Vec.type](x: T) {
```
ducky64, line 150: Why is this in a different implicit class from the above?
Is it for Vec as an object? A comment here might be helpful, eg this infers based on type.

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,32 @@ package object experimental {
...
   }
 
+  /** This class provides the `Lit` method needed to define a `Vec` literal
+    */
+  object VecLiterals {
+    implicit class AddVecLiteralConstructor[T <: Vec[_]](x: T) {
+      /** Given a generator of a list tuples of the form [Int, Data]
+        * constructs a Vec literal, parallel concept to `BundleLiteral`
+        *
+        * @param elems generates a lit of `(Int, Data)` where the I
```
ducky64, line 142: I think the sentence was accidentally

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,32 @@ package object experimental {
...
+        * @return
+        */
+      def Lit(elems: (Int, Data)*): T = {
+        x._makeLit(elems: _*)
+      }
+    }
+
+    implicit class AddObjectLiteralConstructor[T <: Vec.type](x: T) {
+      def Lit[T <: Data](elems: T*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
+        require(elems.nonEmpty, s"Lit.Vec(...) must have at least one element")
```
ducky64, line 152: This seems like a reasonable require, but as people start writing generators with this they may well want a zero length literal. Might need semantics like `Vec[Nothing]` which may not be well supported. But it'll be a strict API expansion.

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,32 @@ package object experimental {
...
+        val indexElements = elems.zipWithIndex.map { case (element, index) => (index, element)}
+        val widestElement = elems.sortBy(_.getWidth).last
```
ducky64, line 154: does the max operator work here?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -318,6 +371,163 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
...
+            case BundleLitBinding(litMap) => litMap.getOrElse(value,
+              throw new BundleLiteralException(s"Field $fieldName specified with unspecified value")
+            )
+            case VecLitBinding(litMap) => litMap.getOrElse(value,
+              throw new VecLiteralException(s"Field $fieldIndex specified with unspecified value"))
+          }
+          val adjustedLitArg = litArg.cloneWithWidth(sample_element.width)
```
ducky64, line 491: Do we want to do this in the frontend, or should this be left to FIRRTL?

ducky64 (commented):
Lots of style comments, but overall functionally looks sane.

I do have questions about whether the width checks should be done in the frontend, or should be done in FIRRTL when more information is available. Similarly, whether cloneWithWidth is necessary, since apparently Record literals didn't have it (and consistency I'd argue is pretty important)?

chick:
@ducky64 Regarding up front checks: I believe that, in general, error checking should be done as early as possible, there is typically better information about the error when done early. Usually we defer errors until later when we cannot determine them early such as after width inference has happened.

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -90,6 +90,9 @@ trait ChiselRunners extends Assertions with BackendCompilationUtilities {
 /** Spec base class for BDD-style testers. */
 abstract class ChiselFlatSpec extends AnyFlatSpec with ChiselRunners with Matchers
 
+/** Spec base class for BDD-style testers. */
+abstract class ChiselFreeSpec extends AnyFreeSpec with ChiselRunners with Matchers
```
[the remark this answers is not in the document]
chick: Yes, this is used in VecLiteralSpec. I personally prefer FreeSpec semantics

docs/src/appendix/experimental-features.md
```diff
@@ -85,7 +85,85 @@ class Example3 extends RawModule {
...
+  out := Vec(2, UInt(4.W)).Lit(0 -> 1.U, 1 -> 2.U)
+}
+```
+
+```scala mdoc:verilog
+chisel3.stage.ChiselStage.emitVerilog(new VecExample1a)
+```
+
+The following examples all use the explicit form.
+With the explicit form partial specification is allowed, defaulting any unconnected fields to 0 (regardless of type).
```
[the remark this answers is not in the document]
chick: The ability to specify a subset of the `vec` elements is part of the spec and is in tests and doc.
Can you point out where this internal notion of _unspecified_ is? I do not see it mentioned in code, docs or tests.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -51,16 +52,19 @@ sealed abstract class Aggregate extends Data {
...
+          val masked = ((BigInt(1) << width) - 1) & eltLit  // also handles the negative case with two's complement
+          Some((accumulator << width) + masked)
+        case (None, _) => None
+        case (_, None) => None
+      }
     }
+
     topBindingOpt match {
-      case Some(BundleLitBinding(_)) =>
+      case Some(BundleLitBinding(_)) | Some(VecLitBinding(_))=>
```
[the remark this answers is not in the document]
chick: Good catch, fixed

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -83,11 +88,50 @@ sealed abstract class Aggregate extends Data {
     }
   }
 
+  // Returns pairs of all fields, element-level and containers, in a Record and their path names
+  def getRecursiveFields(data: Data, path: String): Seq[(Data, String)] = data match {
+    case data: Record =>
+      data.elements.map { case (fieldName, fieldData) =>
```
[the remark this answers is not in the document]
chick: I'm not sure. The pre-existing Bundle code used fold so it seemed less risky to follow the existing pattern.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -83,11 +88,50 @@ sealed abstract class Aggregate extends Data {
...
+      data.getElements.zipWithIndex.map { case (fieldData, fieldIndex) =>
+        getRecursiveFields(fieldData, path = s"$path($fieldIndex)")
+      }.fold(Seq(data -> path)) {
+        _ ++ _
+      }
+    case data => Seq(data -> path) // we don't support or recurse into other Aggregate types here
+  }
+
+
+  // Returns pairs of corresponding fields between two Records of the same type
```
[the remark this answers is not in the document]
chick: It could be. Once again, it seemed like a good idea to follow the earlier naming convention, it as you say "code motion"

core/src/main/scala/chisel3/experimental/package.scala
chick: Yes, this for handling a different style of constructor that does not require providing each index.
I have added a comment to this effect

core/src/main/scala/chisel3/experimental/package.scala
chick: Good catch, deleted dangling string from comment

core/src/main/scala/chisel3/experimental/package.scala
chick: You are probably right that generators should cover a zero length. Do you have a suggestion as to what should happen in this case? Because this is all experimental, it seems like changing this small aspect of the API can be reasonably changed in the future.

core/src/main/scala/chisel3/Aggregate.scala
chick: I intended this behavior, happy to discuss it in a chisel dev meeting

core/src/main/scala/chisel3/experimental/package.scala
chick: `max` does not work here, but I have re-written as a fold, that avoids the unnecessary `sort`.

jackkoenig:
Need to make sure we check the reset behavior with partial literals, ie.
They should be supported, and any elements without a reset in the literal should not be reset in the Verilog, eg. BundleLiterals examle: https://scastie.scala-lang.org/x3mEHCxxRRermejswgemVA

chick:
@jackkoenig I think the test "Vec literals should only init specified fields when used to partially initialize a reg of vec"
in VecLiteralSpec does this. Can you take a look and confirm?

jackkoenig:
> I think the test "Vec literals should only init specified fields when used to partially initialize a reg of vec"
> in VecLiteralSpec does this. Can you take a look and confirm?

Yeah that test looks good!

core/src/main/scala/chisel3/experimental/package.scala
jackkoenig: Note that we currently don't allow empty `VecInit()`, I see this as related. We *should* support both but I think having this consistent is reasonable and when we figure out how to add support we should add to both.

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,35 @@ package object experimental {
...
+      def Lit(elems: (Int, Data)*): T = {
+        x._makeLit(elems: _*)
+      }
+    }
+
+    implicit class AddObjectLiteralConstructor[T <: Vec.type](x: T) {
+      /** This provides an literal construction method for cases using
+        * object `Vec` as in `Vec.Lit(1.U, 2.U)`
+        */
+      def Lit[T <: Data](elems: T*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
```
jackkoenig, line 154:
```suggestion
    implicit class AddObjectLiteralConstructor(x: Vec.type) {
      /** This provides an literal construction method for cases using
        * object `Vec` as in `Vec.Lit(1.U, 2.U)`
        */
      def Lit[T <: Data](elems: T*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
```
No type parameter needs for `Vec.type`, it's an object with no type parameters.

Why do we have the implicits here but not on the version of the implicit for class `Vec`?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -83,11 +88,50 @@ sealed abstract class Aggregate extends Data {
     }
   }
 
+  // Returns pairs of all fields, element-level and containers, in a Record and their path names
+  def getRecursiveFields(data: Data, path: String): Seq[(Data, String)] = data match {
```
jackkoenig, line 92: Why are we adding 2 new public methods here? I'm guessing these should be `private[chisel3]`?

docs/src/appendix/experimental-features.md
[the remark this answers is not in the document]
ducky64: The internal notion is an artifact of the data structures - where the entry for the element in the binding doesn't exist (as opposed to all entries being initialized to zero on creation, even if the user didn't specify those).

core/src/main/scala/chisel3/experimental/package.scala
ducky64: Is it accurate to describe the `elems` argument as "generates a lot" as opposed to something more like the values of the lit?

ducky64 (commented):
Closed some fixed conversation items, but there appears to be a bunch that still need addressing?

For early checks: yes, early checking is nice but potentially can cause inconsistency since it's (to some degree) reimplementing what FIRRTL is doing. I don't have the right answer, but it's something to watch out for. Especially if the checks look more ad-hoc sprinkled rather than a highly consistent pattern.

jackkoenig:
> For early checks: yes, early checking is nice but potentially can cause inconsistency since it's (to some degree) reimplementing what FIRRTL is doing. I don't have the right answer, but it's something to watch out for. Especially if the checks look more ad-hoc sprinkled rather than a highly consistent pattern.

There are some things, at least, that FIRRTL can't check. For example, multiple literal values for the same element of the Vec literal is just last connect as far as FIRRTL is concerned. 

That aside, I think a better software architecture of Chisel would construct the underlying FIRRTL nodes as it goes (and thus be able to leverage error checking there), but that's not the current architecture. I think it is much preferable to be able to provide a stack trace to the user's source code immediately rather than having to wait for all of elaboration to complete and then FIRRTL to check for errors.

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,35 @@ package object experimental {
...
+    */
+  object VecLiterals {
+    implicit class AddVecLiteralConstructor[T <: Vec[_]](x: T) {
+      /** Given a generator of a list tuples of the form [Int, Data]
+        * constructs a Vec literal, parallel concept to `BundleLiteral`
+        *
+        * @param elems generates a lit of `(Int, Data)`
+        * @return
+        */
+      def Lit(elems: (Int, Data)*): T = {
```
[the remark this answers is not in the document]
chick: Changes made

core/src/main/scala/chisel3/Aggregate.scala
chick: Made them private

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -130,6 +132,35 @@ package object experimental {
...
+      def Lit(elems: (Int, T)*): Vec[T] = {
+        x._makeLit(elems: _*)
+      }
+    }
+
+    implicit class AddObjectLiteralConstructor(x: Vec.type) {
+      /** This provides an literal construction method for cases using
+        * object `Vec` as in `Vec.Lit(1.U, 2.U)`
+        */
+      def Lit[T <: Data](elems: T*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
```
jackkoenig, line 154:
```suggestion
      def Lit[T <: Data](elems: T*): Vec[T] = {
```

I think we should remove these implicits for now. We need a `SourceInfo` macro to handle varargs so that one can do a bit extraction immediately after the `Lit` method. However, given that we're lacking the implicits from BundleLiterals as well, I think all of these `Lit` methods should be consistent.

chick: Made this change.
@ducky64 @jackkoenig let me know if anything else needs to be done

core/src/main/scala/chisel3/Aggregate.scala
chick: This was done, used maxBy

core/src/main/scala/chisel3/Aggregate.scala
chick: Removed

core/src/main/scala/chisel3/Aggregate.scala
chick: I would rather keep this as is. It follows the same pattern as Bundle/Record literals.

core/src/main/scala/chisel3/Aggregate.scala
chick: made separate if block

core/src/main/scala/chisel3/Aggregate.scala
chick: @ducky64 I am not sure what `Data.getElementBinding` is. I can't find anywhere in project

core/src/main/scala/chisel3/experimental/package.scala
chick: done

core/src/main/scala/chisel3/Aggregate.scala
chick: Fixed non-exhaustive match, I don't think it is needed for Record / Vec lookup. I have tests for that

core/src/main/scala/chisel3/experimental/package.scala
chick: done

src/test/scala/chiselTests/BundleLiteralSpec.scala
```diff
@@ -76,6 +75,24 @@ class BundleLiteralSpec extends ChiselFlatSpec with Utils {
     } }
   }
 
+  "bundle literals of bundle literals" should "work" in {
```
jackkoenig, line 78:
```suggestion
  "bundle literals of vec literals" should "work" in {
```

jackkoenig:
I'm happy with this as is, would like to get it out there for people to try it on `3.5-SNAPSHOT`.

jackkoenig (approved):
LGTM! Excellent work!

core/src/main/scala/chisel3/Aggregate.scala
ducky64: It doesn't exist yet, the question was more does it make sense to create a new binding helper function that encapsulates this pattern.

PR #1854, opened 2021-04-04
add helper function to convert chirrtl to firrtl.

sequencer:
We purge Driver too radically, thus lost some API, this PR add API to convert chirrtl to firrtl.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
add api chisel3.stage.convert(chirrtl: cir.Circuit): fir.Circuit

none

#### Desired Merge Strategy

#### Release Notes
add api chisel3.stage.convert(chirrtl: cir.Circuit): fir.Circuit to convert cirrtl to firrtl

- [ ] Did you review?


sequencer:
I'm tuning firesim, doc/test will be added later.

src/main/scala/chisel3/stage/ChiselStage.scala
```diff
@@ -172,6 +172,23 @@ object ChiselStage {
       .get
   }
 
+  def convert(chirrtl: cir.Circuit): fir.Circuit = {
```
edwardcwang, line 175: Scaladoc?

jackkoenig: Yes ScalaDoc that also discourages using this

jackkoenig (commented):
I'm okay with adding this back since `cir.Circuit` is public. If we were to deprecate that and then make that private then this can be deprecated and removed at the same time.

sequencer:
need @jackkoenig  review to this PR :)

jackkoenig (approved):

davidbiancolin:
This is probably a stupid question, but why doesn't emitFirrtl return a tuple of (String, Annotations), where _2 => elaboration-generated annotations? Most non-trivial chisel will generate some annotations right? 

 Is there a blessed way of doing this, instead of calling elaborate and then emit as we used to do with the old driver
 (here: https://github.com/firesim/firesim/blob/master/sim/midas/src/main/scala/midas/passes/SimulationMapping.scala#L96)? 
 
I'd be happy to PR a method, say `emitFirrtlAndElaboratedAnnotations`, to ChiselStage that would return a tuple if that makes sense.

sequencer:
Oh, sorry for the delay response, I totally missed this.

> but why doesn't emitFirrtl return a tuple of (String, Annotations), where _2 => elaboration-generated annotations? Most non-trivial chisel will generate some annotations right?

Sounds really reasonable, but I'm thinking if the use case here is conflict to `NoRunFirrtlCompilerAnnotation`?
This maybe will introduce more codes to firesim.
I'll file a PR later.

PR #1857, opened 2021-04-07
Add "how to tie off to 0" to the Cookbook

mwachs5:
### Contributor Checklist

#### Type of Improvement

 - documentation

#### API Impact

No impact

No impact

#### Desired Merge Strategy

#### Release Notes

Added "tieoff to 0" example to Cookbooks

- [ ] Did you review?

### DIFF
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -66,6 +66,31 @@ class Foo extends RawModule {
 }
 ```
 
+### How can I tieoff a Bundle/Vec to 0?
+
+You can use `asTypeOf` as above, but also use `chiselTypeOf` rather than having to
+figure out the type of the LHS:
+
+```scala mdoc:silent:reset
+import chisel3._
+
+class MyBundle extends Bundle {
+  val foo = UInt(4.W)
+  val bar = Vec(4, UInt(1.W))
+}
+
+class Foo(gen: () => Data) extends RawModule {
+  val bundle = IO(Output(gen))
+  bundle := 0.U.asTypeOf(chiselTypeOf(gen))
+}
+
+class Bar extends RawModule {
+  val foo = Module(new Foo(new MyBundle()))
+}
+```
+
+0.U.asTypeOf(chiselTypeOf(a.bits))
+
 ### How do I create a Vec of Bools from a UInt?
 
 Use [`VecInit`](https://www.chisel-lang.org/api/latest/chisel3/VecInit$.html) given a `Seq[Bool]` generated using the [`asBools`](https://www.chisel-lang.org/api/latest/chisel3/UInt.html#asBools():Seq[chisel3.Bool]) method.
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -79,7 +79,7 @@ class MyBundle extends Bundle {
   val bar = Vec(4, UInt(1.W))
 }
 
-class Foo(gen: () => Data) extends RawModule {
+class Foo(gen: Data) extends RawModule {
   val bundle = IO(Output(gen))
   bundle := 0.U.asTypeOf(chiselTypeOf(gen))
 }
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -12,6 +12,7 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
 * Converting Chisel Types to/from UInt
   * [How do I create a UInt from an instance of a Bundle?](#how-do-i-create-a-uint-from-an-instance-of-a-bundle)
   * [How do I create a Bundle from a UInt?](#how-do-i-create-a-bundle-from-a-uint)
+  * [How can I tieoff a Bundle/Vec to 0?](#how-can-i-tieoff-a-bundle-vec-to-0)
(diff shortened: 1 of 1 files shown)


docs/src/cookbooks/cookbook.md
```diff
@@ -66,6 +66,31 @@ class Foo extends RawModule {
...
+  bundle := 0.U.asTypeOf(chiselTypeOf(gen))
+}
+
+class Bar extends RawModule {
+  val foo = Module(new Foo(new MyBundle()))
+}
+```
+
+0.U.asTypeOf(chiselTypeOf(a.bits))
+
```
mwachs5, line 93:
```suggestion
```

docs/src/cookbooks/cookbook.md
```diff
@@ -66,6 +66,28 @@ class Foo extends RawModule {
 }
 ```
 
+### How can I tieoff a Bundle/Vec to 0?
```
jackkoenig, line 69: Can you add this to the Table of Contents as well?

docs/src/cookbooks/cookbook.md
```diff
@@ -66,6 +66,28 @@ class Foo extends RawModule {
...
+
+```scala mdoc:silent:reset
+import chisel3._
+
+class MyBundle extends Bundle {
+  val foo = UInt(4.W)
+  val bar = Vec(4, UInt(1.W))
+}
+
+class Foo(gen: () => Data) extends RawModule {
```
jackkoenig, line 82:
```suggestion
class Foo(gen: Data) extends RawModule {
```

ekiwi: We probably shouldn't change this here, but is there a good reason why we name parameters like this `gen`?
Something like `outputType: Data` seems much more accurate.

mwachs5: I guess because I always think of it as the "generator", since before you had to do something like:

```IO(Output(new FooBundle())```

I think of anything as a `new` as a generator.
Is this `Chisel._` thinking, not `chisel3` thinking?

Comment (ekiwi) [reply docs/src/cookbooks/cookbook.md:82]: I think of it as the result of doing `chiselTypeOf(...)` or as a directly defined type `UInt(8.w)`

File: docs/src/cookbooks/cookbook.md [hunk 4]
```diff
@@ -66,6 +66,28 @@ class Foo extends RawModule {
...
+  val bundle = IO(Output(gen))
+  bundle := 0.U.asTypeOf(chiselTypeOf(gen))
```
jackkoenig, line 84:
```suggestion
  bundle := 0.U.asTypeOf(gen)
```

docs/src/cookbooks/cookbook.md
jackkoenig: I think `gen` came from older style where it would usually be a 0-arity "generator method", eg.
```scala
class MyBundle[T <: Data](gen: () => T) extends Bundle {
  val foo = gen()
  val bar = gen()
}
```
It still probably makes sense in those contexts, but I agree that in cases like this something like `outputType` makes more sense.

docs/src/cookbooks/cookbook.md
mwachs5: if this is what i should write this whole example is pointless, the point was to show needing to call `chiselTypeOf`

jackkoenig: Haha, right, sorry, yeah that makes sense.

mwachs5: well, if this works without the `chiselTypeOf` then that's probably not a great cookbook... maybe what I mean is:

```suggestion
  bundle := 0.U.asTypeOf(chiselTypeOf(bundle))
```

jackkoenig: Yes I think that's better

mwachs5: I think what I'll do is show both with some comments -- bundle is a hardware type, while gen is already a chisel type

docs/src/cookbooks/cookbook.md
```diff
@@ -12,6 +12,7 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
 * Converting Chisel Types to/from UInt
   * [How do I create a UInt from an instance of a Bundle?](#how-do-i-create-a-uint-from-an-instance-of-a-bundle)
   * [How do I create a Bundle from a UInt?](#how-do-i-create-a-bundle-from-a-uint)
+  * [How can I tieoff a Bundle/Vec to 0?](#how-can-i-tieoff-a-bundle-vec-to-0)
```
jackkoenig, line 15:
```suggestion
  * [How can I tieoff a Bundle/Vec to 0?](#how-can-i-tieoff-a-bundlevec-to-0)
```
I checked this and this change is needed to make it work.

jackkoenig (approved):

sequencer:
I'm thinking is it reasonable to add a `tieoff` helper function for all `Data`?

mwachs5:
I'd say no since the value to tie to is context dependent

sequencer:
For example:
https://github.com/chipsalliance/rocket-chip/pull/2811/files
compatibility chisel always depends on an implicitly tieoff(tieoff `valid` and leave `bits` alone), I think it's kind of a common use case?

PR #1867, opened 2021-04-18
implement equal to BitPat.

sequencer:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

add equal to BitPat to check if equality of two BitPat.

None

#### Desired Merge Strategy

#### Release Notes
implement equal to BitPat.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -108,6 +108,13 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
   def === (that: UInt): Bool = macro SourceInfoTransform.thatArg
   def =/= (that: UInt): Bool = macro SourceInfoTransform.thatArg
 
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case y: BitPat => value == y.value && mask == y.mask && getWidth == y.getWidth
+      case _ => false
+    }
+  }
+
   /** @group SourceInfoTransformMacro */
   def do_=== (that: UInt)
       (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = {


ducky64 (approved):
lgtm

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -108,6 +108,13 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
   def === (that: UInt): Bool = macro SourceInfoTransform.thatArg
   def =/= (that: UInt): Bool = macro SourceInfoTransform.thatArg
 
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case y: BitPat => value == y.value && mask == y.mask && getWidth == y.getWidth
+      case _ => false
```
ekiwi, line 114: I feel like this should maybe throw an exception. What I am afraid of is the common error of using `==` instead of `===` remaining undetected.

ekiwi (changes requested):

src/main/scala/chisel3/util/BitPat.scala
ducky64: I vaguely recall discussing the same for Data types, but for some reason there were issues. Perhaps it may be worth revisiting?

I think it makes sense to keep BitPat consistent with Data though.

sequencer: > I think it makes sense to keep BitPat consistent with Data though.

I think `BitPat` is kind of different from `Data`, since it's a literal only. So I think implement `equal` is safe(comparing if two `BitPat` have a same literal).

> I feel like this should maybe throw an exception.

I think so, I'll throw a exception here.

jackkoenig: I feel like I remember seeing that you're not supposed to every error in `equals`, sort of a JVM-contract kind of thing, will poke around.

sequencer: Adding a runtime warning and return false here sound reasonable?

ekiwi: > I feel like I remember seeing that you're not supposed to every error in `equals`, sort of a JVM-contract kind of thing, will poke around.

If that is the case, then I guess we can remove it. Maybe add a check to the linter that you and Adam are working on?

jackkoenig: > Adding a runtime warning and return false here sound reasonable?

I don't think a runtime warning is great either. The problem is that `equals` is called by all kinds of things you may not expect (eg. Maps, Sets).

Universal equality is a huge wart in Scala that they are fixing (opt-in) in Scala 3 (see http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html). I'm not sure that potentially violating JVM standards is a great idea here.

> Maybe add a check to the linter that you and Adam are working on?

Linting seems reasonable to me here. Scala sometimes warns when types can never be equal, I wonder if we can hook into that?

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -108,6 +108,13 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
   def === (that: UInt): Bool = macro SourceInfoTransform.thatArg
   def =/= (that: UInt): Bool = macro SourceInfoTransform.thatArg
 
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case y: BitPat => value == y.value && mask == y.mask && getWidth == y.getWidth
+      case _ => throw new RuntimeException("BitPat can only compare to another BitPat.")
```
jackkoenig, line 114: This should just return false, can consider using the runtime deprecated mechanism to print a warning at the end of Chisel elab.

src/main/scala/chisel3/util/BitPat.scala
sequencer: > Linting seems reasonable to me here. Scala sometimes warns when types can never be equal, I wonder if we can hook into that?

I investigated this morning, but didn't found any method to do so(even compiler plugin), I'd like to make it happen in the future PR.

sequencer:
@jackkoenig Should we get this in? or blocked by:
> using the runtime deprecated mechanism to print a warning at the end of Chisel elab.
Here are my options:
1. merge as it is, implement "runtime warning mechanism" later, add warning back.
2. blocked by `runtime warning mechanism`. 
3. use `logger.LazyLogging` from FIRRTL, warn this at runtime.

jackkoenig (approved):

ekiwi (approved):
OK if Jack thinks it is ok.

Issue #1466, opened 2020-06-07
Encountered an exception


mwachs5:
**Type of issue**: bug report 

**Impact**: unknown

**Development Phase**: request

**Other information**
 

Encountering this error during chisel->firrtl elaboration, would hope to get a more useful error message. I am sure I am doing something wrong but the error message isn't helping me diagnose. Will continue to dig into the problem and update the ticket if I can point what the user error was that led to this case:

```
[error] (run-main-0) java.util.NoSuchElementException: None.get
[error] java.util.NoSuchElementException: None.get
[error] 	at scala.None$.get(Option.scala:529)
[error] 	at scala.None$.get(Option.scala:527)
[error] 	at chisel3.internal.HasId.getRef(Builder.scala:119)
[error] ocketchiat chisel3.internal.HasId.getRef$(Builder.scala:119)
[error] 	at chisel3.Data.getRef(Data.scala:276)
[error] 	at chisel3.internal.firrtl.Converter$.convert(Converter.scala:46)
[error] 	at chisel3.internal.firrtl.Converter$.convert(Converter.scala:50)
[error] 	at chisel3.internal.firrtl.Converter$.convertSimpleCommand(Converter.scala:115)
[error] 	at chisel3.internal.firrtl.Converter$.rec$1(Converter.scala:163)
[error] 	at chisel3.internal.firrtl.Converter$.convert(Converter.scala:201)
[error] 	at chisel3.internal.firrtl.Converter$.convert(Converter.scala:266)
[error] 	at chisel3.internal.firrtl.Converter$.$anonfun$convert$5(Converter.scala:273)
[error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
[error] 	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
[error] Nonzero exit code: 1
[error] (Compile / runMain) Nonzero exit code: 1
[error] Total time: 25 s, completed Jun 7, 2020 1:48:54 PM
```

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

clone the `firrtl-error` branch of rocket-chip: https://github.com/chipsalliance/rocket-chip/tree/firrtl-error

```
<usual rocket-chip setup stuff according to its README>
cd vsim
make verilog
```

**What is the current behavior?**

I get a None.get exception

**What is the expected behavior?**

Something detects that the None.get will fail and gives a more useful error message to help identify the user error.

[...]

PR #1878, opened 2021-04-23
Add some error context to Converter .getRefs

jackkoenig:
Add source locator context to any `.getRefs` in the `Converter` so that when users do unfortunately hit an internal bug, they at least know which line to look at so they can do _something_.

Related to #1466 

This is sort of impossible to test because if we do hit this, it's a bug that needs to be fixed. I happen to be working on some changes where I did hit this so I used my local changes for developing this.

I wanted to make the `SourceInfo` argument implicit since we're just threading it through and it's a bit of a pain, but because we have a materializer macro for `SourceInfo`, that's no good (if we forget to have one in scope it'll just point into the Converter.scala file itself lol).

### Contributor Checklist

#### Type of Improvement

  - new error reporting

#### API Impact

No impact, will fail strict binary compatibility checking but all changes are in a package private object so can be waived.

No impact

#### Desired Merge Strategy

   - Squash

#### Release Notes

Add more context to internal errors during Converter

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -3,9 +3,9 @@
 package chisel3.internal.firrtl
 import chisel3._
 import chisel3.experimental._
-import chisel3.internal.sourceinfo.{NoSourceInfo, SourceLine, SourceInfo}
+import chisel3.internal.sourceinfo.{NoSourceInfo, SourceInfo, SourceLine, UnlocatableSourceInfo}
 import firrtl.{ir => fir}
-import chisel3.internal.{castToInt, throwException}
+import chisel3.internal.{HasId, castToInt, throwException}
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Queue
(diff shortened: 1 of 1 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -24,6 +24,12 @@ private[chisel3] object Converter {
     case Percent => ("%%", List.empty)
   }
 
+  def getRef(id: HasId, sourceInfo: SourceInfo): Arg =
+    id.getOptionRef.getOrElse(
+      throwException(s"Internal error! Could not get ref for $id!" +
+        sourceInfo.makeMessage(" See " + _))
+    )
```
seldridge, line 31: Does this error message include some stuff about this being an actual bug and to report it? If not, should it?

seldridge (approved):
lgtm

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
jackkoenig: Yeah it probably should, I'll make the error message more verbose.

jackkoenig:
@seldridge, I've updated the error message, now you get stuff like:
```
Exception in thread "main" chisel3.internal.ChiselException: Internal error! Could not get ref for 'BundleB' @[Jack.scala 35:7]! 'BundleB' was defined in module 'MyModule@15c25153'. This is a bug in Chisel, please file an issue at 'https://github.com/chipsalliance/chisel3/issues/new'.
```

Note that source locator and module hints should both just be omitted if that information isn't available.

PR #1903, opened 2021-05-07
Fix ShiftRegister with 0 delay. 

sequencer:
if ShiftRegisters is empty, scala will complain:
```
  java.util.NoSuchElementException
    scala.collection.LinearSeqOptimized.last(LinearSeqOptimized.scala:150)
```
This PR:
1. fix this issue and return `in` directly when ShiftRegister size is 0.
1. Add test to check ShiftRegister(s) with delay is 0. 

### Contributor Checklist

#### Type of Improvement
- bug fix

#### API Impact
None

No

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -42,7 +42,7 @@ object ShiftRegister
     * val regDelayTwo = ShiftRegister(nextVal, 2, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).last
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).lastOption.getOrElse(in)
 
   /** Returns the n-cycle delayed version of the input signal with reset initialization.
     *
@@ -55,7 +55,7 @@ object ShiftRegister
     * val regDelayTwoReset = ShiftRegister(nextVal, 2, 0.U, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool): T = ShiftRegisters(in, n, resetData, en).last
+  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool): T = ShiftRegisters(in, n, resetData, en).lastOption.getOrElse(in)
 }
 
 
--- a/src/test/scala/chiselTests/Reg.scala
+++ b/src/test/scala/chiselTests/Reg.scala
@@ -7,6 +7,7 @@ import chisel3.util._
 import chisel3.experimental.DataMirror
 import chisel3.stage.ChiselStage
 import chisel3.testers.BasicTester
+import org.scalacheck.Gen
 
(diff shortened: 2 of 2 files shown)


src/test/scala/chiselTests/Reg.scala
```diff
@@ -55,18 +56,18 @@ class ShiftResetTester(n: Int) extends BasicTester {
   val start = 23.U
   val sr = ShiftRegister(cntVal + 23.U, n, 1.U, true.B)
   when(done) {
-    assert(sr === 1.U)
+    assert(sr === (if(n == 0) cntVal + 23.U else 1.U))
```
sequencer, line 59: This checks `n == 0` since no register will be reset.

src/test/scala/chiselTests/Reg.scala
```diff
@@ -55,18 +56,18 @@ class ShiftResetTester(n: Int) extends BasicTester {
...
     stop()
   }
 }
 
 class ShiftRegisterSpec extends ChiselPropSpec {
   property("ShiftRegister should shift") {
-    forAll(smallPosInts) { (shift: Int) => assertTesterPasses{ new ShiftTester(shift) } }
+    forAll(Gen.choose(0, 4)) { (shift: Int) => assertTesterPasses{ new ShiftTester(shift) } }
```
sequencer, line 66: `smallPosInts` only contains `1 to 4` which didn't check 0 case.
However in RocketChip, it used this API:
https://github.com/chipsalliance/rocket-chip/blob/86a2f2cca699f149bcc082ef2828654a0a4e3f4b/src/main/scala/devices/tilelink/Plic.scala#L188
So I change this test here.

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -42,7 +42,7 @@ object ShiftRegister
     * val regDelayTwo = ShiftRegister(nextVal, 2, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).last
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).lastOption.getOrElse(in)
```
sequencer, line 45: `n = 0` -> `ShiftRegisters` returns an empty `Seq`, `ShiftRegister` will directly return the input wire.

albert-magyar:
It looked like the original `ShiftRegisters.apply` API returned `Seq(in)` for `n=0`. Is there a reason to not just leave it consistent with that original base case and return the full sequence with `in` at index zero?

albert-magyar:
> > It looked like the original ShiftRegisters.apply API returned Seq(in) for n=0. Is there a reason to not just leave it consistent with that original base case and return the full sequence with in at index zero?
> 
> I thought returning a sequence of `Regs` is more consistent with API name?
> but I can also change to your suggested implementation.

 Yeah I guess that makes more sense than the original APU that returned `Seq(in)`. Having them all be registers does feel like it could be more convenient.

albert-magyar (approved):
LGTM

PR #1904, opened 2021-05-08
add os-lib to dependency.

sequencer:
Filesystem interaction has been a flaw in Scala for a long time. We have been using native Java IO or NIO library to do these jobs, which was neither elegant nor fast.
[os-lib](https://github.com/com-lihaoyi/os-lib) gives us a better solution:
For example: 
```scala
os.write(Path(wd / "somePath"),"someString")
```
can easily write to a file.
however with "native" Java API(I can never remember it and Google it every time when I need to use it.):
```scala
val pw = new PrintWriter(new File("somePath" ))
pw.write("someString")
pw.close
```
and there is another version in `FileWriter`, I don't know the difference, they are almost same to each other. 

another example is path API:
```scala
val wd = pwd / "folder"
```
vs
```scala
val wd = new (new File(".").getCanonicalPath(), "folder") 
```

os-lib has a very elegant **TYPE-SAFE** Path library and an easy way to work with files. 

Here is a [blog](https://www.lihaoyi.com/post/HowtoworkwithFilesinScala.html) from @lihaoyi, and it introduce this library, I think we can adopt it for the better filesystem API.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
None

None

#### Desired Merge Strategy

#### Release Notes
add os-lib to dependency.

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -72,6 +72,7 @@ lazy val chiselSettings = Seq (
   libraryDependencies ++= Seq(
     "org.scalatest" %% "scalatest" % "3.1.2" % "test",
     "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % "test",
+    "com.lihaoyi" %% "os-lib" % "0.7.7",
   ),
 ) ++ (
   // Tests from other projects may still run concurrently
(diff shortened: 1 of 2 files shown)


sequencer:
#1910 use this API now

ekiwi:
Seems like we added this by accepting https://github.com/chipsalliance/chisel3/pull/1910

PR #1911, opened 2021-05-12
Implement QMC decoder

sequencer:
This PR depends on #1904
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/experimental/decode/DecodeTableAnnotation.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/DecodeTableAnnotation.scala
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.experimental.decode
+
+import firrtl.annotations.{Annotation, ReferenceTarget, SingleTargetAnnotation}
+
+/** DecodeTableAnnotation used to store a decode result for a specific [[TruthTable]].
+  * This is useful for saving large [[TruthTable]] during a elaboration time.
+  *
+  * @note user should manage the correctness of [[minimizedTable]].
+  *
+  * @param target output wire of a decoder.
+  * @param truthTable input [[TruthTable]] encoded in a serialized [[TruthTable]].
+  * @param minimizedTable minimized [[truthTable]] encoded in a serialized [[TruthTable]].
+  */
+case class DecodeTableAnnotation(
+  target:         ReferenceTarget,
+  truthTable:     String,
+  minimizedTable: String)
+    extends SingleTargetAnnotation[ReferenceTarget] {
+  override def duplicate(n: ReferenceTarget): Annotation = this.copy(target = n)
+}
--- a/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
@@ -194,6 +194,7 @@ object QMCMinimizer extends Minimizer {
     if (minterms.nonEmpty) {
       // cover(i): nonessential prime implicants that covers `minterms(i)`
       val cover = minterms.map(m => implicants.filter(_.covers(m)))
+      // all subsets of `cover`, NP algorithm, O(2 ^ len(cover))
       val all = cover.tail.foldLeft(cover.head.map(Set(_)))((c0, c1) => c0.flatMap(a => c1.map(a + _)))
       all.map(_.toList).reduceLeft((a, b) => if (cheaper(a, b)) a else b)
     } else
@@ -247,6 +248,7 @@ object QMCMinimizer extends Minimizer {
         )
       )
 
+      // O(n ^ 3)
       for (i <- 0 to n) {
         for (j <- 0 until n - i) {
           mergeTable(i)(j).foreach(a => mergeTable(i + 1)(j) ++= mergeTable(i)(j + 1).filter(_ similar a).map(_ merge a))
@@ -273,6 +275,7 @@ object QMCMinimizer extends Minimizer {
 
       val primeImplicants = mergeTable.flatten.flatten.filter(_.isPrime).sortWith(_ < _)
 
+      // O(len(primeImplicants) ^ 4)
       val (essentialPrimeImplicants, nonessentialPrimeImplicants, uncoveredImplicants) =
         getEssentialPrimeImplicants(primeImplicants, implicants)
 
--- a/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
@@ -220,7 +220,7 @@ object QMCMinimizer extends Minimizer {
     val m = outputs.head.getWidth
 
     // for all outputs
-    table.copy(table = (0 until m).flatMap(i => {
+    val minimized = (0 until m).flatMap(i => {
       val outputBp = BitPat("b" + "?" * (m - i - 1) + "1" + "?" * i)
 
       // Minterms, implicants that makes the output to be 1
@@ -277,6 +277,19 @@ object QMCMinimizer extends Minimizer {
         getEssentialPrimeImplicants(primeImplicants, implicants)
 
       (essentialPrimeImplicants ++ getCover(nonessentialPrimeImplicants, uncoveredImplicants)).map(a => (a.bp, outputBp))
-    }).toMap)
+    })
+
+    minimized.tail.foldLeft(table.copy(table = Map(minimized.head))) { case (tb, t) =>
+      if (tb.table.exists(x => x._1 == t._1)) {
+        tb.copy(table = tb.table.map { case (k, v) =>
+          if (k == t._1) {
+            def ones(bitPat: BitPat) = bitPat.toString.drop(7).dropRight(1).zipWithIndex.collect{case ('1', x) => x}
+            (k, BitPat("b" + (0 until v.getWidth).map(i => if ((ones(v) ++ ones(t._2)).contains(i)) "1" else "?").mkString))
+          } else (k, v)
+        })
+      } else {
+        tb.copy(table = tb.table + t)
+      }
+    }
   }
 }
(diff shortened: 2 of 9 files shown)


src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
```diff
@@ -0,0 +1,285 @@
...
+      }
+
+      val primeImplicants = mergeTable.flatten.flatten.filter(_.isPrime).sortWith(_ < _)
+
+      // O(len(primeImplicants) ^ 4)
+      val (essentialPrimeImplicants, nonessentialPrimeImplicants, uncoveredImplicants) =
+        getEssentialPrimeImplicants(primeImplicants, implicants)
+
+      (essentialPrimeImplicants ++ getCover(nonessentialPrimeImplicants, uncoveredImplicants)).map(a => (a.bp, outputBp))
+    }).toMap)
```
yqszxx, line 283: toMap can potentially cause minimized term to be dropped when same term appears in more than one output product terms.

yqszxx (changes requested):
Need to fix returned truth table map first

src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
sequencer: Fixed in 888629cd

src/main/scala/chisel3/util/experimental/decode/decoder.scala
```diff
@@ -0,0 +1,53 @@
...
+                        |${truthTable.table}
+                        |""".stripMargin)
+        truthTable
+      }, {
+        val startTime = System.nanoTime()
+        val minimizedTable = minimizer.minimize(truthTable)
+        val totalTime = System.nanoTime() - startTime
+        val totalTimeInSeconds = totalTime / 1e9
+        val info = f"Logic Minimize with $minimizer finished in ${totalTimeInSeconds} second"
+        if (totalTimeInSeconds > 5)
```
aswaterman, line 28: I don't like the idea of having a built-in timeout, especially with a hard-coded limit.  I suggest removing this entirely, but if you insist on keeping it, make the limit a parameter that defaults to infinity.

aswaterman:
I'm essentially OK with this on technical grounds, provided we still agree we are going to replace it with invocation of Espresso.  I'd like @jackkoenig to perform the actual detailed code review, though.

src/main/scala/chisel3/util/experimental/decode/decoder.scala
sequencer: Got it!

src/main/scala/chisel3/util/experimental/decode/decoder.scala
```diff
@@ -0,0 +1,49 @@
...
+        truthTable
+      }, {
+        val startTime = System.nanoTime()
+        val minimizedTable = minimizer.minimize(truthTable)
+        val totalTime = System.nanoTime() - startTime
+        val totalTimeInSeconds = totalTime / 1e9
+        val info = f"Logic Minimize with $minimizer finished in $totalTimeInSeconds second"
+        logger.warn(info)
+        minimizedTable
+      }
```
sequencer, line 30: @aswaterman I directly user logger.warn here, my concern is if user use QMC to minimize their extra huge TruthTable, NP problem may hang entire elaboration. I personally wanna give a timeout here.
And another idea is we can use an async API to warn user every 10 seconds when they are stack at the `minimize` function. Do you think this is OK?

src/main/scala/chisel3/util/experimental/decode/decoder.scala
```diff
@@ -0,0 +1,60 @@
...
+          if(elapsed > timeout) logger.error(s"Minimizer has been executed for ${(System.nanoTime() - startTime) / 1e9} seconds.")
+        }
+        val minimizedTable = Await.result(future, Duration.Inf)
+        val totalTime = System.nanoTime() - startTime
+        val totalTimeInSeconds = totalTime / 1e9
+        val info = f"Logic Minimize with $minimizer finished in $totalTimeInSeconds second"
+        logger.warn(info)
+        minimizedTable
+      }
+    )
```
sequencer, line 42: @aswaterman 
I use a `timeout = 5` as default parameter to resolve this, if this minimize function is not finished every 5s, user will be warned. This parameter can be configured to higher if trying to minimize a huge decode table.
My point here is QMC is a NP-hard algorithm, to avoid user directly throw a giant table to espresso or QMC, this notice warns user they are doing a crazy thing.

aswaterman: @sequencer I am still not OK with this, for three reasons:
- Since Chisel is embedded in a general-purpose programming language, it is already possible for users to write arbitrary code, including their own infinite loops, NP-hard problems, etc.  It is inconsistent for this one case to emit a warning for long-running code, but not the other cases.
- There are already mechanisms available for debugging [seemingly] non-terminating `java` processes, like `kill -3`.
- This change makes the output non-deterministic, which is a problem for some flows.

Please remove the timeout.

sequencer: Sure!

src/main/scala/chisel3/util/experimental/decode/DecodeTableAnnotation.scala
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.experimental.decode
+
+import firrtl.annotations.{Annotation, ReferenceTarget, SingleTargetAnnotation}
+
+case class DecodeTableAnnotation(
```
jackkoenig, line 7: Can you add some comments describing what this is for?

jackkoenig (commented):
Overall this looks pretty good (although I didn't review the QMC implementation in detail, I'll have to trust that it's tested well, maybe a few more tests would be helpful).

Just a few comments/questions.

sequencer:
Really appreciate Jack’s review! Thanks for your time!

> although I didn't review the QMC implementation in detail, I'll have to trust that it's tested well, maybe a few more tests would be helpful

We(PLCT) did a long review to rocket QMC, I think I and @yqszxx have fully Understood  the internal of those codes, so even there exist some bugs, we are able to fix them.

As for the test, I think we can steal some test cases from espresso, however this may increase our test yield time to test some bigger cases. Is this necessary?

I’m in Beijing these days, so I might now be able to fix those requested changes asap. Sorry for that.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  def copy(table: Map[BitPat, BitPat] = this.table, default: BitPat = this.default) = new TruthTable(table, default)
+
+  override def equals(y: Any): Boolean = {
+    y match {
+      case y: TruthTable => toString == y.toString
+      case _ => false
+    }
+  }
```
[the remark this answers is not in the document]
sequencer: I remember I used SortedMap before , and failed in a 2.12 but successes in 2.13, I will retry this.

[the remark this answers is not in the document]
sequencer: I just take 2 hours to try this.
I found the main problem is we cannot compare two `SortedMap[BitPat, BitPat]`. since `BitPat` in `SortedMap` is a reference.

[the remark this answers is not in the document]
sequencer: So I personally prefer using String comparing here. I use sorted the serialized string, which provide a safer way to implement this.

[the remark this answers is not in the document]
jackkoenig: Yeah it's a bit awkward, but that derives from the referential equality of `BitPat` so I think this is the best solution.

src/test/scala/chiselTests/util/experimental/DecoderSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+        DecodeTableAnnotation(ReferenceTarget("", "", Nil, "", Nil),
+          """10->1
+            |01->1
+            |    0""".stripMargin,
+          """10->1
+            |    0""".stripMargin
+        )
+      )
+    )
+  }
```
jackkoenig, line 39: What about passing when given a correct `DecodeTableAnnotation`? They're for caching right or is that not yet implemented?

src/test/scala/chiselTests/util/experimental/minimizer/MinimizerSpec.scala
```diff
@@ -0,0 +1,256 @@
...
+        // BitPat("b000") -> BitPat("b0"),
+        BitPat("b001") -> BitPat("b?"),
+        BitPat("b010") -> BitPat("b?"),
+        // BitPat("b011") -> BitPat("b0"),
+        BitPat("b100") -> BitPat("b1"),
+        BitPat("b101") -> BitPat("b1"),
+        // BitPat("b110") -> BitPat("b0"),
+        BitPat("b111") -> BitPat("b1")
+      ),
+      BitPat("b0")
```
jackkoenig, line 47: What are the comments for? Is it indicating that when you leave a pattern off, it defaults to requiring the output to be 0?

sequencer: > What are the comments for? Is it indicating that when you leave a pattern off, it defaults to requiring the output to be 0?

it defaults to requiring the output to be 0?
Leave this to default to 0 and optimized by decode algorithm.

sequencer: I’ll add more comments here tonight.

src/test/scala/chiselTests/util/experimental/DecoderSpec.scala
sequencer: It should work, I forgot to add tests. I’ll add that tonight.

sequencer: Fixed in 7ec97426

jackkoenig (approved):

sequencer:
Thanks @jackkoenig ! Decoder finally landed!

yqszxx:
🎉Hooray~

PR #1922, opened 2021-05-19
Implicit driver

sequencer:
Inspired by #1921, adding `ImplicitDriver`(or some other better name).
I only use API from `object ChiselStage` currently, but I think maybe we can change the API from that object to make it be able to consume `args` and `annos`

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
Adding API `chisel3.util.experimental.ImplicitDriver`

None

#### Desired Merge Strategy

#### Release Notes
New Implicit Driver API:
```
import chisel3.util.experimental.ImplicitDriver
  object Main extends App {
    println((new Hello()).emitVerilog)
}
```

- [ ] Did you review?


ekiwi:
What happens if you try to call these methods inside a Chisel module though?

sequencer:
> What happens if you try to call these methods inside a Chisel module though?

I just wanna say this, I think it's safe since its in its own context. But we can also guard this to make sure we are not under a Builder context

seldridge:
> What happens if you try to call these methods inside a Chisel module though?

Just works: https://scastie.scala-lang.org/CSWanEZjT2ehEWgGJpKHew

Admittedly, it's weird, but I don't think there's a reason to try to prevent this. It's reasonable to call `ChiselStage` from inside elaboration.

ekiwi:
> Just works: https://scastie.scala-lang.org/CSWanEZjT2ehEWgGJpKHew

This is pretty amazing!

src/main/scala/chisel3/util/experimental/package.scala
```diff
@@ -0,0 +1,34 @@
+package chisel3.util
+
+import chisel3.Module
+
+package object experimental {
```
seldridge, line 5: I'm all for just dropping this in the `chisel3` package. This should be, by default available if you do `import chisel3._`.

sequencer:
I just did a Frankenstein example: see  e103932

sequencer:
Parallel Chisel!

src/main/scala/chisel3/util/experimental/package.scala
```diff
@@ -0,0 +1,34 @@
...
+      chisel3.stage.ChiselStage.emitSystemVerilog(module)
+    }
+
+    def emitFirrtl = {
+      chisel3.stage.ChiselStage.emitFirrtl(module)
+    }
+
+    def emitChirrtl = {
+      chisel3.stage.ChiselStage.emitChirrtl(module)
+    }
```
seldridge, line 32: We can bikeshed the names of these, but other ideas would be:

- `toVerilogString`
- `asVerilogString`

I would also suggest adding an entry-point for standard compilation:

```scala
  def compile(args: String *): AnnotationSeq = {
    (new ChiselStage).execute(args.toArray, Seq(ChiselGeneratorAnnotation(() => a)))
  }
```

Doing `(new Foo).compile("-X", "low")` is pretty slick.

seldridge (commented):
This is mind altering. Nice solution.

src/main/scala/chisel3/util/experimental/package.scala
sequencer: Got it:) Let's skip the experimental

src/main/scala/chisel3/util/experimental/package.scala
sequencer: Got it!

seldridge: Re: bikeshed... I actually don't care. I think `emit*` is good and matches what we've used for a long time. No reason to change.

src/main/scala/chisel3/util/experimental/package.scala
sequencer: Oh, that's another problem to drop it to chisel3 package:
`package object chisel3` is under `core` project, which doesn't depend on `chisel3` project:
So basically from current build architecture, `chisel3._` has no access to `stage`

src/main/scala/chisel3/util/experimental/package.scala
sequencer: How about API like this? I think we should also expose annotation interface to users:
```scala
    def execute(args: String*)(annos: AnnotationSeq = Nil): AnnotationSeq =
      (new ChiselStage)
        .execute(
          args.toArray,
          annos ++ Seq(new ChiselGeneratorAnnotation(() => module))
        )

    def compile(args: String*): AnnotationSeq = execute(args: _*)(Nil)
```
But which means executing is a little strange with 2 brackets.
```scala
    (new GCD).execute("xxx", "xxx")(Seq(SomeAnno0, SomeAnno1))
```

src/main/scala/chisel3/util/experimental/package.scala
sequencer: How about move it to `chisel3.stage` package? I think that's a good place for this API to stay.
I have no I idea how to add this API to chisel3 package.

src/main/scala/chisel3/util/experimental/package.scala
sequencer: > Re: bikeshed... I actually don't care. I think emit* is good and matches what we've used for a long time. No reason to change.

Personally I think `emit` sounds like a emission to file which has side effect, while `toXXX` has not. But I think both are OK if we need to align with our older APIs.

src/main/scala/chisel3/util/experimental/package.scala
schoeberl: Why is this splitting between ```core``` and the other project in the first place? This is why I stumbled over not being able to implement ```emitVerilog()```  just as a method on ```RawModule```. A method on ```RawModule``` would be easier to read than using this implicit magic ;-)

src/main/scala/chisel3/util/experimental/package.scala
schoeberl: I would like to keep ```emitVerilog()``` for generating the files. Then I would choose a different/better name for just returning it as a string. I think we need two different names anyway.

ekiwi: how about `generateVerilogString`?

src/main/scala/chisel3/util/experimental/package.scala
sequencer: Yes, I also think `RawModule` is much reasonable.
I also wonder why splitting chisel3 into different packages, I think this is the question is to @jackkoenig

src/main/scala/chisel3/util/experimental/package.scala
sequencer: maybe 
```
emitVerilog(path: os.Path = os.pwd): Unit
verilog(): String
```
sounds more simple?

schoeberl: Why not keep the parameters as they have been for ```emitVerilog```. Keeping the possibility of setting other options as well. I also would not like to force the user to use this (new) os library.

ekiwi: Personally, I like to have an active verb in the function name, since the function is going to perform a lot of work. `verilog` looks like a getter which imho should not be performing any substantial work.

schoeberl: I like ```generateVerilogString``` as it is clear what it is. Maybe a bit lengthy. If we copy Java/Scala convention we could shorten it to ```toVerilogString```.

sequencer: something maybe
```
generateVerilogString(): String
generateVerilogFile(directory: os.Path): os.Path
```
?

sequencer: I think @ekiwi is reasonable that we need a long word to indicate this process is expensive ;p
I personally prefer a API with a specific name(Thanks to IDEA code completion)

PR #1923, opened 2021-05-19
Reduce boilerplate for generating Bundle/Vec literals

kammoh:
### Contributor Checklist

#### Type of Improvement

 - new feature/API

#### API Impact

Extends `BundleLiterals` and `VecLiterals` `Lit` to work directly on hardware nodes (as well as chisel types), without requiring an explicit conversion to chisel type. IMO this results to a more readable and DRYer code, both in RTL and `chiseltest`.

None

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/package.scala
+++ b/core/src/main/scala/chisel3/experimental/package.scala
@@ -127,7 +127,7 @@ package object experimental {
   object BundleLiterals {
     implicit class AddBundleLiteralConstructor[T <: Record](x: T) {
       def Lit(elems: (T => (Data, Data))*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
-        x._makeLit(elems: _*)
+        (if(x.isSynthesizable) chiselTypeOf(x) else x)._makeLit(elems: _*)
       }
     }
   }
@@ -143,7 +143,7 @@ package object experimental {
         * @return
         */
       def Lit(elems: (Int, T)*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = {
-        x._makeLit(elems: _*)
+        (if(x.isSynthesizable) chiselTypeOf(x) else x)._makeLit(elems: _*)
       }
     }
 
--- a/src/test/scala/chiselTests/BundleLiteralSpec.scala
+++ b/src/test/scala/chiselTests/BundleLiteralSpec.scala
@@ -31,16 +31,16 @@ class BundleLiteralSpec extends ChiselFlatSpec with Utils {
   "bundle literals" should "pack" in {
     assertTesterPasses{ new BasicTester {
       val bundleLit = (new MyBundle).Lit(_.a -> 42.U, _.b -> false.B, _.c -> MyEnum.sB)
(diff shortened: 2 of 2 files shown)


jackkoenig:
This sort of cuts against one of the core design principles of Chisel 3 (vs. Chisel 2), namely that there is a clear divide between "Chisel types" and "Hardware values". To bring back @azidar's favorite example:

What does `Reg(UInt(3))` mean [in Chisel 2]?
1. Register of type `UInt`, width of 3
2. Register whose next cycle's value is 3
3. Register whose initial value is 3

Trick question, answer is:
4. Register with no initial value and width of 2

Now, I agree that the proposal in this PR is much less ambiguous than Adam's example above. The problem I see with this proposal is consistency. If we only have the clear distinction in cases where it could be ambiguous, then we end up with a mishmash of when you need types vs. when you don't.

This being said, `chiselTypeOf(...)` is soooo verbose and it drives me crazy. I think we should at least shorten it to `typeOf`, and perhaps think about putting it as a method rather than a function, ie. `.typeOf` or `.getType`. This won't make things quite as concise as you're proposing here, but might reduce the verbosity pain a little bit.

kammoh:
Thanks for the feedback @jackbackrack!
I understand the distinction between "Chisel types" and "Hardware nodes" and agree that moving towards a clear divide in the chisel3 codebase was a great step forward. Also IMO chisel3 API improvements (`Reg(UInt(3.W))`, `RegInit(3.U)`) have eliminated much of the ambiguity compared to Chisel2.  
On the other hand, I'm not really convinced that _conversions_ from "hardware type" to "Chisel type" have to be always made explicitly by the hardware designer.
While a surprise conversion in the opposite direction (chisel -> hardware) should never ever be made implicitly, I think in circumstances where the perceived/expected/imagined type is non-hardware, an implied conversions from hardware to chisel type can be understood clearly and unambiguously both by the designer and the toolchain. I agree this conversion is not desirable in situations where a similar constructs operate on both hardware and chisel types (with different semantics and hardware translation) e.g. `Reg(2.U)`, which can easily be confused with `RegInit(2.U)` and should be disallowed by the API (as is it's currently done. Only wished these could be checked at compile time, maybe by the compiler plugin?).
In this case, where the generated output is a literal, I feel there's even less chance of confusion or ambiguity.

ducky64:
So the conversion is not made implicitly because converting a hardware object to a Chisel type discards data, and silently discarding user input can be a great recipe for unintuitive behavior (which was a contributing factor to why `Reg(UInt(3))` is so bad, it discarded the literal 3 portion, so what you see is not what you get).

While `.Lit` is much less risky (and has lower potential for confusion) than `Reg(...)`, I also share the concern about a special-case carve-out and consistency issues. Larger question here is the trade-off between conciseness and consistency - I think we've been bitten by inconsistency before, and I personally lean towards consistency, but better ergonomics around `chiselTypeOf` (eg, with `.getType`) might be a good middle ground here.

kammoh:
I value API consistency quite a lot and close this PR if favor of a more generic ergonomic replacement for `chiselTypeOf`!

kammoh:
Thanks for your feedback @ducky64 
I value API consistency quite a lot and close this PR if favor of a more generic ergonomic replacement for `chiselTypeOf`!

PR #1925, opened 2021-05-20
Get rid of Docker.

sequencer:
This patch changes CI from self-maintained docker to Arch Linux, which provides all binary dependency from its upstream repository.
This makes possible to automatically check our publish is compatible to upstream stable version, and ease the maintenance burden of chisel-tools docker image.
It also switch to the graalvm to speed up our test.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
None

None

#### Desired Merge Strategy
- Squash

#### Release Notes
Switch CI to Arch Linux

- [ ] Did you review?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,21 +16,19 @@ jobs:
     strategy:
       matrix:
         scala: [2.13.6, 2.12.13]
-    container:
-      image: ucbbar/chisel3-tools
-      options: --user github --entrypoint /bin/bash
-    env:
-      CONTAINER_HOME: /home/github
+    container: archlinux/archlinux:latest
 
     steps:
+      - name: Install dependencies
+        run: |
+          echo -e '[archlinuxcn]\nServer = https://repo.archlinuxcn.org/$arch' >> /etc/pacman.conf
+          pacman-key --init
+          pacman -Syu --noconfirm archlinuxcn-keyring
+          pacman -Syu --noconfirm jdk11-graalvm-bin
+          pacman -Syu --noconfirm git sbt mill verilator z3 make clang
+          archlinux-java set java-11-graalvm
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Setup Scala
-        uses: olafurpg/setup-scala@v10
-        with:
-          java-version: adopt@1.8
-      - name: Cache Scala
-        uses: coursier/cache-action@v5
       - name: Documentation (Scala 2.12 only)
         if: matrix.scala == '2.12.13'
         run: sbt ++${{ matrix.scala }} docs/mdoc
@@ -55,19 +53,14 @@ jobs:
   publish:
     needs: [all_tests_passed]
     runs-on: ubuntu-latest
+    container: archlinux/archlinux:latest
     if: github.event_name == 'push'
 
     steps:
+      - name: Install dependencies
+        run: pacman -Syu --noconfirm git sbt gnupg
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Setup Scala
-        uses: olafurpg/setup-scala@v10
-        with:
-          java-version: adopt@1.8
-      - name: Cache Scala
-        uses: coursier/cache-action@v5
-      - name: Setup GPG (for Publish)
-        uses: olafurpg/setup-gpg@v3
       - name: Publish
         run: sbt ci-release
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,12 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
+          echo -e '[archlinuxcn]\nServer = https://repo.archlinuxcn.org/$arch' >> /etc/pacman.conf
+          pacman-key --init
+          pacman -Syu --noconfirm archlinuxcn-keyring
+          pacman -Syu --noconfirm jdk11-graalvm-bin
           pacman -Syu --noconfirm git sbt mill verilator z3 make clang
+          archlinux-java set java-11-graalvm
       - name: Checkout
         uses: actions/checkout@v2
       - name: Documentation (Scala 2.12 only)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,12 +21,7 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          echo -e '[archlinuxcn]\nServer = https://repo.archlinuxcn.org/$arch' >> /etc/pacman.conf
-          pacman-key --init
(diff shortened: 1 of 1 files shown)


ekiwi:
I don't think we support Verilator `4.2xx` yet.

sequencer:
> I don't think we support Verilator `4.2xx` yet.

I think we have already supported it, and actually I'm using it, which works correctly with RocketChip.
My real intention is that we should always support up-to-date published stable version like Z3 and verilator.
If we only support a specific version, that sounds strange.

sequencer:
If sticking to a old version, we cannot get any performance improvement or bug-fixes.

jackkoenig (changes requested):
This seems to be quite a bit slower than the old CI, the test steps are taking > 30 min compared to 18-20 min in the existing CI: https://github.com/chipsalliance/chisel3/runs/2650158099

sequencer:
Yes, I was trying to use graalvm to speed up, but seems to have a back action :(

sequencer:
WTF, why so slow??? even JDK 15 is slow too, I double I should blame llvm now.

sequencer:
I think we should switch to Z3 in the future, verilator is too slow for CI :(

edwardcwang:
How does the arch docker image work stability-wise? I think it would be better to use a stable release like Ubuntu for the docker base instead, if the arch image is going to constantly be autoupdating.

jackkoenig:
Perhaps we should stick with Ubuntu but just get rid of the nested Docker which means we need to install a few dependencies:
* mill - https://github.com/jodersky/setup-mill
* Z3 - Can wget from releases: https://github.com/Z3Prover/z3/releases or via action https://github.com/pavpanchekha/setup-z3
* Verilator - a little trickier, could revive https://github.com/drom/action-verilator or run it directly from a Docker container https://hub.docker.com/r/verilator/verilator/tags

sequencer:
@jackkoenig I agree, I think basically my main target is removing the chisel-tools docker.
Issue verilator/verilator#2980 submitted, hope they can accept a PR for GitHub Action release to let us wget from an official website. 
@ekiwi and @edwardcwang are right, maybe this is not a good flow to run CI.
But why so slow??? Do I need to blame docker?

ekiwi:
>     * Verilator - a little trickier, could revive

If we can use Ubuntu 20.04 LTS then the Verilator that is packaged with that should be OK to use. Its what we use by default in the `chiseltest` CI.

sequencer:
> > ```
> > * Verilator - a little trickier, could revive
> > ```
> 
> If we can use Ubuntu 20.04 LTS then the Verilator that is packaged with that should be OK to use. Its what we use by default in the `chiseltest` CI.

I think we should not rely anything on a specific distro, since it's always too old. This is the reason why I use Arch Linux and Debian, which always ships upstream stable release. I think we should be able to choose version on our own, and always bump version to upstream stable or some version we think they are stable.

ekiwi:
> I think we should not rely anything on a specific distro, since it's always too old. This is the reason why I use Arch Linux and Debian, which always ships upstream stable release. I think we should be able to choose version on our own, and always bump version to upstream stable or some version we think they are stable. Rather than Ubuntu developer think they are stable, actually they just steal packages from Debian, and pretends they are releasing "stable" release.

I understand your frustration and personally I use Fedora which is tracking upstream more closely. However, we do need to at least support Ubuntu 20.04 since that is the most recent LTS release of the most popular distro. Just like we cannot just remove Scala `2.12` support we need to slowly move to more recent Verilator versions.

Is there a specific feature that you need from Verilator that isn't available in `v4.028`?

sequencer:
> Is there a specific feature that you need from Verilator that isn't available in v4.028?

Not really. I just use software from the distro repository.
  
How about making some tradeoff? I admit most of users are using Ubuntu and CentOS, that's a pain issue we have to face. I think after we have GitHub Action based verilator, we can specific multiple releases on our own, and CI success depends on both success or Ubuntu version success. But we need to know if Chisel is broken in the new stable Verilator.

I think if this broken happens, we need to PR verilator to fix. We have no reason to only support old verilator since most of our users are sticking on a old version. We need to support a widely used version(Ubuntu) and upstream stable version(quick break and quick fix.) Or even(just a possibility) to give a verilator version matrix to make this more robustness.

ekiwi:
> I think after we have GitHub Action based verilator, we can specific multiple releases on our own, and CI success depends on both success or Ubuntu version success. But we need to know if Chisel is broken in the new stable Verilator.

I guess the question here is whether this test should be in `chisel` or in `chiseltest` (I am beginning to wonder if it was ever a good idea to split the two into separate repositories).
If you want to add it to chisel, you can have a look at what we do in `chiseltest` in order to test all supported Verilator versions (`v4.2xx` needs a PR to chiseltest to work with the API changes): https://github.com/ucb-bar/chisel-testers2/blob/de8704717513c54412bc09d08280e7306fd0d217/.github/workflows/test.yml#L53

sequencer:
> I guess the question here is whether this test should be in chisel or in chiseltest (I am beginning to wonder if it was ever a good idea to split the two into separate repositories).

Since last week, I start to think if is a good idea to port sanity tests in chisel to model checking, I think it can provide a far more better performance and coverage than verilator.

I still doubt if chiseltest is a good recommendation to users, since it sacrifices performance for join/fork API, I doubt its usability in large designs. 

> If you want to add it to chisel, you can have a look at what we do in chiseltest in order to test all supported Verilator versions (v4.2xx needs a PR to chiseltest to work with the API changes): https://github.com/ucb-bar/chisel-testers2/blob/de8704717513c54412bc09d08280e7306fd0d217/.github/workflows/test.yml#L53

I think that's a good idea, I think we can make it cleaner, serve as GitHub Action plugin which we can use with
```yaml
- name: Install verilator
  uses: xxx/setup-verilator
  with:
    version: 4.xxx
```

ekiwi:
> I still doubt if chiseltest is a good recommendation to users, since it sacrifices performance for join/fork API, I doubt its usability in large designs.

I am planning to separate out the simulator interface layer in chiseltest, so that it can be used independently from the frontend. We really should only have to maintain exactly one binding to Verilator that is tested to work on all supported versions. If I  succeed in this, then maybe the tests in chisel could depend on chiseltest and just reuse the more low-level verilator interface.

sequencer:
> I am planning to separate out the simulator interface layer in chiseltest, so that it can be used independently from the frontend. We really should only have to maintain exactly one binding to Verilator that is tested to work on all supported versions. If I succeed in this, then maybe the tests in chisel could depend on chiseltest and just reuse the more low-level verilator interface.

Haha That's what I'm doing now! I think we really need this.

PR #1943, opened 2021-05-27
Implement User API Module.finishInstantiate.

sequencer:
This PR adds `finishInstantiate`, which can be override by user and add code blocks at last of the Module.

Currently, A library-writer can add code blocks at the begin of Module by providing trait, and let user extend it. However it is impossible to append a code block to last of user codes.
There are some important usage, for example,
`DataMirror` all user IO, and annotate them.
Let user define a list of registers/wires, and annotate them automaticlly.
Auto connection, like what we did in diplomacy.

 
### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
add `Module.finishInstantiate`

None

#### Desired Merge Strategy

#### Release Notes
add `Module.finishInstantiate` to let user insert code block at the end of Module.

- [ ] Did you review?


jackkoenig (changes requested):
This is a substantial API expansion that I think we should be careful with. It's not clear if inheritance is the right approach because it doesn't compose. Say I have 2 APIs that each need to override `finishInstantiate`, then it depends on which order you mix in the traits for which implementation you get:
```scala
trait Foo {
  def func(): Unit = {}
}

trait Fizz extends Foo {
  override def func(): Unit = { 
    println("Howdy!")
  }
}
trait Buzz extends Foo {
  override def func(): Unit = { 
    println("Hi!")
  }
}

(new Foo with Fizz with Buzz {}).func() // Hi!
(new Foo with Buzz with Fizz {}).func() // Howdy!
```
https://scastie.scala-lang.org/oln3zJXmTJaTpTo3GvabEA

Say we do something different, rather than a single method to override, you can call something like `registerFinishAction(f: () => Unit)` that registers functions to call at the end. You _still_ run into ordering issues, what if you have an action that don't touches all ports but then another action runs after it that adds another port?

Perhaps we could constrain these "finish actions" to not be able to generate any hardware, only inject annotations. That sounds a lot like aspects to me 🙂. Maybe what we want here is a lighter-weight aspect API, I'm not sure.

I'd be curious what @azidar thinks.
I would also like to hear from potential users like @hcook.

mwachs5:
this sounds like LazyModule's `InModuleBody`... but agree it doesn't make sense at least as a one-time call. 

Why is this generally needed? There already exist several mechanisms as mentioned for this sort of thing (Aspects, diplomacy's LazyModule), what would a typical use case be?

sequencer:
> @mwachs5 this sounds like `LazyModule`'s `InModuleBody`.

For user-API this is `InModuleBody`, but for library writer, it is the [instantiate](https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/diplomacy/LazyModule.scala#L273-L329) function.

> @jackkoenig That sounds a lot like aspects to me. 
> Maybe what we want here is a lighter-weight aspect API, I'm not sure.
> @mwachs5 Why is this generally needed? There already exist several mechanisms as mentioned for this sort of thing (Aspects, diplomacy's LazyModule), what would a typical use case be?

I admit this API is something similar to `Aspect`. Here are some different points:
`Aspect` can only access public API in a module, while this API is more flexible, since user is defining this API in their own module, only put it at last block to instantiate.
This API is more light-weight, there are no additional passes.

while `diplomacy` is a pretty big library, I'd like it to be extracted from Rocket-Chip and server as a standalone library to users, but even it may becomes a standalone library in the future, it is too big for this demand(we only need to put a block in the end of entire Module), and I'd like diplomacy use this new future in the future, which can clean up its code bases.

> @jackkoenig Perhaps we could constrain these "finish actions" to not be able to generate any hardware, only inject annotations. 

But I still think we can construct hardware here, Diplomacy has the `instantiate` function, and it requires constructing hardware.

> @jackkoenig Say we do something different, rather than a single method to override, you can call something like `registerFinishAction(f: () => Unit)` that registers functions to call at the end. You still run into ordering issues, what if you have an action that don't touches all ports but then another action runs after it that adds another port?

That sounds better! but `abstract override` could also meet the sequence demand, here is a [blog](https://eed3si9n.com/constraining-class-linearization-in-Scala) to illustrate(currently I'm using this), but I admit this is some black magic in Scala, and we should not encourage users to do so. So in my next reversion, I will use your suggesting to rewrite this API.

sequencer:
@jackkoenig I think I still need this API:
For a use case that make an inner module being able to insert block to the outer module.
We can discuss this later in dev-meeting.

PR #1954, opened 2021-06-07
Stop Emitting BlackBoxResourceAnno

seldridge:
### Contributor Checklist

#### Type of Improvement

Other.

#### API Impact

Change Chisel to never emit `BlackBoxResourceAnno`. This improves the separation between Chisel and a FIRRTL compiler. The semantics of locating the file associated with a `BlackBoxResourceAnno` is that of Java Resources. This is fine if a FIRRTL compiler is implemented on top of the JVM. However, if the FIRRTL compiler is _not_ implemented on the JVM, then it makes the Chisel classpath part of what is necessary to pass to the FIRRTL compiler (and that a FIRRTL compiler needs to handle Java resource finding).

No.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/BlackBoxUtils.scala
+++ b/src/main/scala/chisel3/util/BlackBoxUtils.scala
@@ -6,7 +6,6 @@ import chisel3._
 import chisel3.experimental.{ChiselAnnotation, RunFirrtlTransform}
 import firrtl.transforms.{BlackBoxPathAnno, BlackBoxResourceAnno, BlackBoxInlineAnno, BlackBoxSourceHelper,
   BlackBoxNotFoundException}
-import java.io.{File, FileNotFoundException}
 
 trait HasBlackBoxResource extends BlackBox {
   self: BlackBox =>
--- a/src/main/scala/chisel3/util/BlackBoxUtils.scala
+++ b/src/main/scala/chisel3/util/BlackBoxUtils.scala
@@ -16,10 +16,12 @@ private[util] object BlackBoxHelpers {
     def fromResource(resourceName: String, moduleName: ModuleName) = try {
       val blackBoxFile = os.resource / os.RelPath(resourceName.dropWhile(_ == '/'))
       val contents = os.read(blackBoxFile)
-      if (contents.size > BigInt(2).pow(20))
-        logger.warn(s"Included black box resource $resourceName, which will be converted to an inline annotation, is greater than 1 MiB. This may affect compiler performance. Consider including this resource via a black box path.")
-
-
+      if (contents.size > BigInt(2).pow(20)) {
+        val message =
+          s"Black box resource $resourceName, which will be converted to an inline annotation, is greater than 1 MiB." +
+            "This may affect compiler performance. Consider including this resource via a black box path."
+        logger.warn(message)
+      }
       BlackBoxInlineAnno(moduleName, blackBoxFile.last, contents)
     } catch {
       case e: os.ResourceNotFoundException =>
--- a/src/main/scala/chisel3/util/BlackBoxUtils.scala
+++ b/src/main/scala/chisel3/util/BlackBoxUtils.scala
@@ -4,12 +4,31 @@ package chisel3.util
 
 import chisel3._
 import chisel3.experimental.{ChiselAnnotation, RunFirrtlTransform}
-import firrtl.transforms.{BlackBoxPathAnno, BlackBoxResourceAnno, BlackBoxInlineAnno, BlackBoxSourceHelper}
+import firrtl.transforms.{BlackBoxPathAnno, BlackBoxResourceAnno, BlackBoxInlineAnno, BlackBoxSourceHelper,
+  BlackBoxNotFoundException}
+import java.io.{File, FileNotFoundException}
+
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/util/BlackBoxUtils.scala
```diff
@@ -19,7 +38,11 @@ trait HasBlackBoxResource extends BlackBox {
     */
   def addResource(blackBoxResource: String): Unit = {
     val anno = new ChiselAnnotation with RunFirrtlTransform {
-      def toFirrtl = BlackBoxResourceAnno(self.toNamed, blackBoxResource)
+      def toFirrtl = {
+        ResourceHelpers.getResourceAsString(blackBoxResource) match {
+          case ResourceHelpers.ResourceResult(fileName, contents) => BlackBoxInlineAnno(self.toNamed, fileName, contents)
```
sequencer, line 43:
```
        val blackboxFile = os.resource / blackBoxResource
        BlackBoxInlineAnno(self.toNamed, blackboxFile.baseName, os.read(blackboxFile))
```
Have a try to type-safe way to read from file:)

seldridge: Switched to `os-lib`. :+1:

seldridge:
> is possible to replace `HasBlackBoxPath` as well?

Yes. I was aiming to keep this PR narrowly focused, though.

Fundamentally, only one of the three blackbox annotations are actually needed. They all encode the same information with differences in how the path is encoded and at what time.

> and a potential issue we might need to concern is: if there is a really huge blackbox verilog, something like 16 core Boom Verilog(maybe 1G+), will this become another performance issue to JVM? It will directly consume 1G from JVM, and case class has been copying this 1G in each transform.

Possibly. Using `BlackBoxPathAnno` may be a better way to encode this and could be a more sensible way to do what this PR does.

> Maybe another work around is that we can copy all blackboxes to targetDir at the end of elaboration(or reflect it to a chisel API, and do this during elaboration), finally set directly to the blackbox file to there.

This could probably be done with some `CustomFileEmission` where the annotation produces a side effect. This would then happen at the end of the first stage (either Chisel or FIRRTL depending how its being run).

sequencer:
> This could probably be done with some CustomFileEmission where the annotation produces a side effect. This would then happen at the end of the first stage (either Chisel or FIRRTL depending how its being run).

“happen at the end of the first stage” sounds reasonable! if FIRRTL doesn’t need any modification in the later passes, I think we can free it this memory as early as possible to save the effort of copying potential large text file.

seldridge:
Converting these to a path has the downside of not being able to represent the full space of URLs that resources can resolve to (e.g., if this is getting dug out of some JAR file which may be package in a library). If we run into a problem with this requiring too much memory, I can look into an alternative solution.

mwachs5 (commented):
is there anything in the website docs that should be updated for this change? https://www.chisel-lang.org/chisel3/docs/explanations/blackboxes.html#providing-implementations-for-blackboxes

seldridge:
@mwachs wrote:
> is there anything in the website docs that should be updated for this change?

No, I don't think so. 😄 What's cool about this is that the Chisel-level API doesn't change at all. Users can continue to use `hasResource` with no observable change except to the Annotation file that Chisel produces in a split Chisel/FIRRTL build.

jackkoenig:
I think this is okay to backport, it could affect some advanced users doing stuff in FIRRTL. That being said, since it is technically equivalent, it _shouldn't_ cause any problems.

seldridge:
@chick, @jackkoenig: I tested this locally using https://github.com/seldridge/chisel-resource-test. This was initially _not_ working. However, after cleaning my environment (again) it worked. This is pulling the JAR out of dsptools.

src/main/scala/chisel3/util/ExtModuleUtils.scala
```diff
@@ -18,7 +19,13 @@ trait HasExtModuleResource extends ExtModule {
     */
   def addResource(blackBoxResource: String): Unit = {
     val anno = new ChiselAnnotation with RunFirrtlTransform {
-      def toFirrtl = BlackBoxResourceAnno(self.toNamed, blackBoxResource)
+      def toFirrtl = try {
```
chick, line 22: Quibble:  Perhaps this code could be DRYed out a bit by merging the identical code between BlackBox and ExtModule into a trait

chick (approved):
This looks good to me. I like the tests. One quibble (which can be ignored regarding DRY code).
Only other consideration. Since we can anticipate that large files in a resource dir could cause problems whether
we could issue some warning based on size or copy time that would alert users to this situation. Just a thought

seldridge:
@chick. Can you take one more look? Updated with your changes. To DRY it out, I used a implicit class on the object to get a `BlackBoxInlineAnno.fromResource(...)` API which I thought would be the most readable if someone inspects the `addResource` code.

src/main/scala/chisel3/util/ExtModuleUtils.scala
seldridge: Done.

chick (approved):
LGTM

PR #1957, opened 2021-06-10
Add API for setting MultiClock in current Module scope

colin4124:
### Contributor Checklist

#### Type of Improvement

new feature/API

#### API Impact

Not affect existing API, it does much like `withClockAndReset` `withClock` `withReset`,  but set Clock/Reset in current Module scope.

Because `Module(new xxx_module)`, the `Module()` would set `Builder.currentClock` to `None`,  `setClock`  does not need to set `None` again.

Not Change.

#### Desired Merge Strategy

Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/MultiClock.scala
+++ b/core/src/main/scala/chisel3/MultiClock.scala
@@ -6,6 +6,41 @@ import chisel3.internal._
 
 import scala.language.experimental.macros
 
+object setClockAndReset {
+ /** Creates a new Clock and Reset scope in current Module
+    *
+    * @param clock the new implicit Clock
+    * @param reset the new implicit Reset
+    * @return Unit, affect current Module
+    */
+  def apply(clock: Clock, reset: Reset): Unit = {
+    setClock(clock)
+    setReset(reset)
+  }
+}
+
+object setClock {
+  /** Creates a new Clock scope in current Module
+    *
+    * @param clock the new implicit Clock
+    * @return Unit, affect current Module
+    */
+  def apply[T](clock: Clock): Unit =  {
+    Builder.currentClock = Some(clock)
+  }
+}
+
+object setReset {
+  /** Creates a new Reset scope in current Module
+    *
+    * @param reset the new implicit Reset
+    * @return Unit, affect current Module
+    */
+  def apply[T](reset: Reset): Unit = {
+    Builder.currentReset = Some(reset)
+  }
+}
+
 object withClockAndReset {
   /** Creates a new Clock and Reset scope
     *
--- a/src/test/scala/chiselTests/Clock.scala
+++ b/src/test/scala/chiselTests/Clock.scala
@@ -24,6 +24,16 @@ class WithClockAndNoReset extends RawModule {
   out := a
 }
 
+class SetClockAndNoReset extends RawModule {
+  val clock1 = IO(Input(Clock()))
+  val clock2 = IO(Input(Clock()))
+  val in = IO(Input(Bool()))
+  val out = IO(Output(Bool()))
+  setClock(clock2)
+  val a = RegNext(in)
+
+  out := a
+}
 
 class ClockSpec extends ChiselPropSpec {
   property("Bool.asClock.asUInt should pass a signal through unaltered") {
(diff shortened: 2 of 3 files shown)


sequencer:
Why not directly connect clock and reset from the outsider module?
Wondering the real use cases.

edwardcwang:
Also, missing unit tests

colin4124:
@sequencer 

```scala
class Top extends RawModule {
  val myclk = Input(Clock())
  val myrst = Input(Bool())

  withClockAndReset(myclk, myrst) {
    val a =  RegInit(0.U)
    val b =  RegInit(0.U)
    // My Moudle Logic etc.
    ...
  }
}

I don't want to all my module logic code indent in `withClockAndReset` scope. I prefer this:

```scala
class Top extends RawModule {
  val myclk = Input(Clock())
  val myrst = Input(Bool())

  setClockAndReset(myclk, myrst)

  val a =  RegInit(0.U)
  val b =  RegInit(0.U)

  // My Moudle Logic etc.
  ...
}
```

It is not a necessary， it just for  convenience and  may be an alternative.

edwardcwang:
If you want to write your synchronous logic without indents why not just stick them in a regular module and wrap instantiation in `withClockAndReset`? What I like about `withClockAndReset` is that the scope makes it very clear what is going on whereas a function call looks like it's manipulating magic state.

```scala
val a =  RegInit(0.U)
val b =  RegInit(0.U)
```

colin4124:
@edwardcwang I think so `withClockAndReset { ... }` is more clear, and `setClockAndReset` have side effect.

I just give a one more option to user, and this API is very convenient and doesn't make any bad effect, it just syntactic sugar.

Such as some one would use:

```scala
my_list foreach { println(_) } 
```

and other one prefer:

```scala
my_list foreach println
```

sequencer:
I personally think this has two edges:
the good one is, registers can be refered more easily, for example original with has this shortage:
```
val someWire = Wire(UInt(32.W))
withClock(clock2) {
  val a = RegInit(0.U(32.W)) // this reg cannot be refered other of withClock scope
  someWire := a // bind a to external wire, then we can read this wire.
  ... // other operations 
}
```
However, this API may introduce dangerous operations. For example, a very deep function calling this new API and forgot to release to original `clock` and `reset` may lead to a some potential problem, for example, sub-modules will be instantiated under this `clock` and `reset`, and you may not be able to find them. while original method provides a scope management for `clock` and `reset`, which helps user understanding which clock domain they are operating on.

colin4124:
@sequencer 

When you call `Module(new a_sub_module)`, it would set current clock/reset scope to None, so sub-module not use its parent's clock and reset implicitly. 

Use `setClockAndReset` in `RawModule` If a module have only one clock domain, it is called in front of the module logic, and then write your code as you use `extends Module`.

jackkoenig (changes requested):
I'm not okay with exposing mutation to the user like this. While the standard use-case of having a single clock and reset for a `RawModule` is reasonable, I don't think this is the way to accomplish it. We already have support for modules with a single clock and reset: `MultiIOModule` (becoming `Module` in Chisel 3.5).

I'm guessing the fundamental problem here is that you want to control the _name_ of the single clock and reset in a `MultiIOModule`, and I think we could accomplish this in a different way. For example, instead of making the implicit clock and reset in `MultiIOModule` special, we could add a trait for having the implicit clock and reset that provides virtual methods the user must implement, giving them control over the naming, something like:

```scala
trait ImplicitClockAndReset {
  def clock: Clock
  def reset: Reset
  // or
  def clockAndReset: (Clock, Reset)
}
class MyRawModule extends RawModule with ImplicitClockAndReset {
  val myclk = IO(Input(Clock()))
  val myrst = IO(Input(AsyncReset()))

  def clock = myclk
  def reset = myreset
  // Or
  def clockAndReset = (clock, reset)
  ...
}
```

Something like this should fulfill your desired functionality without the negative potential for users doing weird things like changing the clock and reset arbitrarily, or accidentally changing it in a function call that forgets to set it back.

jackkoenig:
I'll note that there are potential consequences of initialization order here, so this idea may not work (if those defs get called too early, the vals won't yet have been initialized). Another things we could do is just expose the ability to _name_ `clock` and `reset` but keep the vals as `clock` and `reset`.

colin4124:
@jackkoenig Yes, I  want to control the name of the single clock and reset.

Application scenarios: only one clock domain. Once it set, no need to set it back. `Module(..)` will take care of it.

Maybe abuse of setting clock but more flexible,  I understand what you concern.

PR #1968, opened 2021-06-21
create and extend annotatable BaseSim class for verification nodes

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This extends the user API. Users can now annotate `Assert`/`Assume`/`Cover` classes.

There should be no change.

#### Desired Merge Strategy

#### Release Notes

Creates the `BaseSim` abstract class for simulation-only components. Creates and allows for annotation of experimental verification `Assert`/`Assume`/`Cover` classes.

- [x] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/verification/package.scala
+++ b/core/src/main/scala/chisel3/experimental/verification/package.scala
@@ -12,15 +12,15 @@ package object verification {
   /** Base simulation-only component. */
   abstract class BaseSim extends NamedComponent
 
-  case class Assert(predicate: Bool, msg: String = "") extends BaseSim
-  case class Assume(predicate: Bool, msg: String = "") extends BaseSim
-  case class Cover(predicate: Bool, msg: String = "") extends BaseSim
+  final class Assert(predicate: Bool, msg: String = "") extends BaseSim
+  final class Assume(predicate: Bool, msg: String = "") extends BaseSim
+  final class Cover(predicate: Bool, msg: String = "") extends BaseSim
 
   object assert {
     def apply(predicate: Bool, msg: String = "")(
       implicit sourceInfo: SourceInfo,
       compileOptions: CompileOptions): Assert = {
-      val a = Assert(predicate, msg)
+      val a = new Assert(predicate, msg)
       when (!Module.reset.asBool) {
         val clock = Module.clock
         Builder.pushCommand(Verification(a, Formal.Assert, sourceInfo, clock.ref, predicate.ref, msg))
@@ -33,7 +33,7 @@ package object verification {
     def apply(predicate: Bool, msg: String = "")(
       implicit sourceInfo: SourceInfo,
       compileOptions: CompileOptions): Assume = {
-      val a = Assume(predicate, msg)
+      val a = new Assume(predicate, msg)
       when (!Module.reset.asBool) {
         val clock = Module.clock
         Builder.pushCommand(Verification(a, Formal.Assume, sourceInfo, clock.ref, predicate.ref, msg))
@@ -47,7 +47,7 @@ package object verification {
       implicit sourceInfo: SourceInfo,
       compileOptions: CompileOptions): Cover = {
       val clock = Module.clock
-      val c = Cover(predicate, msg)
+      val c = new Cover(predicate, msg)
       when (!Module.reset.asBool) {
         Builder.pushCommand(Verification(c, Formal.Cover, sourceInfo, clock.ref, predicate.ref, msg))
       }
--- a/core/src/main/scala/chisel3/experimental/verification/package.scala
+++ b/core/src/main/scala/chisel3/experimental/verification/package.scala
@@ -3,17 +3,19 @@
 package chisel3.experimental
 
 import chisel3._
-import chisel3.internal.{Builder, NamedComponent}
+import chisel3.internal.Builder
 import chisel3.internal.firrtl.{Formal, Verification}
 import chisel3.internal.sourceinfo.SourceInfo
 
 package object verification {
 
-  /** Base simulation-only component. */
-  abstract class BaseSim extends NamedComponent
-
+  /** Named class for assertions. */
   final class Assert(predicate: Bool, msg: String = "") extends BaseSim
+
+  /** Named class for assumes. */
   final class Assume(predicate: Bool, msg: String = "") extends BaseSim
+
+  /** Named class for covers. */
   final class Cover(predicate: Bool, msg: String = "") extends BaseSim
 
   object assert {
--- a/core/src/main/scala/chisel3/experimental/verification/package.scala
+++ b/core/src/main/scala/chisel3/experimental/verification/package.scala
@@ -3,41 +3,55 @@
 package chisel3.experimental
 
 import chisel3._
-import chisel3.internal.Builder
+import chisel3.internal.{Builder, NamedComponent}
 import chisel3.internal.firrtl.{Formal, Verification}
 import chisel3.internal.sourceinfo.SourceInfo
 
 package object verification {
+
+  /** Base simulation-only component. */
+  abstract class BaseSim extends NamedComponent
+
+  case class Assert(predicate: Bool, msg: String = "") extends BaseSim
+  case class Assume(predicate: Bool, msg: String = "") extends BaseSim
+  case class Cover(predicate: Bool, msg: String = "") extends BaseSim
+
   object assert {
     def apply(predicate: Bool, msg: String = "")(
       implicit sourceInfo: SourceInfo,
-      compileOptions: CompileOptions): Unit = {
+      compileOptions: CompileOptions): Assert = {
+      val a = Assert(predicate, msg)
       when (!Module.reset.asBool) {
         val clock = Module.clock
-        Builder.pushCommand(Verification(Formal.Assert, sourceInfo, clock.ref, predicate.ref, msg))
+        Builder.pushCommand(Verification(a, Formal.Assert, sourceInfo, clock.ref, predicate.ref, msg))
       }
+      a
     }
   }
 
   object assume {
     def apply(predicate: Bool, msg: String = "")(
       implicit sourceInfo: SourceInfo,
-      compileOptions: CompileOptions): Unit = {
+      compileOptions: CompileOptions): Assume = {
+      val a = Assume(predicate, msg)
       when (!Module.reset.asBool) {
         val clock = Module.clock
-        Builder.pushCommand(Verification(Formal.Assume, sourceInfo, clock.ref, predicate.ref, msg))
+        Builder.pushCommand(Verification(a, Formal.Assume, sourceInfo, clock.ref, predicate.ref, msg))
       }
+      a
     }
   }
 
   object cover {
     def apply(predicate: Bool, msg: String = "")(
       implicit sourceInfo: SourceInfo,
-      compileOptions: CompileOptions): Unit = {
+      compileOptions: CompileOptions): Cover = {
       val clock = Module.clock
+      val c = Cover(predicate, msg)
       when (!Module.reset.asBool) {
-        Builder.pushCommand(Verification(Formal.Cover, sourceInfo, clock.ref, predicate.ref, msg))
+        Builder.pushCommand(Verification(c, Formal.Cover, sourceInfo, clock.ref, predicate.ref, msg))
       }
+      c
     }
   }
 }
--- a/core/src/main/scala/chisel3/experimental/verification/package.scala
+++ b/core/src/main/scala/chisel3/experimental/verification/package.scala
@@ -10,19 +10,19 @@ import chisel3.internal.sourceinfo.SourceInfo
 package object verification {
 
   /** Named class for assertions. */
-  final class Assert(predicate: Bool, msg: String = "") extends BaseSim
+  final class Assert(val predicate: Bool) extends BaseSim
 
   /** Named class for assumes. */
-  final class Assume(predicate: Bool, msg: String = "") extends BaseSim
+  final class Assume(val predicate: Bool) extends BaseSim
 
   /** Named class for covers. */
-  final class Cover(predicate: Bool, msg: String = "") extends BaseSim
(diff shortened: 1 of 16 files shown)


debs-sifive:
CC @jackkoenig would you mind taking a peek?

sequencer:
Not related to this PR, I think not only verification APIs, even `printf` needs a `id` for being able to be annotated.

core/src/main/scala/chisel3/experimental/verification/package.scala
```diff
@@ -3,41 +3,55 @@
...
 import chisel3.internal.sourceinfo.SourceInfo
 
 package object verification {
+
+  /** Base simulation-only component. */
+  abstract class BaseSim extends NamedComponent
+
+  case class Assert(predicate: Bool, msg: String = "") extends BaseSim
+  case class Assume(predicate: Bool, msg: String = "") extends BaseSim
+  case class Cover(predicate: Bool, msg: String = "") extends BaseSim
```
jackkoenig, line 17: These should not be `case class`es because all `Data` (`Bool` included) implement referential equality and are thus not appropriate for the structural equality implications of `case classes`.

They should be `final class`es though to ensure they are not user extensible by accident.

src/test/scala/chiselTests/experimental/verification/VerificationSpec.scala
```diff
@@ -20,25 +24,81 @@ class VerificationModule extends Module {
...
+    assertContains(lines, "when _T_9 : @[VerificationSpec.scala")
+    assertContains(lines, "assert(clock, _T_7, UInt<1>(\"h1\"), \"\") @[VerificationSpec.scala")
+  }
+
+  property("annotation of verification constructs should work") {
+    val testDir = new File("test_run_dir", "VerificationAnnotationTests")
+    (new ChiselStage).emitFirrtl(
+      gen = new VerificationAnnotationTest.AnnotationTest,
+      args = Array ("-td", testDir.getPath)
+    )
```
jackkoenig, line 90: I think this test should check that the verification nodes have the right names in the FIRRTL

jackkoenig (changes requested):
@sequencer is right that we also need to be able to annotate the `printf` and `stop` nodes. I don't think it has to be in this PR, but it does perhaps imply that base type `BaseSim` should not live in package `chisel3.experimental.verification`. I'm not sure that it **shouldn't** but the point is that `printf` and `stop` will also be returning objects that are subtypes of `BaseSim`.

Other than that thought, looks good, just a couple of minor issues

sequencer:
Another thing is verification should be able to consume `Printable` to provide a simpler user api. Maybe not in PR, but since SiFive start to refactor this, I think this should also be mentioned.

davidbiancolin:
Oooh, exciting!

src/test/scala/chiselTests/experimental/verification/VerificationSpec.scala
debs-sifive: oh, i didn't even update the emitter to emit the names. Are there guidelines I should follow for emission?

Should I reuse `node`?
```
node assm = assume(clock, _T_3, UInt<1>(1), "") @[VerificationSpec.scala 56:29]
```

or make something new?
```
verif assm : assume(clock, _T_3, UInt<1>(1), "") @[VerificationSpec.scala 56:29]
```

Should I also do something special on the SV emission with the names? Like turn them into labels?

ekiwi: The syntax for named statements in firrtl looks like this:

```
assume(clock, _T_3, UInt<1>(1), "") : assm @[VerificationSpec.scala 56:29]
```

debs-sifive: Thanks Kevin! I guess messing with whether or not the name comes out as a label has to be done in FIRRTL...

debs-sifive:
> Another thing is verification should be able to consume Printable to provide a simpler user api

@sequencer another proposal that has been discussed: instead of verification consuming Printable, we should get rid of the `msg` field in verification components entirely. On top of verification/Printable, we can then create a user-facing API that will correlate Printables to verification components through MultiTargetAnnotations (hence this annotation PR).

But for that to work, as you said, Printables must also be annotate-able.

src/main/scala/chisel3/internal/firrtl/Emitter.scala
```diff
@@ -80,8 +80,7 @@ private class Emitter(circuit: Circuit) {
         val printfArgs = Seq(e.clock.fullName(ctx), "UInt<1>(1)",
           "\"" + printf.format(fmt) + "\"") ++ args
         printfArgs mkString ("printf(", ", ", ")")
-      case e: Verification[_] => s"${e.op}(${e.clock.fullName(ctx)}, ${e.predicate.fullName(ctx)}, " +
-        s"UInt<1>(1), " + "\"" + s"${printf.format(e.message)}" + "\")"
+      case e: Verification[_] => s"${e.op}(${e.clock.fullName(ctx)}, ${e.predicate.fullName(ctx)}, UInt<1>(1), \"${printf.format(e.message)}\") : ${e.id.toNamed.name}"
```
jackkoenig, line 83: This fails in Scala 2.12 due to a super long running bug (finally fixed in Scala 2.13) that you can't simply escape double quotes inside of string interpolated blocks, for example: https://scastie.scala-lang.org/s4MlbfAAT4OLse6I0A6oqA

You can either do what was originally done here with the string contentions, or use triple quotes with string interpolation, eg. `s"""..."""` (and just not escape the `"` inside)

jackkoenig: Also, need to make sure we're doing the equivalent in the `Converter`.

src/test/scala/chiselTests/experimental/verification/VerificationSpec.scala
ekiwi: If you provide a non-empty name (i.e. not `""`), `firrtl` will always include it when serializing.
I am not familiar with the `chisel` code base, but my understanding was that it has its own Chirrtl serializer, no?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -202,6 +202,18 @@ private[chisel3] trait HasId extends InstanceId {
...
+    */
+  private[chisel3] def forceOptionalName(namespace: Namespace): Unit =
+    if(_ref.isEmpty) {
+      val candidate_name = computeName(None, None)
+      if (!candidate_name.isEmpty) {
+        val available_name = namespace.name(candidate_name.get)
+        setRef(Ref(available_name))
+      }
+    }
+
```
debs-sifive, line 216: @jackkoenig please check this implementation

debs-sifive: not sure if i should be providing different arguments to `computeName`

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -220,7 +232,10 @@ private[chisel3] trait HasId extends InstanceId {
     case Some(p) => p._component match {
       case Some(c) => _ref match {
         case Some(arg) => arg fullName c
-        case None => computeName(None, None).get
+        case None => computeName(None, None) match {
+          case Some(name) => name
+          case None => throwException("trying to access a name that cannot be computed")
+        }
```
debs-sifive, line 238: @jackkoenig this is my attempt at throwing an exception for annotating stuff without names. Please let me know if there's a better place to put this/word this.

core/src/main/scala/chisel3/internal/Builder.scala
ekiwi: Are you trying to allow `assert` statements to not have a name?
If so, I think we should just name them all, even if it isn't strictly necessary in order to simplify things.

debs-sifive: yeah, the hope is to avoid IR size issues (probably a sifive-specific problem)

jackkoenig: And just general clutter of naming them all, but of course we could and it would be a bit simpler.

core/src/main/scala/chisel3/experimental/verification/package.scala
```diff
@@ -8,36 +8,52 @@ import chisel3.internal.firrtl.{Formal, Verification}
...
+
+  /** Named class for assertions. */
+  final class Assert(predicate: Bool, msg: String = "") extends BaseSim
+
+  /** Named class for assumes. */
+  final class Assume(predicate: Bool, msg: String = "") extends BaseSim
+
+  /** Named class for covers. */
+  final class Cover(predicate: Bool, msg: String = "") extends BaseSim
+
```
jackkoenig, line 20: Do you want the `predicate` and `msg` to be accessible? If so they should be `vals` (`case classes` make their parameters vals automatically, but since we changed them to regular classes, this is no longer true).

core/src/main/scala/chisel3/internal/Builder.scala
ekiwi: I don't think it would have any significant overhead since there are so few of them. If all verification statements have names, that would also open up the possibility to switch to generating SystemVerilog `property` assertions instead of concurrent assertions.
So personally I would recommend to avoid the complexity and just go all in on naming verification statements.

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: Seems reasonable. We probably ought to just error instead of calling `computeName(None, None)` but that's unrelated to this PR so I think what you're here doing is good.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -202,6 +202,18 @@ private[chisel3] trait HasId extends InstanceId {
...
+    * [[autoSeed]] exists.
+    */
+  private[chisel3] def forceOptionalName(namespace: Namespace): Unit =
+    if(_ref.isEmpty) {
+      val candidate_name = computeName(None, None)
+      if (!candidate_name.isEmpty) {
+        val available_name = namespace.name(candidate_name.get)
+        setRef(Ref(available_name))
+      }
+    }
```
jackkoenig:
```suggestion
  private[chisel3] def maybeName(namespace: Namespace): Unit =
    if (_ref.isEmpty) {
      val candidate_name = computeName(None, None)
      if (candidate_name.nonEmpty) {
        val available_name = namespace.name(candidate_name.get)
        setRef(Ref(available_name))
      }
    }
```
A couple of nits: `forceOptionalName` doesn't really make sense, so `maybeName` seems more accurate or perhaps `nameIfSuggestedOrSeeded` if we want to get very explicit 🙂

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -3,13 +3,13 @@
...
 
 import firrtl.{ir => fir}
-
 import chisel3._
 import chisel3.internal._
 import chisel3.internal.sourceinfo.SourceInfo
 import chisel3.experimental._
 import _root_.firrtl.{ir => firrtlir}
 import _root_.firrtl.PrimOps
+import chisel3.DontCare._parent
```
jackkoenig, line 12:
```suggestion
```
This import appears to be unused and is a weird one that probably should never be used 🧐

core/src/main/scala/chisel3/internal/Builder.scala
debs-sifive: ok, i'll go ahead and name everything then

core/src/main/scala/chisel3/internal/Builder.scala
debs-sifive: resolving b/c i switched to naming everything

core/src/main/scala/chisel3/experimental/verification/package.scala
ekiwi: Please don't make `msg` accessible. That will just make it harder to deprecate it later.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
debs-sifive: oops, blaming this one on intelliJ >_>

core/src/main/scala/chisel3/internal/Builder.scala
debs-sifive: i switched to naming everything, so i am going to just not make this function for now.

core/src/main/scala/chisel3/experimental/verification/package.scala
debs-sifive: i will just remove `msg`, since we're not doing anything with these classes yet.

jackkoenig (approved):
LGTM! Excellent work!

src/test/scala/chiselTests/experimental/verification/VerificationSpec.scala
```diff
@@ -20,25 +24,131 @@ class VerificationModule extends Module {
...
+    // compile circuit
+    val testDir = new File("test_run_dir", "VerificationAnnotationTests")
+    (new ChiselStage).emitSystemVerilog(
+      gen = new AnnotationTest,
+      args = Array("-td", testDir.getPath)
+    )
+
+    // read in annotation file
+    val annoFile = new File(testDir, "AnnotationTest.anno.json")
+    annoFile.exists()
```
jackkoenig, line 92:
```suggestion
    annoFile should exist
```
`.exists` just returns `Boolean`, you need to actually assert that it's true.

jackkoenig: Same in a couple of other places

ekiwi (approved):
LGTM! :shipit:

ekiwi:
Before we merge this, it would be good to change the PR title to something that describes the changes better @debs-sifive .

PR #1974, opened 2021-06-24
Fix CloneModuleAsRecord support for .toTarget

jackkoenig:
This fixes an issue where you cannot currently annotate the fields of the `Record` returned by `CloneModuleAsRecord`.
This is due to the fact that `CloneModuleAsRecord` violates many assumptions in the Chisel internals and currently only works for generating FIRRTL by coincidence. I wanted to make it as first class as possible without a full refactor of the internals.

Basically, there are 3 pieces of bookkeeping (often with duplicated information) that need to be in sync and "correct" with regard to the baked-in assumptions in the compiler:
* binding - used mostly for correctness checking 
*  `_ref` - the "IR" information for most things, used for generating FIRRTL and for generating targets
* `_parent` - the parent module, used for generating targets, correctness checking, and in converting `_ref` to FIRRTL

The current implementation returns a Record that is the "IO" of a Module that doesn't actually exist with a WireBinding. The conversion to FIRRTL works because the `Emitter` and `Converter` (speaking of duplicated logic lol) both think that this Record is an Aggregate Wire whose name just happens to correspond to the name of the module instance that was created.

The fundamental issue is that this `Record` is weird--the `Record` itself isn't a real entity, connecting to it only works because we split up connections in `import chisel3._`, never emitting bulk connections. My goal in this PR was to support these `Records` as "first class" as I can, which I did in the following way:

I added a new private `BaseModule` subtype called `ModuleClone` so that the the cloned `Records` can have a true parent with the correct instance and module names. This necessitated 1 very reasonable change that `generateComponent()` now returns an `Option` so that we can legally instantiate modules that don't have a definition. 

~~The hackiness here is that we need to name the instance of the module from the name the Record ultimately gets. This is the weird `setRef` stuff you see in the overridden `setRef` in `ClonePorts`.~~

~~Speaking of overriding `setRef` in `ClonePorts`, this is definitely weird. I added a new `Arg` type `ModuleCloneIO` (analogous to `ModuleIO` with slightly different behavior, look at IR.scala and see them side-by-side). The problem is that `ModuleIO` is usually set when closing the parent Module of the ports, but the `ClonePorts` `Record` is actually created inside the _parent_ of the module whose ports it represents, so we need a way of it knowing to do this. Worse, you only want to do this for the first instance of `ClonePorts` since clones of *that* could be valid `Wires` or `Regs` and you definitely don't want to accidentally set their refs to be `ModuleCloneIO`~~

(Writing this made me come up with another idea. I'm keeping this text but there's a better way inbound)

The main hackiness in this PR is that we need to set the `_parent` for all of the `Data` elements within the `ClonePorts` `Record`. Currently in Chisel, `_parent` is set upon construction for `Data` and `Modules`. I think this basic pattern should probably change (There's no reason to set `_parent` for literals or unbound types for example), but I didn't want to try to refactor internals in a bug fix.

The main contribution of this PR is that by having a **real** `Module` and a **real** `Arg` for the `ClonePorts` `Records`, most assumptions baked into Chisel's internals are now correct. It's still not perfect because the `Record` itself still exposes `.toTarget` which returns `ReferenceTarget` and it'd be nice if we could instead return an `InstanceTarget` or `ModuleTarget`. 

### Contributor Checklist

#### Type of Improvement

 - bug fix
 - code refactoring

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix `.toTarget` and related naming APIs for `CloneModuleAsRecord` `Records` so that they can be annotated

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -64,13 +64,18 @@ object Module extends SourceInfoDoc {
     Builder.currentClock = saveClock   // Back to clock and reset scope
     Builder.currentReset = saveReset
 
-    val component = module.generateComponent()
-    Builder.components += component
+    // Only add the component if the module generates one
+    val componentOpt = module.generateComponent()
+    for (component <- componentOpt) {
+      Builder.components += component
+    }
 
     Builder.setPrefix(savePrefix)
 
     // Handle connections at enclosing scope
-    if(!Builder.currentModule.isEmpty) {
+    // We use _component because Modules that don't generate them may still have one
+    if (Builder.currentModule.isDefined && module._component.isDefined) {
+      val component = module._component.get
       pushCommand(DefInstance(sourceInfo, module, component.ports))
       module.initializeInParent(compileOptions)
     }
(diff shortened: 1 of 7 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -24,16 +24,24 @@ private[chisel3] object Converter {
...
-      val link = "https://github.com/chipsalliance/chisel3/issues/new"
-      val msg = s"Internal error! Could not get ref for '$id'$loc!$module " +
-                s"This is a bug in Chisel, please file an issue at '$link'."
-      throwException(msg)
+      reportInternalError(s"Could not get ref for '$id'$loc!$module")
     }
 
+  private def clonedModuleIOError(mod: BaseModule, name: String, sourceInfo: SourceInfo): Nothing = {
+    val loc = sourceInfo.makeMessage(" " + _)
+    reportInternalError(s"Trying to convert a cloned IO of $mod inside of $mod itself$loc!")
```
mwachs5, line 42: should there be a space between `itself$loc`?

core/src/main/scala/chisel3/Module.scala
```diff
@@ -178,20 +183,73 @@ package internal {
...
+          case (Some(mod), Ref(name)) =>
+            // Our _parent is a ModuleClone that needs its ref to match ours for .toAbsoluteTarget
+            _parent.foreach(_.setRef(Ref(name), force=true))
+            // Return a specialize-ref that will do the right thing
+            ModuleCloneIO(mod, name)
+          case _ => imm
+        }
+        super.setRef(immx, force)
+      }
+    }
```
jackkoenig, line 229: This is a crime and I'm going to try a different approach

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
jackkoenig: Nope, the `makeMessage` above is basically the same as `.map(" " + _).getOrElse("")` on an `Option[String]`. `loc` will be empty if undefined or has a space before it if it is defined.

jackkoenig:
@azidar I've pushed another commit that cleans up the hacky bit, expressing it in a more direct and obvious way. I think this is ready for review.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -308,9 +366,12 @@ package experimental {
 
     /** Legalized name of this module. */
     final lazy val name = try {
-      // If this is a module aspect, it should share the same name as the original module
-      // Thus, the desired name should be returned without uniquification
-      if(this.isInstanceOf[ModuleAspect]) desiredName else Builder.globalNamespace.name(desiredName)
+      // ModuleAspects and ModuleClones are not "true modules" and thus should share
```
azidar, line 369: I like this comment.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -178,20 +183,73 @@ package internal {
...
     private[chisel3] class ClonePorts (elts: Data*)(implicit compileOptions: CompileOptions) extends Record {
       val elements = ListMap(elts.map(d => d.instanceName -> d.cloneTypeFull): _*)
       def apply(field: String) = elements(field)
       override def cloneType = (new ClonePorts(elts: _*)).asInstanceOf[this.type]
     }
 
+    // Recursively set the parent of the start Data and any children (eg. in an Aggregate)
+    private def setAllParents(start: Data, parent: Option[BaseModule]): Unit = {
+      def rec(data: Data): Unit = {
+        data._parent = parent
```
azidar, line 230: It's weird, I never thought to make _parent a var. I think this is surprisingly really good, and useful for stuff.

azidar (approved):
LGTM

core/src/main/scala/chisel3/Module.scala
seldridge: 51yifu

jackkoenig: Yeah, I think something we should consider is either moving `_parent` into the `binding` or `ref` or otherwise not set it by default (there's no reason that `UInt(8.W)` should have a `_parent`). I didn't want to tackle any refactoring in this PR (since I want to backport it) so I took a more minimal change approach.

PR #1992, opened 2021-06-29
have printf return BaseSim subclass so it can be named/annotated

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This change extends the user API, as users can now annotate `printf` calls.

All `printf`s will be named in FIRRTL output. There should be no impact on generated RTL.

#### Desired Merge Strategy

#### Release Notes

`printf` nodes will all be given default names unless one is suggested/provided by the user. `printf` calls can be annotated.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -2,14 +2,11 @@
 
 package chisel3
 
+import chisel3.experimental.BaseSim
 import chisel3.internal.firrtl.Component
 
 import scala.collection.mutable
-
-import java.util.{
-  MissingFormatArgumentException,
-  UnknownFormatConversionException
-}
+import java.util.{MissingFormatArgumentException, UnknownFormatConversionException}
 
 /** Superclass of things that can be printed in the resulting circuit
   *
@@ -46,7 +43,7 @@ import java.util.{
 //   Currently impossible because unpack is called before the name is selected
 //   Could be implemented by adding a new format specifier to Firrtl (eg. %m)
 // TODO Should we provide more functions like map and mkPrintable?
-sealed abstract class Printable {
+sealed abstract class Printable extends BaseSim {
   /** Unpack into format String and a List of String arguments (identifiers)
     * @note This must be called after elaboration when Chisel nodes actually
     *   have names
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -2,11 +2,14 @@
 
 package chisel3
 
-import chisel3.experimental.BaseSim
 import chisel3.internal.firrtl.Component
 
 import scala.collection.mutable
-import java.util.{MissingFormatArgumentException, UnknownFormatConversionException}
+
+import java.util.{
+  MissingFormatArgumentException,
+  UnknownFormatConversionException
+}
 
 /** Superclass of things that can be printed in the resulting circuit
   *
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -43,7 +43,7 @@ import java.util.{MissingFormatArgumentException, UnknownFormatConversionExcepti
 //   Currently impossible because unpack is called before the name is selected
 //   Could be implemented by adding a new format specifier to Firrtl (eg. %m)
 // TODO Should we provide more functions like map and mkPrintable?
-sealed abstract class Printable extends BaseSim {
+sealed abstract class Printable {
   /** Unpack into format String and a List of String arguments (identifiers)
     * @note This must be called after elaboration when Chisel nodes actually
     *   have names
--- a/core/src/main/scala/chisel3/Printf.scala
+++ b/core/src/main/scala/chisel3/Printf.scala
@@ -8,9 +8,6 @@ import chisel3.internal.Builder.pushCommand
 import chisel3.internal.sourceinfo.SourceInfo
 import chisel3.experimental.BaseSim
 
-/** Named class for [[printf]]s. */
-final class PrintfId(val pable: Printable) extends BaseSim
-
 /** Prints a message in simulation
   *
   * See apply methods for use
@@ -35,6 +32,9 @@ object printf {
     }
     formatIn map escaped mkString ""
   }
+
+  /** Named class for [[printf]]s. */
+  final class Printf(val pable: Printable) extends BaseSim
 
   /** Prints a message in simulation
     *
@@ -73,7 +73,7 @@ object printf {
     * @param fmt printf format string
     * @param data format string varargs containing data to print
     */
-  def apply(fmt: String, data: Bits*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): PrintfId =
+  def apply(fmt: String, data: Bits*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
     apply(Printable.pack(fmt, data:_*))
   /** Prints a message in simulation
     *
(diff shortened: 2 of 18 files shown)


debs-sifive:
cc @jackkoenig

ekiwi:
This might be interesting for Firesim @davidbiancolin

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -46,7 +43,7 @@ import java.util.{
 //   Currently impossible because unpack is called before the name is selected
 //   Could be implemented by adding a new format specifier to Firrtl (eg. %m)
 // TODO Should we provide more functions like map and mkPrintable?
-sealed abstract class Printable {
+sealed abstract class Printable extends BaseSim {
```
jackkoenig, line 46: I don't think it's `Printable` that we want to be annotatable--`Printable` is sort of like a "Hardware String" so it's a little funky to annotate that as opposed to the `Printf` statement itself.

 I think we just want `Printf` and `Stop` to have types that extend `BaseSim` like you did with `assert`, `assume`, and `cover`.

sequencer:
and `Stop`？

core/src/main/scala/chisel3/Printable.scala
sequencer: +1

sequencer: And another thought is, like what you commented in https://github.com/chipsalliance/chisel3/pull/1968#issuecomment-865336046.
Should we use annotation to represent contents for `Printf` and `Stop`?

core/src/main/scala/chisel3/Printf.scala
```diff
@@ -3,11 +3,13 @@
...
 import scala.language.experimental.macros
-
 import chisel3.internal._
 import chisel3.internal.Builder.pushCommand
-import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.SourceInfo
+import chisel3.experimental.BaseSim
+
+/** Named class for [[printf]]s. */
+final class PrintfId(val pable: Printable) extends BaseSim
```
debs-sifive, line 12: Anyone have a better name suggestion here? Didn't want to call it `Printf` and have it collide with the IR node.

mwachs5: `NamedPrintf`?

core/src/main/scala/chisel3/Printable.scala
debs-sifive: @sequencer I'm going to not deal with `stop` for now, because it's not strictly necessary for asserts/assumes/covers to be correlated with `stop`s.

jackkoenig (approved):
LGTM! Nice work!

core/src/main/scala/chisel3/Printf.scala
jackkoenig: A collision with the IR node is fine, no users should ever run into that and we can easily distinguish internally

sequencer (approved):
FireSim will be happy ;p

ekiwi:
@debs-sifive could you update the PR title please?

debs-sifive:
Title is updated, but please hold off on merging for a bit — I'm attempting to rename `PrintfId` to `Printf` (not sure if it will work nicely though)

ekiwi:
> Title is updated, but please hold off on merging for a bit — I'm attempting to rename `PrintfId` to `Printf` (not sure if it will work nicely though)

Sounds good. Please remove the `DO NOT MERGE` label once you are ready (or ask one of us to remove it in case you are lacking the permissions).

core/src/main/scala/chisel3/Printf.scala
```diff
@@ -34,6 +33,9 @@ object printf {
     formatIn map escaped mkString ""
   }
 
+  /** Named class for [[printf]]s. */
+  final class Printf(val pable: Printable) extends BaseSim
+
```
debs-sifive, line 38: @jackkoenig I moved this into the `printf` object, because I was getting really weird errors having it out in the base `chisel3` package:

```
[error] /Users/debs/Documents/chisel3/src/main/scala/chisel3/aop/Select.scala:416:33: type Printf is not a member of package chisel3
[error]   case class Printf(id: chisel3.Printf, preds: Seq[Predicate], pable: Printable, clock: Clock) extends Serializeable {
[error]                                 ^
[error] /Users/debs/Documents/chisel3/src/main/scala/chisel3/compatibility.scala:302:24: object printf is not a member of package chisel3
[error] did you mean Printf?
[error]   val printf = chisel3.printf
[error]       
...
[error] /Users/debs/Documents/chisel3/src/main/scala/chisel3/internal/firrtl/Emitter.scala:81:18: missing argument list for method printf in object Predef
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `printf _` or `printf(_,_)` instead of `printf`.
[error]           "\"" + printf.format(fmt) + "\"") ++ args
[error]                  ^
[error] /Users/debs/Documents/chisel3/src/main/scala/chisel3/internal/firrtl/Emitter.scala:84:92: missing argument list for method printf in object Predef
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `printf _` or `printf(_,_)` instead of `printf`.
[error]         s"""${e.op}(${e.clock.fullName(ctx)}, ${e.predicate.fullName(ctx)}, UInt<1>(1), "${printf.format(e.message)}") : ${e.name}"""
[error]   
```

Do you think this is a suitable place for this class to live?

core/src/main/scala/chisel3/experimental/verification/package.scala
```diff
@@ -9,16 +9,10 @@ import chisel3.internal.sourceinfo.SourceInfo
...
-  /** Named class for assumes. */
-  final class Assume(val predicate: Bool) extends BaseSim
-
-  /** Named class for covers. */
-  final class Cover(val predicate: Bool) extends BaseSim
-
   object assert {
+    /** Named class for assertions. */
+    final class Assert(val predicate: Bool) extends BaseSim
+
```
debs-sifive, line 15: also moved these into their corresponding objects for consistency with [this comment](https://github.com/chipsalliance/chisel3/pull/1992/files#r662476858)

core/src/main/scala/chisel3/Printf.scala
jackkoenig: We could resolve these issues but I kind of like having each of them within their respective objects. It's a bit neater than our usually strategy of dumping everything in package `chisel3`.

debs-sifive: kk, i'm happy with it if you are! Can you help me merge if everything else looks good?

core/src/main/scala/chisel3/experimental/verification/package.scala
```diff
@@ -9,16 +9,10 @@ import chisel3.internal.sourceinfo.SourceInfo
...
-
-  /** Named class for assumes. */
-  final class Assume(val predicate: Bool) extends BaseSim
-
-  /** Named class for covers. */
-  final class Cover(val predicate: Bool) extends BaseSim
-
   object assert {
+    /** Named class for assertions. */
+    final class Assert(val predicate: Bool) extends BaseSim
```
jackkoenig, line 14:
```suggestion
    final class Assert extends BaseSim
```
Or
```scala
    final class(private[chisel3] val predicate: Bool) Assert extends BaseSim
```

I can't remember if we talked about this previously (I couldn't find any previous PR comments about this), but at the moment I'm leaning against publicly exposing the `predicate`, at least for now. My concern is about what behavior a user might expect vs. what they'll see. For example:

```scala
var assertRef: assert.Assert = _
when (foo) {
  assertRef = assert(fizz =/= buzz, "<useful message>")
}
// Use assertRef.predicate
```
My worry is that the users will expect `assertRef.predicate` to be equal to `!reset || !foo || (fizz =/= buzz)` rather than just `fizz =/= buzz`. I will grant that it could be useful though so I'm open to being convinced.

I'm curious what @ekiwi thinks about this API.

Perhaps if users want what I'm assuming they will expect, they can also capture `when.cond` so the onus is just on them there. For the record, I don't think we _want_ to capture `!reset || !foo || (fizz =/= buzz)` because it will either always generate FIRRTL nodes that are almost always unused or will be somewhat hard to get right (requires some extension of https://github.com/chipsalliance/chisel3/pull/1518)

ekiwi: I think you bring up an interesting point. For now I think the main goal is to get annotation support into the `3.5` release, so I would keep `predicate` private. When you come up with a use-case for accessing predicate at SiFive, then you can always make another PR that just exposes predicate along with the user-story.

PR #1999, opened 2021-07-01
Fix ChiselEnum warnings and use Logger for warnings instead of println

jackkoenig:
Add ChiselEnum.safe factory method and avoid warning

Previously, ChiselEnum would warn any time a UInt is converted to an
Enum. There was no way to suppress this warning. Now there is a factory
method (`.safe`) that does not warn and returns (Enum, Bool) where the
Bool is the result of calling .isValid on an Enum object. The regular
UInt cast is also now smarter and will not warn if all bitvectors of the
width of the Enum are legal states.

I also added a useful testing utility to capture Logging information

I need to update the mdoc to include this information (and a few stylistic changes), will push another commit with that.

### Contributor Checklist

#### Type of Improvement

   - bug fix     
  - new feature/API   

#### API Impact

Adds `ChiselEnum.safe` for safely casting from `UInt` to an Enum

No impact

#### Desired Merge Strategy

 - Rebase

#### Release Notes

Stop warning when casting from `UInt` to a `ChiselEnum` if all bitvectors of the given width are legal states. Also provide `ChiselEnum.safe` for safely casting from `UInt` to an enum if some states are illegal.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -131,7 +131,7 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
     if (litOption.isDefined) {
       true.B
     } else {
-      factory.all.map(this === _).reduce(_ || _)
+      if (factory.isTotal) true.B else factory.all.map(this === _).reduce(_ || _)
     }
   }
 
@@ -233,6 +233,12 @@ abstract class EnumFactory {
 
   private[chisel3] val enumTypeName = getClass.getName.init
 
+  // Do all bitvectors of this Enum's width represent legal states?
+  private[chisel3] def isTotal: Boolean = {
+    (this.getWidth < 31) && // guard against Integer overflow
+      (enumRecords.size == (1 << this.getWidth))
+  }
+
   private[chisel3] def globalAnnotation: EnumDefChiselAnnotation =
     EnumDefChiselAnnotation(enumTypeName, (enumNames, enumValues).zipped.toMap)
 
@@ -277,7 +283,7 @@ abstract class EnumFactory {
 
   def apply(): Type = new Type
 
-  def apply(n: UInt)(implicit sourceInfo: SourceInfo, connectionCompileOptions: CompileOptions): Type = {
+  private def castImpl(n: UInt, warn: Boolean)(implicit sourceInfo: SourceInfo, connectionCompileOptions: CompileOptions): Type = {
     if (n.litOption.isDefined) {
       enumInstances.find(_.litValue == n.litValue) match {
         case Some(result) => result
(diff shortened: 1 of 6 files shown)


core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -233,6 +233,9 @@ abstract class EnumFactory {
 
   private[chisel3] val enumTypeName = getClass.getName.init
 
+  // Do all bitvectors of this Enum's width represent legal states?
+  private[chisel3] def isTotal: Boolean = enumRecords.size == (1L << this.getWidth)
```
seldridge, line 237: I think `ArrayBuffer.size` returns an `Int`. So, there may be some weird behavior here for widths > 32 bits. I don't anticipate this actually coming up, though.

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -288,15 +291,37 @@ abstract class EnumFactory {
     } else if (n.getWidth > this.getWidth) {
       throwException(s"The UInt being cast to $enumTypeName is wider than $enumTypeName's width ($getWidth)")
     } else {
-      Builder.warning(s"Casting non-literal UInt to $enumTypeName. You can check that its value is legal by calling isValid")
-
+      // Check if its possible for `n` to hit an illegal state
+      val canHitIllegalState = warn && !this.isTotal
+      if (canHitIllegalState) {
```
seldridge, line 296: Nit: consider inlining since this has one use.

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -288,15 +291,37 @@ abstract class EnumFactory {
...
     }
   }
+
+  /** Cast an [[UInt]] to the type of this Enum
+    *
+    * @note will give a Chisel elaboration time warning if the argument could hit invalid states
+    * @param n the UInt to cast
+    * @return the equivalent Enum to the value of the cast UInt
+    */
+  def apply(n: UInt)(implicit sourceInfo: SourceInfo, connectionCompileOptions: CompileOptions): Type = castImpl(n, warn = true)
```
seldridge, line 313: I like the clear use of `warn = true`. :+1:

core/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -184,31 +185,31 @@ private[chisel3] class ErrorLog {
   }
 
   /** Throw an exception if any errors have yet occurred. */
-  def checkpoint(): Unit = {
+  def checkpoint(logger: Logger): Unit = {
     deprecations.foreach { case ((message, sourceLoc), count) =>
-      println(s"${ErrorLog.depTag} $sourceLoc ($count calls): $message")
+      logger.warn(s"${ErrorLog.depTag} $sourceLoc ($count calls): $message")
```
seldridge, line 190: ❤️ ❤️ ❤️

seldridge (approved):

core/src/main/scala/chisel3/StrongEnum.scala
jackkoenig: It was longer before I factored `isTotal` out into a method which is why I named it originally, I'll inline it.

core/src/main/scala/chisel3/StrongEnum.scala
jackkoenig: Fair point on `.size`, I'm just in the habit of always using `Long` when left-shifting but I think it's irrelevant here. We could add an assertion that `this.getWidth` is < 31 and call it a day.

jackkoenig: Done by guarding against Integer overflow

core/src/main/scala/chisel3/StrongEnum.scala
jackkoenig: Done.

jackkoenig:
I pushed some docs updates too.

seldridge (approved):

PR #2004, opened 2021-07-04
Refactor tests

sequencer:
It deprecate `chisel3.testers`, which should not be a user API.
API `TreadleBackend` in `tests` package is not visible to `chisel3` package, so I move all APIs to `chisel3Tests` package, and refactor the imports.
After adding 7687c1d
sbt will add the compiler plugin automatically from chisel project???
Tuned this for the whole weekend, and gave up.

### Contributor Checklist

#### Type of Improvement
- code refactoring

#### API Impact
deprecate APIs

None

#### Desired Merge Strategy

#### Release Notes
None

- [ ] Did you review?


ekiwi:
Could you investigate whether there is a way to enable/disable certain tests depending on whether the action was triggered by a push or a PR? I would prefer to avoid having two completely separate github action files since that increases the maintenance burden.

sequencer:
> Could you investigate whether there is a way to enable/disable certain tests depending on whether the action was triggered by a push or a PR? I would prefer to avoid having two completely separate github action files since that increases the maintenance burden.

Done:)

sequencer:
I'm not familiar with sbt,
I'm finding a way to make `no-plugin-test` depends on `chisel3 / Test` while not using the plugin.
sbt sucks...

sequencer:
OMG never image we can use symlink to do this hack! Thanks @jackkoenig

src/main/scala/chisel3/testers/BasicTester.scala
```diff
@@ -9,6 +9,7 @@ import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.SourceInfo
 
+@deprecated("This is not a User API, will be removed in Chisel 3.6", "3.5")
 class BasicTester extends Module() {
```
sequencer, line 13: Deprecation here.

src/main/scala/chisel3/testers/TesterDriver.scala
```diff
@@ -13,6 +13,7 @@ import firrtl.options.{Dependency, Phase, PhaseManager, TargetDirAnnotation, Uns
 import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlStage}
 import firrtl.transforms.BlackBoxSourceHelper.writeResourceToDirectory
 
+@deprecated("This is not a User API, will be removed in Chisel 3.6", "3.5")
 object TesterDriver extends BackendCompilationUtilities {
```
sequencer, line 17: Deprecation here.

sequencer:
subsumed by #2043

PR #2015, opened 2021-07-07
Update docs on dontTouch as optimization barrier

seldridge:
Change the Scaladoc for the dontTouch utility.  Indicate that this is an
_optimization barrier_ and not just a guarantee that the signal won't be
removed.  The optimization barrier interpretation is the current
implementation in the Scala FIRRTL Compiler.

### Contributor Checklist

- [n/a] Did you add at least one test demonstrating the PR?

#### Type of Improvement

- documentation                      

#### API Impact

None.

None.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/dontTouch.scala
+++ b/core/src/main/scala/chisel3/dontTouch.scala
@@ -5,7 +5,8 @@ package chisel3
 import chisel3.experimental.{ChiselAnnotation, annotate, requireIsHardware}
 import firrtl.transforms.DontTouchAnnotation
 
-/** Marks that a signal should not be removed by Chisel and Firrtl optimization passes
+/** Marks that a signal is an optimization barrier to Chisel and the FIRRTL compiler. This has the effect of
+  * guaranteeing that a signal will not be removed.
   *
   * @example {{{
   * class MyModule extends Module {
@@ -21,9 +22,11 @@ import firrtl.transforms.DontTouchAnnotation
   * @note Calling this on [[Data]] creates an annotation that Chisel emits to a separate annotations
   * file. This file must be passed to FIRRTL independently of the `.fir` file. The execute methods
   * in [[chisel3.Driver]] will pass the annotations to FIRRTL automatically.
+  * @note Because this is an optimization barrier, constants will not be propagated through a signal marked as
+  * dontTouch.
   */
 object dontTouch {
-  /** Marks a signal to be preserved in Chisel and Firrtl
+  /** Mark a signal as an optimization barrier to Chisel and FIRRTL.
     *
     * @note Requires the argument to be bound to hardware
     * @param data The signal to be marked


core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -5,7 +5,8 @@ package chisel3
 import chisel3.experimental.{ChiselAnnotation, annotate, requireIsHardware}
 import firrtl.transforms.DontTouchAnnotation
 
-/** Marks that a signal should not be removed by Chisel and Firrtl optimization passes
+/** Marks that a signal is an optimization barrier to Chisel and the FIRRTL compiler. This has the effect of
```
mwachs5, line 8: Is it obvious what "optimization barrier" means?
```suggestion
/** Marks that a signal is an optimization barrier to Chisel and the FIRRTL compiler.
  * The logical function on either side of the signal will be maintained. This has the effect of
```

ekiwi: I was wondering the same thing.

seldridge: That would be helpful. I'm not 100% how to describe this, though, and I'm not sure if the suggestion captures it, either. (Since I'd say that optimizing across it would still preserve the logical function on either side to still be the same.) 

What about moving the `optimization barrier` language into a note and saying that this will both preserve the signal and prevent constants from being propagate through it?

ekiwi: > prevent constants from being propagate through it?

Why would we want to guarantee that? I thought the whole point of `dontTouch` is to keep the signal around so that you can `bind` to it and see it in a waveform viewer. Why should a user care if we constant prop through the signal?
To me that would seem to restrictive for a Chisel-level API.

seldridge: > Why would we want to guarantee that?

:joy: [Because that's what this API is](https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/transforms/ConstantPropagation.scala#L579). :joy: I had the same reaction as you. Designers have wanted this and will use it to control downstream compilation. I agree that it's kind of a weird API and the name makes it more confusing.

Ideally, we'd have an API that gives you fine grained control over the product of: (1) signal is preserved and (2) optimization barrier.

This PR is narrowly about documenting exactly what the existing API does.

jackkoenig: > Why would we want to guarantee that?

Because that's what designers have asked for over time.

> the name makes it more confusing.

The name echos the same feature in synthesis tools ([1], [2], and see user guides for things like Design Compiler).

I absolutely agree that we need to distinguish between value semantics of a signal vs. optimization barrier, but historically we've just had the one mechanism so it ended up doing both 🙃 

[1] https://www.micro-ip.com/STA/dictionary_512_17/set_dont_touch.html
[2] https://inst.eecs.berkeley.edu/~cs250/fa09/handouts/tut5-dc.pdf

jackkoenig (approved):

PR #2030, opened 2021-07-09
Added flush capability to Queue

anniej-sifive:
### Contributor Checklist

#### Type of Improvement

 - code refactoring  

- new feature/API    

#### API Impact

This extends the Queue API by adding flush capability to Queues. This flush ability defaults to false to maintain compatibility with previous queues.

If the flush capability is on it will create a flush wire and relevant logic. 

#### Desired Merge Strategy

#### Release Notes

This extends the Queue API by adding flush capability to Queues. 

- [ ] Did you review?


src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -265,14 +281,23 @@ class Queue[T <: Data](val gen: T,
...
 
   val ptr_diff = enq_ptr.value - deq_ptr.value
+  val flush = Wire(Bool())
+
+  if (hasFlush) { 
+    flush := io.flush.get //to mitigate returning None from io.flush.get in logic below
+  } 
+  else { 
+    flush := false.B 
+  }
```
hcook, line 291: I would consider
```
flush := io.flush.getOrElse(false.B)
```

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -265,14 +281,23 @@ class Queue[T <: Data](val gen: T,
   }
 
   val ptr_diff = enq_ptr.value - deq_ptr.value
+  val flush = Wire(Bool())
```
mwachs5, line 284: does this need to be a Wire, or can you just do `val flush := ... `?

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: agree, and this can be a `val flush = ...` right?

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -177,6 +178,9 @@ class QueueIO[T <: Data](private val gen: T, val entries: Int) extends Bundle
   val deq = Flipped(DeqIO(gen))
   /** The current amount of data in the queue */
   val count = Output(UInt(log2Ceil(entries + 1).W))
+  /** Input to choose if the queue should flush entries (Optional IO for a flushable Queue)*/
```
mwachs5, line 181:
```suggestion
  /** When asserted, reset the enqueue and dequeue pointers, effectively flushing the queue (Optional IO for a flushable Queue)*/ 
```

I think it's important to clarify whether this is actually erasing the data or not, since we aren't making any guarantees about that which might have security implications

anniej-sifive: Okay yes that makes sense thank you!

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -265,14 +281,23 @@ class Queue[T <: Data](val gen: T,
...
+
   if (isPow2(entries)) {
-    io.count := Mux(maybe_full && ptr_match, entries.U, 0.U) | ptr_diff
+    io.count := Mux(maybe_full && !flush && ptr_match, entries.U, 0.U) | ptr_diff
   } else {
     io.count := Mux(ptr_match,
-                    Mux(maybe_full,
-                      entries.asUInt, 0.U),
+                    Mux(maybe_full && !flush,
+                      entries.asUInt(), 0.U),
```
jackkoenig, line 298:
```suggestion
                      entries.asUInt, 0.U),
```
Presumably Scala 2.13 is warning you about this, but I would rather fix the definition of `.asUInt` than add these parentheses.

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+import chisel3.util._
+import chisel3.util.random.LFSR
+
+class ThingsPassThroughFlushQueueTester(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+  
+  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))  
+  q.io.flush.get := false.B
+  val elems = VecInit(elements.map {
+    _.asUInt()
+  })
```
jackkoenig, line 16:
```suggestion
  val elems = VecInit(elements.map(_.U))
```
We generally use `.U` for `Int => UInt`, and `.asUInt` for `Data => UInt`

README.md
```diff
@@ -171,6 +171,8 @@ git clone https://github.com/chipsalliance/chisel3.git
 cd chisel3
 sbt compile
 ```
+In order to run the following unit tests, make sure you have Verilator installed and loaded
+(Check this by running 'which verilator')
```
debs-sifive, line 175: I think this README change was already performed in a separate PR. Can you remove it from this PR to keep it clean?

mwachs5: you may need to rebase your changes or merge in latest master

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,6 +244,17 @@ class Queue[T <: Data](val gen: T,
     maybe_full := do_enq
```
sequencer, line 244: Seems there exists a unhandled double write behavior?
What if `do_enq =/= do_deq && io.flush.get`?
`enqueue`(but not `dequeue`) and `flush` at the same cycle.

sequencer (changes requested):

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,6 +244,17 @@ class Queue[T <: Data](val gen: T,
...
+      deq_ptr.reset()
+    }
+
+    when(io.flush.get) {
+      maybe_full := false.B
+    }
+  }
+
   io.deq.valid := !empty
   io.enq.ready := !full
```
mwachs5, line 259: what should behavior of `ready`, and `valid`, be relative to the flush? As @sequencer noted above, can you flush at the same time that you actually enqueue something (seems like the answer shoudl be "no"). I think it's ok to `valid` something the same time though

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,91 @@
+package chiselTests
+
+import org.scalacheck._
+
+import chisel3._
+import chisel3.testers.BasicTester
+import chisel3.util._
+import chisel3.util.random.LFSR
+
+class ThingsPassThroughFlushQueueTester(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
```
mwachs5, line 10: please add a test of what happens if you try to write same cycle as trying to flush (should not work) and if you try to read same cycle as flush (should be OK, i think?)

src/main/scala/chisel3/util/Decoupled.scala
anniej-sifive: Yeah Deborah and I looked at it and we agree, and I'm working on making a test case and updating the logic now!

sequencer: If don't support this, I think `assert(enq && maybe_full)` is enough.
If support, just make `enq` and `deq` in the `otherwise` block of `flush` is OK.(but it will increase the critical path. with a `nand` gate.)

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,8 +244,19 @@ class Queue[T <: Data](val gen: T,
     maybe_full := do_enq
   }
 
+  if (hasFlush) { //flushable queue specific logic
```
debs-sifive, line 247: can we make this comment more specific, like
```suggestion
  if (hasFlush) {
  // when flush is on, empty the queue
```

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,8 +244,19 @@ class Queue[T <: Data](val gen: T,
...
+    }
+
+    when(io.flush.get) {
+      maybe_full := false.B
+    }
+  }
+
   io.deq.valid := !empty
-  io.enq.ready := !full
+  io.enq.ready := !full && !flush
```
debs-sifive, line 259:
```suggestion
// cannot enqueue elements while flush is active
  io.enq.ready := !full && !flush
```

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -265,11 +281,13 @@ class Queue[T <: Data](val gen: T,
   }
 
   val ptr_diff = enq_ptr.value - deq_ptr.value
+  
+
```
debs-sifive, line 285: spacing issue
```suggestion
 
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,181 @@
...
```
debs-sifive, line 10: can you add a scaladoc description r.e. what each test does?

anniej-sifive: okay I'll do that

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,181 @@
...
+}
+
+class FullQueueFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+    val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
+    val elems = VecInit(elements.map(_.U))
+
+  val inCnt = Counter(elements.length + 1)
+  val currDepthCnt = Counter(queueDepth + 1)
+  
+  //testing a flush when queue is empty
```
debs-sifive, line 113: when queue is full in this case?

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,181 @@
...
+    inCnt.inc()
+  }
+  when(q.io.deq.fire()) {
+    outCnt.inc()
+  }
+
+  when(flush) {
+    //check that queue gets flushed
+    assert((q.io.count === 0.U) || q.io.deq.valid, s"in count: ${q.io.count} and queue length: $queueDepth" ) 
+    assert(!q.io.enq.ready, s"in count: ${q.io.count} and queue length: $queueDepth")
```
debs-sifive, line 63: can you also clean up these assert strings to be a bit more descriptive, or else remove them if they're just for debug? Might be better to say something like

```suggestion
    assert((q.io.count === 0.U) || q.io.deq.valid, p"Expected to be able to dequeue a flushed queue if it had elements prior to the flush, but got dequeue valid ${q.io.deq.valid}." ) 
    assert(!q.io.enq.ready, p"Expected enqueue to not be ready when flush is active, but got enqueue ${q.io.enq.ready}.") 
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,181 @@
...
+    inCnt.inc()
+  }
+  when(q.io.deq.fire()) {
+    outCnt.inc()
+  }
+
+  when(flush) {
+    //check that queue gets flushed at the beginning with no elements
+    assert(!q.io.deq.valid, s"in count: ${q.io.count} and queue length: $queueDepth" ) 
+    assert(!q.io.enq.ready, s"in count: ${q.io.count} and queue length: $queueDepth")
```
debs-sifive, line 98: same for these and following asserts

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: @sequencer can you re-review to see if your concern is now addressed?

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,8 +244,21 @@ class Queue[T <: Data](val gen: T,
...
+    when(io.flush.get) {
+      enq_ptr.reset()
+      deq_ptr.reset()
+    }
+
+    when(io.flush.get) {
+      maybe_full := false.B
+    }
+  }
+
```
mwachs5, line 258:
```suggestion
  io.flush.foreach { flush => 
  // when flush is on, empty the queue
    when(flush) {
      enq_ptr.reset()
      deq_ptr.reset()
      maybe_full := false.B
    }
  }

```

mwachs5: or, do you even need to conditionalize this on the `io.flush.foreach`? You already have a well-defined `flush` signal...

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -239,8 +244,21 @@ class Queue[T <: Data](val gen: T,
...
   io.deq.valid := !empty
-  io.enq.ready := !full
+// cannot enqueue elements while flush is active
```
mwachs5, line 260:
```suggestion
  // cannot enqueue elements while flush is active
```

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -187,7 +191,7 @@ class QueueIO[T <: Data](private val gen: T, val entries: Int) extends Bundle
   * @param flow True if the inputs can be consumed on the same cycle (the inputs "flow" through the queue immediately).
   * The ''valid'' signals are coupled.
   * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element.
-  *
+  * @param hasFlush True if generated queue requires a flush feature
```
sequencer, line 194: I think we should leave a comment here to indicate `flush` signal being couple to `ready` signal.
This comment can help user to debug the combinational cycle issue introduced by this signal.

aswaterman (changes requested):
I don't think the interface should couple `flush` to `enq.ready`.  We should just instead define that flush is the higher-priority operation: semantically, the enqueue occurs first, then the flush happens afterwards.

Both interfaces are well-defined, but this one has fewer potentially problematic timing arcs.

The user can externally combine the flush signal with enq.valid if so desired.

src/main/scala/chisel3/util/Decoupled.scala
sequencer: Agree with @aswaterman's comment, it will be much better if we don't add additional arc from `flush` to `ready`.

I still suggest using my previous review in https://github.com/chipsalliance/chisel3/pull/2030#discussion_r668208072

anniej-sifive:
@sequencer @aswaterman Thank you for your feedback! I just committed changes in Decoupled.scala that hopefully cover the concerns you brought up if you want to take a look.

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+ * 
+  * @param elements The sequence of elements used in the queue
+  * @param queueDepth The max number of entries in the queue
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class QueueGetsFlushedTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+    val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
+    val elems = VecInit(elements.map(_.U))
```
debs-sifive, line 54:
```suggestion
  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
  val elems = VecInit(elements.map(_.U))
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+ * 
+  * @param elements The sequence of elements used in the queue
+  * @param queueDepth The max number of entries in the queue
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class FullQueueFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+    val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
+    val elems = VecInit(elements.map(_.U))
```
debs-sifive, line 147:
```suggestion
  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
  val elems = VecInit(elements.map(_.U))
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+    currDepthCnt.inc() //counts how many items have been enqueued
+  }
+
+  when(flushRegister) {
+    currDepthCnt.reset() //resets the number of items currently inside queue
+    //check that queue gets flushed when queue is full
+    assert(q.io.count === 0.U)
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush") 
+    assert(q.io.enq.ready, "Expected enqueue to be ready when flush is high because queue should be empty")
+
```
debs-sifive, line 172:
```suggestion
```

src/test/scala/chiselTests/QueueFlushSpec.scala
sequencer: we seems need to add `scalafmt` check. ;p

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -215,28 +220,35 @@ class Queue[T <: Data](val gen: T,
     }
   }
 
-  val io = IO(new QueueIO(genType, entries))
-
+  val io = IO(new QueueIO(genType, entries, hasFlush))
   val ram = if (useSyncReadMem) SyncReadMem(entries, genType, SyncReadMem.WriteFirst) else Mem(entries, genType)
   val enq_ptr = Counter(entries)
   val deq_ptr = Counter(entries)
   val maybe_full = RegInit(false.B)
```
sequencer, line 227: Try to think about using logic below 
```
val maybe_full = RegNext(!flush && do_enq && do_deq)
```
to get rid of a Mux and a XORr gate.

sequencer:
Only left one possible logic improvement in `maybe_full`, but not sure about it, others looks good. Good job!

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -215,28 +220,35 @@ class Queue[T <: Data](val gen: T,
...
+      ram(enq_ptr.value) := io.enq.bits
+      enq_ptr.inc()
+    }
+    when (do_deq) {
+      deq_ptr.inc()
+    }
+    when (do_enq =/= do_deq) {
+      maybe_full := do_enq
+    }
   }
```
aswaterman, line 252: We should get rid of the arc from `flush` to `ram` because we don't care about the contents of `ram` following a flush.  This is especially desirable since this is a high fan-out path.  To do this, avoid the `otherwise` and rewrite as:

      when (do_enq) { 
        ram(enq_ptr.value) := io.enq.bits
        enq_ptr.inc()
      }
      when (do_deq) { 
        deq_ptr.inc()
      }
      when (do_enq =/= do_deq) { 
        maybe_full := do_enq
      }
      when (flush) {
        enq_ptr.reset()
        deq_ptr.reset()
        maybe_full := false.B
      }

src/main/scala/chisel3/util/Decoupled.scala
aswaterman: I don't think this suggestion is correct: it means that `maybe_full` cannot become true unless a dequeue occurs, which doesn't make sense.

sequencer: Oh, I thought about this, my bad!

src/main/scala/chisel3/util/Decoupled.scala
anniej-sifive: Okay I implemented it like you suggested! Thanks!

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -215,18 +220,19 @@ class Queue[T <: Data](val gen: T,
...
   val maybe_full = RegInit(false.B)
-
   val ptr_match = enq_ptr.value === deq_ptr.value
   val empty = ptr_match && !maybe_full
   val full = ptr_match && maybe_full
   val do_enq = WireDefault(io.enq.fire())
   val do_deq = WireDefault(io.deq.fire())
+  val flush = io.flush.getOrElse(false.B) 
+
+  // when flush is on, empty the queue
```
mwachs5, line 235: This comment seems misplaced now

mwachs5: you might also want to say below where you move the comment to the when(flush) clause:

```
// When flush is asserted, empty the queue.
// Semantically, any enqueues happen before the flush.
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+class QueueGetsFlushedTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
+  val elems = VecInit(elements.map(_.U))
+
+  val inCnt = Counter(elements.length + 1)
+  val outCnt = Counter(elements.length + 1)
+  
+  val halfCnt = (queueDepth + 1)/2
+  val cycleCounter = Counter(elements.length + 1)
+  //testing a flush when
```
mwachs5, line 61:
```suggestion
  //testing a flush when ... ?
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+  val flushRegister = RegInit(false.B)
+  flushRegister := flush
+  q.io.enq.valid := (inCnt.value < elements.length.U)
+  q.io.deq.ready := LFSR(16)(tap)
+  cycleCounter.inc() //counts every cycle
+
+  q.io.enq.bits := elems(inCnt.value)
+  when(q.io.enq.fire()) {
+    inCnt.inc()
+    inCnt.inc()
```
mwachs5, line 73: what does it mean to have this twice here? Do you really mean to increment the counter twice?

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+
+/** Test queue can flush at random times
+ * 
+  * @param elements The sequence of elements used in the queue
+  * @param queueDepth The max number of entries in the queue
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class QueueGetsFlushedTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
```
mwachs5, line 52: why isn't this test checking that we are still enqueueing and dequeuing the elements that we expect? I don't see anywhere in these test cases doing what we did above:

```
  when(q.io.deq.fire()) {
    //ensure that what comes out is what comes in
    assert(elems(outCnt.value) === q.io.deq.bits)
    outCnt.inc()
  }
```

in the presence of flushes. this seems like a pretty big test hole.

This would also help demonstrate the expected behavior because you'd have to write the logic to demonstrate the expected behavior of enqueues and dequeues as the same time as flushes (I still don't see any tests that demonstrate that -- specifically directed tests for "i can still enqueue into a non-full queue while flushing, but I don't expect to read what I enqueued" and "I can dequeue from a non-empty queue while flushing, and I expect to read what I enqueued".

mwachs5 (changes requested):
still want to see directed tests for making sure the contents are as we expect in the presence of flushes

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -70,19 +70,28 @@ class QueueGetsFlushedTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int
...
     inCnt.inc()
-    inCnt.inc()
+    currQCnt := currQCnt + 1.U //counts how many items have been enqueued
   }
   when(q.io.deq.fire()) {
-    outCnt.inc()
+    //ensure that what comes out is what comes in
+    assert(flushRegister === false.B)
+    assert(currQCnt <= queueDepth.U)
+    assert(elems(outCnt /*- 1.U*/) === q.io.deq.bits)
```
debs-sifive, line 79:
```suggestion
    assert(elems(outCnt) === q.io.deq.bits)
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -70,19 +70,28 @@ class QueueGetsFlushedTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int
...
-    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue because queue is empty") 
-    assert(q.io.enq.ready, "Expected enqueue to be ready when flush is high because queue is empty")
+  when(flush) {
+    outCnt := outCnt + Mux(q.io.enq.fire(), (currQCnt + 1.U), currQCnt)
+    currQCnt := 0.U //resets the number of items currently inside queue
+  }
+  when(flushRegister) { //Internal signal maybe_full is a register so some signals update on the next cycle
+    //check that queue gets flushed when queue is full
     assert(q.io.count === 0.U)
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush")
```
debs-sifive, line 92: maybe it would make more sense to move this to the `when(flush)` block of logic? I think we are expected to be able to dequeue on the same cycle that flush goes high.

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -100,10 +109,54 @@ class EmptyFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: I
...
+  when(q.io.enq.fire()) {
+    inCnt.inc()
+  }
+  when(q.io.deq.fire()) {
+    assert(elems(outCnt) === q.io.deq.bits)
+    outCnt := outCnt + 1.U
+  }
+  when(flushRegister) {
+    //check that queue gets flushed at the beginning with no elements
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue because it should be empty")
```
debs-sifive, line 136: same as above

src/test/scala/chiselTests/QueueFlushSpec.scala
debs-sifive: i think this comment is still wrong
```suggestion
  //testing a flush when queue is full
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -145,32 +199,95 @@ class EmptyFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: I
...
+  }
   when(flushRegister) {
-    currDepthCnt.reset() //resets the number of items currently inside queue
+    outCnt := outCnt + currQCnt
+    currQCnt := 0.U //resets the number of items currently inside queue
     //check that queue gets flushed when queue is full
+    assert(q.io.deq.fire() === false.B)
     assert(q.io.count === 0.U)
-    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush") 
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush")
```
debs-sifive, line 288: same as above

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,364 @@
...
+          new ThingsPassThroughFlushQueueTester(se._2, depth, se._1, tap, isSync)
+        }
+      }
+    }
+  }
+  property("Queue should flush when requested") {
+    forAll(vecSizes, safeUIntN(20), Gen.choose(0, 15), Gen.oneOf(true, false)) { (depth, se, tap, isSync) =>
+      whenever(se._1 >= 1 && depth >= 1 && se._2.nonEmpty) {
+        assertTesterPasses {
+          new QueueGetsFlushedTester(List(0, 2, 2, 4, 4, 6, 0, 5, 1, 0, 1, 4, 2, 0, 4, 2, 7, 2, 2, 7), 2, 3, 10, true)
```
debs-sifive, line 314: probably want to change this back to taking in generated arguments

anniej-sifive: Oh yes oops I totally missed that. Thanks!

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,364 @@
...
+  }
+  when(flush) {
+    outCnt := outCnt + Mux(q.io.enq.fire(), (currQCnt + 1.U), currQCnt)
+    currQCnt := 0.U //resets the number of items currently inside queue
+  }
+  when(flushRegister) { //Internal signal maybe_full is a register so some signals update on the next cycle
+    //check that queue gets flushed when queue is full
+    assert(q.io.count === 0.U)
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush")
+    assert(q.io.enq.ready, "Expected enqueue to be ready when flush is high because queue should be empty")
```
mwachs5, line 93:
```suggestion
    assert(q.io.enq.ready, "Expected enqueue to be ready when flush was asserted the previous cycle because queue should be empty")
```

src/test/scala/chiselTests/QueueFlushSpec.scala
mwachs5: i think this check is still good though, it's basically ensuring the queue is empty. I would just change the message, and add another check above

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,364 @@
...
+    }
+  }
+  when(flush) {
+    outCnt := outCnt + Mux(q.io.enq.fire(), (currQCnt + 1.U), currQCnt)
+    currQCnt := 0.U //resets the number of items currently inside queue
+  }
+  when(flushRegister) { //Internal signal maybe_full is a register so some signals update on the next cycle
+    //check that queue gets flushed when queue is full
+    assert(q.io.count === 0.U)
+    assert(!q.io.deq.valid, "Expected to not be able to dequeue a flushed queue if it had elements prior to the flush")
```
mwachs5, line 92:
```suggestion
    assert(!q.io.deq.valid, "Expected to not be able to dequeue from a queue that was flushed the previous cycle (it should be empty)")
```

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,364 @@
...
+    assert(currQCnt <= queueDepth.U)
+    assert(elems(outCnt /*- 1.U*/) === q.io.deq.bits)
+    outCnt := outCnt + 1.U
+    when (currQCnt > 0.U) {
+      currQCnt := currQCnt - 1.U
+    }
+  }
+  when(flush) {
+    outCnt := outCnt + Mux(q.io.enq.fire(), (currQCnt + 1.U), currQCnt)
+    currQCnt := 0.U //resets the number of items currently inside queue
```
mwachs5, line 87:
```suggestion
    currQCnt := 0.U //resets the number of items currently inside queue
    assert(currQCount === 0.U || q.io.deq.valid)
```

or something. actually this doesn't even need to be gated on the `flush`. basically i want a check that flush does not impact queue or enqueue behavior *while* it is asserted.

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: this is no longer the case, i think we can resolve this comment

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: i think we can resolve this comment

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+
+/** Test elements can be enqueued and dequeued
+ * 
+  * @param elements The sequence of elements used in the queue
+  * @param queueDepth The max number of entries in the queue
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class ThingsPassThroughFlushQueueTester(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
```
mwachs5, line 19:
```suggestion

/** Test elements can be enqueued and dequeued
 * 
  * @param elements The sequence of elements used in the queue
  * @param queueDepth The max number of entries in the queue
  * @param bitWidth Integer size of the data type used in the queue
  * @param tap Integer tap('seed') for the LFSR
  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
  */
abstract class FlushQueueTesterBase(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true)) 
  ...
}


class ThingsPassThroughFlushQueueTester(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends FlushQueueTesterBase {
```

you could do a common test base class...

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+}
+/** Test elements can be enqueued and dequeued
+ * 
+  * @param elements The sequence of elements used in the queue
+  * @param queueDepth The max number of entries in the queue
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class ThingsPassThroughFlushQueueTester(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
```
mwachs5, line 57: is there a reason you did not make this one extend the base queue test?

anniej-sifive: Yes this test case does not test flush capabilities, it tests regular enqueue and dequeue functionality

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+abstract class FlushQueueTesterBase(elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends BasicTester {
+  val q = Module(new Queue(UInt(bitWidth.W), queueDepth, hasFlush = true))
+  val elems = VecInit(elements.map(_.U))
+  val inCnt = Counter(elements.length + 1)
+  val outCnt = RegInit(0.U(log2Ceil(elements.length).W))
+  val currQCnt = RegInit(0.U(log2Ceil(5).W))
+
+  val flush: Bool // Wire(Bool())//LFSR(16)((tap + 3) % 16)  //testing a flush when flush is called randomly
```
mwachs5, line 26: can you remove this commented code?

mwachs5: in fact, if `flush` is not used in this base class at all, let's just remove it.

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  val flushRegister = RegInit(false.B)
```
mwachs5, line 27: or, you could keep `flush` and make this `RegNext(flush)` <--- not quite sure on that code

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class EmptyFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends FlushQueueTesterBase(elements, queueDepth, bitWidth, tap, useSyncReadMem) {
+  val cycleCounter = Counter(elements.length + 1)
+  cycleCounter.inc() //counts every cycle
+
+  //testing a flush when queue is empty
+  val flush = (cycleCounter.value === 0.U && inCnt.value === 0.U) //flushed only before anything is enqueued  
+  q.io.flush.get := flush
+  flushRegister := flush
```
mwachs5, line 127: i think you can pull these two up into the parent class, if you make:

`val flush := false.B` and then use last-connect assign in the subclasses:

`flush := ...`

src/test/scala/chiselTests/QueueFlushSpec.scala
mwachs5: (I think I prefer keeping `flush` here and pulling MORE into the parent class)

src/test/scala/chiselTests/QueueFlushSpec.scala
mwachs5: (read on as I decided removing `flush` here isn't the best approach)

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class EnqueueEmptyFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends FlushQueueTesterBase(elements, queueDepth, bitWidth, tap, useSyncReadMem) {
+  val cycleCounter = Counter(elements.length + 1)
+  val outCounter = Counter(elements.length + 1)
+
+  //testing an enqueue during a flush
+  val flush = (cycleCounter.value === 0.U && inCnt.value === 0.U) //flushed only before anything is enqueued  
+  q.io.flush.get := flush
+  flushRegister := flush
```
mwachs5, line 152: this can go into parent class

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  * @param bitWidth Integer size of the data type used in the queue
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class FullQueueFlushEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends FlushQueueTesterBase(elements, queueDepth, bitWidth, tap, useSyncReadMem) {
+
+  //testing a flush when queue is full
+  val flush = (currQCnt === queueDepth.U)
+  q.io.flush.get := flush
+  flushRegister := flush
```
mwachs5, line 176: this can go into parent class

src/test/scala/chiselTests/QueueFlushSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  * @param tap Integer tap('seed') for the LFSR
+  * @param useSyncReadMem True uses SyncReadMem instead of Mem as an internal memory element
+  */
+class DequeueFullQueueEdgecaseTester (elements: Seq[Int], queueDepth: Int, bitWidth: Int, tap: Int, useSyncReadMem: Boolean) extends FlushQueueTesterBase(elements, queueDepth, bitWidth, tap, useSyncReadMem) {
+  //Queue should be able to dequeue when queue is not empty and flush is high
+
+  //testing a flush when dequeue is called
+  val flush = currQCnt === (queueDepth/2).U
+  q.io.flush.get := flush
+  flushRegister := flush
```
mwachs5, line 208: this can go into parent

mwachs5:
@aswaterman , @sequencer can you clear your requested changes?

mwachs5 (approved):

aswaterman (approved):

PR #2043, opened 2021-07-13
Speed up CI

sequencer:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         system: ["ubuntu-20.04"]
-        jvm: ["graalvm@21.1.0"]
+        jvm: ["graalvm@21.1.0", "adopt@1.8"]
         scala: ["2.13.6", "2.12.13"]
         verilator: ["4.204"]
         z3: ["4.8.10"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,8 +14,8 @@ jobs:
     name: ci
     strategy:
       matrix:
-        system: ["ubuntu-20.04", "macos-latest"]
-        jvm: ["adopt@1.8", "graalvm@21.1.0"]
+        system: ["ubuntu-20.04"]
+        jvm: ["graalvm@21.1.0"]
         scala: ["2.13.6", "2.12.13"]
         verilator: ["4.204"]
         z3: ["4.8.10"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,8 +14,8 @@ jobs:
     name: ci
     strategy:
       matrix:
-        system: ["ubuntu-20.04"]
-        jvm: ["adopt@1.8"]
+        system: ["ubuntu-20.04", "macos-latest"]
+        jvm: ["adopt@1.8", "graalvm@21.1.0"]
         scala: ["2.13.6", "2.12.13"]
         verilator: ["4.204"]
         z3: ["4.8.10"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,8 +108,16 @@ jobs:
       - name: Documentation (Scala 2.12 only)
         if: matrix.scala == '2.12.13'
         run: sbt ++${{ matrix.scala }} docs/mdoc
-      - name: Test
+      - name: Test with treadle (PR only)
+        if: github.event_name == 'pull_request'
         run: sbt ++${{ matrix.scala }} test noPluginTests/test
+        env:
+          TEST_BACKEND: treadle
+      - name: Test (Push only)
+        if: github.event_name == 'push'
+        run: sbt ++${{ matrix.scala }} test noPluginTests/test
+        env:
+          TEST_BACKEND: verilator
       - name: Binary compatibility
         run: sbt ++${{ matrix.scala }} mimaReportBinaryIssues
 
--- a/src/test/scala/chiselTests/testers/VerilatorBackend.scala
+++ b/src/test/scala/chiselTests/testers/VerilatorBackend.scala
@@ -4,12 +4,12 @@ package chiselTests.testers
 
 import chisel3.stage.ChiselGeneratorAnnotation
 import chisel3.stage.phases.{Convert, Emitter}
-import chisel3.testers.TesterDriver._
-import chiselTests.testers.TesterDriver.{AddImplicitTesterDirectory, finishWrapper}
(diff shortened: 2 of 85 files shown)


sequencer:
Reopen #2036 due to CI issue.

sequencer:
Test in Mac failed with verilator/verilator#3031
So just gave up Mac CI :(

.github/workflows/test.yml
```diff
@@ -83,8 +83,16 @@ jobs:
...
+      - name: Test with treadle (PR only)
+        if: github.event_name == 'pull_request'
         run: sbt ++${{ matrix.scala }} test noPluginTests/test
+        env:
+          TEST_BACKEND: treadle
+      - name: Test (Push only)
+        if: github.event_name == 'push'
+        run: sbt ++${{ matrix.scala }} test noPluginTests/test
+        env:
+          TEST_BACKEND: verilator
```
jackkoenig, line 95: Instead of having two tests guarded by an `if`, can we instead just set the environment variable in bash? I don't like that this could result in different tests being run for PR vs. push if someone updates one but not the other by accident. Also this suppresses tests for all events but PR and push and that's not right, you also want it on tagging and there might be other events too.

Something like:
```suggestion
      - name: Test
        run:
          - if [ $GITHUB_EVENT_NAME = "pull_request" ]; then export TEST_BACKEND="treadle"; else export TEST_BACKEND="verilator"; fi
          - sbt ++${{ matrix.scala }} test noPluginTests/test
```

or a shorter line:

```suggestion
      - name: Test
        run:
          - [ $GITHUB_EVENT_NAME = "pull_request" ] && TB="treadle" || TB="verilator"
          - export TEST_BAKCNED=$TB
          - sbt ++${{ matrix.scala }} test noPluginTests/test
```

src/test/scala/chiselTests/testers/VerilatorBackend.scala
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests.testers
+
+import chisel3.stage.ChiselGeneratorAnnotation
+import chisel3.stage.phases.{Convert, Emitter}
+import chiselTests.testers.TesterDriver.{AddImplicitTesterDirectory, cppToExe, executeExpectingSuccess, finishWrapper, verilogToCpp}
+import firrtl.AnnotationSeq
+import firrtl.options.{Dependency, PhaseManager, TargetDirAnnotation}
+import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlStage}
```
jackkoenig, line 10: Can you put this in `src/test/scala/chisel3/testers` instead of in package `chiselTests`? By putting stuff in package `chiselTest` you're also moving `TestUtils` which I don't think you should.

src/main/scala/chisel3/testers/BasicTester.scala
```diff
@@ -9,6 +9,7 @@ import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.SourceInfo
 
+@deprecated("This is not a User API, will be removed in Chisel 3.6", "3.5")
 class BasicTester extends Module() {
   // The testbench has no IOs, rather it should communicate using printf, assert, and stop.
   val io = IO(new Bundle() {})
```
jackkoenig, line 15: My main concern with deprecating this is it is the only API exposing `stop()` (which is I guess the real problem). A user cannot replicate the behavior of this unless we add `stop`. 

Also we should point people to chiseltest rather than just saying "This is not a User API"

src/test/scala/chiselTests/testers/VerilatorBackend.scala
sequencer: > Can you put this (and the other files being migrated into testing) in src/test/scala/chisel3/testers instead of in package chiselTests?

Yes, bug wondering why?

> By putting stuff in package chiselTest you're also moving TestUtils which I don't think you should.

I kept and deprecated original `TestUtils`, which I think this doesn't break anything?

src/main/scala/chisel3/testers/BasicTester.scala
sequencer: > A user cannot replicate the behavior of this unless we add stop.

`chisel3.stop` was a user API, originally:
```
  def stop()(implicit sourceInfo: SourceInfo) {
    // TODO: rewrite this using library-style SourceInfo passing.
    when (!reset.asBool) {
      pushCommand(Stop(sourceInfo, clock.ref, 0))
    }
  }
```
I think this is `chisel3.stop` too?

> Also we should point people to chiseltest rather than just saying "This is not a User API"

Yes, I should change the deprecation wording.

src/main/scala/chisel3/testers/TesterDriver.scala
```diff
@@ -13,6 +13,7 @@ import firrtl.options.{Dependency, Phase, PhaseManager, TargetDirAnnotation, Uns
 import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlStage}
 import firrtl.transforms.BlackBoxSourceHelper.writeResourceToDirectory
 
+@deprecated("This is not a User API, will be removed in Chisel 3.6", "3.5")
```
jackkoenig, line 16: Let's do these deprecations separately. These aren't great APIs and they're not commonly used but they _are_ used and thus we need a migration path if we're going to plan to remove them.

So let's not do the deprecation here, but get this PR merged.

You could just copy these things into `src/test/scala` and modify there, but alternatively you could factor out the functionality into a package private trait that `TesterDriver` extends and then do the same thing in `src/test/scala/chisel3` and provide the modified functionality that way.

sequencer:
> https://github.com/chipsalliance/chisel3/pull/2043/checks?check_run_id=3358132535#step:13:5720

WTF here?

sequencer:
Wed, 18 Aug 2021 08:54:43 GMT
[info] ImplicitModule directly in a RawModule in an ImplicitModule
6626
Wed, 18 Aug 2021 08:54:43 GMT
Elaborating design...
6627
Wed, 18 Aug 2021 08:54:43 GMT
[info] - should fail
6628
Wed, 18 Aug 2021 09:40:57 GMT
[info] - Infering widths on huge Vecs should not cause a stack overflow
6629
Wed, 18 Aug 2021 09:40:57 GMT
Elaborating design...
6630
Wed, 18 Aug 2021 09:40:57 GMT
Done elaborating.

jackkoenig:
Replaced by https://github.com/chipsalliance/chisel3/pull/2341

PR #2046, opened 2021-07-21
Bundles can no longer be instantiated with bound hardware

### LINKED ISSUE #1894: Autoclonetype 2 supports bound hardware as fields of Bundles
h/t @johnsbrew for reporting this in https://github.com/chipsalliance/chisel3/issues/1890

**Type of issue**: bug report

**Other information**
 

See https://scastie.scala-lang.org/DOeXfeVUQWiRcnmeB0JstQ

The following works
```scala
class MyBundle(val foo: UInt) extends Bundle

  val in  = IO(Input(new MyBundle(123.U))) // This should error
  val out = IO(Output(new MyBundle(UInt(8.W))))

  out := in
}
```

This should give an error

jared-barocsi:
### Contributor Checklist

#### Type of Improvement

Bug fix

#### API Impact

Changes existing API: Code previously taking advantage of `Bundle`s with bound hardware elements will crash.

No impact

#### Desired Merge Strategy

Squash and merge

#### Release Notes
`Bundle`s can no longer be instantiated with value bindings or bound hardware. Fixes #1894

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -963,6 +963,9 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     for (m <- getPublicFields(classOf[Bundle])) {
       getBundleField(m) match {
         case Some(d: Data) =>
+          if (!d.binding.isEmpty)
+            throwException(s"Bundle can only take type elements (got value binding on '${m.getName()}')")
+
           if (nameMap contains m.getName) {
             require(nameMap(m.getName) eq d)
           } else {
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -963,8 +963,9 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     for (m <- getPublicFields(classOf[Bundle])) {
       getBundleField(m) match {
         case Some(d: Data) =>
-          if (!d.binding.isEmpty)
-            throwException(s"Bundle can only take type elements (got value binding on '${m.getName()}')")
+          if (d.binding.nonEmpty) {
+            throwException(s"Bundle can only take type elements (got value binding '$d' on '${m.getName()}')")
+          }
 
           if (nameMap contains m.getName) {
             require(nameMap(m.getName) eq d)
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -963,9 +963,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     for (m <- getPublicFields(classOf[Bundle])) {
       getBundleField(m) match {
         case Some(d: Data) =>
-          if (d.binding.nonEmpty) {
-            throwException(s"Bundle can only take type elements (got value binding '$d' on '${m.getName()}')")
-          }
+          requireIsChiselType(d)
 
           if (nameMap contains m.getName) {
             require(nameMap(m.getName) eq d)
--- a/src/test/scala/chiselTests/BundleSpec.scala
+++ b/src/test/scala/chiselTests/BundleSpec.scala
@@ -129,6 +129,17 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils with Utils {
     }).getMessage should include("aliased fields")
   }
 
+  "Bundles" should "not have bound hardware" in {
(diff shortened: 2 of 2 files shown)


jackkoenig:
I've marked this `API Modification` because (as the removed test indicates), thing that used to "work" no longer do. That being said, I would argue this was an API hole, not a feature, since only very specific things worked and otherwise you usually got an error (rebinding if you passed such a Bundle to a binding op).

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -142,7 +153,11 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils with Utils {
...
 
-  "Bound Data" should "have priority in setting ref over unbound Data" in {
+  /**
+   * Broken in Chisel 3.5
+   * Bundles should not be able to be instantiated using any kind of bound hardware.
+   */
+  /*"Bound Data" should "have priority in setting ref over unbound Data" in {
     class MyModule extends RawModule {
       val foo = IO(new Bundle {
         val x = Output(UInt(8.W))
```
jackkoenig, line 163: Better to delete the test than comment it out

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -963,6 +963,9 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     for (m <- getPublicFields(classOf[Bundle])) {
       getBundleField(m) match {
         case Some(d: Data) =>
+          if (!d.binding.isEmpty)
+            throwException(s"Bundle can only take type elements (got value binding on '${m.getName()}')")
+
```
jackkoenig, line 968:
```suggestion
          if (d.binding.nonEmpty) {
            throwException(s"Bundle can only take type elements (got value binding on '${m.getName()}')")
          }

```
Very much a nit but generally prefer `.nonEmpty` to `!_.isEmpty`, also, best to have the curly braces to avoid issues like the "goto fail" bug.

jackkoenig (commented):
Generally looks good, a couple of minor suggestions.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -963,6 +963,10 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     for (m <- getPublicFields(classOf[Bundle])) {
       getBundleField(m) match {
         case Some(d: Data) =>
+          if (d.binding.nonEmpty) {
+            throwException(s"Bundle can only take type elements (got value binding '$d' on '${m.getName()}')")
+          }
```
sequencer, line 968: Just another idea, maybe `requireIsChiselType` can catch this?

jackkoenig: I think this suggestion conveys the intent a little more clearly.

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -129,6 +129,17 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils with Utils {
...
+    (the[ChiselException] thrownBy extractCause[ChiselException] {
+      ChiselStage.elaborate { new Module {
+        class MyBundle(val foo: UInt) extends Bundle
+        val in  = IO(Input(new MyBundle(123.U))) // This should error: value passed in instead of type
+        val out = IO(Output(new MyBundle(UInt(8.W))))
+
+        out := in
+      } }
+    }).getMessage should include("got value binding")
+  }
```
jackkoenig, line 142: It's possible (but I'm not certain), that children of unbound aggregates still get their binding set to 
It'd be good to add 1 extra test  for a bound aggregate getting passed
```suggestion
  "Bundles" should "not have bound hardware" in {
    (the[ChiselException] thrownBy extractCause[ChiselException] {
      ChiselStage.elaborate { new Module {
        class MyBundle(val foo: UInt) extends Bundle
        val in  = IO(Input(new MyBundle(123.U))) // This should error: value passed in instead of type
        val out = IO(Output(new MyBundle(UInt(8.W))))

        out := in
      } }
    }).getMessage should include("got value binding")
    (the[ChiselException] thrownBy extractCause[ChiselException] {
      ChiselStage.elaborate { new Module {
        class MyBundle(val foo: UInt) extends Bundle
        val out = IO(Output(Vec(2, UInt(8.W))
        val in  = IO(Input(new MyBundle(out(0))) // This should error: value passed in instead of type

        out := in
      } }
    }).getMessage should include("got value binding")
  }
```

jackkoenig (approved):

Issue #1893, opened 2021-04-29
Autoclonetype 2 crashes on Bundles with higher-kinded type parameters


jackkoenig:
h/t to @johnsbrew for reporting in https://github.com/chipsalliance/chisel3/issues/1890

**Type of issue**: bug report

**Impact**: no functional change

**Development Phase**: request

**Other information**
 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

I've reproduced the bug, working 3.4.3 with autoclonetype2 disabled (https://scastie.scala-lang.org/AHe8v359S0iHca0O7yhWlQ), failing 3.4.3 with autoclonetype enabled (https://scastie.scala-lang.org/vGRFbOS2TyuUIeNpC1HQOg).

**What is the current behavior?**
Using higher-kinded type parameters causes a type error when using autoclonetype2
```scala
class DataGen[T <: Data](gen: T) {
  def newType: T = gen.cloneType
}

class MyBundle[A <: Data, B <: DataGen[A]](gen: B) extends Bundle {
  val foo = gen.newType
  
  //override def cloneType = new MyBundle[A, B](gen).asInstanceOf[this.type]
}
```
This causes
```
scala.reflect.internal.Types$TypeError: inferred type arguments [Nothing,B] do not conform to class MyBundle's type parameter bounds [A <: chisel3.Data,B <: Playground.DataGen[A]]
```

**What is the expected behavior?**

It should definitely not crash, it should probably work correctly. If not possible, then it should issue a warning or a more precise error.

PR #2052, opened 2021-07-23
Remove all Bundle cloneTypes and chiselRuntimeDeprecate its use

mwachs5:
TODO: 
- [x] Fix docs "a note on cloneType
- [x] Add a test that the deprecation warning actually does stuff with/without the plugin?
- [x] Fix the runtime deprecation which I apparently reverted

### Contributor Checklist

#### Type of Improvement

- code cleanup 

#### API Impact

This runtime-deprecates the cloneType method used by runtime reflection.
A user will get a runtime deprecation warning if they do not define their
own cloneType method on a Bundle that can't be inferred (e.g. is using private vals in the generation) without the plugin.
If you ARE using the plugin, the user gets an error if they *do* implement an `override def cloneType`.

No change in generated verilog

#### Desired Merge Strategy

#### Release Notes

All usages of Bundle's overriding their own def cloneType are removed from chisel3 code. It is a runtime deprecation warning to rely on runtime reflection without the plugin. It is a compile error to override def cloneType for a Bundle when using the plugin.

- [ ] Did you review?


src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -10,25 +10,26 @@ trait BundleSpecUtils {
...
-    override def cloneType: this.type = (new BundleFoo).asInstanceOf[this.type]
   }
   class BundleBar extends Bundle {
     val bar = UInt(16.W)
-    override def cloneType: this.type = (new BundleBar).asInstanceOf[this.type]
+  }
+
+  class BundleBaz(w: Int) extends Bundle {
+    val baz = UInt(w.W)
+    // This is runtime deprecated, which we test below
```
mwachs5, line 27:
```suggestion
    // This is currently a compiler warning when using the plugin, which we test below.
    // It will be made an error in future releases.
```

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -154,4 +155,16 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils with Utils {
     }
     ChiselStage.emitChirrtl(new MyModule)
   }
+
+  // This tests the interaction of override def cloneType and the plugin.
+  // As of now, this is only a warning, which we can't really check with scalatest.
```
mwachs5, line 161:
```suggestion
  // As of now, this is only a compiler warning, which we can't really check with scalatest.
```

sequencer:
Seem it is possible to run Chisel 3.5 without autoclonetype2?
Similar to namepulugin, I think we should add deprecation in 3.4 and forbid in 3.5? 
if user forgot to add plugin, we should indicate them to change their build script.

docs/src/explanations/bundles-and-vecs.md
```diff
@@ -128,48 +128,17 @@ class ModuleProgrammaticMixedVec(x: Int, y: Int) extends Module {
 }
 ```
 
-### A note on `cloneType`
+### A note on `cloneType` (For Chisel < 3.5)
```
sequencer, line 131: For something that autoclonetype cannot handle like #1893, should we provide another API to user to ask them implement this on their own? I think a backup API should exist for some corner cases.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1046,6 +1046,10 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     */
   protected def _cloneTypeImpl: Bundle = {
     assert(Builder.allowReflectiveAutoCloneType, "reflective autoclonetype is disallowed, this should only happen in testing")
+    Builder.deprecated(
+      "The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  " +
+      "Use the Compiler Plugin to infer cloneType"
+    )
```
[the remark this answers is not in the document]
mwachs5: I thought that sort of deprecation pattern was discouraged -- shouldnt we only be deprecating on major releases, then removing in next major release?

It seems like in 3.5 we are perhaps deprecating running without plugin. then in 3.6 we could remove. I don't think we can just suddenly deprecate something that major on 3.4.4 since users might go from 3.4 to 3.5 directly

docs/src/explanations/bundles-and-vecs.md
mwachs5: That's a good question. @jackkoenig , what do you think? Do we anticipate legitimate uses of needing to write your own `def cloneType` in the short and/or long terM?

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -69,11 +69,16 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
         case con: DefDef if con.symbol.isPrimaryConstructor =>
           primaryConstructor = Some(con)
         case d: DefDef if isNullaryMethodNamed("_cloneTypeImpl", d) =>
-          val msg = "Users cannot override _cloneTypeImpl. Let the compiler plugin generate it. If you must, override cloneType instead."
+          val msg = "Users cannot override _cloneTypeImpl. Let the compiler plugin generate it."
```
mwachs5, line 72: are there still "If you must..." scenarios?

docs/src/explanations/bundles-and-vecs.md
jackkoenig: #1893 aside (which we should just fix), you need to write your own `.cloneType` for `Records` as well so the note is still needed but maybe with a warning that this _shouldn't_ be needed except in rare cases.

core/src/main/scala/chisel3/Aggregate.scala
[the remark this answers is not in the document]
jackkoenig: In dev today (see notes https://github.com/chipsalliance/chisel3/discussions/2055) we discussed a change to deprecation policy that deprecations on minor versions are now the standard with removal in the subsequent major version. Our reasoning is that this enables better development practice, cleaner code, and more agility.

A basic problem we've had for a long time is that we often deprecate and never remove, and sometimes even deprecate things without replacement. The new development flow is that, if an API is to be changed, it is deleted on master and backported to he latest major version. This deletion on master will keep the main codebase cleaner, while also requiring the developer to update all tests *thus ensuring we know what the code migration path looks like and that it is appropriate to deprecate the API*.

The backport will then fail (because it won't be binary compatible), and the developer is then required to deprecate the API on the backport. This has a massive benefit that the internal hacks that are often necessary to maintain some deprecated APIs will only have to be made on the backport branch and will never exist on master.

Finally, it's true that people may sometimes skip minor versions. It will become part of our release notes and announcement notes _which minor version users need to bump to before bumping to the major release_. This requires a bit of additional discipline but I believe the benefits to be worth it.

docs/src/explanations/bundles-and-vecs.md
mwachs5: right, but say we say you only need it in rare cases, and then the plugin makes it an error to havea. cloneType method... how do we let the plugin allow/recognize the "rare cases"? Or is it that the plugin should not hit its error scenario if it doesn't recognize the need for the plugin in a certain case (meaning you need your own `def cloneType` for the rare case)?

jackkoenig: My current thinking is that those "rare cases" are `Records` and bugs. I can't think of any case when one should _want_ to implement their own `cloneType` method, so maybe we just make it an error and fix the bugs that come up. If some use case comes up in the future we can provide an out (perhaps a trait to mixin or a Scala annotation), but I don't think we should do that work unless necessary.

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -8,32 +8,28 @@ import chisel3.testers.BasicTester
 
 trait BundleSpecUtils {
   class BundleFooBar extends Bundle {
-    val foo = UInt(16.W)
-    val bar = UInt(16.W)
+    val foo                           = UInt(16.W)
+    val bar                           = UInt(16.W)
     override def cloneType: this.type = (new BundleFooBar).asInstanceOf[this.type]
```
jackkoenig:
```suggestion
    val foo = UInt(16.W)
    val bar = UInt(16.W)
```

jackkoenig (commented):
I want to merge this, but I think we need https://github.com/chipsalliance/chisel3/issues/1893 fixed before we can force people to use the compiler plugin.

We also obviously need some more docs on that, but that can be another PR before 3.5 release.

src/test/scala/chiselTests/BundleSpec.scala
mwachs5: oh noes i turned on my code formatter will clean this up.

That was there and I nicely got a compiler error of `msg` :)

mwachs5:
I updated the PR text and website documentation to reflect that this PR is now an *error* to add a def cloneType when using the compiler plugin. Not sure the strategy for backporting the release notes to the release branches, but I can clean it up before it goes in. Understood this is blocked on the higher-kinded types bug.

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
mwachs5: I don't think so. This if you must is gone

mwachs5:
So i (or maybe @jackkoenig ) marked this as milestone 1.5, but this needs a deprecation backport. Maybe we need some new mergify rules to handle this case?

mwachs5:
@mergifyio backport 3.4.x

Opening a backport PR and will adjust the compiler plugin behavior (and commented-out test) to be a warning, not an error

jackkoenig (approved):

PR #1518, opened 2020-07-20
Check whether signals escape their when scopes

albert-magyar:
**Related issue**: #1409 freechipsproject/firrtl#1505 #1512 
**Type of change**: improvement
**Impact**: freechipsproject/firrtl#1528 will make code that violates when scopes illegal. This will help enforce this at the Chisel level, but some code (like the old counter implementation from before #1408) will now fail to elaborate.
**Development Phase**:  implementation

**Release Notes**:
Previous versions of Chisel and FIRRTL did not adequately enforce the scoping of declarations to `when` blocks. Using a reference to a signal that has escaped a closed scope will result in an elaboration-time error.

PR #2064, opened 2021-08-06
Error on MPort Escaping When, Fix SyncReadMem.read(addr, en) to not do this

seldridge:
Extend the stricter checking introduced in #1518 by removing the special casing around mports escaping when scopes.  Now, Chisel code that tries to do this will result in an error.

There was one known usage of this in the Chisel standard library: `SyncReadMem.read(addr, en)`.  Change this to use a dummy wire instead.

This attempts to fix code generation where an `mport` is used directly in a later block with a different enable.  An example of this is shown here: https://scastie.scala-lang.org/seldridge/pt0ARTQtTuWkTJfoLfB05g/37.

With this patch, the `Bar` module will now produce the following CHIRRTL:

```scala
circuit Bar :
  module Bar :
    input clock : Clock
    input reset : UInt<1>
    input addr : UInt<3>
    input cond : UInt<1>
    output out : UInt<1>

    smem mem : UInt<1> [8] 
    wire _r_WIRE : UInt 
    wire _r_WIRE_1 : UInt<1> 
    _r_WIRE is invalid 
    _r_WIRE_1 is invalid 
    when cond : 
      _r_WIRE <= addr 
      node _r_T = or(_r_WIRE, UInt<3>("h0")) 
      node _r_T_1 = bits(_r_T, 2, 0) 
      read mport r_MPORT = mem[_r_T_1], clock 
      _r_WIRE_1 <= r_MPORT 
    reg r_REG : UInt<1>, clock with :
      reset => (UInt<1>("h0"), r_REG) 
    r_REG <= cond 
    reg r : UInt<1>, clock with :
      reset => (UInt<1>("h0"), r) 
    when r_REG : 
      r <= _r_WIRE_1 
    out <= r
```

### DIFF
--- a/core/src/main/scala/chisel3/Mem.scala
+++ b/core/src/main/scala/chisel3/Mem.scala
@@ -209,12 +209,13 @@ sealed class SyncReadMem[T <: Data] private (t: T, n: BigInt, val readUnderWrite
   /** @group SourceInfoTransformMacro */
   def do_read(addr: UInt, enable: Bool)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
     val a = Wire(UInt())
+    val port = Wire(t)
     a := DontCare
-    var port: Option[T] = None
+    port := DontCare
     when (enable) {
       a := addr
-      port = Some(read(a))
+      port := read(a)
     }
-    port.get
+    port
   }
 }
--- a/core/src/main/scala/chisel3/internal/MonoConnect.scala
+++ b/core/src/main/scala/chisel3/internal/MonoConnect.scala
@@ -64,7 +64,6 @@ private[chisel3] object MonoConnect {
 
   def checkWhenVisibility(x: Data): Boolean = {
     x.topBinding match {
-      case mp: MemoryPortBinding => true // TODO (albert-magyar): remove this "bridge" for odd enable logic of current CHIRRTL memories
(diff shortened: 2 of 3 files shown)


seldridge:
Ugh... There's a problem with this. The previous behavior of exfiltrating an _inferred_ memport would result in the ability to infer a write port if this shows up on the LHS of a connect. Changing this to a wire may introduce combinational cycles in existing code that relied on this behavior.

A small example showing this is here: https://scastie.scala-lang.org/seldridge/OHHRA7S6R5KaIFw9bkUfFA/141

seldridge:
There is a fundamental problem with this.  The connect semantics involving assigning the mport to a wire _inside a when block_ results in a lowered structure like:

```scala
foo <= validif(en, mem.r.data) // mem.r.data is the data from a read port
```

While this will be optimized away to `foo <= mem.r.data`, this will be directly lowered out of CHIRRTL to `foo <= mem.r.data`.  Leaving this as `validif` is not the same and will trip combinational cycle detection if the enable is combinationally driven by the memory's data.

There is no clean way out of this (other than likely what has been suggested by others trying to solve this problem).  A smaller improvement would just be to wrap the exfiltrated memory port in a wire and return that to avoid re-use of the mport in different contexts, e.g., to infer a readwrite port.

seldridge:
I'm going to close this as this approach will not work. Changing the emitted CHIRRTL here to support both a port declaration and a port accessor (how CIRCT models this) would be a way to handle this.

PR #2065, opened 2021-08-10
Add multiple dimensions to VecInit fill and iterate 

anniej-sifive:
### Contributor Checklist

#### Type of Improvement

- new feature/API   

#### API Impact

This extends the tabulate and fill functions for VecInit to create multi-dimensional Vecs 
Ex. Vec[Vec[UInt]]

This should not change the backend code generation as it only provides different ways to fill a multi-dimensional Vec.

#### Desired Merge Strategy

#### Release Notes

This PR adds builds upon companion object factory methods for the VecInit class, .fill and .iterate, to include multi-dimensional Vecs.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -590,6 +590,33 @@ object VecInit extends SourceInfoDoc {
   /** @group SourceInfoTransformMacro */
   def do_tabulate[T <: Data](n: Int)(gen: (Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] =
     apply((0 until n).map(i => gen(i)))
+
+  /** Creates a new [[Vec]] of length `n` composed of the result of the given
+   * function applied to an element of data type T.
+   * 
+   * @param n number of elements in the vector
+   * @param gen function that takes in an element T and returns an output 
+   * element of the same type
+   */
+  def fill[T <: Data](n: Int)(gen: => T): Vec[T] = macro VecTransform.fill
+
+  /** @group SourceInfoTransformMacro */
+  def do_fill[T <: Data](n: Int)(gen: => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] =
+    apply(Seq.fill(n)(gen))
+
+  /** Creates a new [[Vec]] of length `n` composed of the result of the given
+   * function applied to an element of data type T.
+   * 
+   * @param start First element in the Vec
+   * @param len Lenth of elements in the Vec
+   * @param f Function that applies the element T from previous index and returns the output 
+   * element to the next index
+   */
+  def iterate[T <: Data](start: T, len: Int)(f: (T) => T): Vec[T] = macro VecTransform.iterate
+  
+  /** @group SourceInfoTransformMacro */
+  def do_iterate[T <: Data](start: T, len: Int)(f: (T) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[T] = 
+    apply(Seq.iterate(start, len)(f)) 
 }
 
 /** A trait for [[Vec]]s containing common hardware generators for collection
@@ -964,7 +991,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
       getBundleField(m) match {
         case Some(d: Data) =>
           requireIsChiselType(d)
-
+          
           if (nameMap contains m.getName) {
             require(nameMap(m.getName) eq d)
           } else {
@@ -1268,3 +1295,4 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     */
   override def toPrintable: Printable = toPrintableHelper(elements.toList.reverse)
 }
+
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -528,7 +528,6 @@ object VecInit extends SourceInfoDoc {
     elts.foreach(requireIsHardware(_, "vec element"))
 
     val vec = Wire(Vec(elts.length, cloneSupertype(elts, "Vec")))
-    val tpe = cloneSupertype(elts, "Vec")
     val op = getConnectOpFromDirectionality(vec.head)
     
     (vec zip elts).foreach{ x => 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -991,7 +991,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
       getBundleField(m) match {
         case Some(d: Data) =>
           requireIsChiselType(d)
-          
+
           if (nameMap contains m.getName) {
             require(nameMap(m.getName) eq d)
           } else {
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -522,6 +522,22 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
 }
 
 object VecInit extends SourceInfoDoc {
+
+  def getConnectOpFromDirectionality[T <: Data](proto: T): (T, T) => Unit = macro VecTransform.get_connect_op
+
+  /** @group SourceInfoTransformMacro */
+  def do_getConnectOpFromDirectionality[T <: Data](proto: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): (T, T) => Unit = proto.direction match {
+    case ActualDirection.Input | ActualDirection.Output | ActualDirection.Unspecified =>
+      // When internal wires are involved, driver / sink must be specified explicitly, otherwise
+      // the system is unable to infer which is driver / sink
+      (x, y) => x := y
+    case ActualDirection.Bidirectional(_) =>
+      // For bidirectional, must issue a bulk connect so subelements are resolved correctly.
+      // Bulk connecting two wires may not succeed because Chisel frontend does not infer
+      // directions.
+      (x, y) => x <> y
+  }
+
   /** Creates a new [[Vec]] composed of elements of the input Seq of [[Data]]
     * nodes.
     *
@@ -547,18 +563,11 @@ object VecInit extends SourceInfoDoc {
     elts.foreach(requireIsHardware(_, "vec element"))
 
     val vec = Wire(Vec(elts.length, cloneSupertype(elts, "Vec")))
-
-    // TODO: try to remove the logic for this mess
-    elts.head.direction match {
-      case ActualDirection.Input | ActualDirection.Output | ActualDirection.Unspecified =>
-        // When internal wires are involved, driver / sink must be specified explicitly, otherwise
-        // the system is unable to infer which is driver / sink
-        (vec zip elts).foreach(x => x._1 := x._2)
-      case ActualDirection.Bidirectional(_) =>
-        // For bidirectional, must issue a bulk connect so subelements are resolved correctly.
-        // Bulk connecting two wires may not succeed because Chisel frontend does not infer
-        // directions.
-        (vec zip elts).foreach(x => x._1 <> x._2)
+    val tpe = cloneSupertype(elts, "Vec")
+    val op = getConnectOpFromDirectionality(vec.head) //elts.head
+    
+    (vec zip elts).foreach{ x => 
+      op(x._1, x._2)
     }
     vec
   }
@@ -600,15 +609,28 @@ object VecInit extends SourceInfoDoc {
     * @param gen function that takes in an Int (the index) and returns a
     * [[Data]] that becomes the output element
     */
-  def tabulate[T <: Data](n: Int, m: Int)(gen: (Int) => T): Vec[Vec[T]] = macro VecTransform.tabulate2D
+  def tabulate[T <: Data](n: Int, m: Int)(gen: (Int, Int) => T): Vec[Vec[T]] = macro VecTransform.tabulate2D
 
   /** @group SourceInfoTransformMacro */
-  def do_tabulate[T <: Data](n: Int, m: Int)(gen: (Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] = {
-    val myVec = Wire(Vec(n, Vec(m, cloneSupertype((0 until m).map(i => gen(i)), "Vec"))))
-    myVec.foreach {
-      _ := VecInit.tabulate(m)(gen)
-    }
-    myVec
+  def do_tabulate[T <: Data](n: Int, m: Int)(gen: (Int, Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] = {    
+    // TODO make this lazy (requires LazyList and cross compilation, beyond the scope of this PR)
+    val elts = Seq.tabulate(n, m)(gen)
+    val flatElts = elts.flatten
+
+    require(flatElts.nonEmpty, "Vec hardware values are not allowed to be empty")
+    flatElts.foreach(requireIsHardware(_, "vec element"))
+  
+    val tpe = cloneSupertype(flatElts, "Vec.tabulate")
+    val myVec = Wire(Vec(n, Vec(m, tpe)))
+    val op = getConnectOpFromDirectionality(myVec.head.head)
+    // We could get fancy here to have reuse across the multiple dimensions, but being more verbose is more clear
+    for(
+        (xs1D, ys1D) <- myVec zip elts;
+        (x, y) <- xs1D zip ys1D
+      ) { 
+        op(x, y) 
+      }
+    myVec 
   }
 
   /** Creates a new 3D [[Vec]] of length `n by m by p` composed of the results of the given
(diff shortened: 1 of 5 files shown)


mwachs5:
Can you update the PR title?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -604,6 +665,66 @@ object VecInit extends SourceInfoDoc {
...
+  /** @group SourceInfoTransformMacro */
+  def do_fill[T <: Data](n: Int, m: Int, p: Int)(gen: => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[Vec[T]]] = {
+    val myVec = Wire(Vec(n, Vec(m, Vec(p, cloneSupertype(Seq.fill(p)(gen), "Vec")))))
+    myVec.foreach { 
+      _ := VecInit.fill(m, p)(gen)
+    }
+    myVec
+  }
+
+  /** Creates a new 4D [[Vec]] of length `n by m by p by q` composed of the result of the given
```
mwachs5, line 707: is there any particular reason you went to 4D, not 3D or 5D? Is this paralleling something else in the codebase and/or scala?

anniej-sifive: Scala seqs can go to 4D, so I think its to follow that pattern (https://www.scala-lang.org/api/2.13.3/scala/collection/SeqFactory.html)

mwachs5: Ah, I See that scala Seq.tabulate and so on have up to 5D versions. But there they don't have different names, they just have different numbers of parameters to distinguish... any reason to not follow that pattern

https://www.scala-lang.org/api/2.13.6/scala/collection/Seq$.html#tabulate[A](n1:Int,n2:Int,n3:Int,n4:Int,n5:Int)(f:(Int,Int,Int,Int,Int)=%3EA):CC[CC[CC[CC[CC[A]]]]@scala.annotation.unchecked.uncheckedVariance]

mwachs5: I guess I might not encourage people to have 5D versions in hardware, and might stop at 3D for chisel... but willing to be convinced otherwise

mwachs5 (commented):
could you add something to the docs / cookbook about the 2d usage, so that it's more discoverable?

core/src/main/scala/chisel3/Aggregate.scala
anniej-sifive: Ah oh I guess I forgot that it went to 5D, @jackkoenig suggested to go to 4D and I assumed thats kinda where it came from... I could add a 5D if you want it?

mwachs5: i guess i'd be more inclined to go lower, not higher, but I'd defer to @jackkoenig on what he thinks is best. I'm not sure how realistic 5-d generated hardware would be / how happy tools like VCS would be about trying to do something with that

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -591,6 +591,67 @@ object VecInit extends SourceInfoDoc {
...
+  def tabulate[T <: Data](n: Int, m: Int)(gen: (Int) => T): Vec[Vec[T]] = macro VecTransform.tabulate2D
+
+  /** @group SourceInfoTransformMacro */
+  def do_tabulate[T <: Data](n: Int, m: Int)(gen: (Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] = {
+    val myVec = Wire(Vec(n, Vec(m, cloneSupertype((0 until m).map(i => gen(i)), "Vec"))))
+    myVec.foreach {
+      _ := VecInit.tabulate(m)(gen)
+    }
+    myVec
+  }
```
jackkoenig, line 612: Using `:=` is actually not correct--`VecInit.tabulate` needs to support bidirectional aggregates and `:=` is a unidirectional connection. Now unfortunately, just switching this implementation to using `<>` also won't work due to the fact that `<>` is commutative and thus it is ambiguous which direction connections should go if both sides of a `<>` are wires (if one side is a port, that side "anchors" the connection making it unambiguous). For example:

[This](https://scastie.scala-lang.org/E8fkK8CKTzKrEBSI4ZprdQ) works (both sides ports)
```scala
  val enq = IO(Vec(4, Decoupled(UInt(8.W))))
  val deq = IO(Flipped(Vec(4, Decoupled(UInt(8.W)))))
  deq <> enq
```

[This](https://scastie.scala-lang.org/YuQsQaVTQ1aHzkxpqEhUvA) works (one side port, one side wire)
```scala
  val enq = IO(Vec(4, Decoupled(UInt(8.W))))
  val deq = IO(Flipped(Vec(4, Decoupled(UInt(8.W)))))
  // Effectively what VecInit does
  val wire = Wire(Vec(4, Decoupled(UInt(8.W))))
  wire <> enq
  
  deq <> wire
```

But [this](https://scastie.scala-lang.org/rEH7I0kIR8WmVH5rZK1yPw) doesn't work (both sides wire):
```scala
  val enq = IO(Vec(4, Decoupled(UInt(8.W))))
  val deq = IO(Flipped(Vec(4, Decoupled(UInt(8.W)))))
  // Effectively what VecInit does
  val wire = Wire(Vec(4, Decoupled(UInt(8.W))))
  wire <> enq
  val wire2 = Wire(Vec(4, Flipped(Decoupled(UInt(8.W)))))
  wire2 <> deq
  
  wire2 <> wire
```

In order to make this higher dimensional `fills` and `tabulates` work, they cannot create a Wire and delegate to a function that also creates a Wire. I think the only way to do this correctly is to create the wire in each `tabulate` and then connect every single element (no delegation).

Also, looking at the fact that the 1D tabulate delegates to `do_apply` which has [this weird logic](https://github.com/chipsalliance/chisel3/blob/713de6b823d8707246935b9e31ed2fbafeaeca32/core/src/main/scala/chisel3/Aggregate.scala#L552), I think we should preserve that functionality as well. I'm envisioning factoring that out into a private higher-order function like so:

```scala
private def getConnectOpFromDirectionality[T <: Data](proto: T): (T, T) => Unit = proto.direction match {
  case ActualDirection.Input | ActualDirection.Output | ActualDirection.Unspecified =>
    // When internal wires are involved, driver / sink must be specified explicitly, otherwise
    // the system is unable to infer which is driver / sink
    (x, y) => x := y
  case ActualDirection.Bidirectional(_) =>
    // For bidirectional, must issue a bulk connect so subelements are resolved correctly.
    // Bulk connecting two wires may not succeed because Chisel frontend does not infer
    // directions.
    (x, y) => x <> y
}
```

Then you could use that for this function as:
```suggestion
  // Also gen should take more Ints, 1 per dimensionality
  def do_tabulate[T <: Data](n: Int, m: Int)(gen: (Int, Int) => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] = {
    // TODO make this lazy (requires LazyList and cross compilation, beyond the scope of this PR)
    val elts = Seq.tabulate(n, m)(gen)
   
    val flatElts = elts.flatten
    
    // These checks are copied from do_apply, perhaps some room for factoring out
    require(flatElts.nonEmpty, "Vec hardware values are not allowed to be empty")
    flatElts.foreach(requireIsHardware(_, "vec element"))
    
    val tpe = cloneSupertype(flatElts, "Vec.tabulate")
    val myVec = Wire(Vec(n, Vec(m, tpe)))
    
    val op = getConnectOpFromDirectionality(tpe)
    // We could get fancy here to have reuse across the multiple dimensions, but being more verbose is more clear
    for {
      (xs1D, ys1D) <- myVec zip elts
      (x, y) <- xs1D zip ys1D
    } do op(x, y)
    myVec
  }
```

I know this is a lot, and this is made complex by the lack of a non-commutative `<>`, but it is what it is.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -604,6 +665,66 @@ object VecInit extends SourceInfoDoc {
...
+  def fill[T <: Data](n: Int, m: Int)(gen: => T): Vec[Vec[T]] = macro VecTransform.fill2D
+
+  /** @group SourceInfoTransformMacro */
+  def do_fill[T <: Data](n: Int, m: Int)(gen: => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] = {
+    val myVec = Wire(Vec(n, Vec(m, cloneSupertype(Seq.fill(m)(gen), "Vec"))))
+    myVec.foreach {
+      _ := VecInit.fill(m)(gen)
+    }
+    myVec
+  }
```
jackkoenig, line 685: Note that `fill` is just a special case of `tabulate` (where you ignore the indices) so this can be implemented as just:

```suggestion
  def do_fill[T <: Data](n: Int, m: Int)(gen: => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Vec[Vec[T]] =
    do_tabulate(n, m)((_, _) => gen)
```

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: I was thinking 4D because it's what `Seq` did in Scala 2.12. I agree that 5D is a bit much and maybe 4D is too much. Maybe let's go to 3D for the time being?

jackkoenig (changes requested):
A bit of a long review because there are some subtleties here. I'd also like to see tests that check for bidirectional support.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: Instead of the nested `for`, we could actually just flatten and iterate:
```scala
for ((x, y) <- myVec.flatten zip flatElts) {
  op(x, y)
}
```
This requires fewer intermediate names and would work basically the same way for the higher dimensional versions (with extra `.flatten`s)

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -557,6 +568,74 @@ object VecInit extends SourceInfoDoc {
...
+    val tpe = cloneSupertype(flatElts, "Vec.tabulate")
+    val myVec = Wire(Vec(n, Vec(m, tpe)))
+    val op = getConnectOpFromDirectionality(myVec.head.head)
+    // We could get fancy here to have reuse across the multiple dimensions, but being more verbose is more clear
+    for(
+        (xs1D, ys1D) <- myVec zip elts;
+        (x, y) <- xs1D zip ys1D
+      ) { 
+        op(x, y) 
+      }
```
jackkoenig, line 600:
```suggestion
    for (
      (xs1D, ys1D) <- myVec zip elts;
      (x, y) <- xs1D zip ys1D
    ) { 
      op(x, y) 
    }
```
This tabbing is a bit weird, try to keep it to 2 spaces whenever something needs to be tabbed

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -557,6 +568,74 @@ object VecInit extends SourceInfoDoc {
...
+    val myVec = Wire(Vec(n, Vec(m, Vec(p, tpe))))
+    val op = getConnectOpFromDirectionality(myVec.head.head.head)
+    
+    for(
+        (xs2D, ys2D) <- myVec zip elts;
+        (xs1D, ys1D) <- xs2D zip ys2D;
+        (x, y) <- xs1D zip ys1D
+      ) { 
+        op(x, y) 
+      }
```
jackkoenig, line 634:
```suggestion
    for (
      (xs2D, ys2D) <- myVec zip elts;
      (xs1D, ys1D) <- xs2D zip ys2D;
      (x, y) <- xs1D zip ys1D
    ) { 
      op(x, y) 
    }
```

docs/src/cookbooks/cookbook.md
```diff
@@ -157,6 +158,39 @@ See the [DataView cookbook](dataview#how-do-i-connect-a-subset-of-bundle-fields)
...
+  val foo = UInt(4.W)
+  val bar = UInt(4.W)
+}
+
+class Foo extends Module {
+  
+  val twoDVec = VecInit.fill(2,3)(5.U)
+  val threeDVec = VecInit.fill(1,2,4)(new MyBundle)
+
+  val indexTiedVec = VecInit.tabulate(2,2){ (x, y) => (x+y).asUInt }
```
jackkoenig, line 178:
```suggestion
  val indexTiedVec = VecInit.tabulate(2, 2){ (x, y) => (x + y).U }
```
While `.asUInt` is defined on `Int`, we typically use just `.U`.

Also general Scala style is to have spaces around operators and after commas

docs/src/cookbooks/cookbook.md
```diff
@@ -157,6 +158,39 @@ See the [DataView cookbook](dataview#how-do-i-connect-a-subset-of-bundle-fields)
...
+  assert(indexTiedVec(0)(0) === 0.U)
+  assert(indexTiedVec(0)(1) === 1.U)
+  assert(indexTiedVec(1)(0) === 1.U)
+  assert(indexTiedVec(1)(1) === 2.U)
+
+  val indexTiedVec3D = VecInit.tabulate(1,2,3){ (x, y, z) => (x*y*z).asUInt }
```
jackkoenig, line 184:
```suggestion
  val indexTiedVec3D = VecInit.tabulate(1, 2, 3){ (x, y, z) => (x * y * z).U }
```

docs/src/cookbooks/cookbook.md
```diff
@@ -157,6 +158,39 @@ See the [DataView cookbook](dataview#how-do-i-connect-a-subset-of-bundle-fields)
...
+  assert(indexTiedVec(0)(0)(0) === 0.U)
+  assert(indexTiedVec(1)(1)(1) === 1.U)
+  assert(indexTiedVec(1)(1)(2) === 2.U)
+  assert(indexTiedVec(1)(1)(3) === 3.U)
+  assert(indexTiedVec(1)(2)(3) === 6.U)
+
+}
+```
+
```
jackkoenig, line 193: It would be good to actually run this test too:

````suggestion
```scala mdoc:invisible
// Hidden but will make sure this actually compiles
ChiselStage.emitVerilog(new Foo)
```
````

File: macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala [hunk 9]
```diff
@@ -72,6 +72,9 @@ class MuxTransform(val c: Context) extends SourceInfoTransformMacro {
 object VecTransform
 class VecTransform(val c: Context) extends SourceInfoTransformMacro {
   import c.universe._
+  def get_connect_op(proto: c.Tree): c.Tree = {
+    q"$thisObj.do_getConnectOpFromDirectionality($proto)($implicitSourceInfo, $implicitCompileOptions)"
+  }
```
jackkoenig, line 77:
```suggestion
```
No need for this since that method should be private.

src/test/scala/chiselTests/Vec.scala
```diff
@@ -114,6 +115,88 @@ class FillTester(n: Int, value: Int) extends BasicTester {
...
+    val mod = Module(new PassthroughModule)
+    val vec2D = VecInit.fill(n, m)(mod.io)
+    val tester = Module(new PassthroughModuleTester)
+    for {
+        vec1D <- vec2D
+        module <- vec1D
+    } yield {
+        module <> tester.io
+        assert(module.out === 123.U)
+    }
```
jackkoenig, line 179:
```suggestion
    val vec2D = VecInit.fill(n, m)(Module(new PassthroughModule).io)
    for {
        vec1D <- vec2D
        module <- vec1D
    } yield {
        module <> Module(new PassthroughModuleTester).io
    }
```

Your version technically does test bidirectionally but it was a bit weird because it basically expands out the same module's IO multiple times and then ignores all of them but the last one (last connect semantics); take a look at the CHIRRTL to see what I mean).

This version creates multiple Modules and tests each one. Also note that `PassthroughModuleTester` already has the assertion inside of it.

src/test/scala/chiselTests/Vec.scala
```diff
@@ -114,6 +115,88 @@ class FillTester(n: Int, value: Int) extends BasicTester {
...
+        vec2D <- vec3D
+        vec1D <- vec2D
+        module <- vec1D
+    } yield {
+        module <> tester.io
+        assert(module.out === 123.U)
+    }
+    stop()
+  }
+}
```
jackkoenig, line 198: Also there are no tests for the `.tabulate`. I'd do something similar but it might be useful to factor in the indices in the tabulate into the test somehow.

jackkoenig (commented):
Some nits and then a couple of suggestions on the tests.

src/test/scala/chiselTests/Vec.scala
anniej-sifive: The fill's call tabulate, bc they are a special case of tabulate, so I figured by calling 2D/3D fill I would be catching tabulate within that.

jackkoenig: They do right now, but part of the goal of tests is catching if someone were to refactor things and get it slightly wrong. Perhaps after a refactor `fill` won't call `tabulate` anymore and I'll bet the person doing the refactor won't bother to check the tests.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -513,18 +528,11 @@ object VecInit extends SourceInfoDoc {
...
-      case ActualDirection.Input | ActualDirection.Output | ActualDirection.Unspecified =>
-        // When internal wires are involved, driver / sink must be specified explicitly, otherwise
-        // the system is unable to infer which is driver / sink
-        (vec zip elts).foreach(x => x._1 := x._2)
-      case ActualDirection.Bidirectional(_) =>
-        // For bidirectional, must issue a bulk connect so subelements are resolved correctly.
-        // Bulk connecting two wires may not succeed because Chisel frontend does not infer
-        // directions.
-        (vec zip elts).foreach(x => x._1 <> x._2)
+    val tpe = cloneSupertype(elts, "Vec")
```
jackkoenig, line 531:
```suggestion
```
`tpe` is unused.

jackkoenig (approved):
LGTM! Great work @anniej-sifive!

PR #2073, opened 2021-08-16
Emit FIRRTL `BulkConnect`s for `Record`s, `Vec`s when possible

debs-sifive:
### Contributor Checklist

cc @jackkoenig 

#### Type of Improvement

- backend code generation

#### API Impact
No visible impact to user.

FIRRTL BulkConnects (`<-`) are emitted when possible for `Record`s and `Vec`s, instead of element-wise connections. This should decrease the FIRRTL emission size.

No impact on generated Verilog.

#### Desired Merge Strategy

#### Release Notes

FIRRTL `BulkConnect`s for `Record`s, `Vec`s are emitted when possible, decreasing FIRRTL code emission size.

- [ ] Did you review?


core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -178,6 +188,35 @@ private[chisel3] object MonoConnect {
       case (sink, source) => throw MismatchedException(sink.toString, source.toString)
     }
 
+  /** Check whether two [[Vec]]s can be bulk connected (<-) in FIRRTL. */
+  private[chisel3] def canBulkConnectVecs(sink_v: Vec[Data @unchecked],
```
debs-sifive, line 192: This function and `canBulkConnectRecords` are also used in `BiConnect`. Is there a better place I should move these out to, or can they just live in `MonoConnect`?

src/test/scala/chiselTests/Vec.scala
```diff
@@ -308,4 +308,19 @@ class VecSpec extends ChiselPropSpec with Utils {
         }}
     }
   }
+
+  property("Vec connections should emit FIRRTL bulk connects when possible") {
```
mwachs5, line 312: should we include some of the "when not possible" cases in the tests?

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -178,6 +188,35 @@ private[chisel3] object MonoConnect {
       case (sink, source) => throw MismatchedException(sink.toString, source.toString)
     }
 
+  /** Check whether two [[Vec]]s can be bulk connected (<-) in FIRRTL. */
```
mwachs5, line 191: can you add a bit more color here? What are the cases when they CANNOT be bulk-connected?

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -178,6 +188,35 @@ private[chisel3] object MonoConnect {
...
+      case _ => true
+    }
+    val elemsMatch = CheckTypes.validConnect(
+      Converter.extractType(sink_v, sourceInfo),
+      Converter.extractType(source_v, sourceInfo),
+    )
+    canWriteBinding && elemsMatch
+  }
+
+  /** Check whether two [[Record]]s can be bulk connected (<-) in FIRRTL. */
```
mwachs5, line 206: same comment as above -- when can Records not be bulk connected?

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -178,6 +188,35 @@ private[chisel3] object MonoConnect {
...
+  /** Check whether two [[Vec]]s can be bulk connected (<-) in FIRRTL. */
+  private[chisel3] def canBulkConnectVecs(sink_v: Vec[Data @unchecked],
+                                          source_v: Vec[Data @unchecked],
+                                          sourceInfo: SourceInfo): Boolean = {
+    val canWriteBinding = source_v.topBinding match {
+      case _: ReadOnlyBinding => false
+      case _ => true
+    }
+    val elemsMatch = CheckTypes.validConnect(
+      Converter.extractType(sink_v, sourceInfo),
```
mwachs5, line 200: do the Vecs need to be parameterized exactly in order to be bulk connected? is that what is captured in `extractType`?

src/test/scala/chiselTests/MixedVecSpec.scala
```diff
@@ -277,4 +277,19 @@ class MixedVecSpec extends ChiselPropSpec with Utils {
...
+        val outMono = Output(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+        val inBi = Input(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+        val outBi = Output(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+      })
+      io.outMono := io.inMono
+      io.outBi <> io.inBi
+    })
+    chirrtl should include ("io.outMono <- io.inMono @[MixedVecSpec.scala")
+    chirrtl should include ("io.outBi <- io.inBi @[MixedVecSpec.scala")
+  }
```
mwachs5, line 294: It would be nice to see a case where the MixedVecs are parameterized differently. Can they be bulk connected in that case? If not it would be nice to see the check that they are NOT bulk connected.

src/test/scala/chiselTests/Vec.scala
jackkoenig: Yes we should

src/test/scala/chiselTests/MixedVecSpec.scala
```diff
@@ -277,4 +277,19 @@ class MixedVecSpec extends ChiselPropSpec with Utils {
...
+        val inMono = Input(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+        val outMono = Output(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+        val inBi = Input(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+        val outBi = Output(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+      })
+      io.outMono := io.inMono
+      io.outBi <> io.inBi
+    })
+    chirrtl should include ("io.outMono <- io.inMono @[MixedVecSpec.scala")
+    chirrtl should include ("io.outBi <- io.inBi @[MixedVecSpec.scala")
```
jackkoenig, line 293:
```suggestion
    chirrtl should include ("io.outMono <= io.inMono @[MixedVecSpec.scala")
    chirrtl should include ("io.outBi <= io.inBi @[MixedVecSpec.scala")
```
`import chisel3._` **must** emit `<=`

core/src/main/scala/chisel3/internal/MonoConnect.scala
jackkoenig, line 191:
```suggestion
  /** Check whether two [[Vec]]s can be bulk connected (<=) in FIRRTL. */
```
`<-` is the unsafe (stringly-typed) connect in FIRRTL. `<=` is the stricter form. They are both bidirectional bulk connects. We need to emit `<=` from `import chisel3._`

core/src/main/scala/chisel3/internal/MonoConnect.scala
jackkoenig: They need not be exact, implicit truncation remains the law of the land and that is correctly handled by `validConnect`.

@debs-sifive One thing that we also need to check is the Flows. The method needed for this is not exposed from FIRRTL so we might need to make some utility methods in FIRRTL: https://github.com/chipsalliance/firrtl/blob/4991ee209b4ae179afcb3e043bbe7ba92006931d/src/main/scala/firrtl/passes/CheckFlows.scala#L64. This is a bit tricky because those seem to require Kind and throw exceptions so we can't just reuse out of the box anyway

For development purposes, we can just write these utilities in Chisel and then migrate them to FIRRTL.

We also need to check "visibility" (eg. https://github.com/chipsalliance/chisel3/blob/b6929c9ad438db26055707b8ec5c66e4f70a22b8/core/src/main/scala/chisel3/internal/MonoConnect.scala#L217, need to make sure we're not issuing connects if one of the sides is not part of the current module).

core/src/main/scala/chisel3/internal/MonoConnect.scala
debs-sifive: Looks like `<=` is emitted by `chisel3.internal.firrtl.Connect`, and `<-` by `chisel3.internal.firrtl.BulkConnect`: https://github.com/chipsalliance/chisel3/blob/master/src/main/scala/chisel3/internal/firrtl/Emitter.scala#L74

So I want to be calling `chisel3.internal.firrtl.Connect`? Or should I be changing the behavior of `chisel3.internal.firrtl.BulkConnect`?

jackkoenig: the `chisel3.internal.firrtl.*Connect` names are misnomers, both `<=` and `<-` are bulk connects, `<-` is called "PartialConnect" in FIRRTL

jackkoenig: Do not change the behavior of `BulkConnect` (since that is what is issued by `import Chisel._`), instead you can just issue `chisel3.internal.firrtl.Connect`

src/test/scala/chiselTests/Vec.scala
jackkoenig: Some cases we should test:

1. `import chisel3._` Bundle that instantiates an `import Chisel._` Bundle that relies on the Stringly-typed connections for the connect to be legal, this should not emit a bulk connect. [Some inspiration](https://scastie.scala-lang.org/wy301om6RQii8u6iSe1DRw) -- this example is a little weird because I'm surprised that `out.buzz.foo <= in.buzz.foo` in the emitted CHIRRTL uses `<=` instead of `<-`)
2. A case where `<>` will magically work but the FIRRTL types would be wrong for `<=`, eg. https://scastie.scala-lang.org/9xV1AAU0SFq1UvQLVbGFFw -- `enq <= deq` is illegal while Chisel can do the connecting by splitting it up.

jackkoenig: Similar principles apply in Vecs, in fact you could make Vecs of these Bundles

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -178,6 +188,168 @@ private[chisel3] object MonoConnect {
       case (sink, source) => throw MismatchedException(sink.toString, source.toString)
     }
 
+  /** Check [[Aggregate]] visibility. */
+  private def aggregateConnectContextCheck(implicit sourceInfo: SourceInfo, connectCompileOptions: CompileOptions,
```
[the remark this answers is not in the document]
adkian-sifive: I wonder whether that will affect error message granularity - I found that the way it's laid out right now is nice for pointing out exactly which error condition was hit. Though I think it probably won't matter for this particular example when there's only one exception case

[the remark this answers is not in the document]
debs-sifive: Ideally you should get the same error message — if we fail to emit a FIRRTL BulkConnect, we should try emitting as we did before (element-wise connect). And if that also fails, you should get the same element-wise failure message.

PR #2074, opened 2021-08-16
[docs] Add example of stripping directions from type

jackkoenig:
This question came up on gitter: https://gitter.im/freechipsproject/chisel3?at=6117c83e7555e333510edb83

### Contributor Checklist

#### Type of Improvement

  - documentation

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash
 
- [x] Did you review?

### DIFF
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -27,6 +27,8 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
   * [How do I get Chisel to name signals properly in blocks like when/withClockAndReset?](#how-do-i-get-chisel-to-name-signals-properly-in-blocks-like-whenwithclockandreset)
   * [How do I get Chisel to name the results of vector reads properly?](#how-do-i-get-chisel-to-name-the-results-of-vector-reads-properly)
   * [How can I dynamically set/parametrize the name of a module?](#how-can-i-dynamically-setparametrize-the-name-of-a-module)
+* Directionality
+  * [How do I strip directions from a bidirectional Bundle (or other Data)?](#how-do-i-strip-directions-from-a-bidirectional-bundle-or-other-data)
 
 ## Type Conversions
 
@@ -461,4 +463,53 @@ ChiselStage.emitVerilog(new Salt)
 
 ```scala mdoc:verilog
 ChiselStage.emitVerilog(new Salt)
+```
+
+## Directionality
+
+### How do I strip directions from a bidirectional Bundle (or other Data)?
+
+Given a bidirectional port like a `Decoupled`, you will get an error if you try to connect it directly
+to a register:
+
+```scala mdoc:silent
+import chisel3.util.Decoupled
+class BadRegConnect extends Module {
+  val io = IO(new Bundle {
+    val enq = Decoupled(UInt(8.W))
+  })
+  
+  val monitor = RegNext(io.enq)
+}
+```
+
+```scala mdoc:crash
+ChiselStage.emitVerilog(new BadRegConnect)
+```
+
+While there is no construct to "strip direction" in Chisel3, wrapping a type in `Output(...)` will
+coerce the direction to `Output` (even if the input is bidirectional).
+This will have the desired result when used to construct a Register:
+
+```scala mdoc:silent
+import chisel3.util.Decoupled
+class CoercedRegConnect extends Module {
+  val io = IO(new Bundle {
+    val enq = Flipped(Decoupled(UInt(8.W)))
+  })
+  
+  // Make a Reg of the correct type instead of using RegNext
+  val monitor = Reg(Output(chiselTypeOf(io.enq)))
+  monitor := io.enq // monoconnect will ignore directions of right-hand side
+  // dontTouch so that it shows up in the Verilog
+  dontTouch(monitor)
+}
+```
+
+<!-- Just make sure it actually works -->
+```scala mdoc:invisible
+ChiselStage.emitVerilog(new CoercedRegConnect {
+  // Provide default connections that would just muddy the example
+  io.enq.ready := true.B
+})
 ```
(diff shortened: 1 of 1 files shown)


docs/src/cookbooks/cookbook.md
```diff
@@ -462,3 +464,52 @@ ChiselStage.emitVerilog(new Salt)
...
+This will have the desired result when used to construct a Register:
+
+```scala mdoc:silent
+import chisel3.util.Decoupled
+class CoercedRegConnect extends Module {
+  val io = IO(new Bundle {
+    val enq = Flipped(Decoupled(UInt(8.W)))
+  })
+  
+  // Make a Reg of the correct type instead of using RegNext
```
mwachs5, line 501: I think this example isn't very clear without the resulting verilog. Is this saying that you will only make registers for the outputs? Or will you get a register for the `monitor.valid` as well?

docs/src/cookbooks/cookbook.md
```diff
@@ -462,3 +464,52 @@ ChiselStage.emitVerilog(new Salt)
...
+  val monitor = Reg(Output(chiselTypeOf(io.enq)))
+  monitor := io.enq // monoconnect will ignore directions of right-hand side
```
mwachs5, line 503:
```suggestion
  monitor := io.enq // monoconnect will ignore directions of right-hand side
```

This comment is confusing. *why* will monoconnect ignore the directions. Is it because monoconnect always ignores it? Or are you saying "monoconnect always looks to the LHS for the directions, and in case they are all outputs / Sinks, therefore we will grab everything from io.enq regardless of the source/sink directionality of io.enq"

docs/src/cookbooks/cookbook.md
jackkoenig: I didn't include the Verilog because of all of the RANDOMIZATION stuff which is pretty distracting 🙁

docs/src/cookbooks/cookbook.md
jackkoenig: Would "monoconnect ignores directions" be better? Or no comment?

docs/src/cookbooks/cookbook.md
mwachs5: maybe just a comment then that makes it clear you *will* get a Reg for the `valid` as well?

docs/src/cookbooks/cookbook.md
mwachs5: Well, why does it have to be `Output` instead of `Input`?

jackkoenig: Because `Output` is the default direction 🙃

jackkoenig: I take it back, `Input(...)` works fine too: https://scastie.scala-lang.org/4BrQKZAoT8S9x4kxDcGNlQ

mwachs5: 🤯

mwachs5: this is making me feel like we should create a `StripDirection` alias for `Output`

docs/src/cookbooks/cookbook.md
```diff
@@ -462,3 +464,52 @@ ChiselStage.emitVerilog(new Salt)
...
+}
+```
+
+```scala mdoc:crash
+ChiselStage.emitVerilog(new BadRegConnect)
+```
+
+While there is no construct to "strip direction" in Chisel3, wrapping a type in `Output(...)` will
+coerce the direction to `Output` (even if the input is bidirectional).
+This will have the desired result when used to construct a Register:
```
mwachs5, line 492:
```suggestion
While there is no construct to "strip direction" in Chisel3, wrapping a type in `Output(...)`
(the default direction in Chisel3) will
remove the directionality information of the individual elements.
This will have the desired result when used to construct a Register:
```

docs/src/cookbooks/cookbook.md
mwachs5, line 501:
```suggestion
  // Make a Reg which contains all of the bundle's signals, regardless of their directionality
```

docs/src/cookbooks/cookbook.md
```diff
@@ -462,3 +464,52 @@ ChiselStage.emitVerilog(new Salt)
...
+to a register:
+
+```scala mdoc:silent
+import chisel3.util.Decoupled
+class BadRegConnect extends Module {
+  val io = IO(new Bundle {
+    val enq = Decoupled(UInt(8.W))
+  })
+  
+  val monitor = RegNext(io.enq)
```
mwachs5, line 482:
```suggestion
  val monitor = Reg(chiselTypeOf(io.enq))
  monitor := io.enq
```

Can you do this without `RegNext`, to make it less different from the thing that works below?

docs/src/cookbooks/cookbook.md
mwachs5: I guess I still don't get the monoconnect rules in general given this conversation.

docs/src/cookbooks/cookbook.md
jackkoenig: sure

docs/src/cookbooks/cookbook.md
jackkoenig: > "remove the directionality information of the individual elements"

That's not strictly true, it does set them all to `Output` but reg is fine with that 😬

docs/src/cookbooks/cookbook.md
jackkoenig: `<>` will see that the RHS is bidirectional and use that instead of just having things flow from right to left.

docs/src/cookbooks/cookbook.md
mwachs5: ok editing the above

docs/src/cookbooks/cookbook.md
mwachs5: ok, then maybe just "monoconnect ignores directions from the right hand side" if that is accurate statement...

jackkoenig:
@mwachs5 I accepted your suggestions and tried to improve the comment for the `:=`

mwachs5 (approved):

PR #2077, opened 2021-08-17
Signal Trace API

sequencer:
This API provides a function to inspect the final `Target` via `Data` name, which provides a better functionality for verifications flow to get the REAL hierarchy and signal name with chisel in-memory `Data`.
More tests and demos is TBD.
 
Basically, this can improves a symbol table from `HasId`(`Data`, `Module`) to `Target`, you can `tap` in the code during the elaboration time, and `view` it `Target` out of `Builder`.
For different simulation backends, this can enable internal signal peek/poke flow via:
1. Verilator has [config](https://verilator.org/guide/latest/extensions.html?highlight=config#cmdoption-verilator-public_flat_rw) file.
2. VCS can directly see internal signal if a full name/hierarchy is provided.
3. just convert the Target to treadle symbol table is enough to use treadle API.

Even there are some rename pass, `Target` will still have a correct name. which enhance this the robustness.  
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/Trace.scala
+++ b/core/src/main/scala/chisel3/experimental/Trace.scala
@@ -1,0 +1,69 @@
+package chisel3.experimental
+
+import chisel3.internal.HasId
+import chisel3.{Aggregate, Data, Element, Module}
+import firrtl.AnnotationSeq
+import firrtl.annotations.{Annotation, CompleteTarget, SingleTargetAnnotation}
+import firrtl.transforms.DontTouchAllTargets
+
+/** The util that records the reference map from original [[Data]]/[[Module]] annotated in Chisel and final FIRRTL.
+  * @example
+  * {{{
+  *   class Dut extends Module {
+  *     val a = WireDefault(Bool())
+  *     Trace.traceName(a)
+  *   }
+  *   val annos = (new ChiselStage).execute(Seq(ChiselGeneratorAnnotation(() => new Dut)))
+  *   val dut = annos.collectFirst { case DesignAnnotation(dut) => dut }.get.asInstanceOf[CollideModule]
+  *   // get final reference of `a` Seq(ReferenceTarget("Dut", "Dut", Seq.empty, "a", Seq.empty))
+  *   val firrtlReferenceOfDutA = finalTarget(annos)(dut.a)
+  * }}}
+  * */
+object Trace {
+
+  /** Trace a Instance name. */
+  def traceName(x: Module): Unit = {
+    annotate(new ChiselAnnotation {
+      def toFirrtl: Annotation = TraceNameAnnotation(x.toAbsoluteTarget, x.toAbsoluteTarget)
+    })
+  }
+
+  /** Trace a Data name. */
+  def traceName(x: Data): Unit = {
+    x match {
+      case aggregate: Aggregate =>
+        annotate(new ChiselAnnotation {
+          def toFirrtl: Annotation = TraceNameAnnotation(aggregate.toAbsoluteTarget, aggregate.toAbsoluteTarget)
+        })
+        aggregate.getElements.foreach(traceName)
+      case element: Element =>
+        annotate(new ChiselAnnotation {
+          def toFirrtl: Annotation = TraceNameAnnotation(element.toAbsoluteTarget, element.toAbsoluteTarget)
+        })
+    }
+  }
+
+  /** An Annotation that records the original target annotate from Chisel.
+    *
+    * @param target target that should be renamed by [[firrtl.RenameMap]] in the firrtl transforms.
+    * @param chiselTarget original annotated target in Chisel, which should not be changed or renamed in FIRRTL.
+    */
+  private case class TraceNameAnnotation[T <: CompleteTarget](target: T, chiselTarget: T)
+    extends SingleTargetAnnotation[T]
+    with DontTouchAllTargets {
+    def duplicate(n: T): Annotation = this.copy(target = n)
+  }
+
+  /** Get [[CompleteTarget]] of the target `x` for `annos`.
+    * This API can be used to find the final reference to a signal or module which is marked by `traceName`
+    */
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      Seq(
+        ChiselGeneratorAnnotation(() => new Module1)
+      )
+    )
+    val dut = annos.collectFirst { case DesignAnnotation(dut) => dut }.get.asInstanceOf[Module1]
+    // out of Builder.
+    val oneTarget = chisel3.experimental.DataMirror.trace.view(dut.m0.r.a.a, annos)
+    val ioTarget = chisel3.experimental.DataMirror.trace.view(dut.m0.i.b(1)(2), annos)
+    oneTarget.head.toString should be ("~Module1|Module0>r_a_a")
+    ioTarget.head.toString should be ("~Module1|Module0>i_b_1_2")
```
ekiwi, line 55: What are you trying to test here? Are you trying to predict the name of signals after the circuit has been lowered to LoFirrtl? If so, you need to take into account collisions as we do in the `LowerTypes` pass.

jackkoenig (dismissed):
I like the idea here a lot. I think the most important thing to do here though is to be very focused on use cases and the user API of those use cases. I'm not sure if this is the right API yet without seeing an example of a useful API built using it (eg. a Verilator config file generation API).

My request is that in `src/test/scala`, you show what it might look like to generate Verilator config files using this stuff. We could then include that code in chisel3 and/or firrtl as well, but I'd at least like to see what it looks like so we can be sure if `tap` and `view` are the right API or not.

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
sequencer: I thought `RenameMap` already handled collisions?

ekiwi (changes requested):
This needs tests that demonstrate what happens if the firrtl compiler needs to rename a signal. There are two common situations that need to be tested:

1. when a bundle is renamed because it collides with an existing signal name (see the [LowerTypes spec](https://github.com/chipsalliance/firrtl/blob/master/src/test/scala/firrtl/passes/LowerTypesSpec.scala))
2. when a module is inline (see the [InlineInstances spec](https://github.com/chipsalliance/firrtl/blob/master/src/test/scala/firrtlTests/InlineInstancesTests.scala))

ekiwi (changes requested):
The tests you added are good!

Is there a reason why this `trace` functionality needs to be part of the `DataMirror` object? If you could put the `DataMirror` code back into its old location and add your new `trace` API to a new file, that would make reviewing it a lot easier.

sequencer:
@ekiwi done, trace instance name might contains some bug. I'll work on this this weekend/next week.

ekiwi:
> @ekiwi done, trace instance name might contains some bug. I'll work on this this weekend/next week.

Sound good. Let me know when you have ironed out the bugs and then we should be able to merge this soon. Good job!

core/src/main/scala/chisel3/experimental/HasChiselTarget.scala
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental
+import firrtl.annotations.{CompleteTarget, SingleTargetAnnotation}
+
+/** An Annotation that records the original target annotate from Chisel. */
+trait HasChiselTarget[T <: CompleteTarget] { x: SingleTargetAnnotation[T] =>
```
ekiwi, line 7: Why did you make this trait, when it is only used by a single class?
Seems like your code would be simpler without making this a separate trait.

core/src/main/scala/chisel3/experimental/trace.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /** Trace a Data name. */
+  def traceName(x: Data): Unit = {
+    x match {
+      case aggregate: Aggregate =>
+        annotate(new ChiselAnnotation {
+          def toFirrtl: Annotation = TraceNameAnnotation(aggregate.toAbsoluteTarget, aggregate.toAbsoluteTarget)
+        })
+        annotate(new ChiselAnnotation {
```
davidbiancolin, line 26: You should probably use DontTouchAllTargets or the associated traits instead of emitting auxiliary dontTouch annotations.

sequencer:
@ekiwi ready for review.

ekiwi (changes requested):
Going into the right direction. Some small changes still needed.

core/src/main/scala/chisel3/experimental/Trace.scala
```diff
@@ -6,8 +6,19 @@ import firrtl.AnnotationSeq
 import firrtl.annotations.{Annotation, CompleteTarget, SingleTargetAnnotation}
 import firrtl.transforms.DontTouchAllTargets
 
-/** The util that records the reference map from original [[Data]]/[[Module]] annotated in Chisel and final FIRRTL. */
-object trace {
+/** The util that records the reference map from original [[Data]]/[[Module]] annotated in Chisel and final FIRRTL.
+  * @example
+  * {{{
```
ekiwi, line 11: Nice example, but isn't it missing a call to `traceName`?

ekiwi (changes requested):
Just need to fix the example.

core/src/main/scala/chisel3/experimental/Trace.scala
sequencer: oh my fault.

ekiwi (approved):
LGTM! :ship: 

Just make sure to squash everything into a single commit.

sequencer:
ask @jackkoenig for approval ;p

azidar:
As I'm filling in for @jackkoenig because he is on vacation, I think this is ok to merge. I'm not in love with the user-facing API, but I think since its experimental its ok, and I think this will be very useful.

PR #2089, opened 2021-08-26
graduate new assert/assume/cover statements from experimental into the chisel3 namespace

ekiwi:
### Contributor Checklist

#### Type of Improvement

 - new feature/API

#### API Impact

- `BaseSim` renamed to `VerificationStatement` (however `BaseSim` was introduced in the `3.5` dev cycle, so not a breaking change)
- `chisel3.assert` now returns an id that makes it annotatable
- new `chisel3.assume` that has all the features provided by `chisel3.assert`
- new `chisel3.cover`, takes a message but it is not used yet (adding a printf with the message makes little sense)
- `chisel3.stop(Int)` is now deprecated, the return code was never properly respected by the Verilog emitter
- `chisel3.stop()` now takes an optional `message` string which is not used yet
- `chisel3.stop` is now annotatable just like all other verification statements
- all functions in `chisel3.experimental.verification` are now deprecated and alias to their new implementation in the `chisel3` namespace

- `chisel3.stop` now generates the firrtl `assert` instead of `stop(1)`

#### Desired Merge Strategy

- squash

- [ ] Did you review?


core/src/main/scala/chisel3/Assert.scala
```diff
@@ -77,6 +90,38 @@ object assert {
...
+    val a = new Assume(predicate)
+    when (!Module.reset.asBool) {
+      val clock = Module.clock
+      Builder.pushCommand(Verification(a, Formal.Assume, sourceInfo, clock.ref, predicate.ref, ""))
+    }
+    a
+  }
+}
+
+object cover {
```
debs-sifive, line 109: nit: perhaps these should live in their own files, or else rename this file to not be `Assert.scala`

debs-sifive (commented):
This LGTM! (Also sorry for the delay, was OOO for a bit)

core/src/main/scala/chisel3/Assert.scala
ekiwi: It already did have the `stop` implementation in it. But yeah, renaming it to `Verification.scala` might be a good idea. (I am also happily taking any alternative naming suggestions :) )

ekiwi:
@azidar: I think I need your help to fix the `SelectSpec`. Not sure, why it is failing now. I think the problem is that `assert` now uses an `assert` instead of a `stop(1)`.

core/src/main/scala/chisel3/Assert.scala
ekiwi: done

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -0,0 +1,235 @@
...
+    val fmt = message match {
+      case Some(msg) =>
+        s"$kind failed: $msg\n    at $lineMsg\n"
+      case None => s"$kind failed\n    at $lineMsg\n"
+    }
+    when(!cond) {
+      printf.printfWithoutReset(fmt, data:_*)
+    }
+  }
+}
```
mwachs5, line 235:
```suggestion
}

```

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -166,9 +166,4 @@ package object experimental {
   val prefix = chisel3.internal.prefix
   // Use to remove prefixes not in provided scope
   val noPrefix = chisel3.internal.noPrefix
-
-  /** Base simulation-only component. */
-  abstract class BaseSim extends NamedComponent {
```
mwachs5, line 171: is a printf, assume, etc still a `NamedComponent`?

ekiwi: Yes. I just renamed the base class to `VerificationStatement` and moved it into the `chisel3` package.

core/src/main/scala/chisel3/VerificationStatement.scala
ekiwi: thanks. fixed.

debs-sifive:
Can you clarify a bit on how this assert/assume/cover is related (or not) to the [experimental verif stuff](https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala/chisel3/experimental/verification/package.scala)?

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -0,0 +1,236 @@
...
+    *
+    * May be called outside of a Module (like defined in a function), so
+    * functions using assert make the standard Module assumptions (single clock
+    * and single reset).
+    *
+    * @param cond condition that will be sampled on every clock tick
+    * @param message a string describing the cover event
+    */
+  // Macros currently can't take default arguments, so we need two functions to emulate defaults.
+  def apply(cond: Bool, message: String)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Cover = macro _applyMacroWithMessage
```
debs-sifive, line 149: why does this one not also take a `Bits*`?

ekiwi: I did not find a reasonable way to use a `printf` statement with a `cover` statement.
The failure case with a `cover` is if it cannot be covered. In that case, there are no values to print.
If you have an idea on how a `printf` statement could make sense with cover, I would be happy to hear about it.

ekiwi:
> Can you clarify a bit on how this assert/assume/cover is related (or not) to the [experimental verif stuff](https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala/chisel3/experimental/verification/package.scala)?

It is essentially subsuming it and adding new capabilities to `assert` and `assume` to have the same `printf` mechanism that we used to have with the `stop(1)` based `chisel3.asset`. I hope that this way of exposing the new verification statements will be mostly source compatible with Chisel 3.4 code.

core/src/main/scala/chisel3/VerificationStatement.scala
debs-sifive: ah ok never mind, i see what you mean now!

ekiwi:
> LGTM, again. I guess the last question is if the cover `message` isn't being used at all, should it be added, or should we wait until there is a use case to avoid dead code?

I am working on a followup PR which will automatically attach a `VerificationInfoAnnotation` to every assert, assume, cover and stop. That annotation will have a `message` field.

My current prototype looks like:
```
/** meta-data of assert, assume, cover and stop statements which can be used by downstream tools to analyze verification results */
case class VerificationInfoAnnotation(statement: ReferenceTarget, print: Option[ReferenceTarget], info: VerificationInfo) extends MultiTargetAnnotation {

  override def targets = Seq(Seq(statement), print.toSeq)
  override def duplicate(n: Seq[Seq[Target]]) = n match {
    case Seq(Seq(s: ReferenceTarget), Seq(p: ReferenceTarget)) => copy(statement=s, print=Some(p))
    case Seq(Seq(s: ReferenceTarget), Seq()) => copy(statement=s, print=None)
    case other => throw new RuntimeException(s"Unexpected argument to duplicate: $other")
  }
}

case class VerificationInfo(message: String, filename: String, line: Int, cond: String, kind: String) {
  require(line >= 0)
  require(Set("assert", "assume", "cover", "stop").contains(kind))
}
```

I am thinking about also adding some stack trace information on how the assertion was created. That way, I might be able to throw an exception in chiseltest that contains the stack trace of when the assertion was created.

azidar (approved):
LGTM

PR #2092, opened 2021-08-27
SyncReadMem: fix bug with read(addr) and add some formal tests

ekiwi:
This essentially fixes a bug where in the following code, the read enable would always be false:
```.scala
class BrokenMem extends MultiIOModule {
  val addr = IO(Input(UInt(5.W)))

  val mem = SyncReadMem(32, UInt(8.W))
  out := mem.read(addr)
}
```

You can see the full example [here](https://scastie.scala-lang.org/wLIQK1bKQvaD6FkMTb04Vg)

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

- none

- read enables now work in the one situation that I described above

#### Desired Merge Strategy
- squash

- [x] Did you mark the proper milestone (Bug fix: `3.3.x`, [small] API extension: `3.4.x`, API modification or big change: `3.5.0`)?
- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Mem.scala
+++ b/core/src/main/scala/chisel3/Mem.scala
@@ -213,8 +213,14 @@ sealed class SyncReadMem[T <: Data] private (t: T, n: BigInt, val readUnderWrite
     var port: Option[T] = None
     when (enable) {
       a := addr
-      port = Some(read(a))
+      port = Some(super.do_read(a))
     }
     port.get
   }
+
+  /** @group SourceInfoTransformMacro*/
+  override def do_read(idx: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions) =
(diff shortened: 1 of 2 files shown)


seldridge (approved):
This looks great.

I was originally confused that the SFC was correctly handling this. However, the SFC will not handle this correctly and will generate a constant-read memory (which is what this is describing).

@ekiwi: Can you take care of fixing the backports here (which won't support ChiselTest?). It's probably fine to just remove the test portion from the backport.

seldridge:
I'm running some spot-checks on Verilog for large designs to sanity check this. I'll remove the `DO NOT MERGE` label, once I've finished these.

seldridge:
I tested this with two internal cores and the output Verilog was identical. Good to go from me. Thanks for the fix, @ekiwi.

CC: @jackkoenig

PR #2031, opened 2021-07-09
Fix chisel3 <> for Bundles that contain compatibility Bundles (Take 2)

jackkoenig:
PR #2023 fixed a composition issue for chisel3 biconnects delegating to
FIRRTL partial connect when compatibility mode Bundles are elements of
chisel3 Bundles. It missed an important case though that caused
previously working code to break.

The bug is fixed by doing the automatic flipping for compatibility mode
Bundles that have "Input" as a direction in addition to those that are
"Flipped".

### Contributor Checklist

- [NA] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [NA] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

 - bug fix

#### API Impact

No impact

#### Backend Code Generation Impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

(Not applicable - fixing an unreleased bug)

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (3.2.x, 3.3.x, 3.4.x, 3.5.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

Issue #1953, opened 2021-06-07
Can't have a Vec(new Bundle{}) ?


mwachs5:
**Type of issue**: bug report 

**Impact**: unknown

**Development Phase**: request

**Other information**
 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
See this Scastie:
https://scastie.scala-lang.org/2qc4i466TRS9gvWIAnLmRQ

Note this is using Chisel 3.4.3 with the naming plugin enabled (same behavior seen without plugin enabled: https://scastie.scala-lang.org/XlNASbYNT6WDa641zN9kNQ)

Same code here:
```
import chisel3._
import chisel3.stage.ChiselStage

class Example extends MultiIOModule {
  val in = IO(Input(Bool()))
  val out= IO(Output(Bool()))

  val thisWorks = Reg(new Bundle{})
  val thisDoesNotWork = Reg(Vec(4, new Bundle{})) //I get a None.get error message
    
  out := in
}

```

**What is the current behavior?**

If I try to build a Vec(x, Bundle{}) I get a `None.get` error.

**What is the expected behavior?**

Either I get a good error message about something I am doing wrong (unclear whether Vec-of-empty-bundle is allowed, when a Reg-of-empty-bundle works fine), or it works fine and I get registers-of-empty-bundles.

**Please tell us about your environment:**

Chisel 3.4.3, with naming plugin enabled
**What is the use case for changing the behavior?**

It is nice to be able to build up Bundles of optional fields or sequential fields which might end up that the Bundle itself has nothing in it and have things just work without special casing

mwachs5:
This error is stemming from:

https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala/chisel3/Aggregate.scala#L235

```
    // Since all children are the same, we can just use the sample_element rather than all children    
    // .get is safe because None means mixed directions, we only pass 1 so that's not possible 
 direction = ActualDirection.fromChildren(Set(sample_element.direction), resolvedDirection).get
```

mwachs5:
rewriting the code to make the direction clear works to have an empty bundle in a Vec:
https://scastie.scala-lang.org/OlWvps2cRRa19sFOjdzUsg

schoeberl:
What is the meaning of a ```Reg``` or ```Vec``` of an empty ```Bundle```. Maybe ```Reg((new Bundle{})``` should err.

jackkoenig:
[...]

PR #2093, opened 2021-08-27
Fix chisel3 <> for compatibility Bundles (Take 3)

jackkoenig:
Third time's the charm?

Previous incomplete fixes in #2023 and #2031.

The legality of a FIRRTL connection is determined by type and flow.
Chisel does not have access to true flow information. Previous fix
attempts tried to use ActualDirection as a stand-in for flow, but it is
incorrect in many cases. This new approach checks the flows of the
lvalue and rvalues in the connect and flips the connection if either
the lvalue cannot be a sink or the rvalue cannot be a source.

**Note** I marked this for backporting to 3.2.x because the previous 2 PRs were and this is a fix to those fixes.

### Contributor Checklist

#### Type of Improvement

 - bug fix   

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix handling of `import Chisel._` Bundles when bulk connected inside of `import chisel3._` code

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -11,6 +11,8 @@ import chisel3.internal.firrtl.{Connect, DefInvalid}
 import scala.language.experimental.macros
 import chisel3.internal.sourceinfo._
 
+import scala.annotation.tailrec
+
 /**
 * BiConnect.connect executes a bidirectional connection element-wise.
 *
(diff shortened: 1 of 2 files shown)


mwachs5:
any chance this solves https://github.com/chipsalliance/chisel3/issues/1953?

jackkoenig:
> any chance this solves #1953?

I doubt it, but what's really confusing is I can't even seem to reproduce that bug locally and I have no idea why. Here's a template repo modeled on that issue but works just fine (with or without the plugin) 🤷‍♂️  https://github.com/jackkoenig/chisel3-issue-1953

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -120,15 +122,24 @@ private[chisel3] object BiConnect {
...
         if (notStrict) {
-          // chisel3 <> is commutative but FIRRTL <- is not
-          val flipped = {
-            import ActualDirection._
-            // Everything is flipped when it's the port of a child
-            val childPort = left_r._parent.get != context_mod
-            val isFlipped = Seq(Bidirectional(Flipped), Input).contains(left_r.direction)
-            isFlipped ^ childPort
+          // Traces flow from a child Data to its parent
+          @tailrec def traceFlow(acc: Boolean, data: Data): Boolean = {
```
azidar, line 126: what is `acc` short for? maybe use a more descriptive word?

azidar (approved):
LGTM, minus changing `acc`

core/src/main/scala/chisel3/internal/BiConnect.scala
jackkoenig: I use it for `accumulator` although this isn't really an accumulation, more of an "accumulated result", any thoughts for a better name?

seldridge: You could consider `accumulatedFlow`. [A similar method in CIRCT uses that name](https://github.com/llvm/circt/blob/7e06c705c1ad1fff384fefe740ddae3ff2bf07f1/lib/Dialect/FIRRTL/FIRRTLOps.cpp#L98) after receiving similar feedback that `acc` was unclear.

jackkoenig: Done.

PR #2119, opened 2021-09-13
Bp 2076

sequencer:
I found I cannot force-force PR opened by mergify. So I reopened this to subsume #2090


azidar:
@sequencer I don't know why its failing, it seems like it's a DirectMissingImplementation error? https://github.com/lightbend/mima/blob/b20f56d9af7543ecbf7c78fe997ca0cf4e3b2a69/core/src/main/scala/com/typesafe/tools/mima/lib/analyze/method/MethodChecker.scala#L105

Does this need to be back ported?

PR #2124, opened 2021-09-16
Deprecate auto-application of empty argument lists to parameterless functions

jared-barocsi:
Scala 3 now enforces a distinction between nullary parameter-less functions and nullary functions with empty argument lists. Previously, Scala 2 would auto-apply an empty parameter list onto any invocation of a nullary parameter-less function.

As such, in order to migrate to Scala 3, any nullary function that doesn't induce a side effect, e.g. `asUInt`, must not include a parameter list in its invocation. To deprecate the use of empty parameter lists for these functions, dummy versions of these functions are included which explicitly allow use of an empty parameter list (for example, a default dummy argument or a vararg).

### Contributor Checklist

#### Type of Improvement

code refactoring

#### API Impact

No effect 

No impact

#### Desired Merge Strategy

Squash and merge

- [ ] Did you review?


core/src/main/scala/chisel3/Bits.scala
```diff
@@ -25,7 +25,9 @@ private[chisel3] sealed trait ToBoolable extends Element {
     *
     * @note The width must be known and equal to 1
     */
-  final def asBool(): Bool = macro SourceInfoWhiteboxTransform.noArg
+  final def asBool: Bool = macro SourceInfoWhiteboxTransform.noArg
+  @deprecated("Calling a nullary function with an empty argument list will be removed by official release", "chisel3")
```
mwachs5, line 29:
```suggestion
  @deprecated("Calling a nullary function with an empty argument list is not valid in Scala3. Use .asBool without parentheses", "3.5")
```

I think a good deprecation message should as clearly as possible what you should do if you get this deprecation message. And I thought the second argument is the version in which this was first deprecated?

core/src/main/scala/chisel3/When.scala
```diff
@@ -111,7 +111,9 @@ final class WhenContext private[chisel3] (
   def otherwise(block: => Any)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Unit =
     new WhenContext(sourceInfo, None, block, firrtlDepth + 1, cond ++: altConds)
 
-  def active(): Boolean = scopeOpen
+  def active: Boolean = scopeOpen
+  @deprecated("Calling a nullary function with an empty argument list will be removed by official release", "chisel3")
+  def active(dummy: Int = 0): Boolean = active
```
mwachs5, line 116: why did you sometimes use this Int-with-default form and sometimes use the variable argument list?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -485,7 +485,7 @@ private[chisel3] object Builder extends LazyLogging {
   }
 
   // Returns the prefix stack at this moment
-  def getPrefix(): Prefix = chiselContext.get().prefixStack
+  def getPrefix: Prefix = chiselContext.get().prefixStack
```
mwachs5, line 488: why does this one not need a deprecation version?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -572,7 +572,7 @@ private[chisel3] object Builder extends LazyLogging {
     dynamicContext.whenStack = s
   }
 
-  def currentWhen(): Option[WhenContext] = dynamicContext.whenStack.headOption
+  def currentWhen: Option[WhenContext] = dynamicContext.whenStack.headOption
```
mwachs5, line 575: why does this one not need a deprecation version?

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -37,7 +37,9 @@ object ReadyValidIO {
 
     /** Indicates if IO is both ready and valid
      */
-    def fire(): Bool = target.ready && target.valid
+    def fire: Bool = target.ready && target.valid
```
mwachs5, line 40: I think there is something in the docs that use `fire`. Which makes me wonder if we need to update the mdoc as part of this PR

davidbiancolin:
Inb4 "`asUInt` _does_ have a side effect".

aswaterman:
> Inb4 "`asUInt` _does_ have a side effect".

trigger warning

core/src/main/scala/chisel3/Bits.scala
jackkoenig: Yes you should always tell the user what to do. I don't think it's necessary to say "Use `.asBool` without parentheses", you could just say "Use form without parentheses" or something, this would make it easier to copy-paste the deprecation method throughout.

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -876,10 +905,10 @@ sealed class SInt private[chisel3] (width: Width) extends Bits(width) with Num[S
   /** @group SourceInfoTransformMacro */
   def do_=== (that: SInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = compop(sourceInfo, EqualOp, that)
 
-//  final def abs(): UInt = macro SourceInfoTransform.noArg
+//  final def abs(): UInt = macro SourceInfoTransform.noArgDummy
 
   def do_abs(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): SInt = {
-    Mux(this < 0.S, (-this), this)
+    Mux(this < 0.S, (this.unary_-(sourceInfo, compileOptions)), this)
```
jackkoenig, line 911: Why is this change necessary? Do you get deprecation warnings any time you do `-x` now??

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: Because it's [package] private, we can do whatever we want. That being said, this includes a `.get()` of the dynamic context so I'd say keep the `()`. 

```suggestion
  def getPrefix(): Prefix = chiselContext.get().prefixStack
```

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: Same package private thing, doesn't need to be deprecated.

jackkoenig:
> Inb4 "`asUInt` _does_ have a side effect".

Go away.

In all seriousness though for anyone wondering though--while basically any Chisel API does have side effects, we go to great lengths to hide those side effects from the user such that Chisel APIs _appear to be side effect free_ and therefore will follow the Scala style of not having parentheses.

jackkoenig (changes requested):
Please follow Megan's suggestion on the deprecation messages (ie. they need to tell the user what to do). As I said in the comment, I don't think they need to mention the name of the method, but do need to say something like "use form without parentheses" or "use form of method without parentheses" or something.

Please also make the dummy argument style consistent. Since varargs are the only style that work for macros, I'd apply that throughout.

Also, please leave a blank line between a method definition and the start of the following definition (including when it's a deprecation annotation).

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -222,7 +224,9 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
     * @return this $coll with each bit inverted
     * @group Bitwise
     */
-  final def unary_~ (): Bits = macro SourceInfoWhiteboxTransform.noArg
+  final def unary_~ : Bits = macro SourceInfoWhiteboxTransform.noArg
+  @deprecated("Calling a nullary function with an empty argument list will be removed by official release", "chisel3")
+  final def unary_~(dummy: Int*): Bits = macro SourceInfoWhiteboxTransform.noArgDummy
```
jackkoenig, line 229:
```suggestion
  final def unary_~ : Bits = macro SourceInfoWhiteboxTransform.noArg

  @deprecated("Calling a nullary function with an empty argument list will be removed by official release", "chisel3")
  final def unary_~(dummy: Int*): Bits = macro SourceInfoWhiteboxTransform.noArgDummy
```

core/src/main/scala/chisel3/Bits.scala
jared-barocsi: The compiler was complaining about ambiguity between this and the dummy form.

jackkoenig: Right, so that ambiguity is a huge problem because it basically means users can't use `-` in a unary way any more on `SInts`. This should've been caught by the tests but apparently we never finished this particular test (https://github.com/chipsalliance/chisel3/blob/958904cb2f2f65d02b2ab3ec6d9ec2e06d04e482/src/test/scala/chiselTests/SIntOps.scala#L47).

Please make that test be active and we need to find a way to not have the ambiguity because obviously `-mySInt` needs to work. Do you know why this same issue doesn't show up for stuff like `unary_!` on `Bool`? I tried it out locally, confirmed this issue but was not able to see the same issue on `Bool` which is very weird.

mwachs5: @jared-barocsi did you add the test to be active? Seems like we still have not gotten to the bottom of this question?

jared-barocsi: @mwachs5 Yes, I just now fixed the ambiguity issue

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -304,10 +310,14 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
   def do_>> (that: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bits
 
   /** Returns the contents of this wire as a [[scala.collection.Seq]] of [[Bool]]. */
-  final def toBools(): Seq[Bool] = macro SourceInfoTransform.noArg
+  final def toBools: Seq[Bool] = macro SourceInfoTransform.noArg
+  @deprecated("Calling this function with an empty argument list is invalid in Scala 3. Use the form without parentheses instead", "3.5")
+  final def toBools(dummy: Int*): Seq[Bool] = macro SourceInfoWhiteboxTransform.noArgDummy
```
jackkoenig, line 315: Interestingly, `.toBools` is supposed to be gone and I'm a bit confused why it's here... Definitely not the fault of this PR but weird.

src/main/scala/chisel3/util/Decoupled.scala
jackkoenig: I checked and there are no warnings in the mdoc coming from this PR

jackkoenig (changes requested):
Every deprecation warning needs the version to be `"Chisel 3.5"`, not just `"3.5"` because the Scala compiler either doesn't know or at least doesn't report the name of the project that deprecation warnings come from.

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: cool

core/src/main/scala/chisel3/Bits.scala
mwachs5: I think its still only a deprecation warning?

mwachs5: EDIT: seems like it still exists for `Bits`: https://github.com/chipsalliance/chisel3/blob/4b7b771eeced366345779a75987ce552558a1c7e/core/src/main/scala/chisel3/Bits.scala#L307
Should we get rid of that?

mwachs5: oh. This is the line you are talking about 😂 . Yeah lets get rid of this one in another PR

sequencer: > Interestingly, .toBools is supposed to be gone and I'm a bit confused why it's here... Definitely not the fault of this PR but weird.

My bad, I think did didn't full remove it in https://github.com/chipsalliance/chisel3/pull/1730

I think these should be removed, I can create another PR, or @jared-barocsi can do it in this PR.

mwachs5: @sequencer please create another PR (I opened #2143 to track), as it's not really related to this PR

jackkoenig (approved):

chick:
@jackkoenig @sequencer I think I might have figured this out. Looks like the macro expansions of do_unary_- are not defined for FixedPoint and Interval, not sure how everything worked before here but I'm testing a fix now

sequencer:
super fast!

PR #2132, opened 2021-09-22
Revert default scala version to 2.12 for 3.5 release.

davidbiancolin:
Chipyard builds Chisel + FIRRTL from source using SBT, and it is difficult to override scala versions from an enclosing SBT project when the `lazy val = commonSettings` pattern is used. This changes the default in Chisel to be 2.12 to match FIRRTL, making it easier to build both project from source. 

It might be possible to provide the default value through a settingsKey, so i could prototype that if that is preferable. We plan on moving to published dependencies in the near future, but this fixes the immediate pain of trying out 3.5 and friends. 

### Contributor Checklist

#### Type of Improvement

Other

#### Desired Merge Strategy

Squash

- [ ] Did you review?


ekiwi (approved):
Since all the other projects are still on `2.12` by default, I think this is a good change. Once we fork of the release branch, we can switch all projects to `2.13` by default.

PR #2134, opened 2021-09-23
ci: use Tabby OSS CAD Suite to get solvers and verilator

ekiwi:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,6 @@ jobs:
         system: ["ubuntu-20.04"]
         jvm: ["adopt@1.8"]
         scala: ["2.13.6", "2.12.15"]
-        verilator: ["4.204"]
         espresso: ["2.4"]
     runs-on: ${{ matrix.system }}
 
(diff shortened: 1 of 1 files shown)


sequencer:
we have been struggling switching between different GitHub Actions(to avoid depending on specific version and speed up cache). I personally wanna propose `nix` to provide an environment independent flow, what's @jackkoenig's idea?

ekiwi:
> (to avoid depending on specific version and speed up cache)

This PR does not require caching since it only takes ~16s to download, extract and install the binaries from YosysHQ. It does not let you choose versions independently, however, we can always upgrade to more recent releases which will feature newer versions of all included tools.

sequencer (approved):
I think this is OK, but in the future, I prefer using nix + GitHub CI to test different versions for better compatibilities.

jackkoenig (approved):
I agree this is reasonable given the way we currently (and have historically) tested Chisel

> in the future, I prefer using nix + GitHub CI to test different versions for better compatibilities.

I agree in spirit but not with the specifics of "nix". Nix is great and makes it easier to test lots of versions, but unfortunately one of the most annoying aspects of maintaining a project like Chisel is that you really need to test the _User's_ environment and most users are using Ubuntu, Mac, or Windows. We could have some hybrid approach where Nix is used to test the matrix of versions (for convenience) and then we have just a single version for "User environment testing". That would probably be the ideal CI setup.

ekiwi:
@jackkoenig  Could you mark the correct milestone for this please? I am not sure if CI changes are normally backported or not.

PR #2188, opened 2021-10-18
TruthTable: use ListMap instead of Map

ljwljwljwljw:
**Map** is an unordered data structure, so using **Map** as a parameter of **TruthTable** will cause the decoder to generate a non-reproducible circuit. This will cause the modules that use the decoder unable to be Deduped.

### Contributor Checklist

#### Type of Improvement
- bug fix

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
@@ -5,6 +5,7 @@ package chisel3.util.experimental.decode
 import chisel3.util.BitPat
 
 import scala.annotation.tailrec
+import scala.collection.immutable.ListMap
 import scala.math.Ordered.orderingToOrdered
 import scala.language.implicitConversions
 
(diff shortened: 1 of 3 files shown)


jackkoenig:
Another option is to just make it a `Seq` and then create a `Map` internally if/when lookups are necessary.

ljwljwljwljw:
> Sorry for late reply. IMHO, I think we can
> 
> 1. make `BitPat` extends from `Ordered`
> 2. have this API to consume any Iterable tuple2: `TruthTable.apply(table: Iterable[BitPat, BitPat], defaults) = new TruthTable(table.sorted, defaults)`.
> 3. changing `table` to `Seq` to keep it ordered
> 4. changing `TruthTable` to private to keep away with further compatibility issues.

This can solve the decoder issue, but it does not solve the pla issue (directly use pla without decoder)?

https://scastie.scala-lang.org/knx9264IRSS8D30xccABkw

sequencer:
Yes PLA issue should be resolved as well.

sequencer:
subsumed by #2217

PR #2192, opened 2021-10-19
Make computeName package private and update callsites

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

Public `HasId.computeName` is being deprecated in 3.4.x, and will break binary compatibility.
#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -4,7 +4,7 @@ resolvers += Classpaths.sbtPluginReleases
 
 resolvers += "jgit-repo" at "https://download.eclipse.org/jgit/maven"
 
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.9.0")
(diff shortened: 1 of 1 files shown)


jackkoenig:
We're going to apply the policy described (https://github.com/chipsalliance/chisel3/discussions/2055#discussioncomment-1067786) of break then deprecate in backport.

I'm not 100% sure if I should mark milestone as `3.4.x` so that Mergify does the automatic backport. Or if I should use the mergify command to create it manually. I'm leaning toward the latter but this is making me think that we should either switch to backporting via labels or at least add it as an additional option for this use case (ie. the use case of break then deprecate on backport).

jackkoenig:
This is an API modification, but it is an API that never should've been public.

Mergify will backport once this PR is merged and then @adkian-sifive can deprecate the API on 3.4.x.

PR #2211, opened 2021-10-27
BitSet API

sequencer:
reimplement #2126, `BitSet` is a set of `BitPat`, and `BitPat` will extend from `BitSet` since essentially it is a subset of `BitSet`.
 
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact
API modification for `BitPat(width=)`(ask @jackkoenig for better solution to avoid this)

None

#### Desired Merge Strategy

#### Release Notes
Add `BitSet` API.

- [ ] Did you review?


sequencer:
poke @jackkoenig :)

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+    */
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case that: BitSet => if (this.getWidth == that.getWidth) this.cover(that) && that.cover(this) else false
+      case _ => false
+    }
+  }
+}
+
+object BitSet {
```
jackkoenig, line 78: A ScalaDoc showing some examples (especially construction) would be helpful.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+  val emptyBitSet: BitSet = new BitSet {
+    override val terms = Set()
+  }
+
+  def apply(str: String): BitSet = {
```
jackkoenig, line 83:
```suggestion
  def apply(bitpats: BitPat*): BitSet = {
```
Given apply is supposed to be the canonical constructor, I think having it accept a varargs of `BitPats` is better than accepting a String that is implicitly split on `'\n'` and then parsed into `BitPat`s.

I do believe the String with newlines is useful, but I would call it `BitSet.fromString`

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+    new BitSet {
+      val terms = str
+        .split('\n')
+        .map(str => BitPat(str))
+        .toSet
+      assert(terms.map(_.width).size <= 1)
```
jackkoenig, line 89:
```suggestion
      require(terms.map(_.width).size <= 1, s"All BitPats must be the same size! Got ${...}")
```
1. `assert` implies checking an assumption that fails if the library writer (us) makes a mistake, `require` is used for checking user input
2. User input checks must always have a useful error message, otherwise it's extremely hard to figure out what you did wrong as a user.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+    * Generate a decoder circuit that matches the input to each bitSet.
+    *
+    * The resulting circuit functions like the following but is optimized with a logic minifier
+    * when(input === bitSets(0)) { output := b000001 }
+    * .elsewhen (input === bitSets(1)) { output := b000010 }
+    * ....
+    * .otherwise { if (errorBit) output := b100000 else output := DontCare }
```
jackkoenig, line 100:
```suggestion
    * The resulting circuit functions like the following but is optimized with a logic minifier
    * {{{
    * when(input === bitSets(0)) { output := b000001 }
    * .elsewhen (input === bitSets(1)) { output := b000010 }
    * ....
    * .otherwise { if (errorBit) output := b100000 else output := DontCare }
    * }}}
```
Please create proper ScalaDoc. I recommend running `sbt unidoc` and inspecting the docs coming out to make sure they look the way you expect.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+    * Test whether two BitSetFamily matches the same set of value
+    *
+    * Caution: This method can be very expensive compared to ordinary == operator between two Objects
+    */
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case that: BitSet => if (this.getWidth == that.getWidth) this.cover(that) && that.cover(this) else false
```
jackkoenig, line 72:
```suggestion
      case that: BitSet => this.getWidth == that.getWidth && this.cover(that) && that.cover(this)
```
Scala short circuits boolean operators so this is exactly equivalent and IMO a little more clear.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+trait BitSet { bsf =>
+  val terms: Set[BitPat]
+
+  /**
+    * Get specified width of said BitSet
+    *
+    * All BitPat contained in one set should have the same width
+    */
+  def getWidth: Int = {
+    assert(terms.map(_.width).size <= 1)
```
jackkoenig, line 14: Same comment about the other `assert`.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util
+
+trait BitSet { bsf =>
+  val terms: Set[BitPat]
```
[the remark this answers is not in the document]
jackkoenig: Also, `Set` isn't ordered is it necessary that we make this a `Set` or could it be a `Seq`?

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
...
+        {
+          bitSets.zipWithIndex.flatMap {
+            case (family, i) =>
+              family.terms.map(bs =>
+                s"${bs.toString.replace("-", "?")}->${if (errorBit) "0"}${"0" * (bitSets.size - i - 1)}1${"0" * i}"
+              )
+          } ++ Seq(s"${if (errorBit) "1"}${"?" * bitSets.size}")
+        }.mkString("\n")
+      )
+    )
```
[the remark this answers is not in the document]
jackkoenig: This does cut against my request that `terms` not be public. Perhaps instead `terms` should be public but then it should have a ScalaDoc describing what the terms mean. Is it that a particular bitvector is part of the BitSet if it exists in at least one of the terms?

jackkoenig (changes requested):
I know I'm being a real stickler here, but this feels like a very foundational API so I'm just trying to make sure we get it right.

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util
+
+trait BitSet { bsf =>
```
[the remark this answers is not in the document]
jackkoenig: > Why using bsf instead of more canonical self or just built-in this?

Ah I realize now it has to do with the anonymous inner classes and needing to distinguish. Nevertheless I think `self` is more canonical or `outer` is perhaps more precise.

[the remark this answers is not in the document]
sequencer: I switch to `outer`

src/main/scala/chisel3/util/BitSet.scala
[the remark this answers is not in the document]
sequencer: I switched to 
```
protected[chisel3] def terms: Set[BitPat]
```
since I moved BitSet decode API to decoder.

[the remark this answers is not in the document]
sequencer: > Also, Set isn't ordered, is it necessary that we make this a Set or could it be a Seq?

I'd like to extend `BitPat`  from `Ordered` and then we can sort `Set[BitPat]` when require the order.
I still hesitate making terms a `Seq`, since I doubt should we make `terms` public? what if user wanna extends from `BitSet` to implement their own features?

src/main/scala/chisel3/util/BitSet.scala
sequencer: My bad!

src/main/scala/chisel3/util/BitSet.scala
sequencer: Yes!

src/main/scala/chisel3/util/BitSet.scala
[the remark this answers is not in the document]
sequencer: I made `terms` a `protected[chisel3]` function, since api from set theory makes operated `terms` hard to understand.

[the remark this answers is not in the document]
sequencer: Emmm, I revert `terms` to `public`, making it private will makes the `AddressSet` API hard to implement.

src/main/scala/chisel3/util/BitSet.scala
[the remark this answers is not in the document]
sequencer: > Do we really want the API of BitSet to expose that it's made up of Set[BitPat]s?

I think so, if user wanna implement their own `BitSet` API(like `AddressSet`), `terms` one must be public. And `terms` is the only API which users can access the raw data in a `BitSet`. So I'd like to make this API public.

> Also, Set isn't ordered, is it necessary that we make this a Set or could it be a Seq?

I think Set is OK since there won't be comparing issue, and I have made the serialize API stable via:
```
  override def toString: String = terms.toSeq.sortBy((t: BitPat) => (t.mask, t.value)).mkString("\n")
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -160,14 +159,76 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
...
+      BitSet.emptyBitSet
+    } else {
+      new BitPat(this.value | that.value, this.mask | that.mask, this.width.max(that.width))
+    }
+  }
+
+  /**
+    * Subtract that from this BitPat
+    * @param that subtrahend
+    */
```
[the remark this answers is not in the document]
jackkoenig: This was marked resolved but not addressed. What does it mean to subtract two BitPats?

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -160,14 +159,76 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
...
+          }
+      }
+    }
+  }
+
+  override def isEmpty: Boolean = false
+
+  /** Generate raw string of a BitSat. */
+  def rawString: String = Seq
+    .tabulate(width) { i =>
```
jackkoenig, line 224:
```suggestion
  /** Generate raw string of a BitPat. */
  def rawString: String =
    Seq.tabulate(width) { i =>
```

src/main/scala/chisel3/util/BitSet.scala
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util
+
+/** A Set of [[BitPat]] represents a set of bit vector with mask. */
+trait BitSet { outer =>
```
jackkoenig, line 6: This being an unsealed trait makes me nervous. Can we put it in the same file with `BitPat` and seal it?

jackkoenig (changes requested):
A couple of minor comments and unresolved old comments. This looks awesome, let's get this in.

I am a little nervous about `BitSet` just being a new API. Could/Should we make it experimental?

src/main/scala/chisel3/util/BitSet.scala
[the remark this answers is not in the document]
jackkoenig: > what if user wanna extends from BitSet to implement their own features?

Why do we want the user to extend this? Almost no APIs in Chisel are extensible so I'm not understanding why we should make this particular one extensible.

src/main/scala/chisel3/util/BitSet.scala
sequencer: I'm still thinking should this be sealed or public. 

If we seal it, user can't extend from it, if they need to add new functions, they have to use implicit class or wrapper another function to it. There is a use case which users can extend a `BitSet` in their own class, adding more function to it, for example: `AddressSet`, some functions from that class are too specific which should not be absorbed by chisel upstream, but an `AddressSet` is still a `BitSet`, and being able to use the `BitSet` decoder API.

src/main/scala/chisel3/util/BitSet.scala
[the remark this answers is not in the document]
sequencer: Yes, I think you are right.
I'll seal it :)

sequencer:
> Could/Should we make it experimental?

I think we should, however a non-experimental API `BitPat` depends on an experimental seems strange. So I experimental the `BitSet` object, and seal the `BitSet` API.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -161,14 +273,83 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
...
+              val nvalue = (remove.value ^ bit) & nmask
+              val nwidth = remove.width
+              new BitPat(nvalue, nmask, nwidth)
+            }
+          }
+      }
+    }
+  }
+
+  override def isEmpty: Boolean = false
```
jackkoenig, line 341: If this is always false, and terms are always `BitPat`, can `isEmpty` on trait `BitSet` ever be true?

sequencer: Oh. I think here is a bug. Should be mask = 0 here! My bad I’ll add a test to this.

jackkoenig:
I'd rather we could find a way to make the new stuff experimental, but as far as I can tell it's very awkward no matter what we do. So I'm fine with `BitSet` being non-experimental. I remain a bit nervous about there being issues with it that get hard to fix as a normal API, so please use this a bunch before we release 3.5.0 to find any issues. Regardless, let's keep both it and BitPat sealed.

jackkoenig:
> I'd rather we could find a way to make the new stuff experimental, but as far as I can tell it's very awkward no matter what we do. So I'm fine with BitSet being non-experimental. I remain a bit nervous about there being issues with it that get hard to fix as a normal API, so please use this a bunch before we release 3.5.0 to find any issues. Regardless, let's keep both it and BitPat sealed.

What I mean by this, is that even if we make `BitSet` experimental, some of the new `BitPat` methods are defined on non-experimental `BitPat` and thus we're sort of stuck with. They also return non-experimental `BitSet`.

That being said, I think an imperfect amount of experimental is still better than none. So I'm now thinking that we should keep object and class `BitSet` as experimental. It's slightly awkward, but it at least calls out that these new APIs are new. We can see how they do in 3.5 and make them regular chisel3.util in 3.6 (and let's make that be faster than 3.4 -> 3.5 was 🙂 )

src/main/scala/chisel3/util/BitPat.scala
sequencer: `isEmpty` for `BitPat` always returns `false` since `BitPat` cannot represent a semantic that no bit vector should belong can be covered by this `BitPat`. a `mask = 0` represents all bits are `?` which means all bit vector can be covered by this, while `mask = -1`(all bits are 1) represents only one point are covered.
Thus `BitPat` `isEmpty` is always false.

sequencer: In `BitSet` API:
```
def isEmpty: Boolean = terms.forall(_.isEmpty)
```
this means only empty set or a set which terms are all empty set are empty.

sequencer: For detail test, see:
https://github.com/chipsalliance/chisel3/blob/34f313e9d8c30171bc36c7a1f6993fb95efd9dd4/src/test/scala/chiselTests/util/BitSetSpec.scala#L20-L40

jackkoenig: Okay cool. This makes sense. That being said it'd be nice to see a couple of sanity checks like that `BitSet.empty.isEmpty` is true and that `BitSet("1100").intersect(BitSet("0011")).isEmpty` is also true.

jackkoenig: Ah didn't see the examples when I responded. Okay good, thanks!

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -111,6 +111,118 @@ object BitPat {
...
+    def terms: Set[BitPat]
+
+    /** Get specified width of said BitSet */
+    def getWidth: Int = {
+      require(terms.map(_.width).size <= 1, s"All BitPats must be the same size! Got $this")
+      // set width = 0 if terms is empty.
+      terms.headOption.map(_.width).getOrElse(0)
+    }
+
+    override def toString: String = terms.toSeq.sortBy((t: BitPat) => (t.mask, t.value)).mkString("\n")
```
jackkoenig, line 157: Ironically, this ties in to our discussion in https://github.com/chipsalliance/chisel3/pull/2217#discussion_r769245027. I think it'd be useful if we had a private implicit val typeclass implementation of the Order which we could use here as well.

sequencer: LOL yes, I remember I add order to BitPat is because I have been sorting it everywhere.

sequencer: I'm both OK to sort it everywhere to directly give a specific order to `BitPat`.

sequencer: I'd like to block this PR by #2217, I need to use that implicit import.

sequencer: this PR will be blocked by #2217, I need to get rid of that sortBy.

jackkoenig (approved):
LGTM! Just need to fix the CI failure.

sequencer:
I think I need to add additional tests on this PR, I will work on this today.

PR #2217, opened 2021-10-28
Refactor TruthTable to use Seq.

### LINKED ISSUE #2216: no default option for chisel3.util.experimental.decode.TruthTable
See **line 30** of `chisel3.util.experimental.decode.TruthTable.scala`:
```scala
object TruthTable {
  /** Parse TruthTable from its string representation. */
  def apply(tableString: String): TruthTable = {
    TruthTable(
      tableString
        .split("\n")
        .filter(_.contains("->"))
        .map(_.split("->").map(str => BitPat(s"b$str")))
        .map(bps => bps(0) -> bps(1))
        .toSeq,
      BitPat(s"b${tableString.split("\n").filterNot(_.contains("->")).head.replace(" ", "")}")
    )
  }

  /** Convert a table and default output into a [[TruthTable]]. */
  def apply(table: Iterable[(BitPat, BitPat)], default: BitPat): TruthTable = {
    require(table.map(_._1.getWidth).toSet.size == 1, "input width not equal.")
    require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
    val outputWidth = table.map(_._2.getWidth).head
    new TruthTable(table.toSeq.groupBy(_._1.toString).map { case (key, values) =>
      // merge same input inputs.
      values.head._1 -> BitPat(s"b${
        Seq.tabulate(outputWidth) { i =>
          val outputSet = values.map(_._2)
            .map(_.rawString)
            .map(_ (i))
            .toSet
            .filterNot(_ == '?')
          require(outputSet.size != 2, s"TruthTable conflict in :\n${values.map { case (i, o) => s"${i.rawString}->${o.rawString}" }.mkString("\n")}")
          outputSet.headOption.getOrElse('?')
        }.mkString
      }")
    }, default)
  }
```
There is no default option for the first `apply`.

sequencer:
In current TruthTable implementation, unordered Map will introduce non-deduped circuit.
This PR refactor TruthTable to use Seq instead of Map to produce a reproducible behavior in decoder.
After this PR:
https://scastie.scala-lang.org/Ee26d2aTTKqN6sFA6EJ6GQ
will return true.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -120,7 +120,7 @@ object BitPat {
   * "b10001".U === BitPat("b101??") // evaluates to false.B
   * }}}
   */
-sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends SourceInfoDoc {
+sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends SourceInfoDoc with Ordered[BitPat] {
   def getWidth: Int = width
   def apply(x: Int): BitPat = macro SourceInfoTransform.xArg
   def apply(x: Int, y: Int): BitPat = macro SourceInfoTransform.xyArg
@@ -169,6 +169,9 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
       case (_, false) => "?"
     }
   }.mkString
+
+  import scala.math.Ordered.orderingToOrdered
+  def compare(that: BitPat): Int = (getWidth, value, mask) compare (that.getWidth, that.value, that.mask)
 
   override def toString = s"BitPat($rawString)"
 }
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -8,6 +8,12 @@ import chisel3.internal.sourceinfo.{SourceInfo, SourceInfoTransform}
 
 
 object BitPat {
+
+  private[chisel3] implicit val bitPatOrder = new Ordering[BitPat] {
+    import scala.math.Ordered.orderingToOrdered
+    def compare(x: BitPat, y: BitPat): Int = (x.getWidth, x.value, x.mask) compare (y.getWidth, y.value, y.mask)
+  }
+
   /** Parses a bit pattern string into (bits, mask, width).
     *
     * @return bits the literal value, with don't cares being 0
@@ -120,7 +126,7 @@ object BitPat {
   * "b10001".U === BitPat("b101??") // evaluates to false.B
   * }}}
   */
-sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends SourceInfoDoc with Ordered[BitPat] {
+sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends SourceInfoDoc {
   def getWidth: Int = width
   def apply(x: Int): BitPat = macro SourceInfoTransform.xArg
   def apply(x: Int, y: Int): BitPat = macro SourceInfoTransform.xyArg
@@ -169,9 +175,6 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
       case (_, false) => "?"
     }
   }.mkString
-
-  import scala.math.Ordered.orderingToOrdered
-  def compare(that: BitPat): Int = (getWidth, value, mask) compare (that.getWidth, that.value, that.mask)
 
   override def toString = s"BitPat($rawString)"
 }
--- a/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
@@ -46,7 +46,7 @@ object TruthTable {
         }.mkString
       }")
     }.toSeq
-
+    import BitPat.bitPatOrder
     new TruthTable(if(sort) mergedTable.sorted else mergedTable, default, sort)
   }
 
--- a/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
@@ -4,7 +4,7 @@ package chisel3.util.experimental.decode
 
 import chisel3.util.BitPat
 
-sealed class TruthTable private (val table: Seq[(BitPat, BitPat)], val default: BitPat, val sorted: Boolean) {
+sealed class TruthTable private (val table: Seq[(BitPat, BitPat)], val default: BitPat, val sort: Boolean) {
   def inputWidth = table.head._1.getWidth
 
   def outputWidth = table.head._2.getWidth
@@ -16,7 +16,7 @@ sealed class TruthTable private (val table: Seq[(BitPat, BitPat)], val default: 
     (table.map(writeRow) ++ Seq(s"${" "*(inputWidth + 2)}${default.rawString}")).mkString("\n")
   }
 
-  def copy(table: Seq[(BitPat, BitPat)] = this.table, default: BitPat = this.default, sorted: Boolean = this.sorted) = TruthTable(table, default, sorted)
+  def copy(table: Seq[(BitPat, BitPat)] = this.table, default: BitPat = this.default, sort: Boolean = this.sort) = TruthTable(table, default, sort)
 
   override def equals(y: Any): Boolean = {
     y match {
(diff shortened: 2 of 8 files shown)


sequencer:
I got a better idea with Definition/Instance API.

jackkoenig (changes requested):
Don't use `.toMap` because it destroys the order you just created.

jackkoenig:
I think this PR is reasonable, but I want to checked a couple of things before we merge it.
1. What problem is this solving?
2. Are you sure that order and sorting is the right way to solve the issue? I'm assuming the order of BitPats can affect the result from espresso, is there utility in allowing the user to provide different orders to get different results?
3. Are you wanting to enforce uniqueness? The current code changing from `Map` to `Seq` no longer enforces the invariant that the keys be unique.

sequencer:
BTW I wanna slip a API change(only affect RC1) from a158e0e. I personally think this is a unsafe design, but I'm OK to remain as what it is.

jackkoenig:
> I wanna slip a API change(only affect RC1) from [a158e0e](https://github.com/chipsalliance/chisel3/commit/a158e0e79c15cd77c95b364799208249a3259f6f). I personally think this is a unsafe design, but I'm OK to remain as what it is.

Given this only affects RC1, I think the change is fine.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -28,25 +28,12 @@ final class TruthTable(val table: Map[BitPat, BitPat], val default: BitPat) {
...
-      BitPat(s"b${tableString.split("\n").filterNot(_.contains("->")).head.replace(" ", "")}")
-    )
-  }
-
   /** Convert a table and default output into a [[TruthTable]]. */
-  def apply(table: Iterable[(BitPat, BitPat)], default: BitPat): TruthTable = {
+  def apply(table: Iterable[(BitPat, BitPat)], default: BitPat, sorted: Boolean = true): TruthTable = {
     require(table.map(_._1.getWidth).toSet.size == 1, "input width not equal.")
     require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
     val outputWidth = table.map(_._2.getWidth).head
```
jackkoenig, line 35: Do we really want to trust the user has provided a sorted table? I think it might just be better to sort unconditionally.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -170,5 +170,8 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, width: Int) extends Sou
     }
   }.mkString
 
+  import scala.math.Ordered.orderingToOrdered
+  def compare(that: BitPat): Int = (getWidth, value, mask) compare (that.getWidth, that.value, that.mask)
```
jackkoenig, line 174: It makes me slightly nervous to expose the order of `BitPat` as a public API. Are we 100% sure this is the right, canonical way to order `BitPat`? It makes me a little nervous since if it's the wrong ordering, it's kind of awkward to change if users are sorting `BitPat` themselves. 

I'm not saying this is wrong, but could we instead use the `Ordering` typeclass and keep it private for purposes of sorting internally? No other type provided by Chisel provides `Ordered` so it just makes me a little nervous to do it here for the notionally unrelated purpose of making decoder output deterministic.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
sequencer: I added the default option sort=true
By default in the apply function it will sort user’s input. If user turn it off, which means user should take the unsorted risc(I should document this)

src/main/scala/chisel3/util/BitPat.scala
sequencer: > Are we 100% sure this is the right, canonical way to order BitPat?

this makes the order of BitPat unique. But I can still remove this to make BitPat unordered, and use sortBy in the place where we need this order. I think both are ok. Let you decide ;p

jackkoenig: Order being unique is important, but not sufficient, I could make the order be the `.hashCode` order but obviously that wouldn't be very intuitive 🙂. I'd rather `Order` not be part of the public API, but yes it should be used internally to sort things. You could do this with `sortBy`, but alternatively you could have a private (or package private) implicit val implementation of https://www.scala-lang.org/api/2.12.13/scala/math/Ordering.html and put this functionality there. Then you could import it and still use `.sorted`.

jackkoenig:
@sequencer do you know why the chiseltest integration test is failing?

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
jackkoenig: Ah I see, yeah please document what it does. Also, I'd call it `sort` rather than `sorted` because the latter implies to me that the input is already sorted rather than that we will sort for you.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -4,7 +4,7 @@ package chisel3.util.experimental.decode
 
 import chisel3.util.BitPat
 
-final class TruthTable(val table: Map[BitPat, BitPat], val default: BitPat) {
+sealed class TruthTable(val table: Seq[(BitPat, BitPat)], val default: BitPat) {
```
jackkoenig, line 7:
```suggestion
sealed class TruthTable private (val table: Seq[(BitPat, BitPat)], val default: BitPat) {
```
So that users need to construct this via the companion object apply method.

sequencer:
> @sequencer do you know why the chiseltest integration test is failing?

my bad, I fixed this.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -14,10 +14,10 @@ final class TruthTable(val table: Map[BitPat, BitPat], val default: BitPat) {
     def writeRow(map: (BitPat, BitPat)): String =
       s"${map._1.rawString}->${map._2.rawString}"
 
-    (table.map(writeRow) ++ Seq(s"${" "*(inputWidth + 2)}${default.rawString}")).toSeq.sorted.mkString("\n")
+    (table.map(writeRow) ++ Seq(s"${" "*(inputWidth + 2)}${default.rawString}")).mkString("\n")
   }
 
-  def copy(table: Map[BitPat, BitPat] = this.table, default: BitPat = this.default) = new TruthTable(table, default)
+  def copy(table: Seq[(BitPat, BitPat)] = this.table, default: BitPat = this.default) = new TruthTable(table, default)
```
jackkoenig, line 20: This method is an easy way to sneak in an unsorted `table`, should we sort here?

src/test/scala/chiselTests/util/experimental/TruthTableSpec.scala
```diff
@@ -52,12 +53,32 @@ class TruthTableSpec extends AnyFlatSpec {
...
+        (0 until 16).map{
+          i => BitPat(i.U(4.W)) -> BitPat((1<<i).U(16.W))
+        },
+        BitPat.dontCare(16)
+      )
+
+      io.out := decoder.qmc(io.in, table)
+    }
+    assert(chisel3.stage.ChiselStage.emitChirrtl(new Foo) == chisel3.stage.ChiselStage.emitChirrtl(new Foo))
+  }
```
jackkoenig, line 83: Good test 🙂

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
sequencer: nice, I'm thinking passing the sort option to internal state in `TurhtTable`

src/main/scala/chisel3/util/BitPat.scala
jackkoenig: To be more specific, what I'm asking for here is that instead of extending `Ordered`, we provide a private implementation of `Ordering` like so:
```scala
object BitSet {
  ...
  private implicit val bitSetOrder = new Ordering[BitSet] {
    def compare(that: BitPat): Int = (getWidth, value, mask) compare (that.getWidth, that.value, that.mask)
  }
  ...
}
```
Then we can import it and still use `.sort` internally.

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
jackkoenig: Do we need a copy method? Is it so bad to just make a new one via the factory method?

sequencer: This is used in QMC, since QMC need copy from:
https://github.com/chipsalliance/chisel3/blob/2b0bc0ecbc9c53882e2104ecd1e1387039be27f3/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala#L294-L306

I wonder should we expose this to user? currently I didn't come up with a user case.

src/main/scala/chisel3/util/BitPat.scala
sequencer: wow. I didn't think we can do this!
I will refactor it!

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
sequencer: sure ;p

jackkoenig (approved):
LGTM!

PR #2238, opened 2021-11-05
Use `0` instead of `?` to represent "not in ON set" in the results of QMC minimizer

yqszxx:
This PR changes the symbol which represent "not in ON set" in the results of QMC minimizer from `?` to `0`. The current behavior is correct when using minimizer alone and aligned with the behavior of espresso minimizer.
This PR does not change the behavior of the `decode` api nor verilog codes generated by it because the assumpsion made in QMC previously (`0` and `?` are treated the same way) about the `pla` api is correct and `decode` always use `pla` to generate actual hardware.

### Contributor Checklist

#### Type of Improvement

- bug fix

Does this change any generated Verilog?
No.

#### Desired Merge Strategy

- [ ] Did you review?


sequencer (changes requested):
This will reduce the optimization possibility for QMC.

PR #2239, opened 2021-11-08
refactor strongenum(not deprecate yet).

sequencer:


sequencer:
ask @jackkoenig to see if this changing is appropriate. If true, I'll continue fixing the tests.

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -2,74 +2,18 @@
...
-    * @param definition a map describing which integer values correspond to which enum names
-    */
-  case class EnumDefAnnotation(typeName: String, definition: Map[String, BigInt]) extends NoTargetAnnotation
-
-  case class EnumDefChiselAnnotation(typeName: String, definition: Map[String, BigInt]) extends ChiselAnnotation {
-    override def toFirrtl: Annotation = EnumDefAnnotation(typeName, definition)
-  }
-}
-import EnumAnnotations._
-
```
jackkoenig, line 72: I'm down with the basic idea, but why delete these? I think they should still exist but be deprecated. We should also consider having a command-line option to continue generating them (at least for 3.5) but we can maybe just see if anyone complains before adding that.

sequencer: Should I'll add them back
deprecate them form Chisel 3.5
add a command-line option to emit them(but by default turn it off)
then remove them in Chisel 3.6.

sequencer: @jackkoenig In #2240, I deprecate them from 3.4.x, and remove it in 3.5. Is this OK? or we should deprecate it since 3.5 and remove them in 3.6?

jackkoenig: Because it's so late in 3.5 release process (deleting anything else is another risk that might push back `3.5.0` release), I think we should deprecate in 3.5 and delete in 3.6.

Let's also make the 3.5->3.6 time much faster, 6 months or less this time!

sequencer:
blocked by #2310

PR #2240, opened 2021-11-08
deprecate strongenum annos

sequencer:


jackkoenig:
This is fine to merge once we've merged the other, related PRs

jackkoenig (changes requested):
Because we're now so related in the 3.5 release process, we should instead deprecate in 3.5. Sorry for now helping this catch the train but we don't want 3.5.0 to slip any more.

sequencer:
Sure. I'll deprecate it in 3.5

sequencer:
Work on this in another PR.

PR #2250, opened 2021-11-18
Instance pla

sequencer:
This PR switch pla with definition and instance API to reduce the generated pla code size.
### Contributor Checklist

#### Type of Improvement
- backend code generation

#### API Impact
None

PLA will be module rather than inlined.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/pla.scala
+++ b/src/main/scala/chisel3/util/pla.scala
@@ -3,37 +3,126 @@
 package chisel3.util
 
 import chisel3._
+import chisel3.experimental.hierarchy.{instantiable, public, Definition, Instance}
+
+/** Construct a [[https://en.wikipedia.org/wiki/Programmable_logic_array]] from specified table.
+  *
+  * Each position in the input matrix corresponds to an input variable where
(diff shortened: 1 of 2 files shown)


mwachs5 (commented):
as I said in #2253 , starting to use `Instance`s in a design is not something to do lightly. An `Instance` is *not* a `BaseModule` so code that is written to expect modules (like Injecting Aspects, Select API, etc) can break as soon as there is one `Instance` in a design.

Could there be a different form of the `apply()` which uses the  `Instance` cache?

I would not be as bothered by this as the `Queue` since I am guessing PLA are less common, but it seems heavy handed to take something `experimental` and put it into `util` without an opt-in feel

sequencer:
> Could there be a different form of the apply() which uses the Instance cache?

I guess we can maintain another package inside `hierarchy` package?

sequencer:
> since I am guessing PLA are less common. 

Not actually, in our designs, decoder almost exist everywhere, and PLA is a quite common part.

PR #2260, opened 2021-11-24
Make inline memory initialization the default

### LINKED ISSUE #1293: loadMemoryFromFile() should place readmemh() inline for better compatibility
`loadMemoryFromFile()` creates a separate file and uses a SV bind statement. A simple example:

```
import chisel3._
import chisel3.util.experimental.loadMemoryFromFile

class Foo(val bits: Int, val size: Int, filename: String) extends Module {
  val io = IO(new Bundle {
    val nia = Input(UInt(bits.W))
    val insn = Output(UInt(32.W))
  })

  val memory = Mem(size, UInt(32.W))
  io.insn := memory(io.nia >> 2);
  loadMemoryFromFile(memory, filename)
}

object FooObj extends App {
  chisel3.Driver.execute(Array[String](), () => new Foo(32, 1024, "insns.hex"))
}
```

```
# cat Foo.Foo.memory.v
module BindsTo_0_Foo(
  input         clock,
  input         reset,
  input  [31:0] io_nia,
  output [31:0] io_insn
);

initial begin
  $readmemh("insns.hex", Foo.memory);
end
                      endmodule

bind Foo BindsTo_0_Foo BindsTo_0_Foo_Inst(.*);
```

Yosys doesn't like this, and likely there are other tools that don't either. Is there any reason we don't just place it inline?

carlosedp:
Fixes #1293

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This PR changes the default memory initialization based on files by using the inline `readmem[bh]` statements instead of the previous SV Bind module and also moves the API to `chisel3.util`.

The bind module initialization is available on `chisel3.util.experimental` as `loadMemoryFromFileBind`.

No changes on backend, just API  to generate initialization.

#### Desired Merge Strategy

#### Release Notes

This PR changes the default memory initialization based on external files by using the inline `readmem[bh]` statements instead of the previous SV Bind module and also moves the API to `chisel3.util`.

- [ ] Did you review?


sequencer:
This seems contains a breaking change? There should be a deprecation flow.

ekiwi:
> This seems contains a breaking change? There should be a deprecation flow.

Good point. How would you suggest we approach this?

src/main/scala/chisel3/util/LoadMemoryTransform.scala
```diff
@@ -0,0 +1,92 @@
...
+  *
+  * A binary file can be similarly constructed.
+  * Chisel does not validate the file format or existence. It is supposed to be in a path accessible by the synthesis
+  * tool together with the generated Verilog.
+  *
+  * @see Chisel3 Wiki entry on
+  * [[https://github.com/freechipsproject/chisel3/wiki/Chisel-Memories#loading-memories-in-simulation "Loading Memories
+  * in Simulation"]]
+  */
+object loadMemoryFromFile {
```
sequencer, line 75: I didn't see any `LoadMemoryTransform` here?
Is the file name correct?
I guess `loadMemoryFromFile` is just a rename from `loadMemoryFromFileInline`?
This introduce a API breaking change, and seems didn't do anything else?

sequencer:
Seems currently this PR is a draft PR @carlosedp? Sorry I merged master into this(I thought it's done)
For that breaking change, I think you can don't touch the original API, and implement your new API, after merging, creating a new deprecating PR to deprecate the API you think it should be removed, then, we can discuss: should we deprecate that API?(I personally like getting that API removed)

ekiwi:
> For that breaking change, I think you can don't touch the original API, and implement your new API, after merging, creating a new deprecating PR to deprecate the API you think it should be removed, then, we can discuss: should we deprecate that API?(I personally like getting that API removed)

@sequencer : I told @carlosedp to make a PR in order to get the discussion started. Essentially what this PR does is to make a new public API `chisel3.util.loadMemoryFromFile` which will use the emitter option to generate an inline readmem call. The old `experimental` API is then removed.

ekiwi:
While we are talking about this, I just remembered one of the missing features which are needed before we can make "inline" memory loading the default:
- currently there is no way to initialize a memory of a "complex" (i.e. bundle or vector) type, which is something that was supported by the old functionality (albeit in a rather awkward fashion)

carlosedp:
Ah yes, you mean the problem pointed-out by https://github.com/chipsalliance/chisel3/issues/1289 right?

sequencer:
> I told @carlosedp to make a PR in order to get the discussion started. Essentially what this PR does is to make a new public API chisel3.util.loadMemoryFromFile which will use the emitter option to generate an inline readmem call. The old experimental API is then removed.

I see, I didn't object keeping deprecation flow for `experimental` API. I think if @jackkoenig approve this, we should directly merge this ;p

carlosedp:
Ping... news about this?

PR #2261, opened 2021-11-25
Create ChiselEnum1H where elements have one-hot encoding IDs

carlosedp:
This PR adds a One-Hot encoded ChiselEnum where the elements have 1H instead of sequential IDs:

```scala
// It simplifies:
object InstructionType extends ChiselEnum {
  val INST_I = Value((1 << 0).U)
  val INST_S = Value((1 << 1).U)
  val INST_B = Value((1 << 2).U)
  val INST_U = Value((1 << 3).U)
  val INST_J = Value((1 << 4).U)
  val INST_Z = Value((1 << 5).U)
  val INST_R = Value((1 << 6).U)
  val IN_ERR = Value((1 << 7).U)
}

// into:

object InstructionType extends ChiselEnum1H {
	val INST_I, INST_S, INST_U, INST_B, INST_J, INST_Z, INST_R, IN_ERR = Value
}

// Allowing the use in Mux1H for example as:
  val selectorVal = Wire(UInt(InstructionType.getWidth.W))
  selectorVal := InstructionType.INST_I.asUInt
  val hotValue = chisel3.util.Mux1H(
    selectorVal,
    Seq(
      2.U,
      4.U,
      8.U,
      11.U,
      15.U
    )
  )
```

The PR is for discussion and in case of greenlit, I'll add test cases for it.

### Contributor Checklist

#### Type of Improvement

- documentation
- new feature/API

#### API Impact

The PR creates a ChiselEnum where elements have One-Hot IDs.

None.

#### Desired Merge Strategy

#### Release Notes

This PR adds a One-Hot encoded ChiselEnum where the elements have One-Hot IDs instead of sequential. This change pairs well with `Mux1H`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -233,6 +233,9 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
   def toPrintable: Printable = FullName(this) // TODO: Find a better pretty printer
 }
 
+trait EnumFactory1H extends EnumFactory{
+  override val oneHot: Boolean = true
+}
 
 abstract class EnumFactory {
   class Type extends EnumType(this)
@@ -242,6 +245,7 @@ abstract class EnumFactory {
 
   private var id: BigInt = 0
   private[chisel3] var width: Width = 0.W
+  val oneHot: Boolean = false
 
   private case class EnumRecord(inst: Type, name: String)
   private val enumRecords = mutable.ArrayBuffer.empty[EnumRecord]
@@ -275,13 +279,13 @@ abstract class EnumFactory {
   protected def do_Value(name: String): Type = {
     val result = new Type
 
+    val resId = if (!oneHot) id else BigInt(1 << id.toInt)
+
     // We have to use UnknownWidth here, because we don't actually know what the final width will be
-    result.bindToLiteral(id, UnknownWidth())
-
-    enumRecords.append(EnumRecord(result, name))
-
-    width = (1 max id.bitLength).W
+    result.bindToLiteral(resId, UnknownWidth())
+    width = (1 max resId.bitLength).W
     id += 1
+    enumRecords.append(EnumRecord(result, name))
 
     result
   }
@@ -343,7 +347,6 @@ abstract class EnumFactory {
     (t, t.isValid)
   }
 }
-
 
 private[chisel3] object EnumMacros {
   def ValImpl(c: Context) : c.Tree = {
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -279,7 +279,7 @@ abstract class EnumFactory {
   protected def do_Value(name: String): Type = {
     val result = new Type
 
-    val resId = if (!oneHot) id else BigInt(1 << id.toInt)
+    val resId = if (!oneHot) id else BigInt(1) << id.toInt
 
     // We have to use UnknownWidth here, because we don't actually know what the final width will be
     result.bindToLiteral(resId, UnknownWidth())
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -234,7 +234,7 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
 }
 
 trait EnumFactory1H extends EnumFactory{
-  override val oneHot: Boolean = true
+  final override val oneHot: Boolean = true
 }
 
 abstract class EnumFactory {
@@ -245,7 +245,7 @@ abstract class EnumFactory {
 
   private var id: BigInt = 0
   private[chisel3] var width: Width = 0.W
-  val oneHot: Boolean = false
+  private[chisel3] val oneHot: Boolean = false
 
   private case class EnumRecord(inst: Type, name: String)
   private val enumRecords = mutable.ArrayBuffer.empty[EnumRecord]
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -86,6 +86,10 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
     }
   }
 
+  def toIndex: Int = {
+    factory.all.indexOf(this)
+  }
+
   override def cloneType: this.type = factory().asInstanceOf[this.type]
 
   private[chisel3] def compop(sourceInfo: SourceInfo, op: PrimOp, other: EnumType): Bool = {
--- a/core/src/main/scala/chisel3/experimental/package.scala
+++ b/core/src/main/scala/chisel3/experimental/package.scala
@@ -24,6 +24,13 @@ package object experimental {
 
   type ChiselEnum = EnumFactory
 
+  /**
+   * ChiselEnum1H extends the ChiselEnum type for creating enums with
+   * one-hot encoding IDs.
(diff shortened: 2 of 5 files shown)


core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -23,6 +23,13 @@ package object experimental {
 
   type ChiselEnum = EnumFactory
 
+  /**
+   * ChiselEnum1H extends the ChiselEnum type for creating enums with
+   * one-hot encoding IDs.
+   * This enum type pairs well with [[Mux1H]] for parallel mux processing.
+   */
```
sequencer, line 30: I guess ScalaDoc doesn't like this, use
```
/**
  *
  */
```
instead.

src/main/scala/chisel3/util/Mux.scala
```diff
@@ -17,6 +18,21 @@ import chisel3._
...
+  *   val s0, s1, s2, s3 = Value
+  * }
+  * val selectorVal = Wire(UInt(selector.getWidth.W))
+  * selectorVal := selector.s1.asUInt
+  * val hotValue = chisel3.util.Mux1H(selectorVal.asUInt,
+      Seq(
+  *     2.U,
+  *     4.U,
+  *     8.U,
+  *     11.U,
```
sequencer, line 34: I think we need add a sugar for Mux1H to consume a Enum type and return a Enum type.
These magic num is strange, while seems not correct: `11.U -> 16.U`

sequencer:
My further question to this is: how to design the encoding API? 1H API implicitly contains an assertion: `PopCount === 1.U`. 
And how do we resolve this general encoding(ECC, 8b/10b, Gray Code, Hamming Code, etc...) issue?
Should we directly add corresponding assertion to this, which makes verification easier?
Or should we annotate the these encoding signal with an Annotation? and optionally add a Transform to these signals?
And how to propagate the encoding type? In `bind` function or propagate in FIRRTL transform?

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -232,6 +232,9 @@ abstract class EnumType(private val factory: EnumFactory, selfAnnotating: Boolea
   def toPrintable: Printable = FullName(this) // TODO: Find a better pretty printer
 }
 
+trait EnumFactory1H extends EnumFactory{
+  override val oneHot: Boolean = true
```
sequencer, line 236: Should be final override

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -241,6 +244,7 @@ abstract class EnumFactory {
 
   private var id: BigInt = 0
   private[chisel3] var width: Width = 0.W
+  val oneHot: Boolean = false
```
sequencer, line 247: We may need private this

src/main/scala/chisel3/util/Mux.scala
sequencer: This can be done in future PR(just comment for recording)

sequencer (changes requested):
Generally LGTM. 
I strongly depend on EnumOH, it is really good for code simplification. 
The only thing we should do is to private  the non-user api. 
in the future, we should figure out a better  api to integrate EnumOH to MuxOH.

core/src/main/scala/chisel3/StrongEnum.scala
carlosedp: Done!

core/src/main/scala/chisel3/StrongEnum.scala
carlosedp: Done!

sequencer:
@carlosedp would you mind adding some test to demonstrate the usage?

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -275,13 +279,13 @@ abstract class EnumFactory {
   protected def do_Value(name: String): Type = {
     val result = new Type
 
-    // We have to use UnknownWidth here, because we don't actually know what the final width will be
-    result.bindToLiteral(id, UnknownWidth())
-
-    enumRecords.append(EnumRecord(result, name))
+    val resId = if (!oneHot) id else BigInt(1 << id.toInt)
```
jackkoenig, line 282:
```suggestion
    val resId = if (!oneHot) id else BigInt(1) << id.toInt
```
You need to make sure to left shift the `BigInt`, otherwise you're left shifting the `Int` which can overflow.

carlosedp: Thanks Jack!

carlosedp:
Ping... news on this?

src/main/scala/chisel3/util/Mux.scala
sequencer: I think we need add a sugar for Mux1H to consume a Enum type and return a Enum type.
These magic num is strange, while seems not correct: 11.U -> 16.U

src/test/scala/chiselTests/OneHotMuxSpec.scala
```diff
@@ -304,3 +308,22 @@ class UIntToOHTester extends BasicTester {
...
+     val s0, s1, s2, s3 = Value
+   }
+   val selectorVal = Wire(UInt(selector.getWidth.W))
+   selectorVal := selector.s1.asUInt
+   val hotValue = chisel3.util.Mux1H(selectorVal.asUInt,
+      Seq(
+       2.U,
+       4.U,
+       8.U,
+       11.U,
```
sequencer, line 324: 11 -> 16

sequencer (changes requested):
a little nitpick.

src/test/scala/chiselTests/OneHotMuxSpec.scala
carlosedp: Done :)

carlosedp:
In the sample below, I really don't like the line to set the selector width...

```scala
   object selector extends chisel3.experimental.ChiselEnum1H {
     val s0, s1, s2, s3 = Value
   }
   val selectorVal = Wire(UInt(selector.getWidth.W)) // <--- don't like this
   selectorVal := selector.s1.asUInt
   val hotValue = chisel3.util.Mux1H(selectorVal.asUInt,
      Seq(
       2.U,
       4.U,
       8.U,
       16.U,
     ))
```

Any ideas on improving this?

ekiwi:
> Any ideas on improving this?

Some thoughts:

1) It seems to me like Mux1H should be changed to accept `ChiselEnum1H` instead of a `UInt` directly.
2) You should be able to directly use your `selector` type as a type and thus write something like `val selectorVal = Wire(selector)`.
3) It might also make sense to have a `toIndex` function for `ChiselEnum1H` such that e.g. `assert(selector.INST_S.toIndex === 1.U)`

carlosedp:
For 1, I could send another PR... ok?
On 2, you mean the way it's already implemented I could use like this?
Working on 3.

ekiwi:
> For 1, I could send another PR... ok?

I would prefer to see everything work together in a single PR.



> On 2, you mean the way it's already implemented I could use like this?

Yes. Can you try to change your example and see if it works?



> Working on 3.

Awesome

carlosedp:
There it go! Have Mux1H accept ChiselEnum1H and added the toIndex function.
Also example is pretty simple now:

```scala
  object selector extends chisel3.experimental.ChiselEnum1H {
    val s0, s1, s2, s3 = Value
  }
  val selectorVal = selector.s1
  val hotValue = chisel3.util.Mux1H(selectorVal,
      Seq(
      2.U,
      4.U,
      8.U,
      16.U,
    ))
  // hotValue is 4.U

```

PR #2271, opened 2021-12-01
Require the chisel3 compiler plugin

jackkoenig:
As the chisel3 compiler plugin is now required, we can delete unused
code for reflective autoclonetype as well as the noPluginTests.

Fixes https://github.com/chipsalliance/chisel3/issues/1802

The method of checking that compiler plugin is in use is checking that `_usingPlugin` is set for all `Bundle`s. This is a runtime check and something the user could fake by implemented the protected method themselves. It's not a perfect check since it also wouldn't catch code that doesn't use any `Bundles`, but I think it's at least a decent attempt, maybe we can come up with better ways if this is insufficient.

There's more we can delete (we can probably delete/tweak signal naming code), but this is a pretty good chunk.

Deleting the 3 `val`/`var`s in `Bundle` that were needed for reflective autoclonetype does reduce the size of every `Bundle` object by 12-bytes which is a nice little bonus here.

I need to do a parallel PR deprecating not using the plugin in 3.4.

### Contributor Checklist

#### Type of Improvement

   - code refactoring  
  
#### API Impact

This breaks any Chisel code being compiled without the compiler plugin. Users not using the plugin then starting to use it may run into the slight API change due to autoclonetype2 described in the Chisel v3.4.3 release notes: https://github.com/chipsalliance/chisel3/releases/tag/v3.4.3.

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

The chisel3 compiler plugin is now required. See https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects for instructions on how to add it to your project.

- [ ] Did you review?


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1142,224 +1136,10 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
-    val clone = classMirror.reflectConstructor(ctor).apply(ctorParamsVals:_*).asInstanceOf[this.type]
-    clone._outerInst = this._outerInst
-
-    validateClone(clone,
-      "Constructor argument values were inferred:" +
-        " ensure that variable names are consistent and have the same value throughout the constructor chain," +
-        " and that the constructor is deterministic."
-    )
-    clone
+    throwException(s"Internal Error! This should have been implemented by the chisel3-plugin")
```
mwachs5, line 1142:
```suggestion
    throwException(s"Internal Error! This should have been implemented by the chisel3-plugin. Please file an issue against chisel3")
```

mwachs5 (approved):
🎉

PR #2273, opened 2021-12-01
Deprecate nonuse of the chisel3 compiler plugin

jackkoenig:
Companion to https://github.com/chipsalliance/chisel3/pull/2271, deprecating nonuse of the plugin for release in 3.4.5.

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -731,6 +731,13 @@ package experimental {
   * }}}
   */
 abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
+  if (!_usingPlugin) {
+    Builder.deprecated(
+      "Compiling Chisel code without the compiler plugin is now deprecated. " +
+      "Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects."
+    )
+  }
+
   override def className: String = this.getClass.getSimpleName match {
     case name if name.startsWith("$anon$") => "AnonymousBundle"  // fallback for anonymous Bundle case
     case "" => "AnonymousBundle"  // ditto, but on other platforms


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -731,6 +731,13 @@ package experimental {
   * }}}
   */
 abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
+  if (!_usingPlugin) {
+    Builder.deprecated(
```
mwachs5, line 735: does this dump to the screen every time you build a `Bundle`??? How many times are you going to get this message?

jackkoenig: Good question, I'm looking into it.

jackkoenig: Building [`DecoupledGCD` from chisel-template](https://github.com/freechipsproject/chisel-template/blob/f5f33c69f04a64531cbdb31581e09b95583fba91/src/main/scala/gcd/DecoupledGCD.scala#L26) prints the following with this PR:

```
[deprecated] DecoupledGCD.scala:8 (4 calls): Compiling Chisel code without the compiler plugin is now deprecated. Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects.
[deprecated] DecoupledGCD.scala:27 (3 calls): Compiling Chisel code without the compiler plugin is now deprecated. Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects.
[deprecated] DecoupledGCD.scala:27 (3 calls): The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  Use the Compiler Plugin to infer cloneType
[deprecated] DecoupledGCD.scala:13 (3 calls): Compiling Chisel code without the compiler plugin is now deprecated. Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects.
[deprecated] DecoupledGCD.scala:28 (2 calls): Compiling Chisel code without the compiler plugin is now deprecated. Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects.
[deprecated] DecoupledGCD.scala:28 (2 calls): The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  Use the Compiler Plugin to infer cloneType
```

Basically, it will print 1 line for every single `Bundle` you use that was compiled without the plugin. This could obviously get extremely annoying extremely fast. Worse than that, it's including Bundles defined in `chisel3.util` although pointing to the user's use of the built-in (eg. `Decoupled`).

One might also notice that there are 2 different deprecation warnings though, there is also
```
[deprecated] DecoupledGCD.scala:28 (2 calls): The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  Use the Compiler Plugin to infer cloneType
```

When I don't include this change and build the same config:
```
[deprecated] DecoupledGCD.scala:27 (3 calls): The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  Use the Compiler Plugin to infer cloneType
[deprecated] DecoupledGCD.scala:28 (2 calls): The runtime reflection inference for cloneType (_cloneTypeImpl) is deprecated as of Chisel 3.5:  Use the Compiler Plugin to infer cloneType
```

Thus, not using the plugin is _already_ deprecated, so there's no need for this PR.

PR #2274, opened 2021-12-01
Make stuff in IR.scala package private

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

 - code cleanup

#### API Impact
Stuff that should not have been part of the public API in IR.scala was made package private.

None
#### Desired Merge Strategy

#### Release Notes

Classes in IR.scala were made package private and removed from the public API.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -221,7 +221,7 @@ object Width {
   def apply(): Width = UnknownWidth()
 }
 
-sealed private[chisel3] abstract class Width {
+sealed abstract class Width {
   type W = Int
   def min(that: Width): Width = this.op(that, _ min _)
   def max(that: Width): Width = this.op(that, _ max _)
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -15,11 +15,11 @@ import scala.collection.immutable.NumericRange
 import scala.math.BigDecimal.RoundingMode
 
 
-case class PrimOp(name: String) {
+private[chisel3] case class PrimOp(name: String) {
   override def toString: String = name
 }
 
-object PrimOp {
+private[chisel3] object PrimOp {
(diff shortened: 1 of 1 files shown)


jackkoenig:
Unfortunately, `Width` is actually used in the public API (which is annoying): https://github.com/chipsalliance/chisel3/blob/392ea3c9b5b04e374eeb1bf3b0d87ac9fbf45513/core/src/main/scala/chisel3/package.scala#L154

So at least that needs to stay public but separately we should pull that into package chisel3.

sequencer:
Should these API being deprecated at first? Although I believe there will be no user using those API.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -242,7 +243,7 @@ sealed case class UnknownWidth() extends Width {
   override def toString: String = ""
 }
 
-sealed case class KnownWidth(value: Int) extends Width {
+sealed private[chisel3] case class KnownWidth(value: Int) extends Width {
```
jackkoenig, line 246:
```suggestion
sealed case class KnownWidth(value: Int) extends Width {
```
This also needs to remain public since `.widthOption` returns `Width` which could be a `KnownWidth` (see used by chiseltest: https://github.com/ucb-bar/chisel-testers2/blob/b5a34f18be52d7879ffbdb0468c724a4d539f900/src/main/scala/chiseltest/iotesters/PeekPokeTester.scala#L123)

jackkoenig:
> Should these API being deprecated at first? Although I believe there will be no user using those API.

I would usually say yes, but I think in this case we can just go for it and see if anything breaks. We don't publish anything in `chisel3.internal` to the public API docs. Unfortunately a few of these nodes are used in the public API (Width, BinaryPoint, and Circuit) but for the rest, maybe we should just see if we can make them private.

jackkoenig:
Tested with all upstream Chisel projects, I'm going to merge this for RC2 and hope for the best since this stuff never should have been public. If we get push back we can revert before .0.

jackkoenig (approved):

PR #2277, opened 2021-12-02
Implement DataViews for Seq and Tuple

jackkoenig:
Fixes https://github.com/chipsalliance/chisel3/issues/864 (at long last!)

Implemented in new chisel3.experimental.conversions package. Includes:
* Implementations of DataProduct
* Implementations of DataView from Seq to Vec and Tuple to HWTuple
* Implicit conversions from Seq to Vec and Tuple to HWTuple

Examples are buried in the tests, but this makes it possible to write stuff like:
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val sel = IO(Input(Bool()))
  val y, z = IO(Output(UInt(8.W)))
  (y, z) := Mux(sel, (a, b), (c, d))
}
```
and
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val out = IO(Output(Vec(4, UInt(8.W)))
  out := Seq(a, b, c, d)
}
```

I still haven't implemented more than just Tuple2, I intend to do several, maybe up to like Tuple7 or something.

Opening this before finishing it because I'd like feedback on the approach. 

#### Do we like the implicit conversions?

They feel natural but I know folks are generally skeptical of implicit conversions.
An alternative is to provide `.asData` as an extension method, something like:
```scala
implicit class CanConvertToData[A : DataProduct](a: A) {
  def asData[B <: Data](implicit dataview: DataView[A, B]): B = a.viewAs[B]
}
```

This would change the above examples slightly:
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val sel = IO(Input(Bool()))
  val y, z = IO(Output(UInt(8.W)))
  (y, z).asData := Mux(sel, (a, b).asData, (c, d).asData)
}
```
and
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val out = IO(Output(Vec(4, UInt(8.W)))
  out := Seq(a, b, c, d).asData
}
```

You can see the total impact of this change on existing tests in this diff: https://gist.github.com/jackkoenig/ae5d20b18b4df2871aa174f0a38551a0

There is at least one other example where the implicit conversion doesn't work great but `.asData` does:
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val w, x, y, z = IO(Output(UInt(8.W)))

  // A little annoying that we need the type annotation on VecInit to get the implicit conversion to work
  // Note that one can just use the Seq on the RHS so there is an alternative (may lack discoverability)
  // We could also overload `VecInit` instead of relying on the implicit conversion
  Seq((w, x), (y, z)) := VecInit[HWTuple2[UInt, UInt]]((a, b), (c, d))
}
```
vs.
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val w, x, y, z = IO(Output(UInt(8.W)))

  Seq((w, x), (y, z)).asData := VecInit((a, b).asData, (c, d).asData)
}
```

As alluded to in the comment though, `VecInit` actually becomes obsolete if we have the implicit conversion, so perhaps this example isn't particularly useful.

#### Do we like the package?

```scala
import chisel3.experimental.conversions._
```

They could live directly `chisel3.experimental.dataview`, or it could be a subpackage of `dataview`. I think the `DataProducts` and the `DataViews` should live together but it's not clear if the implicit conversions (assuming we keep them) should live in the same package.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/experimental/conversions/package.scala
+++ b/src/main/scala/chisel3/experimental/conversions/package.scala
@@ -2,72 +2,18 @@
 package chisel3.experimental
 
 import chisel3._
-import chisel3.experimental.DataMirror.internal.chiselTypeClone
 import chisel3.experimental.dataview._
 import scala.language.implicitConversions
 
-/** Default implementations of [[DataProduct]] and [[DataView]] as well as implicit conversions to [[Data]] for some
-  * Scala standard library types
+/** Implicit conversions from some Scala standard library types and [[Data]]
   *
   * @note As this leans heavily on the experimental [[DataView]] feature, these APIs are experimental and subject to change
   */
 package object conversions {
 
-  // ****************************** Seq ******************************
-
-  /** Provides implementations of [[DataProduct]] for any `Seq[A]` where `A` has an implementation of `DataProduct`. */
-  implicit def seqDataProduct[A : DataProduct]: DataProduct[Seq[A]] = new DataProduct[Seq[A]] {
-    def dataIterator(a: Seq[A], path: String): Iterator[(Data, String)] = {
-      val dpa = implicitly[DataProduct[A]]
-      a.iterator
-        .zipWithIndex
-        .flatMap { case (elt, idx) =>
-          dpa.dataIterator(elt, s"$path[$idx]")
-        }
-    }
-  }
-
-  /** Provides `DataView[Seq[A], Vec[B]]` for all `A` such that there exists `DataView[A, B]` */
-  implicit def seqAsVecDataView[A : DataProduct, B <: Data](implicit dv: DataView[A, B]): DataView[Seq[A], Vec[B]] =
-    // TODO this would need a better way to determine the prototype for the Vec
-    DataView.mapping[Seq[A], Vec[B]](
-      xs => Vec(xs.size, chiselTypeClone(xs.head.viewAs[B])), // xs.head is not correct in general
-      { case (s, v) => s.zip(v).map { case (a, b) => a.viewAs[B] -> b } }
-    )
-
   /** Implicit conversion between `Seq` and `Vec` */
-  implicit def seq2Vec[A : DataProduct, B <: Data](xs: Seq[A])(implicit dv: DataView[A, B]): Vec[B] =
+  implicit def seq2vec[A : DataProduct, B <: Data](xs: Seq[A])(implicit dv: DataView[A, B]): Vec[B] =
     xs.viewAs[Vec[B]]
-
-  // ****************************** Tuple2 ******************************
-
-  /** [[Data]] equivalent of Scala's [[Tuple2]]
-    *
-    * Users may not instantiate this class directly. Instead they should use the implicit conversion from `Tuple2`
-    */
-  final class HWTuple2[+A <: Data, +B <: Data] private[conversions] (val _1: A, val _2: B) extends Bundle
-
-  /** Provides implementations of [[DataView]] for `(A, B)` to `HWTuple2[A, B]`  */
-  implicit def view[T1 : DataProduct, T2 : DataProduct, V1 <: Data, V2 <: Data](
-    implicit v1: DataView[T1, V1], v2: DataView[T2, V2]
-  ): DataView[(T1, T2), HWTuple2[V1, V2]] =
-    DataView.mapping(
-      { case (a, b) => new HWTuple2(a.viewAs[V1].cloneType, b.viewAs[V2].cloneType)},
-      { case ((a, b), hwt) =>
-        Seq(a.viewAs[V1] -> hwt._1,
-          b.viewAs[V2] -> hwt._2)
-      }
-    )
-
-  /** Provides implementations of [[DataProduct]] for any [[Tuple2]] where each field has an implementation of `DataProduct`. */
-  implicit def tuple2DataProduct[A : DataProduct, B : DataProduct]: DataProduct[(A, B)] = new DataProduct[(A, B)] {
-    def dataIterator(tup: (A, B), path: String): Iterator[(Data, String)] = {
-      val dpa = implicitly[DataProduct[A]]
-      val dpb = implicitly[DataProduct[B]]
-      val (a, b) = tup
-      dpa.dataIterator(a, s"$path._1") ++ dpb.dataIterator(b, s"$path._2")
-    }
-  }
 
   /** Implicit conversion between `(A, B)` and `HWTuple2[A, B]` */
   implicit def tuple2hwtuple[T1 : DataProduct, T2 : DataProduct, V1 <: Data, V2 <: Data](
--- a/src/main/scala/chisel3/experimental/conversions/package.scala
+++ b/src/main/scala/chisel3/experimental/conversions/package.scala
@@ -1,0 +1,80 @@
+
+package chisel3.experimental
+
+import chisel3._
+import chisel3.experimental.DataMirror.internal.chiselTypeClone
+import chisel3.experimental.dataview._
+import scala.language.implicitConversions
+
+/** Default implementations of [[DataProduct]] and [[DataView]] as well as implicit conversions to [[Data]] for some
+  * Scala standard library types
+  *
+  * @note As this leans heavily on the experimental [[DataView]] feature, these APIs are experimental and subject to change
+  */
+package object conversions {
+
+  // ****************************** Seq ******************************
+
+  /** Provides implementations of [[DataProduct]] for any `Seq[A]` where `A` has an implementation of `DataProduct`. */
+  implicit def seqDataProduct[A : DataProduct]: DataProduct[Seq[A]] = new DataProduct[Seq[A]] {
+    def dataIterator(a: Seq[A], path: String): Iterator[(Data, String)] = {
+      val dpa = implicitly[DataProduct[A]]
+      a.iterator
+        .zipWithIndex
+        .flatMap { case (elt, idx) =>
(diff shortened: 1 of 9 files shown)


jackkoenig:
I'm curious what users think @mwachs5, @aswaterman, @carlosedp, @ingallsj, @schoeberl, @sequencer

mwachs5:
> Do we like the implicit conversions?

One of my biggest user issues with the implicit conversions is its hard to understand why *my* code doesn't work when *your* code does. If there is a function it's easy to figure out in IDE how to include the function, but if an implicit conversion is missing, it's hard to discover.

What is the advantage of the example:
```
  val a, b, c, d = IO(Input(UInt(8.W)))
  val out = IO(Output(Vec(4, UInt(8.W)))
  out := Seq(a, b, c, d).asData
}
```
Is it just saving the few characters of writing `VecInit`, or does this change the output in some way? If the former this seems like a pattern that would again just confuse people.

The `Mux` example is much more compelling IMHO, can you shoudl what that woudl look like if we forced the `.toData` thing?

```
  val a, b, c, d = IO(Input(UInt(8.W)))
  val sel = IO(Input(Bool()))
  val y, z = IO(Output(UInt(8.W)))
  (y, z) := Mux(sel, (a, b), (c, d))
}
```

jackkoenig:
> > Do we like the implicit conversions?
> 
> One of my biggest user issues with the implicit conversions is its hard to understand why _my_ code doesn't work when _your_ code does. If there is a function it's easy to figure out in IDE how to include the function, but if an implicit conversion is missing, it's hard to discover.

Yeah it's a fair concern.

I should revise my proposal a bit, I think we should definitely have `.asData` whether or not we have the implicit conversions. Then, if something isn't working for a user, they can always use the explicit `.asData`. That doesn't resolve your concern, but maybe it helps.


> What is the advantage of the example:
> 
> ```
> class MyModule extends Module {
>   val a, b, c, d = IO(Input(UInt(8.W)))
>   val out = IO(Output(Vec(4, UInt(8.W)))
>   out := Seq(a, b, c, d).asData
> }
> ```
> 
> Is it just saving the few characters of writing `VecInit`, or does this change the output in some way? If the former this seems like a pattern that would again just confuse people.

In this case, it's true that `VecInit(a, b, c, d)` would accomplish the same thing. But if we flipped the directions of everything, `DataView` allows connecting to the underlying elements:
```scala
// This would work
Seq(a, b, c, d).asData := out
// This would just connect to an intermediate Wire and thus be dropped
VecInit(a, b, c, d) := out // bad! don't do this!
```

> The `Mux` example is much more compelling IMHO, can you shoudl what that woudl look like if we forced the `.toData` thing?
> 
> ```
> class MyModule extends Module {
>   val a, b, c, d = IO(Input(UInt(8.W)))
>   val sel = IO(Input(Bool()))
>   val y, z = IO(Output(UInt(8.W)))
>   (y, z) := Mux(sel, (a, b), (c, d))
> }
> ```

I had it above but it would look like this:
```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val sel = IO(Input(Bool()))
  val y, z = IO(Output(UInt(8.W)))
  (y, z).asData := Mux(sel, (a, b).asData, (c, d).asData)
}
```

This is actually the case that made me add the implicit conversion in the first place. The nice thing about the implicit conversion is that any place a user writes code that accepts `Data` or `T <: Data`, it will _just work_ if passed Seqs or Tuples of `Data`. Without the implicit conversion, people can of course manually cast each thing, and method writers could of course add their own version of the method that will do the conversion. For example, in the case of Mux:
```scala
object Mux {
  // Simplified since we also have SourceInfo and CompileOptions implicits here, and a do_ macro
  def apply[T <: Data](cond: Bool, con: T, alt: T): T = ???
}
```
One could add the following additional methods so that it will do the conversion rather than requiring an implicit:
```scala
  def apply[A : DataProduct, T <: Data](cond: Bool, con: T, alt: T)(implicit ev: DataView[A, T]) : T =
    apply(cond, con.viewAs[T], alt.viewAs[T])
```
I think this is fine to do inside of Chisel itself, but I'm not super keen to suggest users do the same. Thus in practice, most user-defined APIs would require explicit `.asData` for non-Data-but-viewable-as-Data arguments

jackkoenig:
It's occurred to me that `.asData` is identical to just `.viewAs` except I'm stylistically leaving off the explicit `[T]` which works so long as `T` is unambiguous (as it normally is). Makes me question whether `.viewAs` is the right method name because the following also works:

```scala
  val a, b, c, d = IO(Input(UInt(8.W)))
  val sel = IO(Input(Bool()))
  val y, z = IO(Output(UInt(8.W)))
  (y, z).viewAs := Mux(sel, (a, b). viewAs, (c, d). viewAs)
}
```

mwachs5:
> This is actually the case that made me add the implicit conversion in the first place. 

I agree I really like the implicit conversion in the Mux case

I think you should include your Seq/Vec example from below it is also nice and compelling. 
```
// This would work
Seq(a, b, c, d).asData := out
// This would just connect to an intermediate Wire and thus be dropped
VecInit(a, b, c, d) := out // bad! don't do this!
```

jackkoenig:
Alright, after further thought, I'm going to move all of the `DataProduct` and `DataView` implementations into the companion objects of `DataProduct` and `DataView` so that they are available in the implicit scope (ie. without a specific import). The implicit conversions will remain in package `chisel3.experimental.conversions` (requiring the specific import) so that we can evaluate how well they work. My intuition is that they are a good thing (for these particular cases, definitely not for all cases), but we should see how it works in practice.

src/main/scala/chisel3/experimental/conversions/package.scala
```diff
@@ -0,0 +1,26 @@
...
+  * @note As this leans heavily on the experimental [[DataView]] feature, these APIs are experimental and subject to change
+  */
+package object conversions {
+
+  /** Implicit conversion between `Seq` and `Vec` */
+  implicit def seq2Vec[A : DataProduct, B <: Data](xs: Seq[A])(implicit dv: DataView[A, B]): Vec[B] =
+    xs.viewAs[Vec[B]]
+
+  /** Implicit conversion between `(A, B)` and `HWTuple2[A, B]` */
+  implicit def tuple2hwtuple[T1 : DataProduct, T2 : DataProduct, V1 <: Data, V2 <: Data](
```
mwachs5, line 19: why is the above `seq2Vec` and this one is not `tuple2HwTuple`

jackkoenig: I'll make them consistent

sequencer:
I checkout and tried this feature, I really like this conversation! I think implicitly conversion will be more friendly to most of newbies to Chisel.
Maybe a documentation(just copy&paste from PR description) for end user will be much better.

> One of my biggest user issues with the implicit conversions is its hard to understand why my code doesn't work when your code does. If there is a function it's easy to figure out in IDE how to include the function, but if an implicit conversion is missing, it's hard to discover.

By default, I forcing users from our lab using IntelliJ IDEA, and think they are supported by Jetbrains by default(I think it really can teach them Scala, haha).
Without implicit conversion, users must understand why they should `asData`, which is a tough job for me to explain the magic of dataview APIs.
So as a trade-off, I think a implicit conversion should be documented and optional import by user(I only need to ask them: just import this for better user experience).

carlosedp:
I'm all for the implicit conversions from a user perspective.
Having the casts all over will make things more complex where one will ask things like "should `.asData` go here... or there" making debug more confusing.

Other than that, I like the implementation making very straightforward to assign tuples/hwtuples too.

jackkoenig:
Okay I think this is ready. I've made a few minor changes but I think there's wide agreement on the approach here.
`DataProducts` and `DataViews` for `Seq` and `Tuple2`-`Tuple10` are now defined in the companion objects for `DataProduct` and `DataView`. This means that users who import `chisel3.experimental.dataview._` will have access to `.viewAs` for `Seqs` and `Tuples` with no additional import. The implicit conversions are in a new package, `chisel3.experiemental.conversions` such that if a user imports from that package, they will get the nice syntax for tuples and the like.

Interestingly, while the implicit conversions seem to work fantastically for Tuples, they seem to work less well for Seq, as evidenced by this test:
```scala
   val a, b, c, d = IO(Input(UInt(8.W)))
   val sel = IO(Input(Bool()))
   val y, z = IO(Output(UInt(8.W)))
   // Unclear why the implicit conversion does not work in this case for Seq
   // That being said, it's easy enough to cast via `.viewAs` with or without
   Seq(y, z) := Mux(sel, Seq(a, b).viewAs, Seq(c, d).viewAs[Vec[UInt]])
 }
```
I do not know why this is, but I wrote this test to capture something interesting. The implicit conversion **does** work for the LHS, ie. `:=` is only working here because of the implicit conversion, yet for some reason it does not work for the arguments to `Mux`. I then show that one can use `.viewAs` without explicit type parameters if there's only 1 `DataView` in scope (which is the normal case). You can of course, do the normally recommended thing and give type parameters.

I'm wondering if we should add `.asData` that is basically an alias for `.viewAs` but would return self if called on a `Data` and otherwise would do the same thing as `.viewAs`. Any thoughts? I do see that as a separate PR.

mwachs5 (approved):

carlosedp:
@jackkoenig it would be great to grab those nice examples and put them into the cookbook

jackkoenig:
> @jackkoenig it would be great to grab those nice examples and put them into the cookbook

Agreed, will do in follow on PR.

PR #2279, opened 2021-12-04
[docs] Recipe for a single wire for a bundle in a module interface

oharboe:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -25,6 +25,8 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
 * [How do I do subword assignment (assign to some bits in a UInt)?](#how-do-i-do-subword-assignment-assign-to-some-bits-in-a-uint)
 * [How do I create an optional I/O?](#how-do-i-create-an-optional-io)
 * [How do I minimize the number of bits used in an output vector](#how-do-i-minimize-the-number-of-bits-used-in-an-output-vector)
+* [How do I create a single input/output wire for a Bundle?]
+(#how-do-i-create-a-bundle-from-a-uint)
 * Predictable Naming
   * [How do I get Chisel to name signals properly in blocks like when/withClockAndReset?](#how-do-i-get-chisel-to-name-signals-properly-in-blocks-like-whenwithclockandreset)
   * [How do I get Chisel to name the results of vector reads properly?](#how-do-i-get-chisel-to-name-the-results-of-vector-reads-properly)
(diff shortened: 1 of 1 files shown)


docs/src/cookbooks/cookbook.md
```diff
@@ -404,6 +406,39 @@ class ModuleWithOptionalIO(flag: Boolean) extends Module {
 }
 ```
 
+### How do I create a single input/output wire for a bundle?
+
+If you have a bundle with many elements in a module interface, the interface can become
+quite verbose. To simplify the module interface, the entire bundle
+can be represented with a single wire.
+
```
jackkoenig, line 414: This recipe is very similar to "How do I create a UInt from an instance of a Bundle?" and "How do I create a Bundle from a UInt?", but I totally agree that having the right framing of recipes is important for discoverability. 

What do you think about phrasing it as something like "How do I pack an Aggregate into a single input/output port?" and then pointing it to "How do I create a UInt from an instance of a Bundle?"? We could also tweak the example up there to provide a port as well (instead of just a wire).

docs/src/cookbooks/cookbook.md
```diff
@@ -404,6 +406,39 @@ class ModuleWithOptionalIO(flag: Boolean) extends Module {
...
+Consider:
+
+```scala mdoc:silent:reset
+import chisel3._
+
+class DooHickeyBundle extends Bundle {
```
mwachs5, line 420: this is a bit nitpicky, but `DooHickey` is like `Foo` which we use most places, is there a reason to not use that here? I think `DooHickey` might be confusing for people who aren't native (american?) english speakers, while Foo is practically a term of art at this point...

docs/src/cookbooks/cookbook.md
```diff
@@ -404,6 +406,39 @@ class ModuleWithOptionalIO(flag: Boolean) extends Module {
...
+  // Lots of individual wires
+  val a = Vec(5, Bool())
+}
+
+class DooHickey() extends Module {
+  // many wires
+  val b = IO(Input(new DooHickeyBundle))
+  // single module interface wire wire
```
mwachs5, line 428:
```suggestion
  // one wide wire
```

docs/src/cookbooks/cookbook.md
oharboe: "Aggregate" is not used elsewhere in the cookbook, but rather "Bundle" is.

Tweaked exampel as suggested.

oharboe:
@mwachs5 @jackkoenig Anything else?

docs/src/cookbooks/cookbook.md
```diff
@@ -62,26 +64,35 @@ class Foo extends RawModule {
 
 Use the [`asTypeOf`](https://www.chisel-lang.org/api/latest/chisel3/UInt.html#asTypeOf[T%3C:chisel3.Data](that:T):T) method to reinterpret the [`UInt`](https://www.chisel-lang.org/api/latest/chisel3/UInt.html) as the type of the [`Bundle`](https://www.chisel-lang.org/api/latest/chisel3/Bundle.html).
 
+If you have a bundle with many elements in a module interface,
```
mwachs5, line 67: I am not sure what you are trying to do in this PR. You replaced an example that went with the text above, `"Use the .asTyepeOf"` with a different example that is showing your different point. Why did you delete the example that was there before?

mwachs5 (changes requested):
You are deleting an example that went with the text before and interjecting a different example / use case. Shouldn't this go under a different heading? I am also not sure why this is explicitly "* [How do I create a single input/output wire for a Bundle?]" and that is not even the title of the section that you are linking to.

I suggest you leave the existing examples alone and add a new top-level "recipe" heading to show whatever it is you are trying to show.

PR #2292, opened 2021-12-08
reduceTree() now operates on Seq

oharboe:
preserves input/output information of the type being reduced.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -327,11 +327,11 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
   def do_reduceTree(redOp: (T, T) => T, layerOp: (T) => T = (x: T) => x)
                    (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions) : T = {
     require(!isEmpty, "Cannot apply reduction on a vec of size 0")
-    var curLayer = this
+    var curLayer : Seq[T] = this
     while (curLayer.length > 1) {
-      curLayer = VecInit(curLayer.grouped(2).map( x =>
+      curLayer = curLayer.grouped(2).map( x =>
         if (x.length == 1) layerOp(x(0)) else redOp(x(0), x(1))
-      ).toSeq)
+      ).toSeq
     }
     curLayer(0)
   }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -327,11 +327,11 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int)
   def do_reduceTree(redOp: (T, T) => T, layerOp: (T) => T = (x: T) => x)
                    (implicit sourceInfo: SourceInfo, compileOptions: CompileOptions) : T = {
     require(!isEmpty, "Cannot apply reduction on a vec of size 0")
-    var curLayer = this
+    var curLayer:Seq[T] = this
```
jackkoenig, line 330:
```suggestion
    var curLayer: Seq[T] = this
```
Nit

src/test/scala/chiselTests/Aggregate.scala
```diff
@@ -0,0 +1,47 @@
...
+  stop()
+}
+
+
+
+class AggregateSpec extends ChiselPropSpec {
+  property("Aggregate functions should preserve input/output type") {
+      assertTesterPasses { new AggregateTester() }
+  }
+}
```
jackkoenig, line 47: Since `.reduceTree` is a method on `Vec` (not on `Aggregate`), could you move this test to `Vec.scala` in `VecSpec`? Also it would be helpful to give more precise names to the classes, eg.

`ReduceTreeModule` instead of `Foo` and `ReduceTreeTester` instead of `AggregateTester`?

jackkoenig (commented):
Generally looks great and will reduce the size of the hardware AST as well which is nice. Just a couple of comments.

src/test/scala/chiselTests/Aggregate.scala
oharboe: I'll

jackkoenig (approved):
LGTM! Thanks!

schoeberl:
Thank's for this fix. I hope it will get into 3.5, so I can include my tree arbitration example in the Chisel book.

BTW, what is the meaning of layerOp. I see it can be used when one node is promoted to the next layer without the normal reduction op. Is there any use case for this?

oharboe:
I would say that the layerOp is an example of the use of functional
programming that is starting to be a little bit too application specific.
You can articulate the same using Scala and functional programming but have
more fine grained control over how you insert pipelining in e.g. a tree
reduction.  I dont know how common its use might be in the wild...

ons. 22. des. 2021, 13:21 skrev Martin Schoeberl ***@***.***>:

PR #2297, opened 2021-12-09
ShiftRegister() and ShiftRegisters() now has a flatten option

oharboe:
flattening is default on, when off a module is used to implement
the shift registers.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/Reg.scala
+++ b/src/main/scala/chisel3/util/Reg.scala
@@ -42,7 +42,7 @@ object ShiftRegister
     * val regDelayTwo = ShiftRegister(nextVal, 2, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).lastOption.getOrElse(in)
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B, flatten: Boolean = true): T = ShiftRegisters(in, n, en, flatten).lastOption.getOrElse(in)
 
   /** Returns the n-cycle delayed version of the input signal with reset initialization.
     *
@@ -55,7 +55,7 @@ object ShiftRegister
     * val regDelayTwoReset = ShiftRegister(nextVal, 2, 0.U, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool): T = ShiftRegisters(in, n, resetData, en).lastOption.getOrElse(in)
+  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool, flatten: Boolean): T = ShiftRegisters(in, n, resetData, en, flatten).lastOption.getOrElse(in)
 }
 
 
@@ -68,8 +68,20 @@ object ShiftRegisters
     * @param en enable the shift
     *
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): Seq[T] =
-    Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B, flatten: Boolean = true): Seq[T] = {
+    if (flatten) {
+      Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+    } else  {
+      class ShiftRegisters extends Module {
+        val in = IO(Input(UInt()))
+        val out = IO(Output(Vec(n, UInt())))
+        out := ShiftRegisters(in.asUInt, n, en, true)
+      }
+      val shift = Module(new ShiftRegisters)
+      shift.in := in
+      shift.out.map(_.asTypeOf(in))
+    }
+  }
 
   /** Returns delayed input signal registers with reset initialization from 1 to n.
     *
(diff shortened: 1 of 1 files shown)


sequencer:
I don't object adding `Module` to `ShiftRegister(s)`, this is good for post synthesis flow, since PnR may need take additional care to this.
I personally think there are two method to do so:
1. adding a module and provide a `inline` option, but default to be `false` to avoid affecting other users. add `firrtl.passes.InlineAnnotation` to the correspond instance if true.
2. rather than adding a new `Module`, you can also use `chisel3.util.experimental.group` to gather register into a Module.

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -79,6 +91,20 @@ object ShiftRegisters
...
     */
-  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool): Seq[T] =
-    Seq.iterate(in, n + 1)(util.RegEnable(_, resetData, en)).drop(1)
+  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool, flatten: Boolean): Seq[T] = {
+    if (flatten) {
+      Seq.iterate(in, n + 1)(util.RegEnable(_, resetData, en)).drop(1)
+    } else  {
+      class ShiftRegistersX extends Module {
+        val resetData = IO(Input(Bool()))
+        val in = IO(Input(UInt()))
```
mwachs5, line 100: This should not be an IO

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -68,8 +68,20 @@ object ShiftRegisters
     * @param en enable the shift
     *
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): Seq[T] =
-    Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B, flatten: Boolean = true): Seq[T] = {
+    if (flatten) {
+      Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+    } else  {
+      class ShiftRegisters extends Module {
```
mwachs5, line 75: Can you call this ShiftRegister or ShiftReg? The "s" is implied

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -42,7 +42,7 @@ object ShiftRegister
     * val regDelayTwo = ShiftRegister(nextVal, 2, ena)
     * }}}
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): T = ShiftRegisters(in, n, en).lastOption.getOrElse(in)
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B, flatten: Boolean = true): T = ShiftRegisters(in, n, en, flatten).lastOption.getOrElse(in)
```
mwachs5, line 45: I think "inline" is a better name for this parameter, it's more industry standard

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -79,6 +91,20 @@ object ShiftRegisters
...
     *
     */
-  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool): Seq[T] =
-    Seq.iterate(in, n + 1)(util.RegEnable(_, resetData, en)).drop(1)
+  def apply[T <: Data](in: T, n: Int, resetData: T, en: Bool, flatten: Boolean): Seq[T] = {
+    if (flatten) {
+      Seq.iterate(in, n + 1)(util.RegEnable(_, resetData, en)).drop(1)
+    } else  {
+      class ShiftRegistersX extends Module {
+        val resetData = IO(Input(Bool()))
```
mwachs5, line 99: Doesn't this need en as an input too?

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -68,8 +68,20 @@ object ShiftRegisters
...
     *
     */
-  def apply[T <: Data](in: T, n: Int, en: Bool = true.B): Seq[T] =
-    Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+  def apply[T <: Data](in: T, n: Int, en: Bool = true.B, flatten: Boolean = true): Seq[T] = {
+    if (flatten) {
+      Seq.iterate(in, n + 1)(util.RegEnable(_, en)).drop(1)
+    } else  {
+      class ShiftRegisters extends Module {
+        val in = IO(Input(UInt()))
```
mwachs5, line 76: Doesn't this need en as an input too?

mwachs5 (changes requested):
You will need to write tests

mwachs5:
I also like the @sequencer suggestion of the inlineAnnotation, but we should look at the difference in the output from the current solution to one using that to see if the changes are acceptable

oharboe:
> You will need to write tests

Yep. This is just for discussion for now though.

I want to know how to handle indexing of vectors, where I think it would be similarly helpful to control flattening.

oharboe:
This needs a more cross-cutting solution. In the spirit of this pull request, there should be a way to control flattening of  Vec() index access. Vec() access for something like Vec(16, UInt(64.W)) generates a giant fog of logic and should be representable as a module.

PR #2304, opened 2021-12-15
Improve exception message for aliased bundle fields

chick:
Improve exception message for aliased bundle fields
- Shows groups of field names that share a common id (i.e. aliased)
- Show, as much as possible, them in the order that fields appear in bundle
- Updated BundleSpec's relevant tests

### Contributor Checklist

#### Type of Improvement

Improves error message

#### API Impact

None

None

#### Desired Merge Strategy

Squash

#### Release Notes

The exception thrown when a bundle contains aliased field will now show the field names of the
aliased fields.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -33,7 +33,7 @@ sealed abstract class Aggregate extends Data {
         case b: Bundle =>
           // show groups of names of fields with duplicate id's
           // The sorts make the displayed order of fields deterministic and matching the order of occurrence in the Bundle.
-          // it's a bit convoluted but happens rarely and makes error easier to understand for user
+          // It's a bit convoluted but happens rarely and makes the error message easier to understand
           val dupNames = duplicates.toSeq.sortBy(_._id).map { duplicate =>
             b.elements
               .collect { case x if x._2._id == duplicate._id => x }
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -42,7 +42,7 @@ sealed abstract class Aggregate extends Data {
               .mkString("(", ",", ")")
           }.mkString(",")
           throw new AliasedAggregateFieldException(
-            s"Bundle ${b.getClass} contains aliased fields named ${dupNames}"
+            s"${b.className} contains aliased fields named ${dupNames}"
           )
         case _ =>
           throw new AliasedAggregateFieldException(
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -30,7 +30,7 @@ sealed abstract class Aggregate extends Data {
     val duplicates = getElements.groupBy(identity).collect { case (x, elts) if elts.size > 1 => x }
     if (!duplicates.isEmpty) {
       this match {
-        case b: Bundle =>
+        case b: Record =>
           // show groups of names of fields with duplicate id's
           // The sorts make the displayed order of fields deterministic and matching the order of occurrence in the Bundle.
           // It's a bit convoluted but happens rarely and makes the error message easier to understand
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -29,7 +29,26 @@ sealed abstract class Aggregate extends Data {
...
+              .toSeq.sortBy(_._2._id)
+              .map(_._1).reverse
+              .mkString("(", ",", ")")
+          }.mkString(",")
+          throw new AliasedAggregateFieldException(
+            s"Bundle $this contains aliased fields named ${dupNames}"
+          )
+        case _ =>
+          throw new AliasedAggregateFieldException(
+            s"Aggregate $this contains aliased fields $duplicates ${duplicates.mkString(",")}"
```
mwachs5, line 49: can we show hitting this in your test case below? I think the existing case only hits the above check?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -29,7 +29,26 @@ sealed abstract class Aggregate extends Data {
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
     val duplicates = getElements.groupBy(identity).collect { case (x, elts) if elts.size > 1 => x }
     if (!duplicates.isEmpty) {
-      throw new AliasedAggregateFieldException(s"Aggregate $this contains aliased fields $duplicates")
+      this match {
+        case b: Bundle =>
+          // show groups of names of fields with duplicate id's
+          // sorts are to make order of fields shown, deterministic and matching order of occurrence in bundle
```
mwachs5, line 35:
```suggestion
          // The sorts make the displayed order of fields deterministic and matching order of occurrence in the Bundle
```

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -29,7 +29,26 @@ sealed abstract class Aggregate extends Data {
...
+          // it's a bit convoluted but happens rarely and makes error easier to understand for user
```
mwachs5, line 36:
```suggestion
          // it's a bit convoluted but happens rarely and makes the error message easier to understand
```

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: You could hit the bottom with a `Record`, but I think it would be better for the logic for `Record` to be the same as for `Bundle` up above. You can use value `className` in the String error to give more precise `Record` vs. concrete `Bundle` subclass in the error message.

If you fold `Record` handling into the above, this should be unreachable but still good to leave with a decent error message.

chick: I changed code to check for Record which will include Bundle.
I think the `case _ =>` should be considered future proofing against changes that inadvertently allow aliased fields or new types that might have them

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -29,7 +29,26 @@ sealed abstract class Aggregate extends Data {
...
+          // It's a bit convoluted but happens rarely and makes the error message easier to understand
+          val dupNames = duplicates.toSeq.sortBy(_._id).map { duplicate =>
+            b.elements
+              .collect { case x if x._2._id == duplicate._id => x }
+              .toSeq.sortBy(_._2._id)
+              .map(_._1).reverse
+              .mkString("(", ",", ")")
+          }.mkString(",")
+          throw new AliasedAggregateFieldException(
+            s"Bundle ${b.getClass} contains aliased fields named ${dupNames}"
```
jackkoenig, line 45:
```suggestion
            s"${b.className} contains aliased fields named ${dupNames}"
```
Not `Bundle` because it could be a `Record` and `.getClass` isn't quite right, we use `.className` (defined in this file) for other String printing.

jackkoenig (approved):
LGTM! I actually used this last night to debug issues in DSPTools!

PR #2306, opened 2021-12-16
Chisel plugin bundle elements handler

chick:
### Contributor Checklist

#### Type of Improvement

Performance improvement
Code refactoring

#### API Impact

Could affect unusual Bundle constructions.
Some error messages have been changed

This should not impact code generation

#### Release Notes

This eliminates the use of reflection in Aggregate.scala when building the `elements` method of a Bundle.
This should result in improvement in elaboration times for large designs.

- [ ] Did you review?


core/src/main/scala/chisel3/BundleComparator.scala
```diff
@@ -0,0 +1,39 @@
+package chisel3
+
+/** This comparator compares the new and old elements accessor of a `Bbundle``'s fields
```
mwachs5, line 3:
```suggestion
/** This comparator compares the new and old elements accessor of a [[Bundle]]'s fields
```

src/main/scala/chisel3/util/HasDemoBundle.scala
```diff
@@ -0,0 +1,426 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+//package chisel3
+//
+//import scala.language.reflectiveCalls
+//
+//import chisel3.experimental.{ChiselEnum, FixedPoint}
+//import chisel3.stage.ChiselStage
+//import chisel3.util.Decoupled
+//
```
jackkoenig, line 10: I'm assuming you didn't mean to include this file?

chick: Sheepishly did mean to leave it in there for the moment, I guess I could tuck it somewhere else, but in case there are problems I uncomment specific blocks and am able to edit compile debug much quicker. Happy to remove it though

jackkoenig: We can keep it for now but delete before merging, no worries

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1088,12 +1088,50 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+      throw ExpectedChiselTypeException(s"Bundle: $this contains hardware fields: " + hardwareFields.mkString(","))
+    }
+    VectorMap(
+      _elementsImpl
+        .toSeq
+        .flatMap {
+          case (name, data: Data) =>
+              Some(name -> data)
+          case (name, data: Some[Data]) =>
+              Some(name -> data.get)
```
jackkoenig, line 1116:
```suggestion
          case (name, opt: Some(data: Data)) =>
              Some(name -> data)
```
Due to type erasure, you can't match on the type parameter of the `Some`, but you can unbox it and just look inside.

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    object Enum0 extends ChiselEnum {
+      val s0, s1, s2 = Value
+    }
+
+    ChiselStage.emitFirrtl(new Module1)
+  }
+}
+
```
jackkoenig, line 476: The breadth of tests here is good but it would be nice to have them a little cleaned up, eg. getting rid of commented out code and giving things qualitative names instead of stuff like `DebugProblem13Module`.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -8,10 +8,17 @@ import nsc.plugins.{Plugin, PluginComponent}
...
+  var buildElementsAccessor: Boolean = true,
+  var pluginDebugBundlePattern: String = "" // For example try .*Bpip.*"
+) {
   def useBundlePluginOpt = "useBundlePlugin"
   def useBundlePluginFullOpt = s"-P:${ChiselPlugin.name}:$useBundlePluginOpt"
-
+  def buildElementAccessorOpt = "buildElementAccessor"
+  def buildElementAccessorFullOpt = s"-P:chiselplugin:$buildElementAccessorOpt"
+  def pluginDebugBundlePatternOpt = "enablePluginDebug"
+  def pluginDebugBundlePatternFullOpt = s"-P:chiselplugin:$pluginDebugBundlePatternOpt:"
```
jackkoenig, line 21: Can you add some comments here to describe how to use the debugpattern stuff? Is that something we want merged or is it just for debugging during development?

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -63,62 +97,177 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
+            allParentFields ++ currentFields
+          }
+
+          val elementArgs = getAllBundleFields(bundle.symbol)
+
+          val elementsImplSym =
+            bundle.symbol.newMethod(TermName("_elementsImpl"), bundle.symbol.pos.focus, Flag.OVERRIDE | Flag.PROTECTED)
+          // Handwritten cloneTypes don't have the Method flag set, unclear if it matters
+          elementsImplSym.resetFlag(Flags.METHOD)
+          // Need to set the type to chisel3.Bundle for the override to work
```
jackkoenig, line 250: I believe these comments are wrong

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -63,62 +97,177 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
-
-        // Create the symbol for the method and have it be associated with the Bundle class
-        val cloneTypeSym =  bundle.symbol.newMethod(TermName("_cloneTypeImpl"), bundle.symbol.pos.focus, Flag.OVERRIDE | Flag.PROTECTED)
-        // Handwritten cloneTypes don't have the Method flag set, unclear if it matters
-        cloneTypeSym.resetFlag(Flags.METHOD)
-        // Need to set the type to chisel3.Bundle for the override to work
-        cloneTypeSym.setInfo(NullaryMethodType(bundleTpe))
-
-        val cloneTypeImpl = localTyper.typed(DefDef(cloneTypeSym, neww))
+        var additionalMethods: Seq[Tree] = Seq()
```
jackkoenig, line 132: Can we do this immutably instead of with a `var`?

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -63,62 +97,177 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
+        }
 
         // ==================== Generate _usingPlugin ====================
         // Unclear why quasiquotes work here but didn't for cloneTypeSym, maybe they could.
         val usingPlugin = localTyper.typed(q"override protected def _usingPlugin: Boolean = true")
+        additionalMethods ++= Seq(usingPlugin)
 
         val withMethods = deriveClassDef(bundle) { t =>
-          deriveTemplate(t)(_ :+ cloneTypeImpl :+ usingPlugin)
+          deriveTemplate(t)(_ ++ additionalMethods)
```
jackkoenig, line 270: Note that instead of `var additionalMethods`, if you make `elementsImpl` be of type `Option[Tree]`, this line can just be:
```scala
deriveTemplate(t)(_ :+ cloneTypeImpl :+ usingPlugin ++ elementsImpl)
```

jackkoenig (changes requested):

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,80 @@
+# Notes on the Compiler Plug-in
+
+The Chisel plugin provides some features that are too difficult, or not possbile, 
+to implement through regular code. These features
```
mwachs5, line 4: Something is missing in this sentence

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Compiler plugin Features
+- Records that this plugin ran on a bundle by adding a method `_usingPlugin: Boolean = true`
+- Automatically generating the `cloneType` methods of Bundles
+- Generating a function that returns list of the hardware fields in a bundle
```
mwachs5, line 9:
```suggestion
- Records that this plugin ran on a bundle by adding a method `_usingPlugin: Boolean = true`
- Automatically generates the `cloneType` methods of Bundles
- Generates a function that returns a list of the hardware fields in a bundle
```

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,80 @@
+# Notes on the Compiler Plug-in
```
mwachs5, line 1: you need to do an editing pass on this file actually looking at it in a markdown viewer, many issues in `.md` format

core/src/main/scala/chisel3/Aggregate.scala
chick: Will be in next commit

src/main/scala/chisel3/util/HasDemoBundle.scala
```diff
@@ -0,0 +1,523 @@
...
+//  object Enum0 extends ChiselEnum {
+//    val s0, s1, s2 = Value
+//  }
+//
+//
+//
+//  def main(args: Array[String]): Unit = {
+//    ChiselStage.emitFirrtl(new Module1)
+//  }
+//}
```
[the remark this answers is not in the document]
chick: Next commit will review

src/test/scala/chiselTests/BundleSpec.scala
```diff
@@ -70,24 +74,24 @@ class BundleSpec extends ChiselFlatSpec with BundleSpecUtils with Utils {
...
+//    (the[ChiselException] thrownBy extractCause[ChiselException] {
+//      ChiselStage.elaborate {
+//        new Module {
+//          val io = IO(new Bundle {
+//            val b = new BadSeqBundle
+//          })
+//        }
+//      }
+//    }).getMessage should include("Public Seq members cannot be used to define Bundle elements")
+//  }
```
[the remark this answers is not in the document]
chick: Working on this, I've got a problem here, passing it on to the Aggregate code

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
chick: Done

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
chick: Yes, this is will be cleaner
Done

src/main/scala/chisel3/util/HasDemoBundle.scala
chick: Will Delete in next commit

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
chick: These comments were part of the code previous to this PR.
I just moved them. What do you suggest here?

docs/src/appendix/chisel-plugin.md
chick: This change will be in next commit

src/test/scala/chiselTests/BundleSpec.scala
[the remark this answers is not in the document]
chick: Ok, I think I have it fixed, passes this test now

src/test/scala/chiselTests/BundleElementsSpec.scala
chick: Did major cleanup here, mostly just naming things based on the test involved

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
chick: I removed this from here. Added a docs/src/appendix/chisel-plugin.md to describe debug techniques

docs/src/appendix/chisel-plugin.md
chick: I made this a bit cleaner, it's not too thorough but has some better debugging hints. Used IntelliJ's md rendering.
I hope that is sufficient.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -8,10 +8,15 @@ import nsc.plugins.{Plugin, PluginComponent}
...
+private[plugin] case class ChiselPluginArguments(
+  val skipFiles:                mutable.HashSet[String] = mutable.HashSet.empty,
+  var buildElementsAccessor:    Boolean = true,
+  var pluginDebugBundlePattern: String = "" // For example try .*Bpip.*"
+) {
   def useBundlePluginOpt = "useBundlePlugin"
   def useBundlePluginFullOpt = s"-P:${ChiselPlugin.name}:$useBundlePluginOpt"
-
+  def buildElementAccessorOpt = "buildElementAccessor"
+  def buildElementAccessorFullOpt = s"-P:chiselplugin:$buildElementAccessorOpt"
```
jackkoenig, line 19:
```suggestion
  def genBundleElementsOpt = "genBundleElements"
  def genBundleElementsOpt = s"-P:chiselplugin:$genBundleElementsOpt"
```
I would reframe this more toward the user API than the internals (what does "Accessor" mean to the user?) I'm basing this off the 3.4.3 option `-P:chiselplugin:useBundlePlugin`, but am open to other ideas.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -8,10 +8,15 @@ import nsc.plugins.{Plugin, PluginComponent}
 import scala.reflect.internal.util.NoPosition
 import scala.collection.mutable
 
-private[plugin] case class ChiselPluginArguments(val skipFiles: mutable.HashSet[String] = mutable.HashSet.empty) {
+private[plugin] case class ChiselPluginArguments(
+  val skipFiles:                mutable.HashSet[String] = mutable.HashSet.empty,
+  var buildElementsAccessor:    Boolean = true,
```
jackkoenig, line 13:
```suggestion
  var buildElementsAccessor:    Boolean = false,
```
If the option is to enable it, then it needs to be false by default.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -8,10 +8,15 @@ import nsc.plugins.{Plugin, PluginComponent}
...
+  var pluginDebugBundlePattern: String = "" // For example try .*Bpip.*"
```
jackkoenig, line 14: There doesn't seem to be a way to set this via a scalacOption

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -0,0 +1,460 @@
...
+object ContainsImplicitTestingClass {
+  implicit class BooleanToAugmentedBoolean(private val x: Boolean) extends AnyVal {
+    def toInt: Int = if (x) 1 else 0
+
+    // this one's snagged from scalaz
+    def option[T](z: => T): Option[T] = if (x) Some(z) else None
+  }
+}
```
jackkoenig, line 460:
```suggestion
```

These are tests and adding this syntactic sugar from rocket-chip is kind of weird.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1243,6 +1289,40 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
...
+                  case _ => // don't care about non-Data Seq
+                }
+              case _ => // not a Seq
+            }
+          }
+      }
+    }
+    VectorMap(nameMap.toSeq.sortWith { case ((an, a), (bn, b)) => (a._id > b._id) || ((a eq b) && (an > bn)) }: _*)
+  }
+
```
jackkoenig, line 1325: This was useful for development but this needs to be private if we keep it.

core/src/main/scala/chisel3/BundleComparator.scala
```diff
@@ -0,0 +1,39 @@
...
+        if (showComparison) {
+          println(f"$color$a%30s (${b._id}%06x) $c%30s (${d._id}%06x) ${Console.RESET}")
+        }
+    }
+    if (showComparison) {
+      println("=" * header.length)
+    }
+    discrepancyFound
+  }
+}
```
jackkoenig, line 39: This definitely needs to not be part of the public API, if it's absolutely needed it should be in package `chisel3.internal`. That being said, I don't think this is necessary to keep. If you are that worried about needing it for debugging future issues, it would be better if we can move this and the implementation of the old elements into the test project ([TestUtils](https://github.com/chipsalliance/chisel3/blob/master/src/test/scala/chisel3/testers/TestUtils.scala) is in package chisel3 so can access package private stuff like `getPublicFields`).

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Generating `cloneType` method
+After a period of testing this is now on by default. This flag was used to control that prior to full adoption.
+```
+-P:chiselplugin:useBundlePlugin
+```
+
+### Generating the `_elementsImpl` method
+For generating the elementsImpl method, the default is off.
+Turn it on by using the javac flag
```
jackkoenig, line 19:
```suggestion
Turn it on by using the scalac flag (scalacOptions in SBT)
```

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
jackkoenig: It looks more like you copy-pasted them because that comment is correct for `_cloneTypeImpl` but not correct for `_elementsImpl`, the return type is not `chisel3.Bundle`, it's `SeqMap[String, Any]` (see next line).

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
+# Notes on the Compiler Plug-in
+
+The Chisel plugin provides some operations that are too difficult, or not possbile, 
+to implement through regular Scala code.
+
+## Compiler plugin operations
+1. Automatically generates the `cloneType` methods of Bundles
+2. Records that this plugin ran on a bundle by adding a method `_usingPlugin: Boolean = true`
+3. Generates a function that returns list of the hardware fields in a bundle `_elementsImpl`
+
```
jackkoenig, line 10: I think this is useful information, but it feels like a weird mix of "useful to users" and "useful only to contributors". I think it would be better to either split this into 2 files, or at least call out with a `##` level title ("Notes for Contributors") or something.

mwachs5: Agreed -- for example should this appear on the website?

jackkoenig: I'd generally say the contributor only stuff should not live on the website, but I'm not to fussed either way, it just needs to be clear what is what.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
chick: Works for me, will do this

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
chick: I'm removing this, my docs say to just create this in the code during a debugging session

core/src/main/scala/chisel3/BundleComparator.scala
```diff
@@ -0,0 +1,39 @@
+package chisel3
```
mwachs5, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0

package chisel3
```

docs/src/appendix/appendix.md
```diff
@@ -9,6 +9,7 @@ section: "chisel3"
 This section covers some less-common Chisel topics.
 
 * [Differences between Chisel3 and Chisel2](chisel3-vs-chisel2)
+* [Notes on Chisel Plugin](chisel-plugin)
```
mwachs5, line 12: this will cause this to show up on website and will need a corresponding chisel-lang website PR

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
...
+### Generating `cloneType` method
+After a period of testing this is now on by default. This flag was used to control that prior to full adoption.
```
mwachs5, line 12:
```suggestion
After a period of testing this is now on by default (as of Chisel Foo.Baz.Biddle). This flag was used to control that prior to full adoption.
```

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
...
+For generating the elementsImpl method, the default is off.
+Turn it on by using the javac flag
+```
+-P:chiselplugin:buildElementAccessor
+```
+
+#### Detecting Bundles with Seq[Data]
+Currently having this is a run time error
+Here is a block of code that could be added (with some refinement in 
+the detection mechanism)
```
mwachs5, line 27:
```suggestion

Trying to have a `val Seq[Data]` (as opposed to a `val Vec[Data]` in a `Bundle` is
a run time error.
Here is a block of code that could be added to the plugin to detect this case at compile time (with some refinement in 
the detection mechanism):
```

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+### Generating the `_elementsImpl` method
+For generating the elementsImpl method, the default is off.
+Turn it on by using the javac flag
+```
+-P:chiselplugin:buildElementAccessor
+```
+
+#### Detecting Bundles with Seq[Data]
```
mwachs5, line 24: I'm confused what this section is saying. does the compiler plugin detect this or not? What does "currently" mean?

docs/src/appendix/chisel-plugin.md
mwachs5, line 10: 4. Future work -- detecting Seq[Data] in Bundles? I'm confused about what that section is saying

docs/src/appendix/chisel-plugin.md
```diff
@@ -0,0 +1,76 @@
...
+  //      elide these fields, and will be caught in Aggregate.scala at runtime
+  if (member.isAccessor && typeIsSeqOfData(member.tpe) && !isIgnoreSeqInBundle(bundleSymbol)) {
+    global.reporter.error(
+      member.pos,
+      s"Bundle.field ${bundleSymbol.name}.${member.name} cannot be a Seq[Data]. " +
+        "Use Vec or MixedVec or mix in trait IgnoreSeqInBundle"
+    )
+  }
+```
+## Notes about working on the `_elementsImpl` generator for the plugin in `BundleComponent.scala`
```
mwachs5, line 40:
```suggestion
### Notes about working on the `_elementsImpl` generator for the plugin in `BundleComponent.scala`
```

docs/src/appendix/chisel-plugin.md
chick: will do

src/test/scala/chiselTests/BundleElementsSpec.scala
chick: I will clean this up and eliminate the implicit class

core/src/main/scala/chisel3/Aggregate.scala
chick: Making this private

core/src/main/scala/chisel3/BundleComparator.scala
chick: I've removed this

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
chick: Yeah, I was just leaving this until the end.

chick: done

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
chick: remove comments

build.sbt
```diff
@@ -198,6 +198,7 @@ lazy val chisel = (project in file(".")).
   settings(
     mimaPreviousArtifacts := Set(),
     libraryDependencies += defaultVersions("treadle") % "test",
+    scalacOptions += "-P:chiselplugin:genBundleElements",
```
jackkoenig, line 201: I think we cannot enable this for regular Chisel (but can in the tests) for the same reasons we couldn't enable autoclonetype2, it'll cause issues for any Bundle extending a Bundle in `chisel3.util` (see https://github.com/chipsalliance/chisel3/releases/tag/v3.4.3)

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -0,0 +1,457 @@
...
+    fieldNames.toSeq
+      .zipAll(bundle.elements.map(_._1).toList, "Missing Bundle Field", "Extra BundleField")
+      .forall {
+        case (fromBundle, fromList) =>
+          if (fromBundle != fromList) {
+            println(s"""Bundle: ${bundle}, field mismatch "$fromBundle" to "$fromList" """)
+          }
+          fromBundle == fromList
+      }
+  }
```
jackkoenig, line 456: I really would prefer if this matched just not names but also the actual Data themselves. I think a utility like the following would be slightly more verbose for each check but actually meaningfully checks that we aren't accidentally cloning the Data or something.
```scala
def checkBundle[T <: Bundle](bun: T)((checks: T => (String, Data))*): Unit = {
  bun.elements.toSeq should equal checks.map(bun).toSeq
}

checkBundle(new BpipOneField)(
  "bpipSuperTraitGood" -> _.bpipSuperTraitGood,
  "bpipTraitGood" -> _.bpipTraitGood,
  ...
)```

File: core/src/main/scala/chisel3/BundleComparator.scala (hunk 22 above)
Comment (chick) [reply core/src/main/scala/chisel3/BundleComparator.scala]: BundleComparator was deleted

File: docs/src/appendix/chisel-plugin.md (hunk 25 above)
Comment (chick) [reply docs/src/appendix/chisel-plugin.md:27]: will add this to next commit on this PR

File: docs/src/appendix/chisel-plugin.md (hunk 21 above)
Comment (chick) [reply docs/src/appendix/chisel-plugin.md:10]: This section was just enumerating the things the plugin currently does.
I've added a bit more verbiage to this effect.

File: docs/src/appendix/chisel-plugin.md (hunk 27 above)
Comment (chick) [reply docs/src/appendix/chisel-plugin.md:40]: done

File: src/test/scala/chiselTests/BundleElementsSpec.scala (hunk 29 above)
Comment (chick) [reply src/test/scala/chiselTests/BundleElementsSpec.scala]: done, I have pushed this commit into PR

File: docs/src/appendix/chisel-plugin.md (hunk 26 above)
Comment (chick) [reply docs/src/appendix/chisel-plugin.md:24]: I've cleaned up the language a bit here. Basically says this feature (3. on earlier list) is off by default and uses a scalac flag to turn it on.

File: docs/src/appendix/appendix.md (hunk 23 above)
Comment (chick) [reply docs/src/appendix/appendix.md:12]: Removed the link.
Moved the chisel3-plugin.md to a README.md in the plugin folder
Noted that the information there is for developers working on chisel3 internals

File: docs/src/appendix/chisel-plugin.md (hunk 24 above)
Comment (chick) [reply docs/src/appendix/chisel-plugin.md:12]: Referenced date and pr where current behavior was implemented

File: build.sbt (hunk 28 above)
Comment (chick) [reply build.sbt:201]: @jackkoenig can you point me to the place where this should go

Comment (jackkoenig) [reply build.sbt:201]: It just needs to be `Test / scalacOptions += "-P:chiselplugin:genBundleElements",`

File: src/test/scala/chiselTests/BundleElementsSpec.scala [hunk 30]
```diff
@@ -0,0 +1,515 @@
...
+        } else {
+          None
+        }
+    }
+    if (errorsStrings.nonEmpty) {
+      assert(false, s"Bundle: ${bun.getClass.getName}: " + errorsStrings.mkString("\n"))
+      false
+    } else {
+      true
+    }
```
jackkoenig, line 513:
```suggestion
    assert(errorsStrings.isEmpty, s"Bundle: ${bun.getClass.getName}: " + errorsStrings.mkString("\n"))
```
You're not using the return value so I don't see any point in having this return `Boolean` instead of just Unit

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -0,0 +1,515 @@
...
+      val io = IO(new Bundle {
+        val foo = Input(UInt(8.W))
+        val x = new Bundle {
+          val y = if (false) Some(Input(UInt(8.W))) else None
+        }
+      })
+      assertElementsMatchExpected(io)("x" -> _.x, "foo" -> _.foo)
+//TDDO: Figure out why the following line fails.
+//      assertElementsMatchExpected(io.x)("y" -> _.y.get)
+      assertElementsMatchExpected(io.x)()
```
jackkoenig, line 400: It fails because you've made `y` `None`, if you want it to pass with the get, you need to make the argument to the `if` be true.

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -0,0 +1,515 @@
...
+  out := DontCare
+
+  assertElementsMatchExpected(out)("superQux" -> _.superQux, "superFoo" -> _.superFoo)
+}
+
+/* Testing whether gen fields superFoo and superQux can be found when they are
+ * declared in a superclass
+ *
+ */
+class BundleElementsSpec extends AnyFreeSpec with Matchers {
```
jackkoenig, line 226: These tests look great, I would like to see 1 negative test just to sanity check that `assertElementsMatchExpected` is checking something. Could just be a test that is intentionally wrong and you just expect the assertion failure.

build.sbt
chick: Will do

src/test/scala/chiselTests/BundleElementsSpec.scala
chick: Adding a new block of tests 
"plugin tests should fail properly in the following cases"
that demonstrates the proper failure modes of checkAssertion

src/test/scala/chiselTests/BundleElementsSpec.scala
chick: Agreed, reworking as suggested

src/test/scala/chiselTests/BundleElementsSpec.scala
chick: Yep, stared at this too long. Removed comment

docs/src/appendix/chisel-plugin.md
chick: @jackkoenig @mwachs5 I agree this is weird. I moved this purposely into the README instead of placing it where it would be picked up by website. This should be a short term option that quickly (maybe by 3.5.2) becomes the default.
I tried to simplify and condense things a bit. Not sure how much more to do here

build.sc
```diff
@@ -58,7 +58,8 @@ trait CommonModule extends CrossSbtModule with PublishModule with ScalafmtModule
   override def scalacOptions = T {
     super.scalacOptions() ++ Agg(
       "-deprecation",
-      "-feature"
+      "-feature",
+      "-P:chiselplugin:genBundleElements"
```
jackkoenig, line 62: This also needs to only affect the tests, put it in `object test` below.

jackkoenig:
I have not done super robust benchmarking here, but for a pretty typical large SiFive design, this makes Chisel ~16% faster. Pretty good!

jackkoenig (approved):
LGTM! Excellent work @chick!

Issue #2144, opened 2021-10-04
Vec.reduceTree produces suboptimal reduction tree


nicmcd:
**Type of issue**: bug report

**Impact**: no functional change | increase efficiency in produced verilog

**What is the current behavior?**
Vec[Bool].reduceTree produces suboptimal reductions.
```
// chisel
class OrReduce(n: Int) extends Module {
  val io = IO(new Bundle {
    val in = Input(Vec(n, Bool()))
    val out = Output(Bool())
  })
  io.out := io.in.reduceTree(_ | _)
}

// generated verilog
module OrReduce(
  input   clock,
  input   reset,
  input   io_in_0,
  input   io_in_1,
  input   io_in_2,
  input   io_in_3,
  input   io_in_4,
  input   io_in_5,
  input   io_in_6,
  input   io_in_7,
  output  io_out
);
  assign io_out = io_in_0 | io_in_1 | (io_in_2 | io_in_3) | (io_in_4 | io_in_5 | (io_in_6 | io_in_7)); // @[OrReduce.scala 13:32]                                                                                  
endmodule
```
This OR reduction has 6 gates of delay.

**What is the expected behavior?**
The produced reduction should be as follows:
```
assign io_out = ((io_in_0 | io_in_1) | (io_in_2 | io_in_3)) | ((io_in_4 | io_in_5) | (io_in_6 | io_in_7));
```
This has only 3 gate delays.

**Please tell us about your environment:**
chisel = 3.4.3
scala = 2.12.13

**What is the use case for changing the behavior?**
Bitwise reductions are very common. If a function name implies the algorithm used for reduction (i.e., reduceTree) then the produced hardware should match the named algorithm.

sequencer:
I think synthesis tool should handle this directly, since the fanout selection is actually a physical sense synthesis job, chisel doesn’t aware of physical information, which make chisel impossible to know how to generate a better circuit with optimized fanout.

nicmcd:
I know that a synthesis tool *may* optimize the reduction tree, but if Chisel is going to rely on this then logically there should be no reduceTree() function at all since the reduce() will be equivalently reduced. So the question is... what is the purpose of the reduceTree() function?

seldridge:
Several things:

[...]

PR #2318, opened 2021-12-23
Generate a balanced tree with reduceTree

schoeberl:
### Contributor Checklist

#### Type of Improvement

- additional test

#### API Impact

non

non

#### Desired Merge Strategy

- [ ] Did you review?


schoeberl:
Shall I also move the other two ```reduceTree''' tests into this ```Spec'''?

Shall I include the fix for the unbalanced issue #2144 in this PR or in a different one?

schoeberl:
fixes  #2144

Ready for review, Maybe @seldridge. However, I am missing the right to request a reviewer.

sequencer:
I think there are two issues here:
1. if element of vector is not associative, this implementation will change the behavior.
2. implementation implemented a binary tree, which might be sub-optimized in ASIC designs.

schoeberl:
Hi @sequencer, thanks for the review. However, I do not really understand your two questions. (1) what do you mean of vector associative and (2) I think a binary tree is the best solution we can do. The original implementation did something along this but unbalanced.

sequencer:
Sorry for the delay response, my main concern here is:
1. the behavior of `reduceTree` is changed, that's my main concern, basically if the operation in `reduceTree` is not "associative": for example `+` operation to a IEEE754 floating data structure, this will break the legacy code.
If we need to make this change: we must have the deprecation flow.
2. For digital circuit designs: binary tree might not be the best solution, since high-fanout optimization can be leveraged to generate high density circuits, for example `OR4` use 10 transistors, while binary tree with 3*`OR2` use 18 transistors. Although this is the optimization what synthesis tools should take care of, I think arbitrary tree support to `reduceTree` might be a better solution.

So from my point of view, I think: behavior of the `reduceTree` should not be changed silently, since this might introduce bugs to users.
If we really want have a binary reduce tree support in `reduceTree`: I think we need to add another option to make this configureable, and use the current implementation by default to avoid subtle bug to users.

jackkoenig:
I wanted to visualize the change here so I sketched up some code to generate a dotfile while running the reduction:
```scala
import chisel3._

import java.io._

class TreeReduce extends Module {
  var i = 0L
  val map = new mutable.HashMap[Data, String]()
  val getName: Data => String = (d: Data) =>
    map.getOrElseUpdate(d, {
      val name = s"_T_$i"
      i += 1
      name
    })
  val in = IO(Input(Vec(5, UInt(8.W))))
  val out = IO(Output(UInt()))

  // Prepopulate names
  in.zipWithIndex.foreach { case (d, i) => map(d) = s"""\"in[$i]\"""" }

  val fw = new FileWriter(new File("output.dot"))
  fw.write("digraph g {\nnode [shape=plaintext];\n\n")

  out := in.reduceTree { (x, y) =>
    val res = x +% y
    fw.write(s"${getName(x)} -> ${getName(res)};\n")
    fw.write(s"${getName(y)} -> ${getName(res)};\n")
    res
  }
  fw.write("}\n")
  fw.close()
}
object Main extends App {
  println(getVerilogString(new TreeReduce))
}
```

### .reduce

reduce_5

### .reduceTree on master

reduceTree_5

### .reduceTree on this PR

**Note** that `in[0]` is no longer in the top-left, sorry about that. I didn't feel like figuring out how to make GraphViz do what I wanted here.

reduceTreeNew_5

schoeberl:
Thank you for this drawing, I should have done it ;-) Yes, this is correct. And in my opinion, the current version is unfair, as in[4] has only one hop, while the others have three hops). The PR version resolves any difference at the first reduce step towards a power of 2 number of elements. At a power of 2, it is a perfect tree then.

Can we at some point merge this?

jackkoenig:
Yeah I think we should merge this, I was trying to think of how to express the metric of "fairness" that makes this PR better than the current implementation and I think it's enough to just say that it reduces both the mean and standard deviation of the levels of logic each value goes through:

  | reduce | master | this PR
-- | -- | -- | --
  | 4 | 3 | 3
  | 4 | 3 | 3
  | 3 | 3 | 2
  | 2 | 3 | 2
  | 1 | 1 | 2
**Mean** | 2.8 | 2.6 | 2.4
**SD** | 1.3 | 0.9 | 0.5

jackkoenig (approved):

schoeberl:
About the fairness: with this PR the difference between paths is a maximum of 1, independent of the size. In the current version, the difference grows with the size of the Vec.

BTW, I cannot merge. I have no write access.

jackkoenig:
I've resolved the conflicts, run ScalaFmt, and fixed a minor inconsistency in code generation between Scala 2.12 and 2.13 (that I noticed elsewhere but figured we might as well fix while touching this code).

schoeberl:
Interesting little detail. Thanks.

PR #2327, opened 2022-01-05
Add BitPat.fromEnum to convert values from ChiselEnum

carlosedp:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

No impact to current API, the PR adds a new `fromEnum` function to BitPat generating BitPat from existing ChiselEnum values.

No change.

#### Desired Merge Strategy

#### Release Notes

This PR adds `BitPat.fromEnum` function enabling `BitPat` generation from existing `ChiselEnum` values.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -88,6 +88,13 @@ object BitPat {
     x.value.asUInt(x.getWidth.W)
   }
 
+    /** Creates a [[BitPat]] from a [[ChiselEnum]] literal.
+    *
+    * @param n the [[ChiselEnum]]
+    */
+  def fromEnum[T <: Data](n: T): UInt =
+    n.litValue.U((n.getWidth).W)
+
   /** Allows UInts to be used where a BitPat is expected, useful for when an
     * interface is defined with BitPats but not all cases need the partial
     * matching capability.
--- a/src/test/scala/chiselTests/util/BitPatSpec.scala
+++ b/src/test/scala/chiselTests/util/BitPatSpec.scala
@@ -3,9 +3,13 @@
 package chiselTests.util
 
 import chisel3.util.BitPat
+import chisel3.experimental.ChiselEnum
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
+object Enum extends ChiselEnum {
+  val VAL1, VAL2, VAL3 = Value
+}
 
 class BitPatSpec extends AnyFlatSpec with Matchers {
   behavior of classOf[BitPat].toString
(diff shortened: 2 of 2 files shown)


ekiwi:
Since `apply` is already overloaded for `UInt`, maybe it would make sense to also just overload it for `EnumType`. This way your example would be even easier to write.

carlosedp:
That's neat @ekiwi! Even better... thanks!

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -107,6 +101,14 @@ object BitPat {
     apply("b" + x.litValue.toString(2).reverse.padTo(len, "0").reverse.mkString)
   }
 
+    /**
+    * Allows [[ChiselEnum]] to be used where a BitPat is expected.
+    */
+  def apply(x: EnumType): BitPat = {
+    val len = if (x.isWidthKnown) x.getWidth else 0
+    apply("b" + x.litValue.toString(2).reverse.padTo(len, "0").reverse.mkString)
```
ekiwi, line 109: You could still re-use the apply function, i.e., do something like:
```
def apply(x: EnumType): BitPat = apply(x.litValue.U(n.getWidth).W)
```

carlosedp: 👍🏻

sequencer:
If directly cast to `UInt`, if might have negative influence to circuit PPA, since no mask is used, while actually there is `mask` for `MuxOH`, and we cannot ignore it.  

Think about another option:
what if an API like this?
```
def apply(x: EnumType*): BitPat = ???
```
There is a common use case that we need match multiple Enum which a single circuit.
For example, we need to use `BitPat` match `s1` and `s0` at the same time.
if `s1` and `s0` is OneHot type, we just set correspond bits to 1, while set other bits to 0.

sequencer:
This is an algorithm problem, if you have multiple encodings, for example, s0 -> 010, s1 -> 110, s2 -> 001.
How do you cover one(some) of them with minimal set.
It's simple in OneHot(just extract correspond bit).
However, for arbitrary user defined encoding, this problem is extended to the [set cover problem](https://en.wikipedia.org/wiki/Set_cover_problem), which is NP-hard.
Luckily, that's the problem, which is already solved by our decoder, you can construct a `TruthTable` in below:
```
s0 -> 1
s1 -> 1
s2 -> 0
default -> ?
```
to match a selection to `s0` and `s1`
The generated circuit should be optimized to minimal circuit to use.
In this case, state is always small and we don't wanna this simple function depends on `espresso`, so you can directly select `qmc` to handle this job.

carlosedp:
As an aside, I think we might need more documentation on [espresso](https://github.com/chipsalliance/espresso) requirements, install instructions and Chisel dependency on it (like what can and can't be made without it).

PR #2333, opened 2022-01-10
Optional clock param for memory ports

adkian-sifive:
Warn if clock at memory instantiation differs from clock bound at port creation, add tests to MultiClockSpec

### Contributor Checklist

#### Type of Improvement

  - code refactoring                  

   - new feature/API                   
 
#### API Impact

Apply functions for creating read/write accessors to Mem and SyncReadMem have an optional parameter `clock`. If this parameter is not passed, the clock for the port is inferred from the context. If the clock passed to the ports (implicit or otherwise) does not match the clock at memory instantiation, a warning is issued. 

No impact

#### Desired Merge Strategy

#### Release Notes

`Mem` and `SyncReadMem` read/write ports accept an optional parameter `clock` for binding a clock that is different from the implicit clock in the current context. 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Mem.scala
+++ b/core/src/main/scala/chisel3/Mem.scala
@@ -91,7 +91,7 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
   def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg
 
   def do_apply(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
-    do_apply_impl(idx, Builder.forcedClock, MemPortDirection.INFER, false)
+    do_apply_impl(idx, clock, MemPortDirection.INFER, false)
 
   /** Creates a read accessor into the memory with dynamic addressing. See the
     * class documentation of the memory for more detailed information.
--- a/core/src/main/scala/chisel3/Mem.scala
+++ b/core/src/main/scala/chisel3/Mem.scala
@@ -108,10 +108,20 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
   def do_read(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
     do_apply_impl(idx, clock, MemPortDirection.READ, false)
 
-  protected def do_apply_impl(idx: UInt, clock: Clock, dir: MemPortDirection, warn: Boolean)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+  protected def do_apply_impl(
+    idx:   UInt,
+    clock: Clock,
+    dir:   MemPortDirection,
+    warn:  Boolean
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): T = {
     if (warn && clock != clockInst) {
-      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. " +
-        "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0")
+      Builder.warning(
+        "The clock used to initialize the memory is different than the one used to initialize the port. " +
+          "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0"
+      )
     }
     makePort(sourceInfo, idx, dir, clock)
   }
@@ -127,10 +137,19 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
   def write(idx: UInt, data: T, clock: Clock)(implicit compileOptions: CompileOptions): Unit =
     write_impl(idx, data, clock, false)
 
-  private def write_impl(idx: UInt, data: T, clock: Clock, warn: Boolean)(implicit compileOptions: CompileOptions): Unit = {
+  private def write_impl(
+    idx:   UInt,
+    data:  T,
+    clock: Clock,
+    warn:  Boolean
+  )(
+    implicit compileOptions: CompileOptions
+  ): Unit = {
     if (warn && clock != clockInst) {
-      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. " +
-        "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0")
+      Builder.warning(
+        "The clock used to initialize the memory is different than the one used to initialize the port. " +
+          "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0"
+      )
     }
     implicit val sourceInfo = UnlocatableSourceInfo
     makePort(UnlocatableSourceInfo, idx, MemPortDirection.WRITE, clock) := data
@@ -146,23 +165,42 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
     * @note this is only allowed if the memory's element data type is a Vec
     */
   def write(
-    idx: UInt,
+    idx:  UInt,
     data: T,
     mask: Seq[Bool]
   )(
     implicit evidence: T <:< Vec[_],
-    compileOptions: CompileOptions
+    compileOptions:    CompileOptions
   ): Unit =
     masked_write_impl(idx, data, mask, Builder.forcedClock, true)
 
-  def write(idx: UInt, data: T, mask: Seq[Bool], clock: Clock) (implicit evidence: T <:< Vec[_], compileOptions: CompileOptions): Unit =
+  def write(
+    idx:   UInt,
+    data:  T,
+    mask:  Seq[Bool],
+    clock: Clock
+  )(
+    implicit evidence: T <:< Vec[_],
+    compileOptions:    CompileOptions
+  ): Unit =
     masked_write_impl(idx, data, mask, clock, false)
 
-  private def masked_write_impl(idx: UInt, data: T, mask: Seq[Bool], clock: Clock, warn: Boolean) (implicit evidence: T <:< Vec[_], compileOptions: CompileOptions): Unit = {
+  private def masked_write_impl(
+    idx:   UInt,
+    data:  T,
+    mask:  Seq[Bool],
+    clock: Clock,
+    warn:  Boolean
+  )(
+    implicit evidence: T <:< Vec[_],
+    compileOptions:    CompileOptions
+  ): Unit = {
     implicit val sourceInfo = UnlocatableSourceInfo
     if (warn && clock != clockInst) {
-      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. " +
-        "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0")
+      Builder.warning(
+        "The clock used to initialize the memory is different than the one used to initialize the port. " +
+          "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0"
+      )
     }
     val accessor = makePort(sourceInfo, idx, MemPortDirection.WRITE, clock).asInstanceOf[Vec[Data]]
     val dataVec = data.asInstanceOf[Vec[Data]]
@@ -178,9 +216,9 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
 
   private def makePort(
     sourceInfo: SourceInfo,
-    idx: UInt,
-    dir: MemPortDirection,
-    clock: Clock
+    idx:        UInt,
+    dir:        MemPortDirection,
+    clock:      Clock
   )(
     implicit compileOptions: CompileOptions
   ): T = {
@@ -290,7 +328,15 @@ sealed class SyncReadMem[T <: Data] private (t: T, n: BigInt, val readUnderWrite
   def read(x: UInt, y: Bool, z: Clock): T = macro SourceInfoTransform.xyzArg
 
   /** @group SourceInfoTransformMacro */
-  def do_read(addr: UInt, enable: Bool, clock: Clock, warn: Boolean)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+  def do_read(
+    addr:   UInt,
+    enable: Bool,
+    clock:  Clock,
+    warn:   Boolean
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): T = {
     val a = Wire(UInt())
     a := DontCare
     var port: Option[T] = None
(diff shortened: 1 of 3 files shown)


ekiwi:
How do you disable the warning if you want the memory ports to be in different clock domains?
Like how do you disable the warning, e.g., for the AsyncFifo in Rocket-Chip?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,42 +60,52 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
     */
   def apply(x: Int): T = macro SourceInfoTransform.xArg
 
+  def do_apply_impl(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+    if (clock != clockInst) {
+      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. This behavior will be depracated in 3.6.0")
```
mwachs5, line 65:
```suggestion
      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. This behavior will be deprecated in 3.6.0")
```

Don't we have a Chisel runtime deprecation we can use here?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,56 +60,57 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
     */
   def apply(x: Int): T = macro SourceInfoTransform.xArg
 
+  def do_apply_impl(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
```
jackkoenig, line 63:
```suggestion
  private def do_apply_impl(idx: UInt, clock: Clock, warn: Boolean)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
```

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -45,7 +45,7 @@ object Mem {
     do_apply(BigInt(size), t)(sourceInfo, compileOptions)
 }
 
-sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends HasId with NamedComponent with SourceInfoDoc {
+sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt, clockInst: Clock = Builder.forcedClock) extends HasId with NamedComponent with SourceInfoDoc {
   _parent.foreach(_.addId(this))
```
jackkoenig, line 50:
```suggestion
sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends HasId with NamedComponent with SourceInfoDoc {
  _parent.foreach(_.addId(this))

  private val clockInst: Clock = Builder.forcedClock
```

This class is sealed and we never seem to actually pass a different clock as an argument so I think it's more clear just to make it a private val.

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,56 +60,57 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
     */
   def apply(x: Int): T = macro SourceInfoTransform.xArg
 
+  def do_apply_impl(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+    if (clock != clockInst) {
+      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. This behavior will be deprecated in 3.6.0")
```
jackkoenig, line 65:
```suggestion
      Builder.warning("The clock used to initialize the memory is different than the one used to initialize the port. " + 
        "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0")
```

core/src/main/scala/chisel3/Mem.scala
jackkoenig: Should be private and we need the ability to suppress the warning

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,56 +60,57 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
     */
   def apply(x: Int): T = macro SourceInfoTransform.xArg
 
+  def do_apply_impl(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
+    if (clock != clockInst) {
```
jackkoenig, line 64:
```suggestion
    if (warn && clock != clockInst) {
```

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,56 +60,57 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
...
     * See the class documentation of the memory for more detailed information.
     */
-  def apply(x: UInt): T = macro SourceInfoTransform.xArg
+  def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg
 
   /** @group SourceInfoTransformMacro */
-  def do_apply(idx: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
-    makePort(sourceInfo, idx, MemPortDirection.INFER)
+  def do_apply(idx: UInt, clock: Clock = Builder.forcedClock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
+    do_apply_impl(idx, clock)
```
jackkoenig, line 89:
```suggestion
  def do_apply(idx: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
    do_apply_impl(idx, Builder.forcedClock, true)
    
  def do_apply(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
    do_apply_impl(idx, clock, false)
```

You need different versions of these because you need to suppress the warning when the clock is passed manually.

Same with versions below.

jackkoenig (changes requested):

core/src/main/scala/chisel3/Mem.scala
jackkoenig: This message (along with my other suggestions) should address @ekiwi's concern.

adkian-sifive:
@ekiwi @jackkoenig Thanks for the review - I've added the warn condition. I'm thinking I can either use the same pattern for reads and writes (with `do_read_impl`/`do_write_impl`), or add a `direction` parameter to `do_apply_impl` and call that from the read/write handlers - thoughts?

core/src/main/scala/chisel3/Mem.scala
adkian-sifive: I think I need protected here since I need to access it from the SyncReadMem subclass

jackkoenig: > I think I need protected here since I need to access it from the SyncReadMem subclass

Ah yes, correct, but definitely protected instead of public.

jackkoenig:
@ekiwi any idea what's up with this chiseltest failure? https://github.com/chipsalliance/chisel3/runs/4768585511?check_suite_focus=true

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -74,28 +83,45 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
     * See the class documentation of the memory for more detailed information.
     */
   def apply(x: UInt): T = macro SourceInfoTransform.xArg
+  def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg
```
jackkoenig, line 86: This is sort of an undocumented convention, but can you put this apply method above the `do_apply` it expands to? That makes it much easier to find the implementation of the method.

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -60,42 +61,67 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt) extends H
...
+        "If this is intentional, please pass the clock explicitly when creating the port. This behavior will be an error in 3.6.0")
+    }
+    makePort(sourceInfo, idx, dir, clock)
+  }
+
+  /** Creates a read/write accessor into the memory with dynamic addressing.
+    * See the class documentation of the memory for more detailed information.
+    */
+  def apply(x: UInt): T = macro SourceInfoTransform.xArg
+  def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg
```
jackkoenig, line 76: I don't think I phrased this as precisely as I meant to. Can you make sure each `def <method>` is immediately above the corresponding `def do_<method>` (with only an empty line between them)?

More like:
```scala
  def apply(x: UInt): T = macro SourceInfoTransform.xArg

  def do_apply(idx: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = ...

   def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg

  def do_apply(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = ...
```

Does that make sense?

adkian-sifive: Oh i see what you mean

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -85,7 +86,12 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
 
   /** @group SourceInfoTransformMacro */
   def do_apply(idx: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
-    makePort(sourceInfo, idx, MemPortDirection.INFER)
+    do_apply_impl(idx, Builder.forcedClock, MemPortDirection.INFER, true)
+
+  def apply(x: UInt, y: Clock): T = macro SourceInfoTransform.xyArg
+
+  def do_apply(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
+    do_apply_impl(idx, Builder.forcedClock, MemPortDirection.INFER, false)
```
jackkoenig, line 94:
```suggestion
  def do_apply(idx: UInt, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
    do_apply_impl(idx, clock, MemPortDirection.INFER, false)
```

adkian-sifive: Fixed

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -244,23 +323,44 @@ object SyncReadMem {
...
+    do_read(idx, en, Builder.forcedClock, true)
+
+  def read(x: UInt, y: Bool, z: Clock): T = macro SourceInfoTransform.xyzArg
+
+  /** @group SourceInfoTransformMacro */
+  def do_read(idx: UInt, en: Bool, clock: Clock)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T =
+    do_read(idx, en, clock, false)
+
+  /** @group SourceInfoTransformMacro */
+  def do_read(
```
adkian-sifive, line 346: Should this be `private do_read_impl`?

macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala
```diff
@@ -200,6 +200,10 @@ class SourceInfoTransform(val c: Context) extends AutoSourceTransform {
     q"$thisObj.$doFuncTerm($x, $y)($implicitSourceInfo, $implicitCompileOptions)"
   }
 
+  def xyzArg(x: c.Tree, y: c.Tree, z: c.Tree): c.Tree = {
+    q"$thisObj.$doFuncTerm($x, $y, $z)($implicitSourceInfo, $implicitCompileOptions)"
+  }
```
jackkoenig, line 205:
```suggestion
  def xyzArg(idx: c.Tree, en: c.Tree, clock: c.Tree): c.Tree = {
    q"$thisObj.$doFuncTerm($idx, $en, $clock)($implicitSourceInfo, $implicitCompileOptions)"
  }
```

src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -129,6 +129,73 @@ class MultiClockSpec extends ChiselFlatSpec {
...
+
+      val mem = withClock(myClock) { Mem(4, UInt(8.W)) }
+      val port0 = mem(0.U, myClock)
+    }
+
+    val (logMemInst, _) = grabLog(ChiselStage.elaborate(new MyModuleMemInstClock))
+    logMemInst should include("memory is different")
+
+    val (logPortInst, _) = grabLog(ChiselStage.elaborate(new MyModulePortInstClock))
+    (logPortInst should not).include("memory is different")
```
jackkoenig, line 158:
```suggestion
    logPortInst shouldNot include("memory is different")
```

src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -129,6 +129,73 @@ class MultiClockSpec extends ChiselFlatSpec {
...
+
+    val (logNoWarn, _) = grabLog(ChiselStage.elaborate(new MyModuleNoWarn))
+    (logNoWarn should not).include("memory is different")
```
jackkoenig, line 161:
```suggestion
    logNoWarn shouldNot include("memory is different")
```

src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -129,6 +129,73 @@ class MultiClockSpec extends ChiselFlatSpec {
     })
   }
 
+  "Differing clocks at memory and port instantiation" should "warn" in {
```
jackkoenig, line 132: The follow 3 cases should be split into differently named tests. This name only applies to the first one, the next 2 are both **not** warnings

src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -129,6 +129,73 @@ class MultiClockSpec extends ChiselFlatSpec {
...
+    logMemInst should include("memory is different")
+
+    val (logPortInst, _) = grabLog(ChiselStage.elaborate(new MyModulePortInstClock))
+    (logPortInst should not).include("memory is different")
+
+    val (logNoWarn, _) = grabLog(ChiselStage.elaborate(new MyModuleNoWarn))
+    (logNoWarn should not).include("memory is different")
+  }
+
+  "Differing clocks at sync memory and port instantiation" should "warn" in {
```
jackkoenig, line 164: Same issue here

src/test/scala/chiselTests/MultiClockSpec.scala
```diff
@@ -129,6 +129,73 @@ class MultiClockSpec extends ChiselFlatSpec {
...
+
+    val (logMemInst, _) = grabLog(ChiselStage.elaborate(new MyModuleMemInstClock))
+    logMemInst should include("memory is different")
+
+    val (logPortInst, _) = grabLog(ChiselStage.elaborate(new MyModulePortInstClock))
+    (logPortInst should not).include("memory is different")
+
+    val (logExplicitClk, _) = grabLog(ChiselStage.elaborate(new MyModuleExplicitClk))
+    (logExplicitClk should not).include("memory is different")
+  }
```
jackkoenig, line 197: These tests are good, and it's a bit annoying, but every changed method needs to be tested, each `read` and `write` for both with and without the warning.

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -122,9 +171,38 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt)
   )(
     implicit evidence: T <:< Vec[_],
     compileOptions:    CompileOptions
+  ): Unit =
+    masked_write_impl(idx, data, mask, Builder.forcedClock, true)
+
+  def write(
```
jackkoenig, line 177: Please add ScalaDoc like the `write` method above does (for every added public, non-do_ method). You can basically copy the ScalaDoc but add something about the `clock` parameter and what it's for.

jackkoenig (changes requested):
This looks good. Mainly nits and requests for ScalaDoc and more tests.

jackkoenig (approved):
LGTM! Nice work @adkian-sifive!

PR #2363, opened 2022-01-26
Fix Decoder bug for constant 0 and DC

sequencer:
### Minimal Reproducible Tests
Should be optimized to 0
```
00->0
01->?
10->0
11->0
    ?
```
Should be optimized to `DontCare`
```
00->?
01->?
10->?
11->?
    ?
```

### Reason of this bug
QMC will optimize to empty for case 0 and case `DontCare`, since it only cares `ON set`.
Espresso by default(`-of`) will return `ON set` as well, so it will also be empty on these two cases.
Espresso will also return another result by `-ofd`(However that's the return which PLA cannot consume).
```
.type fd
.i 2
.o 1
.p 1
01 2
.end
```
So this behavior is because `PLA`, which is the circuit backend of these logic minimizer, only consumes `ON set`.
Both QMC and espresso algorithm are optimizing to a PLA, which means `?` and `0` will finally land to 'OFF set'. This will have a correct behavior for PLA. 
This seems less optimized: circuit result should be `DontCare`, but not `0`, however, if the minimizer backend is PLA, it cannot be a `DontCare`. 
So when user using `Minimizer` output decoder, they be get less optimized result.(Maybe we should private `Minimizer`?)
 
### Two options to resolve 
1. Throw an exception if a TruthTable encountered a constant 0 or don't care result. This won't break compatibility, but need user handle empty TruthTable exception on their own. In this case A potential bug will be figured out during elaboration: 
*some control logic is not coupling to decode input(instruction or microOp)*, rather than tie 0, it will result an error similar to `unconnected signal` since, essentially, this control signal is not driven. 

2. Return `0.U`.
I thought this is bad, but since the backend is PLA: we can only return 0 to PLA to have a better logic. This does fix the issue. But won't trap the 'unconnected signal' error.

### So the problem becomes:
Should we return `0.U` directly? The current implementation of this PR is return `0.U` directly.
Or throw this error to user. It may be able to figure out some corner bug, but unexperienced user might happy with this.

### Contributor Checklist

#### Type of Improvement
- bug fix

#### API Impact
none

none

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/minimizer/MinimizerSpec.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/minimizer/MinimizerSpec.scala
@@ -272,4 +272,39 @@ trait MinimizerSpec extends AnyFlatSpec with ChiselScalatestTester with Formal {
       BitPat(s"b${Seq(N, X, X, X, X, X, X, X, X, A2_X, A1_X, IMM_X, DW_X, FN_X, N, M_X, X, X, X, X, X, X, X, CSR_X, X, X, X, X).reduce(_ + _)}")
     ))
   }
+
+  "output is 0" should "pass" in {
+    minimizerTest(TruthTable.fromString(
+      """00->0
+        |01->?
+        |10->0
+        |11->0
+        |    ?
+        |""".stripMargin
+
+    ))
+  }
+  "output is 1" should "pass" in {
+    minimizerTest(TruthTable.fromString(
+      """00->1
+        |01->?
+        |10->1
+        |11->1
+        |    ?
+        |""".stripMargin
+
+    ))
+  }
+  // I know this seems to be crazy, but if user is crazy as well...
+  "output is dont care" should "pass" in {
+    minimizerTest(TruthTable.fromString(
+      """00->?
+        |01->?
+        |10->?
+        |11->?
+        |    ?
+        |""".stripMargin
+
+    ))
+  }
 }
--- a/src/main/scala/chisel3/util/experimental/decode/EspressoMinimizer.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/EspressoMinimizer.scala
@@ -57,11 +57,20 @@ object EspressoMinimizer extends Minimizer with LazyLogging {
     def readTable(espressoTable: String) = {
       def bitPat(espresso: String): BitPat = BitPat("b" + espresso.replace('-', '?'))
 
-      espressoTable
+      val out = espressoTable
         .split("\n")
         .filterNot(_.startsWith("."))
         .map(_.split(' '))
         .map(row => bitPat(row(0)) -> bitPat(row(1)))
+      // special case for 0 and DontCare, if output is not couple to input
+      if (out.isEmpty)
+        Array(
+          (
+            BitPat(s"b${"?" * table.inputWidth}"),
+            BitPat(s"b${"0" * table.outputWidth}")
+          )
+        )
+      else out
     }
 
     val input = writeTable(table)
--- a/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/QMCMinimizer.scala
@@ -304,24 +304,37 @@ object QMCMinimizer extends Minimizer {
       )
     })
 
-    minimized.tail.foldLeft(table.copy(table = Seq(minimized.head))) {
-      case (tb, t) =>
-        if (tb.table.exists(x => x._1 == t._1)) {
-          tb.copy(table = tb.table.map {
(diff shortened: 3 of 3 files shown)


aswaterman:
I think this is OK for practical reasons: we don’t know how wide to make the result, because we didn’t get any sample inputs.

(I don’t agree that an empty truth table is meaningless, though: it means the same thing as a nonempty truth table for which all of the RHS values are ?, i.e., the implementation can correctly return whatever value it wants.)

sequencer:
> I don’t agree that an empty truth table is meaningless, though: it means the same thing as a nonempty truth table for which all of the RHS values are ?, i.e., the implementation can correctly return whatever value it wants.

I do agree this, if decoder use this empty `TruthTable`, it should return `DontCare`.

There are 3 options:
1. throw an exception if user pass an empty `TruthTable` to decoder.
This won't break compatibility, but need user handle empty `TruthTable` exception on their own.

2. let `TruthTable` return `DontCare` if catching an empty `TruthTable`.
This will break compatibility, since `DontCare` is not a `UInt`. But essentially this is the best idea to make circuit being able to be optimized by synthesis flow.

3. return `0.U`.
IIRC, SFC will always make `DontCare` as `0.U`, so this is also reasonable. But I think this changes the semantic, so I personally don't like this.

So I choose the first option which directly throw the exception, and let user handle this like what I did in chipsalliance/rocket-chip#2836.

aswaterman:
Yeah, I think this is a pragmatic choice.

jackkoenig:
I think this is pragmatic, but it doesn't really fix https://github.com/chipsalliance/rocket-chip/pull/2836 does it? What's the real issue there?

sequencer:
> What's the real issue there?

I realized this is a bug that if a `TruthTable` will be minimized to empty(although it wasn't empty before)

integration-tests/src/test/scala/chiselTests/util/experimental/minimizer/MinimizerSpec.scala
```diff
@@ -272,4 +272,39 @@ trait MinimizerSpec extends AnyFlatSpec with ChiselScalatestTester with Formal {
...
+        |110000000010?????????????1010011->0
+        |110000000011?????????????1010011->0
+        |110100000000?????????????1010011->0
+        |110100000001?????????????1010011->0
+        |110100000010?????????????1010011->0
+        |110100000011?????????????1010011->0
+        |111000000000?????000?????1010011->0
+        |111000000000?????001?????1010011->0
+        |111100000000?????000?????1010011->0
+        |                                  ?""".stripMargin))
```
sequencer, line 308: From this table(copy from RC). it semantically means just return 0 is enough.
However Decoder didn't cover this.

sequencer:
After decoding, output might not couple to input, see the newly added bug, which I pulled from RC.
There will be such corner cases:
1. minimized result should be a constant(not couple to input), output is `0` or `1`
2. all out bits are `?`, output is `DontCare`
3. input conflict

For the first case: I think pull down/up signal to a constant is reasonable. I think this might be a bug to `TruthTable` which I should fix.
For the second case: `apply` should return `DontCare`, but type is incorrect, I'd like to throw exception to user like what I proposed.
For the third case: I just checked it can be detect in current codes. I think this won't be a bug.

sequencer:
Sure... It seems that corner case found both QMC and espresso bug...
Working a fix...

sequencer:
Here are three different tables:
output is constant 0:
```
espresso input table:
.i 2
.o 1
.type fr
00 0
01 -
10 0
11 0

espresso output table:
.i 2
.o 1
.p 0
.e
```

output is constant 1:
```
output is 1
Elaborating design...
espresso input table:
.i 2
.o 1
.type fr
00 1
01 -
10 1
11 1

espresso output table:
.i 2
.o 1
.p 1
-- 1
.e


??->1
    ?
```

output is dont care
```
espresso input table:
.i 2
.o 1
.type fr
00 -
01 -
10 -
11 -

espresso output table:
.i 2
.o 1
.p 0
.e
```

we can found both constant 1 and don'tcare behave correct, but constant 0 is wrong.

build.sc
```diff
@@ -8,45 +8,41 @@ import mill.contrib.buildinfo.BuildInfo
 
 object chisel3 extends mill.Cross[chisel3CrossModule]("2.13.6", "2.12.14")
```
jackkoenig, line 10: It would be better to make these build changes separately

jackkoenig (approved):
LGTM thanks @sequencer!

jackkoenig:
I think you're getting bitten by Scalafmt

build.sc
sequencer: I should create another PR for mill. I forgot it. ;p

PR #2364, opened 2022-01-26
reduceTree() cleanup

oharboe:
a bit cleaner implementation of the fix in 55cb58877aca898f1ccae5edf29aeede9d1b71ba

### Contributor Checklist

- [ ] Did you review?


jackkoenig (changes requested):
I think the existing implementation is actually fine. `Iterators` can be hard to reason about because they are only traversable once, so mixing a `var` with an `Iterator` is a recipe for disaster. I know this because I have been bitten by it 🙂 https://stackoverflow.com/questions/61829366/why-does-this-iterator-infinitely-loop.

PR #2365, opened 2022-01-26
Update supported val modifiers for @public

azidar:
### Contributor Checklist

#### Type of Improvement

  - bug fix

#### API Impact

Now supports more modifiers on the val's marked with `@public`, including `final` and `override`.

#### Desired Merge Strategy

#### Release Notes

Support more `val` modifiers like `final` and `override` for `@public`.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
+++ b/src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
@@ -303,7 +303,11 @@ class InstanceSpec extends ChiselFunSpec with Utils {
         val value = 10
         val overriddenVal = 10
       }
-      @instantiable class SubClass() extends SupClass {
+      trait SupTrait {
+        def x: Int
+        def y: Int
+      }
+      @instantiable class SubClass() extends SupClass with SupTrait {
         // This errors
         //@public private val privateVal = 10
         // This errors
@@ -312,6 +316,8 @@ class InstanceSpec extends ChiselFunSpec with Utils {
         @public final val finalVal = 12
         @public lazy val lazyValue = 12
         @public val value = value
+        @public final override lazy val x: Int = 3
+        @public override final lazy val y: Int = 4
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -298,6 +298,21 @@ class InstanceSpec extends ChiselFunSpec with Utils {
       annos should contain(MarkAnnotation("~Top|Top/i:HasEither>x".rt, "xright"))
       annos should contain(MarkAnnotation("~Top|Top/i:HasEither>y".rt, "yleft"))
     }
+    it("3.12: should properly support val modifiers") {
+      class SupClass extends Module {
+        val value = 10
+        val overriddenVal = 10
+      }
+      @instantiable class SubClass() extends SupClass {
+        // This errors
```
mwachs5, line 307: you could add a `should not compile` check if you want to check the "this errors" case

mwachs5: what about `final`?

azidar: Yeah I tried it, but I think that functionality interacts incorrectly with the `@public` macros because it itself is a macro, so it ends up not working.

src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -298,6 +298,22 @@ class InstanceSpec extends ChiselFunSpec with Utils {
...
+        //@public private val privateVal = 10
+        // This errors
+        //@public protected val protectedVal = 10
+        @public override val overriddenVal = 12
+        @public final val finalVal = 12
+        @public lazy val lazyValue = 12
+        @public val value = value
```
jackkoenig, line 314: Can you also try mixing some modifiers? eg.

```scala
trait Foo {
  def x: Int
  def y: Int
}

class Bar extends Foo {
  final override lazy val x: Int = 3
  override final lazy val y: Int = 4
}
```

jackkoenig (approved):
LGTM! Thanks!

PR #2376, opened 2022-02-01
Improve error reporting

jackkoenig:
Inspired by @oharboe's comments here: https://github.com/ucb-bar/chiseltest/pull/501#issuecomment-1024878100.

This gives a little bit of refinement to error reporting.
1. Making sure that our stack trace trimming doesn't accidentally add stack traces to exceptions that don't have them otherwise.
2. Providing a name to the exception thrown by collected errors so that it doesn't have `$$anon$1` in the error message
3. Add `--throw-on-first-error` to give users the option to throw exceptions instead of aggregating recoverable errors

### Contributor Checklist

#### Type of Improvement

 - code cleanup  

#### API Impact

This changes how errors are reported to make them a little more clear

No impact

#### Desired Merge Strategy

   - Squash

#### Release Notes

Improve recoverable error reporting and add `--throw-on-first-error` for throwing exceptions instead of aggregating recoverable errors.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Error.scala
+++ b/core/src/main/scala/chisel3/internal/Error.scala
@@ -246,7 +246,10 @@ private[chisel3] class ErrorLog {
     }
 
     if (!allErrors.isEmpty) {
-      throw new Errors("Fatal errors during hardware elaboration. Look above for error list.")
+      throw new Errors(
+        "Fatal errors during hardware elaboration. Look above for error list. " +
+          "Rerun with --throw-on-first-error if you wish to see a stack trace."
+      )
     } else {
       // No fatal errors, clear accumulated warnings since they've been reported
       errors.clear()
--- a/core/src/main/scala/chisel3/internal/Error.scala
+++ b/core/src/main/scala/chisel3/internal/Error.scala
@@ -46,31 +46,36 @@ object ExceptionHelpers {
       }
 
       // Step 1: Remove elements from the top in the package trimlist
-      ((a: Array[StackTraceElement]) => a.dropWhile(inTrimlist))
-        // Step 2: Optionally remove elements from the bottom until the anchor
-        .andThen(_.reverse)
-        .andThen(a =>
-          anchor match {
-            case Some(b) => a.dropWhile(ste => !ste.getClassName.startsWith(b))
-            case None    => a
-          }
-        )
-        // Step 3: Remove elements from the bottom in the package trimlist
-        .andThen(_.dropWhile(inTrimlist))
-        // Step 4: Reverse back to the original order
-        .andThen(_.reverse.toArray)
-        // Step 5: Add ellipsis stack trace elements and "--full-stacktrace" info
-        .andThen(a =>
-          ellipsis() +:
-            a :+
-            ellipsis() :+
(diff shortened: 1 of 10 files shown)


core/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -46,31 +47,36 @@ object ExceptionHelpers {
...
+                Some("Stack trace trimmed to user code only. Rerun with --full-stacktrace to see the full stack trace")
+              )
+          )
+          // Step 5: Mutate the stack trace in this exception
+          .andThen(throwable.setStackTrace(_))
+
+      val stackTrace = throwable.getStackTrace
+      if (stackTrace.nonEmpty) {
+        trimStackTrace(stackTrace)
+      }
```
seldridge, line 79:
```suggestion
      val stackTrace = throwable.getStackTrace
      if (stackTrace.nonEmpty) 
        trimStackTrace(stackTrace)
```

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -61,6 +61,24 @@ case object PrintFullStackTraceAnnotation
 
 }
 
+/** On recoverable errors, this will cause Chisel to throw an exception instead of continuing.
+  */
+case object ThrowOnFirstErrorAnnotation
```
seldridge, line 66: Is this expected to be used as an annotation? If not, then this can be made `private[chisel3.stage]`.

src/test/scala/chiselTests/stage/ChiselStageSpec.scala
```diff
@@ -207,4 +241,53 @@ class ChiselStageSpec extends AnyFlatSpec with Matchers with Utils {
     exception.getStackTrace.mkString("\n") should include("java")
   }
 
+  it should "NOT add a stack trace to an exception with no stack trace" in {
+    val exception = intercept[java.lang.Exception] {
+      (new ChiselStage).emitChirrtl(new UserExceptionNoStackTrace)
```
seldridge, line 246: Nit: only `elaborate` should be needed here (though there is no cost to using `emitChirrtl`.

seldridge (approved):
lgtm

Nice usage of `info` in the tests.

core/src/main/scala/chisel3/internal/Error.scala
jackkoenig: I actually prefer the explicit curly braces for mutating conditional blocks. It's a little thing but GOTO Fail will never leave my brain.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
jackkoenig: It will be needed by chiseltest which only exposes Annotations and not arguments, but good suggestion in general

src/test/scala/chiselTests/stage/ChiselStageSpec.scala
jackkoenig: It seems that `class ChiselStage` doesn't have `elaborate` 🤦‍♂️

PR #2381, opened 2022-02-01
Emit FIRRTL bulkconnects whenever possible

jared-barocsi:
This PR aims to emit a FIRRTL bulkconnect when connecting `Aggregate`s either mono- or bi-directionally. This does not cover all possible cases (e.g. ExpandConnects + Dedup), but instead serves as a baseline for most cases when a bulk connect would work.

### Contributor Checklist

- [x Did you specify the type of improvement?
- [] Did you add appropriate documentation in `docs/src`?

#### Type of Improvement

- bug fix
- performance improvement

#### API Impact

No impact

No impact

#### Desired Merge Strategy

Squash and merge

#### Release Notes
Emit FIRRTL bulk connects whenever possible when using a `MonoConnect` or `BiConnect`

- [ ] Did you review?


core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -218,6 +219,91 @@ private[chisel3] object BiConnect {
...
+    // TODO do we need elementwise check for bundles?
+//    val elemsMatch = if(sink.elements.size == source.elements.size) {
+//      val elemValidConnect = sink.elements.zip(source.elements).map {
+//        case (sink, source) => CheckTypes.validConnect(
+//          Converter.extractType(sink._2, sourceInfo),
+//          Converter.extractType(source._2, sourceInfo))
+//        case _ => false
+//      }
+//      elemValidConnect.foldLeft(true)(_ && _)
+//    } else false
```
jackkoenig, line 255: Please delete any commented code

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -122,29 +128,24 @@ private[chisel3] object BiConnect {
...
+
+        // chisel3 <> is commutative but FIRRTL <- is not
+        val flipConnection =
+          !MonoConnect.canBeSink(left_r, context_mod) || !MonoConnect.canBeSource(right_r, context_mod)
+        val (newLeft, newRight) = if (flipConnection) (right_r, left_r) else (left_r, right_r)
+
+        // Check whether Records can be bulk connected (all elements can be connected)
+        if (canBulkConnectAggregates(newLeft, newRight, sourceInfo, connectCompileOptions, context_mod)) {
+          val leftReified:  Data = if (isView(newLeft)) reifySingleData(newLeft).get else newLeft
+          val rightReified: Data = if (isView(newRight)) reifySingleData(newRight).get else newRight
```
jackkoenig, line 145: You're reifying here but also again inside of `canBulkConnectAggregates`, I think you only need to reify once and pass it to `canBulkConnectAggregates`. Now I do note the `.get` when you reify here, ie. `reifySingleData(newLeft).get`. Note that if `reifySingleData` returns `None`, then you can't bulk connect. So if you just check if each Data is a view and `reifySingleData` if it is, then you can guard against trying to bulk connect if either is `None`.

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -218,6 +219,91 @@ private[chisel3] object BiConnect {
...
+          case Some(data: Aggregate) =>
+            sinkReified = data
+            Converter.extractType(data, sourceInfo)
+          case None => None
+        }
+        else Converter.extractType(sink, sourceInfo)
+
+      if (sourceType != sinkType || sourceType == None)
+        return false
+    }
```
jackkoenig, line 283: I think all of this view/reification logic is unnecessary if you reify outside of this method.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -190,6 +206,138 @@ private[chisel3] object MonoConnect {
       case (sink, source) => throw MismatchedException(sink, source)
     }
 
+  /** Check [[Aggregate]] visibility. */
+  private[chisel3] def aggregateConnectContextCheck(
```
jackkoenig, line 210: What does `true` mean and what does `false` mean? It would be helpful if you could expand this comment to say that it checks contexts (and throws exceptions) as well as with the information about what the return value means.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -190,6 +206,138 @@ private[chisel3] object MonoConnect {
...
+    sourceInfo:            SourceInfo,
+    connectCompileOptions: CompileOptions,
+    context_mod:           RawModule
+  ): Boolean = {
+    // Assuming we're using a <>, check if a bulk connect is valid in that case
+    val biConnectCheck =
+      BiConnect.canBulkConnectAggregates(sink, source, sourceInfo, connectCompileOptions, context_mod)
+
+    // Check that the Aggregate's child signals are all strictly inputs (not bidirectional)
+    val sinkIsInputCheck: Boolean = sink.direction == ActualDirection.Output
```
jackkoenig, line 336: What happens if you try to bulk connect internal wires?
```scala
val w1 = Wire(Vec(2, UInt(8.W)))
val w2 = Wire(Vec(2, UInt(8.W)))
w2 := w1
```
This should be a legal bulk connect.

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -218,6 +219,91 @@ private[chisel3] object BiConnect {
...
+    // sink must be writable
+    val bindingCheck = sink.topBinding match {
+      case _: ReadOnlyBinding => false
+      case _ => true
+    }
+
+    // check data can flow between provided aggregates
+    val flow_check = MonoConnect.canBeSink(sink, context_mod) && MonoConnect.canBeSource(source, context_mod)
+
+    typeCheck && contextCheck && bindingCheck && flow_check
```
jackkoenig, line 304: Because `&&` can early out if false, it would be better to make all of the above `vals` be `defs` (so long as each one is only accessed once) so that the logic is only invoked if each of the proceeding checks were true. We do need to make sure though that we only access each one of these once and put a comment on each one to make sure it's clear why they're defs.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -129,12 +131,19 @@ private[chisel3] object MonoConnect {
...
         if (sink_v.length != source_v.length) { throw MismatchedVecException }
-        for (idx <- 0 until sink_v.length) {
-          try {
-            implicit val compileOptions = connectCompileOptions
-            connect(sourceInfo, connectCompileOptions, sink_v(idx), source_v(idx), context_mod)
-          } catch {
-            case MonoConnectException(message) => throw MonoConnectException(s"($idx)$message")
+        if (canBulkConnectAggregates(sink_v, source_v, sourceInfo, connectCompileOptions, context_mod)) {
+          val sinkReified:   Data = if (isView(sink_v)) reifySingleData(sink_v).get else sink_v
+          val sourceReified: Data = if (isView(source_v)) reifySingleData(source_v).get else source_v
```
jackkoenig, line 136: Same reification comment as above.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -150,19 +159,26 @@ private[chisel3] object MonoConnect {
...
-        for ((field, sink_sub) <- sink_r.elements) {
-          try {
-            source_r.elements.get(field) match {
-              case Some(source_sub) => connect(sourceInfo, connectCompileOptions, sink_sub, source_sub, context_mod)
-              case None => {
-                if (connectCompileOptions.connectFieldsMustMatch) {
-                  throw MissingFieldException(field)
+        if (canBulkConnectAggregates(sink_r, source_r, sourceInfo, connectCompileOptions, context_mod)) {
+          val sinkReified:   Data = if (isView(sink_r)) reifySingleData(sink_r).get else sink_r
+          val sourceReified: Data = if (isView(source_r)) reifySingleData(source_r).get else source_r
```
jackkoenig, line 164: Same reification comment as above.

jackkoenig (changes requested):
This is very exciting! Nice work! I have some suggestions for code improvements as well as 1 suggested case that I suspect isn't currently supported.

jackkoenig (approved):
LGTM. Great work! Thanks!

jackkoenig:
I meant to mark this for backporting but forgot.

jackkoenig:
@Mergifyio backport 3.5.x

PR #2387, opened 2022-02-02
Make truthtable being able to accept unknown input width.

sequencer:
During the migration to RC.
I encountered this issue: sometime input or output width are unknown.
I originally think the input to decoder was a set of MicroOp, which is certainly a known width. in RC CSR, it shows the usage as an address decoder. I think it's reasonable to have an input unknown width.
However in the output side, the result of decode logic is always a control logic, padding it is an anti-pattern design.
So, this PR makes TruthTable accepting an unknown width to fit to the use cases like RC CSR.

### Contributor Checklist

#### Type of Improvement
- bug fix

#### API Impact
TruthTable will accept unknown width input now.
None

#### Desired Merge Strategy

#### Release Notes
TruthTable will accept unknown width input now.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
@@ -31,10 +31,14 @@ object TruthTable {
 
   /** Convert a table and default output into a [[TruthTable]]. */
   def apply(table: Iterable[(BitPat, BitPat)], default: BitPat, sort: Boolean = true): TruthTable = {
-    require(table.map(_._1.getWidth).toSet.size == 1, "input width not equal.")
+    val inputWidth = table.map(_._1.getWidth).max
     require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
     val outputWidth = table.map(_._2.getWidth).head
-    val mergedTable = table
+    val mergedTable = table.map {
+      case (in, out) =>
+        // pad input signals.
+        (BitPat.dontCare(inputWidth - in.getWidth) ## in, out)
+    }
       .groupBy(_._1.toString)
       .map {
         case (key, values) =>
--- a/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
@@ -35,9 +35,10 @@ object TruthTable {
     require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
     val outputWidth = table.map(_._2.getWidth).head
     val mergedTable = table.map {
-      case (in, out) =>
-        // pad input signals.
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -31,10 +31,14 @@ object TruthTable {
...
-    require(table.map(_._1.getWidth).toSet.size == 1, "input width not equal.")
+    val inputWidth = table.map(_._1.getWidth).max
     require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
     val outputWidth = table.map(_._2.getWidth).head
-    val mergedTable = table
+    val mergedTable = table.map {
+      case (in, out) =>
+        // pad input signals.
+        (BitPat.dontCare(inputWidth - in.getWidth) ## in, out)
+    }
```
jackkoenig, line 41:
```suggestion
    val mergedTable = table.map {
      // pad input signals if necessary
      case (in, out) if inputWidth > in.width =>
        (BitPat.dontCare(inputWidth - in.width) ## in, out)
      case (in, out) => (in, out)
    }
```
I'm not sure that there's any negative to unconditionally padding but it might be worth only padding if necessary? Any thoughts?

I also switched it to use `.width` instead of `.getWidth` because the latter reminds me of `Data.getWidth` which can fail. `.width` is more clear that this is always known (for BitPats).

src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
```diff
@@ -31,10 +31,14 @@ object TruthTable {
...
   def apply(table: Iterable[(BitPat, BitPat)], default: BitPat, sort: Boolean = true): TruthTable = {
-    require(table.map(_._1.getWidth).toSet.size == 1, "input width not equal.")
+    val inputWidth = table.map(_._1.getWidth).max
     require(table.map(_._2.getWidth).toSet.size == 1, "output width not equal.")
     val outputWidth = table.map(_._2.getWidth).head
-    val mergedTable = table
+    val mergedTable = table.map {
+      case (in, out) =>
+        // pad input signals.
+        (BitPat.dontCare(inputWidth - in.getWidth) ## in, out)
```
jackkoenig, line 40: Also is don't care padding correct? If someone creates a `BitPat` from `0.U`, the implication to me would be zero padding, not dontcare padding.

sequencer: My bad... I'll change this after wake up.

sequencer: Sorry, it seems I took 6 days to weak up :(

jackkoenig (approved):
LGTM!

PR #2390, opened 2022-02-03
Tweak Bundle._elementsImpl

jackkoenig:
* Change type of `Bundle._elementsImpl` to Iterable

It was previously SeqMap (ListMap on Scala 2.12). This change gives us
more freedom to optimize the implementation without breaking binary
compatibility. It is scala.collection.Iterable because it is perfectly
fine to return mutable collections (like Arrays) since the only use is
to Iterate on them.

* Disallow users implementing `Bundle._elementsImpl`

Currently, it would result in a runtime linkage error. This turns it
into a compile-time error.

### Contributor Checklist

Tweak to unreleased feature, no release notes necessary

#### Type of Improvement

 - code refactoring    

#### API Impact

This widens the type of `Bundle._elementsImpl` to `Iterable` so that we can be more flexible in the future without breaking binary compatibility. Technically user visible because `_elementsImpl` is `protected`, but if they were to override it they would previously have gotten a runtime linkage error and with this PR they will get a compile-time error.

None

#### Desired Merge Strategy

   - Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -1254,7 +1254,7 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
   /*
    * This method will be overwritten by the Chisel-Plugin
    */
-  protected def _elementsImpl: SeqMap[String, Any] = {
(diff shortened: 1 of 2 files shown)


chick (approved):
That looks pretty simple and clean.
Approved

PR #2396, opened 2022-02-03
Fix bundle elements performance regression

jackkoenig:
I previously measured that https://github.com/chipsalliance/chisel3/pull/2306 improved the performance of Chisel3 on a typical large SiFive design by 16%, more careful measurements show that number to actually be **30%**.

**However,** this benefit only comes if the user uses the scalacOption `-P:chiselplugin:genBundleElements` which tells the Chisel compiler plugin to generate `Bundle.elements`. I then tried this typical large SiFive design without that scalacOption and I found Chisel to run 21% **slower**. Considering most users will bump to 3.5.1 and won't apply that new scalacOption immediately (if they ever do), this slowdown is unacceptable.

This PR changes 2 small things:
1. Only call _elementsImpl once in Bundle.elements -- this mitigates most of the slowdown but not all of it. It also speed up the case when folks are using `-P:chiselplugin:genBundleElements` (although probably a very small amount). 
2. Make the reflective `_elementsImpl` use the exact old implementation of `Bundle.elements`, `Bundle.elements` then matches on the output of `_elementsImpl` and just returns immediately when its the reflective implementation.

This appears to fully mitigate the slowdown.

I tested this on SiFive designs as well as by disabling Bundle.elements generation in the chisel3 tests and running `sbt test`.

### Contributor Checklist

No release notes necessary because this feature is not yet released.

#### Type of Improvement

 - bug fix       
 - performance improvement 

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -1208,7 +1208,9 @@ abstract class Bundle(implicit compileOptions: CompileOptions) extends Record {
     * }}}
     */
   final lazy val elements: SeqMap[String, Data] = {
-    val hardwareFields = _elementsImpl.flatMap {
+    // _elementsImpl is a method, only call it once
+    val elementsImpl = _elementsImpl
+    val hardwareFields = elementsImpl.flatMap {
       case (name, data: Data) =>
         if (data.isSynthesizable) {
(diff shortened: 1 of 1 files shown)


chick (approved):
I like the solution

PR #2407, opened 2022-02-10
Add scanLeftOr and scanRightOr utilies

CircuitCoder:
### Contributor Checklist

#### Type of Improvement
new feature/API

#### API Impact
Add utility `LSBOr` and `MSBOr` similar to rocket-chip's `leftOr` and `rightOr`. These functions are generally very useful and can be used to implementing other useful utilities such as bitfield arbitration

N/A

#### Desired Merge Strategy
Rebase

#### Release Notes
Add experimental utility `LSBOr` and `MSBOr`.

- [ ] Did you review?

### DIFF
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
@@ -2,13 +2,12 @@
 
 import chisel3._
 import chisel3.util._
-import chisel3.experimental.util.algorithm._
 import chiseltest._
 import chiseltest.formal._
 import org.scalatest.flatspec.AnyFlatSpec
 import scala.math.min
 
-class scanLeftOrTestModule(width: Int) extends Module {
+class ScanLeftOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
 
   var lsb = false.B
@@ -24,7 +23,7 @@ class scanLeftOrTestModule(width: Int) extends Module {
   assert(testee === ref)
 }
 
-class scanRightOrTestModule(width: Int) extends Module {
+class ScanRightOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
 
   val ref = Reverse(scanLeftOr(Reverse(input)))
@@ -34,15 +33,15 @@ class scanRightOrTestModule(width: Int) extends Module {
 }
 
 class scanOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
-  "scanLeftOr" should "correctly computes" in {
+  "scanLeftOr" should "compute correctly" in {
     for(i <- 1 to 16) {
-      verify(new scanLeftOrTestModule(i), Seq(BoundedCheck(1)))
+      verify(new ScanLeftOrTestModule(i), Seq(BoundedCheck(1)))
     }
   }
 
-  "scanRightOr" should "correctly computes" in {
+  "scanRightOr" should "compute correctly" in {
     for(i <- 1 to 16) {
-      verify(new scanRightOrTestModule(i), Seq(BoundedCheck(1)))
+      verify(new ScanRightOrTestModule(i), Seq(BoundedCheck(1)))
     }
   }
 }
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
@@ -8,28 +8,7 @@ import chiseltest.formal._
 import org.scalatest.flatspec.AnyFlatSpec
 import scala.math.min
 
-// Copied from rocket-core
-object RocketImpl {
-  // Fill 1s from low bits to high bits
-  def leftOR(x: UInt): UInt = leftOR(x, x.getWidth, x.getWidth)
-  def leftOR(x: UInt, width: Integer, cap: Integer = 999999): UInt = {
-    val stop = min(width, cap)
-    def helper(s: Int, x: UInt): UInt =
-      if (s >= stop) x else helper(s+s, x | (x << s)(width-1,0))
-    helper(1, x)(width-1, 0)
-  }
-
-  // Fill 1s form high bits to low bits
-  def rightOR(x: UInt): UInt = rightOR(x, x.getWidth, x.getWidth)
-  def rightOR(x: UInt, width: Integer, cap: Integer = 999999): UInt = {
-    val stop = min(width, cap)
-    def helper(s: Int, x: UInt): UInt =
-      if (s >= stop) x else helper(s+s, x | (x >> s))
-    helper(1, x)(width-1, 0)
-  }
-}
-
-class LSBOrTestModule(width: Int) extends Module {
+class scanLeftOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
 
   var lsb = false.B
@@ -39,35 +18,31 @@ class LSBOrTestModule(width: Int) extends Module {
     cur
   }
   val ref = VecInit(vec).asUInt
-  val rocketRef = RocketImpl.leftOR(input)
 
-  val testee = LSBOr(input)
+  val testee = scanLeftOr(input)
 
   assert(testee === ref)
-  assert(testee === rocketRef)
 }
 
-class MSBOrTestModule(width: Int) extends Module {
+class scanRightOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
 
-  val ref = Reverse(LSBOr(Reverse(input)))
-  val rocketRef = RocketImpl.rightOR(input)
-  val testee = MSBOr(input)
+  val ref = Reverse(scanLeftOr(Reverse(input)))
+  val testee = scanRightOr(input)
 
   assert(testee === ref)
-  assert(testee === rocketRef)
 }
 
-class LSBMSBOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
-  "LSBOr" should "correctly computes" in {
+class scanOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
+  "scanLeftOr" should "correctly computes" in {
     for(i <- 1 to 16) {
-      verify(new LSBOrTestModule(i), Seq(BoundedCheck(1)))
+      verify(new scanLeftOrTestModule(i), Seq(BoundedCheck(1)))
     }
   }
 
-  "MSBOr" should "correctly computes" in {
+  "scanRightOr" should "correctly computes" in {
     for(i <- 1 to 16) {
-      verify(new MSBOrTestModule(i), Seq(BoundedCheck(1)))
+      verify(new scanRightOrTestModule(i), Seq(BoundedCheck(1)))
     }
   }
 }
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
@@ -1,0 +1,45 @@
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.util.algorithm._
+import chiseltest._
+import chiseltest.formal._
+import org.scalatest.flatspec.AnyFlatSpec
+
+class LSBOrTestModule(width: Int) extends Module {
+  val input = IO(Input(UInt(width.W)))
+
+  var lsb = false.B
+  val vec = for(b <- input.asBools) yield {
+    val cur = b || lsb
+    lsb = cur
+    cur
+  }
+  val ref = VecInit(vec).asUInt
+
+  val testee = LSBOr(input)
+
+  assert(testee === ref)
+}
+
+class MSBOrTestModule(width: Int) extends Module {
+  val input = IO(Input(UInt(width.W)))
+
+  val ref = Reverse(LSBOr(Reverse(input)))
+  val testee = MSBOr(input)
+
+  assert(testee === ref)
+}
+
+class LSBMSBOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
+  "LSBOr" should "correctly computes" in {
+    for(i <- 1 to 16) {
+      verify(new LSBOrTestModule(i), Seq(BoundedCheck(1)))
+    }
+  }
+
+  "MSBOr" should "correctly computes" in {
+    for(i <- 1 to 16) {
+      verify(new MSBOrTestModule(i), Seq(BoundedCheck(1)))
+    }
+  }
+}
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 import chisel3._
 import chisel3.util._
 import chisel3.experimental.util.algorithm._
--- a/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
+++ b/integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
@@ -6,6 +6,28 @@ import chisel3.experimental.util.algorithm._
 import chiseltest._
 import chiseltest.formal._
 import org.scalatest.flatspec.AnyFlatSpec
+import scala.math.min
+
+// Copied from rocket-core
+object RocketImpl {
+  // Fill 1s from low bits to high bits
+  def leftOR(x: UInt): UInt = leftOR(x, x.getWidth, x.getWidth)
+  def leftOR(x: UInt, width: Integer, cap: Integer = 999999): UInt = {
+    val stop = min(width, cap)
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= stop) x else helper(s+s, x | (x << s)(width-1,0))
+    helper(1, x)(width-1, 0)
+  }
+
+  // Fill 1s form high bits to low bits
+  def rightOR(x: UInt): UInt = rightOR(x, x.getWidth, x.getWidth)
+  def rightOR(x: UInt, width: Integer, cap: Integer = 999999): UInt = {
+    val stop = min(width, cap)
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= stop) x else helper(s+s, x | (x >> s))
+    helper(1, x)(width-1, 0)
+  }
+}
 
 class LSBOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
@@ -17,19 +39,23 @@ class LSBOrTestModule(width: Int) extends Module {
     cur
   }
   val ref = VecInit(vec).asUInt
+  val rocketRef = RocketImpl.leftOR(input)
 
   val testee = LSBOr(input)
 
   assert(testee === ref)
+  assert(testee === rocketRef)
 }
 
 class MSBOrTestModule(width: Int) extends Module {
   val input = IO(Input(UInt(width.W)))
 
   val ref = Reverse(LSBOr(Reverse(input)))
+  val rocketRef = RocketImpl.rightOR(input)
   val testee = MSBOr(input)
 
   assert(testee === ref)
+  assert(testee === rocketRef)
 }
 
 class LSBMSBOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
--- a/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
+++ b/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
@@ -25,6 +25,7 @@ object scanLeftOr {
     helper(1, data)(width - 1, 0)
   }
 }
+/** Map each bits to logical or of itself and all bits with higher index. 
   * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
   * This is consistent with the `right` as in "shift right" performed on bits, which means "start at the left (most significant bit) and move to the right".
   * @example {{{
--- a/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
+++ b/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
@@ -4,42 +4,43 @@ package chisel3.experimental.util.algorithm
 
 import chisel3._
 
-/** Map each bits to logical or of itself and all bits less siginificant than it.
+/** Map each bits to logical or of itself and all bits with lower index.
+  * Here 'left' means 'with lower index', as in arrays and lists, not to be confused with the 'left' as in 'shift left'
   * @example {{{
-  * LSBOr("b00001000".U) // Returns "b11111000".U
-  * LSBOr("b00010100".U) // Returns "b11111100".U
-  * LSBOr("b00000000".U) // Returns "b00000000".U
+  * scanLeftOr("b00001000".U) // Returns "b11111000".U
+  * scanLeftOr("b00010100".U) // Returns "b11111100".U
+  * scanLeftOr("b00000000".U) // Returns "b00000000".U
   * }}}
-  * This circuit seems to be high fan out, but synthesis tool should handle this.
   */
-object LSBOr {
+object scanLeftOr {
   def apply(data: UInt): UInt = {
     val width = data.widthOption match {
       case Some(w) => w
-      case None    => throw new IllegalArgumentException("Cannot call LSBOr on data with unknown width.")
+      case None    => throw new IllegalArgumentException("Cannot call scanLeftOr on data with unknown width.")
     }
-    VecInit(Seq.tabulate(width) { i: Int =>
-      VecInit(data.asBools().dropRight(width - i - 1)).asUInt().orR()
-    }).asUInt()
+
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= width) x else helper(s + s, x | (x << s)(width - 1, 0))
+    helper(1, data)(width - 1, 0)
   }
 }
 
-/** Map each bits to logical or of itself and all bits more siginificant than it.
+/** Map each bits to logical or of itself and all bits with higher index.
+  * Here 'right' means 'with higher index', as in arrays and lists, not to be confused with the 'right' as in 'shift right'
   * @example {{{
-  * MSBOr("b00001000".U) // Returns "b00001111".U
-  * MSBOr("b00010100".U) // Returns "b00011111".U
-  * MSBOr("b00000000".U) // Returns "b00000000".U
+  * scanRightOr("b00001000".U) // Returns "b00001111".U
+  * scanRightOr("b00010100".U) // Returns "b00011111".U
+  * scanRightOr("b00000000".U) // Returns "b00000000".U
   * }}}
-  * This circuit seems to be high fan out, but synthesis tool should handle this.
   */
-object MSBOr {
+object scanRightOr {
   def apply(data: UInt): UInt = {
     val width = data.widthOption match {
       case Some(w) => w
-      case None    => throw new IllegalArgumentException("Cannot call MSBOr on data with unknown width.")
+      case None    => throw new IllegalArgumentException("Cannot call scanRightOr on data with unknown width.")
     }
-    VecInit(Seq.tabulate(width) { i: Int =>
-      VecInit(data.asBools().drop(i)).asUInt().orR()
-    }).asUInt()
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= width) x else helper(s + s, x | (x >> s))
+    helper(1, data)(width - 1, 0)
   }
 }
--- a/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
+++ b/src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
-package chisel3.experimental.util.algorithm
+package chisel3.util
 
 import chisel3._
 
-/** Map each bits to logical or of itself and all bits with lower index.
-  * Here `scanLeft` means "start at the left and look to the right, where left is the lowest index", a common operation on arrays and lists.
-  * This is consistent with the `left` as in "shift left" performed on bits, which means "start at the right (least significant bit) and move to the left".
+/** Map each bit to the logical OR of itself and all bits with lower index
+  *
+  * Here `scanLeft` means "start from the left and iterate to the right, where left is the lowest index", a common operation on arrays and lists.
   * @example {{{
   * scanLeftOr("b00001000".U(8.W)) // Returns "b11111000".U
   * scanLeftOr("b00010100".U(8.W)) // Returns "b11111100".U
@@ -25,9 +25,10 @@ object scanLeftOr {
     helper(1, data)(width - 1, 0)
   }
 }
-/** Map each bits to logical or of itself and all bits with higher index. 
-  * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
-  * This is consistent with the `right` as in "shift right" performed on bits, which means "start at the left (most significant bit) and move to the right".
+
+/** Map each bit to the logical OR of itself and all bits with higher index
+  *
+  * Here `scanRight` means "start from the right and iterate to the left, where right is the highest index", a common operation on arrays and lists.
   * @example {{{
   * scanRightOr("b00001000".U) // Returns "b00001111".U
   * scanRightOr("b00010100".U) // Returns "b00011111".U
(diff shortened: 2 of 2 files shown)


integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
+import chisel3._
```
sequencer, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0

import chisel3._
```

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
...
+
+class LSBOrTestModule(width: Int) extends Module {
+  val input = IO(Input(UInt(width.W)))
+
+  var lsb = false.B
+  val vec = for(b <- input.asBools) yield {
+    val cur = b || lsb
+    lsb = cur
+    cur
+  }
```
sequencer, line 16: This is neat.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/** Map each bits to logical or of itself and all bits less siginificant than it.
+  * @example {{{
+  * LSBOr("b00001000".U) // Returns "b11111000".U
+  * LSBOr("b00010100".U) // Returns "b11111100".U
+  * LSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  */
+object LSBOr {
+  def apply(data: UInt): UInt = VecInit(Seq.tabulate(data.getWidth) { i: Int =>
```
sequencer, line 13: use `widthOption` to catch and throw an `Exception` when `!data.widthKnown`.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/** Map each bits to logical or of itself and all bits more siginificant than it.
+  * @example {{{
+  * MSBOr("b00001000".U) // Returns "b00001111".U
+  * MSBOr("b00010100".U) // Returns "b00011111".U
+  * MSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  */
+object MSBOr {
+  def apply(data: UInt): UInt = VecInit(Seq.tabulate(data.getWidth) { i: Int =>
```
sequencer, line 26: use `widthOption` to catch and throw an `Exception` when `!data.widthKnown`.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,29 @@
+package chisel3.experimental.util.algorithm
```
sequencer, line 1: I think putting these kind of hardware algorithm to `algorithm` is a good idea.
 
```suggestion
// SPDX-License-Identifier: Apache-2.0

package chisel3.experimental.util.algorithm
```

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import chisel3._
+
+/** Map each bits to logical or of itself and all bits less siginificant than it.
+  * @example {{{
+  * LSBOr("b00001000".U) // Returns "b11111000".U
+  * LSBOr("b00010100".U) // Returns "b11111100".U
+  * LSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  */
```
sequencer, line 11: Good doc,
I remember I had an issue chipsalliance/rocket-chip#2616 years ago discussing this, this circuit seems to be high fan-out, but it can be optimized by the synthesis tool. Maybe adding a note here to users for the fanout issue?
```suggestion
/** Map each bits to logical or of itself and all bits less siginificant than it.
  * @example {{{
  * LSBOr("b00001000".U) // Returns "b11111000".U
  * LSBOr("b00010100".U) // Returns "b11111100".U
  * LSBOr("b00000000".U) // Returns "b00000000".U
  * }}}
  * This circuit seems to be high fan out, but synthesis tool should handle this.
  */
```

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,29 @@
...
+  }).asUInt()
+}
+
+/** Map each bits to logical or of itself and all bits more siginificant than it.
+  * @example {{{
+  * MSBOr("b00001000".U) // Returns "b00001111".U
+  * MSBOr("b00010100".U) // Returns "b00011111".U
+  * MSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  */
```
sequencer, line 24:
```suggestion
/** Map each bits to logical or of itself and all bits more siginificant than it.
  * @example {{{
  * MSBOr("b00001000".U) // Returns "b00001111".U
  * MSBOr("b00010100".U) // Returns "b00011111".U
  * MSBOr("b00000000".U) // Returns "b00000000".U
  * }}}
  * This circuit seems to be high fan out, but synthesis tool should handle this.
  */
```

sequencer (commented):
Generally LGTM, name of this: `LSBOr` or `lsbOr` or `LsbOr` should have an alignment to chisel standard library.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
CircuitCoder: Width check is added in the newest commit.

sequencer (approved):
LGTM.

sequencer:
I think we need discuss the package name on dev-meeting, hold it till next Monday

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,47 @@
...
+      verify(new LSBOrTestModule(i), Seq(BoundedCheck(1)))
+    }
+  }
+
+  "MSBOr" should "correctly computes" in {
+    for(i <- 1 to 16) {
+      verify(new MSBOrTestModule(i), Seq(BoundedCheck(1)))
+    }
+  }
+}
```
sequencer, line 47: From the deb meeting discussion:
A formal checking between RC utils and this test should exist.
We suggest we need to add those API to test and formal check logic equivalence.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
...
+
+/** Map each bits to logical or of itself and all bits less siginificant than it.
+  * @example {{{
+  * LSBOr("b00001000".U) // Returns "b11111000".U
+  * LSBOr("b00010100".U) // Returns "b11111100".U
+  * LSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  * This circuit seems to be high fan out, but synthesis tool should handle this.
+  */
+object LSBOr {
```
sequencer, line 15: It should be renamed to `leftOr` to coordinate to RC.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
...
+
+/** Map each bits to logical or of itself and all bits more siginificant than it.
+  * @example {{{
+  * MSBOr("b00001000".U) // Returns "b00001111".U
+  * MSBOr("b00010100".U) // Returns "b00011111".U
+  * MSBOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  * This circuit seems to be high fan out, but synthesis tool should handle this.
+  */
+object MSBOr {
```
sequencer, line 35: It should be renamed to `rightOr` to coordinate to RC.

sequencer (changes requested):
feedback form dev-meeting.

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
CircuitCoder: Done in 677fa13a

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
CircuitCoder: IMO the naming leftOr has some discrepancy with "left" in "shift left". `leftOr` is actually or-ing all bits in the right (LSB). This can be a source of confusion. Actually I got it wrong when adding cross-check with rocket-core's implementation.

BTW, if we really want to keep in line with RC's naming, should we use `leftOR` instead of `leftOr`?

sequencer: > BTW, if we really want to keep in line with RC's naming, should we use leftOR instead of leftOr?

yeah, Chisel use `Or` everywhere, so `OR` is an anti-pattern in Chisel(but not in RC).

> `leftOr` is actually or-ing all bits in the right (LSB). This can be a source of confusion.

I agree, that is confusing as well(actually every time I use this function I need to take a look at its implementation). Personally I like the `lsbOr` better, which means `OR` LSB data of each bits. I think for the naming part, we may not need to align with RC.
what's opinion from @jackkoenig and @ekiwi? I'm OK at `lsbOr` and `msbOr` for this function.

ekiwi: My suggestion here is: make a PR to rocket-chip to change the name of their functions from `leftOR` to `lsbOr` and from `rightOR` to `msbOr`. As soon as that PR is approved, we can then go ahead and add the function to chisel3. If the rocket-chip folks do not approve of changing the name, that means that suddenly there are two copies of the same function, one in rocket-chip and one in chisel.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
...
+object LSBOr {
+  def apply(data: UInt): UInt = {
+    val width = data.widthOption match {
+      case Some(w) => w
+      case None    => throw new IllegalArgumentException("Cannot call LSBOr on data with unknown width.")
+    }
+    VecInit(Seq.tabulate(width) { i: Int =>
+      VecInit(data.asBools().dropRight(width - i - 1)).asUInt().orR()
+    }).asUInt()
+  }
```
jackkoenig, line 24: Things to consider about the implementation. @terpstra noted that the version in rocket-chip is optimized for simulation speed (minimizing the amount of generated Verilog). This is obviously very sensitive to both the number of FIRRTL nodes generated and to the optimizations in the FIRRTL implementation (maybe CIRCT can/will do something different here).

Another thing to consider (h/t @sequencer) is how well things will constant propagate if you were to call this then just index the MSB. It would be ideal if that were optimized but is not clear if that's reasonable to expect a FIRRTL compiler to do.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: h/t @seldridge for some archeology, the original name of this was `highOR`, then changed to `leftOR` when `rightOR` was added. Food for thought.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
sequencer: There was a common usage: "find first one" using this API. However can simulator detect this pattern or not is unclear, I think optimization pass to detect this common circuit should happen in ventilator or VCS passes.
Node for this compact mode is about N(indexing) + N(orR) ~2N, generally amd64 backend can leverage more CISC benefits here.
Essentially, this is the prefix tree for each bit, a prefix graph for entire circuit. node for prefix tree depends on the the tree implementation, but mostly higher than 2N, if my calculation is correct.

Maybe implementing a prefix tree in Chisel sounds a final idea? I'm not so sure about this. How to design API with a prefix graph.(maybe an implicit class? not so sure)

ekiwi: What would we loose by just copying over the rocket-chip solution? That seems to be by far the simplest solution. Are there actually noticable QoR differences?

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: From dev today, our current favorite seems to be `prefixOrFromLsb` (and `prefixOrFromMsb` for the other one).

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: From dev today, I think we should stick with the rocket-chip implementation (even if we don't expose the `width` arguments). Looking at how many nodes it creates, this implementation is `O(n^2)` vs. `O(log(n))` for the rocket-chip implementation.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: I chatted about this a bit with @terpstra. He argued that `fromLsb` is clear, although arguably `LsbFirst` is another alternative to consider. That being said, he argued (and I agree) that `left` and `right` are actually unambiguous. It is no different than a `foldLeft`, `reduceLeft`, whatever. It is a bit unfortunately that number like `54321` decomposes to `List(1, 2, 3, 4, 5)`, but this is universal. Iterate on the bits of a bitvector in Verilog and you get the exact same thing.

All of this is to say, I now think that `prefixOrLeft` and `prefixOrRight` are perfectly acceptable options. I think they are a lot better than `leftOr` because, (1) it matches the style of `foldLeft`, and (2) `prefixOr` is much more clearly the operation than just an `Or`. With that in mind, I'll accept `prefixOrFromLsb`, but I actually prefer `prefixOrLeft` now.

ekiwi: :+1:  on `prefixOrLeft`

seldridge: Note: `prefixOr` is a made-up name that doesn't appear to have prior art.  The line of reasoning here appears to be this looks like a `prefixSum` and since sum is `xor` in GF(2), then it's reasonable to call this `prefixXor`.  If we use a different reduction, then it's reasonable to call this `prefix<binary-operator>`.  I would prefer to not invent our own thing here and stick with "scan" which Wikipedia (without any citations) calls the generalization of `prefixSum` to other binary operators.

seldridge: 😅 And concretely, I'm proposing `scanLeftOr`.

jackkoenig: My argument in favor of using `Left` relies on the universal naming convention. This makes my arguing for `prefixOr` is a bit inconsistent since we made that one up. `scanLeftOr` is probably the most "conventionally consistent name), since a reasonable implementation of this behavior uses it:
ie. 
```scala
data.asBools.scanLeft(false.B)(_ || _).tail.asUInt
```
This is really just a special scanLeft on the bits of a `UInt`...

aswaterman: I think calling it a `scan` is appropriate.  (It's less immediately obvious what `prefixOr` means, and it kind of sounds like it's going to increase the width of the input by prefixing something to it.)

Although I agree there's a case to be made for `left` and `right`, it's inevitably going to confuse people, because we are defining this on UInt, and the leftmost part of a UInt, when written out, is the MSB.

jackkoenig: @aswaterman had an interesting suggestion, what about `Up` and `Down`? While I think the convention of `Left` and `Right` is unambiguous on sequences, and I am comfortable extending the convention to bit vectors like `UInt`, I think this is an interesting idea.

So, yet another proposal is the convention of "Up" to mean from least-significant bit from most-significant bit, and "Down" to mean the opposite. Thus the name of this function could be: `scanUpOr`.

My main concern about "Up" and "Down" is that it would be inconsistent with the methods on `Vec` which I really do not think should deviate from the normal functional convention given that Vecs are Scala Seqs. There would be a little bit of extra convention learning that `scanUp` is equivalent to `.asBools.scanLeft`, but I think that's reasonable if there's general consensus that "Up"/"Down" is much more clear than "Left"/"Right" in the context of bit vectors like `UInt`.

jackkoenig (changes requested):
Okay, done bike shedding. We should
1. Name them `scanLeftOr` and `scanRightOr` to use consistent terminology with sequence traversals. 
2. As far as the implementation, we should use the original from rocket-chip which is [`O(lg(n))`](https://scastie.scala-lang.org/xMfhgUawRzKv3qu7uk5m4w) in node count vs. [`O(n^2)`](https://scastie.scala-lang.org/mV6xbfFrTxKzqTKBUj3UkA) for this current implementation (note my implementation using `scanLeft` is [`O(n)`](https://scastie.scala-lang.org/6KrzfWN8Rf6mR04RhtRXYg)

Per QoR concerns, if someone provides compelling physical design results we can revisit. That being said, SiFive has been building high performance designs using the rocket-chip utility so I doubt there's an issue in proprietary ASIC tools but I will grant that there may be problems using open-source.

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,73 @@
...
+import chisel3.util._
+import chisel3.experimental.util.algorithm._
+import chiseltest._
+import chiseltest.formal._
+import org.scalatest.flatspec.AnyFlatSpec
+import scala.math.min
+
+// Copied from rocket-core
+object RocketImpl {
+  // Fill 1s from low bits to high bits
```
mwachs5, line 13: Can we use scaladoc comments so that these are well documented in IDEs?

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.util.algorithm._
+import chiseltest._
+import chiseltest.formal._
+import org.scalatest.flatspec.AnyFlatSpec
+import scala.math.min
+
+// Copied from rocket-core
```
mwachs5, line 11: what is `rocket-core`? Can you provide a better link, and are there any license implications to "copied from"?

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
jackkoenig: This is a test, no ScalaDoc needed

mwachs5: oh nvm this is in the tests so probably not as important

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental.util.algorithm
+
+import chisel3._
+
+/** Map each bits to logical or of itself and all bits less siginificant than it.
```
mwachs5, line 7:
```suggestion
/** Map each bits to logical OR of itself and all bits less significant than it.
```

mwachs5:
Ok coming in just at the end of this, I'm confused by:

> Name them scanLeftOr and scanRightOr to use consistent terminology with sequence traversals.

So is the "Left" relative to a UInt written out in binary: `ob1010` the "leftmost" bit is a `1` or in the `0` in the to-bools vec version of it Seq(0, 1, 0, 1)? I don't really care and think it's fine as long as we document it very heavily.

jackkoenig:
> So is the "Left" relative to a UInt w/ value 0xA written out in binary: b1010 the "leftmost" bit is a 1 or in the 0 in the to-bools vec version of it Seq(0, 1, 0, 1)? I don't really care and think it's fine as long as we document it very heavily.
>
> Reading the above I think we're saying it's the second one.

"Left" means low index to high index in sequences (in every language with traversals, think `foldLeft`), so the analogy is from low-order bit to high-order bit (think about what the 0th index is on `b1010`). Waterman's suggestion of Up and Down is interesting and I kind of like it, but there's an argument of having additional terminology to learn rather than just learning Left/Right. If we add Up/Down for use on UInts, it suggests we should have consistent terminology on Vecs, but Vec extends Seq so via inheritance it is stuck with Left/Right terminology.

Thus my argument is to stick with Left/Right and just explain it. It's consistent and we're not making up new terminology.

> Can we just make it you can't actually call this on a UInt, only on a Vec[Bool]? Then the confusion basically goes away. 

No, the whole purpose of this API is that it's useful on `UInt` and can be done using word-level logic that is more efficient than doing it on a `Vec`, see the table I made in https://github.com/chipsalliance/chisel3/pull/2407#pullrequestreview-897992375. Word-level is `O(lg(n))` lines of chirrtl/verilog while the best bit-level implementation is `O(n)` where n is the width of your bitvector.

mwachs5:
> Thus my argument is to stick with Left/Right and just explain it. It's consistent and we're not making up new terminology.

SGTM. Let me know when I can review the scaladoc :)

CircuitCoder:
> "Left" means low index to high index in sequences

Actually that's a great way to explain the usage or "left" and "right" here. I've added it to the documentation.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
CircuitCoder: The renaming is done in e6fab506430c1bce4ce5e0de64cba8e08658d3f6

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
CircuitCoder: The implementation has been changed to stay in line with rocket-chip in e6fab506430c1bce4ce5e0de64cba8e08658d3f6.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+/** Map each bits to logical or of itself and all bits with higher index.
+  * Here 'right' means 'with higher index', as in arrays and lists, not to be confused with the 'right' as in 'shift right'
+  * scanRightOr("b00001000".U) // Returns "b00001111".U
+  * scanRightOr("b00010100".U) // Returns "b00011111".U
+  * scanRightOr("b00000000".U) // Returns "b00000000".U
+  * }}}
+  * This circuit seems to be high fan out, but synthesis tool should handle this.
```
sequencer, line 34: synthesis tool cannot handle this now.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
...
+  */
+object scanRightOr {
+  def apply(data: UInt): UInt = {
+    val width = data.widthOption match {
+      case Some(w) => w
+      case None    => throw new IllegalArgumentException("Cannot call scanRightOr on data with unknown width.")
+    }
+    def helper(s: Int, x: UInt): UInt =
```
sequencer, line 42: make it private.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
...
+  * }}}
+  */
+object scanLeftOr {
+  def apply(data: UInt): UInt = {
+    val width = data.widthOption match {
+      case Some(w) => w
+      case None    => throw new IllegalArgumentException("Cannot call scanLeftOr on data with unknown width.")
+    }
+
+    def helper(s: Int, x: UInt): UInt =
```
sequencer, line 22: make it private

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
CircuitCoder: Oh that's a comment for the earlier impl. I just forgot to remove it.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
...
+      if (s >= width) x else helper(s + s, x | (x << s)(width - 1, 0))
+    helper(1, data)(width - 1, 0)
+  }
+}
+
+/** Map each bits to logical or of itself and all bits with higher index.
+  * Here 'right' means 'with higher index', as in arrays and lists, not to be confused with the 'right' as in 'shift right'
```
mwachs5, line 29: similar comment as above

sequencer, line 29:
```suggestion
  * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
  * This is consistent with the `right` as in "shift right" performed on bits, which means "start at the left (most significant bit) and move to the right".
```

sequencer:
I do think we need a CI to reformat on pulls:(

mwachs5 (commented):
can we update the title of this PR to match the new names?

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
...
+      case Some(w) => w
+      case None    => throw new IllegalArgumentException("Cannot call scanLeftOr on data with unknown width.")
+    }
+
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= width) x else helper(s + s, x | (x << s)(width - 1, 0))
+    helper(1, data)(width - 1, 0)
+  }
+}
+  * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
```
mwachs5, line 28:
```suggestion
  /** Map each bits to logical or of itself and all bits with higher index.
    *Here `scanRight` means "start from the right and look to the left, where right is the highest index", a common operation on arrays and lists.
```

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental.util.algorithm
+
+import chisel3._
+
+/** Map each bits to logical or of itself and all bits with lower index.
+  * Here `scanLeft` means "start at the left and look to the right, where left is the lowest index", a common operation on arrays and lists.
```
mwachs5, line 8:
```suggestion
  * Here `scanLeft` means "start from the left and look to the right, where left is the lowest index", a common operation on arrays and lists.
```

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig, line 28:
```suggestion
/** Map each bits to logical or of itself and all bits with higher index. 
  * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
```

jackkoenig: You beat me by a few seconds, but mine had a space between `*` and `Here` so I merged mine 🙂

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,48 @@
...
+  val input = IO(Input(UInt(width.W)))
+
+  val ref = Reverse(scanLeftOr(Reverse(input)))
+  val testee = scanRightOr(input)
+
+  assert(testee === ref)
+}
+
+class scanOrTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
+  "scanLeftOr" should "correctly computes" in {
```
mwachs5, line 37:
```suggestion
  "scanLeftOr" should "compute correctly" in {
```

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,48 @@
...
+    for(i <- 1 to 16) {
+      verify(new scanLeftOrTestModule(i), Seq(BoundedCheck(1)))
+    }
+  }
+
+  "scanRightOr" should "correctly computes" in {
```
mwachs5, line 43:
```suggestion
  "scanRightOr" should "compute correctly" in {
```

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.util.algorithm._
+import chiseltest._
+import chiseltest.formal._
+import org.scalatest.flatspec.AnyFlatSpec
+import scala.math.min
+
+class scanLeftOrTestModule(width: Int) extends Module {
```
mwachs5, line 11:
```suggestion
class ScanLeftOrTestModule(width: Int) extends Module {
```

integration-tests/src/test/scala/chiselTests/util/experimental/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,48 @@
...
+    cur
+  }
+  val ref = VecInit(vec).asUInt
+
+  val testee = scanLeftOr(input)
+
+  assert(testee === ref)
+}
+
+class scanRightOrTestModule(width: Int) extends Module {
```
mwachs5, line 27:
```suggestion
class ScanRightOrTestModule(width: Int) extends Module {
```

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental.util.algorithm
```
mwachs5, line 3: is there a reason this is `experimental`? And what is defining characteristic of what makes a `util` an `algorithm`?

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,47 @@
...
+      case None    => throw new IllegalArgumentException("Cannot call scanLeftOr on data with unknown width.")
+    }
+
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= width) x else helper(s + s, x | (x << s)(width - 1, 0))
+    helper(1, data)(width - 1, 0)
+  }
+}
+/** Map each bits to logical or of itself and all bits with higher index. 
+  * Here `scanRight` means "start at the right and look to the left, where right is the highest index", a common operation on arrays and lists.
```
mwachs5, line 29: but what do you think about "from the right" vs "at the right"

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: I'm confident enough that this is a useful and correct API to argue it doesn't need `experimental`. I also think `algorithm` is unnecessary but don't feel too strongly either way. Directly in `chisel3.util` might improve discoverability although we probably ought to just put some of these into a doc on the webpage.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
```diff
@@ -0,0 +1,47 @@
...
+  def apply(data: UInt): UInt = {
+    val width = data.widthOption match {
+      case Some(w) => w
+      case None    => throw new IllegalArgumentException("Cannot call scanRightOr on data with unknown width.")
+    }
+    def helper(s: Int, x: UInt): UInt =
+      if (s >= width) x else helper(s + s, x | (x >> s))
+    helper(1, data)(width - 1, 0)
+  }
+}
```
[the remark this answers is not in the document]
jackkoenig: I think we would make a cookbook for all of the utilities on UInt but I don't want to block this PR on it.

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
jackkoenig: I prefer `from`

src/main/scala/chisel3/experimental/util/algorithm/Bitwise.scala
sequencer: +1 on `util`

jackkoenig (approved):
LGTM. I deleted the comparison to "shift left" because I think it's more confusing than helpful. I would argue that one _could_ think about "shift left" as shifting from left to right where left is low-order bits, but no one actually does.

I also changed the package to `chisel3.util`. I'm confident enough in this API that there's no need for experimental.

jackkoenig:
Thank you @CircuitCoder for your contribution and perseverance at getting this very useful API merged!

PR #2412, opened 2022-02-13
[DRAFT] Contextual and Contexts

azidar:
This is a proposal for a new type, `Contextual`, which allows Chisel users to define values differently for instances of the same definition, but access them only outside the instance (thus not affecting hardware generation).

This is useful for attaching different metadata to different identical instances. In the examples, we demonstrate setting absolute indexes for instances, and setting a 'who's your neighbor' value. 

I am really happy with this general form of the implementation; this feature is incredibly powerful, conceptually consistent with D/I, and has pure abstractions.

### Contributor Checklist

- [ ] Did you review?


seldridge (commented):
One thing that seems off to me here is that `Definition` is introducing a level of indirection for contexts that seems superfluous.  In this PR: module declarations set default context, module definitions override default/remove/provide missing context, and instances are references to definitions with all context set.  Is this correct?

It would seem cleaner to remove `Definition` from this and have modules declared with default context and instances that set/override context.  I.e., can the context be moved onto the instance as opposed to on the module definition?

Zooming out: it feels like adding a `Contextual` on a `Module` is a way of describing a parametric FIRRTL module where the parameterization cannot affect the hardware description.  (Put differently, the choice of parameterization cannot affect the Chisel generator.)  `withContext` is then "per-instance argument passing".  The initial backing data store for this information is `Contextual` that stores values on a per-instance basis.  The eventual backing data store for these argument values is non-local FIRRTL annotations, but could be parametric FIRRTL IR as we evolve the FIRRTL specification.

This PR treats contextuals like ports.  They are defined like ports with a special type.  However, they are represented and used in a totally different way.  Is there a way to make the parameter passing more explicit and by-name?  E.g., something like:

```scala
@instaniable
class Bar extends Module {
  @public
  val p = Parameter[Int]
}

class Foo extends Module {
  val barDef = Definition(new Bar)
  val bar1 = Instance(barDef)
  val bar2 = Instance(barDef)
  bar1.p := 42
  bar2.p := 1337
}
```

What do you think?

seldridge:
Another example to think about is how does this work for parameters that should be represented in-IR.  This is relatively straightforward to start with for external modules because they already are kind-of parametric definitions.  Two uses cases I envision:
1. Passing a generator-time parameter around
2. Passing an IR parameter around

Start with an external module that has an integer parameter. (The following does not compile and uses some weird syntax that kind of matches how blackboxes work now...)

```scala
@instantiable
class Bar extends ExtModule(Map("intParam" -> IntParam)) {
  @public
  val in = IO(Input(Bool()))
  @public
  val out = IO(Output(Bool()))
}
```

This can then either be instantiated with the parameter bound at generator time:

```scala
class Foo(generatorParam: Int) extends RawModule {
  val in = IO(Input(Bool()))
  val out = IO(Output(Bool()))

  val barDef = Definition(new Bar)
  val bar = Instance(barDef, Seq(generatorParam))
  bar.in := in
  out := bar.out
}

ChiselStage.emitChirrtl(new Foo(42))
```

Or, the parameter is threaded through from the top module:

```scala
@instantiable
class Foo2 extends RawModule {
  @public
  val in = IO(Input(Bool()))
  @public
  val out = IO(Output(Bool()))
  @public
  val p = Paramater[Int]

  val barDef = Definition(new Bar)
  val bar = Instance(barDef, Seq(p))
  bar.in := in
  out := bar.out
}
```

mwachs5:
But @seldridge I think you are *maybe* missing a use case 3?
These are parameters that are specific to how a thing is used, not really a property of the thing itself. While this PR gives an abstraction to couple them, we could achieve the same thing by writing aspect-oriented code over the (sub)circuit and extracting whatever information that we need. Maybe that's what we should just do and abandon any sort of object-oriented, distribution-of-concerns into the units themselves?

To me this is really the concept to conserve, not so much anything about a parameter that the module ever needs to know...

seldridge:
> These are parameters that are specific to how a thing is used, not really a property of the thing itself.

The "how a thing is used", in Chisel at least, is equivalent to capturing some context information (a parameter) from the parent.  E.g., you can use this to communicate clock domain information or (as the PR shows) the initial `[x, y]` coordinates for placement of some region.

> While this PR gives an abstraction to couple them, we could achieve the same thing by writing aspect-oriented code over the (sub)circuit and extracting whatever information that we need. Maybe that's what we should just do and abandon any sort of object-oriented, distribution-of-concerns into the units themselves?

😅 Don't do that...

> To me this is really the concept to conserve, not so much anything about a parameter that the module ever needs to know...

My main worry is that we are actually dealing with a parameterization problem and I am concerned that this is building yet-another way to kind-of do parameterization.  (We really screwed up here with FIRRTL.  I thought annotations were amazing only to realize years later that they are a retrofitted solution to give parametric qualities to an un-parametric IR.  Failure to recognize the need for parameterization and that we were building a solution to a parameterization problem was extremely costly.)

That said, I recognize the need for an immediate solution here (and a solution that works for diplomatic graphs) and it would be ideal if whatever is used for D/I also works for Diplomacy.  I will be less nervous if we treat this as a tactical solution to fix immediate problems and we are free to strategically redesign this in the future.

src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -1044,4 +1047,196 @@ class InstanceSpec extends ChiselFunSpec with Utils {
...
+        }
+        val answers = Map(
+          ("~Top|Root/i0:Middle/i0:Leaf" -> Index(0)),
+          ("~Top|Root/i0:Middle/i1:Leaf" -> Index(1)),
+          ("~Top|Root/i1:Middle/i0:Leaf" -> Index(2)),
+          ("~Top|Root/i1:Middle/i1:Leaf" -> Index(3))
+        )
+        Select.allInstancesOf[Leaf](definition).foreach { case l: Instance[Leaf] =>
+          l.index should be (answers(l.toTarget.toString))
+        }
```
mwachs5, line 1084: As we discussed IRL, I am concerned about the fact that this doesn't really give a stable result if you run it twice. I think we talked about storing the stable result in an auxiliary data structure, but then how is this better/different than building an Aspect over this thing and putting the mapping from Instance -> Index into the auxiliary data structure?

mwachs5 (commented):
At a high level, I am wondering how far we can get if we say the only thing that is Contextual is yourself:

```
class Foo extends Module(p: SomeParam) with MustHaveContext

val someParam = 3
val  myself = Contextual[Foo]
val thing = s"This is a complicated String that uses ${myself}"

val thingYouCareAbout = (someParam, thing)
}
```

and any calls to Instance of Definition[MustHaveContext] has to provide / fill in the Contextual with that Instance I agree i do not know how to implement this Scala-wise, but
do we need to support arbitrary contextuals on anything *but* the Instance itself?

PR #2419, opened 2022-02-16
Cache all lookups, added tests

azidar:
### Contributor Checklist

#### Type of Improvement

 - bug fix

#### API Impact

No change to APIs, just fixes a bug where looking up the same value from the same Hierarchy[T] object returned a new object every time.

#### Release Notes

Bugfix: looking up the same value from the same Hierarchy[T] object now returns the same object, rather than a new object every time.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/hierarchy/Hierarchy.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/Hierarchy.scala
@@ -22,7 +22,10 @@ sealed trait Hierarchy[+A] {
   }
 
   /** Updated by calls to [[_lookup]], to avoid recloning returned Data's */
-  private[chisel3] val cache = HashMap[Data, Data]()
+  private val cache = HashMap[Any, Any]()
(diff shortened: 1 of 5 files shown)


core/src/main/scala/chisel3/experimental/hierarchy/Hierarchy.scala
```diff
@@ -21,8 +21,11 @@ sealed trait Hierarchy[+A] {
     case Clone(i: IsClone[A]) => i.getProto
   }
 
-  /** Updated by calls to [[_lookup]], to avoid recloning returned Data's */
-  private[chisel3] val cache = HashMap[Data, Data]()
+  /** Updated by calls to [[_lookup]], to avoid recloning returned values */
+  private val cache = HashMap[Any, Any]()
```
jackkoenig, line 25: I'm not sure if I can construct the case where it will matter, but my intuition is still that this should be an `IdentityHashMap`

PR #2434, opened 2022-03-07
std: add AddressSet from rocket-chip

ekiwi:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/stdlib/src/chisel3/std/AddressSet.scala
+++ b/stdlib/src/chisel3/std/AddressSet.scala
@@ -1,0 +1,134 @@
+package chisel3.std
+
+import chisel3._
+import chisel3.util.log2Floor
+
+/** AddressSets specify the address space managed by the manager
+  * Base is the base address, and mask are the bits consumed by the manager
+  * e.g: base=0x200, mask=0xff describes a device managing 0x200-0x2ff
+  * e.g: base=0x1000, mask=0xf0f describes a device managing 0x1000-0x100f, 0x1100-0x110f, ...
+  */
+case class AddressSet(base: BigInt, mask: BigInt) extends Ordered[AddressSet] {
+  // Forbid misaligned base address (and empty sets)
+  require((base & mask) == 0, s"Mis-aligned AddressSets are forbidden, got: ${this.toString}")
+  require(
+    base >= 0,
+    s"AddressSet negative base is ambiguous: $base"
+  ) // TL2 address widths are not fixed => negative is ambiguous
+  // We do allow negative mask (=> ignore all high bits)
+
+  def contains(x: BigInt): Boolean = ((x ^ base) & ~mask) == 0
+  def contains(x: UInt):   Bool = ((x ^ base.U).zext & (~mask).S) === 0.S
+
+  // turn x into an address contained in this set
+  def legalize(x: UInt): UInt = base.U | (mask.U & x)
+
+  // overlap iff bitwise: both care (~mask0 & ~mask1) => both equal (base0=base1)
+  def overlaps(x: AddressSet): Boolean = (~(mask | x.mask) & (base ^ x.base)) == 0
+  // contains iff bitwise: x.mask => mask && contains(x.base)
+  def contains(x: AddressSet): Boolean = ((x.mask | (base ^ x.base)) & ~mask) == 0
+
+  // The number of bytes to which the manager must be aligned
+  def alignment: BigInt = (mask + 1) & ~mask
+  // Is this a contiguous memory range
+  def contiguous: Boolean = alignment == mask + 1
+
+  def finite: Boolean = mask >= 0
+  def max: BigInt = { require(finite, "Max cannot be calculated on infinite mask"); base | mask }
+
+  // Widen the match function to ignore all bits in imask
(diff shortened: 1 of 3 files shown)


sequencer:
This has been implemented by `BitSet` API and `decoder.bitset` can handle this.

sequencer:
FYI: https://github.com/chipsalliance/chisel3/blob/master/src/test/scala/chiselTests/util/BitSetSpec.scala#L95-L117

sequencer:
I think we need a wrapper from bus library to `decoder.bitset`

ekiwi:
How would you translate `AddressSet(0x200, 0xff)` to the `BitSet` API? Like how do you specify the offset?

sequencer:
Basically `AddressSet` is a `BitSet` with `apply` function to convert address `offset` and `range`.
`AddressSet(0x200, 0xff)`->`BitSet(BitPat(0x200, 0xff))`

ekiwi:
> Basically `AddressSet` is a `BitSet` with `apply` function to convert address `offset` and `range`.

I see. So you can basically think of `value` as being and `offset`. Do you know if there is a good way to expose the mask to the user?

I am trying to see how `BitSet` could be used with the [AXI4 SRAM](https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/amba/axi4/SRAM.scala) from rocket chip. Things that are missing so far are:
- [`.mask`](https://github.com/chipsalliance/rocket-chip/blob/b36060ebe7bd821ea4b4d1dd2aa9e6d38e7d01b4/src/main/scala/diplomacy/SRAM.scala#L31)
- [`.contains(UInt): Bool`](https://github.com/chipsalliance/rocket-chip/blob/b36060ebe7bd821ea4b4d1dd2aa9e6d38e7d01b4/src/main/scala/amba/axi4/SRAM.scala#L67)

sequencer:
RC Address Decoder use a heuristic algorithm to solve the partition problem, however the PPA result is bad, from evaluation at my side, decoder API provides about 30% PPA improvement in some case, so I directly drop it.
I think what we need here is a factory to convert address and address range to `BitSet`.

ekiwi:
> Yes, that's the key issue, we need a address factory, to convert any `AddressSet` to `BitSet`.

But then we would still need `AddressSet`. I thought the whole point was to reduce all uses of `AddressSet` with `BitSet` in order to avoid duplicate functionality.

sequencer:
> > Yes, that's the key issue, we need a address factory, to convert any `AddressSet` to `BitSet`.
> 
> But then we would still need `AddressSet`. I thought the whole point was to reduce all uses of `AddressSet` with `BitSet` in order to avoid duplicate functionality.

Yes, I do think so, we need a new API and implementations for `AddressSet`.
`mask` is trivial.
`contains` should be implemented via decoder. I must forgot to implement that.
I'll add it to `BitSet`, and `AddressSet` can directly wrap that.

sequencer:
Just thought about the `contains`, which is trivial as well:
```
def contains(data: UInt): decoder.bitset(data, Seq(this))
```
`this` is the `BitSet` need to test on the contains.

stdlib/src/chisel3/std/AddressSet.scala
```diff
@@ -0,0 +1,134 @@
+package chisel3.std
+
+import chisel3._
+import chisel3.util.log2Floor
+
+/** AddressSets specify the address space managed by the manager
```
mwachs5, line 6: If this is getting upstreamed it needs solid scaladoc on each public function

sequencer:
Note: there is a big difference between `BitSet` and RC `AddressSet`:
RC `AddressSet` can be unknown width, as you can see from `val Everything = AddressSet(0, -1)` API, while chisel BitSet doesn't support.

There are a lot of useful APIs for interconnection library, some are covered by `BitSet`, while some are not:
- `def contains(x: BigInt): Boolean` can be replaced by `BitSet.cover`
- `def contains(x: UInt): Bool` is not implemented, but I think we can implement by `decoder.BitSet` with a more general API, which can test a hard signal can be multiple `AddressSet` covered or not, this will significantly improve the PPA.
- `def legalize(x: UInt): UInt` can be dropped.
- `def overlaps(x: AddressSet)` is `def overlap(that: BitSet)`
- `def contains(x: AddressSet)` is `def cover(that: BitSet)`
- `def alignment: BigInt` is a useful helper, but it should not exist in `BitSet`
- `def contiguous: Boolean` is a useful helper, but it should not exist in `BitSet`
- `def finite: Boolean` is dropped, `BitSet` must be known width.
- `def max: BigInt` is a useful helper, but it should not exist in `BitSet`
- `def widen(imask: BigInt)` IIRC, is only useful in some test, I think we should keep the API simple.
- `def intersect(x: AddressSet)` is `def intersect(that: BitSet): BitSet`
- `def subtract(x: AddressSet) ` is `def subtract(that: BitSet): BitSet`
- `def compare(x: AddressSet): Int` is useless, IIRC, is is for the ordering, which used in address decoder.
- `def toString`, we certainly need this!

most API in `object AddressSet` is designed for decoder, I think they can safely dropped.

PR #2441, opened 2022-03-09
[Work in Progress] AXI4 bundle definitions for the standard library

ekiwi:
Very much work in progress, just collecting some thought for now.

### Contributor Checklist

- [ ] Did you review?


jerryz123:
> should the name of the bundles end in IO (like in DecoupledIO, QueueIO etc.) or Bundle (as is common in the rocket chip code base) or something else?

I prefer the rocket-chip style, as the bundle doesn't have to be used as an IO, its a bit more intuitive. 

> should there be separate classes for the read and write address channels? Their fields are essentially the same.

I think these can inherit the same AXI4ABundle class.

> aw or addrWriteChannel or addressWriteChannel or addressWrite or addrWrite? (similar decision for the other channels)

Concise naming is easier to read and match with the spec. `AXI4AW/AXI4AR/AXI4W/AXI4R/AXI4B`. 

I think at a minimum, this needs enough functionality to replace the rocket-chip AXI4 implementations.

stdlib/src/main/scala/chisel3/std/axi4/Axi4IO.scala
```diff
@@ -0,0 +1,119 @@
...
+  /** Burst length. 1 to 16 for Axi3. 1 to 256 for Axi4 Incr burst. */
+  val length = UInt(8.W)
+
+  /** Size of the transfer. 2**size in bytes. */
+  val size = UInt(3.W)
+  val burst = Axi4Burst()
+  val lock = Axi4Lock()
+  val cache = new Axi4MemoryType
+  val prot = new Axi4AccessPermission
+  val qos = UInt(4.W)
```
jwright6323, line 80: Note that the spec calls for this to be 4 bits wide, but 3rd party implementations sometimes use fewer or more bits, so it may be useful to leave as a param

stdlib/src/main/scala/chisel3/std/axi4/Axi4IO.scala
```diff
@@ -0,0 +1,119 @@
...
+object Axi4Response extends ChiselEnum {
+  val Okay = Value("b00".U(2.W))
+  val ExclusiveOkay = Value("b01".U(2.W))
+  val ServerError = Value("b10".U(2.W))
+  val DecodeError = Value("b11".U(2.W))
+}
+
+class Axi4AddressChannel(params: Axi4IOParameters, userBits: Int = 0) extends Bundle {
+  require(userBits >= 0, "number of user bits must be non-negative")
+  val id = UInt(params.idBits.W)
```
jwright6323, line 68: Technically the read and write ID bits can be different from each other for a single port, so it may be beneficial to give them different params.

ekiwi: Thanks I will add that.

stdlib/src/main/scala/chisel3/std/axi4/Axi4IO.scala
ekiwi: I could make it a parameter or we could try to rely on the firrtl compiler to optimize the port width. Since QoS is always 4-bit according to the standard, what would you prefer?

ekiwi:
TODO from @jwright6323 : Turns out that the `Axi4MemoryType` Bundle needs to be different for the read vs. the write direction.

sequencer:
@ekiwi would you mind cherry-picking `chisel.std` definition in build systems(sbt mill) firstly, creating another PR, and #2518 can directly use that?

stdlib/src/main/scala/chisel3/std/axi4/Axi4IO.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  /** Size of the transfer. 2**size in bytes. */
+  val size = UInt(3.W)
+  val burst = Axi4Burst()
+  val lock = Axi4Lock()
+  val cache = new Axi4MemoryType
+  val prot = new Axi4AccessPermission
+  val qos = UInt(4.W)
+  private val regionWidth = if (params.enableRegionBits) { 4 }
+  else { 0 }
```
sequencer, line 82: nitpick on format ;p
```suggestion
  private val regionWidth = if (params.enableRegionBits) 4 else 0
```

stdlib/src/test/scala/chisel3/std/axi4/TestMem.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  val mem = SyncReadMem(memoryEntries, Vec(beatBytes, UInt(8.W)))
+
+  private def decodeAddress(address: UInt): (UInt, Bool) = {
+    val a = Cat(maskBits.zip((address >> log2Ceil(beatBytes)).asBools).filter(_._1).map(_._2).reverse)
+    val sel = ((address ^ baseAddress.U).zext & (~mask).S) === 0.S
+    (a, sel)
+  }
+  val (r_addr, r_sel0) = decodeAddress(io.readAddress.bits.address)
+  val (w_addr, w_sel0) = decodeAddress(io.writeAddress.bits.address)
```
sequencer, line 48: #2449 merged, I think we can have a try to `BitSet. fromRange` and `decoder.bitset` to provide a better PPA.

ekiwi: Yes, there are probably a lot of possible optimizations. Note though, that the aim of this module is only to serve as a testing ground for the bundle definitions. If we never use the bundle to generate HW, it will not be tested properly.

sequencer: Sure!

ekiwi: If we ever decide to actually offer the IPs in the library (and just in `test`) we will have to be concerned more about PPA.
But also, since this would be the standard library we would have to make sure not to optimize for one particular backend. Things should synthesize reasonably well for FPGA and various ASIC technologies.

jiegec:
FWIW, I have written a Bundle with DataView to convert AXI4 bundle (from rocket chip) to standard names:

```scala
// use standard names
class StandardAXI4Bundle(
    val addrBits: Int,
    val dataBits: Int,
    val idBits: Int
) extends Bundle {
  val AWREADY = Input(Bool())
  val AWVALID = Output(Bool())
  val AWID = Output(UInt(idBits.W))
  val AWADDR = Output(UInt(addrBits.W))
  val AWLEN = Output(UInt(8.W))
  val AWSIZE = Output(UInt(3.W))
  val AWBURST = Output(UInt(2.W))
  val AWLOCK = Output(UInt(1.W))
  val AWCACHE = Output(UInt(4.W))
  val AWPROT = Output(UInt(3.W))
  val AWQOS = Output(UInt(4.W))

  val WREADY = Input(Bool())
  val WVALID = Output(Bool())
  val WDATA = Output(UInt(dataBits.W))
  val WSTRB = Output(UInt((dataBits / 8).W))
  val WLAST = Output(Bool())

  val BREADY = Output(Bool())
  val BVALID = Input(Bool())
  val BID = Input(UInt(idBits.W))
  val BRESP = Input(UInt(2.W))

  val ARREADY = Input(Bool())
  val ARVALID = Output(Bool())
  val ARID = Output(UInt(idBits.W))
  val ARADDR = Output(UInt(addrBits.W))
  val ARLEN = Output(UInt(8.W))
  val ARSIZE = Output(UInt(3.W))
  val ARBURST = Output(UInt(2.W))
  val ARLOCK = Output(UInt(1.W))
  val ARCACHE = Output(UInt(4.W))
  val ARPROT = Output(UInt(3.W))
  val ARQOS = Output(UInt(4.W))

  val RREADY = Output(Bool())
  val RVALID = Input(Bool())
  val RID = Input(UInt(idBits.W))
  val RDATA = Input(UInt(dataBits.W))
  val RRESP = Input(UInt(2.W))
  val RLAST = Input(Bool())
}

object StandardAXI4Bundle {
  implicit val axiView = DataView[StandardAXI4Bundle, AXI4Bundle](
    vab =>
      new AXI4Bundle(
        AXI4BundleParameters(vab.addrBits, vab.dataBits, vab.idBits)
      ),
    // AW
    _.AWREADY -> _.aw.ready,
    _.AWVALID -> _.aw.valid,
    _.AWID -> _.aw.bits.id,
    _.AWADDR -> _.aw.bits.addr,
    _.AWLEN -> _.aw.bits.len,
    _.AWSIZE -> _.aw.bits.size,
    _.AWBURST -> _.aw.bits.burst,
    _.AWLOCK -> _.aw.bits.lock,
    _.AWCACHE -> _.aw.bits.cache,
    _.AWPROT -> _.aw.bits.prot,
    _.AWQOS -> _.aw.bits.qos,
    // W
    _.WREADY -> _.w.ready,
    _.WVALID -> _.w.valid,
    _.WDATA -> _.w.bits.data,
    _.WSTRB -> _.w.bits.strb,
    _.WLAST -> _.w.bits.last,
    // B
    _.BREADY -> _.b.ready,
    _.BVALID -> _.b.valid,
    _.BID -> _.b.bits.id,
    _.BRESP -> _.b.bits.resp,
    // AR
    _.ARREADY -> _.ar.ready,
    _.ARVALID -> _.ar.valid,
    _.ARID -> _.ar.bits.id,
    _.ARADDR -> _.ar.bits.addr,
    _.ARLEN -> _.ar.bits.len,
    _.ARSIZE -> _.ar.bits.size,
    _.ARBURST -> _.ar.bits.burst,
    _.ARLOCK -> _.ar.bits.lock,
    _.ARCACHE -> _.ar.bits.cache,
    _.ARPROT -> _.ar.bits.prot,
    _.ARQOS -> _.ar.bits.qos,
    // R
    _.RREADY -> _.r.ready,
    _.RVALID -> _.r.valid,
    _.RID -> _.r.bits.id,
    _.RDATA -> _.r.bits.data,
    _.RRESP -> _.r.bits.resp,
    _.RLAST -> _.r.bits.last
  )
  implicit val axiView2 = StandardAXI4Bundle.axiView.invert(ab =>
    new StandardAXI4Bundle(
      ab.params.addrBits,
      ab.params.dataBits,
      ab.params.idBits
    )
  )
}
```

I hope it helps!

ekiwi:
> FWIW, I wrote a Bundle with DataView to convert AXI4 bundle (from rocket chip) to standard names:

Cool. Is that code licensed under Apache 2.0 ?

jiegec:
> > FWIW, I wrote a Bundle with DataView to convert AXI4 bundle (from rocket chip) to standard names:
> 
> Cool. Is that code licensed under Apache 2.0 ?

Sure.

chiplet:
FWIW here's a library of AXI4-Lite types, some interconnect, and some peripherals:
https://gitlab.com/a-core/a-core_chisel/amba/-/tree/master/src/main/scala/axi4l

I agree that it would be a good idea to add the bundle definitions to `stdlib` since currently everyone seems to be writing their own types.

One thing to consider is how the interface clock and reset signals are defined. IIRC the spec names them `ACLK` and `ARESETn` respectively. The spec also states that the reset signal is asynchronous with active-low polarity, which is currently quite painful to implement in Chisel.

Should the bundle definitions for the standard library types aim for full spec compliance, or would it be more useful to take an opinionated approach when it comes to clocks / resets? Using Chisel's implicit clock and reset would simplify implementations but it would also make it more difficult to integrate existing implementations written in e.g. verilog/VHDL that are fully spec compliant.

ekiwi:
> Should the bundle definitions for the standard library types aim for full spec compliance, or would it be more useful to take an opinionated approach when it comes to clocks / resets? Using Chisel's implicit clock and reset would simplify implementations but it would also make it more difficult to integrate existing implementations written in e.g. verilog/VHDL that are fully spec compliant.

I think we should rely on Chisel's implicit clock and reset whenever possible and not artificially limit ourselves. In order to connect to external components, we will provide adapters that ensure that you will get ports that are named according to the AXI standard. I think - as part of this adapter - we can also include some glue logic to expose the clock and reset with the standard names and polarities.

jwright6323:
Even though ACLK and ARESETn are marked as required, they're frequently omitted or named something else when there are multiple buses on the same clock/reset domain

chiplet:
> I think we should rely on Chisel's implicit clock and reset whenever possible and not artificially limit ourselves.

Now that I think about it, another good reason for going with implicit clocks is that it dramatically simplifies testing from Chisel. IIRC chiseltest only supports implicit clocks and resets for its testers so defining custom clock and reset would force designs to use some other verification framework.

unlsycn:
It seems to be impossible to convert the Axi4IO to ExternalAxi4IO by using DataView without subword assignment, since bundles such as Axi4Response cannot be viewed as a UInt.

PR #2443, opened 2022-03-11
Initialize lastGrant to make round robin arbitor have a determined in…

lingscale:
…itial state.

### DIFF
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -78,7 +78,7 @@ abstract class LockingArbiterLike[T <: Data](gen: T, n: Int, count: Int, needsLo
 
 class LockingRRArbiter[T <: Data](gen: T, n: Int, count: Int, needsLock: Option[T => Bool] = None)
     extends LockingArbiterLike[T](gen, n, count, needsLock) {
-  lazy val lastGrant = RegEnable(io.chosen, io.out.fire)
+  lazy val lastGrant = RegEnable(io.chosen, 0.U, io.out.fire)
   lazy val grantMask = (0 until n).map(_.asUInt > lastGrant)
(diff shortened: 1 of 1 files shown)


aswaterman:
See https://github.com/chipsalliance/chisel3/pull/2329

lingscale:
Thank you, in fact I care about reset certainty of real chip more than simulation behavior. I'll think more carefully.

jiegec:
FWIW, if you do care about this and do not want to change the source of chisel3, use the following firrtl transform:

```scala
import firrtl.CircuitState
import firrtl.DependencyAPIMigration
import firrtl.Transform
import firrtl.ir.ExtModule
import firrtl.ir._

/** Zero-initialized registers if no reset is specified.
  */
object ZeroInit extends Transform with DependencyAPIMigration {
  override def prerequisites = Seq()

  override def invalidates(a: Transform) = false

  override protected def execute(state: CircuitState): CircuitState = {
    state.copy(circuit = state.circuit.mapModule(onModule(_)))
  }

  private def onModule(m: DefModule): DefModule =
    m match {
      case e: ExtModule => e
      case mod: Module =>
        mod.copy(body = mod.body.mapStmt(onStmt))
    }

  private def onStmt(s: Statement): Statement = s match {
    // special handle for RRArbiter lastGrant
    // see https://github.com/chipsalliance/chisel3/pull/2329
    case i: DefRegister => {
      if (i.name == "lastGrant" && i.tpe == UIntType(IntWidth(1))) {
        println(s"Hacking lastGrant")
        i.copy(reset = Reference("reset"), init = UIntLiteral(0, IntWidth(1)))
      } else {
        i
      }
    }
    case other => other.mapStmt(onStmt(_))
  }
}
```

The code can be adapted to any possible width of `lastGrant`.

PR #2444, opened 2022-03-11
[docs] Add Cookbook section on aliased Bundle fields

jackkoenig:
See [post mdoc rendered markdown here](https://gist.github.com/jackkoenig/299c16ba1be0144340ca580a4d2336d0#aliased-bundle-fields).

### Contributor Checklist

#### Type of Improvement

   - documentation        

#### API Impact

No impact

No impact

#### Desired Merge Strategy

   - Squash:

- [ ] Did you review?

### DIFF
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -20,6 +20,8 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
   * [Can I make a 2D or 3D Vector?](#can-i-make-a-2D-or-3D-Vector)
   * [How do I create a Vector of Registers?](#how-do-i-create-a-vector-of-registers)
   * [How do I create a Reg of type Vec?](#how-do-i-create-a-reg-of-type-vec)
+* Bundles
+  * [How do I deal with aliased Bundle fields?](#aliased-bundle-fields)
 * [How do I create a finite state machine?](#how-do-i-create-a-finite-state-machine-fsm)
 * [How do I unpack a value ("reverse concatenation") like in Verilog?](#how-do-i-unpack-a-value-reverse-concatenation-like-in-verilog)
 * [How do I do subword assignment (assign to some bits in a UInt)?](#how-do-i-do-subword-assignment-assign-to-some-bits-in-a-uint)
@@ -31,7 +33,7 @@ Please note that these examples make use of [Chisel's scala-style printing](../e
   * [How can I dynamically set/parametrize the name of a module?](#how-can-i-dynamically-setparametrize-the-name-of-a-module)
 * Directionality
   * [How do I strip directions from a bidirectional Bundle (or other Data)?](#how-do-i-strip-directions-from-a-bidirectional-bundle-or-other-data)
-  
+
 ## Type Conversions
 
 ### How do I create a UInt from an instance of a Bundle?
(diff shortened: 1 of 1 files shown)


docs/src/cookbooks/cookbook.md
```diff
@@ -231,6 +233,134 @@ class Foo extends RawModule {
...
+```
+
+Following the `gen` pattern when creating Bundles can result in some opaque error messages:
+
+```scala mdoc
+class AliasedBundle[T <: Data](gen: T) extends Bundle {
+  val foo = gen
+  val bar = gen
+}
+class AlsoAliasedBundle[T <: Data](val gen: T) extends Bundle {
```
mwachs5, line 257: Can you move this example later and say something like "you'd also get this error in this case..." since you don't talk about this code here it's a bit confusing

docs/src/cookbooks/cookbook.md
```diff
@@ -231,6 +233,134 @@ class Foo extends RawModule {
...
+Functionally the same as (1) but with more subtle syntax, you can use [Scala by-name function parameters](https://docs.scala-lang.org/tour/by-name-parameters.html):
+
+```scala mdoc
+class UsingByNameParameters[T <: Data](gen: => T) extends Bundle {
+  val foo = gen
+  val bar = gen
+}
+```
+
+With this usage, you not include `() =>` when passing the argument:
```
mwachs5, line 322: You do not include

mwachs5 (commented):
At a high level, should we rename this article to "how to properly use gen args in your bundles". Also, should we pick one recipe to recommend? As this is a cookbook not exhaustive reference on all the options? 

But thanks for writing this, it actually helped me understand why I never quite knew which pattern to use... they are all equivalent!

docs/src/cookbooks/cookbook.md
```diff
@@ -231,6 +233,134 @@ class Foo extends RawModule {
...
+
+```scala mdoc:invisible
+getVerilogString(new Top(new DirectionedBundle(UInt(8.W))))
+```
+
+This approach is admittedly a little ugly and may mislead others reading the code because it implies that this Bundle is intended to be used as an `Output`.
+
+#### 4. Call `.cloneType` directly
+
+You can also just call `.cloneType` on your `gen` argument directly.
```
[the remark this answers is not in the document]
jackkoenig: In Chisel2, the distinction between types and values was super blurry so users often ended up just peppering `.cloneType` everywhere to make sure things worked correctly. The idea in Chisel3 was to have a stronger distinction here--binding operators only take types, and chiselTypeOf only converts from hardware values back to types.

That being said, my use of `.cloneType` here is back to the Chisel2 style and is admittedly just avoiding this issue.

docs/src/cookbooks/cookbook.md
jackkoenig: Sure.

jackkoenig:
> At a high level, should we rename this article to "how to properly use gen args in your bundles". Also, should we pick one recipe to recommend? As this is a cookbook not exhaustive reference on all the options?

Probably, but I was trying to avoid taking a position at the moment. I think we should decide on a style to recommend, but it's tricky because the style that leads to this problem is really really common. I think I'd like to merge this example as is since it is useful, but we should think about what style we want to recommend and then go with it.

To elaborate a little bit on why I'm avoiding taking a position, here's a little discussion:

------------------------

### Field

Note that this is a long running option and there was an old proposal to add `Field(...)` as a new undirectioned operator analogous to `Output` that you could use in this case. The rub there is that to be consistent it would be required on all members of the `Bundle`, eg. if you had something like
```scala
class MyBundle[T <: Data](gen: T) extends Bundle {
  val foo = UInt(8.W)
  val bar = UInt(32.W)
  val fizz = gen
  val buzz = Vec(2, UInt(8.W))
}
```

If you want to add another field to `MyBundle` that uses `gen` and we had `Field`, the new `Bundle` would be:
```scala
class MyBundle2[T <: Data](gen: T) extends Bundle {
  val foo = Field(UInt(8.W))
  val bar = Field(UInt(32.W))
  val fizz = Field(gen)
  val buzz = Field(Vec(2, UInt(8.W)))
  val other = Field(gen)
}
```
People understandably find this a bit annoying. Also note that this same problem applies to the current legal approach of just using `Output`. 

### Thunkification `() => T`

So on the flip side, say we recommend `() => T` or `=> T`. These probably make the most sense but have their own issues being infectious from bottom up. Say we have `MyBundle` from above and we use it in some larger design:

```scala
// Some datastructure of parameters to my generator
case class GeneratorParams[T <: Data](x: Int, y: Int, gen: T)

class MyModule[T <: Data](params: GeneratorParams) extends Module {
 ...
  val foo = IO(new MyBundle(params.gen))
```
Now the problem, is when someone updates `MyBundle` to `MyBundle3`:
```scala
class MyBundle[T <: Data](gen: () => T) extends Bundle {
  val foo = UInt(8.W)
  val bar = UInt(32.W)
  val fizz = gen()
  val buzz = Vec(2, UInt(8.W))
  val other = gen()
}
```

The will likely naively update all calls like so:
```scala
class MyModule[T <: Data](params: GeneratorParams) extends Module {
 ...
  val foo = IO(new MyBundle(() => params.gen))
                         // ^ Problem
```
This will give the exact same aliasing error. The problem is that the "thunkification" of `gen` must go all the way up to the top.

docs/src/cookbooks/cookbook.md
jackkoenig: Fixed.

mwachs5 (approved):

PR #2449, opened 2022-03-15
BitSetRange API

CircuitCoder:
Add a new implementation of `chisel3.util.experimental.BitSet`, `BitSetRange`, matching continuous bit ranges. Use cases include representing address ranges.

**This is currently a WIP**.
TODO: A `matches` method is added to all `BitSet` for testing, which currently has a naive implementation. We need to change that to a better implementation (e.g. using decoder)
TODO: Maybe add a `BitSet.range` as an alias of `BitSetRange.apply`?
TODO: Maybe add a overload `BitSetRange.apply(Range)` to permit code such as `BitSetRange(0x1000 until 0x5000)`
TODO: Documentations are not finished.

### Contributor Checklist

#### Type of Improvement

new API

#### API Impact

Add a new class `BitSetRange` representing `BitSet` corresponding to continuous bit ranges. 
Add a new object `BitSetRange` for constructing `BitSetRange` instances.

#### Desired Merge Strategy

Squash

- [ ] Did you review?


src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -165,6 +166,8 @@ package experimental {
     /** whether this [[BitSet]] is empty (i.e. no value matches) */
     def isEmpty: Boolean = terms.forall(_.isEmpty)
 
+    def matches(input: UInt) = VecInit(terms.map(_ === input).toSeq).asUInt.orR
+
```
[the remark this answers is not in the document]
CircuitCoder: So that means one single `matches` call will match multiple inputs?

integration-tests/src/test/scala/chiselTests/util/experimental/BitPat.scala
```diff
@@ -0,0 +1,32 @@
+package chiselTests.util.experimental
```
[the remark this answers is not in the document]
CircuitCoder: Fixed in 6d012f70

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -226,6 +229,51 @@ package experimental {
     }
   }
 
+  sealed class BitSetRange(val start: BigInt, val length: BigInt, val width: Int) extends BitSet { outer =>
+    override def terms = {
+      val collected = mutable.Set[BitPat]();
+      var ptr = start;
+      var left = length;
+      while (left > 0) {
+        var cur_pow = left.bitLength - 1
```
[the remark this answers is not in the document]
CircuitCoder: Fixed in 6d012f70

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -226,6 +229,51 @@ package experimental {
...
+        if (ptr != 0) {
+          val max_pow = ptr.lowestSetBit
+          if (max_pow < cur_pow) cur_pow = max_pow
+        }
+
+        val inc = BigInt(1) << cur_pow
+        assert((ptr & inc - 1) == 0)
```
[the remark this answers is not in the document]
CircuitCoder: Added in 6d012f70

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -226,6 +229,51 @@ package experimental {
...
+      }
+
+      collected.toSet
+    }
+
+    // TODO: width!
+    override def toString: String = s"BitSetRange(0x${start.toString(16)} - 0x${(start + length).toString(16)})"
+  }
+
+  object BitSetRange {
```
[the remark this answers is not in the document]
CircuitCoder: Yes I'll add them after the API goes through discussion.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -165,6 +166,8 @@ package experimental {
     /** whether this [[BitSet]] is empty (i.e. no value matches) */
     def isEmpty: Boolean = terms.forall(_.isEmpty)
 
+    def matches(input: UInt) = VecInit(terms.map(_ === input).toSeq).asUInt.orR
```
sequencer, line 169: Maybe override `BitSetRange.matches` to use `input >= start.U && input < (start + length).U`? It should leverage the power of synthesis to generate two adder, while `BitSet.matches` will generate a high fanout `OR` and a lot of `XOR` for misaligned `BitSetRange`.

sequencer:
Generally LGTM, would you mind have a try to copy-paste this PR to RocketChip? replacing its address decoder, to see if it can pass the rocket-chip CI.

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: I believe this difference in behavior may surprise users? For example, `rangeA.matches(input)` generates two comparisons, while
```
rangeA.intersects(rangeB).matches(input)
```
generates a decoder, although the intersection is always a range itself.

I suggest that we leave `matches` as-is in BitSetRange, while adding a new function (e.g. `matchAsCmp`) to BitSetRange, so users will be alerted when they find out `rangeA.intersects(rangeB)` has no such member function, and no unexpected behavior will emerge.

Later, we may add APIs such as `BitSetRange.intersectRange(BitSetRange)` or `BitSet.tryAsRange`.

CircuitCoder:
Thoughts on converting from generic BitSet to BitSetRange:

We can also convert ANY BitSet into a list of continuous ranges. The problem is that the number of ranges maybe very large (e.g. `BitPat("???????????????0")`). So an API similar to `BitSet.toRanges: Seq[BitSetRange]` may take ages.

An alternative is `BitSet.toRanges: Iterator[BitSetRange]`, then the potentially large number of ranges is the API users' problem. They can choose to handle it when working with dynamic data, or just collect into a `Seq` when they are sure there is not many regions.

The steps to generate each range is as follows:
- Calculate the lower bound of the BitSet, which is quite easy to do: fill in 0 into ? in all terms, and pick the smallest one. Let `l` be that lower bound.
- Calculate the lower bound of `complement of (the original bitset unions [0, l))`. Let that be `r`.
- [l, r) is the lowest continuous region. Subtract that region from the bitset in the iterator.

Another idea is add a `maxRegionNumber` argument. But I personally prefers the one using iterators, as it provides more flexibility.

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: Also, we may change matches into using something similar to:
```scala
object MatchAlgo extends Enumeration {
  val Auto, Cmp, Decoder = Value()
}
def matches(input: UInt, algo: MatchAlgo = MatchAlgo.Auto)
```

And allow using comparisons on all BitSets (See https://github.com/chipsalliance/chisel3/pull/2449#issuecomment-1075889265)

sequencer: Yes, this sounds to be a better idea.
IMO: 
1. `matches` for`BitPat` is only the `===` alias.
2. `BitSet` should use `Decoder`.
3. `BitSetRange` should use `Compare`(I prefer `Compare` than `Cmp` ;p)
If user explicitly uses the `Compare` option for `BitSet`, `BitSet` should call `bitSet.toBitSetRanges.map(_.matches(input)).reduce(_ || _)`.

sequencer:
Here is my thought: 
1. we do need a `BitSetRange` type, semantic of which is: "a continues `BitSet`", here are two reasons:
- for `matches` function, `Compare` can provide better PPA since it leverage the adder rather than pure PLA to match on signals. 
- `toString` function can be overridden to print the Range rather than the confusing truth table.
2. Generally, I like the `Iterator` idea.
- It's a good idea to throw the NPC problem to users. 
- only consuming a part of `BitSet` is irrational, for most of usages, all converted data is needed. Whether this to be a user API need to be discussed, I think it need to be decided by @jackkoenig.
- another propose is having another API: `toBitSetRange(giveUpAt: Int = someReasonableThreshold): Seq[BitSetRange]`, this API should use the `Iterator` to generate and finally return a `Seq` of `BitSetRange` or throw an error.

CircuitCoder:
@sequencer An initial version of `BitSet.toRanges` is implemented in c8013dd0. May requires additional perfing, but since the BitSet within the iterator should never increase in term count, it should mostly be fine.

I'm working on providing some sort of randomized testing.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -224,8 +227,95 @@ package experimental {
...
+
+    /**
+     * Calculate the inverse of this pattern set.
+     * 
+     * @return A BitSet matching all value (of the given with) iff it doesn't match this pattern.
+     */
+    def inverse: BitSet = {
+      val total = BitPat("b" + ("?".repeat(this.getWidth)))
+      total.subtract(this)
+    }
```
sequencer, line 239: It may run forever, but this API is necessary, I think the time complexity may need to be documented? or at least tell users this API is time consuming.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -224,8 +227,95 @@ package experimental {
...
+
+    /**
+      * Calculate lowest posible value to match this pattern seset.
+      *
+      * @throws UnsupportedOperationException if this this is an empty pattern set.
+      * @return The matched value
+      */
+    def lowerbound: BigInt = terms.map(_.lowerbound).min
```
sequencer, line 247: I think we should private this, since I have no idea the use case.
If you can give some common use cases, we can public at it is.

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -224,8 +227,95 @@ package experimental {
...
+    /**
+      * Convert arbitrary BitSet into continous ranges
+      * 
+      * @note Due to the potential large number of generated ranges, this method return an iterator.
+      * API user should be careful when the number of the ranges are not guarenteed.
+      *
+      * @return An iterator yielding all continous ranges in this pattern. The ranges are guaranteed to be increasing,
+      *     non-overlaping, and all of them will precisely cover the entire BitSet.
+      */
+    def toRanges: Iterator[BitSetRange] = new BitSetRangeIterator(this)
```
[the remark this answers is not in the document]
sequencer: I think this API is useful!

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -224,8 +227,95 @@ package experimental {
...
+      require(w >= max_known_length, "Cannot construct a BitSetRange with width smaller than its range end")
+      val ret = new BitSetRange(start, length, w)
+
+      if (!allowUnaligned) require(ret.terms.size == 1, "Unaligned BitSetRange")
+
+      ret
+    }
   }
 
+  class BitSetRangeIterator(var remaining: BitSet) extends Iterator[BitSetRange] {
```
sequencer, line 307: private this class. :)

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -224,8 +227,95 @@ package experimental {
...
+      require(length > 0, "Cannot construct a empty BitSetRange")
+      val max_known_length = (start + length - 1).bitLength
+      val w = if (width >= 0) width else max_known_length
+      require(w >= max_known_length, "Cannot construct a BitSetRange with width smaller than its range end")
+      val ret = new BitSetRange(start, length, w)
+
+      if (!allowUnaligned) require(ret.terms.size == 1, "Unaligned BitSetRange")
+
+      ret
+    }
```
sequencer, line 304: This factory is useful, but I think for simplicity, we can move this API to `BitSet.range(start: BitInt, length: BigInt)`.
`width` can be refactored as `BitSet.expandWidthTo(width: Int)`.
`allowUnaligned` can also be refactored as `BitSetRange.isAligned: Bool` function.

We may need(or even change to) another API: `BitSet.range(range: Range)`

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: This is just a simple reduction into `BitSet.subtract`. If I'm not mistaken, the maximum number of generated terms is width * len(term), which doesn't seems that bad :tm:

Actually, if we are to document about the complexity, we may also want to document about the complexity of `BitSet.subtract`?

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: The reason this method is `public` is that another class (BitSetRangeIterator) need to invoke it. If we want to hide this method, we need to extract it into a `private object BitSetUtils`.

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: So how are we going to implement expandWidthTo? Fill in `?` or `0`? In the ranges case, we should obvious fill '0', but for more general bit patterns, this is not exactly clear what we should do.

IMHO two bit pattern of different width is two completely different thing with different schematic. There is no *obvious correct* way of converting a shorter one into a longer one.

src/main/scala/chisel3/util/BitPat.scala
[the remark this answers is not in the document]
CircuitCoder: Renaming is done in 0c30bd61

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: Changed to private in 0c30bd61

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -347,6 +437,8 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, val width: Int)
 
   override def isEmpty: Boolean = false
 
+  override def lowerbound: BigInt = value & mask
```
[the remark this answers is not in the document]
CircuitCoder: Done in 0c30bd61

src/main/scala/chisel3/util/BitPat.scala
sequencer: yeah, I agree, expanding width is tricky here. So maybe we still should keep the width as an arg.

src/main/scala/chisel3/util/BitPat.scala
sequencer: yes, basically all complexity should come from `subtract`. emmm. that's time consuming as well. I think mention it on ScalaDoc is good enough yet.

CircuitCoder:
Tests are added in 6f079f7b.

Actually I encountered an unexpected result from `BitSet.getWidth`. An empty BitSet returns zero for its width, but the BitSet may came from subtracting two BitSets with positive width. One would expect whenever I subtract another BitSet won't change one BitSet's width.

@sequencer Is this considered a bug, or a expected behavior? If this is indeed unexpected, then we need to keep a separate record of one BitSet's width. Also, `BitSet.empty` need to take an width argument (or we can keep a special value (e.g. -1) as a wildcard, and throw an exception on getWidth?)

Besides, are we going to allow binary subtraction / union / intersection between BitSets with **different** width? If we do, then how are we going to pad the shorter one?

jackkoenig:
I'm going to try to summarize a really good discussion we had in Chisel dev today (thanks to @sequencer and @seldridge for the great session!).

I'm going to use the term `AddressSet` instead of `BitSetRange` for purposes of this writeup for some clarity but just note they refer to the same idea.

With the current implementation, it does not make sense to have `AddressSet` extend `BitSet` because the underlying primitive (and unit of reuse) of `BitSet` is `terms` of type `Set[BitPat]`. The entire purpose of `AddressSet` is that some BitSet-like operations can be done more efficiently on that representation than they can be on the equivalent `Set[BitPat]` representation. But this benefit is thrown away whenever we do an operation that converts the `AddressSet` to `Set[BitPat]`, more-or-less defeating the purpose.

That being said, I do see the utility in being able to do something like take an `AddressSet`, intersect it with a `BitSet` and then check if a `UInt` matches the union. I see 3 options:

#### 1. BitSet and AddressSet are separate APIs (no inheritance relationship)

This is the simplest to implement and saves us from having to make difficult implementation decisions that will be discussed in the later options. This does mean that users will have to handle any composition of `AddressSet` with `BitSet` themselves. For example, if checking that a `UInt` matches the union of an `AddressSet` and a `BitSet`, a user will have to check the `UInt` against each and then `&&` the result.

#### 2. BitSet and AddressSet implement a common trait (HWSet? ChiselSet?)

Similarly how we introduced `BitSet` to make it possible to represent certain operations on `BitPat` that cannot be represented with a single `BitPat`, we need the ability to represent this operations on the combination of `BitSets` and `AddressSets`. The reason for this new type is that the `BitSet` primitive being `Set[BitPat]` does not adequately capture such combinations (it _can_ capture it but you get bad circuits as your outputs). Thus we can have other concrete subtypes of this new type called things like `Union`, `Intersection`, and `Subtraction` (or `Difference`). 

Thus, the union of an `AddressSet` and a `BitSet` will return a `Union`. This saves the higher-level user intent so that later uses of this `Union` (eg. decoding or checking if a `UInt` matches it) can still be implemented efficiently. Note that this does **not** mean we always return a `Union` when you union two `BitSet`, we can continue to use more optimal solutions whenever combining just `BitPat`s together.

Furthermore, this has highlighted to me that the current `Set[BitPat]` datastructure in `BitSet` is sort of a "normalized form" (it looks sort of like disjunctive normal form if you squint but perhaps `BitPat` is the real DNF). My real argument here is that when combining `AddressSet` with `BitSet` we do **not** want to normalize because we will get a worse circuit if we do.

#### 3. AddressSet extends BitSet but we deprecate and remove the `terms: Set[BitPat]` method

This one is basically the same in every way as (2), except that instead of introducing a new supertype (that we have to come up with a good name for), we just reclaim `BitSet` as the super type.

Originally this was my favorite option but with my recent realization that `BitSet` being a `Set[BitPat]` looks very similar to a normalized form (although may not be) makes me question if we should keep the `Set[BitPat]` representation around. I'm not really sure here though because we could just create a new concrete subtype of `BitSet` that captures `Set[BitPat]`.

CircuitCoder:
After some discussion with @sequencer (thanks for his patience since I had so much misunderstanding), now I really like the idea of the common trait approach. If I'm understanding correctly, this approach effectively decouples the construction, optimization, and consumption of a BitSet. Operations on BitSets become more similar to operations on immutable data structures, where we construct a tree without premature normalization/optimization. Optimization is the transformation between equivalent trees, and consumption has the semantics of simple recursions on trees.

If that's what we end up doing, I believe the current BitSet is the same as the union of some BitPats? So I guess (3) will work just fine. BitPat and AddressSet will be the leaf nodes in the tree and differs in the way we consume them (with pattern match or with comparison).

Some other thoughts on what to do now:

The original motivation behind the API is that there lacks a way to directly construct a BitSet matching ranges. How we can optimize the consumption of such ranges is sort of an afterthought. So an API like `BitSet.fromRanges` may fit this purpose better. This is reflected in the current implementation: `BitSetRange` is merely a type-level guarantee that a BitSet is continuous. Then now we have this extra information, we can do a little bit of optimization.

Since some library features (crossbar, router, etc) are blocked by the lack of a way to construct such BitSet, maybe we can add that feature first? Some options are:

- Remove `BitSetRange` entirely, move current `BitSetRange.terms` to `BitSet.fromRange`. Continous BitSets are immediately broken into BitPats during construction. This is the bare minimum implementation of such API, and discards any optimization opportunity until we switch to the new approach.
- Keep `BitSetRange`, maybe allow more optimization, e.g.:
  - Check if results are continuous in BitSet operations (intersect, union, etc.). If they are, return the corresponding continuous version instead.
  - Add methods like `BitSetRange.intercsectRange(BitSetRange): BitSetRange`, `BitSetRange.unionRange(BitSetRange): Maybe[BitSetRange]`.
- Just switch to the tree approach right away. The downside is that right now optimizations and consumptions have to be implemented in the frontend, so optimizations can only follow simple heuristics.

jackkoenig:
We've discussed this in the developers' meeting. In order to enable development without having to be 100% certain on the API, we should merge this to master and **not** backport to 3.5.x. This means we can iterate a bit using the SNAPSHOTs and see how we feel about the API.

@CircuitCoder 

> if I'm understanding correctly, this approach effectively decouples the construction, optimization, and consumption of a BitSet. Operations on BitSets become more similar to operations on immutable data structures, where we construct a tree without premature normalization/optimization. Optimization is the transformation between equivalent trees, and consumption has the semantics of simple recursions on trees.
>
> If that's what we end up doing, I believe the current BitSet is the same as the union of some BitPats? So I guess (3) will work just fine. BitPat and AddressSet will be the leaf nodes in the tree and differs in the way we consume them (with pattern match or with comparison).

Your understanding here is 100% correct, so as you say, it's really just how should we proceed.

I'm totally fine with merging some way of doing this even if it blasts down to `Set[BitPat]` and we figure out the optimization approach later. I think `BitSet.fromRange` is the most pragmatic approach to get something working right away so I am fine with it. My only concern is that because the API necessarily constructs a `BitSet` (and `BitSet` has `terms: Set[BitPat]`) we will either need to change this API before release of Chisel 3.6 or deprecate `terms: Set[BitPat]`. We don't have to make this decision right now, I am more than happy to approve something just to get the ball rolling.

I'm not super keen on keeping `BitSetRange` but having special handling to allow more optimization. If we're going to allow more optimization, we should switch to the tree approach. I think it's totally fine to add `BitSet.fromRange` (lacking optimization) for now, and then go straight to the tree implementation. While I agree that it will be hard to do as good of a job at the moment just in the frontend, I think that the heuristics we would follow would actually be pretty decent in practice and still leave the door open for direct IR support in the future and a better overall implementation. The neat thing about the tree approach (even if implemented in the frontend) is that a later IR implementation can be substituted in without any API changes.

**TLDR** Let's go ahead with merging `BitSet.fromRange` to master so we can play with the API, and we can discuss how we want to eventually implement the more complete, tree-based version.

sequencer:
@CircuitCoder any updates on this PR?

CircuitCoder:
Just moved all logics from `BitSetRange` into `BitSet.fromRange` in 2aefb34da2439c2b0abafa94105b3417b4c8f706. Thie PR should be ready to be merged!

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+    /**
+      * Construct a [[BitSet]] matching a range of value
+      *
+      * @param start The smallest matching value
+      * @param length The length of the matching range
+      * @param width The width of the constructed [[BitSet]]. If not given, the returned [[BitSet]] have the width of the maximum possible matching value.
+      * @return A [[BitSet]] matching exactly all inputs in range [start, start + length)
+      */
+    def fromRange(
+      start:  BigInt,
```
sequencer, line 156: How about adding another alias `def fromRange(range: Range, width: Int)`?

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+      length: BigInt,
+      width:  Int = -1
+    ): BitSet = {
+      require(length > 0, "Cannot construct a empty BitSetRange")
+      val max_known_length = (start + length - 1).bitLength
```
sequencer, line 161: sorry for my nitpick
```suggestion
      val maxKnownLength = (start + length - 1).bitLength
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+      val w = if (width >= 0) width else max_known_length
```
sequencer, line 162:
```suggestion
      val w = if (width >= 0) width else maxKnownLength
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+      require(
+        w >= max_known_length,
```
sequencer, line 164:
```suggestion
        w >= maxKnownLength,
```

sequencer (approved):
generally LGTM, ask @jackkoenig to review again.
In 3.7, we should move the`BitSet` manipulation logic into MFC,
but currently, we can just use this to create new circuit related to interconnection.

src/main/scala/chisel3/util/BitPat.scala
CircuitCoder: Fixed in c16799fd

src/main/scala/chisel3/util/BitPat.scala
sequencer: Maybe in this PR, we shouldn't add this.

integration-tests/src/test/scala/chiselTests/util/experimental/BitPat.scala
```diff
@@ -0,0 +1,34 @@
...
+  val testee = range.matches(input)
+  val ref = input >= start.U && input < (start + length).U
+
+  assert(testee === ref)
+}
+
+class BitSetRangeTest extends AnyFlatSpec with ChiselScalatestTester with Formal {
+  val rng = new Random(0x19260817)
+  val cases = 128
+  "BitSet.fromRange" should "be identical as comparesions" in {
```
jackkoenig, line 25:
```suggestion
  "BitSet.fromRange" should "be formally equivalent with bound checks" in {
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
       bs.getWidth
       bs
     }
+
+    /**
+      * Construct a [[BitSet]] matching a range of value
```
jackkoenig, line 148:
```suggestion
    /**  Construct a [[BitSet]] matching a range of value
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+      require(
+        w >= maxKnownLength,
+        s"Cannot construct a BitSetRange with width($w) smaller than its range end(b${(start + length - 1).toString(2)})"
+      )
+
+      // Break down to individual bitpats
+      val atoms = {
+        val collected = mutable.Set[BitPat]();
+        var ptr = start;
+        var left = length;
```
jackkoenig, line 172:
```suggestion
        val collected = mutable.Set[BitPat]()
        var ptr = start
        var left = length
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+      *
+      * @param start The smallest matching value
+      * @param length The length of the matching range
+      * @param width The width of the constructed [[BitSet]]. If not given, the returned [[BitSet]] have the width of the maximum possible matching value.
+      * @return A [[BitSet]] matching exactly all inputs in range [start, start + length)
+      */
+    def fromRange(
+      start:  BigInt,
+      length: BigInt,
+      width:  Int = -1
```
jackkoenig, line 158: Rather than using negative values and default values to indicate no `width` argument, it would be better to just overload this function with 1 version taking a width argument and the other one not, eg.
```scala
def fromRange(
   start:  BigInt,
   length: BigInt): BitSet = fromRange(start, length, (start + length - 1).bitLength)


def fromRange(
   start:  BigInt,
   length: BigInt,
   width:  Int
 ): BitSet = ???
```

src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -142,6 +143,57 @@ package experimental {
...
+    ): BitSet = {
```
jackkoenig, line 159: Is `BitSet` the type we want to return here? I'm fine with merging this to master as is but before we release `3.6` I want us to think about if we either want to deprecate `def terms` of `BitSet` or if we want to add a new type above `BitSet` that is not required to be represented using `Set[BitPat]`.

jackkoenig (commented):
I've left some nitty comments and an API suggestion (I don't like using `-1` as an indicator of "width not specified"). I'm not going to request changes because I will be gone for the next 2 weeks but I hope my suggestions are addressed before merging.

src/main/scala/chisel3/util/BitPat.scala
sequencer: I filed #2516 to private `terms` API in 3.6, and #2517 to deprecate in 3.5.
As for the BitSet type, I think in the future(maybe 3.7) it should be a trait to represent BitSet or result of BitSet operations.

PR #2465, opened 2022-03-30
Use var List instead of ListBuffer to save memory

jackkoenig:
This reduces memory use of every HasId by 64 bytes.

Every instance of HasId (including all Data) had 2 ListBuffer vals for
recording post-naming hooks, yet this feature is almost never used.
These are now vars of type List which allows the common case of Nil to
add no incremental memory use per instance of HasId.

I measured the 64 byte memory reduction with [JAMM](https://github.com/jbellis/jamm) (published as `io.github.stephankoelle:jamm:0.4.1`). JAMM reports that a ListBuffer uses 32 bytes of shallow memory. They use 48 bytes deep but looking at the code, the 4 vars inside of them are `Nil`, `null`, `false`, and `0` so the deep memory use is structurally shared.

I also measured this to reduce memory consumption of our Chisel generator of a large design from 5 GiB to 4.2 GiB. Not bad for only a few lines of change!

### Contributor Checklist

#### Type of Improvement

  - performance improvement      

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Reduce memory use of every `HasId` (includes `Data`) by 64 bytes

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -115,10 +115,10 @@ private[chisel3] trait HasId extends InstanceId {
   private var prefix_seed: Prefix = Nil
 
   // Post-seed hooks called to carry the suggested seeds to other candidates as needed
-  private val suggest_postseed_hooks = scala.collection.mutable.ListBuffer.empty[String => Unit]
+  private var suggest_postseed_hooks: List[String => Unit] = Nil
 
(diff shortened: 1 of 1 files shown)


seldridge (approved):
51yifu

jackkoenig:
Ah shoot, I overstated the savings slightly but they're still solid. ~12% instead of ~15% savings.

PR #2468, opened 2022-04-01
Prevent FIRRTL bulk connects on BlackBox Bundles.

### LINKED ISSUE #2457: Exception on Bulk Connect to BlackBox IO Bundle
Our code stopped working after upgrading to Chisel 3.5.2. The culprit is a bidirectional connect on a BlackBox IO Bundle.
Looking into it, I think the algorithm added in 3553a1583403824718923a6cc530cec3b38f5704 attempts to create a bulk connect to the BlackBox Bundle, which then causes the crash. I guess `BiConnect.canBulkConnectAggregates` would have to prevent this?

**Type of issue**: bug report

**Impact**: unknown

**Other information**
 

https://scastie.scala-lang.org/Fri405TeQYu0KhJiGvkpww

So far I could only find one scenario actually triggering the error: A BiDir connect on a flipped BlackBox IO Bundle containing at least one output.
 
```
import chisel3._

case class MyBundle() extends Bundle {
  val O: Bool = Output(Bool())
}

class MyModule() extends RawModule {

  val io: MyBundle = IO(Flipped(MyBundle()))
  val bb: MyBlackBox = Module(new MyBlackBox())

  bb.io <> io
}

class MyBlackBox() extends BlackBox {
   val io: MyBundle = IO(Flipped(MyBundle()))
}

print((new ChiselStage).emitFirrtl(new MyModule()))
```

`chisel3.internal.ChiselException: Internal Error! Could not get ref for 'chisel3.internal.package$ViewParentAPI@5ef91709' @[main.scala 15:9]! This is a bug in Chisel, please file an issue at 'https://github.com/chipsalliance/chisel3/issues/new'`

fzi-hielscher:
### Contributor Checklist

This is an attempt to fix #2457 . It should prevent crashes due to FIRRTL bulk connects being emitted on BlackBox IO Bundles.
I was not entirely certain on the best way to catch these cases in the `canBulkConnectAggregates` method. My assumption is that BlackBoxes do not contain any other Bundles than the one IO Bundle. So I am simply checking whether the parent Module of the sink or source Bundle is a BlackBox.

#### Type of Improvement

- bug fix

#### API Impact

None

Ideally, none. In the worst case it may unnecessarily prevent some FIRRTL bulk connects form being emitted, which should not have any functional impact.

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -285,7 +285,13 @@ private[chisel3] object BiConnect {
       case _ => true
     }
 
-    typeCheck && contextCheck && bindingCheck && flow_check && sourceNotLiteralCheck
+    // do not bulk connect the 'io' pseudo-bundle of a BlackBox since it will be decomposed in FIRRTL
+    def blackBoxCheck = Seq(source, sink).map(_._parent).forall {
+      case Some(_: BlackBox) => false
+      case _ => true
+    }
+
+    typeCheck && contextCheck && bindingCheck && flow_check && sourceNotLiteralCheck && blackBoxCheck
   }
 
   // These functions (finally) issue the connection operation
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM! Thank you for this fix!

PR #2471, opened 2022-04-01
Optimize memory use of naming prefixes

jackkoenig:
* Use a single field instead of two in HasId (4-bytes per HasId)
* Set the prefix to Nil after setting ref to free up memory

Draft PR for now because I'm having a bit of trouble measuring a memory improvement from this (perhaps I should write a synthetic benchmark), but analytically, it is easy to see how it should be beneficial.

This also has minor bug potential because the behavior of `_computeName` will change if someone calls it after `forceName` is called on a `HasId`. Auditing the code, that doesn't happen, but it might be worth having `_computeName` get the name from the `_ref` if it's available or perhaps error if called when it shouldn't be.

### Contributor Checklist

- [ ] Did you review?


seldridge (approved):
lgtm

Why have two mutable vars when you can just use one and overwrite it? 😉

PR #2477, opened 2022-04-06
git-subtree add sifive/chisel-circt

seldridge:
This commit is the first step towards enabling CIRCT-compiled Chisel natively inside Chisel (as opposed to relying on a separate library).

This does three things:

1) It uses `git-subtree` to add https://github.com/sifive/chisel-circt (with history) as a directory called `circt/`.
2) It moves and updates the contents of `circt/` into `src/main/scala/circt/` or `src/test/scala/circt/`.
3) It removes unnecessary files (build.sbt/build.sc) from the circt tree and updates the README to reference the fact that this is no longer a separate dependency.

This PR is expected to fail until we have a way of adding CIRCT as a dependency to the CI.


ekiwi:
Should this be in an `experimental` package or otherwise marked as not meant for general usage until stable binary releases for `firtool` are available for Mac and Linux?

seldridge:
Good point, Kevin. My plan is to block this PR (keep it as a draft) until we have binary releases available for CIRCT.

ekiwi:
> Good point, Kevin. My plan is to block this PR (keep it as a draft) until we have binary releases available for CIRCT.

Sounds good. Do you think it would be feasible to have some code that automatically determines the platform we are on and then downloads the latest firtool release (or a compatible version) for the current platform from github?
That way it would be just as easy as using the old firrtl compiler.

seldridge:
> Do you think it would be feasible to have some code that automatically determines the platform we are on and then downloads the latest firtool release (or a compatible version) for the current platform from github?

This sounds clever, but is likely over-engineered. 😅 

I think it's reasonable to handle this the same way we handle any other binary dependency, e.g., verilator with ChiselTest: give an excellent warning message if the tool isn't installed, but don't try to do a local install automatically unless it's part of some larger build system (like how Rocket Chip used to build a Verilator version for you).

mwachs5:
firtool binaries is now published on each release of CIRCT, so I plan to resurrect this PR

seldridge:
> firtool binaries is now published on each release of CIRCT, so I plan to resurrect this PR

Specifically, this was added in https://github.com/llvm/circt/pull/3617 and is automatically published as part of the release assets for CIRCT. E.g., the SiFive 1.11.0 release of CIRCT binaries are available here: https://github.com/llvm/circt/releases/tag/sifive%2F1%2F11%2F0 This is also what I'm using for CI in this PR.

seldridge:
Logistically, any feedback on this PR that actually requests code modifications I'll address on upstream `chisel-circt` and then fold into this PR. I'd like to keep this PR as a pure: (1) `git subtree add`, (2) `mv circt/ src/main/scala/`, (3) enable CI, and (4) tag a final release of `chisel-circt` and archive it. If there are more nitty concerns with the code that can wait, I'm signing myself up to fix these after this lands.

mwachs5:
ok, so you would prefer to merge outstanding chisel-circt PRs first before merging into chisel3?

seldridge:
> ok, so you would prefer to merge outstanding chisel-circt PRs first before merging into chisel3?

I'd rather not block this on those PRs. We can move them over here once this lands.

I was instead saying if Jack comes in and wants me to rewrite `CIRCTStage.scala`, I would instead go do that on `sifive/chisel-circt` and then rework this PR here to incorporate that. I'm aiming to have a clean break where `sifive/chisel-circt` ends development and is archived and that end point is where development in `chisel3` starts.

src/main/scala/circt/stage/phases/CIRCT.scala
```diff
@@ -0,0 +1,171 @@
...
+      case RunFirrtlTransformAnnotation(transform) =>
+        transform match {
+          /* Inlining/Flattening happen by default, so these can be dropped. */
+          case _: firrtl.passes.InlineInstances | _: firrtl.transforms.Flatten => false
+          /* ReplSeqMem is converted to a firtool option */
+          case _: firrtl.passes.memlib.ReplSeqMem =>
+            blackbox = true
+            false
+          /* Any emitters should not be passed to firtool. */
+          case _: firrtl.Emitter => false
```
ekiwi, line 77: How do you specify the target if you emitters are ignored? Like how would I distinguish between generating Firrtl or SystemVerilog?

mwachs5:
I agree that if there are tweaks we want to make to chisel-circt we can do that here once this is on.

What is the story for backporting and backwards compatability binary checking for this chunk of code? For now will we just have this on `master` branch and worry about the binary compatability story once it gets into 3.6?

jackkoenig:
> What is the story for backporting and backwards compatability binary checking for this chunk of code? For now will we just have this on master branch and worry about the binary compatability story once it gets into 3.6?

No backporting, this is a Chisel 3.6 addition.

mwachs5:
Ok, if this is a 3.6 addition and we are stopping any development on chisel-circt, that is a bit tricky for anyone who wants to use 3.5.x and chisel-circt and needs chisel-circt improvements... But I'm Ok with a workflow that makes parallel PRs to chisel-circt and this repo until this is actually released w chisel

src/main/scala/circt/stage/phases/CIRCT.scala
seldridge: There's a `CIRCTTargetAnnotation` that will control it. They main issue here is that there isn't a good way of adding multiple `RunFirrtlTransformAnnotation(_: firrtl.Emitter)`s to CIRCT as it's trying to work towards a fixed target, e.g., FIRRTL Dialect, HW Dialect, SV Dialect, or SV.

It's reasonable that if there was one `RunFirrtlTransformAnnotation(_ : firrtl.Emitter)` that this could be used to tell CIRCT what to compile to. However, it seemed cleaner to just strip this and force users to use the new annotation.

seldridge:
> that is a bit tricky for anyone who wants to use 3.5.x and chisel-circt and needs chisel-circt improvements

I wouldn't worry about this. `chisel-circt` is ranked 607574 on Maven. `chisel3` is ranked 17385. Also, `chisel-circt` was always an extremely experimental thing. I don't see a good reason to commit to maintenance of `sifive/chisel-circt` for a hypothetical user on 3.5.x when the answer is: (1) use SFC for 3.5.x (which is the official policy) or (2) switch to 3.6 (which should be easy if somebody is already relying on `chisel-circt` which has limitations around supported Chisel APIs).

seldridge:
I force pushed this after landing a PR that @mwachs5 needed. That closes out any known needed development on `sifive/chisel-circt`.

mwachs5 (approved):

src/main/scala/circt/stage/phases/CIRCT.scala
seldridge: Thinking about this a bit more... it would probably be better to actually error on this. I can do this in a follow-up. 

This should be basically dead code right now as the expectation is that a user is going to use `circt.stage.ChiselStage` which is _not_ going to use a `RunFirrtlTransformAnnotation` to tell CIRCT what emitter to use. There's a separate API/Annotation for that, `CIRCTTargetAnnotation`. A user could see weird behavior if they are doing one of the following:
1. Generating a `RunFirrtlTransformAnnotation(_ : Emitter)` inside a `ChiselAnnotation`.
2. Building a custom stage out of Chisel3 phases and CIRCT phases. 
3. Possibly other weird things.

seldridge:
Merging after verbal approval from @jackkoenig.

src/main/scala/circt/Implicits.scala
```diff
@@ -0,0 +1,19 @@
...
+
+    /** Construct an Option from a Boolean. */
+    def option[A](b: => A): Option[A] =
+      if (a)
+        Some(b)
+      else
+        None
+  }
+
+}
```
sequencer, line 19: Sorry for the delay review:
Should we private these to avoid introducing more public APIs?

.github/workflows/test.yml
```diff
@@ -42,6 +42,11 @@ jobs:
         uses: olafurpg/setup-scala@v10
         with:
           java-version: ${{ matrix.jvm }}
+      - name: Install CIRCT
+        run: |
+          mkdir usr
+          wget https://github.com/llvm/circt/releases/download/sifive%2F1%2F11%2F0/circt-bin-ubuntu-20.04.tar.gz -O - | tar -zx -C usr/
+          echo "$(pwd)/usr/bin" >> $GITHUB_PATH
```
sequencer, line 49: Hope we can have a better way to install CIRCT in the system.

src/main/scala/circt/stage/Annotations.scala
```diff
@@ -0,0 +1,129 @@
...
+    extends NoTargetAnnotation
+    with CustomFileEmission {
+
+  override protected def baseFileName(annotations: AnnotationSeq): String = filename
+
+  override def getBytes = value.getBytes
+
+}
+
+object CIRCTHandover extends HasShellOptions {
```
sequencer, line 100: I thought we might eventually get rid of the `Handover` behavior. Am I right?

src/main/scala/circt/stage/phases/AddDefaults.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package circt.stage.phases
+
+import circt.Implicits.BooleanImplicits
+import circt.stage.CIRCTHandover
+
+import firrtl.AnnotationSeq
+import firrtl.options.Phase
+
+class AddDefaults extends Phase {
```
sequencer, line 11: should this be private too?

src/main/scala/circt/stage/phases/MaybeSFC.scala
```diff
@@ -0,0 +1,50 @@
...
+package circt.stage.phases
+
+import circt.stage.{CIRCTHandover, CIRCTOptions}
+
+import firrtl.AnnotationSeq
+import firrtl.options.{Dependency, Phase}
+import firrtl.options.Viewer.view
+import firrtl.stage.{Forms, RunFirrtlTransformAnnotation}
+
+class MaybeSFC extends Phase {
```
sequencer, line 12: and here

sequencer (commented):
Sorry for the late view, but not too late, glad we didn't version 3.6.0 now.
Just some small questions:
1. Is possible to add as less as possible public APIs to make room for future improvements.
2. IIRC, the handover behavior will be removed eventually, what's the roadmap of this.

PR #2496, opened 2022-04-19
Utils: add source locators

mwachs5:
### Contributor Checklist

#### Type of Improvement

   - performance improvement       

   - backend code generation  

#### API Impact

This adds implicits to Cat (and I plan to add the other Utils in this PR). Which I think is a backwards-incompatible API change though no user code should actually have to change.

The resulting code will have more source locators for any utils.

#### Desired Merge Strategy

#### Release Notes

Add source locators for util functions Cat, ...

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/Cat.scala
+++ b/src/main/scala/chisel3/util/Cat.scala
@@ -3,6 +3,7 @@
 package chisel3.util
 
 import chisel3._
+import chisel3.internal.sourceinfo._
 
 /** Concatenates elements of the input, in order, together.
   *
@@ -19,7 +20,7 @@ object Cat {
   /** Concatenates the argument data elements, in argument order, together. The first argument
     * forms the most significant bits, while the last argument forms the least significant bits.
     */
-  def apply[T <: Bits](a: T, r: T*): UInt = apply(a :: r.toList)
+  def apply[T <: Bits](a: T, r: T*)(implicit sourceInfo: SourceInfo): UInt = apply(a :: r.toList)
 
   /** Concatenates the data elements of the input sequence, in reverse sequence order, together.
     * The first element of the sequence forms the most significant bits, while the last element
@@ -28,5 +29,5 @@ object Cat {
     * Equivalent to r(0) ## r(1) ## ... ## r(n-1).
     * @note This returns a `0.U` if applied to a zero-element `Vec`.
     */
-  def apply[T <: Bits](r: Seq[T]): UInt = SeqUtils.asUInt(r.reverse)
+  def apply[T <: Bits](r: Seq[T])(implicit sourceInfo: SourceInfo): UInt = SeqUtils.asUInt(r.reverse)
 }
--- a/src/test/scala/chiselTests/util/CatSpec.scala
+++ b/src/test/scala/chiselTests/util/CatSpec.scala
@@ -60,4 +60,15 @@ class CatSpec extends ChiselFlatSpec {
     chirrtl should include("node hi_hi = cat(in[0], in[1])")
   }
 
+  it should "have a source locator" in {
+    class MyModule extends RawModule {
+      val in = IO(Input(Vec(8, UInt(8.W))))
+      val out = IO(Output(UInt()))
+
+      // noPrefix to avoid `out` as prefix
+      out := noPrefix(Cat(in))
+    }
+    val chirrtl = ChiselStage.emitChirrtl(new MyModule)
+    chirrtl should include("cat(in[6], in[7]) @[CatSpec.scala")
+  }
 }


src/main/scala/chisel3/util/Cat.scala
```diff
@@ -3,6 +3,7 @@
 package chisel3.util
 
 import chisel3._
+import chisel3.internal.sourceinfo._
```
jackkoenig, line 6:
```suggestion
import chisel3.internal.sourceinfo.SourceInfo
```

src/main/scala/chisel3/util/Cat.scala
```diff
@@ -19,7 +20,7 @@ object Cat {
   /** Concatenates the argument data elements, in argument order, together. The first argument
     * forms the most significant bits, while the last argument forms the least significant bits.
     */
-  def apply[T <: Bits](a: T, r: T*): UInt = apply(a :: r.toList)
+  def apply[T <: Bits](a: T, r: T*)(implicit sourceInfo: SourceInfo): UInt = apply(a :: r.toList)
```
jackkoenig, line 23: Actually, we need to use `do_apply` macros here because one could do a bit extraction immediately after a `Cat`.

mwachs5: I do not know what that means, can you point to an example?

jackkoenig: It's used a lot in `Bits.scala`: https://github.com/chipsalliance/chisel3/blob/44165a259bb16733a41798edca6b554b13f1d54a/core/src/main/scala/chisel3/Bits.scala#L73.

Here's a link to me explaining what this is and how it works in the Chisel internals talk at the CCC last year: https://youtu.be/2_WPFk8lriE?t=1166

mwachs5: ok, i believe I have this right now.

src/main/scala/chisel3/util/Bitwise.scala
```diff
@@ -24,13 +26,27 @@ object FillInterleaved {
     *
     * Output data-equivalent to in(size(in)-1) (n times) ## ... ## in(1) (n times) ## in(0) (n times)
     */
-  def apply(n: Int, in: UInt): UInt = apply(n, in.asBools)
+  final def apply(n: Int, in: UInt): UInt = macro SourceInfoTransform.nInArg
```
mwachs5, line 29: i realized i was inconsistent in marking these as `final def`. Is that correct/required if these are `object`s?

jackkoenig: I don't know what difference it makes on `object`s, trying to figure that out

jackkoenig: I asked on Scala gitter. It doesn't make any difference to use `final` and thus it's discouraged since doing it suggests that it _does_ mean something to anyone reading the code.

mwachs5: cleaned this up.

src/main/scala/chisel3/util/Bitwise.scala
```diff
@@ -24,13 +26,27 @@ object FillInterleaved {
     *
     * Output data-equivalent to in(size(in)-1) (n times) ## ... ## in(1) (n times) ## in(0) (n times)
     */
-  def apply(n: Int, in: UInt): UInt = apply(n, in.asBools)
+  def apply(n: Int, in: UInt): UInt = macro SourceInfoTransform.nInArg
+
+  def do_apply(n: Int, in: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt =
```
jackkoenig, line 31:
```suggestion
  /** @group SourceInfoTransformMacro */
  def do_apply(n: Int, in: UInt)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): UInt =
```
This grouping helps a lot in the ScalaDoc, eg. https://www.chisel-lang.org/api/latest/chisel3/UInt.html

jackkoenig: We should add this to all new `do_` methods.

mwachs5: do I need to create the group first for the `Fill` thing? I think I need to do more than just adding the `@group` right, I need to copy all the "here be dragons" stuff ...

mwachs5: figured it out. fixed.

chick (approved):
Looks good to me. I did not go deep in analysis 
but code reads well and tests look good

PR #2497, opened 2022-04-20
fix for scalacPluginClasspath in build.sc

sequencer:
refer to
https://github.com/com-lihaoyi/mill/blob/61698d6dab93d18b050fab31fcc7d9c57716c430/scalalib/src/ScalaModule.scala#L25-L32
these parameters are included. adding those will lead to strange
behavior.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -105,9 +105,9 @@ class chisel3CrossModule(val crossScalaVersion: String) extends CommonModule wit
   }
 
   object test extends Tests with TestModule.ScalaTest {
-    override def scalacPluginClasspath = m.scalacPluginClasspath
+    override def scalacPluginClasspath = T { m.scalacPluginClasspath() }
 
-    override  def scalacOptions = T {
+    override def scalacOptions = T {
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -105,12 +105,6 @@ class chisel3CrossModule(val crossScalaVersion: String) extends CommonModule wit
   }
 
   object test extends Tests with TestModule.ScalaTest {
-    override def scalacPluginClasspath = m.scalacPluginClasspath
-
-    override  def scalacOptions = T {
-      super.scalacOptions() ++ Agg("-P:chiselplugin:genBundleElements")
-    }
-
```
jackkoenig, line 113: You don't want to delete the `scalacOption` because we **only** specify this argument for the tests, not for main Chisel (because it would be a binary incompatible change on a minor release, we'll make it mandatory in 3.6)

sequencer: yep, I hit this issue, I need more time to debug the mill build system.

PR #2501, opened 2022-04-23
fix plugin while building with mill.

sequencer:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -104,12 +104,12 @@ class chisel3CrossModule(val crossScalaVersion: String) extends CommonModule wit
     )
   }
 
-  object test extends Tests with TestModule.ScalaTest {
-    override def scalacPluginClasspath = m.scalacPluginClasspath
+  override def scalacOptions = T {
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -104,12 +104,12 @@ class chisel3CrossModule(val crossScalaVersion: String) extends CommonModule wit
     )
   }
 
-  object test extends Tests with TestModule.ScalaTest {
-    override def scalacPluginClasspath = m.scalacPluginClasspath
+  override def scalacOptions = T {
+    super.scalacOptions() ++ Agg(s"-Xplugin:${plugin.jar().path}", "-P:chiselplugin:genBundleElements")
```
jackkoenig, line 108: We can start generating Bundle elements on master for 3.6, but we should probably just do it by making it always on rather than using the argument. Regardless this is fine as long as it isn't backported to 3.5.x

jackkoenig, line 108: We can start generating Bundle elements on master for 3.6, but we should probably just do it by making it always on rather than using the argument. Regardless this is fine as long as it isn't backported to 3.5.x

jackkoenig (approved):

PR #2512, opened 2022-04-29
Support separately elaborating definition and instance in ChiselStage

debs-sifive:
### Contributor Checklist

#### Type of Improvement

  - new feature/API

#### TECH DEBT

Need to improve the API to instantiate previously-defined `Definition`s in the future (right now, we are passing in a `Definition` into the `Top` of the second generator call). Tracked in https://github.com/chipsalliance/chisel3/issues/2514

Need to improve the code that handles module name collisions. Right now I am making a bunch of `ImportedDefinitionAnnotation`s on the imported `Definition` and all of its submodules; this list of `Annotation`s must get passed around between elaboration stages to ensure no name collisions. Long-term solution may require the support of namespaces in FIRRTL.

#### API Impact

None.

None.

#### Desired Merge Strategy

#### Release Notes

Support for separately elaborating a Definition and Instance.

- [x] Did you review?


core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -107,9 +108,33 @@ object Instance extends SourceInfoDoc {
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
   ): Instance[T] = {
+    // Check to see if the module is already defined
```
mwachs5, line 111:
```suggestion
    // Check to see if the module is already defined either internally or externally
```

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -107,9 +108,33 @@ object Instance extends SourceInfoDoc {
...
+    val existingMod = Builder.components.map {
+      case c: DefModule if c.id == definition.proto => Some(c)
+      case c: DefBlackBox if c.name == definition.proto.name => Some(c)
+      case _ => None
+    }.flatten
+
+    if (existingMod.isEmpty) {
+      // Add a module that will get extracted later on so that FIRRTL does not
```
mwachs5, line 119: what does "extracted" mean? Do you mean:
```suggestion
      // Add a Definition that will get emitted as an ExtModule so that FIRRTL does not
```

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -363,8 +364,21 @@ private[chisel3] class ChiselContext() {
   val viewNamespace = Namespace.empty
 }
 
+/** Stores a [[Definition]] that is imported so that its Instances can be
+  * compiled separately.
+  */
+case class ImportedDefinitionAnnotation[T <: BaseModule with IsInstantiable](importedDefinition: Definition[T]) extends NoTargetAnnotation
+
 private[chisel3] class DynamicContext(val annotationSeq: AnnotationSeq, val throwOnFirstError: Boolean) {
+  val importedDefinitionsAnno = annotationSeq.collect{case a: ImportedDefinitionAnnotation[_] => a}
```
mwachs5, line 373:
```suggestion
  val importedDefinitionsAnnos = annotationSeq.collect{case a: ImportedDefinitionAnnotation[_] => a}
```

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -363,8 +364,21 @@ private[chisel3] class ChiselContext() {
...
   val globalNamespace = Namespace.empty
+
+  // Ensure imported Definitions get the same name as their modules so that
+  // they can be properly linked to Instances that are separately compiled
+  importedDefinitionsAnno.foreach { importedDef =>
```
mwachs5, line 378:
```suggestion
  importedDefinitionsAnnos.foreach { importedDef =>
```

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -363,8 +364,21 @@ private[chisel3] class ChiselContext() {
...
+  * compiled separately.
+  */
+case class ImportedDefinitionAnnotation[T <: BaseModule with IsInstantiable](importedDefinition: Definition[T]) extends NoTargetAnnotation
+
 private[chisel3] class DynamicContext(val annotationSeq: AnnotationSeq, val throwOnFirstError: Boolean) {
+  val importedDefinitionsAnno = annotationSeq.collect{case a: ImportedDefinitionAnnotation[_] => a}
   val globalNamespace = Namespace.empty
+
+  // Ensure imported Definitions get the same name as their modules so that
+  // they can be properly linked to Instances that are separately compiled
```
mwachs5, line 377:
```suggestion
  // Ensure imported Definitions emit as ExtModules with the correct name
  // so that instantations will also use the correct name, and prevent any name conflicts with
  // Modules/Definitions in this elaboration
```

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+  describe("(0): Elaborating an Instance separately from its Definition") {
+    it("should result in an instantiation in FIRRTL without a module declaration.") {
+      val testDir = createTestDirectory(this.getClass.getSimpleName).toString
+      val dutAnnos = (new ChiselStage).run(Seq(
+        ChiselGeneratorAnnotation(() => new AddOne),
+        TargetDirAnnotation(testDir),
+      ))
+
+      // Grab DUT definition to pass into testbench
+      val designAnnos = dutAnnos.flatMap { a =>
```
mwachs5, line 27:
```suggestion
      val designDefs = dutAnnos.flatMap { a =>
```

isnt what is returned a Definition[AddOne]?

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+import chisel3._
+import chisel3.stage.{ChiselGeneratorAnnotation, ChiselStage, DesignAnnotation}
+import chisel3.internal.ImportedDefinitionAnnotation
+import chisel3.experimental.hierarchy.{Definition, Instance}
+import firrtl.options.TargetDirAnnotation
+
+import java.nio.file.Paths
+import scala.io.Source
+
+class SeparateElaborationSpec extends ChiselFunSpec with Utils {
```
mwachs5, line 15: can you add another test for the name conflict case? Where you have Definition(AddOne) and then in the second elaboration you also make a `newDef = Definition(AddOne)` + `Instance(newDef)`?

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+        val inst0 = Instance(defn)
+        val inst1 = Instance(defn)
+
+        inst0.in := 0.U
+        inst1.in := 0.U
+      }
+
+      // If there is a repeat module definition, FIRRTL emission will fail
+      val firrtl = (new ChiselStage).emitFirrtl(
+        gen = new Testbench(dutDef),
```
azidar, line 113: Right now, the only way to instantiate a definition from a previous elaboration is to pass it as an argument to the top module in the next elaboration. This means that we must support a better API (e.g. like Instantiate), prior to release.

mwachs5 (commented):
Looks good to me! Do you think this can be backported to 3.5?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -363,8 +372,23 @@ private[chisel3] class ChiselContext() {
   val viewNamespace = Namespace.empty
 }
 
+/** Stores a [[Definition]] that is imported so that its Instances can be
+  * compiled separately.
+  */
+case class ImportedDefinitionAnnotation[T <: BaseModule with IsInstantiable](importedDefinition: Definition[T])
```
azidar, line 378: There must be a better place to put this class than in `Builder.scala`. Is there another place that annotations are defined, within the `core` package?

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+      (tb_rtl should not).include("module AddOne")
+    }
+  }
+
+  describe("(1): Elaborating an Instance and Definition together") {
+    it("should not result in a repeat definition of the module.") {
+      val testDir = createTestDirectory(this.getClass.getSimpleName).toString
+
+      class Top extends Module {
+        val inst = Instance(Definition(new AddOne))
```
azidar, line 70: This is not testing what we talked about. Here, you are creating a definition, then instantiating it. This should always work (regardless of your changes in your PR).

We care about the following case:

```
class Top(dut: Definition[DUT]) {
  val i0 = Instance(Definition(new DUT))
  val i1 = Instance(dut)
}
```

We need to make sure that i0's definition's module's name is different than i1's definition's module's name. Given your code, i1's module name will get the original dut's definition's module's name. We need the test to actually test this.

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+            Some(a.design.asInstanceOf[AddOne].toDefinition)
+          case _ => None
+        }
+      }
+      require(designAnnos.length == 1, s"Exactly one DesignAnnotation should exist, but found: $designAnnos.")
+      val dutDef = designAnnos.head
+
+      class Testbench(defn: Definition[AddOne]) extends Module {
+        val inst0 = Instance(defn)
+        val inst1 = Instance(defn)
```
azidar, line 105: This doesn't test what we talked about. See above comment.

azidar (changes requested):
Please update tests according to our feedback. Also, we should record in the PR description future work to create an API to instantiate a previously-defined Definition, which is not passing the Definition via an argument to the Top of the second elaboration call.

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
debs-sifive: I wrote this test to make sure that if we're doing things the "regular" way, an `ExtModule` is NOT created by `Instance`... But now that you point out this is the regular case, this test case does seem redundant with our regular D/I tests; I will remove it and add the case you stated!

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
debs-sifive: This test is ensuring that I the new code in `Instance` doesn't create 2 ExtModules (one per instance) instead of 1.

core/src/main/scala/chisel3/internal/Builder.scala
debs-sifive: Would it make sense to stick in `Definition.scala`?

azidar: Yeah, I think that's a good place.

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+      val tb_rtl = Source.fromFile(s"$testDir/Testbench.v").getLines.mkString
+      tb_rtl should include("module AddOne_1(")
+      tb_rtl should include("AddOne_1 mod (")
+      (tb_rtl should not).include("module AddOne(")
+      tb_rtl should include("AddOne inst (")
+    }
+
+    it(
+      "(0.b): should not occur between an Instance of a Definition and an Instance of a previously elaborated Definition."
+    ) {
```
debs-sifive, line 72: @mwachs5 @azidar I think you both were asking for the same test case, this is it!

debs-sifive:
@mwachs5 @azidar not sure if it needs to be handled in this PR, but what should we do if there are multiple "previously elaborated" definitions with the same name? There doesn't seem to be a way to catch that right now.

Like if I make two `AddOneParameterized`s of different widths they both still think they are named `AddOneParameterized`.

cc @sequencer this is what we were discussing in chisel meeting

sequencer:
Yes. Currently Builder will handle the name confliction and rename to Queue_1 Queue_2. But in the Stage API there is no way to resolve the naming issue.
Maybe a prefix API might be required here?

mwachs5:
> what should we do if there are multiple "previously elaborated" definitions with the same name? There doesn't seem to be a way to catch that right now.

Can we error?

debs-sifive:
I can make `Builder` error if it sees more than one `ImportedDefinitionAnnotation` of the same module type...

But that will totally prohibit passing in more than 1 definition of the same module type, which is probably not what we want long term.

debs-sifive:
Another option is to only allow for a single `ImportedDefinitionAnnotation` for now and say we only support the testbench use case? 🤷‍♀️

mwachs5:
> I can make Builder error if it sees more than one ImportedDefinitionAnnotation of the same module type...

> But that will totally prohibit passing in more than 1 definition of the same module type, which is probably not what we want long term.

It's not just the Definition of the same module type, that definition can't instantiate anything (e.g. a Queue) with Module or Instance that are going to be used in a larger context.

So far we handle this with annotations e.g. nested Module prefixing annotation. Maybe this is a good reason to keep pushing on the chisel prefixing API vs relying on annotations for this, but I don't think it really changes this PR.

debs-sifive:
R.e. multiple imported definitions with the same name, from discussion with @azidar, for this PR we should:
* pass a list of `ImportedDefinitionAnnotations` between serial stage calls. This should solve our issues the two definitions are elaborated serially, but not if they are elaborated in parallel.
* in Builder, error on multiple `ImportedDefinitionAnnotations` that are holding `Definition`s with the same name

mwachs5:
I chatted with @seldridge a bit here, and we are thinking we are going to make "Namespaces" a more first-class citizen sooner or later. But not sure it's what we want to solve in this PR

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -107,9 +108,33 @@ object Instance extends SourceInfoDoc {
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
   ): Instance[T] = {
+    // Check to see if the module is already defined internally or externally
+    val existingMod = Builder.components.map {
+      case c: DefModule if c.id == definition.proto          => Some(c)
+      case c: DefBlackBox if c.name == definition.proto.name => Some(c)
+      case _ => None
+    }.flatten
```
debs-sifive, line 116: Jack previously pointed out that this is a linear search. Would it be worthwhile to change `Builder.components` from an `ArrayBuffer` to a `HashMap` for better lookup speed?

azidar: I don't think its worthwhile unless you can measure the speedup; your lookup function would be complicated because you'd have to find the `DefBlackBox` case anyways.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -364,7 +364,23 @@ private[chisel3] class ChiselContext() {
 }
 
 private[chisel3] class DynamicContext(val annotationSeq: AnnotationSeq, val throwOnFirstError: Boolean) {
+  val importedDefinitionAnnos = annotationSeq.collect { case a: ImportedDefinitionAnnotation[_] => a }
+
+  // Ensure there are no repeated names for imported Definitions
+  val importedDefinitionNames = importedDefinitionAnnos.map { a => a.importedDefinition.proto.name }
+  if (importedDefinitionNames.distinct.length < importedDefinitionNames.length) {
+    throwException("Imported Definitions must have distinct names.")
```
mwachs5, line 372: It would be really helpful if we could show the ones that are NOT distinct... perhaps something like:

```suggestion
  if (importedDefinitionNames.distinct.length < importedDefinitionNames.length) {
    val duplicates = importedDefinitionNames.distinct.diff(importedDefinitionNames)
    throwException(s"Imported Definitions must have distinct names, the following are duplicates: $duplicates")
```

mwachs5 (commented):
looks good to me, slight suggestion for better failing require error messages!

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -107,9 +108,33 @@ object Instance extends SourceInfoDoc {
...
+
+    if (existingMod.isEmpty) {
+      // Add a Definition that will get emitted as an ExtModule so that FIRRTL
+      // does not complain about a missing element
+      class EmptyExtModule extends ExtModule {
+        override def generateComponent(): Option[Component] = {
+          require(!_closed, "Can't generate module more than once")
```
mwachs5, line 123: is there any way to say what module we are trying to generate more than once? can we print `desiredName` at this point?

mwachs5 (approved):

mwachs5:
Looks good! Do you want me to label with `please merge`?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -364,7 +364,24 @@ private[chisel3] class ChiselContext() {
...
 
 private[chisel3] class DynamicContext(val annotationSeq: AnnotationSeq, val throwOnFirstError: Boolean) {
+  val importedDefinitionAnnos = annotationSeq.collect { case a: ImportedDefinitionAnnotation[_] => a }
+
+  // Ensure there are no repeated names for imported Definitions
+  val importedDefinitionNames = importedDefinitionAnnos.map { a => a.importedDefinition.proto.name }
+  if (importedDefinitionNames.distinct.length < importedDefinitionNames.length) {
+    val duplicates = importedDefinitionNames.diff(importedDefinitionNames.distinct).mkString(", ")
+    throwException(s"Expected distinct imported Definition names but found duplicates for: $duplicates")
+  }
```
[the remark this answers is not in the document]
debs-sifive: Not sure what you mean by "conflict". The top-level module should be able to access the submodules uniquely, since the `Definition`s are now forced to have unique names.

The submodules also won't conflict in the `Builder`, since each `Definition` is built in a separate call to `ChiselStage`.

debs-sifive:
> Looks good! Do you want me to label with please merge?

Yes please! @azidar can you please mark the "changes requested" as resolved if your concerns have been addressed?

core/src/main/scala/chisel3/internal/Builder.scala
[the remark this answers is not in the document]
mwachs5: you could in theory pass in all the Modules as definitions from the first elaboration, then potential conflicts could be detected/avoided. If yo only pass in the top level definition you'd likely have issues in firrtl/verilog linking later which is still only hypothetical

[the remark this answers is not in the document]
azidar: Yes, I think the question is how we would have to do linking. @sequencer and @mwachs5  make good points, where we won't be able to link easily, without renaming conflict names when combining the Verilog. We could avoid this by adding submodule names to the duplicates.

This is probably the best intermediate solution, because it guarantees no name conflicts.

debs-sifive:
I'm changing this to a draft PR since, after some discussion, I think it makes more sense to solidify what we're doing with namespaces/prefix API and apply that to our multiple-definition-elaboration scheme.

Still working on test cases for submodule name conflicts; will push those soon!

src/test/scala/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+        // Tie inputs to a value so ChiselStage does not complain
+        inst0.in := 0.U
+        inst1.in := 0.U
+      }
+
+      val dutDef0_rtl = Source.fromFile(s"$testDir/dutDef0/AddTwoMixedModules.v").getLines.mkString
+      dutDef0_rtl should include("module AddOne(")
+      dutDef0_rtl should include("module AddTwoMixedModules(")
+      val dutDef1_rtl = Source.fromFile(s"$testDir/dutDef1/AddTwoMixedModules_1.v").getLines.mkString
+      dutDef1_rtl should include("module AddOne_2(")
```
debs-sifive, line 287: @azidar is this expected to skip to `AddOne_2` instead of `AddOne_1`?

mwachs5: @debs-sifive I think what is happening is `AddTwoMixedModules` made an `AddOne` (the Definition) and an `AddOne_1`(the Instance) in the first `dutDef0. They may get deduped later but as far as Chisel is concerned you have two names taken up in the namespace by the time you get to `dutDef1`

PR #2515, opened 2022-05-03
Revert #2496 

sequencer:
#2496 introduced a bug which was spotted by my daily CI in
https://github.com/sequencer/playground/runs/6243944547
An MVP should be:
```
RegEnable(next = true.B, init = true.B, enable = true.B)
```
won't compile, and Scala complains:
```
macro applications do not support named and/or default arguments
```
This was should blame to Scala Compiler in https://github.com/scala/scala/blob/21a56430f9902b5b8b81535fd76371caf23c0d10/src/compiler/scala/tools/nsc/typechecker/ContextErrors.scala#L630
which seems to be a limitation to Scala compiler(both 2.12 and 2.13) that the function which used macro cannot use default or named arguments.

Thus #2496 introduced a breaking change to related function calls. 

There exists three options to this:
1. Reverting #2496 to make downstream happy(keep compatibility but the detail source info is lost)
2. Regard this as a breaking change, user should bear with it.(we should add this to release note)
3. Refactoring `SourceInfo` and `CompileOption` macro to Scala Plugin, put it earlier than macro.(refactor a lot but user will be happier)

### Contributor Checklist

#### Type of Improvement
- bug fix

- [ ] Did you review?


jackkoenig:
There _might_ be a 4th option, we could maybe use [`@deprecatedName`](https://www.scala-lang.org/api/2.12.15/scala/deprecatedName.html) to sort of deprecate the use of by-name parameters on these methods.

We really want to add source locators if at all possible but if we can avoid a hard breakage (without a deprecation step) that would definitely be ideal.

ekiwi:
I am curious to hear your thoughts on this @jackkoenig : How hard / how useful would it be to implement source locator propagation through a compiler plugin instead of using macros?

sequencer:
[@deprecatedName](https://www.scala-lang.org/api/2.12.15/scala/deprecatedName.html) sounds to be a great idea!

mwachs5:
wait are we saying that the change to adding source locators means we can't pass named arguments to RegNext and friends? That seems sad in general...

sequencer:
Yes, it seems that Scala compiler doesn't like this, I have seen two downstreams beaten by this:(
https://github.com/chipsalliance/rocket-chip/pull/2986
https://github.com/sifive/sifive-blocks/pull/177

jackkoenig:
> I am curious to hear your thoughts on this @jackkoenig : How hard / how useful would it be to implement source locator propagation through a compiler plugin instead of using macros?

All compiler plugin stuff is difficult, we'd essentially be reimplementing a subset of what def macros do. It's probably doable, but I'm not actually convinced that named arguments would work if we did it in a plugin. Macros paradise was implemented as a plugin prior to Scala 2.13 and it has never supported named arguments so I suspect there's some fundamental limitation here.

jackkoenig:
> wait are we saying that the change to adding source locators means we can't pass named arguments to RegNext and friends? That seems sad in general...

@mwachs5 - Yes, it is a really annoying limitation but it is why Chisel3 moved to things like `UInt(8.W)` instead of `UInt(width = 8)` and `RegInit(0.U)` instead of `Reg(init = 0.U)`. It's a super annoying limitation.

PR #2527, opened 2022-05-13
Make Verification Statements (assert,assume) accept printable

### LINKED ISSUE #2350: Verification Statements do not support Printable

Currently the Verification Statements (i.e. `assert` and `assume`) can take additional arguments that will result in a message being printed when the condition is violated. However, it only supports the "C"-style printing, i.e.: `assert(a, "a=%d", a)`. In addition it would be nice if it also supported the Scala/Python style of printing like `printf` already does, i.e. `assert(a, p"a=$a")`.

girishpai:
Fix #2350
### Contributor Checklist

#### Type of Improvement

- new feature / API 
#### API Impact

Current API with strings should still work for Assert/Assume. The base case is made to work with printable (Like in Printf) and the current API is built on top of this. Call Printable.pack(String,data) before calling the new API for this case.

None

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -41,7 +41,16 @@ object assert {
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
-  ): Assert = macro _applyMacroWithMessage
+  ): Assert = macro _applyMacroWithStringMessage
+
+  def apply(
+    cond:    Bool,
+    message: Printable
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): Assert = macro _applyMacroWithPrintableMessage
+
   def apply(cond: Bool)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Assert =
     macro _applyMacroWithNoMessage
 
@@ -56,6 +65,10 @@ object assert {
 
   import VerificationStatement._
 
+  @deprecated(
+    "This method has been deprecated in favor of _applyMacroWithStringMessage. Please use the same.",
+    "Chisel 3.5"
+  )
   def _applyMacroWithMessage(
     c:              blackbox.Context
   )(cond:           c.Tree,
@@ -66,7 +79,32 @@ object assert {
   ): c.Tree = {
     import c.universe._
     val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
-    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message), ..$data)($sourceInfo, $compileOptions)"
+    q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithStringMessage(
+    c:              blackbox.Context
+  )(cond:           c.Tree,
+    message:        c.Tree,
+    data:           c.Tree*
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithPrintableMessage(
+    c:              blackbox.Context
+  )(cond:           c.Tree,
+    message:        c.Tree
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message))($sourceInfo, $compileOptions)"
   }
 
   def _applyMacroWithNoMessage(
@@ -84,15 +122,14 @@ object assert {
   def _applyWithSourceLine(
     cond:    Bool,
     line:    SourceLineInfo,
-    message: Option[String],
-    data:    Bits*
+    message: Option[Printable]
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
   ): Assert = {
     val id = new Assert()
     when(!Module.reset.asBool()) {
-      failureMessage("Assertion", line, cond, message, data)
+      failureMessage("Assertion", line, cond, message)
       Builder.pushCommand(Verification(id, Formal.Assert, sourceInfo, Module.clock.ref, cond.ref, ""))
     }
     id
@@ -127,7 +164,16 @@ object assume {
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
-  ): Assume = macro _applyMacroWithMessage
+  ): Assume = macro _applyMacroWithStringMessage
+
+  def apply(
+    cond:    Bool,
+    message: Printable
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): Assume = macro _applyMacroWithPrintableMessage
+
   def apply(cond: Bool)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Assume =
     macro _applyMacroWithNoMessage
 
@@ -142,6 +188,10 @@ object assume {
 
   import VerificationStatement._
 
+  @deprecated(
+    "This method has been deprecated in favor of _applyMacroWithStringMessage. Please use the same.",
+    "Chisel 3.5"
+  )
   def _applyMacroWithMessage(
     c:              blackbox.Context
   )(cond:           c.Tree,
@@ -152,7 +202,32 @@ object assume {
   ): c.Tree = {
     import c.universe._
     val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
-    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message), ..$data)($sourceInfo, $compileOptions)"
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some(_root_.chisel3.Printable.pack($message, ..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithStringMessage(
+    c:              blackbox.Context
+  )(cond:           c.Tree,
+    message:        c.Tree,
+    data:           c.Tree*
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some(_root_.chisel3.Printable.pack($message, ..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithPrintableMessage(
+    c:              blackbox.Context
+  )(cond:           c.Tree,
+    message:        c.Tree
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message))($sourceInfo, $compileOptions)"
   }
 
   def _applyMacroWithNoMessage(
@@ -170,15 +245,14 @@ object assume {
   def _applyWithSourceLine(
     cond:    Bool,
     line:    SourceLineInfo,
-    message: Option[String],
-    data:    Bits*
+    message: Option[Printable]
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
   ): Assume = {
     val id = new Assume()
     when(!Module.reset.asBool()) {
-      failureMessage("Assumption", line, cond, message, data)
+      failureMessage("Assumption", line, cond, message)
       Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, ""))
     }
     id
@@ -299,13 +373,11 @@ private object VerificationStatement {
     (p.source.file.name, p.line, p.lineContent.trim)
   }
 
-  // creates a printf to inform the user of a failed assertion or assumption
   def failureMessage(
     kind:     String,
     lineInfo: SourceLineInfo,
     cond:     Bool,
-    message:  Option[String],
-    data:     Seq[Bits]
+    message:  Option[Printable]
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
@@ -314,11 +386,11 @@ private object VerificationStatement {
     val lineMsg = s"$filename:$line $content".replaceAll("%", "%%")
     val fmt = message match {
       case Some(msg) =>
-        s"$kind failed: $msg\n    at $lineMsg\n"
-      case None => s"$kind failed\n    at $lineMsg\n"
+        p"$kind failed: $msg\n    at $lineMsg\n"
+      case None => p"$kind failed\n    at $lineMsg\n"
     }
     when(!cond) {
-      printf.printfWithoutReset(fmt, data: _*)
+      printf.printfWithoutReset(fmt)
     }
   }
 }
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -43,6 +43,25 @@ object assert {
     compileOptions:      CompileOptions
   ): Assert = macro _applyMacroWithStringMessage
 
+  /** Checks for a condition to be valid in the circuit at all times. If the
+    * condition evaluates to false, the circuit simulation stops with an error.
+    *
+    * Does not fire when in reset (defined as the encapsulating Module's
+    * reset). If your definition of reset is not the encapsulating Module's
+    * reset, you will need to gate this externally.
+    *
+    * May be called outside of a Module (like defined in a function), so
+    * functions using assert make the standard Module assumptions (single clock
+    * and single reset).
+    *
+    * @param cond condition, assertion fires (simulation fails) when false
+    * @param message optional chisel Printable type message
+    *
+    * @note See [[printf.apply(fmt:Printable)]] for documentation on printf using Printables
+    * @note currently cannot be used in core Chisel / libraries because macro
+    * defs need to be compiled first and the SBT project is not set up to do
+    * that
+    */
   def apply(
     cond:    Bool,
     message: Printable
@@ -78,10 +97,11 @@ object assert {
     compileOptions: c.Tree
   ): c.Tree = {
     import c.universe._
-    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
     q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
   }
 
+  /** Used by our macros. Do not call directly! */
   def _applyMacroWithStringMessage(
     c:              blackbox.Context
   )(cond:           c.Tree,
@@ -91,10 +111,11 @@ object assert {
     compileOptions: c.Tree
   ): c.Tree = {
     import c.universe._
-    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
     q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
   }
 
+  /** Used by our macros. Do not call directly! */
   def _applyMacroWithPrintableMessage(
     c:              blackbox.Context
   )(cond:           c.Tree,
@@ -103,7 +124,7 @@ object assert {
     compileOptions: c.Tree
   ): c.Tree = {
     import c.universe._
-    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
     q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message))($sourceInfo, $compileOptions)"
   }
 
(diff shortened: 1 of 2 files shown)


girishpai:
@jackkoenig  please note -  have not really added many unit tests - except two of them within the current AssertSpec. Can add more unit tests if you feel there is a need.

mwachs5:
Should we add something to the mdocs about this? just a note "you can use these in your `assert(...)` and `assume(...)` statements too!
https://www.chisel-lang.org/chisel3/docs/explanations/printing.html
I suppose it would have been better if there was an actual doc about asserts and assumes though so maybe we should save that for a followup

girishpai:
> Should we add something to the mdocs about this? just a note "you can use these in your `assert(...)` and `assume(...)` statements too! https://www.chisel-lang.org/chisel3/docs/explanations/printing.html I suppose it would have been better if there was an actual doc about asserts and assumes though so maybe we should save that for a followup

Sure @mwachs5 ! Do I need to know anything specific about updating mdocs? Never done that (or maybe I have forgotten?)

mwachs5:
You can check out the documentation docs here: https://github.com/chipsalliance/chisel3/tree/master/docs#mdoc

The source for example the file I linked above is here:
https://github.com/chipsalliance/chisel3/blob/master/docs/src/explanations/printing.md

girishpai:
> You can check out the documentation docs here: https://github.com/chipsalliance/chisel3/tree/master/docs#mdoc
> 
> The source for example the file I linked above is here: https://github.com/chipsalliance/chisel3/blob/master/docs/src/explanations/printing.md
> 
> Note that particular one doesnt use much `mdoc` functionality but some of the other ones actually compile-check, and some even emit verilog / error messages

Cool - but since as you mentioned we do not have anything about asserts / assumes to begin with - let us do that as a follow up!

sequencer:
See https://github.com/chipsalliance/chisel3/pull/1968#issuecomment-865336046, maybe this needs some implementation decision from @debs-sifive

girishpai:
> See [#1968 (comment)](https://github.com/chipsalliance/chisel3/pull/1968#issuecomment-865336046), maybe this needs some implementation decision from @debs-sifive

I will let @jackkoenig and @debs-sifive  chime in here. As this PR indicates - adding Printable support is reasonably easy and straightforward - not sure how much of an effort the suggestion within https://github.com/chipsalliance/chisel3/pull/1968#issuecomment-865336046 is.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -66,7 +79,32 @@ object assert {
   ): c.Tree = {
     import c.universe._
     val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
-    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message), ..$data)($sourceInfo, $compileOptions)"
+    q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithStringMessage(
```
jackkoenig, line 85:
```suggestion
  private def _applyMacroWithStringMessage(
```
What happens if we make this private (or package private)? I'm not certain that we can but it's worth trying.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -84,15 +122,14 @@ object assert {
   def _applyWithSourceLine(
     cond:    Bool,
     line:    SourceLineInfo,
-    message: Option[String],
-    data:    Bits*
+    message: Option[Printable]
   )(
```
jackkoenig, line 126: This is a binary incompatible change (because this method is injected into user code by our macros). Thus we need to keep around the old version (not deprecated, just marked with a comment to delete in 3.6).

You should be able to overload it though, both versions can have the same name.

girishpai:
> > I will let @jackkoenig and @debs-sifive chime in here. As this PR indicates - adding Printable support is reasonably easy and straightforward - not sure how much of an effort the suggestion within [#1968 (comment)](https://github.com/chipsalliance/chisel3/pull/1968#issuecomment-865336046) is.
> 
> From what I understand this PR just adds support for `p"{wire}"` style printing in addition to `"%d", wire` and that is awesome. The thing that @debs-sifive mentioned can be added in the background. It would just involve adding an annotation from chisel that points to both, the `printf` and the verification statement generated. This is orthogonal to this PR.

Just to add since this accepts any printable - once this other enhancement PR goes in https://github.com/chipsalliance/chisel3/pull/2528 will be able to do
` cf"Chisel Bits Foo = $foo%x  Scala Float = $scalafloatVal%2.2f"`   as well!

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -127,7 +164,16 @@ object assume {
...
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
-  ): Assume = macro _applyMacroWithMessage
+  ): Assume = macro _applyMacroWithStringMessage
+
+  def apply(
```
jackkoenig, line 169: Needs ScalaDoc, see the format String apply above for an example.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -152,7 +202,32 @@ object assume {
   ): c.Tree = {
     import c.universe._
     val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
-    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message), ..$data)($sourceInfo, $compileOptions)"
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some(_root_.chisel3.Printable.pack($message, ..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithStringMessage(
```
jackkoenig, line 208: Same comment as above seeing if this can be private (or at least package private).

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -152,7 +202,32 @@ object assume {
...
+    data:           c.Tree*
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some(_root_.chisel3.Printable.pack($message, ..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithPrintableMessage(
```
jackkoenig, line 221: Same comment as above seeing if this can be private (or at least package private).

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -66,7 +79,32 @@ object assert {
...
+    data:           c.Tree*
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
+    q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  def _applyMacroWithPrintableMessage(
```
jackkoenig, line 98: Same comment as above seeing if this can be private (or at least package private).

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -170,15 +245,14 @@ object assume {
   def _applyWithSourceLine(
```
jackkoenig, line 245: Same comment about binary compatibility

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -41,7 +41,16 @@ object assert {
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
-  ): Assert = macro _applyMacroWithMessage
+  ): Assert = macro _applyMacroWithStringMessage
+
+  def apply(
```
jackkoenig, line 46: This needs ScalaDoc

jackkoenig (changes requested):
Generally looks great, some comments about trying to make methods private, binary compatibility, and need for ScalaDocs

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Good point - let me give this a shot!

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Agreed!

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: @jackkoenig  - sadly this gives an error
`VerificationStatement.scala:44:21: macro implementation must be public
[error]   ): Assert = macro _applyMacroWithStringMessage`

girishpai: Let me try package private ....

girishpai: Same error - this is what I tried
`private[chisel3] def _applyMacroWithStringMessage(`

girishpai: So I guess we will have to live with this being visible - and just probably add a WARNING comment to not use this directly.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: @jackkoenig  tried this 
```def _applyWithSourceLine(
    cond:    Bool,
    line:    SourceLineInfo,
    message: Option[String],
    data:    Bits*
  )(
  ): Assert = {
    val id = new Assert()
    when(!Module.reset.asBool()) {
      failureMessage("Assertion", line, cond,message.map(Printable.pack(_,data : _*))) 
      Builder.pushCommand(Verification(id, Formal.Assert, sourceInfo, Module.clock.ref, cond.ref, ""))
    }
    id
  }
```
And now compile fails telling there is ambiguity - I understand why but yet to figure out a solution. Still digging - but putting it out here. 

```
ambiguous reference to overloaded definition,
[error] both method _applyWithSourceLine in object assert of type (cond: chisel3.Bool, line: chisel3.VerificationStatement.SourceLineInfo, message: Option[chisel3.Printable])(implicit sourceInfo: chisel3.internal.sourceinfo.SourceInfo, implicit compileOptions: chisel3.CompileOptions)chisel3.assert.Assert
[error] and  method _applyWithSourceLine in object assert of type (cond: chisel3.Bool, line: chisel3.VerificationStatement.SourceLineInfo, message: Option[String], data: chisel3.Bits*)(implicit sourceInfo: chisel3.internal.sourceinfo.SourceInfo, implicit compileOptions: chisel3.CompileOptions)chisel3.assert.Assert
[error] match argument types (chisel3.Bool,(String, Int, String),None.type)
[error]   assert((a.andR) === (s"%${w}s".format(BigInt(_a).toString(2)).foldLeft(true)(_ && _ == '1')).asBool) 
```

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Same as above - unfortunately not possible - gives compile error. So for now will just add an explicit message not to use this directly.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Same as above - unfortunately not possible - gives compile error. So for now will just add an explicit message not to use this directly.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Same as above - unfortunately not possible - gives compile error. So for now will just add an explicit message not to use this directly.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: At worse I need to overload the failureMessage! Initially my intention was to reduce duplication as much as possible -since I was not aware of the additional constraint of binary compatibility. 

@jackkoenig  feel free to suggest any other way out of this - otherwise I will go ahead with the obvious solution

girishpai: Ok I take back what I said - basically I need to just explicitly change the name of my new method (append Printable or something) and we should be good! Trying ....

girishpai: Still getting same error - need to take a break from screen and get back to this - not thinking clearly I guess!

girishpai: Ok - that solution of renaming my new _appltSourceLinePrintable  to append with Printable worked - so nothing within this file references the old _applySourceLine method. 

```
/** This will be removed in Chisel 3.6 in favor of the Printable version */
  def _applyWithSourceLine(
    cond:    Bool,
    line:    SourceLineInfo,
    message: Option[String],
    data:    Bits*
  )(
  ): Assume = {
    val id = new Assume()
    when(!Module.reset.asBool()) {
      failureMessage("Assumption", line, cond, message.map(Printable.pack(_,data: _*)))
      Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, ""))
    }
    id
  }

  /** Used by our macros. Do not call directly! */
  def _applyWithSourceLinePrintable(
    cond:    Bool,
    line:    SourceLineInfo,
    message: Option[Printable]
  )(
  ): Assume = {
    val id = new Assume()
    when(!Module.reset.asBool()) {
      failureMessage("Assumption", line, cond, message)
      Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, ""))
    }
    id
  }


```
Still curious why the previous version did not work.

girishpai: oh so basically the optionality of data is what causes the ambiguity - despite two different Option types - None is common to both. 

```
def _applyWithSourceLine(
    cond:    Bool,
    line:    SourceLineInfo,
    message: Option[String],
    data:    Bits*
  )(
```

girishpai: @jackkoenig  sadly looks like we might have to live with the uglier solution since using the same name does not work (feel free to suggest alternate solutions)

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added - used the one present already as a template and made slight modifications! Hope that's fine.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Done - due to some compile issues could not use the same name (basically overloading failed) - hence had to rename the new one to be different.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added a comment indicating not to use this directly and only for use within the macros.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added a comment indicating not to use this directly and only for use within the macros.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added a comment indicating not to use this directly and only for use within the macros.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added - used the one present already as a template and made slight modifications! Hope that's fine.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Went ahead with this change to rename the new version with Printable appended and keeping the old one.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: Added a comment indicating not to use this directly and only for use within the macros.

girishpai:
@jackkoenig  incorporated most of your feedback - except the one to make these new macro definitions private - compiler complained and hence have just added a comment on top of each of them - explicitly asking not to use directly.

core/src/main/scala/chisel3/VerificationStatement.scala
jackkoenig: Yeah we just have to use 2 names. Due to type erasure, `Option[String]` and `Option[Printable]` are actually the same type at runtime so it does make sense that we can't overload here.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -299,13 +449,11 @@ private object VerificationStatement {
...
   }
 
-  // creates a printf to inform the user of a failed assertion or assumption
   def failureMessage(
     kind:     String,
     lineInfo: SourceLineInfo,
     cond:     Bool,
-    message:  Option[String],
-    data:     Seq[Bits]
+    message:  Option[Printable]
```
jackkoenig, line 456: This is a binary incompatible change, you need the old one (deprecate it)

jackkoenig: Actually this is part of a private object, it will show up as a binary incompatibility but since users can't call it, it doesn't matter.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -65,8 +97,35 @@ object assert {
     compileOptions: c.Tree
   ): c.Tree = {
     import c.universe._
-    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLine"))
-    q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some($message), ..$data)($sourceInfo, $compileOptions)"
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
+    q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
+  }
+
+  /** Used by our macros. Do not call directly! */
```
mwachs5, line 104: You could use the same or create a new  `@group` to make this show up in a 'Here be dragons" section of the API docs:

```
```

This renders in the API docs like so:

Screen Shot 2022-05-13 at 12 52 18 PM

https://github.com/chipsalliance/chisel3/blob/83eeda2efebb1e3e3d66365d388f5de627766d47/macros/src/main/scala/chisel3/SourceInfoDoc.scala#L22

src/test/scala/chiselTests/Assert.scala
```diff
@@ -61,6 +61,17 @@ class BadUnescapedPercentAssertTester extends BasicTester {
   stop()
 }
```
jackkoenig, line 63: It would be helpful to add some tests for `assume` as well.

src/test/scala/chiselTests/Assert.scala
```diff
@@ -61,6 +61,17 @@ class BadUnescapedPercentAssertTester extends BasicTester {
...
 
+class PrintableFormattedAssertTester extends BasicTester {
+  val foobar = Wire(UInt(32.W))
+  foobar := 123.U
+  assert(foobar === 123.U, p"Error! Wire foobar =/= ${Hexadecimal(foobar)} This is 100%% wrong.\n")
+  stop()
+}
+
+class PrintableBadUnescapedPercentAssertTester extends BasicTester {
+  assert(1.U === 1.U, p"I'm 110% sure this is an invalid message")
```
jackkoenig, line 72: Actually this is valid, this is part of the reason we need to add the new String interpolator (ie. your `cf` PR), because it's perfectly valid to have `%` in a `p"..."` (eg. https://scastie.scala-lang.org/HdvWBGEERamX0v7yFc9CSw)

core/src/main/scala/chisel3/VerificationStatement.scala
jackkoenig: Yeah this would be a good thing to do.

jackkoenig: Actually, this gave me an idea, we should create `object internal` in each of these objects and then put all of the `_` methods in there. That way it's clear that there's are internal things that users shouldn't use.

The current ScalaDoc for `object assert` is a mess:
Screen Shot 2022-05-13 at 1 16 02 PM

Putting the internal methods in `object internal` helps a ton:

Screen Shot 2022-05-13 at 1 16 39 PM

Since it would be binary incompatible to just move them all, we may still want to do the grouping though because we'll need to deprecate a few of the methods before moving them.

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: So @jackkoenig  we are good with what we have? No need to add the old one right?

jackkoenig: Correct, sorry for the confusion. I had checked binary compatibility and it flagged this, but there are things in Scala that have to be implemented in Java as public APIs even though there private from Scala's perspective and are thus "false positives" if that makes sense.

src/test/scala/chiselTests/Assert.scala
girishpai: @jackkoenig  cool - let me create a new Assume.scala file and add to that one! 

Also - do I need to do anything different to add these tests in CI or will it automatically get picked up?

src/test/scala/chiselTests/Assert.scala
girishpai: Oh ok - I did not pay attention to this since I just merely copy-pasted what we had before and replaced the message with p".." - wonder how is the test passing then -  I think it is expecting a failure of some sorts

src/test/scala/chiselTests/Assert.scala
mwachs5: they will get picked up as long as they are in the same file and extend the scala test as the others do

core/src/main/scala/chisel3/VerificationStatement.scala
girishpai: @mwachs5  @jackkoenig   - thanks for the pointers / tips - but just for clarity what do you feel is a MUST for this PR? Obviously long term we should incorporate all of these suggestions!

mwachs5: i think we can easily add a @group to this PR

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -56,6 +84,10 @@ object assert {
 
   import VerificationStatement._
 
+  @deprecated(
+    "This method has been deprecated in favor of _applyMacroWithStringMessage. Please use the same.",
+    "Chisel 3.5"
+  )
```
sequencer, line 90: This is not a public API? can we deprecate it?

jackkoenig: It is deprecated, or are you saying deprecate and just say "do not use"?

sequencer: From my understanding, it like any other Macro transformed APIs, they are not supported being called by users(but still public) since this PR removed APIs which used this macro, I thought it can be safely removed(rather deprecated)

src/test/scala/chiselTests/Assert.scala
girishpai: @jackkoenig  @mwachs5 created a spec for Assume similar to that for Assert (in fact first version is just a copy and replace assert with assume - since as per spec in simulation they should be identical). 

But looks like for some reason `assume(false.B)` does not trigger a failure while `assert(false.B)` does. 

This warning could potentially be the reason 
```
Warning: 1 verification statements (assert, assume or cover) were removed 

when compiling to Verilog because the 

basic Verilog standard does not support them.

 If this was not intended, compile to System Verilog instead using the `-X sverilog` compiler flag.
```

core/src/main/scala/chisel3/VerificationStatement.scala
jackkoenig: So while they absolutely shouldn't, users could technically define their own methods that [macro] call these: https://scastie.scala-lang.org/MmzrWNwvSyasKAUJPOIzgw. In any case I think we should deprecate on minor version and remove in 3.6.

src/test/scala/chiselTests/Assert.scala
mwachs5: Well i guess for the assume test we need to emit to SV instead of verilog?

girishpai: Yes - confirmed that it is indeed the reason by inspecting the verilog. Let me try changing the compiler first if that is straightforward - else will change the test to just inspect the generated FIRRTL/verilog

@jackkoenig - feel free to share specific pointers to change the compiler (like by passing an annotation or something).

girishpai: @mwachs5  - not sure if your suggestion is same as that of @jackkoenig - I am assuming  it is! (Pun intended)

jackkoenig: @girishpai it specifies the compiler here: https://github.com/chipsalliance/chisel3/blob/2b48fd15a7711dcd44334fbbc538667a102a581a/src/main/scala/chisel3/testers/TesterDriver.scala#L57

Note that changing this will have pretty wide ranging impacts so might affect lots of other tests. It's probably better to do something that just checks the emitted FIRRTL like so: https://github.com/chipsalliance/chisel3/blob/6d4e6edc9a62c166af95b41b6aea9634d34275ed/src/test/scala/chiselTests/BulkConnectSpec.scala#L20

girishpai: @jackkoenig  thanks! Most likely will go with inspecting FIRRTL - but for future purposes might be a good idea to provide support to override the Emitter at the Spec/test level.

core/src/main/scala/chisel3/VerificationStatement.scala
sequencer: That's quite reasonable, although I doubt if there exists any user doing so. Thanks for the explanation! Let's go through the deprecation flow ;p

jackkoenig:
Replaced by https://github.com/chipsalliance/chisel3/pull/2663

PR #2528, opened 2022-05-13
Add formatted Printable interpolator `cf`

girishpai:
### Contributor Checklist

#### Type of Improvement

- new feature / API

#### Desired Merge Strategy

- 
#### Release Notes

F interpolator for Chisel (cf) analogous to the one available within Scala 

- [x] Did you review?


girishpai:
@jackkoenig  - still working on adding more tests (especially negative tests) - but was hoping to start getting feedback on the main code.

girishpai:
There is definite scope for refactoring the existing PrintableSpec.scala - like the checking part - which has a lot of common things (thanks to @debs-sifive  for pointing that out). As of now I have just reused the same pattern as that of p"" within that file. 

To keep things focussed - I propose on doing the actual refactoring in a future PR and for this PR just ensure all interesting cases are covered based on new unit tests using the old template.

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+     *   3. If format specifers target regular scala data types - java.lang.String.format method is called with specifier and argument to get the formatted string.
+     *
+     * Yet to document all the corener cases which this code tries to cover - but here are few of them
+     * 1. Allow calling %f or %2.2f (or something similar) on Int type data (Int, Short, Long) - this requires doing an explicit cast of the data before calling String.format
+     * 2. For bits - if no format specifier given (%s i.e) call .toPrintable to pick the default one. For regular data - let the String.format handle canonical case of %s too.
+     * Still testing this code - I am pretty sure this can be refactored and made more efficient / readable. WIP.
+     *
+     * /
+     */
+     */
```
jackkoenig, line 260: Before merging, let's make this ScalaDoc more like the one defined on[ Scala's `f"..."` interpolator](https://www.scala-lang.org/api/2.13.8/scala/StringContext.html#f[A%3E:Any](args:A*):String).

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+    def cf(args: Any*): Printable = {
```
jackkoenig, line 261: This is very bikesheddy, but I think we should figure out what we want to call this. The current name `cf"..."` has the nice quality of basically saying "Chisel format" (and sort of refers to the relationship with `f"..."`). Alternatively, we could just call it `c"..."` for "Chisel". Regardless, I think this sort of replaces `p"..."`, maybe we should consider deprecating it in favor of this one.

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +261,208 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+    val firrtl = ChiselStage.emitChirrtl(new MyModule)
+    getPrintfs(firrtl) match {
+      case Seq(Printf("myInt = 1234", Seq())) =>
+      case e                                  => fail()
+    }
+  }
+  it should "generate proper printf for simple Decimal printing with cf format specifier" in {
+    class MyModule extends BasicTester {
+      val myWire = WireDefault(1234.U)
+      printf(cf"myWire = ${Decimal(myWire)}")
```
jackkoenig, line 290:
```suggestion
      printf(cf"myWire = $myWire%d")
```
The purpose of `cf` is to support this style 🙂

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +261,208 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+    class MyBundle extends Bundle {
+      val foo = UInt(32.W)
+    }
+    class MyModule extends BasicTester {
+      override def desiredName: String = "MyModule"
+      val myWire = Wire(new MyBundle)
+      val myInst = Module(new MySubModule)
+      printf(cf"${Name(myWire.foo)}")
+      printf(cf"${FullName(myWire.foo)}")
+      printf(cf"${FullName(myInst.io.fizz)}")
```
jackkoenig, line 335:
```suggestion
      printf(cf"${myWire.foo}%n")
      printf(cf"${myWire.foo}%N")
      printf(cf"${myInst.io.fizz}%N")
```

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +261,208 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+      val l1: Long = 253
+      printf(cf"F1 = $f1 D1 = $i1%2.4f F1 formatted = $f1%2.2f s1 = $s1%2.2f l1 = $l1%3.6f")
+
+    }
+    val firrtl = ChiselStage.emitChirrtl(new MyModule)
+    getPrintfs(firrtl) match {
+      case Seq(Printf("F1 = 20.45156 D1 = 10.0000 F1 formatted = 20.45 s1 = 15.00 l1 = 253.000000", Seq())) =>
+      case e                                                                                                => { println("e = ", e); fail() }
+    }
+  }
```
jackkoenig, line 415: I'm a little confused why you're only checkin %f, what about other format specifiers like `%d` and `%x`?

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+          // Example cf"This is $val%2.2f here" - parts - Seq("This is ","%2.2f here") - the format specifier here is %2.2f.
+          val idx_of_fmt_str =
+            if (
+              !part.isEmpty() && part.charAt(0) == '%' && (part.size == 1 || (part.size >= 2 && part.charAt(1) != '%'))
+            ) part.indexWhere { _.isLetter }
+            else -1
+
+          // If no format specifier - pick default - %s
+
+          val fmt = if (idx_of_fmt_str >= 0) part.substring(0, idx_of_fmt_str + 1) else "%s"
```
jackkoenig, line 285: I don't think it's right to use `%s` as some "default" format-specifier. `%s` means "format as String" which is actually not the same thing as "format as Printable" which is how it seems to be used below. I think you should make `fmt` be `Option[String]` where `None` is the default (no specifier provided) and only the `None` case is `.toPrintable`.

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+              val castedT = (fmt.takeRight(1), t) match {
+                case ("f", v: Int) => v.asInstanceOf[Double]
+                case ("f", v: Short) => v.asInstanceOf[Double]
+                case ("f", v: Byte) => v.asInstanceOf[Double]
+                case ("f", v: Long) => v.asInstanceOf[Double]
+                case (_, t) => t
+              }
+
+              PString(String.format(fmt, castedT.asInstanceOf[AnyRef]))
+            }
```
jackkoenig, line 329:
```suggestion
            case t => PString(fmt.format(t))
```
I don't really understand this, I suspect that the above should work and support other cases (eg. `%d`) that you're not handling here.

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+              fmt match {
+                case "%n" => Name(d)
+                case "%N" => FullName(d)
+                case "%s" => d.toPrintable
+                case x => {
+                  val msg = s"Illegal format specifier '$x'!\n"
+                  throw new UnknownFormatConversionException(msg)
+                }
+              }
+            }
```
jackkoenig, line 311:
```suggestion
            case d: Data => Printable.pack(fmt, d)
```
Please correct me if I'm wrong, but I think `Printable.pack` already handles everything here (with the exception of `%s` which I think should be implemented as `.toString`.

jackkoenig: The only caveat is that delegating individual pieces to `Printable.pack` will give a better error message than this PR currently does, but a worse message than it could. A more ideal version of this PR would be to have `Printable.pack` delegate to this function (ie. I'm suggesting moving much of the `Printable.pack` logic here (including it's nice error messages with carats) and then changing `Printable.pack` to just construct a `StringContext` and call this method.

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+                case ("f", v: Long) => v.asInstanceOf[Double]
+                case (_, t) => t
+              }
+
+              PString(String.format(fmt, castedT.asInstanceOf[AnyRef]))
+            }
+          }
+
+          // Remove format specifier from parts string
+          val modP = part.zipWithIndex.filter { _._2 > idx_of_fmt_str }.map { _._1 }.mkString
```
jackkoenig, line 333:
```suggestion
          val modP = part.zipWithIndex.collect { case (p, idx) if idx > idx_of_fmt_str => p }.mkString
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -234,6 +235,117 @@ package object chisel3 {
...
+
+          (modP, Some(fmtArgs))
+        }
+      }
+      // Combine the 1st part with the rest of the modified (format specifier removed) parts
+      val combParts = parts(0) +: partsAndSpecifierSeq.map { _._1 }
```
jackkoenig, line 339:
```suggestion
      val combParts = parts(0) +: partsAndSpecifierSeq.map(_._1)
```
Just FYI, Scala style is to generally use `(...)` unless you need `{ ... }` which is only in cases when you need a `{ case ... => }` or when you need multiple statements in your anonymous function.

src/test/scala/chiselTests/PrintableSpec.scala
jackkoenig: Also need to test things like `$mywire%x`, and I think that the the current implementation doesn't work if you have normal Strings after the modifier, eg. `printf(cf"myWire = $myWire%d!!!\n)`

jackkoenig (changes requested):
Good start, needs some iteration.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: As discussed - these set of tests were copied over from "p" to ensure cf is indeed a superset.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: As discussed - have a version of this in one the tests below - this particular test is the "p" version - just replacing with cf. 

That being said - still more tests will be added to cover additional cases.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: Part of the grander test plan :) - Started with known challenging cases which I had to deal with during development.

core/src/main/scala/chisel3/package.scala
girishpai: Agreed - picked %s as default just based on what is done for normal scala f interpolator.

core/src/main/scala/chisel3/package.scala
girishpai: Somehow the String.format does not accept ints being passed when doing %f - expects only double or float types. 

As discussed - will remove all of this and let the user cast if needed.

core/src/main/scala/chisel3/package.scala
girishpai: @jackkoenig - my WIP latest commit - does this - please take a look (and we can discuss this more later) - basically the only thing which needs to be done here is handle % (make it %% - since the "cf" API relies on the user to do this explicitly).

core/src/main/scala/chisel3/package.scala
girishpai: @jackkoenig  my latest commit does this - for now named it packCF - also a known limitation is handling %% - WIP.... do take a look and give feedback!

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -19,6 +19,7 @@ private[chisel3] object Converter {
       val (fmts, args) = pables.map(p => unpack(p, ctx)).unzip
       (fmts.mkString, args.flatten.toSeq)
     case PString(str) => (str.replaceAll("%", "%%"), List.empty)
+    case PStringNew(str) => (str, List.empty)
```
girishpai, line 22: This is needed because for cf API we do not need to do this - since the user / caller has to explicitly handle this. The previous "p" API  it was not done by user.

core/src/main/scala/chisel3/package.scala
girishpai: Also due to the fact that escaping % needs to be explicitly done by user for "cf" API - that part is not needed to be done explicitly when emittting firrtl.

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
girishpai: This is no longer needed as we handle literal percent by emitting `Percent` type in PercentSlitter function.

core/src/main/scala/chisel3/package.scala
girishpai: Delegated Printable.pack to use the cf API - retained all of the error reporting with the previous version and went one step ahead - also report unescaped literal % if found.

core/src/main/scala/chisel3/package.scala
girishpai: All of this casting is removed and now we do a simple call to .format.

core/src/main/scala/chisel3/package.scala
girishpai: Converted to Option[String] and used .toPrintable for Printable and %s for regular types.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: New test cases being developed - please do not review the spec file just as yet.

core/src/main/scala/chisel3/package.scala
girishpai: p calls the cf interpolator now.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: Working on refactoring / adding more tests - please review this file later.

girishpai: Continuing to add more tests - but at least this basic scenario seems to work - including strings right after %x of the wire. @jackkoenig  will continue testing - but wanted to know intuitively why you felt this would not work - since high level algorithm has not changed since your last comment on this.

girishpai: Checked in the basic scenario 

```
it should "correctly print strings after modifier" in {
    class MyModule extends BasicTester {
      val b1 = 10.U
      printf(cf"This is here $b1%x!!!! And should print everything else")
    }
    generateAndCheck(new MyModule,Seq(Printf("This is here %x!!!! And should print everything else",Seq("UInt<4>(\"ha\")"))))

  }
```

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: This scenario is already added in the spec file - marking resolved

girishpai: https://github.com/chipsalliance/chisel3/blob/c3b1412e0932c620dcb6cce38bbe6988673fc355/src/test/scala/chiselTests/PrintableSpec.scala#L284

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+
+      sc.checkLengths(args) // Enforce sc.parts.size == pargs.size + 1
       val parts = sc.parts.map(StringContext.treatEscapes)
-      // Zip sc.parts and pargs together ito flat Seq
-      // eg. Seq(sc.parts(0), pargs(0), sc.parts(1), pargs(1), ...)
+
+      // The 1st part is assumed never to contain a format specifier.
+      // If the 1st part of a string is an argument - then the 1st part will be an empty String.
+      // So we need to parse parts following the 1st one to get the format specifiers if any
+      val partsAfterFirst = parts.slice(1, parts.size)
```
jackkoenig, line 288:
```suggestion
      val partsAfterFirst = parts.tail
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+                case Some(x) => {
+                  val msg = s"Illegal format specifier '$x' for Chisel Data type!\n"
+                  throw new UnknownFormatConversionException(msg)
+                }
+                case None => d.toPrintable
+              }
+            }
+            case p: Printable => {
+              fmt match {
+                case Some("%s") => p.toString()
```
jackkoenig, line 320:
```suggestion
                case Some("%s") => PString(p.toString)
```
Do not include `()` on nullary methods that don't have side effects. This prints a warning (which is easy to miss in our sea of warnings).

Also better to not rely on the implicit conversion from `String` to `Printable`.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+            ) part.indexWhere { _.isLetter }
+            else -1
+
+          val fmt = if (idx_of_fmt_str >= 0) Some(part.substring(0, idx_of_fmt_str + 1)) else None
+          val fmtArgs: Printable = arg match {
+            case d: Data => {
+              fmt match {
+                case Some("%n") => Name(d)
+                case Some("%N") => FullName(d)
+                case Some("%s") => d.toString
```
jackkoenig, line 309:
```suggestion
                case Some("%s") => PString(d.toString)
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+                  throw new UnknownFormatConversionException(msg)
+                }
+                case None => p 
+              }
+            }
+
+            // Generic case - use String.format (for example %d,%2.2f etc on regular Scala types)
+            case t => {
+              PString(fmt.getOrElse("%s").format(t))
+            }
```
jackkoenig, line 332:
```suggestion
            case t =>  PString(t.toString)
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
-      // eg. Seq(sc.parts(0), pargs(0), sc.parts(1), pargs(1), ...)
+
+      // The 1st part is assumed never to contain a format specifier.
+      // If the 1st part of a string is an argument - then the 1st part will be an empty String.
+      // So we need to parse parts following the 1st one to get the format specifiers if any
+      val partsAfterFirst = parts.slice(1, parts.size)
+
+      // Align parts to their potential specifiers
+      val partsAndSpecifierSeq = partsAfterFirst.zip(args).map {
+        case (part, arg) => {
```
jackkoenig, line 292:
```suggestion
      val partsAndSpecifierSeq = partsAfterFirst.zip(args).map { case (part, arg) =>
```
There's no need for the extra indentation and `{`

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+          // Check if part starts with a format specifier (with % - disambiguate with literal % checking the next character if needed to be %)
+          // In the case of %f specifier there is a chance that we need more information - so capture till the 1st letter (a-zA-Z).
+          // Example cf"This is $val%2.2f here" - parts - Seq("This is ","%2.2f here") - the format specifier here is %2.2f.
+          val idx_of_fmt_str =
+            if (
+              !part.isEmpty() && part.charAt(0) == '%' && (part.size == 1 || (part.size >= 2 && part.charAt(1) != '%'))
+            ) part.indexWhere { _.isLetter }
+            else -1
+
+          val fmt = if (idx_of_fmt_str >= 0) Some(part.substring(0, idx_of_fmt_str + 1)) else None
```
jackkoenig, line 303:
```suggestion
  // Somewhere up above
  def extractFormatSpecifier(s: String): (Option[String], String)

  // ...

          val (fmt, rest) = extractFormatSpecifier(part) 
```

This would be much better factored into a function defined up above. What we're really doing is taking a String that might have a format specifier and parsing it out.

This would replace the need for `modP` below.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+          else {
+            iter += 1
+          }
+        }
+
+        // Handle the tail
+        if(curr_start < iter) {
+          buf += Some(PString(s.substring(curr_start,iter)))
+        }
+        buf.toSeq
```
jackkoenig, line 279:
```suggestion
      def percentSplitter(s : String): Seq[Printable] = {
        val pieces = s.split("%%").toList.flatMap { p =>
          if (p.contains('%')) throw new UnknownFormatConversionException("Un-escaped % found")
          // Wrap in PString and intersperse the escaped percentages
          Seq(Percent, PString(p))
        }
        pieces.tail // Don't forget to drop the extra percent we put at the beginning
```
This can be a lot simpler unless I'm missing something. Also I'm not sure why you're using `Option[Printable]` when you only have have `Some`. Also, method names should start with lower case letters

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+            // Generic case - use String.format (for example %d,%2.2f etc on regular Scala types)
+            case t => {
+              PString(fmt.getOrElse("%s").format(t))
+            }
+          }
+
+          // Remove format specifier from parts string
+          val modP = part.zipWithIndex.collect { case (p, idx) if idx > idx_of_fmt_str => p }.mkString
+
+          (modP, Some(fmtArgs))
```
jackkoenig, line 338: Why are you wrapping in `Some` here since there is no alternative path?

jackkoenig:
```suggestion
          fmtArg +: percentSplitter(modP)
```
Also, we can get rid of most of the logic below by just building our `Seq[Printable]` here and using a `flatMap` up above:

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+          // In the case of %f specifier there is a chance that we need more information - so capture till the 1st letter (a-zA-Z).
+          // Example cf"This is $val%2.2f here" - parts - Seq("This is ","%2.2f here") - the format specifier here is %2.2f.
+          val idx_of_fmt_str =
+            if (
+              !part.isEmpty() && part.charAt(0) == '%' && (part.size == 1 || (part.size >= 2 && part.charAt(1) != '%'))
+            ) part.indexWhere { _.isLetter }
+            else -1
+
+          val fmt = if (idx_of_fmt_str >= 0) Some(part.substring(0, idx_of_fmt_str + 1)) else None
+          val fmtArgs: Printable = arg match {
```
jackkoenig, line 304:
```suggestion
          val fmtArg: Printable = arg match {
```
Why plural? It's just the Printable form of the `arg` right?

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,26 +211,140 @@ package object chisel3 {
...
+
+      val pargsPables: Seq[Option[Printable]] = partsAndSpecifierSeq.map { _._2 }
       val seq = for { // append None because sc.parts.size == pargs.size + 1
-        (literal, arg) <- parts.zip(pargs :+ None)
-        optPable <- Seq(Some(PString(literal)), arg)
+        (literal, arg) <- combParts.zip(pargsPables :+ None)
+        optPable <- PercentSplitter(literal) ++ Seq(arg)
         pable <- optPable // Remove Option[_]
       } yield pable
       Printables(seq)
```
jackkoenig, line 350: If you do my suggested flatMap above, this just becomes
```scala
Printables(parts.head +: partsAndSpecifierSeq)
```
Should also rename `partsAndSpecifierSeq`.

jackkoenig (changes requested):
Lots of detailed code nits, we can meet to discuss if that helps.

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +258,115 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+    class MyModule extends BasicTester {
+      printf(cf"This should error out for sure because of % - it should be %%")
+    }
+    a[java.util.UnknownFormatConversionException] should be thrownBy {
+      extractCause[java.util.UnknownFormatConversionException] {
+        ChiselStage.elaborate { new MyModule }
+      }
+    }
+  }
+
```
jackkoenig, line 357: This test is good, it would also be good to have one that only has an unescapted `%` at the end of the `cf"..."` String because this is a common corner case to miss.

girishpai: @jackkoenig  this case is indirectly covered in p"%"  since it calls cf("%%") - but will add one explicitly with cf"%%"

jackkoenig: > @jackkoenig this case is indirectly covered in p"%" since it calls cf("%%") - but will add one explicitly with cf"%%"

I don't think that's true, since `p"%"` should not error but `cf"...%"` should.

girishpai: Agreed - about cf"%" should error - but cf"%%" should not which is what p"%" translates to. Anyway point taken about this scenario!

girishpai: Just realized I misread your first statement to begin with - my bad!

core/src/main/scala/chisel3/package.scala
girishpai: Refactored as suggested - except needed to ensure percentSplitter is called on parts.head as well.

girishpai: Renamed `partsAndSpecifierSeq` to `pables`

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: Added the test case to check if single unescaped % does cause an exception.

core/src/main/scala/chisel3/package.scala
girishpai: Thanks! Incorporated - had to make some minor changes for corner cases (like only if %% exists in whole string)

core/src/main/scala/chisel3/package.scala
girishpai: Refactored as suggested!

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +258,115 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+      val f1 = 20.45156
+      val i1 = 10
+      val s1: Short = 15
+      val l1: Long = 253
+      printf(cf"F1 = $f1 D1 = $i1 F1 formatted = $f1%2.2f s1 = $s1 l1 = $l1")
+
+    }
+    generateAndCheck(new MyModule) {
+      case Seq(Printf("F1 = 20.45156 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253", Seq())) =>
+    }.getOrElse(fail())
```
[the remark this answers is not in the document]
girishpai: Added variables using %x and %s in the same test.

core/src/main/scala/chisel3/package.scala
girishpai: @jackkoenig  - I do not understand / agree with this change. We do need to pass "fmt" to ensure things like $scala_var%2.2f or $scala_var%d works. The above change will only pick $scala_var and pick the default toString value returned.

girishpai:
@jackkoenig  have incorporated all but one of your suggested changes (with some minor additions to fix some corner cases). Please take a look.

core/src/main/scala/chisel3/package.scala
jackkoenig: Yeah I had realized that and forgot to finish my edit. Your current thing is right, although stylistically I'd still just put it on 1 line.

Also you don't need `{...}` on `case`

jackkoenig:
```suggestion
            case t => PString(fmt.getOrElse("%s").format(t))
```
Slight formatting suggestion.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +211,116 @@ package object chisel3 {
...
+      // If not - then flag an error. 
+      // Return seq of Printables (either PString or Percent or both - nothing else
+      def percentSplitter(s : String): Seq[Printable] = {
+          if(s.isEmpty) return Seq(PString(""))
+	        val pieces = s.split("%%").toList.flatMap { p =>
+	          if (p.contains('%')) throw new UnknownFormatConversionException("Un-escaped % found")
+	          // Wrap in PString and intersperse the escaped percentages
+	          Seq(Percent, PString(p))
+	        }
+	        if(pieces.isEmpty) Seq(Percent) else pieces.tail  // Don't forget to drop the extra percent we put at the beginning
```
jackkoenig, line 259:
```suggestion
          if (s.isEmpty) Seq(PString(""))
          else {
	    val pieces = s.split("%%").toList.flatMap { p =>
	      if (p.contains('%')) throw new UnknownFormatConversionException("Un-escaped % found")
	      // Wrap in PString and intersperse the escaped percentages
	      Seq(Percent, PString(p))
	    }
	    if (pieces.isEmpty) Seq(Percent) else pieces.tail  // Don't forget to drop the extra percent we put at the beginning
	  }
```
Some minor formatting (space between `if` and `(`) -- this would probably be done by scalafmt

Also `return` is frowned up on Scala so tweaked this to use `if-else`

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +211,116 @@ package object chisel3 {
...
+            if (
+              !part.isEmpty() && part.charAt(0) == '%' && (part.size == 1 || (part.size >= 2 && part.charAt(1) != '%'))
+            ) part.indexWhere { _.isLetter }
+            else -1
+
+          val fmt = if (idx_of_fmt_str >= 0) Some(part.substring(0, idx_of_fmt_str + 1)) else None
+
+          val formatRemovedString = part.zipWithIndex.collect { case (p, idx) if idx > idx_of_fmt_str => p }.mkString
+
+          (fmt,formatRemovedString)
```
jackkoenig, line 276:
```suggestion
          val endFmtIdx =
            if (part.length > 1 && part(0) == '%' && part(1) != '%') part.indexWhere(_.isLetter)
            else -1
          val (fmt, rest) = part.splitAt(endFmtIdx)
          
          val fmtOpt = if (fmt.nonEmpty) Some(fmt) else None
          (fmtOpt, rest)
```
Overall logic looks good, here's a slight simplification to the condition (check my logic) and the use of `splitAt` also helps.

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -47,88 +49,93 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
         Printf(str, args)
     }
   }
 
+
+  // Generates firrtl, gets Printfs
+  // Returns None if failed match; else calls the partial function which could have its own check
+  // Returns Some(true) to caller 
+  // Not calling fail() here - and letting caller do so - helps in localizing errors correctly. 
+  private def generateAndCheck(gen : => RawModule)(check : PartialFunction[Seq[Printf],Unit]) = {
```
jackkoenig, line 64: Rather than returning an option and having the caller do `.getOrElse(fail())` could we instead return a `Boolean` and have the caller `assert` it?

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -47,88 +49,93 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+    val firrtl = ChiselStage.emitChirrtl(gen)
+    val printfs = getPrintfs(firrtl)
+    if(!check.isDefinedAt(printfs)) {
+      println("Firrtl = \n",firrtl)
+      println("Actual = ",printfs)
```
jackkoenig, line 69: Remove these before merging

girishpai: These messages are useful when failures occur even in the future - let me try to use Scala's Either API return this along with error object (Left in this case).

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: Good point - let me do that.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +211,116 @@ package object chisel3 {
...
+
+      // Align parts to their potential specifiers
+      val pables = partsAfterFirst.zip(args).flatMap {case (part, arg) => {
+          val (fmt,modP) = extractFormatSpecifier(part)
+          val fmtArg: Printable = arg match {
+            case d: Data => {
+              fmt match {
+                case Some("%n") => Name(d)
+                case Some("%N") => FullName(d)
+                case Some("%s") => PString(d.toString)
```
jackkoenig, line 295:
```suggestion
```
Upon further reflection, we should reserve `%s` for String's that map down to Verilog. Let's not add this right now.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +211,116 @@ package object chisel3 {
...
+                case Some(x) => {
+                  val msg = s"Illegal format specifier '$x' for Chisel Data type!\n"
+                  throw new UnknownFormatConversionException(msg)
+                }
+                case None => d.toPrintable
+              }
+            }
+            case p: Printable => {
+              fmt match {
+                case Some("%s") => PString(p.toString)
```
jackkoenig, line 306:
```suggestion
```
Upon further reflection, we should reserve `%s` for String's that map down to Verilog. Let's not add this right now.

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: Actually it might get too complicated - for now will just remove these before merging and in a future PR enhance the test framework!

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: @jackkoenig  - upon converting this function to return boolean and then using an assert at the end of every test - getting this long and cryptic error message - ideally we want the failure msg to be as simple as "test failed at PrintableSpec.scala [line number]

```
  PrintableSpec.this.generateAndCheck(new MyModule())(({
    @SerialVersionUID(value = 0) final <synthetic> class $anonfun extends scala.runtime.AbstractPartialFunction[Seq[PrintableSpec.this.Printf],Unit] with java.io.Serializable {
      def <init>(): <$anon: Seq[PrintableSpec.this.Printf] => Unit> = {
        $anonfun.super.<init>();
        ()
      };
      final override def applyOrElse[A1 <: Seq[PrintableSpec.this.Printf], B1 >: Unit](x1: A1, default: A1 => B1): B1 = ((x1.asInstanceOf[Seq[PrintableSpec.this.Printf]]: Seq[PrintableSpec.this.Printf]): Seq[PrintableSpec.this.Printf] @unchecked) match {
        case scala.`package`.Seq.unapplySeq[PrintableSpec.this.Printf](<unapply-selector>) <unapply> ((str: String, args: Seq[String]): PrintableSpec.this.Printf("An exact string", scala.`package`.Seq.unapplySeq[String](<unapply-selector>) <unapply> ())) => ()
        case (defaultCase$ @ _) => default.apply(x1)
      };
      final def isDefinedAt(x1: Seq[PrintableSpec.this.Printf]): Boolean = ((x1.asInstanceOf[Seq[PrintableSpec.this.Printf]]: Seq[PrintableSpec.this.Printf]): Seq[PrintableSpec.this.Printf] @unchecked) match {
        case scala.`package`.Seq.unapplySeq[PrintableSpec.this.Printf](<unapply-selector>) <unapply> ((str: String, args: Seq[String]): PrintableSpec.this.Printf("An exact string", scala.`package`.Seq.unapplySeq[String](<unapply-selector>) <unapply> ())) => true
        case (defaultCase$ @ _) => false
      }
    };
    new $anonfun()
  }: PartialFunction[Seq[PrintableSpec.this.Printf],Unit])) was false (PrintableSpec.scala:86)
```

girishpai: This is what I did in one the tests 

```
it should "pass exact strings through" in {
    class MyModule extends BasicTester {
      printf(p"An exact strings")
    }
    
    assert(generateAndCheck(new MyModule) {
      case Seq(Printf("An exact string", Seq())) =>
    }) 

```

girishpai: Or maybe I should just add some default message - wanted to avoid boiler plate stuff.

jackkoenig: I think there's a better way. The way `fail()` gives a line number is that it takes an `implicit pos: Position` argument, try adding that to `generateAndCheck`

jackkoenig: `org.scalactic.source.Position` I think is the complete path. It's very similar to Chisel's own `SourceInfo` 😂

girishpai: Slight tangent - why did we have our own SourceInfo as opposed to reusing Position? - ok topic for an offline discussion!

girishpai: btw @jackkoenig  you rock! Thanks for the awesome idea - it does seem to work! Now this generateAndCheck is truly self-contained! Will commit my changes upon further testing...

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -61,59 +61,77 @@ object Printable {
...
-  def pack(fmt: String, data: Data*): Printable = {
+  def pack(fmt : String, data : Data*): Printable = {
     val args = data.toIterator
 
+    // The string received as an input to pack is already
+    // treated  i.e. escape sequences are processed. 
+    // Since StringContext API assumes the parts are un-treated
+    // treatEscapes is called within the implemented custom interpolators. 
+    // The literal \ needs to be escaped before sending to the custom cf interpolator. 
+    val fmtEscapeBackSlash = fmt.replace("\\","\\\\")
```
girishpai, line 72: @jackkoenig  need your inputs - this is the reason why the test within PrintfSpec was failing - basically Scala strings by default are "escape treated" unless using raw interpolator. 

But the parts Seq which StringContext API provides for interpolators expects escape sequences not treated. For most cases this is not an issue (like for \n,\t etc) but for literal \ it is an issue. Hence need to add back in the escape for the literal \. 

 At least that's what I understood so far - feel free to correct me.

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -61,59 +61,79 @@ object Printable {
...
+
 
-    pables += PString(str)
-    Printables(pables)
+    // The string received as an input to pack is already
+    // treated  i.e. escape sequences are processed. 
+    // Since StringContext API assumes the parts are un-treated
+    // treatEscapes is called within the implemented custom interpolators. 
+    // The literal \ needs to be escaped before sending to the custom cf interpolator. 
+
```
girishpai, line 134: @jackkoenig need your inputs - this is the reason why the test within PrintfSpec was failing - basically Scala strings by default are "escape treated" unless using raw interpolator.

But the parts Seq which StringContext API provides for interpolators expects escape sequences not treated. For most cases this is not an issue (like for \n,\t etc) but for literal \ it is an issue. Hence need to add back in the escape for the literal .

At least that's what I understood so far - feel free to correct me.

core/src/main/scala/chisel3/package.scala
girishpai: @jackkoenig  have updated the Scaladoc for this one - please take a look when you can!

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,140 @@ package object chisel3 {
...
-            if !str.isEmpty // Handle empty Strings
-          } yield PString(str)
+      // P interpolator does not treat % differently - hence need to add % before sending to cf.
+      val t = sc.parts.map(_.replaceAll("%", "%%"))
+      StringContext(t: _*).cf(args: _*)
+    }
+
+    /** Custom string interpolator for generating formatted Printables : cf"..."
+      *
+      * Enhanced version of scala's `f` interpolator
```
mwachs5, line 225: scaladoc new lines are not meaningful, it's like `.md` Please use punctuation when you mean a new sentence or add a newline if you mean a new paragraph

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,140 @@ package object chisel3 {
...
+      * Each expression (argument) referenced within the string is
+      * converted to a particular Printable depending
+      * on the format specifier and type.
+      *
+      * For Chisel types referenced within the String
+      * %n - Returns Name Printable
```
mwachs5, line 231: not sure how this list gets formatted, I think scaladoc generally uses wiki syntax, so maybe need some bullets and `<code>` blocks?
https://en.wikipedia.org/wiki/Help:Wikitext#Lists

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,140 @@ package object chisel3 {
...
+      * Default is %s if no specifier given.Call String.format with the argument and specifier.
+      * Wrap the result in PString Printable
+      *
+      * Literal percents need to be escaped with a %.
+      *
+      * For the parts of the StringContext - remove format specifiers and
+      * if literal percents are present - convert them into Percent Printable.
+      * Rest of the string will be wrapped in PString Printable.
+      *
+      * For example:
```
mwachs5, line 249:
```suggestion
      * @example
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,140 @@ package object chisel3 {
...
+      * on the format specifier and type.
+      *
+      * For Chisel types referenced within the String
+      * %n - Returns Name Printable
+      * %N - Returns FullName Printable
+      * %b,%d,%x = Only applicable for Bits type - returns a FirrtlFormat Printable
+      * Default if no specifier given is to call .toPrintable on the Chisel Type
+      *
+      * For Printable type
+      *  No explicit format specifier supported - just return the Printable
```
mwachs5, line 237:
```suggestion
      * _For [[Printable]] type_: No explicit format specifier supported - just return the Printable
```

mwachs5: just a suggestion

core/src/main/scala/chisel3/package.scala
girishpai: @mwachs5  tried using the bullet scheme with ** - still no effect - as in just regular ** gets printed. Do we have some examples within any of our source code trying to render bullets?

mwachs5: I would think you'd want `*` single or `-`? Since the first `*` is for just the scaladoc itself? I'll look to see if we have something...

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,143 @@ package object chisel3 {
...
+      * 
+      * ==== For Chisel types referenced within the String ====
+      * %n - Returns [[Name]] Printable.
+      * %N - Returns [[FullName]] Printable.
+      * %b,%d,%x,%c = Only applicable for types of [[Bits]] or dreived from it. - returns ([[Binary]],[[Decimal]],
+      * [[Hexadecimal]],[[Character]]) Printable respectively. 
+      * Default if no specifier given is to call [[Data.toPrintable]] on the Chisel Type.
+      *
+      *
+      * ==== For [[Printable]] type_:  ====
```
mwachs5, line 237:
```suggestion
      * ==== For [[Printable]] type:  ====
```

core/src/main/scala/chisel3/package.scala
girishpai: Both `*` and `-` did not work. Specifically this is what I did 
```
* ==== For Chisel types referenced within the String ====
      * - %n - Returns [[Name]] Printable.
      * - %N - Returns [[FullName]] Printable.
      * -  %b,%d,%x,%c = Only applicable for types of [[Bits]] or dreived from it. - returns ([[Binary]],[[Decimal]],
      * [[Hexadecimal]],[[Character]]) Printable respectively.
      * Default if no specifier given is to call [[Data.toPrintable]] on the Chisel Type.
```

girishpai: the - gets printed instead of bullet.

girishpai: The first * on the left is part of the scaladoc.

mwachs5: for the lists, try adding newlines around them and then using `-`:

```
The markup for list blocks looks like:

/** Here is an unordered list:
  * 
  *   - First item
  *   - Second item
  *     - Sub-item to the second
  *     - Another sub-item
  *   - Third item
  *
  * Here is an ordered list:
  *
  *   1. First numbered item
  *   1. Second numbered item
  *     i. Sub-item to the second
  *     i. Another sub-item
  *   1. Third item
  */
```

girishpai: @mwachs5  worked!! Yaayyy!! Thanks!!! - Should have reached out earlier - spent ~ 20 mins or so on this - but in the process got enlightened a bit about ScalaDocs too!

girishpai: What was missing was the "extra space" before the -

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -47,87 +49,86 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
 
+  // Generates firrtl, gets Printfs
+  // Returns None if failed match; else calls the partial function which could have its own check
+  // Returns Some(true) to caller
+  // Not calling fail() here - and letting caller do so - helps in localizing errors correctly.
+  private def generateAndCheck(gen: => RawModule, pos: Position)(check: PartialFunction[Seq[Printf], Unit]) = {
+    val firrtl = ChiselStage.emitChirrtl(gen)
+    val printfs = getPrintfs(firrtl)
+    if (!check.isDefinedAt(printfs)) {
+      fail()(pos)
```
jackkoenig, line 67:
```suggestion
  private def generateAndCheck(gen: => RawModule)(check: PartialFunction[Seq[Printf], Unit])(implicit pos: Position) = {
    val firrtl = ChiselStage.emitChirrtl(gen)
    val printfs = getPrintfs(firrtl)
    if (!check.isDefinedAt(printfs)) {
      fail()
```
You should not need to pass `pos` manually, if you use an implicit here, you can get rid of `Position.here` at every call site.

girishpai: @jackkoenig done!

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +254,134 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+      val str1 = "Printable String!"
+      printf(
+        cf"F1 = $f1 D1 = $i1 F1 formatted = $f1%2.2f s1 = $s1 l1 = $l1 str1 = $str1%s i1_string = $i1%s i1_hex=$i1%x"
+      )
+
+    }
+    generateAndCheck(new MyModule) {
+      case Seq(
+            Printf(
+              "F1 = 20.45156 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253 str1 = Printable String! i1_string = 10 i1_hex=a",
```
jackkoenig, line 275: Can you create and assign to a val this `Printf` object on an earlier line so that scalafmt doesn't butcher this so much?

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -63,57 +63,76 @@ object Printable {
...
+    // The string received as an input to pack is already
+    // treated  i.e. escape sequences are processed.
+    // Since StringContext API assumes the parts are un-treated
+    // treatEscapes is called within the implemented custom interpolators.
+    // The literal \ needs to be escaped before sending to the custom cf interpolator.
 
-    pables += PString(str)
-    Printables(pables)
+    val bufEscapeBackSlash = buf.map(_.replace("\\", "\\\\"))
```
jackkoenig, line 134: Can you write a test that fails without this escaping, I know that one thing that checks all ASCII checks it but I'd be nice to call out this behavior with a focused test so that if we break it in the future, the failure is localized and clear.

girishpai: Yup was working on it as we speak!

src/test/scala/chiselTests/PrintableSpec.scala
girishpai: @jackkoenig  what's the heuristic which scalafmt uses to split? I did assign it to a val like this 

```
val expected = Printf("F1 = 20.4517 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253 str1 = Printable String! i1_string = 10 i1_hex=a",Seq())
    generateAndCheck(new MyModule) {
      case Seq(`expected`) =>
    }
```

But still it split it up like this 

```
val expected = Printf(
      "F1 = 20.4517 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253 str1 = Printable String! i1_string = 10 i1_hex=a",
      Seq()
    )
    generateAndCheck(new MyModule) {
      case Seq(`expected`) =>
    }
```

jackkoenig: You could do something like:
```scala
val msg = "F1 = 20.45156 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253 str1 = Printable String! i1_string = 10 i1_hex=a"
generateAndCheck(new MyModule) {
  case Seq(Printf(`msg`), Seq()) =>
}
```

some folks don't like the backticks for comparison instead of binding a value name so you could do:
```scala
val expected = "F1 = 20.45156 D1 = 10 F1 formatted = 20.45 s1 = 15 l1 = 253 str1 = Printable String! i1_string = 10 i1_hex=a"
generateAndCheck(new MyModule) {
  case Seq(Printf(msg), Seq()) if msg == expected =>
}
```

girishpai: @jackkoenig  - I took a reasonable shortcut - pruned the expected string to remove long and short types - did not add much value in terms of coverage - and that seems to have done the trick too. Hope that's fine - else can resort to your suggestion.

core/src/main/scala/chisel3/Printable.scala
girishpai: @jackkoenig  - done. Added two different tests - one with Printable.pack - and should pass due to the above code - and the other one which actually fails when calling StringContext.cf (exposes the API requirement).

src/test/scala/chiselTests/PrintableSpec.scala
```diff
@@ -261,4 +254,145 @@ class PrintableSpec extends AnyFlatSpec with Matchers {
...
+  }
+
+  it should "fail when passing directly to StirngContext.cf a string with  literal \\ correctly escaped  " in {
+    a[StringContext.InvalidEscapeException] should be thrownBy {
+      extractCause[StringContext.InvalidEscapeException] {
+        val s_seq = Seq("Test with literal \\ correctly escaped")
+        StringContext(s_seq: _*).cf(Seq(): _*)
+      }
+    }
+  }
```
jackkoenig, line 388: Why _should_ this fail?

girishpai: This is confirming the expected API when calling cf directly like this - the user has to explicitly pass the "parts" - and ensure that it does not have literal \ where as when calling cf"...." - scala compiler will pass the parts as a raw string with none of the escape sequences processed.

girishpai: Any string literal by default scala processes all escape sequences - unless we use the raw interpolator.

girishpai: Again - feel free to correct me!

girishpai: Hence in this call the parts contain processed literal backslash "\\" becomes '\\' and then the treatEscapes within cf complains that it found a literal \

girishpai: ok - even the editor is not allowing me to show **" \ \ "** becomes '\\' - assume no space between \\ in the string.

core/src/main/scala/chisel3/package.scala
```diff
@@ -210,29 +212,142 @@ package object chisel3 {
...
+      *
+      * @example
+      * {{{
+      *
+      * val w1  = 20.U // Chisel UInt type (which extends Bits)
+      * val f1 = 30.2 // Scala float type.
+      * val pable = cf"w1 = $w1%x f1 = $f1%2.2f. This is 100% clear"
+      *
+      * // pable is as follows
+      * // Printables(List(PString(w1 = ), Binary(UInt<5>(20)), PString( f1 = ), PString(30.20), PString(. This is 100), Percent, PString( clear)))
```
jackkoenig, line 259:
```suggestion
      * val pable = cf"w1 = $w1%x f1 = $f1%2.2f. This is 100%% clear"
      *
      * // pable is as follows
      * // Printables(List(PString(w1 = ), Hexadecimal(UInt<5>(20)), PString( f1 = ), PString(30.20), PString(. This is 100), Percent, PString( clear)))
```

Can you check this? `%x` is hex and you need to escape the `%%`. I'm guessing you originally generated this but it wasn't quite right

girishpai: Good catch - will update.

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -63,57 +63,76 @@ object Printable {
...
         val msg = "has no matching argument!\n" + errorMsg(i)
         // Exception wraps msg in s"Format Specifier '$msg'"
         throw new MissingFormatArgumentException(msg)
       }
-      args.next()
+      val _ = args.next()
     }
+    var iter = 0
+    var curr_start = 0
+    var buf = mutable.ListBuffer.empty[String]
```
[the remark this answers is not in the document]
girishpai: Good point! - Have updated to val.

girishpai:
@mwachs5 - please review / approve when you can! 

@jackkoenig  - not sure who will do the final merge of this PR - I am assuming it is you since I do not have merge privileges.

PR #2533, opened 2022-05-16
Add Sbox in stdlib

ZenithalHourlyRate:
### Contributor Checklist

#### API Impact

Add new API

No

#### Desired Merge Strategy

Squash

#### Release Notes

An efficient SBox implementation for both AES and SM4 is added in stdlib

#### Other notes

This is suggested by https://github.com/chipsalliance/rocket-chip/pull/2950#discussion_r873330412

This is beneficial for all chisel user who wants to implement SBox for their scalar crypto extension

- [ ] Did you review?

### DIFF
--- a/stdlib/src/main/scala/chisel3/std/SBox.scala
+++ b/stdlib/src/main/scala/chisel3/std/SBox.scala
@@ -316,6 +316,7 @@ object SBoxAESDecOut {
 }
 
 /** bottom (outer) linear layer for SM4
+  * See SBoxMid for more doc
   */
 object SBoxSM4Out {
   def apply(x: UInt): UInt = {


sequencer:
If not using a VIA ROM or other custom designs, this seems to be a state-of-the-art SBox implementation(it can be directly synthesis with a good performance), and its idea can also be reused to other cryptography algorithms. So besides providing a implementation of AES and SM4, I prefer a hardware generator which directly describe the algorithm, and use this algorithm to generate the RTL.

stdlib/src/main/scala/chisel3/std/SBox.scala
```diff
@@ -0,0 +1,344 @@
...
+
+/*
+    [BoPe12] Boyar J., Peralta R. "A Small Depth-16 Circuit for the AES
+    S-Box." Proc.SEC 2012. IFIP AICT 376. Springer, pp. 287-298 (2012)
+    DOI: https://doi.org/10.1007/978-3-642-30436-1_24
+    Preprint: https://eprint.iacr.org/2011/332.pdf
+
+    [Ny93] Nyberg K., "Differentially Uniform Mappings for Cryptography",
+    Proc. EUROCRYPT '93, LNCS 765, Springer, pp. 55-64 (1993)
+    DOI: https://doi.org/10.1007/3-540-48285-7_6
```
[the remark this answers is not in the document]
sequencer: emmm...
I think the API should be discussed.
Can you join the dev-meeting today or tomorrow to explain the basic idea of your implementation?
I thought it's kind of an affine transformation? I'm not so sure about it.

stdlib/src/main/scala/chisel3/std/SBox.scala
```diff
@@ -0,0 +1,344 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+//  This is translated from https://github.com/riscv/riscv-crypto
+//
+//    2020-01-29    Markku-Juhani O. Saarinen <mjos@pqshield.com>
+//    Copyright (c) 2020, PQShield Ltd. All rights reserved.
+//    Converted to Chisel in 2022
```
seldridge, line 7: We should get some clarification on what the actual license of the RTL implementation is. Some files in the RTL directory are "do whatever you want" whereas this one doesn't specify. That may then implicitly default to the top-level CC 4.0 license which is incompatible with Apache.

I think this is just oversight as the `riscv/riscv-crypto` repo is supposed to be using CC 4.0 for all RISC-V related spec stuff, but I don't think the intent is to apply this to any RTL or software as well.

ZenithalHourlyRate:
> How could I make use of the S-Box from this PR to simplify my code?

In `class S` you can use `SBoxAESEncOut(SBoxMid(SBoxAESEncIn(io.in)))` instead of `VecInit(StaticTables.S.map(_.U(8.W)))(io.in)`

Also, you can merge the datapath of `S` and `xS` as demoed in
https://github.com/OpenRigil/rocket-chip/blob/076d041e02d08cc78500cdea004276f57f1eca87/src/main/scala/rocket/CryptoNIST.scala#L119-L122
and that is why we seperate `SBoxMid` into a standalone object

stdlib/src/main/scala/chisel3/std/SBox.scala
seldridge: Clarifying that the "do whatever you want" of other RTL files I spot-checked are ISC. This should be permissive enough to sweep it in.

Yes, creating an issue on `riscv/riscv-crypto` would be a good idea.

ZenithalHourlyRate:
> Also, you can merge the datapath of `S` and `xS`

It seems that your impl only has the encryption part so no merge can be done. I think encryption/decryption would both be present most of the time and reusing some datapath is wanted

ZenithalHourlyRate:
> We need document explaining how and where to integrate these modules into custom crypto designs.

Agreed. doc is a TODO.

talking about custom crypto design, I think we could have a package like `stdlib.crypto`?

A util like "GFMul a UInt with Int" is useful for AES (maybe also for other crypto tools, I've not done a survey on this) and can be upstreamed to chisel3, I think. See https://github.com/OpenRigil/rocket-chip/blob/076d041e02d08cc78500cdea004276f57f1eca87/src/main/scala/rocket/CryptoNIST.scala#L125-L145

Other utils in `CryptoNIST.scala` is also useful for other AES implementations, though that needs quite a lot discussion.

stdlib/src/main/scala/chisel3/std/SBox.scala
ZenithalHourlyRate: > Yes, creating an issue on `riscv/riscv-crypto` would be a good idea.

Created. https://github.com/riscv/riscv-crypto/issues/159

mjosaarinen: For what it's worth I'm making a public domain dedication for the sbox code. You have my permission to change the header accordingly. Of course, an acknowledgment of original authorship would be nice.

Background: RISC-V contributions are BSD 2-Clause (it's in the RISC-V membership agreement) and that was the license in the research repo where the S-Boxes came from https://github.com/mjosaarinen/lwaes_isa/blob/master/hdl/sboxes.v  So there always was a free license to use the copyrighted code.

stdlib/src/main/scala/chisel3/std/SBox.scala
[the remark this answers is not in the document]
ekiwi: The way to review it would be to copy the results into the test and then run the generator function, checking if they are correct.

ZenithalHourlyRate:
> Agreed. doc is a TODO.

Doc added.

stdlib/src/main/scala/chisel3/std/SBox.scala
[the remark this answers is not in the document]
ZenithalHourlyRate: > The way to review it would be to copy the results into the test and then run the generator function, checking if they are correct.

Sorry I did not quite get it. What "results" and "generator function" did you mean?  I think the current copy-pasted SBox table in the integration test could be "results" and the "generator" is derived by `z3` in SBoxTester, and they are checked by `z3`.

integration-tests/src/test/scala/chisel3/std/SBox.scala
```diff
@@ -0,0 +1,115 @@
...
+  output := SBoxSM4Out(SBoxMid(SBoxSM4In(input)))
+
+  val table = SM4.sbox
+  val lut = VecInit(table.map(_.U(8.W)).toSeq)
+  val ref = lut(input)
+
+  assert(output === ref)
+}
+
+class SBoxTester extends AnyFlatSpec with ChiselScalatestTester with Formal {
```
ekiwi, line 105: For these cases I would actually recommend _not_ doing a formal check since the inputs space is fairly small (256). You could just do a normal treadle based test and iterate through all input values externally. That would most likely be more efficient than calling out to Z3.

ekiwi (changes requested):
Looks like you are doing the test the way I described. I am just not sure if calling to Z3 makes sense in this case since you already have a lookup table that you compare to, so there is nothing smart that Z3 can do that would make it better than blindly iterating over all 256 inputs.

integration-tests/src/test/scala/chisel3/std/SBox.scala
ZenithalHourlyRate: https://github.com/chipsalliance/chisel3/pull/2955
Treadle has been removed.

sequencer:
since treadle is removed. I think it’s OK for now. But I just realised that The formal tool will be deprecated with SFC as well……………
I think we can depend on SFC formal backend on test for now until we have the LEC tool in CIRCT.

integration-tests/src/test/scala/chisel3/std/SBox.scala
ekiwi: treadle still works for the chiseltest based tests and will continue working for a while

ekiwi: it is also pretty simple to switch to the verilator backend if we have to

ekiwi:
> But I just realised that The formal tool will be deprecated with SFC as well

It is not actually going to get deprecated. At least not the full formal verification package that you can access from `chiseltest`. Only the `firrtl` compiler itself will be deprecated, but I plan on keeping the formal capabilities of chiseltest alive for as long as possible.

PR #2534, opened 2022-05-16
Deprecate implicit .U() and .S() syntax for literal bit extracts

### LINKED ISSUE #2413: Isn't .U syntax supposed to only take .W arguments?
**Type of issue**: bug report

**Impact**: API modification

**Other information**
 

Put this in a `Module`:

```  
val bar = 5
val foo = bar.U(7)
```

Here is a simple Scastie (some commented out code that also all works unexpectedly)
https://scastie.scala-lang.org/nqEQGHEUTziPKNj7NUirBA

This code compiles without error

I thought it should be an error to have no `.W` suffix on the arguments inside the `.U`, since those are not supposed to directly take integer arguments

    - version: 3.5.1
    
Hard-to-catch unintended user errors

jared-barocsi:
Fixes #2413: It is currently possible to pass an integer to the literal .U/.S syntax as a way of immediately bit extracting on a literal value, but this poses confusion and ambiguity as users usually intend to pass a Width.

### Contributor Checklist

#### Type of Improvement

- bug fix
- new feature/API

#### API Impact

Deprecates `.U(<int>)` and `.S(<int>)` as a way of bit extracting on a literal value, instead advising users to explicitly invoke `.U.apply(int)`

No impact

#### Desired Merge Strategy

Squash and merge

#### Release Notes
Deprecate implicit literal bit extracts, e.g. `1.U(8)`, to prevent users from accidentally passing an integer instead of a Width.
Instead, if a bit extract on a literal is intentional then users are advised to first assign the literal to a `val` and then extract from that to be clear about which functionality they are using.

- [x] Did you review?


aswaterman:
Please consider that UInt.extract conflicts with a rocket-chip PML version of same: https://github.com/chipsalliance/rocket-chip/blob/1f05ea88754b421684face06e52260848d27520f/src/main/scala/util/package.scala#L151

I know we rocket-chip folks don’t get to govern Chisel API decisions unilaterally, but if you can find a way to sidestep this collision, please do. One option is just having users explicitly invoke the apply method.

ekiwi:
> Please consider that UInt.extract conflicts with a rocket-chip PML version of same: https://github.com/chipsalliance/rocket-chip/blob/1f05ea88754b421684face06e52260848d27520f/src/main/scala/util/package.scala#L151
> 
> I know we rocket-chip folks don’t get to govern Chisel API decisions unilaterally, but if you can find a way to sidestep this collision, please do. One option is just having users explicitly invoke the apply method.

I wonder if we could just change the behavior of `apply(hi, lo)` in chisel to match the one of the `extract` method in rocket chip and then remove the `extract` from rocket chip? Is there a reason the current behavior of `apply(hi, lo)` is better?

jackkoenig:
> I wonder if we could just change the behavior of `apply(hi, lo)` in chisel to match the one of the `extract` method in rocket chip and then remove the `extract` from rocket chip? Is there a reason the current behavior of `apply(hi, lo)` is better?

The Chisel version errors where the Rocket-Chip returns `0.U`, when `hi == lo-1`. I think it's reasonable to have the rocket-chip behavior; however it would be much better if we could return a true zero-width wire rather than a 1-bit zero. This is perhaps in the realm of "reserving the right to have that behavior" but we really need true zero-width wire support.

jackkoenig:
A reasonable plan here is to just tell users to use `.apply` and not take the name `extract`. Some day when we have zero-width wire support we can adopt the rocket-chip behavior and add `extract` as an alias for `apply`.

Given the addition of `extract` in this PR is just to slightly increase user understandability, I think avoiding the PML collision is a fine reason to just not do it [yet] and just tell users to call `.apply`.

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -177,6 +175,12 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
     }
   }
 
+  final def apply(x: Int, y: Int): UInt = macro SourceInfoTransform.xyArg
```
jackkoenig, line 178: We may end up not adding `extract`, but if we do, make sure not to get rid of ScalaDoc. All public methods need ScalaDoc.

core/src/main/scala/chisel3/package.scala
```diff
@@ -37,13 +39,48 @@ package object chisel3 {
...
-    /** Int to SInt conversion, recommended style for constants.
-      */
-    def S: SInt = SInt.Lit(bigint, Width())
+    /** BigInt to UInt conversion followed by bit extract (bad style, deprecated) */
+    @chiselRuntimeDeprecated
+    @deprecated(
+      "Passing an Int to .U is usually a mistake: It does *not* set the width but does a bit extract. " +
+        "Did you mean .U(<arg>.W)? " +
+        "If you do want bit extraction, use .U.extract(<arg>)",
+      "3.5.3"
```
jackkoenig, line 51:
```suggestion
      "Chisel 3.5.4"
```
You always need to include the name of the project in the version ([see ScalaDoc for `@deprecated` which says this](https://www.scala-lang.org/api/2.13.8/scala/deprecated.html)).

Also Chisel 3.5.3 has already been released so this is Chisel 3.5.4.

core/src/main/scala/chisel3/package.scala
```diff
@@ -37,13 +39,48 @@ package object chisel3 {
...
+      // Note that both source locator and compile options are ignored for apply on literal
+      this.U.extract(x)(DeprecatedSourceInfo, ExplicitCompileOptions.Strict)
+    }
+
+    /** BigInt to UInt conversion followed by bit extract */
+    def U(x: BigInt, y: BigInt): UInt = { // scalastyle:ignore method.name
+      // Note that both source locator and compile options are ignored for apply on literal
+      this.U.do_apply(x, y)(DeprecatedSourceInfo, ExplicitCompileOptions.Strict)
+    }
+
```
jackkoenig, line 63: Should this version also be deprecated?

mwachs5:
> Given the addition of extract in this PR is just to slightly increase user understandability, I think avoiding the PML collision is a fine reason to just not do it [yet] and just tell users to call .apply.

Is this particularly for the 2-argument `hi`, `lo` case or are you suggesting getting rid of the `.extract()` option and suggestion for even the single argument case?

src/test/scala/chiselTests/UIntOps.scala
```diff
@@ -172,15 +172,15 @@ class MatchedRotateLeftAndRight(w: Int = 13) extends BasicTester {
...
+  assert("b101010".U.extract(2) === false.B)
+  assert("b101010".U.extract(3) === true.B)
+  assert("b101010".U.extract(100) === false.B)
   assert("b101010".U(3, 0) === "b1010".U)
   assert("b101010".U(9, 0) === "b0000101010".U)
 
-  assert("b101010".U(6.W)(2) === false.B)
-  assert("b101010".U(6.W)(3) === true.B)
-  assert("b101010".U(6.W)(100) === false.B)
+  assert("b101010".U(6.W).extract(2) === false.B)
```
mwachs5, line 181: Why did these 3 tests need to be changed? I thought that if you put the (6.W) it *is* OK to put the extraction index immediately?

mwachs5:
@jared-barocsi can you change the PR title and release notes to reflect this final version, as there is no longer an `.extract`

core/src/main/scala/chisel3/package.scala
mwachs5: It sounds like we agreed this version does not need to be deprecated? As we seem to think it is not ambiguous?

src/test/scala/chiselTests/SIntOps.scala
```diff
@@ -85,11 +85,11 @@ class SIntOpsTester(c: SIntOps) extends Tester(c) {
...
 class SIntLitExtractTester extends BasicTester {
-  assert(-5.S(1) === true.B)
-  assert(-5.S(2) === false.B)
-  assert(-5.S(100) === true.B)
-  assert(-5.S(3, 0) === "b1011".U)
-  assert(-5.S(9, 0) === "b1111111011".U)
+  assert(-5.S.apply(1) === true.B)
+  assert(-5.S.apply(2) === false.B)
+  assert(-5.S.apply(100) === true.B)
+  assert(-5.S.apply(3, 0) === "b1011".U)
```
mwachs5, line 91: this is currently not consistent with the implementation -- you did not deprecate this version so you should not need the `apply` for the 2-argument version. Which way do you want to go?

src/test/scala/chiselTests/UIntOps.scala
```diff
@@ -172,11 +172,11 @@ class MatchedRotateLeftAndRight(w: Int = 13) extends BasicTester {
...
 class UIntLitExtractTester extends BasicTester {
-  assert("b101010".U(2) === false.B)
-  assert("b101010".U(3) === true.B)
-  assert("b101010".U(100) === false.B)
-  assert("b101010".U(3, 0) === "b1010".U)
-  assert("b101010".U(9, 0) === "b0000101010".U)
+  assert("b101010".U.apply(2) === false.B)
+  assert("b101010".U.apply(3) === true.B)
+  assert("b101010".U.apply(100) === false.B)
+  assert("b101010".U.apply(3, 0) === "b1010".U)
```
mwachs5, line 178: same comment as above -- inconsistent with the implementation to add the `.apply` here

core/src/main/scala/chisel3/package.scala
jackkoenig: 👍

jackkoenig:
> > Given the addition of extract in this PR is just to slightly increase user understandability, I think avoiding the PML collision is a fine reason to just not do it [yet] and just tell users to call .apply.
> 
> Is this particularly for the 2-argument `hi`, `lo` case or are you suggesting getting rid of the `.extract()` option and suggestion for even the single argument case?

For consistency reasons, I think we should only add `extract` if we're adding it for both versions.

src/test/scala/chiselTests/SIntOps.scala
mwachs5: or maybe we should have tests for both versions

jackkoenig (changes requested):
So unfortunately this doesn't currently work.

Given the following:
```scala
  val out = IO(Output(UInt()))
  out := 0.U(8)
}
```
No warning is given, it isn't calling the new deprecated function, it just calls the normal `.U` and then `.apply` on the returned `UInt`.

If you manually pass a `BigInt` you get:
```scala
  val out = IO(Output(UInt()))
  out := 0.U(BigInt(8))
}
```
This gives the following error:
```
[error] .../Jack.scala:6:12: ambiguous reference to overloaded definition,
[error] both method U in class fromBigIntToLiteral of type (x: BigInt)chisel3.Bool
[error] and  method U in class fromBigIntToLiteral of type => chisel3.UInt
[error] match argument types (scala.math.BigInt) and expected result type chisel3.Data
[error]   out := 0.U(BigInt(8))
[error]            ^
```

I'm not sure what exactly we need to do to make this work, but I'm guessing it involves turning `def U` into a macro.

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -94,9 +100,9 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
     * @param x an index
     * @return the specified bit
     */
-  final def apply(x: BigInt): Bool = macro SourceInfoTransform.xArg
+  final def apply(x: BigInt): Bool = macro IntLiteralApplyTransform.safeApply
 
-  /** @group SourceInfoTransformMacro */
+  /** @group IntLiteralApplyTransformMacro */
```
jackkoenig, line 105:
```suggestion
  /** @group SourceInfoTransformMacro */
```
This group is not related to the actual `macro` class, it's purely used for ScalaDoc grouping, leave that bit alone.

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -121,12 +127,12 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends Element wi
...
     * @param x an index
     * @return the specified bit
-    * @note convenience method allowing direct use of [[scala.Int]] without implicits
     */
-  final def apply(x: Int): Bool = macro SourceInfoTransform.xArg
+  final def apply(x: Int): Bool = macro IntLiteralApplyTransform.safeApply
 
-  /** @group SourceInfoTransformMacro */
-  final def do_apply(x: Int)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Bool = apply(BigInt(x))
+  /** @group IntLiteralApplyTransformMacro */
```
jackkoenig, line 133:
```suggestion
  /** @group SourceInfoTransformMacro */
```
This group is not related to the actual `macro` class, it's purely used for ScalaDoc grouping, leave that bit alone.

macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala
```diff
@@ -294,3 +293,33 @@ class SourceInfoWhiteboxTransform(val c: whitebox.Context) extends AutoSourceTra
     q"$thisObj.$doFuncTerm($that)($implicitSourceInfo, $implicitCompileOptions)"
   }
 }
+
+// Workaround for https://github.com/sbt/sbt/issues/3966
+object IntLiteralApplyTransform
+
+/**
+  */
```
jackkoenig, line 301:
```suggestion
```
No purpose in adding an empty ScalaDoc

jackkoenig (changes requested):
This is awesome, works great. I have some minor suggestions but I have checked this out locally and it works great! I couldn't find a way to test for compiler warnings in ScalaTest so I think we'll just go with it as is.

jackkoenig (approved):
LGTM!

PR #2548, opened 2022-05-27
Make ExtModule port naming consistent with Module

jackkoenig:
ExtModule now uses the same namePorts implementation as regular Modules.
Previously, ExtModules only allowed port naming via runtime reflection.
This meant that .suggestName and other naming APIs do not work. It also
breaks FlatIO for ExtModule which is a potential replacement API for
BlackBox's special `val io` handling.

This is an API expansion (see `API Impact` below) but it is arguably a bugfix that the APIs that work for normal modules just don't work for `ExtModule`. I can be convinced that this should be done in 3.6.0 as this fix/change is a necessary piece of my effort to remove runtime reflection naming in 3.6.0, but I think there's value in getting this piece out as a bugfix in 3.5.4.

### Contributor Checklist

#### Type of Improvement

 - bug fix 
 - code cleanup 
 - new feature/API   

#### API Impact

Previously, ExtModule ports could **only** be named by runtime reflection. This meant that compiler plugin naming and APIs like `.suggestName` would not work. This also meant that `FlatIO` didn't work. Now the naming behavior for `ExtModule` ports is consistent with that of regular `Modules`.

No impact

#### Desired Merge Strategy

   - Squash

#### Release Notes

`ExtModule` port naming is now consistent with `Module` port naming. This means things like `.suggestName` work and cases where reflective naming fails will now work via the compiler plugin. Also it makes `FlatIO` work on `ExtModule`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -71,11 +71,8 @@ package experimental {
 
       val names = nameIds(classOf[ExtModule])
 
-      // Name ports based on reflection
-      for (port <- getModulePorts) {
-        require(names.contains(port), s"Unable to name port $port in $this")
-        port.setRef(ModuleIO(this, _namespace.name(names(port))))
-      }
+      // Ports are named in the same way as regular Modules
+      namePorts(names)
 
       // All suggestions are in, force names to every node.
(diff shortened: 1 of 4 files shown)


seldridge (approved):
This unification looks great to me.

PR #2555, opened 2022-06-02
Support VerificationStatement in the naming plugin

jackkoenig:
Previously, verification statements (assert, assume, cover, and printf)
were only named via reflection.

### Contributor Checklist

#### Type of Improvement

- bug fix   

#### API Impact

Verification statements are now named in the same way as normal hardware components. 

In rare cases, some signals may get a new prefix based on the name of a verification statement, but this should only occur in cases where the statement is returned from a function/block and assigned to a val, eg.
```scala
val foo = {
  val wire = Wire(UInt(8.W))
  wire := ...
  assert(wire =/= 123.U)
}
```
In this case, `wire` will now be named `foo_wire`.

#### Desired Merge Strategy

   - Squash
  
#### Release Notes

Handle naming and prefixing of verification statements in the same way as `Data` and `Mems`. This primarily affects the names used by annotations but also will now cause prefixing when a verification statement is returned from a function or block and assigned to a `val`.

- [x] Did you review?

### DIFF
--- a/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
+++ b/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
@@ -78,10 +78,13 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
       }
     }
 
-    private val shouldMatchData:      Type => Boolean = shouldMatchGen(tq"chisel3.Data")
-    private val shouldMatchDataOrMem: Type => Boolean = shouldMatchGen(tq"chisel3.Data", tq"chisel3.MemBase[_]")
-    private val shouldMatchModule:    Type => Boolean = shouldMatchGen(tq"chisel3.experimental.BaseModule")
-    private val shouldMatchInstance:  Type => Boolean = shouldMatchGen(tq"chisel3.experimental.hierarchy.Instance[_]")
+    private val shouldMatchData: Type => Boolean = shouldMatchGen(tq"chisel3.Data")
+    // Checking for all chisel3.internal.NamedComponents, but since it is internal, we instead have
+    // to match the public subtypes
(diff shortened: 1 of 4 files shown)


mwachs5:
Lgtm. I didn't realize asserts returned a node to assign to, that's pretty nonintuitive/ maybe we should get rid of that?

azidar (approved):

jackkoenig:
> Lgtm. I didn't realize asserts returned a node to assign to, that's pretty nonintuitive/ maybe we should get rid of that?

@mwachs5 

It was only fairly recently added. While I agree it's somewhat unintuitive, it's needed for annotating them.

jackkoenig:
> Sure, but not sure why that impacts the name of some wire that is an input to the assertion...? Or maybe I am just misinterpreting your example, the name foo_wire has nothing to do with the wire being the thing returned, as it would always be called foo_wire inside a prefix block assigned to foo

Yeah this is a fair question and we can debate the behavior. This PR does 2 things (illustrated by the 2 tests):
1) Makes the plugin name verification statements so they do not rely on reflective naming
2) Makes verification statements assigned to vals provide a prefix

The combination of these 2 makes the behavior uniform with that of `Data` and `Mem`, but we could do (1) without (2). I decided to make the behavior uniform for 2 reasons.

A) Consistency tends to be more predictable and thus understandable by users
B) There is a [reported regression](https://github.com/chipsalliance/chisel3/issues/1890#issuecomment-830013672) in moving from `@chiselName` to the compiler plugin where **all** vals used to cause prefixes and now only vals of type `Data` and `Mem` do.

I was a bit motivated by the problem/complaint in (B) to _at least_ make the verification statement vals included in the prefixing behavior, but I grant it's only a tiny tiny piece of the overall feature regression there.

mwachs5 (approved):

PR #2556, opened 2022-06-02
[WIP] Make some usages of suggestName runtime deprecated

mwachs5:
### Contributor Checklist

#### Type of Improvement

  - code cleanup      

#### API Impact

This runtime deprecates some usages of the suggestName API and warns that they will be turned into errors in the future. Specifically, you can no longer:

 * call suggestName on the same HasId more than once
 * call suggestName after the name has already been computed via _computeName
 * call suggestName with the same value that the prefix (autoSeed) would have given it
 
No impact on output verilog, though people avoiding these runtimeDeprecations would see them.

#### Desired Merge Strategy

#### Release Notes

Some usages of suggestName are runtime deprecated

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -15,7 +15,6 @@ import _root_.firrtl.{AnnotationSeq, RenameMap}
 import chisel3.experimental.dataview.{reify, reifySingleData}
 import chisel3.internal.Builder.Prefix
 import logger.LazyLogging
-import chisel3.internal.RuntimeDeprecatedTransform._
 
 import scala.collection.mutable
 
@@ -138,12 +137,14 @@ private[chisel3] trait HasId extends InstanceId {
     this
   }
 
-  /** Takes the first seed suggested. Multiple calls to this function will be ignored.
+  /** Takes the first seed suggested. Multiple calls to this function will be ignored (and will become an error in future).
     * If the final computed name conflicts with another name, it may get uniquified by appending
     * a digit at the end.
     *
     * Is a higher priority than [[autoSeed]], in that regardless of whether [[autoSeed]]
     * was called, [[suggestName]] will always take precedence.
+    *
+    * @note calling this after the name has already been computed will become an error in the future.
     *
     * @param seed The seed for the name of this component
     * @return this object
@@ -219,6 +220,8 @@ private[chisel3] trait HasId extends InstanceId {
 
   /** This resolves the precedence of [[autoSeed]] and [[suggestName]]
     *
+    * @note It will become an error in the future to suggestName the same thing that autoSeed would have assigned
+    *
     * @return the current calculation of a name, if it exists
     */
   private[chisel3] def seedOpt: Option[String] = {
@@ -280,26 +283,18 @@ private[chisel3] trait HasId extends InstanceId {
   // Helper for reifying the parent of a view if the view maps to a single Target
   private[chisel3] def reifyParent: BaseModule = reifyTarget.flatMap(_._parent).getOrElse(ViewParent)
 
-  private var _usedInstanceName: Option[String] = None
   // Implementation of public methods.
-  def instanceName: String = {
-    val result = _parent match {
-      case Some(ViewParent) => reifyTarget.map(_.instanceName).getOrElse(this.refName(ViewParent.fakeComponent))
-      case Some(p) =>
-        (p._component, this) match {
-          case (Some(c), _) => refName(c)
-          case (None, d: Data) if d.topBindingOpt == Some(CrossModuleBinding) => _ref.get.localName
-          case (None, _: MemBase[Data]) => _ref.get.localName
-          case (None, _) =>
-            throwException(s"signalName/pathName should be called after circuit elaboration: $this, ${_parent}")
-        }
-      case None => throwException("this cannot happen")
-    }
-    _usedInstanceName.foreach { used =>
-      require(used == result, s"We already used instance name ${used} but this time we came up with ${result}")
-    }
-    _usedInstanceName = Some(result)
-    result
+  def instanceName: String = _parent match {
+    case Some(ViewParent) => reifyTarget.map(_.instanceName).getOrElse(this.refName(ViewParent.fakeComponent))
+    case Some(p) =>
+      (p._component, this) match {
+        case (Some(c), _) => refName(c)
+        case (None, d: Data) if d.topBindingOpt == Some(CrossModuleBinding) => _ref.get.localName
+        case (None, _: MemBase[Data]) => _ref.get.localName
+        case (None, _) =>
+          throwException(s"signalName/pathName should be called after circuit elaboration: $this, ${_parent}")
+      }
+    case None => throwException("this cannot happen")
   }
   def pathName: String = _parent match {
     case None             => instanceName
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -15,6 +15,7 @@ import _root_.firrtl.{AnnotationSeq, RenameMap}
 import chisel3.experimental.dataview.{reify, reifySingleData}
 import chisel3.internal.Builder.Prefix
 import logger.LazyLogging
+import chisel3.internal.RuntimeDeprecatedTransform._
 
 import scala.collection.mutable
 
@@ -148,6 +149,16 @@ private[chisel3] trait HasId extends InstanceId {
     * @return this object
     */
   def suggestName(seed: => String): this.type = {
+    if (suggested_seed.isDefined) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, when already called with ${suggested_seed}) more than once will become an error in Chisel 3.6"
+      )
+    }
+    if (_computedName.isDefined) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, when the name was already computed as ${_computedName})) after the name has already been computed will become an error in Chisel 3.6"
+      )
+    }
     if (suggested_seed.isEmpty) suggested_seed = Some(seed)
     naming_prefix = Builder.getPrefix
     for (hook <- suggest_postseed_hooks.reverse) { hook(seed) }
@@ -165,6 +176,8 @@ private[chisel3] trait HasId extends InstanceId {
     }
   }
 
+  private[chisel3] var _computedName: Option[Option[String]] = None
+
   /** Computes the name of this HasId, if one exists
     * @param defaultPrefix Optionally provide a default prefix for computing the name
     * @param defaultSeed Optionally provide default seed for computing the name
@@ -172,30 +185,35 @@ private[chisel3] trait HasId extends InstanceId {
     */
   private[chisel3] def _computeName(defaultPrefix: Option[String], defaultSeed: Option[String]): Option[String] = {
 
-    /** Computes a name of this signal, given the seed and prefix
-      * @param seed
-      * @param prefix
-      * @return
-      */
-    def buildName(seed: String, prefix: Prefix): String = {
-      val builder = new StringBuilder()
-      prefix.foreach { p =>
-        builder ++= p
-        builder += '_'
-      }
-      builder ++= seed
-      builder.toString
-    }
+    _computedName.getOrElse {
 
-    if (hasSeed) {
-      Some(buildName(seedOpt.get, naming_prefix.reverse))
-    } else {
-      defaultSeed.map { default =>
-        defaultPrefix match {
-          case Some(p) => buildName(default, p :: naming_prefix.reverse)
-          case None    => buildName(default, naming_prefix.reverse)
+      /** Computes a name of this signal, given the seed and prefix
+        * @param seed
+        * @param prefix
+        * @return
+        */
+      def buildName(seed: String, prefix: Prefix): String = {
+        val builder = new StringBuilder()
+        prefix.foreach { p =>
+          builder ++= p
+          builder += '_'
+        }
+        builder ++= seed
+        builder.toString
+      }
+
+      val result = if (hasSeed) {
+        Some(buildName(seedOpt.get, naming_prefix.reverse))
+      } else {
+        defaultSeed.map { default =>
+          defaultPrefix match {
+            case Some(p) => buildName(default, p :: naming_prefix.reverse)
+            case None    => buildName(default, naming_prefix.reverse)
+          }
         }
       }
+      _computedName = Some(result)
+      result
     }
   }
 
@@ -203,7 +221,17 @@ private[chisel3] trait HasId extends InstanceId {
     *
     * @return the current calculation of a name, if it exists
     */
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  private[chisel3] def seedOpt: Option[String] = {
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
+          Builder.deprecated(
+            s"calling suggestName(${suggested}) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6"
+          )
+        }
+    }
+    suggested_seed.orElse(auto_seed)
+  }
 
   /** @return Whether either autoName or suggestName has been called */
   def hasSeed: Boolean = seedOpt.isDefined
@@ -252,18 +280,26 @@ private[chisel3] trait HasId extends InstanceId {
   // Helper for reifying the parent of a view if the view maps to a single Target
   private[chisel3] def reifyParent: BaseModule = reifyTarget.flatMap(_._parent).getOrElse(ViewParent)
 
+  private var _usedInstanceName: Option[String] = None
   // Implementation of public methods.
-  def instanceName: String = _parent match {
-    case Some(ViewParent) => reifyTarget.map(_.instanceName).getOrElse(this.refName(ViewParent.fakeComponent))
-    case Some(p) =>
-      (p._component, this) match {
-        case (Some(c), _) => refName(c)
-        case (None, d: Data) if d.topBindingOpt == Some(CrossModuleBinding) => _ref.get.localName
-        case (None, _: MemBase[Data]) => _ref.get.localName
-        case (None, _) =>
-          throwException(s"signalName/pathName should be called after circuit elaboration: $this, ${_parent}")
-      }
-    case None => throwException("this cannot happen")
+  def instanceName: String = {
+    val result = _parent match {
+      case Some(ViewParent) => reifyTarget.map(_.instanceName).getOrElse(this.refName(ViewParent.fakeComponent))
+      case Some(p) =>
+        (p._component, this) match {
+          case (Some(c), _) => refName(c)
+          case (None, d: Data) if d.topBindingOpt == Some(CrossModuleBinding) => _ref.get.localName
+          case (None, _: MemBase[Data]) => _ref.get.localName
+          case (None, _) =>
+            throwException(s"signalName/pathName should be called after circuit elaboration: $this, ${_parent}")
+        }
+      case None => throwException("this cannot happen")
+    }
+    _usedInstanceName.foreach { used =>
+      require(used == result, s"We already used instance name ${used} but this time we came up with ${result}")
+    }
+    _usedInstanceName = Some(result)
+    result
   }
   def pathName: String = _parent match {
     case None             => instanceName


mwachs5:
Note this currently comments out reflective naming as it kept tripping up the "you are suggesting the same name the prefix would have given you" test

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -161,49 +178,68 @@ private[chisel3] trait HasId extends InstanceId {
     // This could be called with user prefixes, ignore them
     noPrefix {
       suggested_seed = Some(seed)
-      this.suggestName(seed)
+      this.suggestNameInternal(seed)
     }
   }
 
+  private[chisel3] var _computedName: Option[Option[String]] = None
```
jackkoenig, line 185: I'm not super keen on adding this `var` because it increases the memory size of every `HasId` by 4-bytes and every named `Data` by 28-bytes. I took a different approach in https://github.com/chipsalliance/chisel3/pull/2561 to just recalculate the name and check which maybe we can do here.

jackkoenig, line 185: I'm not super keen on adding this `var` because it increases the memory size of every `HasId` by 4-bytes and every named `Data` by 28-bytes. I took a different approach in https://github.com/chipsalliance/chisel3/pull/2561 to just recalculate the name and check which maybe we can do here.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -161,49 +178,68 @@ private[chisel3] trait HasId extends InstanceId {
...
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
+          Builder.deprecated(
+            s"calling suggestName(${suggested}) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6"
+          )
+        }
+    }
+    suggested_seed.orElse(auto_seed)
+  }
```
jackkoenig, line 242: I don't think this is the right place to put this since, internally it seems reasonable to call `seedOpt` multiple times but now it might add a warning multiple times. Can we instead issue the warning upon calling `.suggestName`?

mwachs5:
This has changed in that I now am checking that thing-you-are-suggestNaming is actually `canBeNamed` (brought this back from #2561 ), and no longer checking "has name already been used before suggestname", because i'm not sure the suggestion for that given that we don't want to waste space on _computedName... probably simpler to just check if we have already closed the module

src/main/scala/chisel3/util/Arbiter.scala
```diff
@@ -132,7 +132,6 @@ class RRArbiter[T <: Data](val gen: T, val n: Int) extends LockingRRArbiter[T](g
   * consumer.io.in <> arb.io.out
   * }}}
   */
-@chiselName
```
mwachs5, line 135: oops missed one

src/main/scala/chisel3/util/Decoupled.scala
```diff
@@ -344,7 +344,6 @@ object Queue {
     *   consumer.io.in <> Queue(producer.io.out, 16)
     * }}}
     */
-  @nowarn("cat=deprecation&msg=TransitName")
```
mwachs5, line 347: this should have been deleted earlier

core/src/main/scala/chisel3/Module.scala
```diff
@@ -117,8 +117,8 @@ object Module extends SourceInfoDoc {
   */
 abstract class Module(implicit moduleCompileOptions: CompileOptions) extends RawModule {
   // Implicit clock and reset pins
-  final val clock: Clock = IO(Input(Clock())).suggestName("clock")
-  final val reset: Reset = IO(Input(mkReset)).suggestName("reset")
+  final val clock: Clock = IO(Input(Clock()))._suggestNameInternal("clock")
+  final val reset: Reset = IO(Input(mkReset))._suggestNameInternal("reset")
```
mwachs5, line 121: @azidar do we even still need these? why can't the plugin get these right?

core/src/main/scala/chisel3/internal/Builder.scala
mwachs5: no... because at time of suggestName we don't know what the auto seed is?

core/src/main/scala/chisel3/Module.scala
jackkoenig: The plugin isn't run on `core`.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -178,10 +190,22 @@ private[chisel3] trait HasId extends InstanceId {
...
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  private[chisel3] def seedOpt: Option[String] = {
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
+          Builder.deprecated(
+            s"calling suggestName(${suggested}) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6"
+          )
+        }
+    }
```
jackkoenig, line 206:
```suggestion
    if (suggested_seed.isDefined && auto_seed.isDefined) {
      if (suggested_seed.get == auto_seed.get) {
        Builder.deprecated(
          s"calling suggestName(${suggested}) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6"
        )
      }
    }
```
I think this is worth micro-optimizing since this method is called on every `Data`, sometimes multiple times. Your code _probably_ JIT compiles just fine but I know the above will be fast in practice (because most don't have a `suggested_seed` so it will generally be 1 load and 1 branch rather than potentially allocating a Tuple object.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,31 +129,42 @@ private[chisel3] trait HasId extends InstanceId {
...
     * @return this object
     */
   def suggestName(seed: => String): this.type = {
-    if (suggested_seed.isEmpty) suggested_seed = Some(seed)
-    naming_prefix = Builder.getPrefix
-    this
+    require(Builder.hasDynamicContext, s"suggestName (${seed}) should only be called from a Builder context.")
+    if (suggested_seed.isDefined) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, when already called with ${suggested_seed}) will become an error in Chisel 3.6"
```
jackkoenig, line 154:
```suggestion
        s"Calling .suggestName(\"$seed\"), when already called with \"${suggested_seed.get}\", will become an error in Chisel 3.6"
```

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,31 +129,42 @@ private[chisel3] trait HasId extends InstanceId {
...
+    * @note calling this after the name has already been computed will become an error in the future.
+    *
     * @param seed The seed for the name of this component
     * @return this object
     */
   def suggestName(seed: => String): this.type = {
-    if (suggested_seed.isEmpty) suggested_seed = Some(seed)
-    naming_prefix = Builder.getPrefix
-    this
+    require(Builder.hasDynamicContext, s"suggestName (${seed}) should only be called from a Builder context.")
```
jackkoenig, line 151: I think this should also be a deprecation rather than error (for purposes of backporting to 3.5.x). It currently just silently does nothing: https://scastie.scala-lang.org/nfiyNDBaSJWx4GKY4cxxOg

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    val pm = new PhaseManager(Seq(Dependency[chisel3.stage.phases.Checks], Dependency[chisel3.stage.phases.Elaborate]))
+    val test = pm
+      .transform(Seq(ChiselGeneratorAnnotation(() => new Test()), NoRunFirrtlCompilerAnnotation))
+      .collectFirst {
+        case d: DesignAnnotation[_] => d
+      }
+      .get
+      .design
```
jackkoenig, line 48: FYI there's a much simpler way to express this sort of thing:
```suggestion
var test: Test = null
ChiselStage.elaborate {
  test = new Test
  test
}
```
It's nasty and horrible but pithy 🙂. Scala 3 has a [neat feature](https://dotty.epfl.ch/docs/reference/experimental/cc.html) to make such hacks a compile-time error, but for the foreseeable future we can do this.

That being said I'm suggesting changing this to a warning for this PR anyway.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,31 +129,42 @@ private[chisel3] trait HasId extends InstanceId {
...
+      )
+    }
+    if (!HasId.canBeNamed(this)) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, on something that cannot actually be named: ${this}) will become an error in Chisel 3.6"
+      )
+    }
+    if (_parent.map(_.isClosed).getOrElse(false)) { // not sure what it means to have no parent
+      Builder.deprecated(
+        s"Calling suggestName ($seed, on ${this}, when the containing module (${_parent.get.name}) completed elaboration already will become an error in Chisel 3.6"
```
jackkoenig, line 164:
```suggestion
        s"Calling .suggestName(\"$seed\") on ${this} when the containing module \"${_parent.get.name}\" has already completed elaboration will become an error in Chisel 3.6"
```

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+        val x = WireInit(0.U(3.W)).suggestName("wire")
+        dontTouch(x)
+      }
+    }
+    val (log, _) = grabLog(ChiselStage.emitVerilog(new Test()))
+    log should include(
+      "calling suggestName(wire) had no effect as it is the same as the auto prefixed name"
+    )
+  }
+
```
jackkoenig, line 84: What happened to 4?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -178,10 +190,22 @@ private[chisel3] trait HasId extends InstanceId {
...
     *
     * @return the current calculation of a name, if it exists
     */
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  private[chisel3] def seedOpt: Option[String] = {
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
+          Builder.deprecated(
+            s"calling suggestName(${suggested}) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6"
```
jackkoenig, line 203:
```suggestion
            s"calling .suggestName(\"${suggested}\") had no effect as it is the same as the automatically given name, this will become an error in 3.6"
```

So this warning has a bit of a problem. For test `3` in your tests, the message you get is:
```
[deprecated] DependencyManager.scala:280 (1 calls): calling suggestName(wire) had no effect as it is the same as the auto prefixed name, this will become an error in 3.6
```

We can't give a good source locator here so we should make sure to give no source locator at all. Also we need to give the `.toString` for the `Data` so that it will at least let the user know what module it is in. Not having a source locator is pretty bad though....

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,31 +129,42 @@ private[chisel3] trait HasId extends InstanceId {
...
-    this
+    require(Builder.hasDynamicContext, s"suggestName (${seed}) should only be called from a Builder context.")
+    if (suggested_seed.isDefined) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, when already called with ${suggested_seed}) will become an error in Chisel 3.6"
+      )
+    }
+    if (!HasId.canBeNamed(this)) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, on something that cannot actually be named: ${this}) will become an error in Chisel 3.6"
```
jackkoenig, line 159:
```suggestion
        s"Calling .suggestName(\"$seed\") on \"$this\" (which cannot actually be named) will become an error in Chisel 3.6"
```

jackkoenig (changes requested):
Some minor suggestions, and 1 problem we should sort out (the error message for the suggestName matching the automatically derived name).

core/src/main/scala/chisel3/internal/Builder.scala
mwachs5: hmm, i thought I had a unit test for this error case so I am surprised it silently does nothing...

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
mwachs5: Sigh.

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+      val inst = Instance(defn)
+      inst.suggestName("fuzz")
+      val fuzz = inst
+    }
+    val (log, chirrtl) = grabLog(ChiselStage.emitChirrtl(new Example))
+    log should equal("")
+    chirrtl should include("inst fuzz of AddOne")
+  }
+   */
+
```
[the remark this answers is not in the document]
mwachs5: should I just delete this? or put a ticket to track it?

src/main/scala/chisel3/util/Decoupled.scala
mwachs5: i deleted one of these off master already, it should have been deleted in the "delete TransitName" PR. But sure I can open a different PR for that.

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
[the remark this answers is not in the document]
jackkoenig: I think deleting it and noting it in an issues is good. Or just a small comment saying "TODO test suggestName on Instance when <link to issue> is resolved"

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+      q.suggestName("fuzz")
+    }
+
+    val (log, chirrtl) = grabLog(ChiselStage.emitChirrtl(new Example))
+    log should equal("")
+    chirrtl should include("inst fuzz of PassThrough")
+  }
+
+  /*
+  property("5d. Calling suggestName on an Instance instance should be allowed") {
```
mwachs5, line 158:
```suggestion
  // This test is commented out until https://github.com/chipsalliance/chisel3/issues/2613 is resolved
  property("5d. Calling suggestName on an Instance instance should be allowed") {
```

core/src/main/scala/chisel3/internal/Builder.scala
mwachs5: ok I will make it a deprecation error, but still confused about why you say it silently does nothing given the test for it

jackkoenig: Sorry, by currently I mean in Chisel 3.5.3 it does nothing (see my Scastie). I'm saying it should be a warning in 3.5.4 and an error in 3.6.0.

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
mwachs5: 😱

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -178,10 +192,22 @@ private[chisel3] trait HasId extends InstanceId {
...
     *
     * @return the current calculation of a name, if it exists
     */
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  private[chisel3] def seedOpt: Option[String] = {
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
+          Builder.deprecated(
+            "calling suggestName(\"" + suggested + "\") had no effect as it is the same as the automatically given name, this will become an error in 3.6"
```
mwachs5, line 205: Sorry @jackkoenig your long comment here got rebased away (about source locators).

Suggestions on either of these points?:

A.  I can't put `$this` into the debug string because it then becomes recursive call since this is part of toString.
B. I'm not sure how to do what you are saying about make sure the source locator does not show up... how can I suppress a source locator?

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
```diff
@@ -0,0 +1,240 @@
...
+    // Nasty use of var, only for this test purpose. Don't do stuff like this!
+    var test: Test = null
+    ChiselStage.elaborate {
+      test = new Test
+      test
+    }
+    val (log, _) = grabLog {
+      test.wire.suggestName("somethingElse")
+    }
+    log should include("suggestName(\"somethingElse\") should only be called from a Builder context")
```
mwachs5, line 50: @jackkoenig this test doesn't actually raise any runtime deprecations, what am I missing? I would assume I am not in a Builder context...

jackkoenig: I had to do a similar thing in https://github.com/chipsalliance/chisel3/pull/2550, check it out. It's awkward but this is a warning we'll add and backport, and then promptly delete.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -178,10 +192,22 @@ private[chisel3] trait HasId extends InstanceId {
...
+    *
+    * @note It will become an error in the future to suggestName the same thing that autoSeed would have assigned
     *
     * @return the current calculation of a name, if it exists
     */
-  private[chisel3] def seedOpt: Option[String] = suggested_seed.orElse(auto_seed)
+  private[chisel3] def seedOpt: Option[String] = {
+    suggested_seed.zip(auto_seed).foreach {
+      case (suggested, auto) =>
+        if (suggested == auto) {
```
jackkoenig, line 203:
```suggestion
    if (suggested_seed.isDefined && auto_seed.isDefined) {
      if (suggested_seed.get == auto_seed.get) {
```
I suggested this in a comment that got lost but I think this is worth micro-optimizing to avoid any allocations as this is called on every `Data`.

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: You could do one of the following to get rid of the source locator: (a) Use `Builder.warningNoLoc`, (b) Use `Builder.deprecated(msg, Some(""))`, (c) Enhance `Builder.deprecated` to not require a source locator

(b) sort of works but you end up with:
```
[deprecated]  (1 calls): calling .suggestName("wire")
```
I think we could make it understand empty source locators to get rid of the `(1 calls)` bit.

Alright we can't call `.toString` but we need to give the user some more information, maybe:
```scala
val name = this._parent.get.name + "." + suggested
Builder.deprecated(
  "calling .suggestName(\"" + suggested + "\") on " + name + " had no effect as it is the same as the automatically given name, this will become an error in 3.6",
  Some("")
)
```

mwachs5: how about Some("(unknown)") which is what the deprecated message would do if it couldn't figure out the location

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
mwachs5: excellent, tried that and it worked, thanks!

core/src/main/scala/chisel3/internal/Builder.scala
mwachs5: Now this is:

mwachs5: which I think is OK

src/test/scala/chiselTests/naming/SuggestNameSpec.scala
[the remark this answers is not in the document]
mwachs5: i put the comment to the issue link #2613

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -129,31 +129,42 @@ private[chisel3] trait HasId extends InstanceId {
...
+      Builder.deprecated(
+        s"Calling suggestName ($seed, when already called with ${suggested_seed}) will become an error in Chisel 3.6"
+      )
+    }
+    if (!HasId.canBeNamed(this)) {
+      Builder.deprecated(
+        s"Calling suggestName ($seed, on something that cannot actually be named: ${this}) will become an error in Chisel 3.6"
+      )
+    }
+    if (_parent.map(_.isClosed).getOrElse(false)) { // not sure what it means to have no parent
```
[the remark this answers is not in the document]
mwachs5: i could only get it to give me the one warning about the BuilderContext, so i think that is OK.

mwachs5:
I *think* I've incorporated all the feedback

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -268,7 +303,26 @@ private[chisel3] trait HasId extends InstanceId {
...
+    case m: MemBase[_] => true
+    // These names don't affect hardware
+    case _: VerificationStatement => false
+    // While the above should be comprehensive, since this is used in warning we want to be careful
+    // to never accidentally have a match error
+    case _ => false
+  }
+}
+
+/** Holds the implementation of to for Data and MemBase */
```
mwachs5, line 325:
```suggestion
/** Holds the implementation of toNamed for Data and MemBase */
```

PR #2561, opened 2022-06-03
Add --warn:reflective-naming

jackkoenig:
This new argument (and associated annotation) will turn on a warning
whenever reflective naming changes the name of a signal. This is
provided to help migrate from Chisel 3.5 to 3.6 since reflective naming
is removed in Chisel 3.6.

I also factored `_computeName` out into a resuable function so that you can ask for the name of something given a `seed` and `prefix` without having to have a `HasId` with those fields set.

This mega-conflicts with @mwachs5's https://github.com/chipsalliance/chisel3/pull/2556

### Contributor Checklist

#### Type of Improvement

 - new feature/API

#### API Impact

This adds a new CLI/annotation API to warn when some names will change between Chisel 3.5 and 3.6.

No impact

#### Desired Merge Strategy

  - Rebase

#### Release Notes

Add `--warn:reflective-naming` to enable a warning that indicates when the name of some value in the hardware will change in the migration from Chisel 3.5 to 3.6.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -43,6 +43,22 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends 
 
   val compileOptions = moduleCompileOptions
 
+  // This could be factored into a common utility
+  private def canBeNamed(id: HasId): Boolean = id match {
+    case d: Data =>
+      d.binding match {
+        case Some(_: ConstrainedBinding) => true
+        case _ => false
+      }
+    case b: BaseModule => true
+    case m: MemBase[_] => true
+    // These names don't affect hardware
+    case _: VerificationStatement => false
+    // While the above should be comprehensive, since this is used in warning we want to be careful
+    // to never accidentally have a match error
+    case _ => false
+  }
+
   private[chisel3] override def generateComponent(): Option[Component] = {
     require(!_closed, "Can't generate module more than once")
     _closed = true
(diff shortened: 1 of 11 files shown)


core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -43,6 +43,17 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends
...
 
+  // This could be factored into a common utility
+  private def canBeNamed(id: HasId): Boolean = id match {
+    case d: Data =>
+      d.binding match {
+        case Some(_: ConstrainedBinding) => true
+        case _ => false
+      }
+    case b: BaseModule => true
+    case m: Mem[_]     => true
```
mwachs5, line 54: can Instances be named?

mwachs5: does this need a _ default case? What about Verification statements?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -154,6 +180,44 @@ package object internal {
   /** Marker trait for modules that are not true modules */
   private[chisel3] trait PseudoModule extends BaseModule
 
+  /** Computes a name
+    *
+    * @param prefix The prefix associated with the seed
+    * @param seed The seed for computing the name (if available)
+    * @param defaultPrefix Optional default prefix which is only used if the seed is not defined
+    * @param defaultSeed Optional default (or "fallback") seed
+    * @return the name, if it can be computed
```
mwachs5, line 189: is it `None` if it cannot be computed? Under what scenarios can it not be computed?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -154,6 +180,44 @@ package object internal {
   /** Marker trait for modules that are not true modules */
   private[chisel3] trait PseudoModule extends BaseModule
 
+  /** Computes a name
+    *
+    * @param prefix The prefix associated with the seed
```
mwachs5, line 185: is `prefix` used at all if `seed` is not set?   In other words, is is (defaultPrefix + defaultSeed) or (prefix + seed), but never a combination of the two?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -154,6 +180,44 @@ package object internal {
...
+      builder ++= seed
+      builder.toString
+    }
+
+    seed match {
+      case Some(s) => Some(buildName(s, prefix.reverse))
+      case None =>
+        defaultSeed.map { default =>
+          defaultPrefix match {
+            case Some(p) => buildName(default, p :: prefix.reverse)
```
mwachs5, line 214: shouldn't this be defaultPrefix.reverse?

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: Yes and yes, will fix and add tests

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: It can't be computed if seed and default seed are both `None`.

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: Yeah, for the record since I'm just factoring this method out, I wouldn't worry too much about the implementation as is. We should refactor it but that's not the purpose of this PR.

That being said, I believe `defaultPrefix` is only every used to insert `_` at the beginning of temporaries, I think we can capture that functionality more directly in a follow on PR.

mwachs5 (approved):
LGTM!

jackkoenig:
Forgot to run scalafmt and forgot to add a test for `Instance`, will push in a sec.

PR #2567, opened 2022-06-07
Enforce port name uniqueness

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

- code refactoring
- code cleanup  

#### API Impact

Using `suggestName` with the same name for multiple ports will now throw an error.

#### Desired Merge Strategy

#### Release Notes

Port name duplication and port name collision with other components like wires now throws an error instead of silently sanitizing the names.

- [ ] Did you review?


core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -43,14 +43,27 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends
 
   val compileOptions = moduleCompileOptions
 
+  private[chisel3] def checkPorts(): Unit = {
```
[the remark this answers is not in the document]
adkian-sifive: Actually looks like `namePorts` is used by `ExtModule`, this may be needed here?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -210,15 +210,24 @@ private[chisel3] trait HasId extends InstanceId {
   // Uses a namespace to convert suggestion into a true name
   // Will not do any naming if the reference already assigned.
   // (e.g. tried to suggest a name to part of a Record)
-  private[chisel3] def forceName(prefix: Option[String], default: => String, namespace: Namespace): Unit =
+  private[chisel3] def forceName(
+    prefix:     Option[String],
+    default:    => String,
+    namespace:  Namespace,
+    errorIfDup: Boolean = false,
```
jackkoenig, line 217:
```suggestion
    // Resolve name collisions, typically uses namespace.name
    resolveName:  String => String
```
This logic feels pretty convoluted to me but we can fold the two arguments into one which I think makes the erroring more clear. The exception throwing can then move out of this function (which feels wrong) and into the caller site when naming a port.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -210,15 +210,24 @@ private[chisel3] trait HasId extends InstanceId {
...
+    refBuilder: String => Arg = (x: String) => (Ref(x))
+  ): Unit =
     if (_ref.isEmpty) {
-      val candidate_name = _computeName(prefix, Some(default)).get
+      val candidate_name = _computeName(prefix, Some(default).filterNot(_ => errorIfDup)).get
```
jackkoenig, line 221: I know i suggested this but that was just for development, this logic is not very clear as to what it's trying to do. Notably, the result of `_computeName` can only ever been `None` if the thing has no name and there's no `default`. If we were to expose that `default` is an `Option`, we could change this to a `.getOrElse` (with no `filterNot`) and fold the error from `checkPorts` into here (and make sure to pass `None` as the default for ports).

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -210,15 +210,24 @@ private[chisel3] trait HasId extends InstanceId {
   // Uses a namespace to convert suggestion into a true name
   // Will not do any naming if the reference already assigned.
   // (e.g. tried to suggest a name to part of a Record)
-  private[chisel3] def forceName(prefix: Option[String], default: => String, namespace: Namespace): Unit =
+  private[chisel3] def forceName(
+    prefix:     Option[String],
+    default:    => String,
+    namespace:  Namespace,
+    errorIfDup: Boolean = false,
+    refBuilder: String => Arg = (x: String) => (Ref(x))
```
jackkoenig, line 218:
```suggestion
    refBuilder: String => Arg = Ref(_)
```
or `x => ref`
Just FYI, since you have the type already defined you can use short-hand on functions like this.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -70,7 +83,7 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends
               case MemoryPortBinding(_, _) =>
                 id.forceName(None, default = "MPORT", _namespace)
               case PortBinding(_) =>
-                id.forceName(None, default = "PORT", _namespace)
+                id.forceName(None, default = "PORT", _namespace, true, { x: String => ModuleIO(this, x) })
```
jackkoenig, line 86:
```suggestion
                id.forceName(None, default = "PORT", _namespace, true, x => ModuleIO(this, x))
```
You don't need the `{}` and type. It's fine to include them if you think it makes it more clear, I don't but YMMV.

jackkoenig (changes requested):
`forceName` has always been convoluted but this makes it a bit worse, I've made some suggestions that may or may not be completely clear so lets chat if you want some clarification.

src/test/scala/chiselTests/experimental/ProgrammaticPortsSpec.scala
```diff
@@ -63,11 +66,12 @@ class ProgrammaticPortsSpec extends ChiselFlatSpec with Utils {
...
 
   "SuggestName collisions on ports" should "be illegal" in {
-    a[ChiselException] should be thrownBy extractCause[ChiselException] {
+    val error = intercept[Exception] {
       ChiselStage.elaborate(new Module {
         val foo = IO(UInt(8.W)).suggestName("apple")
         val bar = IO(UInt(8.W)).suggestName("apple")
       })
     }
+    assert(error.toString contains "Cannot have duplicate names apple_1 and apple")
```
mwachs5, line 75: would there be any way to reflect produce the source locator of the things with these names?

jackkoenig: It would be wonderful, but we would need to capture and keep source locators upon naming and that's a bit tricky. It's worth considering in future enhancements though.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -43,14 +43,25 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends
...
+        )
+      }
+    }
+  }
+
   private[chisel3] override def generateComponent(): Option[Component] = {
     require(!_closed, "Can't generate module more than once")
     _closed = true
 
     // Ports get first naming priority, since they are part of a Module's IO spec
```
jackkoenig, line 61: Can you update this comment since it is no longer true?

src/test/scala/chiselTests/experimental/ProgrammaticPortsSpec.scala
```diff
@@ -49,7 +49,9 @@ class ProgrammaticPortsSpec extends ChiselFlatSpec with Utils {
   }
 
   "Ports" should "always win over internal components in naming" in {
```
jackkoenig, line 51: Can you update the description of this test since the behavior has changed?

jackkoenig (dismissed):
A couple of minor requests but otherwise this is good to go!

adkian-sifive:
@mergifyio update

azidar (approved):

sequencer:
This PR breaks playground regression CI in https://github.com/chipsalliance/playground/runs/7890002567, which should be fixed in the RC.

PR #2571, opened 2022-06-08
Added migration for inferModuleReset

azidar:
### Contributor Checklist

I didn't add more comprehensive ScalaDoc. I will add that once we add migration options for all the compilation options.

#### API Impact

Adds another optional field to CompileOptions to migrate inferModuleReset.

#### Desired Merge Strategy

Squash

#### Release Notes

Added an optional field to CompileOptions to help users migrate the inferModuleReset compile option.

- [x] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/CompileOptions.scala
+++ b/core/src/main/scala/chisel3/CompileOptions.scala
@@ -24,7 +24,7 @@ trait CompileOptions {
   val inferModuleReset: Boolean
 
   // If marked true, then any Module which consumes inferModuleReset=false must also mix in chisel3.RequireSyncReset
-  val migrateInferModuleReset: Boolean = false
+  def migrateInferModuleReset: Boolean = false
 }
 
 object CompileOptions {
--- a/core/src/main/scala/chisel3/CompileOptions.scala
+++ b/core/src/main/scala/chisel3/CompileOptions.scala
@@ -22,6 +22,9 @@ trait CompileOptions {
   val explicitInvalidate: Boolean
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
+
+  // If marked true, then any Module which consumes inferModuleReset=false must also mix in chisel3.RequireSyncReset
+  val migrateInferModuleReset: Boolean = false
 }
 
 object CompileOptions {
--- a/core/src/main/scala/chisel3/CompileOptions.scala
+++ b/core/src/main/scala/chisel3/CompileOptions.scala
@@ -23,7 +23,7 @@ trait CompileOptions {
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
 
-  // If marked true, then any Module which consumes inferModuleReset=false must also mix in chisel3.RequireSyncReset
+  /** If marked true, then any Module which consumes `inferModuleReset=false` must also mix in [[RequireSyncReset]] */
   def migrateInferModuleReset: Boolean = false
 }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/CompileOptions.scala
```diff
@@ -22,6 +22,9 @@ trait CompileOptions {
   val explicitInvalidate: Boolean
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
+
+  // If marked true, then any Module which consumes inferModuleReset=false must also mix in chisel3.RequireSyncReset
+  val migrateInferModuleReset: Boolean = false
```
jackkoenig, line 27:
```suggestion
  def migrateInferModuleReset: Boolean = false
```
We really ought to default all virtual methods to def unless it truly is required to be a val. That being said I'm fine with keeping this as a `val` for consistency with the rest of the trait.

src/test/scala/chiselTests/MigrateCompileOptionsSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+    override val migrateInferModuleReset = true
+  }
+
+  it should "error if migrating, but not extended RequireSyncReset" in {
+    implicit val options = migrateIR
+    class Foo extends Module {
+      val io = new Bundle {}
+    }
+    intercept[Exception] {
+      ChiselStage.elaborate(new Foo {})
```
jackkoenig, line 33:
```suggestion
      ChiselStage.elaborate(new Foo)
```
There is no reason to make this an anonymous subclass

src/test/scala/chiselTests/MigrateCompileOptionsSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+  it should "not error if migrating, and you mix with RequireSyncReset" in {
+    implicit val options = migrateIR
+    class Foo extends Module with RequireSyncReset {
+      val io = new Bundle {}
+    }
+    ChiselStage.elaborate(new Foo {})
```
jackkoenig, line 41:
```suggestion
    ChiselStage.elaborate(new Foo)
```

src/test/scala/chiselTests/MigrateCompileOptionsSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+    val connectFieldsMustMatch = false
+    val declaredTypeMustBeUnbound = false
+    val dontTryConnectionsSwapped = false
+    val dontAssumeDirectionality = false
+    val checkSynthesizable = false
+    val explicitInvalidate = false
+    val inferModuleReset = false
+
+    override val migrateInferModuleReset = true
+  }
```
jackkoenig, line 25: If we're going to adopt this new technique of "migration" compile options, I think we should consider some factory methods to construct them from `Strict` and `NotStrict`, I'm thinking something like:

```scala
object CompileOptions {
  def migrationFromStrict(migrateInferModuleReset: Boolean = false): CompileOptions = ...
}
```

Anyway, I don't think it should be part of this PR, but if we are to sort of standardize on this technique I think a little bit of help for users would be warranted, otherwise this is a very power user-y API.

core/src/main/scala/chisel3/CompileOptions.scala
```diff
@@ -22,6 +22,9 @@ trait CompileOptions {
   val explicitInvalidate: Boolean
   // Should the reset type of Module be a Bool or a Reset
   val inferModuleReset: Boolean
+
+  // If marked true, then any Module which consumes inferModuleReset=false must also mix in chisel3.RequireSyncReset
```
jackkoenig, line 26:
```suggestion
  /** If marked true, then any Module which consumes `inferModuleReset=false` must also mix in [[RequireSyncReset]] */
```

jackkoenig (approved):
Some tiny nits and then some food for thought

PR #2580, opened 2022-06-15
Define leading '_' as API for creating temporaries

jackkoenig:
Note, I've rendered the updated docs [here](https://gist.github.com/jackkoenig/5777f77ba7c1a66c574d6f566bc31724).

Chisel and FIRRTL have long used signals with names beginning with an
underscore as an API to specify that the name does not really matter.
Tools like Verilator follow a similar convention and exclude signals
with underscore names from waveform dumps by default. With the
introduction of compiler-plugin prefixing in Chisel 3.4, the convention
remained but was hard for users to use unless the unnnamed signal
existed outside of any prefix domain. Notably, unnamed signals are most
useful when creating wires inside of utility methods which almost always
results in the signal ending up with a prefix.

With this commit, Chisel explicitly recognizes signals whos val names
start with an underscore and preserve that underscore regardless of any
prefixing. Chisel will also ignore such underscores when generating
prefixes based on the temporary signal, preventing accidental double
underscores in the names of signals that are prefixed by the temporary.

Here's a specific example of the change this makes, Given:
```scala
  val out = {
    val port = IO(Output(UInt(8.W)))
    val _sum = foo + bar
    val _sum2 = Wire(UInt(8.W))
    _sum2 := foo + 1.U
    port := _sum + _sum2
    port
  }
}
```
Chisel master (and 3.4.3) currently compile this to:
```
module Example :
  input clock : Clock
  input reset : UInt<1>
  input foo : UInt<8>
  input bar : UInt<8>
  output out : UInt<8>

  node _out__sum_T = add(foo, bar)
  node out__sum = tail(_out__sum_T, 1)
  wire out__sum2 : UInt<8>
  node _out__sum2_T = add(foo, UInt<1>("h1"))
  node _out__sum2_T_1 = tail(_out__sum2_T, 1)
  out__sum2 <= _out__sum2_T_1
  node _out_port_T = add(out__sum, out__sum2)
  node _out_port_T_1 = tail(_out_port_T, 1)
  out <= _out_port_T_1
```

Now this is compiled to:
```
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input foo : UInt<8>
    input bar : UInt<8>
    output out : UInt<8>

    node _out_sum_T = add(foo, bar)
    node _out_sum = tail(_out_sum_T, 1)
    wire _out_sum2 : UInt<8>
    node _out_sum2_T = add(foo, UInt<1>("h1"))
    node _out_sum2_T_1 = tail(_out_sum2_T, 1)
    _out_sum2 <= _out_sum2_T_1
    node _out_port_T = add(_out_sum, _out_sum2)
    node _out_port_T_1 = tail(_out_port_T, 1)
    out <= _out_port_T_1 
```

Note that `out__sum` and `out__sum2` become `_out_sum` and `_out_sum2`. This change comes from Chisel effectively recognizing the leading `_` is not actually part of the name, but is the signifier of a temporary so it is stripped from the "seed" and moved to the front of the final name.

Also note that `_out__sum2_T` becomes `_out_sum2_T`. This functionality looks similar to the previous paragraph, but is actually a totally different change where Chisel, again recognizing that `_` is the signifier of a temporary and not really part of the name, actually ignores the leading `_` for generating prefixes from signals.

Between these two changes, I think the naming behavior is a pretty nice improvement over the current behavior.

### Contributor Checklist

#### Type of Improvement

 - new feature/API    

#### API Impact

This adds a new API to let users creating "temporary" signals. I will note that this might have a s

This could potentially rejigger a lot of names. On the one hand, I see this as a bug fix (and thus backportable) because I think the prior behavior of ignoring users defining signals with leading `_` as sort of a bug. On the other hand, this has the potential to cause a big change in naming when bumping and users may not appreciate that on a minor release.

#### Desired Merge Strategy

  - Squash

#### Release Notes

Define leading '_' as API for creating temporaries. Chisel will now preserve the leading `_` for any val-named signal which signifies to FIRRTL and other backend tools that the name does not matter. Chisel also now ignores the leading `_` in prefixes generated from such "unnamed signals" which reduces the number of cases of double `__` in the middle of signal names.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -66,7 +66,7 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends 
           if (id.isSynthesizable) {
             id.topBinding match {
               case OpBinding(_, _) =>
-                id.forceName(Some(""), default = "T", _namespace)
+                id.forceName(None, default = "_T", _namespace)
               case MemoryPortBinding(_, _) =>
                 id.forceName(None, default = "MPORT", _namespace)
               case PortBinding(_) =>
@@ -74,7 +74,7 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends 
               case RegBinding(_, _) =>
                 id.forceName(None, default = "REG", _namespace)
               case WireBinding(_, _) =>
-                id.forceName(Some(""), default = "WIRE", _namespace)
+                id.forceName(None, default = "_WIRE", _namespace)
               case _ => // don't name literals
             }
           } // else, don't name unbound types
@@ -147,18 +147,39 @@ package object internal {
   /** Marker trait for modules that are not true modules */
   private[chisel3] trait PseudoModule extends BaseModule
 
+  /* Check if a String name is a temporary name */
+  def isTemp(name: String): Boolean = name.nonEmpty && name.head == '_'
+
   /** Creates a name String from a prefix and a seed
     * @param prefix The prefix associated with the seed (must be in correct order, *not* reversed)
     * @param seed The seed for computing the name (if available)
     */
   def buildName(seed: String, prefix: Prefix): String = {
-    val builder = new StringBuilder()
-    prefix.foreach { p =>
-      builder ++= p
-      builder += '_'
+    // Don't bother copying the String if there's no prefix
+    if (prefix.isEmpty) {
+      seed
+    } else {
+      // Using Java's String builder to micro-optimize appending a String excluding 1st character
+      // for temporaries
+      val builder = new java.lang.StringBuilder()
+      // Starting with _ is the indicator of a temporary
+      val temp = isTemp(seed)
+      // Make sure the final result is also a temporary if this is a temporary
+      if (temp) {
+        builder.append('_')
+      }
+      prefix.foreach { p =>
+        builder.append(p)
+        builder.append('_')
+      }
+      if (temp) {
+        // We've moved the leading _ to the front, drop it here
+        builder.append(seed, 1, seed.length)
+      } else {
+        builder.append(seed)
+      }
+      builder.toString
     }
-    builder ++= seed
-    builder.toString
   }
 
   // Private reflective version of "val io" to maintain Chisel.Module semantics without having
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -55,26 +55,26 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends 
       id match {
         case id: ModuleClone[_]   => id.setRefAndPortsRef(_namespace) // special handling
         case id: InstanceClone[_] => id.setAsInstanceRef()
-        case id: BaseModule       => id.forceName(None, default = id.desiredName, _namespace)
-        case id: MemBase[_]       => id.forceName(None, default = "MEM", _namespace)
-        case id: stop.Stop        => id.forceName(None, default = "stop", _namespace)
-        case id: assert.Assert    => id.forceName(None, default = "assert", _namespace)
-        case id: assume.Assume    => id.forceName(None, default = "assume", _namespace)
-        case id: cover.Cover      => id.forceName(None, default = "cover", _namespace)
-        case id: printf.Printf => id.forceName(None, default = "printf", _namespace)
+        case id: BaseModule       => id.forceName(default = id.desiredName, _namespace)
+        case id: MemBase[_]       => id.forceName(default = "MEM", _namespace)
+        case id: stop.Stop        => id.forceName(default = "stop", _namespace)
+        case id: assert.Assert    => id.forceName(default = "assert", _namespace)
+        case id: assume.Assume    => id.forceName(default = "assume", _namespace)
+        case id: cover.Cover      => id.forceName(default = "cover", _namespace)
+        case id: printf.Printf => id.forceName(default = "printf", _namespace)
         case id: Data =>
           if (id.isSynthesizable) {
             id.topBinding match {
               case OpBinding(_, _) =>
-                id.forceName(None, default = "_T", _namespace)
+                id.forceName(default = "_T", _namespace)
               case MemoryPortBinding(_, _) =>
-                id.forceName(None, default = "MPORT", _namespace)
+                id.forceName(default = "MPORT", _namespace)
               case PortBinding(_) =>
-                id.forceName(None, default = "PORT", _namespace)
+                id.forceName(default = "PORT", _namespace)
               case RegBinding(_, _) =>
-                id.forceName(None, default = "REG", _namespace)
+                id.forceName(default = "REG", _namespace)
               case WireBinding(_, _) =>
-                id.forceName(None, default = "_WIRE", _namespace)
+                id.forceName(default = "_WIRE", _namespace)
               case _ => // don't name literals
             }
           } // else, don't name unbound types
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -176,21 +176,13 @@ private[chisel3] trait HasId extends InstanceId {
   }
 
   /** Computes the name of this HasId, if one exists
-    * @param defaultPrefix Optionally provide a default prefix for computing the name
     * @param defaultSeed Optionally provide default seed for computing the name
     * @return the name, if it can be computed
     */
-  private[chisel3] def _computeName(defaultPrefix: Option[String], defaultSeed: Option[String]): Option[String] = {
-    if (hasSeed) {
-      Some(buildName(seedOpt.get, naming_prefix.reverse))
-    } else {
-      defaultSeed.map { default =>
-        defaultPrefix match {
-          case Some(p) => buildName(default, naming_prefix.reverse)
-          case None    => buildName(default, naming_prefix.reverse)
-        }
-      }
-    }
+  private[chisel3] def _computeName(defaultSeed: Option[String]): Option[String] = {
+    seedOpt
+      .orElse(defaultSeed)
+      .map(name => buildName(name, naming_prefix.reverse))
   }
 
   /** This resolves the precedence of [[autoSeed]] and [[suggestName]]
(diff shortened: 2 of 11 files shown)


jackkoenig:
I've marked this 3.6 because it has a fairly large impact on code gen stability when bumping and I am not sure that's a great thing to do in a minor release. That being said I could be convinced that this makes sense in 3.5.4.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -66,15 +66,15 @@ abstract class RawModule(implicit moduleCompileOptions: CompileOptions) extends
           if (id.isSynthesizable) {
             id.topBinding match {
               case OpBinding(_, _) =>
-                id.forceName(Some(""), default = "T", _namespace)
+                id.forceName(None, default = "_T", _namespace)
```
jackkoenig, line 69: I realized that my changes here make the "defaultPrefix" argument of `_computeName` redundant so I need to further clean that up.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -186,7 +186,7 @@ private[chisel3] trait HasId extends InstanceId {
     } else {
       defaultSeed.map { default =>
         defaultPrefix match {
-          case Some(p) => buildName(default, p :: naming_prefix.reverse)
+          case Some(p) => buildName(default, naming_prefix.reverse)
```
jackkoenig, line 189: This code can get a lot simpler now, TODO fix.

docs/src/explanations/naming.md
```diff
@@ -202,6 +202,50 @@ class Example8 extends Module {
 ChiselStage.emitVerilog(new Example8)
 ```
 
+### Behavior for "Unnamed signals" (aka "Temporaries")
+
```
mwachs5, line 206: nice. I was coming here to say "doesn't this deserve a docs update" and here it is!

src/test/scala/chiselTests/naming/PrefixSpec.scala
```diff
@@ -414,4 +414,59 @@ class PrefixSpec extends ChiselPropSpec with Utils {
...
+            4.U // literal so there is no name
+          }
+          b
+        }
+      }
+    }
+    aspectTest(() => new Test) { top: Test =>
+      Select.wires(top).map(_.instanceName) should be(List("a_b_c_d"))
+    }
+  }
```
mwachs5, line 471: should we show that adding a `suggestName` overrides the `_` behavior, whatever its?

seldridge:
> Define leading '_' as API for creating temporaries. Chisel will now preserve the leading _ for any val-named signal which signifies to FIRRTL and other backend tools that the name does not matter. Chisel also now ignores the leading _ in prefixes generated from such "unnamed signals" which reduces the number of cases of double __ in the middle of signal names.

This release notes info is great.

WDYT about codifying some of this into a Chisel spec?

Possibly useful, but the way this is evolving on the FIRRTL Dialect side is that ops have a _name kind_ enum that qualify the utility of keeping a name around and let the compiler reason about this for different optimization levels. Currently there are two values: (1) interesting and (2) droppable. This matches the existing non-underscore (droppable) and named (interesting) convention, but allows for further enumeration going forward and avoids permanently encoding this information in the leading character of a string.

src/test/scala/chiselTests/naming/PrefixSpec.scala
jackkoenig: I'm documenting how `suggestName` affects prefixing

docs/src/explanations/naming.md
```diff
@@ -202,6 +226,84 @@ class Example8 extends Module {
...
+    sum + 1.U
+  }
+  // Comes after so does *not* affect prefix above
+  out0.suggestName("foo")
+
+  // Comes before so *does* affect prefix below
+  out1.suggestName("bar")
+  out1 := {
+    // bar_diff
+    val diff = in0 - in1
```
mwachs5, line 252: but what happens if I `suggestName("mySuggestedName")` on `diff`? Is it `out_mySuggestedName` or just `mySuggestedName`?

TBH I lost track of what it actually does, then it's a question is that a bug or a feature

src/test/scala/chiselTests/naming/PrefixSpec.scala
jackkoenig: Done, see the additional docs (rendered [here](https://gist.github.com/jackkoenig/5777f77ba7c1a66c574d6f566bc31724))

docs/src/explanations/naming.md
jackkoenig: This is [already documented](https://www.chisel-lang.org/chisel3/docs/explanations/naming.html#suggest-a-signals-name-or-the-instance-name-of-a-module)

> If you want to specify the name of a signal, you can always use the .suggestName API. Please note that the suggested name will still be prefixed (including by the plugin). You can always use the noPrefix object to strip this.

docs/src/explanations/naming.md
```diff
@@ -202,6 +225,84 @@ class Example8 extends Module {
 ChiselStage.emitVerilog(new Example8)
 ```
 
+Note that using `.suggestName` does *not* affect prefixes derived from val names;
```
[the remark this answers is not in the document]
jackkoenig: The example is right below this line 😉, pasting from the mdoc-output of this section:

-------------------

Note that using `.suggestName` does *not* affect prefixes derived from val names; however, it *can* affect prefixes derived from connections (eg. `:=)`:
```scala
class ConnectionPrefixExample extends Module {
  val in0 = IO(Input(UInt(2.W)))
  val in1 = IO(Input(UInt(2.W)))

  val out0 = IO(Output(UInt()))
  val out1 = IO(Output(UInt()))

  out0 := {
    // out0_sum
    val sum = in0 + in1
    sum + 1.U
  }
  // Comes after so does *not* affect prefix above
  out0.suggestName("foo")

  // Comes before so *does* affect prefix below
  out1.suggestName("bar")
  out1 := {
    // bar_diff
    val diff = in0 - in1
    diff + 1.U
  }
}
```
```verilog
module ConnectionPrefixExample(
  input        clock,
  input        reset,
  input  [1:0] in0,
  input  [1:0] in1,
  output [1:0] foo,
  output [1:0] bar
);
  wire [1:0] out0_sum = in0 + in1; // @[naming.md 235:19]
  wire [1:0] bar_diff = in0 - in1; // @[naming.md 245:20]
  assign foo = out0_sum + 2'h1; // @[naming.md 236:9]
  assign bar = bar_diff + 2'h1; // @[naming.md 246:10]
endmodule
```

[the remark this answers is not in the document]
mwachs5: but neither of these examples is showing the suggestName on `diff` itself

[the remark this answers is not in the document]
mwachs5: but i think you covered that with the link to the "already covered" case about suggestName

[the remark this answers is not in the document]
mwachs5: Wait sorry my response was to the wrong point, be back in a sec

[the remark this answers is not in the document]
mwachs5: just looking for an example to demonstrate what you are saying in " Note that using `.suggestName` does *not* affect prefixes derived from val names"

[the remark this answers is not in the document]
jackkoenig: I've added another example to that case, here's the updated text:

----------------------

Note that using `.suggestName` does **not** affect prefixes derived from val names;
however, it _can_ affect prefixes derived from connections (eg. `:=`):

```scala
class ConnectionPrefixExample extends Module {
  val in0 = IO(Input(UInt(2.W)))
  val in1 = IO(Input(UInt(2.W)))

  val out0 = {
    val port = IO(Output(UInt()))
    // out0_mul
    val mul = in0 * in1
    port := mul + 1.U
    port
  }
  // The prefix was derived from "out0" above, suggestName does not affect prefix
  out0.suggestName("foo")

  val out1 = IO(Output(UInt()))
  val out2 = IO(Output(UInt()))


  out1 := {
    // out1_sum
    val sum = in0 + in1
    sum + 1.U
  }
  // Comes after so does *not* affect prefix above
  out1.suggestName("bar")

  // Comes before so *does* affect prefix below
  out2.suggestName("fizz")
  out2 := {
    // fizz_diff
    val diff = in0 - in1
    diff + 1.U
  }
}
```
```verilog
module ConnectionPrefixExample(
  input        clock,
  input        reset,
  input  [1:0] in0,
  input  [1:0] in1,
  output [3:0] foo,
  output [1:0] bar,
  output [1:0] fizz
);
  wire [3:0] out0_mul = in0 * in1; // @[naming.md 232:19]
  wire [1:0] out1_sum = in0 + in1; // @[naming.md 245:19]
  wire [1:0] fizz_diff = in0 - in1; // @[naming.md 255:20]
  assign foo = out0_mul + 4'h1; // @[naming.md 233:17]
  assign bar = out1_sum + 2'h1; // @[naming.md 246:9]
  assign fizz = fizz_diff + 2'h1; // @[naming.md 256:10]
endmodule
```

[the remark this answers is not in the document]
jackkoenig: Actually I'm going to tweak it to make it even more obvious by putting the `.suggestName` on `port`.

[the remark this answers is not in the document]
jackkoenig: Okay further updated

jackkoenig:
> WDYT about codifying some of this into a Chisel spec?

@seldridge I think that would be a good thing to do. It would be fairly straightforward to specify the semantics/rules for emitting FIRRTL at least. It's a bit more difficult for defining down to Verilog until we have stronger guarantees on how things lower from FIRRTL to Verilog but perhaps it's time to at least get the ball rolling.

jackkoenig:
@mwachs5 I think this is ready to :shipit:

jackkoenig:
Note this is 3.6.0 though, might be nice to backport some of the docs changes to 3.5 so it shows up on the website

docs/src/explanations/naming.md
```diff
@@ -78,6 +78,29 @@ class Example2 extends Module {
 ChiselStage.emitVerilog(new Example2)
 ```
 
+Prefixing can also be derived from the name of signals on the left-hand side of a connection.
+While this is not implemented via the compiler plugin, the behavior should feel similar:
+
+```scala mdoc
```
mwachs5, line 84: this is very helpful in an explanation doc!

mwachs5 (approved):
This is all crazy complicated but happy to approve the better docs of the status quo if we are open to changing them!

mwachs5:
> 
Note this is 3.6.0 though, might be nice to backport some of the docs changes to 3.5 so it shows up on the website

I agree -- if we can backport the docs and tests without the `_` change that would be good

jackkoenig:
Before we merge this, let me check if this actually impacts names that much. If the impact is very minor, we can backport the full PR. If the impact is huge, then I'll just backport the docs improvements.

jackkoenig:
Okay this actually has zero impact on naming unless you intentionally name things with leading `_`. And since apparently that is not used at all in SiFive internal codebases at least, it seems that this is a reasonable thing to include in a minor release.

PR #2592, opened 2022-06-21
Pass optional name in ImportDefinitionAnno used during Separate elaboration of definition and instance 

girishpai:
### Contributor Checklist

#### Type of Improvement
- code refactoring
- new API 

#### API Impact

- No impact - since the default value of this option is None - and the previous behavior should hold 

- Change is relative to the override name given for the external definition. Previously it used to pick only `definition.proto.name` - now it can hold a different value provided by the user. 

#### Technical Debt
- Ideally the original issue of getting actual names of the external components in the final circuit can / should be solved more cleanly by making Prefixing a first class support within Chisel and/or having the concept of Namespaces within Firrtl. 

#### Desired Merge Strategy
Squash 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
@@ -120,6 +120,6 @@ object Definition extends SourceInfoDoc {
   * compiled separately.
   */
 case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](
-  definition: Definition[T],
-  name:       Option[String] = None)
+  definition:      Definition[T],
+  overrideDefName: Option[String] = None)
     extends NoTargetAnnotation
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
@@ -119,5 +119,5 @@ object Definition extends SourceInfoDoc {
 /** Stores a [[Definition]] that is imported so that its Instances can be
   * compiled separately.
   */
-case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](definition: Definition[T])
+case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](definition: Definition[T], name : Option[String] = None)
     extends NoTargetAnnotation
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
@@ -119,5 +119,7 @@ object Definition extends SourceInfoDoc {
 /** Stores a [[Definition]] that is imported so that its Instances can be
   * compiled separately.
   */
-case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](definition: Definition[T], name : Option[String] = None)
+case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](
+  definition: Definition[T],
+  name:       Option[String] = None)
     extends NoTargetAnnotation
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -344,7 +344,7 @@ private[chisel3] class DynamicContext(
   // Ensure there are no repeated names for imported Definitions - both Proto Names as well as ExtMod Names
   val importAllDefinitionProtoNames = importDefinitionAnnos.map { a => a.definition.proto.name }
   val importDistinctDefinitionProtoNames = importDefinitionMap.keys.toSeq
-  val importAllDefinitionExtModNames = importDefinitionMap.values.toSeq
+  val importAllDefinitionExtModNames = importDefinitionMap.toSeq.map(_._2)
   val importDistinctDefinitionExtModNames = importAllDefinitionExtModNames.distinct
 
   if (importDistinctDefinitionProtoNames.length < importAllDefinitionProtoNames.length) {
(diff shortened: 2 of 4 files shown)


girishpai:
@mwachs5  @azidar @jackkoenig - this PR addresses part of the problem when we have prefixing / name changes involved between different elaborations - i.e. the subsequent elaborations need to be passed "the final name in the circuit" of the definition. What this final name will be is a separate problem to solve (for example when there is a `NestedModulePrefixing` being done - then we need to figure out if this definition gets the prefix or not and pass the name accordingly in the definition. Hope this made sense. 

High level this is what was done 
- Add an optional name parameter for `ImportDefinitionAnnotation` 
- Within the `DynamicContext` class populate a Map from `definition.proto.name` to this optional name. If no Name passed (aka it is None - then use the definition.proto.name` as before)
- Query this Map to get the `desiredName` for the `EmptyExtModule` 

Still working on adding more tests - feel free to comment / suggest changes or request for more clarifications.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -335,10 +336,26 @@ private[chisel3] class DynamicContext(
   val throwOnFirstError: Boolean) {
   val importDefinitionAnnos = annotationSeq.collect { case a: ImportDefinitionAnnotation[_] => a }
 
-  // Ensure there are no repeated names for imported Definitions
-  val importDefinitionNames = importDefinitionAnnos.map { a => a.definition.proto.name }
-  if (importDefinitionNames.distinct.length < importDefinitionNames.length) {
-    val duplicates = importDefinitionNames.diff(importDefinitionNames.distinct).mkString(", ")
+  // Map holding the actual names of extModules
+  val importDefinitionMap: mutable.Map[String, String] = mutable.Map.empty
```
azidar, line 340: Can you make this immutable, built from iterating over importDefinitionAnnos, rather than a mutable map?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -16,6 +16,7 @@ import chisel3.experimental.dataview.{reify, reifySingleData}
 import chisel3.internal.Builder.Prefix
 import logger.LazyLogging
 
+import scala.collection.mutable
 import scala.collection.mutable
```
azidar, line 20: You added mutable twice?

core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
```diff
@@ -119,5 +119,7 @@ object Definition extends SourceInfoDoc {
 /** Stores a [[Definition]] that is imported so that its Instances can be
   * compiled separately.
   */
-case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](definition: Definition[T])
+case class ImportDefinitionAnnotation[T <: BaseModule with IsInstantiable](
+  definition: Definition[T],
+  name:       Option[String] = None)
```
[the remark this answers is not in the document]
girishpai: Done - renamed to `overrideDefName`

core/src/main/scala/chisel3/internal/Builder.scala
girishpai: Done.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -335,20 +335,34 @@ private[chisel3] class DynamicContext(
...
+  val importDistinctDefinitionExtModNames = importAllDefinitionExtModNames.distinct
+
+  if (importDistinctDefinitionProtoNames.length < importAllDefinitionProtoNames.length) {
+    val duplicates = importAllDefinitionProtoNames.diff(importDistinctDefinitionProtoNames).mkString(", ")
     throwException(s"Expected distinct imported Definition names but found duplicates for: $duplicates")
   }
+  if (importDistinctDefinitionExtModNames.length < importAllDefinitionExtModNames.length) {
+    val duplicates = importAllDefinitionExtModNames.diff(importDistinctDefinitionExtModNames).mkString(", ")
+    throwException(s"Expected distinct overrideDef names but found duplicates for: $duplicates")
+  }
```
jackkoenig, line 357: Can you express this in such a way that it doesn't add 5 public fields to this class (which means each of these must stick around for all of Chisel elaboration). `importDefinitionMap` seems to be the only one that is needed outside of the class, can the rest be local vals of a `def` or block?

girishpai: @jackkoenig  - good point! Let me refactor.

girishpai:
@mwachs5 @jackkoenig  @azidar  - please re-review when you all can!

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -335,20 +335,39 @@ private[chisel3] class DynamicContext(
...
 
-  // Ensure there are no repeated names for imported Definitions
-  val importDefinitionNames = importDefinitionAnnos.map { a => a.definition.proto.name }
-  if (importDefinitionNames.distinct.length < importDefinitionNames.length) {
-    val duplicates = importDefinitionNames.diff(importDefinitionNames.distinct).mkString(", ")
-    throwException(s"Expected distinct imported Definition names but found duplicates for: $duplicates")
+  // Map holding the actual names of extModules
+  // Pick the definition name by default in case not passed through annotation.
+  val importDefinitionMap = importDefinitionAnnos
+    .map(a => ((a.definition.proto.name, a.overrideDefName.getOrElse(a.definition.proto.name))))
```
jackkoenig, line 341:
```suggestion
    .map(a => a.definition.proto.name -> a.overrideDefName.getOrElse(a.definition.proto.name))
```

girishpai: Done!

girishpai:
@jackkoenig  FYI we would need this backported to 3.5 (assuming that's what we use internally). Not sure who is going to take care of that.

PR #2610, opened 2022-06-29
Implement trait for Chisel compiler to name arbitrary non-Data types

jared-barocsi:
### Contributor Checklist

#### Type of Improvement

- new feature/API
#### API Impact

Adds a public trait `HasChiselName`.

no impact

#### Desired Merge Strategy

Squash and merge

#### Release Notes

Add `HasChiselName` trait for the chisel compiler to name an arbitrary type like a named component

- [ ] Did you review?


core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -162,6 +162,34 @@ package object experimental {
     */
   trait NoChiselNamePrefix
 
+  /** Treat instances of this type as a named component, and rename using the Chisel compiler
+    * plugin
+    *
+    * The compiler plugin by default automatically names instances of `Data` and `Mem`
+    * recursively. This trait allows the plugin to treat arbitrary types as named components
+    * and prefix them when they otherwise would not be
```
mwachs5, line 170:
```suggestion
    * recursively. This trait allows the plugin to treat arbitrary types as named components
    * and prefix them (and objects constructed within them) when they otherwise would not be.
```

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -162,6 +162,34 @@ package object experimental {
     */
   trait NoChiselNamePrefix
 
+  /** Treat instances of this type as a named component, and rename using the Chisel compiler
+    * plugin
```
mwachs5, line 166:
```suggestion
    * plugin.
```

src/test/scala/chiselTests/NamingAnnotationTest.scala
```diff
@@ -4,12 +4,13 @@ package chiselTests
...
 import chisel3._
 import chisel3.experimental.chiselName
+import chisel3.experimental.HasChiselName
 import chisel3.internal.InstanceId
 import chisel3.stage.ChiselStage
 
 import scala.collection.mutable.ListBuffer
 
-trait NamedModuleTester extends Module {
+trait NamedModuleTester extends Module with HasChiselName {
```
mwachs5, line 13: why do we need `HasChiselName`, isn't `Module` already `HasChiselName`?

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -162,6 +162,34 @@ package object experimental {
...
+    *   // Name of nonData.value will be "nonData_value"
+    *   // Without HasChiselName, the name will be "value"
+    *   val nonData = new NotAData
+    *
+    *   // Name of nonData.value will be "nonData2_value"
+    *   // Without HasChiselName, the name will be "value_1"
+    *   val nonData2 = new NotAData
+    * }
+    */
+  trait HasChiselName
```
jackkoenig, line 191: As we discussed offline, I think we should come up with a better name here. "HasChiselName" implies to me that the subtype has a "name" from Chisel's perspective and thus one might assume it is annotatable/will exist in the output Verilog. What this really means is that the `vals` of this type will be noticed by the Chisel compiler plugin for prefixing things that actually do have names like Wires.

"AffectsChiselPrefix" might be more precise. Any other ideas? @mwachs5 @azidar @seldridge @sequencer?

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -82,7 +82,12 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
     // Checking for all chisel3.internal.NamedComponents, but since it is internal, we instead have
     // to match the public subtypes
     private val shouldMatchNamedComp: Type => Boolean =
-      shouldMatchGen(tq"chisel3.Data", tq"chisel3.MemBase[_]", tq"chisel3.VerificationStatement")
+      shouldMatchGen(
+        tq"chisel3.Data",
+        tq"chisel3.MemBase[_]",
+        tq"chisel3.VerificationStatement",
+        tq"chisel3.experimental.HasChiselName"
+      )
```
jackkoenig, line 90: I need to include some lines from below (lines 182-190) to mention what I don't love about this current implementation:
```scala
        else if (shouldMatchNamedComp(tpe)) {
          val str = stringFromTermName(name)
          // Starting with '_' signifies a temporary, we ignore it for prefixing because we don't
          // want double "__" in names when the user is just specifying a temporary
          val prefix = if (str.head == '_') str.tail else str
          val newRHS = transform(rhs)
          val prefixed = q"chisel3.experimental.prefix.apply[$tpt](name=$prefix)(f=$newRHS)"
          val named = q"chisel3.internal.plugin.autoNameRecursively($str)($prefixed)"
          treeCopy.ValDef(dd, mods, name, tpt, localTyper.typed(named))
```

So as we can see, things that extend `HasChiselName` will now be included as matching in that `if` block. This means they will be wrapped in `autoNameRecursively`, yet they are not of type `HasId` so they aren't _actually_ namable (that `autoNameRecursively` is effectively a no-op).

Technically this has no impact on the output, but it does
1. Slightly slow down compilation
2. Slightly increase the size of the compiled user code
3. Slightly slow down elaboration time
4. Show up in stack traces which may make it appear to be doing something, but it doesn't.

I think we should not generate the `autoNameRecursively` for things that extend `HasChiselName` (but are not also `Data` obviously).

src/test/scala/chiselTests/NamingAnnotationTest.scala
jared-barocsi: `Module` isn't in the named components so `Module` val`s don't get seen by the compiler as one

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -82,9 +82,18 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
     // Checking for all chisel3.internal.NamedComponents, but since it is internal, we instead have
     // to match the public subtypes
     private val shouldMatchNamedComp: Type => Boolean =
-      shouldMatchGen(tq"chisel3.Data", tq"chisel3.MemBase[_]", tq"chisel3.VerificationStatement")
+      shouldMatchGen(
+        tq"chisel3.Data",
+        tq"chisel3.MemBase[_]",
+        tq"chisel3.VerificationStatement",
```
mwachs5, line 88: this all seems good, but does it make sense to make `Data` and `MemBase` etc just also `AffectsChiselPrefix`?

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -162,6 +162,35 @@ package object experimental {
     */
   trait NoChiselNamePrefix
 
+  /** Treat instances of this type as a named component, and rename using the Chisel compiler
+    * plugin.
+    *
+    * The compiler plugin by default automatically prefixes and names instances of `Data` and
+    * `Mem` recursively. This trait allows the plugin to treat arbitrary types as named
+    * components and prefix them (and objects constructed within them) when they otherwise
+    * would not be.
```
jackkoenig, line 171:
```suggestion
  /** Generate prefixes from values of this type in the Chisel compiler plugin
    *
    * Users can mixin this trait to tell the Chisel compiler plugin to include the names of
    * vals of this type when generating prefixes for naming `Data` and `Mem` instances.
    * This is generally useful whenever creating a `class` that contains `Data`, `Mem`,
    * or `Module` instances but does not itself extend `Data` or `Module`.
```
This is a subtle but tricky thing to communicate. I think it's important to not call instances of this type "named components" because they themselves are not named, it's just that values of this type can affect the prefix of things that are named. It might be useful to include a link to the Chisel docs page on this (using the `@see` ScalaDoc tag).

core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -162,6 +162,35 @@ package object experimental {
...
+    *   class NotAData extends AffectsChiselPrefix {
+    *     val value = Wire(Bool())
+    *   }
+    *
+    *   // Name of nonData.value will be "nonData_value"
+    *   // Without AffectsChiselPrefix, the name will be "value"
+    *   val nonData = new NotAData
+    *
+    *   // Name of nonData.value will be "nonData2_value"
+    *   // Without AffectsChiselPrefix, the name will be "value_1"
```
jackkoenig, line 188:
```suggestion
    *   // Name with AffectsChiselPrefix:    "nonData_value"
    *   // Name without AffectsChiselPrefix: "value"
    *   val nonData = new NotAData
    *
    *   // Name with AffectsChiselPrefix:    "nonData2_value"
    *   // Name without AffectsChiselPrefix: "value_1"
```
This is a great example, just a slight tweak to make the different names lineup

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -82,9 +82,18 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
...
+        tq"chisel3.experimental.AffectsChiselPrefix"
```
jackkoenig, line 89:
```suggestion
```
`AffectsChiselPrefix` is not a `NamedComponent`. If you want 1 function to check this, I would change its name.

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -171,22 +180,34 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
       // Check if a subtree is a candidate
       case dd @ ValDef(mods, name, tpt, rhs) if okVal(dd) =>
         val tpe = inferType(tpt)
+        val isData = shouldMatchData(tpe)
+
```
jackkoenig, line 184:
```suggestion
        val isData = shouldMatchData(tpe)
        // Note the use of short-circuiting operations makes these checks cheap
        val isNamedComp = isData || shouldMatchNamedComp(tpe)
        val isPrefixed = isNamedComp || shouldMatchChiselPrefixed(tpe)

```

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -171,22 +180,34 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
...
-        if (shouldMatchData(tpe) && inBundle(dd)) {
+        if (isData && inBundle(dd)) {
           val str = stringFromTermName(name)
           val newRHS = transform(rhs) // chisel3.internal.plugin.autoNameRecursively
           val named = q"chisel3.internal.plugin.autoNameRecursively($str)($newRHS)"
           treeCopy.ValDef(dd, mods, name, tpt, localTyper.typed(named))
         }
         // If a Data or a Memory, get the name and a prefix
-        else if (shouldMatchNamedComp(tpe)) {
+        else if (isData || shouldMatchNamedComp(tpe)) {
```
jackkoenig, line 193:
```suggestion
        else if (isData || isPrefixed) {
```

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -171,22 +180,34 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
...
+
+          val named =
+            if (isData || !shouldMatchChiselPrefixed(tpe))
+              // Generate autoNameRecursively call if:
+              // - the type is a data, or
+              // - the type is not a AffectsChiselPrefix
+              q"chisel3.internal.plugin.autoNameRecursively($str)($prefixed)"
+            else
+              // Do not generate if the type is both not a data and a AffectsChiselPrefix
+              prefixed
```
jackkoenig, line 209:
```suggestion
          val named =
            if (isNamedComp) {
              // Only name named components (not things that are merely prefixed)
              q"chisel3.internal.plugin.autoNameRecursively($str)($prefixed)"
            } else {
              prefixed
            }
```
1. Always use curly braces on multi-line if-else, we don't want to be the next [Goto fail](https://en.wikipedia.org/wiki/Unreachable_code#goto_fail_bug). We can probably make Scalafmt do this and we really should.
2. The comments were a little redundant so I shortened them, the code does basically say what it is doing
3. I minimized the number of lookups with a couple of extra vals above, see other comment

jackkoenig (changes requested):
This looks great and is very close. Wonderful ScalaDoc on the new trait. I made a couple of minor rewording and optimization suggestions but the approach is good!

jackkoenig (approved):
LGTM! Nice work!

PR #2626, opened 2022-07-11
Add an Explanation chapter about Scala types vs Chisel Types vs Hardware

mwachs5:
### Contributor Checklist

#### Type of Improvement

   - documentation    

#### API Impact

No impact.

No Impact.

#### Desired Merge Strategy

#### Release Notes

Add an Explanation article about Scala Type vs Chisel Type vs Hardware.

- [ ] Did you review?


docs/src/explanations/data-types.md
```diff
@@ -134,3 +134,17 @@ a <> b
 a <> b
 a <> c
 ```
+
+## Scala Type vs Chisel Type vs Hardware
+
+The *Scala* type of the Data is recognized by the Scala compiler, such as `Decoupled[UInt]` or `MyBundle`, where 
+```MyBundle(w: Int) extends Bundle {val foo: UInt(w.W), val bar: UInt(w.W)}```
```
mwachs5, line 141:
```suggestion
```
MyBundle(w: Int) extends Bundle {val foo: UInt(w.W), val bar: UInt(w.W)}
```
```

docs/src/explanations/data-types.md
```diff
@@ -134,3 +134,21 @@ a <> b
...
 ```
+
+## Scala Type vs Chisel Type vs Hardware
+
+The *Scala* type of the Data is recognized by the Scala compiler, such as `Decoupled[UInt]` or `MyBundle`, where 
+```
+MyBundle(w: Int) extends Bundle {val foo: UInt(w.W), val bar: UInt(w.W)}
+```
+
+The *Chisel* Type of a Data is all the fields actually present, by names, and their types including widths. For example, `MyBundle(3)` creates a Chisel Type of `Record` with `foo : UInt(3.W),  bar: UInt(3.W))`.
```
mwachs5, line 145:
```suggestion
The *Chisel* type of a Data is all the fields actually present, by names, and their types including widths. For example, `MyBundle(3)` creates a Chisel Type of `Record` with `foo : UInt(3.W),  bar: UInt(3.W))`.
```

mwachs5 (commented):
ref: https://stackoverflow.com/questions/70362890/what-does-vec-type-anonymousbundleio-io-in-module-must-be-a-chisel-type

docs/src/explanations/advanced-data-types.md
```diff
@@ -0,0 +1,141 @@
+---
+layout: docs
+title:  "Advanced Data Types"
+section: "chisel3"
+---
+
+# Advanced Data Types
```
jackkoenig, line 7: I don't love this title for this doc. "Advanced Data Types" implies thinking about more expressive and powerful types (think `Records` compared to `Bundles` or learning about Higher-Kinded Types in Scala). This is more of "Chisel vs. Scala Types") because fundamentally, the Chisel types are types in Chisel's type system, it's just that Chisel's type system is a Scala runtime thing.

docs/src/explanations/advanced-data-types.md
```diff
@@ -0,0 +1,141 @@
...
+// Only work for Chisel Types
+ChiselStage.elaborate(new Wrapper(5, false))
+ChiselStage.elaborate(new Wrapper(6, false))
+ChiselStage.elaborate(new Wrapper(7, false))
+```
+
+Can only `:=` to hardware:
+```scala mdoc:crash
+ChiselStage.elaborate(new Wrapper(1, false))
+```
```
jackkoenig, line 110:
````suggestion
Can only `:=` to hardware:
```scala mdoc:crash
  val chiselType = new MyBundle(3)
  chiselType := DontCare
})
```
````
Here's an example of what I mean from above, this is basically what you're testing and it's really not that bad to just inline the real check.

File: docs/src/explanations/advanced-data-types.md [hunk 5]
```diff
@@ -0,0 +1,141 @@
...
+and pure Chisel types (e.g. `Bool()`). You can get runtime errors passing a Chisel type when hardware is expected, and vice versa.
+
+## Scala Type vs Chisel Type vs Hardware
+
+The *Scala* type of the Data is recognized by the Scala compiler, such as `Decoupled[UInt]` or `MyBundle`, where 
+```
+MyBundle(w: Int) extends Bundle {val foo: UInt(w.W), val bar: UInt(w.W)}
+```
+
+The *Chisel* type of a Data is all the fields actually present, by names, and their types including widths and directions.
```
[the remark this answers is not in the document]
azidar: What I'm wanting to change in the conceptual model is that Chisel types define relative direction (flipped/aligned) of their fields, but Chisel types do NOT define the actual direction of their leafs.

This means a few things.

1. Absolute direction of a leaf field is a property of the hardware it is bound as, NOT as a property of the Chisel type. This is why I want to add `Incoming` and `Outgoing`, because they define the absolute direction of leafs of their Chisel types (subject to relative flips along the path from root-to-leaf field).
2. Along those lines, when Chisel types define absolute direction, `Wire` does not make sense conceptually in this model because a wire's direction is usually defined by how it is being used/connected. In fact, only ports really care about absolute direction.
3. Producer/consumer (who drives who in a connection) is a property of the connection operator, not a property of the "absolute direction" of things, because again that doesn't make sense for wires. *This is a desirable property* because the connection semantics work the same on all hardware of the same type (be it wires or ports or registers etc.)
4. This is why Wes's connection operators are so useful, because they are consistent regardless of the underlying bound hardware, while still being powerful enough to describe many use cases e.g. connecting to monitors.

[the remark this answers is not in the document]
mwachs5: I like these directions, but as it is today for documenting, the ChiselType I believe *does* capture the directionality? Or shoudl I just say `directionality*    *it's complicated`?

docs/src/explanations/advanced-data-types.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+```scala mdoc:silent
+// Work for both
+ChiselStage.elaborate(new Wrapper(0, true))
+ChiselStage.elaborate(new Wrapper(0, false))
+
+// Only work if synthesizable
+ChiselStage.elaborate(new Wrapper(1, true))
```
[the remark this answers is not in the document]
mwachs5: I re-wrote this in a Goofus and Gallant style, weigh in when you get a chance

PR #2628, opened 2022-07-12
New enhanced API for  specifying Chisel to Firrtl Annotations

girishpai:
### Contributor Checklist

#### Type of Improvement
- new feature / API 
- 

#### High level Changes
- With the intent to ensure no existing code breaks / needs an update - a different variable (called newAnnotations) was used. Also wherever this is used a default value of Seq.empty was provided in the function argument list - again to avoid updating existing code. 
-  A new trait called `ChiselMultiAnnotaiton` was created with `toFirrtl` method - which returns a Seq of Annotations. 

@jackkoenig  this is based on our agreed upon strategy (creating a totally new trait as opposed to adding a new method in the existing one) 

#### API Impact

No impact to current code. The current API is useful to generate annotations  containing information which will only be available after a module elaborates.  However this is limited since we can only generate a single annotation per call. In some cases we need ability to generate a Seq of annotations. 

No change to generated verilog. 

#### Desired Merge Strategy
- Squash

#### Release Notes

New API to convert a single `Chisel Annotation` into a Seq of `FirrtlAnnotations` 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -39,7 +39,6 @@ object annotate {
     Builder.annotations += anno
   }
   def apply(annos: ChiselToFirrtlAnnotations): Unit = {
-    println("Adding to new annotations in the builder")
     Builder.newAnnotations += annos
   }
 }
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -20,6 +20,10 @@ trait ChiselAnnotation {
   def toFirrtl: Annotation
 }
 
+trait ChiselToFirrtlAnnotations {
+  def toFirrtlAnnotations: Seq[Annotation]
+}
+
 /** Mixin for [[ChiselAnnotation]] that instantiates an associated FIRRTL Transform when this Annotation is present
   * during a run of
   * [[Driver$.execute(args:Array[String],dut:()=>chisel3\.RawModule)* Driver.execute]].
@@ -33,6 +37,10 @@ trait RunFirrtlTransform extends ChiselAnnotation {
 object annotate {
   def apply(anno: ChiselAnnotation): Unit = {
     Builder.annotations += anno
+  }
+  def apply(annos: ChiselToFirrtlAnnotations): Unit = {
+    println("Adding to new annotations in the builder")
+    Builder.newAnnotations += annos
   }
 }
 
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -20,6 +20,10 @@ trait ChiselAnnotation {
   def toFirrtl: Annotation
 }
 
+/** Enhanced interface for Annotations in Chisel
+  *
+  *  Defines a conversion to corresponding FIRRTL Annotation(s)
+  */
 trait ChiselToFirrtlAnnotations {
   def toFirrtlAnnotations: Seq[Annotation]
 }
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -24,8 +24,8 @@ trait ChiselAnnotation {
   *
   *  Defines a conversion to corresponding FIRRTL Annotation(s)
   */
-trait ChiselToFirrtlAnnotations {
-  def toFirrtlAnnotations: Seq[Annotation]
+trait ChiselMultiAnnotation {
+  def toFirrtl: Seq[Annotation]
 }
 
 /** Mixin for [[ChiselAnnotation]] that instantiates an associated FIRRTL Transform when this Annotation is present
@@ -42,7 +42,7 @@ object annotate {
   def apply(anno: ChiselAnnotation): Unit = {
     Builder.annotations += anno
   }
-  def apply(annos: ChiselToFirrtlAnnotations): Unit = {
+  def apply(annos: ChiselMultiAnnotation): Unit = {
     Builder.newAnnotations += annos
   }
 }
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -872,11 +872,11 @@ case class DefBlackBox(
   @deprecated("Do not use newAnnotations val of Circuit directly - use firrtlAnnotations instead. Will be removed in a future release",
     "Chisel 3.5")
   @since("Chisel 3.5.4")
-  newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
+  newAnnotations: Seq[ChiselMultiAnnotation] = Seq.empty) {
 
   def firrtlAnnotations: Iterable[Annotation] =
     annotations.flatMap(_.toFirrtl.update(renames)) ++ newAnnotations.flatMap(
-      _.toFirrtlAnnotations.flatMap(_.update(renames))
+      _.toFirrtl.flatMap(_.update(renames))
     )
 
   def copy(
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -865,9 +865,8 @@ case class Circuit(
   name:           String,
   components:     Seq[Component],
   annotations:    Seq[ChiselAnnotation],
-  newAnnotations: Seq[ChiselToFirrtlAnnotations],
-  renames:        RenameMap) {
-  println(s"NewAnnoSize = ${newAnnotations.size}")
+  renames:        RenameMap,
+  newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
   def firrtlAnnotations: Iterable[Annotation] =
     annotations.flatMap(_.toFirrtl.update(renames)) ++ newAnnotations.flatMap(
       _.toFirrtlAnnotations.flatMap(_.update(renames))
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -14,7 +14,6 @@ import _root_.firrtl.annotations.Annotation
 import scala.collection.immutable.NumericRange
 import scala.math.BigDecimal.RoundingMode
 import scala.annotation.nowarn
-import dataclass.{data, since}
 
 case class PrimOp(name: String) {
   override def toString: String = name
@@ -862,7 +861,7 @@ case class DefBlackBox(
   params: Map[String, Param])
     extends Component
 
-@data class Circuit(
+case class Circuit(
   name:       String,
   components: Seq[Component],
   @deprecated("Do not use annotations val of Circuit directly - use firrtlAnnotations instead. Will be removed in a future release",
@@ -871,8 +870,11 @@ case class DefBlackBox(
   renames:     RenameMap,
   @deprecated("Do not use newAnnotations val of Circuit directly - use firrtlAnnotations instead. Will be removed in a future release",
     "Chisel 3.5")
-  @since("Chisel 3.5.4")
-  newAnnotations: Seq[ChiselMultiAnnotation] = Seq.empty) {
+
+  newAnnotations: Seq[ChiselMultiAnnotation]) {
+
+  def this(name: String, components: Seq[Component], annotations: Seq[ChiselAnnotation], renames: RenameMap) =
+    this(name, components, annotations, renames, Seq.empty)
 
   def firrtlAnnotations: Iterable[Annotation] =
     annotations.flatMap(_.toFirrtl.update(renames)) ++ newAnnotations.flatMap(
@@ -887,8 +889,12 @@ case class DefBlackBox(
   ) = Circuit(name, components, annotations, renames, newAnnotations)
 
 }
-object Circuit {
+object Circuit
+    extends scala.runtime.AbstractFunction4[String, Seq[Component], Seq[ChiselAnnotation], RenameMap, Circuit] {
   def unapply(c: Circuit): Option[(String, Seq[Component], Seq[ChiselAnnotation], RenameMap)] = {
     Some((c.name, c.components, c.annotations, c.renames))
   }
+
+  def apply(name: String, components: Seq[Component], annotations: Seq[ChiselAnnotation], renames: RenameMap): Circuit =
+    new Circuit(name, components, annotations, renames)
 }
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -871,7 +871,8 @@ case class DefBlackBox(
   renames:     RenameMap,
   @deprecated("Do not use newAnnotations val of Circuit directly - use firrtlAnnotations instead. Will be removed in a future release",
     "Chisel 3.5")
-  @since newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
+  @since("Chisel 3.5.4")
+  newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
 
   def firrtlAnnotations: Iterable[Annotation] =
     annotations.flatMap(_.toFirrtl.update(renames)) ++ newAnnotations.flatMap(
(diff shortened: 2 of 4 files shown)


core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -861,7 +861,14 @@ case class DefBlackBox(
   params: Map[String, Param])
     extends Component
 
-case class Circuit(name: String, components: Seq[Component], annotations: Seq[ChiselAnnotation], renames: RenameMap) {
-  def firrtlAnnotations: Iterable[Annotation] = annotations.flatMap(_.toFirrtl.update(renames))
-
+case class Circuit(
```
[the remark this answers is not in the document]
mwachs5: though I supposed this does not need to block

mwachs5 (commented):
Shouldn't there be something when we do Convert to FIRRTL that actually looks at these newAnnotations?

https://github.com/chipsalliance/chisel3/blob/2b48fd15a7711dcd44334fbbc538667a102a581a/src/main/scala/chisel3/stage/phases/Convert.scala#L30

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -20,6 +20,14 @@ trait ChiselAnnotation {
   def toFirrtl: Annotation
 }
 
+/** Enhanced interface for Annotations in Chisel
+  *
+  *  Defines a conversion to corresponding FIRRTL Annotation(s)
+  */
+trait ChiselToFirrtlAnnotations {
+  def toFirrtlAnnotations: Seq[Annotation]
+}
```
seldridge, line 29: I'm not huge on this trait name, but I can't think of a better one. Maybe `FirrtlAnnotationGenerator` or `FirrtlAnnotationAdder` or `FirrtlAnnotationBuilder`. The method name would then possibly be better as `addFirrtlAnnotations` or something.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -861,7 +861,14 @@ case class DefBlackBox(
...
+  name:           String,
+  components:     Seq[Component],
+  annotations:    Seq[ChiselAnnotation],
+  renames:        RenameMap,
+  newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
+  def firrtlAnnotations: Iterable[Annotation] =
```
girishpai, line 870: > Shouldn't there be something when we do Convert to FIRRTL that actually looks at these newAnnotations?

@mwachs5  - that's handled here - in short firrtlAnnotations now includes flattening of the new Seq of annotations in addition to the ones from the legacy API.

seldridge (commented):
Generally looks good. Some ideas/suggestions throughout.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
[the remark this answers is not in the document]
girishpai: Since we did not have a scaladoc for this class to begin with - would like to de-couple that from this PR (hope that's ok).

src/test/scala/chiselTests/NewAnnotationsSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    // Pass multiple annotations in the same seq - should get emitted out correctly.
+    annotate(new ChiselToFirrtlAnnotations {
+      def toFirrtlAnnotations =
+        Seq(new DontTouchAnnotation(mod1.io.in.toNamed), new DontTouchAnnotation(mod1.io.out.toNamed))
+    })
+  }
+
+  val stage = new ChiselStage
+  "Ensure all annotations continue to be passed / digested correctly with the new API" - {
+    "NoDedup and DontTouch work as expected" in {
```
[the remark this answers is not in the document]
girishpai: @seldridge agreed! Did something quick and dirty by copy-pasting and modifying. Will trim down this test.

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -34,6 +42,9 @@ object annotate {
   def apply(anno: ChiselAnnotation): Unit = {
     Builder.annotations += anno
   }
```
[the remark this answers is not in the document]
girishpai: @seldridge - that seems really neat - let me give it a shot! I obviously took the brute force approach to begin with!

core/src/main/scala/chisel3/Annotation.scala
girishpai: @seldridge  don't have strong opinions on the name - I just went with something quick to reduce mental load!

core/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
girishpai: Although one prime reason for this brute force was for this initial PR to not touch / modify anything related to the older API (handle that as part of a different PR). Either ways if the suggested change does not cause a lot of disruption will go with it! Also will wait on @jackkoenig to weigh in.

core/src/main/scala/chisel3/Annotation.scala
jackkoenig: When we originally added the Annotation API, I actually argued for `() => firrtl.Annotation` as the type but was shot down in the Chisel dev meeting 🙃. That being said your slick trick for `.toRelativeTarget` won't work if we just use nullary functions so that might be a good enough reason to keep this new API as a trait. The coolest thing about your trick isn't the implicit class, it's that the `ChiselAnnotation` object can eagerly grab the current Module which is the biggest problem with `toRelativeTarget` (that grabbing the current module needs to be eager, but the .toRelativeTarget call needs to be lazy).

girishpai: ok - I admit to not going over/understanding  @seldridge  trick yet for `toRelativeTarget` - and will just agree with @jackkoenig 's expertise that the trait is the way to go!

core/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
girishpai: ok this amazing trick worked - as in compile passed - running all the tests  - lets hope it holds!

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -861,7 +861,14 @@ case class DefBlackBox(
...
 
-case class Circuit(name: String, components: Seq[Component], annotations: Seq[ChiselAnnotation], renames: RenameMap) {
-  def firrtlAnnotations: Iterable[Annotation] = annotations.flatMap(_.toFirrtl.update(renames))
-
+case class Circuit(
+  name:           String,
+  components:     Seq[Component],
+  annotations:    Seq[ChiselAnnotation],
+  renames:        RenameMap,
+  newAnnotations: Seq[ChiselToFirrtlAnnotations] = Seq.empty) {
```
jackkoenig, line 869: Adding fields to case classes is both a binary incompatible and source incompatible change. This would not be an issue if we had properly made all of these classes private.... but we didn't and [some users do notice](https://github.com/chipsalliance/chisel3/issues/2307).

To safely backport this to 3.5.x (which I presume is your plan) we either need to [use dataclass](https://github.com/chipsalliance/firrtl/pull/2071/files) to give case class-like behavior in a more friendly way for making modifications like this, or we need to add these new arguments to a 2nd parameter list and implement equals and hashcode ourselves.

core/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
girishpai: unfortunately a bunch of tests failed - taking a look (19 in total) - hopefully can find a quick solution - else will revert to the ugly code which we have.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: This is why I originally added this def, because the annotations need to be renamed anyway (for DataViews) so it's pretty important people don't access the ChiselAnnotations from the class directly.

In fact, could you deprecate the `annotations` field of this class to tell anyone who happens to be using this API to use `firrtlAnnotations` instead?

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: @jackkoenig  let me explore that - I am assuming since the `annotations` variable is publicly visible - the change suggested by @seldridge  to make `annotations`  `Seq[ChiselToFirrtlAnnotations]` instead of `Seq[ChiselAnnotation]` also is disruptive?

girishpai: > @jackkoenig let me explore that - I am assuming since the `annotations` val is publicly visible - the change suggested by @seldridge to make `annotations` `Seq[ChiselToFirrtlAnnotations]` instead of `Seq[ChiselAnnotation]` also is disruptive?

core/src/main/scala/chisel3/Annotation.scala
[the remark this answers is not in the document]
girishpai: @jackkoenig  needed your inputs - though unifying the two annotations is the clean way to do it - might be more disruptive than I thought.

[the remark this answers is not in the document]
girishpai: Discussed with @jackkoenig  - unfortunately due to backward compatibility requirements - for now we have decided to keep it this way (added a TODO to take care of it in the near future)

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: @jackkoenig  please take a look - have used dataclass - however not sure if I have handled copy properly (also do we need an unapply method? )

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: Done - added the deprecated warning on both `annotations` and `newAnnotations` just in case someone tries to use that as well. Hope that's fine.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -861,7 +862,28 @@ case class DefBlackBox(
...
+      _.toFirrtlAnnotations.flatMap(_.update(renames))
+    )
+
+  def copy(
+    name:           String = name,
+    components:     Seq[Component] = components,
+    annotations:    Seq[ChiselAnnotation] = annotations,
+    renames:        RenameMap = renames,
+    newAnnotations: Seq[ChiselToFirrtlAnnotations] = newAnnotations
+  ) = Circuit(name, components, annotations, renames, newAnnotations)
```
girishpai, line 887: @jackkoenig  not sure if we need two different versions of copy (similar to apply)

girishpai: Also not sure if we need an implementation for "unapply"

girishpai: when I tried to overload copy - it gave some error from the data macro - just FYI - did not spend too much time debugging it - if we do need two different copy versions will spend time on that.

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -20,6 +20,14 @@ trait ChiselAnnotation {
   def toFirrtl: Annotation
 }
 
+/** Enhanced interface for Annotations in Chisel
+  *
+  *  Defines a conversion to corresponding FIRRTL Annotation(s)
+  */
+trait ChiselToFirrtlAnnotations {
+  def toFirrtlAnnotations: Seq[Annotation]
```
jackkoenig, line 28:
```suggestion
trait ChiselMultiAnnotation {
  def toFirrtl: Seq[Annotation]
```

Meganit but just making the naming be more in line with the existing `ChiselAnnotation`

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: Copy should exclude any new methods (any after `@since`), also please add `"Chisel 3.5.4"` to the since just to make it obvious. You also do need to define unapply. To check if your changes are binary compatible, you can try making them against 3.5.x and then running `sbt +mimaReportBinaryIncompatibilities`

See https://github.com/chipsalliance/firrtl/blob/c7403e80f965aa3c6f65b176dbb56dbb905cbb31/src/main/scala/firrtl/ir/IR.scala#L769 for example of the copy not having any new fields and of implementing unapply yourself (again not including the new fields).

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: I responded on the `Outdated` thread: https://github.com/chipsalliance/chisel3/pull/2628#discussion_r919492572

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -861,7 +862,28 @@ case class DefBlackBox(
...
     extends Component
 
-case class Circuit(name: String, components: Seq[Component], annotations: Seq[ChiselAnnotation], renames: RenameMap) {
-  def firrtlAnnotations: Iterable[Annotation] = annotations.flatMap(_.toFirrtl.update(renames))
+@data class Circuit(
+  name:       String,
+  components: Seq[Component],
+  @deprecated("Do not use annotations val of Circuit directly - use firrtlAnnotations instead. Will be removed in a future release",
+    "Chisel 3.5")
+  annotations: Seq[ChiselAnnotation],
```
girishpai, line 870: @jackkoenig is this the right way to call deprecated? Do I need to change the version number - since I am gettnig a compile error (due to this deprecation) within our internal sifive codebase. I distinctly remember using Chisel 3.5 for something similar in the recent past and it not complaining.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: @jackkoenig did what you suggested - and I am seeing similar errors as mentioned in the PR you linked after my update - can these be ignored? 
```
error] chisel3-core: Failed binary compatibility check against edu.berkeley.cs:chisel3-core_2.12:3.5.3! Found 4 potential problems (filtered 8)
[error]  * class chisel3.internal.firrtl.Circuit is declared final in current version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("chisel3.internal.firrtl.Circuit")
[error]  * static method tupled()scala.Function1 in class chisel3.internal.firrtl.Circuit does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("chisel3.internal.firrtl.Circuit.tupled")
[error]  * static method curried()scala.Function1 in class chisel3.internal.firrtl.Circuit does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("chisel3.internal.firrtl.Circuit.curried")
[error]  * the type hierarchy of object chisel3.internal.firrtl.Circuit is different in current version. Missing types {scala.runtime.AbstractFunction4}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("chisel3.internal.firrtl.Circuit$")

```

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: Added nowarn and now compile passes.

core/src/main/scala/chisel3/Annotation.scala
girishpai: Changed!

core/src/main/scala/chisel3/internal/firrtl/IR.scala
girishpai: @jackkoenig have incorporated your changes to resolve the compatibility issues - the check is still running locally - and so far so good - will update once it ends (hopefully passes).

girishpai: It passed! 

```
sbt +mimaReportBinaryIssues
....
[success] Total time: 22 s, completed Jul 13, 2022 11:25:27 AM
[info] Setting Scala version to 2.13.7 on 6 projects.
[info] Reapplying settings...
[info] set current project to chisel3 (in build file:/scratch/girishp/vscode/clone_23jun/federation/rocket-chip/chisel3/)
[success] Total time: 0 s, completed Jul 13, 2022 11:25:28 AM
[info] Setting Scala version to 2.13.8 on 6 projects.
[info] Reapplying settings...
[info] set current project to chisel3 (in build file:/scratch/girishp/vscode/clone_23jun/federation/rocket-chip/chisel3/)
[success] Total time: 0 s, completed Jul 13, 2022 11:25:29 AM
[info] Reapplying settings...
[info] set current project to chisel3 (in build file:/scratch/girishp/vscode/clone_23jun/federation/rocket-chip/chisel3/)
```

girishpai: This is on 3.5.x ....

jackkoenig (approved):

PR #2634, opened 2022-07-21
Unify Chisel2 and chisel3 directionality

azidar:
## Background: `Chisel._` vs `chisel3._` directionality

In `Chisel._`, `IO` has an implicit output direction, and `Flipped` is used for two things:

 1) Define relative direction of a field with respect to its parent bundle
 2) To change the implicit output to an implicit input of IO

```scala
val io = IO(new Bundle { // implicit output
  val x = Flipped(UInt()) // x is flipped relative to bundle, so it is an input
  val y = UInt() // y is aligned relative to bundle, so it is an output
})
```

You can declare an input port using the (2) kind of Flipped:

```scala
val io = IO(Flipped(new Bundle { // implicit input
  val x = Flipped(UInt()) // x is flipped relative to bundle, so it is an output
  val y = UInt() // y is aligned relative to bundle, so it is an input
}))
```

In `chisel3`, `IO` does not have an implicit direction, but `Input` and `Output` are specified on types, which coerce all subfields of the type to the same direction. `Flipped` inverts the absolute direction so `Input -> Output` and `Output -> Input`.

```scala
val io = IO(new Bundle { // no implicit direction
  val x = Input(UInt()) // x is an input
  val y = Output(UInt()) // y is an output
}))
```

You can switch directions with `Flipped`:

```scala
val io = IO(Flipped(new Bundle { // Swap all absolute directions
  val x = Input(UInt()) // x is flipped so it is an output
  val y = Output(UInt()) // y is flipped so it is an input
})))
```

## Unifying directionality with new primitives

Both mechanisms of describing field directions can be unified with the following (conceptual) primitives:

- `Flipped`: a field's relative direction is reversed with respect to its parent bundle
- `Aligned`: a field's relative direction is the same with respect to its parent bundle (is implicit)
- `Outgoing`: an IO whose implicit direction is output
- `Incoming`: an IO whose implicit direction is input
- `stripFlipsOf`: a type-generator that aligns all subfields, recursively
- `reverseFlipsOf`: a type-generator that flips all subfields, recursively

Expressing Chisel._ semantics:

- `IO(new Bundle))` becomes `Outgoing(new Bundle)`
- `IO(Flipped(new Bundle))` becomes `Incoming(new Bundle)`; note that `Outgoing(reverseFlipsOf(new Bundle))` is the same, but I think it makes sense to have `Incoming` as well.
- `new Bundle { val x = UInt() }` is `new Bundle { val x = Aligned(UInt()) }`, or unchanged as `Aligned` is implicit
- `new Bundle { val x = Flipped(UInt()) }` is unchanged

Expressing chisel3._ semantics:

- `IO(new Bundle))` becomes `Outgoing(new Bundle)`
- `IO(Flipped(new Bundle))` becomes `Incoming(new Bundle)`
- `new Bundle { val x = Output(UInt()) }` is `new Bundle { val x = Aligned(stripFlipsOf(UInt())) }`
- `new Bundle { val x = Input(UInt()) }` is `new Bundle { val x = Flipped(stripFlipsOf(UInt())) }`

## How this PR unifies these things

Just like we can express Chisel/chisel3 with new primitives, we can represent the new primitives with Chisel/chisel3 APIs. The only thing that was missing was (1) ALWAYS apply the implicit output from IO (or implicit input if a flip is there), and (2) enable mixing Input/Output with Flipped. This means that effectively, `Input` becomes `Flipped(stripsFlipsOf(..))` and `Output` becomes `Aligned(stripsFlipsOf(..))`, and now the semantics can coexist. Basically, this change defines how `Chisel._`'s relative directions resolve in `chisel3`. **This *should* not break any code, because previously it was an error.**

Future work will be to add the new primitives as they are, and refactor Input/Output etc to just call the new primitives. In the meantime, we can change the semantics so that this change is easier, and we can get rid of compatibility mode.

### Contributor Checklist

- [N/A] Did you add appropriate documentation in `docs/src`? **Will do this later when we add more primitives**

#### Type of Improvement

 - new feature/API

#### API Impact

Adds an API, where you can now declare bundles that have both Flipped and Input/Output.

None

#### Desired Merge Strategy

#### Release Notes

Remove restriction that Bundles have to only declare fields with absolute directions (Input/Output) or relative directions (Flipped).

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -527,8 +527,7 @@ package experimental {
       //   even for chisel3 code.
       // This assigns the explicit directions required by both semantics on all Bundles.
       // This recursively walks the tree, and assigns directions if no explicit
-      //   direction given by upper-levels (override Input / Output) AND element is
-      //   directly inside a compatibility Bundle determined by compile options.
+      //   direction given by upper-levels (override Input / Output)
       def assignCompatDir(data: Data): Unit = {
         data match {
           case data: Element => data._assignCompatibilityExplicitDirection
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -543,9 +543,9 @@ package experimental {
                     vec.getElements.foreach(assignCompatDir(_))
                 }
               case SpecifiedDirection.Input | SpecifiedDirection.Output =>
-                // forced assign, nothing to do
-                // Note this is because Input and Output recurse down their types to align all fields to that SpecifiedDirection
-                // Thus, no implicit assigment is necessary.
+              // forced assign, nothing to do
+              // Note this is because Input and Output recurse down their types to align all fields to that SpecifiedDirection
+              // Thus, no implicit assigment is necessary.
             }
         }
       }
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -522,33 +522,35 @@ package experimental {
       */
     protected def _bindIoInPlace(iodef: Data): Unit = {
       // Compatibility code: Chisel2 did not require explicit direction on nodes
-      // (unspecified treated as output, and flip on nothing was input).
-      // This sets assigns the explicit directions required by newer semantics on
-      // Bundles defined in compatibility mode.
+      //   (unspecified treated as output, and flip on nothing was input).
+      // However, we are going to go back to Chisel2 semantics, so we need to make it work
+      //   even for chisel3 code.
+      // This assigns the explicit directions required by both semantics on all Bundles.
       // This recursively walks the tree, and assigns directions if no explicit
-      // direction given by upper-levels (override Input / Output) AND element is
-      // directly inside a compatibility Bundle determined by compile options.
+      //   direction given by upper-levels (override Input / Output) AND element is
+      //   directly inside a compatibility Bundle determined by compile options.
       def assignCompatDir(data: Data, insideCompat: Boolean): Unit = {
         data match {
-          case data: Element if insideCompat => data._assignCompatibilityExplicitDirection
-          case data: Element => // Not inside a compatibility Bundle, nothing to be done
+          case data: Element => data._assignCompatibilityExplicitDirection
           case data: Aggregate =>
             data.specifiedDirection match {
               // Recurse into children to ensure explicit direction set somewhere
               case SpecifiedDirection.Unspecified | SpecifiedDirection.Flip =>
                 data match {
                   case record: Record =>
-                    val compatRecord = !record.compileOptions.dontAssumeDirectionality
-                    record.getElements.foreach(assignCompatDir(_, compatRecord))
+                    record.getElements.foreach(assignCompatDir(_))
                   case vec: Vec[_] =>
-                    vec.getElements.foreach(assignCompatDir(_, insideCompat))
+                    vec.getElements.foreach(assignCompatDir(_))
                 }
-              case SpecifiedDirection.Input | SpecifiedDirection.Output => // forced assign, nothing to do
+              case SpecifiedDirection.Input | SpecifiedDirection.Output =>
+                // forced assign, nothing to do
+                // Note this is because Input and Output recurse down their types to align all fields to that SpecifiedDirection
+                // Thus, no implicit assigment is necessary.
             }
         }
       }
 
-      assignCompatDir(iodef, false)
+      assignCompatDir(iodef)
 
       iodef.bind(PortBinding(this))
       _ports += iodef
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -529,7 +529,7 @@ package experimental {
       // This recursively walks the tree, and assigns directions if no explicit
       //   direction given by upper-levels (override Input / Output) AND element is
       //   directly inside a compatibility Bundle determined by compile options.
-      def assignCompatDir(data: Data, insideCompat: Boolean): Unit = {
+      def assignCompatDir(data: Data): Unit = {
         data match {
           case data: Element => data._assignCompatibilityExplicitDirection
           case data: Aggregate =>
--- a/src/test/scala/chiselTests/Direction.scala
+++ b/src/test/scala/chiselTests/Direction.scala
@@ -83,15 +83,13 @@ class DirectionSpec extends ChiselPropSpec with Matchers with Utils {
     })
   }
 
-  property("Empty Vecs with no direction on the sample_element *should* cause direction errors") {
-    an[Exception] should be thrownBy extractCause[Exception] {
-      ChiselStage.elaborate(new Module {
-        val io = IO(new Bundle {
-          val foo = Input(UInt(8.W))
-          val x = Vec(0, UInt(8.W))
-        })
+  property("Empty Vecs with no direction on the sample_element should not cause direction errors, as Chisel and chisel3 directions are merged") {
+    ChiselStage.elaborate(new Module {
+      val io = IO(new Bundle {
+        val foo = Input(UInt(8.W))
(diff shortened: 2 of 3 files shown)


core/src/main/scala/chisel3/Module.scala
```diff
@@ -522,33 +522,35 @@ package experimental {
...
+                    record.getElements.foreach(assignCompatDir(_))
                   case vec: Vec[_] =>
-                    vec.getElements.foreach(assignCompatDir(_, insideCompat))
+                    vec.getElements.foreach(assignCompatDir(_))
                 }
-              case SpecifiedDirection.Input | SpecifiedDirection.Output => // forced assign, nothing to do
+              case SpecifiedDirection.Input | SpecifiedDirection.Output =>
+              // forced assign, nothing to do
+              // Note this is because Input and Output recurse down their types to align all fields to that SpecifiedDirection
+              // Thus, no implicit assigment is necessary.
```
jackkoenig, line 548:
```suggestion
                // forced assign, nothing to do
                // Note this is because Input and Output recurse down their types to align all fields to that SpecifiedDirection
                // Thus, no implicit assigment is necessary.
```
Just making it more obvious that these comments are referring to this code path

src/test/scala/chiselTests/Direction.scala
```diff
@@ -117,15 +117,15 @@ class DirectionSpec extends ChiselPropSpec with Matchers with Utils {
...
+    "Explicitly directioned but empty Bundles should not cause direction errors because Chisel and chisel3 directionality are merged"
+  ) {
+    ChiselStage.elaborate(new Module {
+      val io = IO(new Bundle {
+        val foo = UInt(8.W)
+        val x = Input(new Bundle {})
       })
-    }
+    })
   }
```
jackkoenig, line 129: Can you add a couple of tests showing the mixed semantics? For example, is the following now legal?
```scala
val io = IO(new Bundle {
  val a = Input(UInt(8.W))
  val b = Output(UInt(8.W))
  val c = UInt(8.W)
  val d = Flipped(UInt(8.W))
})
```

jackkoenig (commented):
Generally looks good but I would like to see more examples showing what is now possible.

Also should there be something in `docs/` for this?

core/src/main/scala/chisel3/Module.scala
azidar: Formatter will disagree with you :P

azidar:
So I could write docs, but I kind of want to wait until we actually add the new operators/primitives, then its easier to describe it. As is, all the docs are still technically correct, and this interoperability doesn't need to be advertised yet.

src/test/scala/chiselTests/Direction.scala
```diff
@@ -327,4 +327,73 @@ class DirectionSpec extends ChiselPropSpec with Matchers with Utils {
...
+    }
+
+    val emitted: String = ChiselStage.emitChirrtl(new MyModule)
+    val firrtl:  String = ChiselStage.convert(new MyModule).serialize
+
+    // Check that emitted directions are correct.
+    Seq(emitted, firrtl).foreach { o =>
+      {
+        // Chisel Emitter formats spacing a little differently than the
+        // FIRRTL Emitter :-(
```
jackkoenig, line 350: 1. There isn't a Chisel emitter anymore so these emit exactly the same
2. Even if they did emit slightly differently, I don't think the purpose of these tests is to test the emitters so just using 1 is fine
3. Surprisingly, this test actually passes on both [3.6-SNAPSHOT](https://scastie.scala-lang.org/JXnauH3yQryVZzRMyNXkCA) and [3.5.3](https://scastie.scala-lang.org/wTroenCmSFmAO6R7xzSncQ) -- so that is weird...

src/test/scala/chiselTests/Direction.scala
```diff
@@ -327,4 +327,73 @@ class DirectionSpec extends ChiselPropSpec with Matchers with Utils {
...
+      io.monitor.bits := io.driver.bits
+      io.monitor.valid := io.driver.valid
+      io.monitor.ready := io.driver.ready
+    }
+
+    val emitted: String = ChiselStage.emitChirrtl(new MyModule)
+    val firrtl:  String = ChiselStage.convert(new MyModule).serialize
+
+    // Check that emitted directions are correct.
+    Seq(emitted, firrtl).foreach { o =>
```
jackkoenig, line 382: Same comment as above, no reason to test the two different emission paths (which are actually the same anyway).

src/test/scala/chiselTests/Direction.scala
azidar: Conversation with @jackkoenig offline, realized this actually got introduced in 3.5.2 in the emitting firrtl bulk connect when possible update.

src/test/scala/chiselTests/Direction.scala
```diff
@@ -327,4 +327,56 @@ class DirectionSpec extends ChiselPropSpec with Matchers with Utils {
...
+    class Decoupled extends Bundle {
+      val bits = UInt(3.W)
+      val valid = Bool()
+      val ready = Flipped(Bool())
+    }
+    class DecoupledAndMonitor extends Bundle {
+      val producer = new Decoupled()
+      val consumer = Flipped(new Decoupled())
+      val monitor = Input(new Decoupled()) // Same as Flipped(stripFlipsIn(..))
+      val driver = Output(new Decoupled()) // Same as stripFlipsIn(..)
```
jackkoenig, line 360: This is a nit, but all of these names are flipped from what I would expect (due to the fact that you are flipping the use of this Bundle. That is a bit weird, but not really the end of the world.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -522,33 +522,35 @@ package experimental {
...
+      //   (unspecified treated as output, and flip on nothing was input).
+      // However, we are going to go back to Chisel2 semantics, so we need to make it work
+      //   even for chisel3 code.
+      // This assigns the explicit directions required by both semantics on all Bundles.
       // This recursively walks the tree, and assigns directions if no explicit
-      // direction given by upper-levels (override Input / Output) AND element is
-      // directly inside a compatibility Bundle determined by compile options.
-      def assignCompatDir(data: Data, insideCompat: Boolean): Unit = {
+      //   direction given by upper-levels (override Input / Output) AND element is
+      //   directly inside a compatibility Bundle determined by compile options.
```
jackkoenig, line 531:
```suggestion
      //   direction given by upper-levels (override Input / Output)
```
I think this `AND` is no longer true right?

jackkoenig (approved):
Overall LGTM, 1 suggested improvement to a comment and 1 nit

mwachs5:
Can you fill out the release notes? What is THIS PR actually doing?

azidar:
> Can you fill out the release notes? What is THIS PR actually doing?

I thought I did?

> Remove restriction that Bundles have to only declare fields with absolute directions (Input/Output) or relative directions (Flipped).

ekiwi:
@azidar Since there seem to be many ways of doing things, would you mind documenting, what the _recommended_ style is? Maybe at least update [the examples in our readme](https://github.com/chipsalliance/chisel3#what-does-chisel-code-look-like) which should probably be updated anyways to take advantage of the `MultiIO` by default feature and not declare useless anonymous Bundles.

azidar:
> Since there seem to be many ways of doing things, would you mind documenting, what the recommended style is? 

I believe your Blinky example is correct, but we haven't actually added `Incoming` or `Outgoing` (or even decided that's the right names for them). That is future work that I'd like to do. That's why right now I don't want to document anything unnecessarily because I don't think the work is ready to be user facing.

Perhaps what would help is if I wrote up an actual proposal that we can discuss?

mwachs5:
Yep release notes look good now

mwachs5:
Did we consider backporting this to 3.5? I know it's strictly a change in behavior, but its a more-things-work-instead-of-being errors which might help in upgrading.

PR #2662, opened 2022-08-05
Add OpaqueType support to Records

adkian-sifive:
`OpaqueType` allows simplifying `Record` references when the `Record` subtypes have single map elements

### Contributor Checklist

#### Type of Improvement

 - new feature/API               

#### API Impact

New override for Records for single element types

If `Record.opaqueType` is overridden to `true`, the generated CHIRRTL will substitute name of the `Record` for that of the subfield, and substitute type of the subfield for type of the `Record`

#### Desired Merge Strategy

#### Release Notes

Add new override in Record to support opaque types

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -925,11 +925,12 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
-  // Classes can override opaqueType if they want to "unbox" Records
-  //  that have maps with a single element. If opaqueType is
-  //  overridden to true, the subfield name of the record is
-  //  skipped and the Record name is used instead. See RecordSpec
-  //  for example usage and expected output
+  /** Classes can override opaqueType if they want to "unbox"
+    * Records that have maps with a single element. If opaqueType
+    * is overridden to true, the subfield name of the record is
+    * skipped and the Record name is used instead. See RecordSpec
+    * for example usage and expected output
+    */
   def opaqueType: Boolean = false
 
   // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
@@ -939,6 +940,7 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
+    assert(!opaqueType || elements.size == 1, "opaqueType cannot be true when the size of elements is not 1")
     for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true), unbox = opaqueType) }
   }
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -925,11 +925,17 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
-  /** Classes can override opaqueType if they want to "unbox"
-    * Records that have maps with a single element. If opaqueType
-    * is overridden to true, the subfield name of the record is
-    * skipped and the Record name is used instead. See RecordSpec
-    * for example usage and expected output
+  /** Indicates if this Record represents an "Opaque Type"
+    *
+    * Opaque types provide a mechanism for user-defined types 
+    * that do not impose any "boxing" overhead in the emitted FIRRTL and Verilog.
+    * You can think about an opaque type Record as a box around
+    * a single element that only exists at Chisel elaboration time.
+    * Put another way, if opaqueType is overridden to true, 
+    * The Record may only contain a single element with an empty name
+    * and there will be no `_` in the name for that element in the emitted Verilog.
+    *
+    * @see RecordSpec in Chisel's tests for example usage and expected output
     */
   def opaqueType: Boolean = false
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -940,7 +940,10 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    assert(!opaqueType || elements.size == 1, "opaqueType cannot be true when the size of elements is not 1")
+    assert(
+      !opaqueType || elements.size == 1,
+      s"Size of elements must be 1 if opaqueType == true, not ${elements.size}: ${elements.keys.mkString(", ")}"
+    )
     for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true), unbox = opaqueType) }
   }
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -934,7 +934,7 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true)) }
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true), unbox = opaqueType) }
   }
 
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -927,11 +927,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
 
   /** Indicates if this Record represents an "Opaque Type"
     *
-    * Opaque types provide a mechanism for user-defined types 
+    * Opaque types provide a mechanism for user-defined types
     * that do not impose any "boxing" overhead in the emitted FIRRTL and Verilog.
     * You can think about an opaque type Record as a box around
     * a single element that only exists at Chisel elaboration time.
-    * Put another way, if opaqueType is overridden to true, 
+    * Put another way, if opaqueType is overridden to true,
     * The Record may only contain a single element with an empty name
     * and there will be no `_` in the name for that element in the emitted Verilog.
     *
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -925,6 +925,11 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
+  // Classes can override opaqueType if they want to "unbox" Records
+  //  that have maps with a single element. If opaqueType is 
+  //  overriden to true, the subfield names of the record are
+  //  skipped, and the Record name is used instead. See RecordSpec
+  //  for example usage and expected output
   def opaqueType: Boolean = false
 
   // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -926,9 +926,9 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
   // Classes can override opaqueType if they want to "unbox" Records
-  //  that have maps with a single element. If opaqueType is 
-  //  overriden to true, the subfield names of the record are
-  //  skipped, and the Record name is used instead. See RecordSpec
+  //  that have maps with a single element. If opaqueType is
+  //  overridden to true, the subfield name of the record is
+  //  skipped and the Record name is used instead. See RecordSpec
   //  for example usage and expected output
   def opaqueType: Boolean = false
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -925,6 +925,8 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
+  def opaqueType: Boolean = false
+
   // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
   // of hardware created when connecting to one of these elements
   private def setElementRefs(): Unit = {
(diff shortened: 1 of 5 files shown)


src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +146,34 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
...
+        val out = IO(Output(new SingleElementRecord))
+
+        out := in1 + in2
+      }
+    }
+    chirrtl should include("input in1 : UInt<8>")
+    chirrtl should include("input in2 : UInt<8>")
+    chirrtl should include("add(in1, in2)")
+  }
+
```
hcook, line 176: Would `out := 0.U(8.W)` be expected to work?

hcook: (And same question for an example where the RHS is not a Literal if it matters)

adkian-sifive: I would say no cause the type of `out` is not actually being changed, only the IR being emitted after module close time will change

jackkoenig: Correct, this still enforces the Chisel type system.

hcook: to clarify, I have been assuming this case would _not_ work and implementing my own use case accordingly, but, if I ever have to do something similar again, it would save me a ton of time if this did work anywhere `opaqueType` is true

jackkoenig: > it would save me a ton of time if my above example did work

This would sort of defeat the purpose of what an `OpaqueType` is supposed to be. If you want the type to act like the type it is wrapping, then why not just use a type alias? You could also alternatively use an implicit conversion.

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -311,7 +313,11 @@ private[chisel3] object Converter {
         case (false, SpecifiedDirection.Flip | SpecifiedDirection.Input) =>
           fir.Field(getRef(elt, info).name, fir.Flip, extractType(elt, false, info))
       }
-      fir.BundleType(d.elements.toIndexedSeq.reverse.map { case (_, e) => eltField(e) })
+      if (!d.opaqueType)
+        fir.BundleType(d.elements.toIndexedSeq.reverse.map { case (_, e) => eltField(e) })
+      else
+        extractType(d.elements.head._2, true, info)
```
mwachs5, line 319: should d.elements have only one item? Should we put an assertion for that?

adkian-sifive: That's a good idea, I'll also add a unit test that should throw when user overrides `opaqueType` incorrectly

src/test/scala/chiselTests/RecordSpec.scala
hcook: You can't use an implicit conversion because `final. def :=(that: Data)` accepts all data

hcook: > then why not just use a type alias?

Because I have specific, runtime determined configuration where the number of elements is 1 (and `opaqueType` would evaluate to `true`). Only in those exact cases do i want `happensToBeOpaqueHere := UInt(8.W)` to work, in all other instances a firrtl binding error is appropriate

hcook: And again, do not consider this a blocking request, i have already worked around it by manually calling another method instead of `:=` at all call sites. But that could have been avoided.

I further bring it up because the PR message says "and substitute type of the subfield for type of the Record" and I would like to understand what that means in practice

hcook: > "only the IR being emitted after module close time will change"

is helpful in narrowing the scope of the answer but I do not really understand what that means to user code. in what situations does FIRRTL make a different decision about allowable bindings based on that change? or none?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -925,14 +925,23 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
...
+
   // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
   // of hardware created when connecting to one of these elements
   private def setElementRefs(): Unit = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true)) }
+    assert(!opaqueType || elements.size == 1, "opaqueType cannot be true when the size of elements is not 1")
```
mwachs5, line 943:
```suggestion
    assert(!opaqueType || elements.size == 1, s"Size of elements must be 1 if opaqueType =1 true, not ${elements.size}: ${elements.keys.mkString(",")} ")
```

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +146,53 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
     e.getMessage should include("contains aliased fields named (bar,foo)")
   }
 
+  they should "be OpaqueType for single element maps" in {
+    class SingleElementRecord extends Record {
+      private val underlying = UInt(8.W)
+      val elements = SeqMap("" -> underlying)
```
mwachs5, line 152: is it important that the string be empty? What if I put ("unused" -> underlying)? Can we write a test for that case?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -925,14 +925,26 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
...
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true)) }
+    assert(
+      !opaqueType || elements.size == 1,
+      s"Size of elements must be 1 if opaqueType == true, not ${elements.size}: ${elements.keys.mkString(", ")}"
+    )
+    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true), unbox = opaqueType) }
```
adkian-sifive, line 947:
```suggestion
    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true), opaque = opaqueType) }
```

src/test/scala/chiselTests/RecordSpec.scala
adkian-sifive: So the way this is implemented in this PR is by handling `OpaqueType` as a special case of `Record` - there's a custom type under the hood that does nothing other than note that a `Record` instance needs to be represented differently in the IR. This means that `OpaqueType` is not fully a native type (yet), and I think while it's _possible_ to support the monoconnect use case, that would mean adding special handling for that in connects which would currently be very hacky IMO

adkian-sifive: Also I think the monoconnect case might fall naturally once we have more ingrained support for `OpaqueType`

hcook: > which would currently be very hacky IMO
 thanks for the explanation, sounds best to wait for it

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -925,14 +925,26 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
...
+
   // Doing this earlier than onModuleClose allows field names to be available for prefixing the names
   // of hardware created when connecting to one of these elements
   private def setElementRefs(): Unit = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true)) }
+    assert(
```
jackkoenig, line 943:
```suggestion
    require(
```
`require` is used for indicating the user has made a mistake, as opposed to `assert` which usually means the developer has made a mistake.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -925,14 +925,26 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record(private[chisel3] implicit val compileOptions: CompileOptions) extends Aggregate {
 
+  /** Classes can override opaqueType if they want to "unbox"
+    * Records that have maps with a single element. If opaqueType
+    * is overridden to true, the subfield name of the record is
+    * skipped and the Record name is used instead. See RecordSpec
+    * for example usage and expected output
+    */
```
jackkoenig, line 933:
```suggestion
  /** Indicates if this Record represents an "Opaque Type"
    *
    * Opaque types provide a mechanism for user-defined types 
    * that do not impose any "boxing" overhead in the emitted FIRRTL and Verilog.
    * You can think about an opaque type Record as a box around
    * a single element that only exists at Chisel elaboration time.
    * Put another way, if opaqueType is overridden to true, 
    * The Record may only contain a single element with an empty name
    * and there will be no `_` in the name for that element in the emitted Verilog.
    *
    * @see RecordSpec in Chisel's tests for example usage and expected output
    */
```

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -925,14 +925,26 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
...
   // of hardware created when connecting to one of these elements
   private def setElementRefs(): Unit = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
     // which can cause collisions
     val _namespace = Namespace.empty
-    for ((name, elt) <- elements) { elt.setRef(this, _namespace.name(name, leadingDigitOk = true)) }
+    assert(
+      !opaqueType || elements.size == 1,
+      s"Size of elements must be 1 if opaqueType == true, not ${elements.size}: ${elements.keys.mkString(", ")}"
```
jackkoenig, line 945:
```suggestion
      !opaqueType || (elements.size == 1 && elements.head._1 == ""),
      s"Opaque types must have exactly one element with an empty name, not ${elements.size}: ${elements.keys.mkString(", ")}
```

src/test/scala/chiselTests/RecordSpec.scala
jackkoenig: The name should be required to be empty, I've added this to my comments

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +146,70 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
...
+    }).getMessage should include("Size of elements must be 1 if opaqueType == true")
+
+    val namedElementChirrtl = ChiselStage.emitChirrtl {
+      new Module {
+        val in = IO(Input(new NamedSingleElementRecord))
+        val out = IO(Output(new SingleElementRecord))
+        out := in
+      }
+    }
+    namedElementChirrtl should include("input in : UInt<8>")
```
jackkoenig, line 210: This particular test should fail.

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +146,70 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
     e.getMessage should include("contains aliased fields named (bar,foo)")
   }
 
+  they should "be OpaqueType for single element maps" in {
```
jackkoenig, line 149: There are lots of tests in here, can you split them up between different named blocks describing what each test is checking?

jackkoenig (changes requested):
Implementation looks good! Some suggested improvements.

jackkoenig (approved):

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +196,25 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
     e.getMessage should include("contains aliased fields named (bar,foo)")
   }
 
+  they should "be OpaqueType for maps with single unnamed elements" in {
+    val singleElementChirrtl = ChiselStage.emitChirrtl { new SingleElementRecordModule }
+    singleElementChirrtl should include("input in1 : UInt<8>")
+    singleElementChirrtl should include("input in2 : UInt<8>")
+    singleElementChirrtl should include("add(in1, in2)")
+  }
```
jackkoenig, line 204: I forgot, can you also add something checking the `.toTarget` of the opaque type? You can see some examples here: https://github.com/chipsalliance/chisel3/blob/0dd0d06445b97565ad0ee985f3510ffd54aa7971/src/test/scala/chiselTests/ToTargetSpec.scala

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -108,6 +108,61 @@ trait RecordSpecUtils {
...
+
+  class ErroneousOverrideModule extends Module {
+    val in = IO(Input(new ErroneousOverride))
+    val out = IO(Output(new ErroneousOverride))
+    out := in
+  }
+
+  var m: SingleElementRecordModule = _
+  ChiselStage.elaborate { m = new SingleElementRecordModule; m }
```
jackkoenig, line 165: You generally don't want to leave artifacts as fields of ScalaTest specs, you should just elaborate this in the scope of each test that needs it.

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -108,6 +108,60 @@ trait RecordSpecUtils {
...
+    override def cloneType: this.type = this
+  }
+
+  class ErroneousOverrideModule extends Module {
+    val in = IO(Input(new ErroneousOverride))
+    val out = IO(Output(new ErroneousOverride))
+    out := in
+  }
+
+  var m: SingleElementRecordModule = _
```
jackkoenig, line 164:
```suggestion
```
It's still weird to have this var, I'm going to inline it into each place that needs it.

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +200,36 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
...
+  }
+
+  they should "throw an error when map contains more than one element and opaqueType is overriden to true" in {
+    (the[Exception] thrownBy extractCause[Exception] {
+      ChiselStage.elaborate { new ErroneousOverrideModule }
+    }).getMessage should include("Opaque types must have exactly one element with an empty name")
+  }
+
+  they should "work with .toTarget" in {
+    ChiselStage.elaborate { m = new SingleElementRecordModule; m }
```
jackkoenig, line 223:
```suggestion
    var m: SingleElementRecordModule = _
    ChiselStage.elaborate { m = new SingleElementRecordModule; m }
```

src/test/scala/chiselTests/RecordSpec.scala
```diff
@@ -146,6 +200,36 @@ class RecordSpec extends ChiselFlatSpec with RecordSpecUtils with Utils {
...
+    val q = m.in1.toTarget.toString
+    assert(q == "~SingleElementRecordModule|SingleElementRecordModule>in1")
+  }
+
+  they should "NOT work with .toTarget on non-data OpaqueType Record" in {
+    ChiselStage.elaborate { m = new SingleElementRecordModule; m }
```
jackkoenig, line 229:
```suggestion
    var m: SingleElementRecordModule = _
    ChiselStage.elaborate { m = new SingleElementRecordModule; m }
```

jackkoenig (approved):
LGTM except one minor tweak

jackkoenig:
Github automerged this even though CI failed.... that's weird. I'll push a fix

PR #2663, opened 2022-08-08
Printables for verification preconditions

adkian-sifive:
Continuation of #2527 

Additions include unit test for `assume` and Scaladoc group for internal macros
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -20,10 +20,6 @@ object assert {
     * reset). If your definition of reset is not the encapsulating Module's
     * reset, you will need to gate this externally.
     *
-    * May be called outside of a Module (like defined in a function), so
-    * functions using assert make the standard Module assumptions (single clock
-    * and single reset).
-    *
     * @param cond condition, assertion fires (simulation fails) when false
     * @param message optional format string to print when the assertion fires
     * @param data optional bits to print in the message formatting
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -11,7 +11,21 @@ import chisel3.internal.sourceinfo.SourceInfo
 
 import scala.reflect.macros.blackbox
 
-object assert {
+/** Scaladoc information for internal verification statement macros
+  * that are used in objects assert, assume and cover.
+  *
+  * @groupdesc VerifPrintMacros
+  *
+  * <p>
+  * '''These internal methods are not part of the public-facing API!'''
+  * </p>
+  * <br>
+  *
+  * @groupprio VerifPrintMacros 1001
+  */
+trait VerifPrintMacrosDoc
+
+object assert extends VerifPrintMacrosDoc {
 
   /** Checks for a condition to be valid in the circuit at rising clock edge
     * when not in reset. If the condition evaluates to false, the circuit
@@ -76,6 +90,7 @@ object assert {
 
   import VerificationStatement._
 
+  /** @group VerifPrintMacros */
   @deprecated(
     "This method has been deprecated in favor of _applyMacroWithStringMessage. Please use the same.",
     "Chisel 3.5"
@@ -93,15 +108,6 @@ object assert {
     q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo, $compileOptions)"
   }
 
-  /** Internal verification statement macros.
-    *
-    * @groupdesc VerifPrintMacros
-    *
-    * <p>
-    * '''These internal methods are not part of the public-facing API!'''
-    * </p>
-    * <br>
-    */
   /** @group VerifPrintMacros */
   def _applyMacroWithStringMessage(
     c:              blackbox.Context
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -41,7 +41,35 @@ object assert {
   )(
     implicit sourceInfo: SourceInfo,
     compileOptions:      CompileOptions
-  ): Assert = macro _applyMacroWithMessage
+  ): Assert = macro _applyMacroWithStringMessage
+
+  /** Checks for a condition to be valid in the circuit at all times. If the
```
ekiwi, line 46: First of all thanks for adding this scaladoc!

Some small suggestions for improvement:

> Checks for a condition to be valid in the circuit at all times.

Technically the condition is only checked at a rising clock edge. If there was an asynchronous signal being fed into the assert, then a change in that signal might be missed.

> Does not fire when in reset (defined as the encapsulating Module's reset).
I would suggest to remove the explanation in parenthesis since the reset could also be changed through `withReset`.

One other thing that might be worth adding would be a short explanation of the behavior when used not with simulation, but with a bounded model checker like Jasper Gold or the formal backend in chiseltest.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -41,7 +41,35 @@ object assert {
...
+    * condition evaluates to false, the circuit simulation stops with an error.
+    *
+    * Does not fire when in reset (defined as the encapsulating Module's
+    * reset). If your definition of reset is not the encapsulating Module's
+    * reset, you will need to gate this externally.
```
mwachs5, line 51:
```suggestion
    * Does not fire when in reset (defined as the current implicit reset, e.g. as set by
    * the enclosing `withReset` or Module.reset.
```

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -41,7 +41,35 @@ object assert {
...
+    *
+    * May be called outside of a Module (like defined in a function), so
+    * functions using assert make the standard Module assumptions (single clock
+    * and single reset).
```
mwachs5, line 55: is this necessary to say? There are lots of things that can be called outside of a Module such as `Reg(...)` and we do not clarify those things

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -41,7 +41,35 @@ object assert {
...
+    *
+    * @param cond condition, assertion fires (simulation fails) when false
```
mwachs5, line 57:
```suggestion
    * @param cond condition, assertion fires (simulation fails) on a rising clock edge when false and reset is not asserted
```

core/src/main/scala/chisel3/VerificationStatement.scala
adkian-sifive: Would it be better to change the reset condition doc to also add `withReset`? Something like

> Does not fire when in reset, where reset is defined as the encapsulating Module's reset or that of the encapsulating `withReset` block

ekiwi: > Would it be better to change the reset condition doc to also add `withReset`? Something like

Personally I feel like having a single place where we explain how resets work in Chisel would be preferable. Then for specific functions we can just write something like `Does not fire when reset is active.` with maybe a link to the central explanation of how implicit `reset` works in Chisel.

adkian-sifive: Agreed

jackkoenig:
The ScalaDoc still leaves much to be desired (see screenshot below). Compare it to how things are grouped on `UInt` (https://javadoc.io/doc/edu.berkeley.cs/chisel3_2.13/latest/chisel3/UInt.html).
Can we:
1. Make sure every `_applyMacro...` function is in the `VerifPrintMacros` group?
2. Ensure the warning description of `VerifPrintMacros` shows up?
3. Put `VerifPrintMacros` below the "Ungrouped" group?

Screen Shot 2022-08-10 at 4 56 31 PM

src/test/scala/chiselTests/Assert.scala
```diff
@@ -61,6 +61,29 @@ class BadUnescapedPercentAssertTester extends BasicTester {
   stop()
 }
 
+class PrintableFormattedAssertTester extends BasicTester {
+  val foobar = Wire(UInt(32.W))
+  foobar := 123.U
+  assert(foobar === 123.U, p"Error! Wire foobar =/= ${Hexadecimal(foobar)} This is 100%% wrong.\n")
```
jackkoenig, line 67:
```suggestion
  assert(foobar === 123.U, cf"Error! Wire foobar =/= $foobar%x This is 100%% wrong.\n")
```

src/test/scala/chiselTests/Assert.scala
```diff
@@ -61,6 +61,29 @@ class BadUnescapedPercentAssertTester extends BasicTester {
...
+  stop()
+}
+
+class PrintableAssumeTester extends Module {
+  val in = IO(Input(UInt(8.W)))
+  val out = IO(Output(UInt(8.W)))
+
+  val w = Wire(UInt(8.W))
+  w := 255.U
+  assume(w === 255.U, p"Assumption failed, Wire w =/= ${Hexadecimal(255.U)}")
```
jackkoenig, line 82:
```suggestion
  assume(w === 255.U, cf"Assumption failed, Wire w =/= $w%x")
```

jackkoenig (changes requested):
My main complaint is about the ScalaDoc--I also made a couple of minor suggestions to the tests to use `cf"..."` instead of `p"..."`. Otherwise looks great, just need to refine the ScalaDoc a little bit!

jackkoenig:
ScalaDoc looks great, thanks!
Screen Shot 2022-08-13 at 10 00 44 AM

jackkoenig (approved):

PR #2671, opened 2022-08-12
Add ability to suppress enum cast warnings

zyedidia:
Casting a UInt to an Enum causes an unsuppressable warning. In some cases the designer can guarantee that the UInt is in fact a valid enum state (for example the UInt was created by casting an Enum value, or has been otherwise verified). In those cases it should be possible for the designer to suppress the warning. Sometimes a cast happens through another function like `asTypeOf`, so using `.safe` to avoid the warning is not practical.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This adds a new object for suppressing enum cast warnings.

No change.

#### Desired Merge Strategy

Squash.

#### Release Notes

The `suppressEnumCastWarning` object can be used to prevent a block from generating enum cast warnings.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -413,7 +413,7 @@ private object UnsafeEnum extends EnumFactory
 object suppressEnumCastWarning {
 
   /**
-    * Execute @block while suppressing enum cast warnings.
+    * Execute block while suppressing enum cast warnings.
     */
   def apply[T](block: => T): T = {
     val parentWarn = Builder.suppressEnumCastWarning
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -410,11 +410,24 @@ private[chisel3] class UnsafeEnum(override val width: Width) extends EnumType(Un
 }
 private object UnsafeEnum extends EnumFactory
 
+/** Suppress enum cast warnings
+ *
+ * Users should use [[EnumFactory.safe <EnumType>.safe]] when possible.
+ *
+ * This is primarily used for casting from [[UInt]] to a Bundle type that contains an Enum.
+ * {{{
+ * class MyBundle extends Bundle {
+ *   val addr = UInt(8.W)
+ *   val op = OpEnum()
+ * }
+ *
+ * // Since this is a cast to a Bundle, cannot use OpCode.safe                                                                                    
+ * val bundle = suppressEnumCastWarning {
+ *   someUInt.asTypeOf(new MyBundle)
+ * }
+ * }}}
+ */
 object suppressEnumCastWarning {
-
-  /**
-    * Execute block while suppressing enum cast warnings.
-    */
   def apply[T](block: => T): T = {
     val parentWarn = Builder.suppressEnumCastWarning
 
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -411,22 +411,22 @@ private[chisel3] class UnsafeEnum(override val width: Width) extends EnumType(Un
 private object UnsafeEnum extends EnumFactory
 
 /** Suppress enum cast warnings
- *
- * Users should use [[EnumFactory.safe <EnumType>.safe]] when possible.
- *
- * This is primarily used for casting from [[UInt]] to a Bundle type that contains an Enum.
- * {{{
- * class MyBundle extends Bundle {
- *   val addr = UInt(8.W)
- *   val op = OpEnum()
- * }
- *
- * // Since this is a cast to a Bundle, cannot use OpCode.safe                                                                                    
- * val bundle = suppressEnumCastWarning {
- *   someUInt.asTypeOf(new MyBundle)
- * }
- * }}}
- */
+  *
+  * Users should use [[EnumFactory.safe <EnumType>.safe]] when possible.
+  *
+  * This is primarily used for casting from [[UInt]] to a Bundle type that contains an Enum.
+  * {{{
+  * class MyBundle extends Bundle {
+  *   val addr = UInt(8.W)
+  *   val op = OpEnum()
+  * }
+  *
+  * // Since this is a cast to a Bundle, cannot use OpCode.safe                                                                                    
+  * val bundle = suppressEnumCastWarning {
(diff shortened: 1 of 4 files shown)


ekiwi:
> Sometimes a cast happens through another function like `asTypeOf`, so using `.safe` to avoid the warning is not practical.

Could you show us some example source code where using `safe` is not practical?

Bluntly turning off all the warnings seems like it could lead to more bugs.

zyedidia:
Sure, here is a API that is similar to MuxLookup that I use for decoders:

```scala
object DecodeLookup {
  def apply[T <: Bundle](b: T, addr: UInt, default: T, mapping: Array[(BitPat, T)]): T = {
    val mapBits = mapping.map(x => (x._1 -> BitPat(x._2.litValue.U(x._2.getWidth.W))))
    val defBits = BitPat(default.litValue.U(default.getWidth.W))
    decoder(addr, decode.TruthTable(mapBits, defBits, false)).asTypeOf(b)
  }
}
```

This converts literal bundles to UInts for use in a truth table so that logic optimization can be applied, and then converts the UInts back to the original bundle type. If the bundle type contains Enums (as it often does), the `asTypeOf` causes warnings, and I don't see how to easily use `.safe` to avoid it. This conversion is in fact safe though because all the UInts were produced by converting bundles, which are guaranteed to have valid enum states.

I'd also like to note that this isn't bluntly turning off all warnings, this is allowing the designer to turn off one warning for a specific scope.

Perhaps it's also worth discussing whether there should be a more general warning suppression mechanism that works at the scope level. Maybe something like Scala's `@nowarn` that can match based on a regular expression.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -408,6 +409,8 @@ private[chisel3] object Builder extends LazyLogging {
     dynamicContextVar.value.get
   }
 
+  var suppressEnumCastWarning: Boolean = false
```
ekiwi, line 412: This should get a small comment on what this is used for.

src/test/scala/chiselTests/StrongEnum.scala
```diff
@@ -514,6 +515,23 @@ class StrongEnumSpec extends ChiselFlatSpec with Utils {
   it should "correctly check if the enumeration is one of the values in a given sequence" in {
     assertTesterPasses(new IsOneOfTester)
   }
+
+  it should "suppress warning using suppressEnumCastWarning" in {
```
ekiwi, line 519: Could you also add a test where one warning is suppressed and the other is not. Then you can check that exactly one warning is generated.

ekiwi (approved):
Thanks for the explanation. LGTM. Just two small improvements before we merge this.

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -409,3 +409,20 @@ private[chisel3] class UnsafeEnum(override val width: Width) extends EnumType(Un
   override def cloneType: this.type = new UnsafeEnum(width).asInstanceOf[this.type]
 }
 private object UnsafeEnum extends EnumFactory
+
+object suppressEnumCastWarning {
+
+  /**
+    * Execute @block while suppressing enum cast warnings.
+    */
```
jackkoenig, line 417: Is this valid ScalaDoc? What does it look like?

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -409,3 +409,20 @@ private[chisel3] class UnsafeEnum(override val width: Width) extends EnumType(Un
   override def cloneType: this.type = new UnsafeEnum(width).asInstanceOf[this.type]
 }
 private object UnsafeEnum extends EnumFactory
+
+object suppressEnumCastWarning {
```
jackkoenig, line 413:
```suggestion
/** Suppress enum cast warnings
  *
  * Users should use [[EnumFactory.safe <EnumType>.safe]] when possible.
  *
  * This is primarily used for casting from [[UInt]] to a Bundle type that contains an Enum.
  * {{{
  * class MyBundle extends Bundle {
  *   val addr = UInt(8.W)
  *   val op = OpEnum()
  * }
  *
  * // Since this is a cast to a Bundle, cannot use OpCode.safe                                                                                    
  * val bundle = suppressEnumCastWarning {
  *   someUInt.asTypeOf(new MyBundle)
  * }
  * }}}
  */
object suppressEnumCastWarning {
```
The best place for ScalaDoc is on the object because that's the first thing you see when you open the doc page, here's a little example of using this

jackkoenig:
It would also be good to add a little mini blurb here: https://github.com/chipsalliance/chisel3/blob/master/docs/src/explanations/chisel-enum.md#casting

core/src/main/scala/chisel3/StrongEnum.scala
```diff
@@ -409,3 +409,33 @@ private[chisel3] class UnsafeEnum(override val width: Width) extends EnumType(Un
...
+  * Users should use [[EnumFactory.safe <EnumType>.safe]] when possible.
+  *
+  * This is primarily used for casting from [[UInt]] to a Bundle type that contains an Enum.
+  * {{{
+  * class MyBundle extends Bundle {
+  *   val addr = UInt(8.W)
+  *   val op = OpEnum()
+  * }
+  *
+  * // Since this is a cast to a Bundle, cannot use OpCode.safe
```
jackkoenig, line 424:
```suggestion
  * // Since this is a cast to a Bundle, cannot use OpCode.safe
```
This is probably my fault in copy-paste but there's a bunch of whitespace on the end of this line

docs/src/explanations/chisel-enum.md
```diff
@@ -167,6 +168,25 @@ val (log2, _) = grabLog(ChiselStage.emitChirrtl(new SafeFromUInt))
...
+
+class SuppressedFromUInt extends Module {
+  val in = IO(Input(UInt(15.W)))
+  val out = IO(Output(new MyBundle()))
+  suppressEnumCastWarning {
+    out := in.asTypeOf(new MyBundle)
+  }
+}
+```
+
```
jackkoenig, line 189:
```scala mdoc:invisible

val (log3, _) = grabLog(ChiselStage.emitChirrtl(new SuppressedFromUInt))
assert(log3.isEmpty)
```

Sorry, one last suggestion but it would be nice to actually check that there's no when generating this, blending testing with documentation but still nice to check the example.

jackkoenig (approved):

PR #2673, opened 2022-08-12
Show equivalent warnings/errors only once

zyedidia:
Sometimes one line of source code can end up generating the same warning or error multiple times. With this change, multiple equivalent warnings/errors generated from the same source code location will only be shown once. This de-duplication uses the same approach as the de-duplication that currently exists for deprecation warnings.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

Multiple equivalent warnings generated from the same location will now only be shown once.

No codegen change.

#### Desired Merge Strategy

Squash

#### Release Notes

Multiple equivalent warnings generated from the same location will now only be shown once.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Error.scala
+++ b/core/src/main/scala/chisel3/internal/Error.scala
@@ -177,21 +177,24 @@ private[chisel3] object ErrorLog {
 }
 
 private[chisel3] class ErrorLog {
-
-  def getLoc: String = {
-    getUserLineNumber match {
+  def getLoc(loc: Option[StackTraceElement]): String = {
+    loc match {
       case Some(elt: StackTraceElement) => s"${elt.getFileName}:${elt.getLineNumber}"
       case None => "(unknown)"
     }
   }
 
   /** Log an error message */
-  def error(m: => String): Unit =
-    errors += (((m, getLoc), new Error(m, getUserLineNumber)))
+  def error(m: => String): Unit = {
+    val loc = getUserLineNumber
+    errors += (((m, getLoc(loc)), new Error(m, loc)))
+  }
 
   /** Log a warning message */
-  def warning(m: => String): Unit =
-    errors += (((m, getLoc), new Warning(m, getUserLineNumber)))
+  def warning(m: => String): Unit = {
+    val loc = getUserLineNumber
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -178,18 +178,26 @@ private[chisel3] object ErrorLog {
...
+    getUserLineNumber match {
+      case Some(elt: StackTraceElement) => s"${elt.getFileName}:${elt.getLineNumber}"
+      case None => "(unknown)"
+    }
+  }
+
   /** Log an error message */
   def error(m: => String): Unit =
-    errors += new Error(m, getUserLineNumber)
+    errors += (((m, getLoc), new Error(m, getUserLineNumber)))
```
jackkoenig, line 190: It would be better if we didn't call `getUserLineNumber` here twice since it involves getting a stack trace (which is expensive). Can `getLoc` instead accept `Option[StackTraceElement]` and just do the filtering on that and then we can call `getUserLineNumber` just once?

src/test/scala/chiselTests/WarningSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+  "Warnings" should "be de-duplicated" in {
+    object MyEnum extends ChiselEnum {
+      val e0, e1, e2 = Value
+    }
+
+    class MyModule extends Module {
+      val in = IO(Input(UInt(2.W)))
+      val out1 = IO(Output(MyEnum()))
+      val out2 = IO(Output(MyEnum()))
+      out1 := MyEnum(in); out2 := MyEnum(in)
```
jackkoenig, line 23:
```suggestion
      def func(out: MyEnum): Unit = {
        out := MyEnum(in)
      }
      func(out1)
      func(out2)
```
While the two warnings on the same line are deduplicated (due to only line number being provided in the source locator here), that may not always be the case (if we were to start using `SourceInfo` in warnings). More importantly though, I think it doesn't show the intended use case which is warnings that do point to the same exact thing which more often comes up from calling a utility multiple times.

jackkoenig (approved):
LGTM! This is great and I'm not sure why we didn't think of this years ago 😂

PR #2700, opened 2022-08-24
Eagerly throw error if certain annotations are used on non-hardware types

zyedidia:
Using `forceName` or `DecodeTable` on a non-hardware value now throws an error at the call-site instead of later in the pipeline when source locator information is no longer available. I will also open a separate pull request to backport this to 3.5.x with a deprecation warning instead of an error.

### Contributor Checklist

#### Type of Improvement

- code cleanup

#### API Impact
Using `forceName` or `DecodeTable` on a non-hardware value now throws an error at the call-site instead of later in the pipeline.

None.

#### Desired Merge Strategy

#### Release Notes

Using `forceName` or `DecodeTable` on a non-hardware value now throws an error at the call-site instead of later in the pipeline.
- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/experimental/decode/decoder.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/decoder.scala
@@ -6,7 +6,6 @@ import chisel3._
 import chisel3.experimental.{annotate, ChiselAnnotation}
 import chisel3.util.{pla, BitPat}
 import chisel3.util.experimental.{getAnnotations, BitSet}
-import chisel3.internal.Builder
 import firrtl.annotations.Annotation
 import logger.LazyLogging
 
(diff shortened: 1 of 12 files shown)


src/main/scala/chisel3/util/experimental/decode/decoder.scala
```diff
@@ -30,6 +31,7 @@ object decoder extends LazyLogging {
       val (plaInput, plaOutput) =
         pla(minimizedTable.table.toSeq, BitPat(minimizedTable.default.value.U(minimizedTable.default.getWidth.W)))
 
+      if (!plaOutput.isSynthesizable) Builder.error(s"Using DecodeTableAnnotation on non-hardware value $plaOutput")
```
jackkoenig, line 34: Since this is checking the output of one of our own APIs rather than user input, I don't think the standard `Builder.error` is the right check here (that's more for user-error). I think an `assert` would be better just as a sanity check.

jackkoenig (approved):
Looks great, just 1 suggested improvement.

PR #2706, opened 2022-08-25
Add lexical scope checks to Assert, Assume and Printf

adkian-sifive:
When passing a `Printable` to `assert`, `assume` or `printf`, the scope of where the `Printable` was constructed should be the same as where it is being referenced. This is currently not checked, which makes it possible to construct cases where invalid FIRRTL is generated. Specifically, if the `Printable` is constructed in an inner module and referenced from an outer module, the resulting FIRRTL can have references to objects like wires or temporaries that are local to the inner module. 

For example, the following snippet

```
class InnerPrintable extends Module {
  val in = IO(Input(UInt(8.W)))

  val w = Wire(UInt(8.W))
  w := 255.U
  val p = p"$w"

  out := in
}

ChiselStage.emitChirrtl { 
  new Module {
    val mod = Module(new InnerPrintable)
    printf(mod.p)
  }
}
```
outputs the problematic FIRRTL `printf`

```
  module InnerPrintable:
    input clock : Clock
    ...

    wire w : UInt<8>
    w <= UInt<8>("hff")
    ...

module Outer_Anon :
    input clock : Clock
    ...

    node _T = bits(reset, 0, 0)
    node _T_1 = eq(_T, UInt<1>("h0"))
    when _T_1 : 
      printf(clock, UInt<1>("h1"), "%d", w) : printf
```
where the wire `w` internal to `InnerPrintable` is being accessed outside of its scope.

This PR adds checks to `assert`, `assume` and `printf` to ensure that the Builder scopes at `Printable` construction and reference are the same.

### Contributor Checklist

#### Type of Improvement

  - bug fix                            

#### API Impact

N/A

Printfs, asserts and assumes will not be present in the generated Verilog if the Printable scopes are invalid.

#### Desired Merge Strategy

#### Release Notes

Printables (and other objects?) used in printf, assert and assume can now only be referenced in the scopes in which they were constructed

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -147,7 +147,7 @@ object Printable {
     }
 
     message match {
-      case Some(x) => getData(x).map(_.requireVisible())
+      case Some(x) => getData(x).foreach(_.requireVisible())
       case _       =>
     }
   }
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -45,6 +45,8 @@ import java.util.{MissingFormatArgumentException, UnknownFormatConversionExcepti
 // TODO Should we provide more functions like map and mkPrintable?
 sealed abstract class Printable {
 
+  val context = internal.Builder.currentModule
+
   /** Unpack into format String and a List of String arguments (identifiers)
     * @note This must be called after elaboration when Chisel nodes actually
     *   have names
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -45,8 +45,6 @@ import java.util.{MissingFormatArgumentException, UnknownFormatConversionExcepti
 // TODO Should we provide more functions like map and mkPrintable?
 sealed abstract class Printable {
 
-  val context = internal.Builder.currentModule
-
   /** Unpack into format String and a List of String arguments (identifiers)
     * @note This must be called after elaboration when Chisel nodes actually
     *   have names
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -135,7 +135,7 @@ object Printable {
     StringContext(bufEscapeBackSlash.toSeq: _*).cf(data: _*)
   }
 
-  private[chisel3] def checkScope(message: Option[Printable]): Unit = {
+  private[chisel3] def checkScope(message: Printable): Unit = {
     def getData(x: Printable): Seq[Data] = {
       x match {
         case y: FirrtlFormat => Seq(y.bits)
@@ -145,11 +145,7 @@ object Printable {
         case _             => Seq() // Handles subtypes PString and Percent
       }
     }
-
-    message match {
-      case Some(x) => getData(x).foreach(_.requireVisible())
-      case _       =>
-    }
+    getData(message).foreach(_.requireVisible())
   }
 }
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/chisel3/Printable.scala
```diff
@@ -45,6 +45,8 @@ import java.util.{MissingFormatArgumentException, UnknownFormatConversionExcepti
 // TODO Should we provide more functions like map and mkPrintable?
 sealed abstract class Printable {
 
+  val context = internal.Builder.currentModule
```
jackkoenig, line 48: The context/lexical scope we need to check is not a property of the `Printable` object, rather it's a property of the `Data` boxed by various `Printables` (eg. `Decimal`). Thus `Printable` itself should not be capturing the Module in which it was created. Furthermore, lexical scope also needs to take `when` scoping into account, see this example Scastie (https://scastie.scala-lang.org/yJB6k07dSm6rTian91M1rA)

As a straw man, I think what you want to do is have a [package private] method (maybe `def checkLexicalScope(moduleContext: RawModule, whenContext: Option[WhenContext]): Unit`) that recurses through the `Printable` structure and checks the lexical scope of each `Data`, reporting an error if the scope is violated.

jackkoenig (changes requested):
This is a great idea but the implementation needs to be more about checking the lexical scope of the individual Data within a `Printable` rather than of the `Printable` objects themselves.

core/src/main/scala/chisel3/Printable.scala
jackkoenig, line 48: Also I think this is unused now right?

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -25,7 +25,26 @@ import scala.reflect.macros.blackbox
   */
 private[chisel3] trait VerifPrintMacrosDoc
 
-object assert extends VerifPrintMacrosDoc {
+private[chisel3] trait ScopeCheck {
+  def checkScope(message: Option[Printable]): Unit = {
```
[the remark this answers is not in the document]
adkian-sifive: If i just take it out the printables with the `message` field None like `p"ok"` error with None.get - I'd have to update the callsites as `checkScope(message.getOrElse(None))` or something which seems more verbose?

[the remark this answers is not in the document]
jackkoenig: You could make the callsite just be `message.foreach(checkScope(_))`

src/test/scala/chiselTests/Assert.scala
```diff
@@ -84,6 +84,78 @@ class PrintableAssumeTester extends Module {
...
+
+class AssertPrintablePortScope extends BasicTester {
+  ChiselStage.elaborate {
+    new Module {
+      val mod = Module(new PrintableScopeTester)
+      assert(1.U === 2.U, mod.printablePort)
+    }
+  }
+  stop()
+}
```
jackkoenig, line 113:
```suggestion
class AssertPrintablePortScope extends BasicTester {
  val mod = Module(new PrintableScopeTester)
  assert(1.U === 2.U, mod.printablePort)
  stop()
}
```
`BasicTester` **is** a Module, it's intended to be a Chisel TestBench testing Chisel inside of it. If you run `elaborate` in side of it, you're doing a separate Chisel elaboration that throws away whatever you built. Thus these tests aren't testing what you think they are unless you just actually instantiate the stuff you want normally as I've shown in this suggestion.

Please fix the tests below as well.

jackkoenig (approved):
LGTM!

PR #2713, opened 2022-08-30
Added :<>=, :<=, :>=, :#=

azidar:
### Background

In general, Chisel's connection operators are very broken; their semantics change depending on `import Chisel._` vs `import chisel3._`. Even `chisel3._`'s semantics are not clean; you cannot reason about what will be connected to what unless you know whether the things you are connecting are Wires or IOs. Part of this stems from the fact that we got rid of relative-flippedness for declaring things, instead relying on using Input/Output.

This PR defines four new operators which have very clean semantics which depend solely on the hardware type of the things you are connecting. Using them, we plan to migrate all existing operators semantics to compositions of them, but for now we define them independently of Chisel/chisel3, and thus they have more verbose syntax to make them compatible with both imports.

As we continue, we may rename them etc., but their semantics are composable which is an important property. In fact, there are only actually two operators which, when composed in different ways, can implement the semantics of other kinds of operators: `:<=` and `:>=`. While these operators may not be the most common ones to use, understanding them will help users reason about how all the connection operators work.

I've written a lot of ScalaDoc to describe their semantics. Future work can write more explanative material (mdoc etc.) to flesh out them in more detail.

Update: I've adding the `Defaulting` type, and it the operators will assign default values to unassigned fields, rather than erroring.

### TODO:

- Update docs to clarify strictness guarantees
- Write mdoc
- Brainstorm names for `Defaulting`
- Improve the tests, I'm not happy with their structure, and they aren't testing `:<=` and `:>=`

### Current requested reviewer feedback

- ScalaDoc explanations are clear (and mdoc once written)

### Contributor Checklist

#### Type of Improvement

 - new feature/API 

#### API Impact

No impact on current code. Adds new operators and documentation.

#### Desired Merge Strategy

#### Release Notes

- Added new operators :<>=, :<=, :>=, :#=
- Added new Scaladoc to :=, <>

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -115,7 +115,6 @@ sealed abstract class Aggregate extends Data {
   }
 
   // Emits the FIRRTL `this <- that`, or `this is invalid` if that == DontCare
-  // Note that `this <= that` will be emitted if both `this` and `that` are not aggregate types
   private[chisel3] def firrtlPartialConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
     // If the source is a DontCare, generate a DefInvalid for the sink,
     //  otherwise, issue a Partial Connect.
@@ -280,6 +279,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * @note the length of this Vec and that Seq must match
     * @param that the Seq to connect from
+    * @group connection
     */
   def <>(that: Seq[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = {
     if (this.length != that.length)
@@ -302,6 +302,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     * @note This is necessary in [[Aggregate]], rather than relying on [[Data.<>]], due to supporting the Seq
     * @note the length of this Vec and that Vec must match
     * @param that the Vec to connect from
+    * @group connection
     */
   def <>(that: Vec[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit =
     this.bulkConnect(that.asInstanceOf[Data])
@@ -315,6 +316,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *  - Equivalent to `this :<>= that`, with the additional restriction that the relative bundle field flips must match
     *
     * @note the length of this Vec must match the length of the input Seq
+    * @group connection
     */
   def :=(that: Seq[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = {
     require(
@@ -335,6 +337,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * @note This is necessary in [[Aggregate]], rather than relying on [[Data.:=]], due to supporting the Seq
     * @note the length of this Vec must match the length of the input Vec
+    * @group connection
     */
   def :=(that: Vec[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = this.connect(that)
 
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -274,7 +274,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * For chisel3._, uses the [[BiConnect]] algorithm; sub-elements of `that` may end up driving sub-elements of `this`
     *  - Complicated semantics, will likely be deprecated in the future
-    * 
+    *
     * For Chisel._, emits the FIRRTL.<- operator
     *  - Equivalent to `this :<>= that` without the restrictions that bundle field names and vector sizes must match
     *
@@ -282,7 +282,10 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     * @param that the Seq to connect from
     */
   def <>(that: Seq[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = {
-    if (this.length != that.length) Builder.error(s"Vec (size ${this.length}) and Seq (size ${that.length}) being bulk connected have different lengths!")
+    if (this.length != that.length)
+      Builder.error(
+        s"Vec (size ${this.length}) and Seq (size ${that.length}) being bulk connected have different lengths!"
+      )
     for ((a, b) <- this.zip(that)) {
       a <> b
     }
@@ -292,7 +295,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * For chisel3._, uses the [[BiConnect]] algorithm; sub-elements of `that` may end up driving sub-elements of `this`
     *  - Complicated semantics, hard to write quickly, will likely be deprecated in the future
-    * 
+    *
     * For Chisel._, emits the FIRRTL.<- operator
     *  - Equivalent to `this :<>= that` without the restrictions that bundle field names and vector sizes must match
     *
@@ -307,7 +310,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * For chisel3._, this operator is mono-directioned; all sub-elements of `this` will be driven by sub-elements of `that`.
     *  - Equivalent to `this :#= that`
-    * 
+    *
     * For Chisel._, this operator connections bi-directionally via emitting the FIRRTL.<=
     *  - Equivalent to `this :<>= that`, with the additional restriction that the relative bundle field flips must match
     *
@@ -326,7 +329,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     *
     * For chisel3._, this operator is mono-directioned; all sub-elements of `this` will be driven by sub-elements of `that`.
     *  - Equivalent to `this :#= that`
-    * 
+    *
     * For Chisel._, this operator connections bi-directionally via emitting the FIRRTL.<=
     *  - Equivalent to `this :<>= that`, with the additional restriction that the relative bundle field flips must match
     *
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -104,13 +104,25 @@ sealed abstract class Aggregate extends Data {
 
   private[chisel3] def width: Width = getElements.map(_.width).foldLeft(0.W)(_ + _)
 
-  private[chisel3] def legacyConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
-    // If the source is a DontCare, generate a DefInvalid for the sink,
-    //  otherwise, issue a Connect.
+  // Emits the FIRRTL `this <= that`, or `this is invalid` if that == DontCare
+  private[chisel3] def firrtlConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
+    // If the source is a DontCare, generate a DefInvalid for the sink, otherwise, issue a Connect.
     if (that == DontCare) {
-      pushCommand(DefInvalid(sourceInfo, Node(this)))
+      pushCommand(DefInvalid(sourceInfo, lref))
     } else {
-      pushCommand(BulkConnect(sourceInfo, Node(this), Node(that)))
+      pushCommand(Connect(sourceInfo, lref, Node(that)))
+    }
+  }
+
+  // Emits the FIRRTL `this <- that`, or `this is invalid` if that == DontCare
+  // Note that `this <= that` will be emitted if both `this` and `that` are not aggregate types
+  private[chisel3] def firrtlPartialConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
+    // If the source is a DontCare, generate a DefInvalid for the sink,
+    //  otherwise, issue a Partial Connect.
+    if (that == DontCare) {
+      pushCommand(DefInvalid(sourceInfo, lref))
+    } else {
+      pushCommand(PartialConnect(sourceInfo, lref, Node(that)))
     }
   }
 
@@ -258,23 +270,46 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
   private[chisel3] final override def allElements: Seq[Element] =
     (sample_element +: self).flatMap(_.allElements)
 
-  /** Strong bulk connect, assigning elements in this Vec from elements in a Seq.
+  /** The "bulk connect operator", assigning elements in this Vec from elements in a Seq.
     *
-    * @note the length of this Vec must match the length of the input Seq
+    * For chisel3._, uses the [[BiConnect]] algorithm; sub-elements of `that` may end up driving sub-elements of `this`
+    *  - Complicated semantics, will likely be deprecated in the future
+    * 
+    * For Chisel._, emits the FIRRTL.<- operator
+    *  - Equivalent to `this :<>= that` without the restrictions that bundle field names and vector sizes must match
+    *
+    * @note the length of this Vec and that Seq must match
+    * @param that the Seq to connect from
     */
   def <>(that: Seq[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = {
-    if (this.length != that.length) {
-      Builder.error("Vec and Seq being bulk connected have different lengths!")
-    }
-    for ((a, b) <- this.zip(that))
+    if (this.length != that.length) Builder.error(s"Vec (size ${this.length}) and Seq (size ${that.length}) being bulk connected have different lengths!")
+    for ((a, b) <- this.zip(that)) {
       a <> b
+    }
   }
 
-  // TODO: eliminate once assign(Seq) isn't ambiguous with assign(Data) since Vec extends Seq and Data
+  /** The "bulk connect operator", assigning elements in this Vec from elements in a Vec.
+    *
+    * For chisel3._, uses the [[BiConnect]] algorithm; sub-elements of `that` may end up driving sub-elements of `this`
+    *  - Complicated semantics, hard to write quickly, will likely be deprecated in the future
+    * 
+    * For Chisel._, emits the FIRRTL.<- operator
+    *  - Equivalent to `this :<>= that` without the restrictions that bundle field names and vector sizes must match
+    *
+    * @note This is necessary in [[Aggregate]], rather than relying on [[Data.<>]], due to supporting the Seq
+    * @note the length of this Vec and that Vec must match
+    * @param that the Vec to connect from
+    */
   def <>(that: Vec[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit =
     this.bulkConnect(that.asInstanceOf[Data])
 
-  /** Strong bulk connect, assigning elements in this Vec from elements in a Seq.
+  /** "The strong connect operator", assigning elements in this Vec from elements in a Seq.
+    *
+    * For chisel3._, this operator is mono-directioned; all sub-elements of `this` will be driven by sub-elements of `that`.
+    *  - Equivalent to `this :#= that`
+    * 
+    * For Chisel._, this operator connections bi-directionally via emitting the FIRRTL.<=
+    *  - Equivalent to `this :<>= that`, with the additional restriction that the relative bundle field flips must match
     *
     * @note the length of this Vec must match the length of the input Seq
     */
@@ -287,7 +322,17 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
       a := b
   }
 
-  // TODO: eliminate once assign(Seq) isn't ambiguous with assign(Data) since Vec extends Seq and Data
+  /** "The strong connect operator", assigning elements in this Vec from elements in a Vec.
+    *
+    * For chisel3._, this operator is mono-directioned; all sub-elements of `this` will be driven by sub-elements of `that`.
+    *  - Equivalent to `this :#= that`
+    * 
+    * For Chisel._, this operator connections bi-directionally via emitting the FIRRTL.<=
+    *  - Equivalent to `this :<>= that`, with the additional restriction that the relative bundle field flips must match
+    *
+    * @note This is necessary in [[Aggregate]], rather than relying on [[Data.:=]], due to supporting the Seq
+    * @note the length of this Vec must match the length of the input Vec
+    */
   def :=(that: Vec[T])(implicit sourceInfo: SourceInfo, moduleCompileOptions: CompileOptions): Unit = this.connect(that)
 
   /** Creates a dynamically indexed read or write accessor into the array.
--- a/core/src/main/scala/chisel3/Connectable.scala
+++ b/core/src/main/scala/chisel3/Connectable.scala
@@ -337,16 +337,18 @@ private[chisel3] object DirectionalConnectionFunctions {
           if(op.assignToConsumer) leafConnect(c, p, co, op)
           if(op.assignToProducer) leafConnect(c, p, po, op)
         }
-      case (path, Some(LeafConnection(Some((c, FlippedWithRoot)), None))) => if(op.noDangles) errors += (s"dangling consumer field $c")
-      case (path, Some(LeafConnection(None, Some((p, AlignedWithRoot))))) => if(op.noDangles) errors += (s"dangling producer field $p")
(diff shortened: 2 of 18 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -104,13 +104,25 @@ sealed abstract class Aggregate extends Data {
...
+    // If the source is a DontCare, generate a DefInvalid for the sink, otherwise, issue a Connect.
+    if (that == DontCare) {
+      pushCommand(DefInvalid(sourceInfo, lref))
+    } else {
+      pushCommand(Connect(sourceInfo, lref, Node(that)))
+    }
+  }
+
+  // Emits the FIRRTL `this <- that`, or `this is invalid` if that == DontCare
+  // Note that `this <= that` will be emitted if both `this` and `that` are not aggregate types
```
seldridge, line 118: Is there a good reason to special case this situation as opposed to letting a FIRRTL compiler do whatever it wants with it? I think that connect and partial connect both have the same behavior here?

azidar: Hmm. I can't remember, does partial connect error if lhs is a source flow, or silently do nothing?

azidar: But the original reason why I special cased it is to match what Chisel was doing before this PR.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -258,23 +270,49 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
...
+      Builder.error(
+        s"Vec (size ${this.length}) and Seq (size ${that.length}) being bulk connected have different lengths!"
+      )
+    for ((a, b) <- this.zip(that)) {
       a <> b
+    }
   }
 
-  // TODO: eliminate once assign(Seq) isn't ambiguous with assign(Data) since Vec extends Seq and Data
+  /** The "bulk connect operator", assigning elements in this Vec from elements in a Vec.
```
mwachs5, line 294: we should put all these in the `@group connect`

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -180,15 +180,15 @@ private[chisel3] object BiConnect {
         ) {
           pushCommand(Connect(sourceInfo, leftReified.get.lref, rightReified.get.lref))
         } else if (!emitStrictConnects) {
```
mwachs5, line 182: should we change `emitStrictConnects` to `emitFirrtlConnects`?

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -251,20 +251,22 @@ private[chisel3] object BiConnect {
     }
   }
 
-  /** Check whether two aggregates can be bulk connected (<=) in FIRRTL. From the
-    * FIRRTL specification, the following must hold for bulk connection:
+  /** Check whether two Data can be bulk connected (<=) in FIRRTL. From the
+    * FIRRTL specification, the following must hold for FIRRTL connection:
     *
-    *   1. The types of the left-hand and right-hand side expressions must be
+    *   1. The Chisel types of the left-hand and right-hand side expressions must be
```
mwachs5, line 257: why would the firrtl specification say anything about the Chisel types?

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -251,20 +251,22 @@ private[chisel3] object BiConnect {
...
     *       equivalent.
     *   2. The bit widths of the two expressions must allow for data to always
     *        flow from a smaller bit width to an equal size or larger bit width.
     *   3. The flow of the left-hand side expression must be sink or duplex
     *   4. Either the flow of the right-hand side expression is source or duplex,
     *      or the right-hand side expression has a passive type.
+    *
```
mwachs5, line 264: 5. Analogs aren't allowed here (something about analogs)

core/src/main/scala/chisel3/internal/BiConnect.scala
azidar: Isn't that the name of the compile option you made?

core/src/main/scala/chisel3/internal/BiConnect.scala
azidar: It doesn't, but in order for us to know if a FIRRTL connection can be emitted, we need to know whether Chisel type is a valid FIRRTL type in this regard

core/src/main/scala/chisel3/internal/BiConnect.scala
azidar: Uh.. do we check for analog??

mwachs5: sorry what do you mean?

core/src/main/scala/chisel3/internal/BiConnect.scala
mwachs5: yeah I am asking if we should change it

core/src/main/scala/chisel3/internal/BiConnect.scala
azidar: In the function below, I don't think we check for Analog..

mwachs5: im pretty sure it will return false if there is an Analog in the bundle. I think this fails:

```
    // check that the aggregates have the same types
    def typeCheck = CheckTypes.validConnect(
      Converter.extractType(sink, sourceInfo),
      Converter.extractType(source, sourceInfo)
    )
```

core/src/main/scala/chisel3/internal/BiConnect.scala
mwachs5: we can do in  a follow-up

core/src/main/scala/chisel3/internal/BiConnect.scala
mwachs5: it eventually calls this:

```
object WrappedType {
  def apply(t: Type) = new WrappedType(t)
  def wt(t:    Type) = apply(t)
  // Check if it is legal for the source type to drive the sink type
  // Which is which matters because ResetType can be driven by itself, Bool, or AsyncResetType, but
  //   it cannot drive Bool nor AsyncResetType
  private def compare(sink: Type, source: Type): Boolean =
    (sink, source) match {
      case (_: UIntType, _: UIntType) => true
      case (_: SIntType, _: SIntType) => true
      case (ClockType, ClockType)           => true
      case (AsyncResetType, AsyncResetType) => true
      case (ResetType, tpe)                 => legalResetType(tpe)
      case (tpe, ResetType)                 => legalResetType(tpe)
      case (_: FixedType, _: FixedType) => true
      case (_: IntervalType, _: IntervalType) => true
      // Analog totally skips out of the Firrtl type system.
      // The only way Analog can play with another Analog component is through Attach.
      // Otherwise, we'd need to special case it during ExpandWhens, Lowering,
      // ExpandConnects, etc.
      case (_: AnalogType, _: AnalogType) => false
...

File: core/src/main/scala/chisel3/Aggregate.scala (hunk 2 above)
Comment (azidar) [reply core/src/main/scala/chisel3/Aggregate.scala:294]: Does that actually work?

File: core/src/main/scala/chisel3/internal/BiConnect.scala (hunk 5 above)
Comment (azidar) [reply core/src/main/scala/chisel3/internal/BiConnect.scala:265]: Thanks!

File: core/src/main/scala/chisel3/internal/BiConnect.scala (hunk 4 above)
Comment (azidar) [reply core/src/main/scala/chisel3/internal/BiConnect.scala:257]: I clarified, thanks!

File: core/src/main/scala/chisel3/Aggregate.scala (hunk 1 above)
Comment (azidar) [reply core/src/main/scala/chisel3/Aggregate.scala:118]: I think we aren't special casing anymore, so clarified the comment.

File: core/src/main/scala/chisel3/Aggregate.scala [hunk 6]
```diff
@@ -104,13 +104,25 @@ sealed abstract class Aggregate extends Data {
...
+  private[chisel3] def firrtlPartialConnect(that: Data)(implicit sourceInfo: SourceInfo): Unit = {
     // If the source is a DontCare, generate a DefInvalid for the sink,
-    //  otherwise, issue a Connect.
+    //  otherwise, issue a Partial Connect.
     if (that == DontCare) {
-      pushCommand(DefInvalid(sourceInfo, Node(this)))
+      pushCommand(DefInvalid(sourceInfo, lref))
     } else {
-      pushCommand(BulkConnect(sourceInfo, Node(this), Node(that)))
+      pushCommand(PartialConnect(sourceInfo, lref, Node(that)))
```
mwachs5, line 125: @azidar what do you feel about my aggressive rename here... do we want to roll back this change?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -258,23 +270,49 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
...
     *
-    * @note the length of this Vec must match the length of the input Seq
+    * For chisel3._, uses the [[BiConnect]] algorithm; sub-elements of `that` may end up driving sub-elements of `this`
+    *  - Complicated semantics, will likely be deprecated in the future
+    *
+    * For Chisel._, emits the FIRRTL.<- operator
+    *  - Equivalent to `this :<>= that` without the restrictions that bundle field names and vector sizes must match
+    *
+    * @note the length of this Vec and that Seq must match
+    * @param that the Seq to connect from
```
mwachs5, line 282:
```suggestion
    * @param that the Seq to connect from
    * @group connect
```

core/src/main/scala/chisel3/Aggregate.scala
mwachs5: it did when I was testing it? why do you think it would not work?

azidar: Done. Thanks!

core/src/main/scala/chisel3/Aggregate.scala
azidar: Done, thanks!

mwachs5:
implements part of https://github.com/chipsalliance/chisel3/issues/2643

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -84,6 +87,7 @@ object Connectable {
       * :<= and :>= do, and then you'll be able to reason your way to understanding what's happening :)
       * @note If the types of consumer and producer also have identical relative flips, then we can emit FIRRTL.<= as it is a stricter version of chisel3.:<>=
       * @note If the widths differ between consumer/producer, the assignment will still occur and truncation, if necessary, is implicit
+      * @note "turk-duck-en" is a meme where a turkey is stuffed with a duck, which is stuffed with a chicken; `:<>=` is a `:=` stuffed with a `<>`
```
mwachs5, line 90:
```suggestion
      * @note "turk-duck-en" is a dish where a turkey is stuffed with a duck, which is stuffed with a chicken; `:<>=` is a `:=` stuffed with a `<>`
```

mwachs5: ultra nit :)

core/src/main/scala/chisel3/Trie.scala
```diff
@@ -0,0 +1,135 @@
+package chisel3
+
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+/** A prefix tree. Prefixes represented by a list of `Token` values. For each
+  * valid path the tree will contain a value of type `Value`.
+  *
+  * To use this, do not extend this trait (it is sealed). Instead, create an empty trie with
```
mwachs5, line 10: can you provide some wikipedia reference or something to what a Trie is?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -62,4 +62,54 @@ object Defaulting {
     */
   def apply[T <: Data](default: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Defaulting[T] =
     new Defaulting(chiselTypeOf(default), default)
+
+  implicit class DefaultingTypeclass[T <: Data](h: T) {
+    def withDefault(f: T => Seq[(Data, Data)])(implicit sourceInfo: SourceInfo): T = {
```
mwachs5, line 67: can you put comments on these. what is this h, l?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -62,4 +62,54 @@ object Defaulting {
...
+        case a: Record =>
+          a.elements.foreach { case (field, e) => recBuildTrie(path :+ field, e) }
+        case x =>
+      }
+    }
+    recBuildTrie(Vector.empty[String], data)
+    trie
+  }
+
+  private[chisel3] def setDefault(h: Data, lit: Data): Unit = {
```
mwachs5, line 101: even though this is private can you scaladoc this. what is h, lit?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -62,4 +62,54 @@ object Defaulting {
...
+    val hwTrie = buildTrie(h)
+    val litTrie = buildTrie(lit)
+    hwTrie.collectDeep {
+      case (path, Some(x)) =>
+        val inLit = litTrie.get(path)
+        inLit.map { y => 
+          require(x.defaultOrNull == null)
```
mwachs5, line 108: can we put messages on these, something like "Can only set the default once but it's was set to bla before and now we're trying to set it to baz" (I assum ethat is what this is requiring? Can we add another test where someone does Bool().withDefault.withDefault to show hitting this error?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -62,4 +62,54 @@ object Defaulting {
...
+      f(h).map { case (h, l) => setDefault(h, l) }
+      h
+    }
+    def withDefault(lit: T)(implicit sourceInfo: SourceInfo): T = {
+      if(!lit.isLit) internal.Builder.error("Must provide a literal")
+      if(h.defaultOrNull != null) internal.Builder.error(s"Cannot set a type's default value twice; setting as $lit, but already set as ${lit.defaultOrNull}")
+      setDefault(h, lit)
+      h
+    }
+    def hasDefault: Boolean = h.defaultOrNull != null
```
mwachs5, line 77: is this a public API? Still think this should be more specific like "hasDefaultInTurduckenConnectOperator" or people will start abusing it in their IO / Wire initilizers, unless that is what they are supposed to do with it?

core/src/main/scala/chisel3/Trie.scala
azidar: Yup!

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      * All producer's leaf/branch alignments (with respect to producer) do not influence the connection.
+      *
+      * The following restrictions apply:
+      *  - The Chisel type of consumer and producer must be the "same shape" recursively:
+      *    - All ground types are the same (UInt and UInt are same, SInt and UInt are not), but widths can be different
+      *    - All vector types are the same length
+      *    - All bundle types have the same field names, but the flips of fields can be different between producer and consumer
+      *  - The leaf fields that are ultimately assigned to, must be assignable. This means they cannot be module inputs or instance outputs.
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, but leave `ready` unconnected
```
mwachs5, line 36:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, but leave `ready` unconnected
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      * @note If the widths differ between consumer/producer, the assignment will still occur and truncation, if necessary, is implicit
+      *
```
mwachs5, line 38: by the way, at least in my `sbt doc` converter to HTML, multiple @notes just get rendered as a big block of text with commas between them 🤷

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      * All consumer's leaf/branch alignments (with respect to consumer) do not influence the connection.
+      *
+      * The following restrictions apply:
+      *  - The Chisel type of consumer and producer must be the "same shape":
+      *    - All ground types are the same (UInt and UInt are same, SInt and UInt are not), but widths can be different
+      *    - All vector types are the same length
+      *    - All bundle types have the same field names, but the flips of fields can be different
+      *  - The leaf fields that are ultimately assigned to, must be assignable. This means they cannot be module inputs or instance outputs.
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `ready` from consumer to producer, but leave `bits` and `valid` unconnected
```
mwachs5, line 62:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `ready` from consumer to producer, but leave `bits` and `valid` unconnected
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      *
+      * For `consumer :<>= producer`, both producer and consumer leafs could be driving or be driven-to:
+      *   - consumer's fields aligned w.r.t. consumer will be driven by corresponding fields of producer
+      *   - producer's fields flipped w.r.t. producer will be driven by corresponding fields of consumer
+      *
+      * Identical to calling both :<= and :>= in sequence (order is irrelevant), e.g.:
+      *   consumer :<= producer
+      *   consumer :>= producer
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, and `ready` from consumer to producer.
```
mwachs5, line 86:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, and `ready` from consumer to producer.
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+
+    /** The "mono-direction connection operator", aka the "coercion operator"
+      *
+      * For `consumer :#= producer`, all leaf fields of consumer (regardless of relative flip) are driven by the corresponding leaf fields of producer (regardless of relative flip)
+      *
+      * Identical to calling :<= and :>=, but swapping consumer/producer for :>=: (order is irrelevant), e.g.:
+      *   consumer :<= producer
+      *   producer :>= consumer
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits`, `valid`, AND `ready` from producer to consumer (despite `ready` being flipped)
```
mwachs5, line 131:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits`, `valid`, AND `ready` from producer to consumer (despite `ready` being flipped)
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      * All producer's leaf/branch alignments (with respect to producer) do not influence the connection.
+      *
+      * The following restrictions apply:
+      *  - The Chisel type of consumer and producer must be the "same shape" recursively:
+      *    - All ground types are the same (UInt and UInt are same, SInt and UInt are not), but widths can be different
+      *    - All vector types are the same length
+      *    - All bundle types have the same field names, but the flips of fields can be different between producer and consumer
+      *  - The leaf fields that are ultimately assigned to, must be assignable. This means they cannot be module inputs or instance outputs.
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, but leave `ready` unconnected
```
mwachs5, line 165:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, but leave `ready` unconnected
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      * All consumer's leaf/branch alignments (with respect to consumer) do not influence the connection.
+      *
+      * The following restrictions apply:
+      *  - The Chisel type of consumer and producer must be the "same shape":
+      *    - All ground types are the same (UInt and UInt are same, SInt and UInt are not), but widths can be different
+      *    - All vector types are the same length
+      *    - All bundle types have the same field names, but the flips of fields can be different
+      *  - The leaf fields that are ultimately assigned to, must be assignable. This means they cannot be module inputs or instance outputs.
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `ready` from consumer to producer, but leave `bits` and `valid` unconnected
```
mwachs5, line 193:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `ready` from consumer to producer, but leave `bits` and `valid` unconnected
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+      *
+      * For `consumer :<>= producer`, both producer and consumer leafs could be driving or be driven-to:
+      *   - consumer's fields aligned w.r.t. consumer will be driven by corresponding fields of producer
+      *   - producer's fields flipped w.r.t. producer will be driven by corresponding fields of consumer
+      *
+      * Identical to calling both :<= and :>= in sequence (order is irrelevant), e.g.:
+      *   consumer :<= producer
+      *   consumer :>= producer
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, and `ready` from consumer to producer.
```
mwachs5, line 219:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits` and `valid` from producer to consumer, and `ready` from consumer to producer.
```

core/src/main/scala/chisel3/Connectable.scala
```diff
@@ -0,0 +1,486 @@
...
+
+    /** The "mono-direction connection operator", aka the "coercion operator"
+      *
+      * For `consumer :#= producer`, all leaf fields of consumer (regardless of relative flip) are driven by the corresponding leaf fields of producer (regardless of relative flip)
+      *
+      * Identical to calling :<= and :>=, but swapping consumer/producer for :>=: (order is irrelevant), e.g.:
+      *   consumer :<= producer
+      *   producer :>= consumer
+      *
+      * @note Connecting two [[Decoupled]]'s would connect `bits`, `valid`, AND `ready` from producer to consumer (despite `ready` being flipped)
```
mwachs5, line 246:
```suggestion
      * @example Connecting two [[Decoupled]]'s would connect `bits`, `valid`, AND `ready` from producer to consumer (despite `ready` being flipped)
```

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,115 @@
...
+
+import scala.collection.immutable.SeqMap
+
+/** Data type for representing any type that has a defined default value
+  *
+  * Uses the underlying opaque type support.
+  *
+  * @note This API is experimental and subject to change
+  */
+final class Defaulting[T <: Data] private (
```
mwachs5, line 16: should we delete this from this PR?

azidar: Probably, but I want to try out this PR first to see how it works out.

src/test/scala/chiselTests/DirectionalBulkConnectSpec.scala
```diff
@@ -731,44 +731,201 @@ class DirectionalBulkConnectSpec extends ChiselFunSpec with Utils {
...
+    it("(?.d) A structurally identical but fully aligned monitor version of a bundle can easily be connected to") {
+      class Decoupled extends Bundle {
+        val valid = Bool()
+        val ready = Flipped(Bool())
+        val data = UInt(32.W)
+      }
+      class MyModule extends Module {
+        val in  = IO(Flipped(new Decoupled()))
+        val out = IO(new Decoupled())
+        val monitor = IO(Output(new Decoupled()))
```
mwachs5, line 856: can you please show a version of this where the underlying type is not the same thing (`Decoupled` vs a `FlatDecoupled` where you don't use `Output` to strip the type, it just has the type manually. I would want to see that it is the same behavior

src/test/scala/chiselTests/DirectionalBulkConnectSpec.scala
```diff
@@ -731,44 +731,201 @@ class DirectionalBulkConnectSpec extends ChiselFunSpec with Utils {
...
+        out :<>= in
+        monitor :#= in
```
mwachs5, line 858: can you also show (including the one I had above) the difference (if any) between `:<=` in the two cases. 

Mostly I want to understand if we are talking about going down the `monitor`'s aligned/not aligned fields, does it matter whether they are "natively" aligned in the Bundle subclass or have been forced to be aligned with the `Output`.

azidar:
Closed in favor of #2806

PR #2715, opened 2022-08-30
Add ChiselSubtypeOf for structural subtyping

zyedidia:
Currently DataView works at the Scala type level, where subtypes for the DataView must be subtypes in the Scala type system. This PR expands the notion of a subtype to include Chisel types that structurally have a parent/child relationship. For example:

```scala
class A extends Bundle {
  val x = UInt(3.W)
  val y = UInt(3.W)
  val z = UInt(3.W)
}
class B extends Bundle {
  val x = UInt(3.W)
  val y = UInt(3.W)
}
```

A and B have no subtyping relationship in the Scala type system, but structurally A is a subtype of B because it contains all of B's public fields (same names and same types). After this PR, it would be possible to connect something of type A to something of type B using DataView:

```scala
b <> a.viewAsSupertype(new B)
```

This would be the same as

```scala
b.x := a.x
b.y := a.y
```

If B is not a supertype of A a Scala compilation error will be thrown (at compile-time not during elaboration), saying `A is not a Chisel subtype of B.` Sometimes knowing whether Chisel types are equivalent is not possible until elaboration time, in which case an error might be thrown then instead of during compilation.

This works by using a macro to reflectively inspect the two types at compile-time to determine if they are structurally subtypes. If not, the macro throws a scala compile error. This replaces the previously used `A <:< B` operator with `ChiselSubTypeOf[A, B]`  in `viewAsSupertype`.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This is binary incompatible because it changes the signature of `viewAsSupertype`. The `viewAsSupertype` function will now also accept more types (allowing structural supertypes instead of just Scala supertypes).

None.

#### Desired Merge Strategy
Squash.

#### Release Notes

The DataView `viewAsSupertype` method now accepts supertypes that are structurally super (A is a structural supertype of B if all of A's public fields are contained within B).

- [ ] Did you review?


macros/src/main/scala/chisel3/internal/SubTypeOf.scala
```diff
@@ -0,0 +1,49 @@
...
+    def subtypeOf(a: Type, b: Type): Boolean = {
+      val mb = b.members.filter(_.isPublic)
+      // Go through every public member of b and make sure a member with the
+      // same name exists in a and it has the same structural type.
+      mb.filter(vb => vb.isTerm && vb.asTerm.isGetter || vb.isType)
+        .forall(vb => {
+          val name = if (vb.isTerm) TermName(vb.name.toString) else TypeName(vb.name.toString)
+          typeEquals(a.member(name).info, vb.info)
+        })
+    }
```
jackkoenig, line 36: This doesn't actually check that the members are of type `Data`, so I can make things not work by adding fields of other Scala types that should be irrelevant, eg. `val d = 3`. See [the plugin](https://github.com/chipsalliance/chisel3/blob/5fdf74f95e64cb69d6097547f20d789a83dbd735/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala#L32) for a complex example of checking the types of things, we could probably do something similar (but cleaner) here.

macros/src/main/scala/chisel3/internal/SubTypeOf.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    if (!subtypeOf(ta.tpe, tb.tpe)) {
+      c.error(
+        empty.pos,
+        s"${ta.tpe} is not a Chisel subtype of ${tb.tpe}. Did you mean .viewAs[${tb.tpe}]? " +
+          "Please see https://www.chisel-lang.org/chisel3/docs/cookbooks/dataview"
```
jackkoenig, line 42: It would be really helpful if we could list the files that don't match

mwachs5 (commented):
is there a reason to change the `viewAsSupertype` function vs adding a new `viewAsStructuralSupertype` or `viewAsChiselSupertype` or something? I think it's confusing to mix the Scala and chisel semantics in this way, and sticking the word `chisel` in there is consistent with `chiselTypeOf`

jackkoenig:
> is there a reason to change the viewAsSupertype function vs adding a new viewAsStructuralSupertype or viewAsChiselSupertype or something? 

This functionality completely subsumes `.viewAsSupertype`. We could of course have both functions for clarity though even if they use the same underlying mechanism.

> I think it's confusing to mix the Scala and chisel semantics in this way, and sticking the word chisel in there is consistent with chiselTypeOf

Incidentally, I'd like to rename `chiselTypeOf` to just `typeOf` (`chisel3.typeOf` after all) to make it less verbose. I feel similar about `.viewAsSupertype` which is pretty annoying to type and `viewAsStructuralSupertype` or `viewAsChiselSupertype` would just be worse. I dunno though, clarity is more important than brevity, but we do want to keep the up the signal-to-noise ratio of code using these APIs...

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,55 @@
...
+    import c.universe._
+
+    val ta = implicitly[c.WeakTypeTag[A]]
+    val tb = implicitly[c.WeakTypeTag[B]]
+    val tdata = implicitly[c.WeakTypeTag[Data]]
+    val empty = q""
+
+    def baseType(t: Type): Type = t.baseType(t.baseClasses(0))
+
+    def couldBeEqual(a: Type, b: Type) =
```
mwachs5, line 21: can we put the return types on these `def`s

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,55 @@
...
+      // If one baseType is a subtype the other baseType, then these two
+      // types could be equal, so we allow it and leave it to elaboration to
+      // figure out. Otherwise, the types must be equal or we throw an error.
+      (baseType(b) <:< baseType(a) || baseType(a) <:< baseType(b)) || a =:= b
+
+    // Returns true if 'a' is a structural subtype of 'b' (i.e. all fields of
+    // 'b' exist within 'a' with the same names and the same types).
+    def subtypeOf(a: Type, b: Type) = {
```
mwachs5, line 29: return type plz

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,55 @@
+package chisel3.experimental
+
+import chisel3._
+
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+
+trait ChiselSubTypeOf[A, B]
```
jackkoenig, line 8: A ScalaDoc describing what the semantics of this are would be great.

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,54 @@
...
+    // Go through every public member of b and make sure a member with the
+    // same name exists in a and it has the same structural type.
+    for (vb <- mb) {
+      val name = TermName(vb.name.toString)
+      val vaTyp = a.member(name).info.resultType
+      val vbTyp = vb.info.resultType
+      if (vaTyp == NoType || vbTyp == NoType || !couldBeEqual(vaTyp, vbTyp)) {
+        val err = if (vaTyp == NoType) s"${a}.${name} does not exist" else s"${vaTyp} != ${vbTyp}"
+        c.error(
+          empty.pos,
```
jackkoenig, line 44:
```suggestion
          c.enclosingPosition,
```
This will give a better source locator pointing to the opening parentheses of `.viewAsSupertype`.

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,54 @@
...
+          s"${a} is not a Chisel subtype of ${b}: mismatch at ${b}.${name}: $err. Did you mean .viewAs[${b}]? " +
+            "Please see https://www.chisel-lang.org/chisel3/docs/cookbooks/dataview"
+        )
+      }
+    }
+
+    return empty
```
jackkoenig, line 51:
```suggestion
   empty
```
No need for `return` keyword.

core/src/main/scala/chisel3/experimental/dataview/SubTypeOf.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import chisel3._
+
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+
+trait ChiselSubTypeOf[A, B]
+
+object ChiselSubTypeOf {
+  def genChiselSubTypeOf[A: c.WeakTypeTag, B: c.WeakTypeTag](c: Context): c.Tree = {
```
jackkoenig, line 11:
```suggestion
object ChiselSubtypeOf {
  def genChiselSubtypeOf[A: c.WeakTypeTag, B: c.WeakTypeTag](c: Context): c.Tree = {
```
Subtype is one word so we shouldn't capitalize `Type`, this also makes it consistent with `.viewAsSupertype`. Can you also name the file the same as the trait? ChiselSubtypeOf`?

jackkoenig (commented):
This looks great, a few minor requests

jackkoenig (approved):
Great work!

core/src/main/scala/chisel3/experimental/dataview/SubtypeOf.scala
```diff
@@ -0,0 +1,73 @@
...
+  *      val y = UInt(3.W)
+  *      val z = UInt(3.W)
+  *    }
+  *    class B extends Bundle {
+  *      val x = UInt(3.W)
+  *      val y = UInt(3.W)
+  *    }
+  *  }}}
+  */
+trait ChiselSubtypeOf[A, B]
```
jackkoenig, line 28:
```suggestion
sealed trait ChiselSubtypeOf[A, B]
```
Forgot that we need to make sure users don't extend this

PR #2717, opened 2022-08-31
Add Instantiate

jackkoenig:
Instantiate is an experimental new API that provides a similar style of
Module instantiation to Module(new MyModule), but using Definition and
Instance under-the-hood to only elaborate Modules with the same
parameters a single time.

One big question is if this should be part of 3.5.x or 3.6.0. I'm fine with 3.5.x but I think we should limit any usage until 3.6.0. Also, depending on how we decide to solve "The DataBox Problem" (see comments in code), if we were to change the implementation of `==` and `hashCode` of `Data` within Chisel, that **definitely** is a major version change. 

@sequencer as promised! Sorry for being so slow here.

#### This is currently a draft because I think it needs a few things before merging:
- [x] Move the cache into the DynamicContext and clear it between elaborations
- [x] Add `SourceInfo` so that `Instances` returned by this API have the right source locator
- [x] Add an example to `docs`
- [x] Test differing `CompileOptions`

#### Other improvements that I think should be made but need not be part of the initial PR:
- [ ] Solve "The DataBox Problem" -- Implement a better mechanism for structural `==` and `hashCode` behavior for `Data`

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

New API

This makes it very convenient for users to emit multiply instantiated modules (although they already can with `Definition` and `Instance`) which is likely to expose bugs in FIRRTL custom transforms. This also has the potential to change disambiguation suffixes in module names since modules do not need to be elaborated and deduped when using this feature.  

#### Desired Merge Strategy

- Squash

#### Release Notes

Add experimental `Instantiate`, a new API for multiply instantiating modules of the same parameters.

- [ ] Did you review?


core/src/main/scala/chisel3/experimental/hierarchy/Instantiate.scala
```diff
@@ -0,0 +1,191 @@
...
+    args: K,
+    f:    K => A
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): Instance[A] = {
+    val tag = implicitly[ru.WeakTypeTag[A]]
+    // Include type of module in key since different modules could have the same arguments
+    // CompileOptions need to be included but are likely a source of confusion for users
+    val key = (boxAllData(args), tag, compileOptions)
```
sequencer, line 120: I wonder if it is possible to let user being able to override the `key` implementation, like they can rewrite the `hashCode` and `equal` API  ?

jackkoenig: We could, but I'm curious what the use case is? We should definitely solve the "The DataBox Problem", but other than making sure that `Data` behaves correctly here, what need is there for changing the key?

jackkoenig:
This is ready for review. I think it's in really good shape other than the "The DataBoxProblem" which I think we should fix separately, and we can of course discuss the name, any ideas better than `Instantiate`?

mwachs5 (approved):

core/src/main/scala/chisel3/experimental/hierarchy/Instantiate.scala
sequencer: I think only users can decide how to cache the Definition. for example:
1. different parameter might have a same RTL but different metadata, or some parameter happen to calculate to a same result, which might lead to different elaboration to a same RTL.
2. On the other hand. Even a same parameter can lead to different RTL when module depends on some global variables, this cannot be avoid by compiler, only user can decide: every time I initiate this module, force it to re-elaborate.

For previous case, user can define the cache key calculation method to explicitly say: when calculation results are same, don't elaborate it. 
For the second case, If user makes some Defination depending on global variables, we have to provide a control to this case: They can configure the key cache with the global variable they need, or just randomize the key to force re-elaborate.

I think this is the reason why I think the cache key should be a user configurable variable.

mwachs5: > I think this is the reason why I think the cache key should be a user configurable variable.

Could the user just not use the Instantiate API in that case? I guess the question is the Instantiate API one way to do this for "well formed" module constructors that don't depend on weird global variables to build themselves, and that do want different things for different passed-in parameters?

sequencer: > Could the user just not use the Instantiate API in that case?

I think it's still reasonable to give user this flexibility, but ask user to give the `cachekey` implementation.  I don't think it's a good solution to provide a "well formed" Chisel module guideline, since there are a lot of code styles, and the born of this API will limit them.
I prefer throwing the "cache key" burden to user, since only the users can know what the cache key is, Jack's implementation provides a neat default implementation, but making some advance users being able to override the `cachekey` can be a good thing to the API usage:
if using this API, users are not allowed to depends on global mutable or obey some code guideline.
vs
if using this API, if users depend on global mutable, or do other thing which is possible to permute the `Module` elaborations, please mix-in a `HasElaborationCache` trait to `Module` to override the default cache key implementation.

jackkoenig:
From discussion in dev meeting on 9/12
* Redefine equality and hashcode on `chisel3.Data` to have structural semantics for unbound types and literals
  * There is some subtly here with how to handle Aggregate literals that have DontCare fields...
* Provide "Key to HashableKey" typeclass to let the user provide a function to change what parts of the arguments to the module will be considered as the key to the Instantiable cache
  * This also needs to take into account the Module type so that this function can differ for different modules (also makes it possible for the user to take into account global variables that are used by a module constructor)

jackkoenig:
I think we should still give the user more power over the hashing function, but that can be done in a backwards compatible way so I'm going to go ahead and merge this.

ekiwi:
> I think we should still give the user more power over the hashing function, but that can be done in a backwards compatible way so I'm going to go ahead and merge this.

I was mostly thinking about creating a mechanism that would allow this feature to be implemented without modifying the `Builder` class. We should chat about this sometime.

sequencer:
I think the #2857 + this can provide a good cache system.

PR #2723, opened 2022-09-02
Add source locators to ports

chick:
### Contributor Checklist

#### Type of Improvement

- backend code generation
- 
#### API Impact

Source locators are now on regularly define ports (not on automatically generated clock and reset).
This should not affect anything downstream

Source locators for ports will be included in generated verilog, but should have change any behavior

#### Desired Merge Strategy

#### Release Notes
With this change firrtl and verilog will contain source locators for ports. This can be very helpful on figuring out where the ports were declared in deeply nested generated `IO`s

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -357,7 +357,7 @@ package experimental {
     private[chisel3] def findPort(name: String): Option[Data] = {
       _ports.find(_._1.seedOpt.contains(name)) match {
         case Some((data, _)) => Some(data)
-        case _ => None
+        case _               => None
       }
     }
 
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -187,7 +187,7 @@ package experimental {
       * requested (so that all calls to ports will return the same information).
       * Internal API.
       */
-    def apply[T <: Data](iodef: => T): T = {
+    def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
       val module = Module.currentModule.get // Impossible to fail
       require(!module.isClosed, "Can't add more ports after module close")
       val prevId = Builder.idGen.value
@@ -337,7 +337,7 @@ package experimental {
       _ids.toSeq
     }
 
-    private val _ports = new ArrayBuffer[Data]()
+    private val _ports = new ArrayBuffer[(Data, SourceInfo)]()
 
     // getPorts unfortunately already used for tester compatibility
     protected[chisel3] def getModulePorts = {
@@ -352,7 +352,9 @@ package experimental {
     // This is dangerous because it can be called before the module is closed and thus there could
     // be more ports and names have not yet been finalized.
     // This should only to be used during the process of closing when it is safe to do so.
-    private[chisel3] def findPort(name: String): Option[Data] = _ports.find(_.seedOpt.contains(name))
+    private[chisel3] def findPort(name: String): Option[Data] = {
+      _ports.map(_._1).find { _.seedOpt.contains(name) }
+    }
 
     protected def portsSize: Int = _ports.size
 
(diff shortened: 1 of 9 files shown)


debs-sifive (approved):
lgtm! (but i do not have write access)

core/src/main/scala/chisel3/Module.scala
```diff
@@ -347,12 +347,16 @@ package experimental {
...
+    protected def portsContains(elem: Data): Boolean = {
+      _ports.exists { port => port._1 == elem }
+    }
 
     // This is dangerous because it can be called before the module is closed and thus there could
     // be more ports and names have not yet been finalized.
     // This should only to be used during the process of closing when it is safe to do so.
-    private[chisel3] def findPort(name: String): Option[Data] = _ports.find(_.seedOpt.contains(name))
+    private[chisel3] def findPort(name: String): Option[Data] = {
+      _ports.map(_._1).find { _.seedOpt.contains(name) }
```
jackkoenig, line 358:
```suggestion
      _ports.find(_._1.seedOpt.contains(name))
```
Unless the JIT can optimize, that `.map` is going to create a copy of `_ports` every time this function is called which would be pretty expensive.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.5.8
+sbt.version=1.6.2
```
jackkoenig, line 1:
```suggestion
sbt.version=1.5.8
```
Please do not change this as part of this PR

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -835,7 +835,7 @@ case class Stop(id: stop.Stop, sourceInfo: SourceInfo, clock: Arg, ret: Int) ext
   "This API should never have been public, for Module port reflection, use DataMirror.modulePorts",
   "Chisel 3.5"
 )
-case class Port(id: Data, dir: SpecifiedDirection)
+case class Port(id: Data, dir: SpecifiedDirection, sourceInfo: SourceInfo = UnlocatableSourceInfo)
```
jackkoenig, line 838:
```suggestion
case class Port(id: Data, dir: SpecifiedDirection, sourceInfo: SourceInfo)
```
I would rather not make this have a default because I want us to be cognizant in our internals about whether or not we have a source locator to pass rather than being able to accidentally forget because of the default.

jackkoenig (changes requested):
Overall this looks great but I do have a few requests.

core/src/main/scala/chisel3/Module.scala
chick: Makes sense. will fix on next push

project/build.properties
chick: Accidentally let this slip through, will revert on next push

core/src/main/scala/chisel3/internal/firrtl/IR.scala
chick: will fix on next push

chick:
@jackkoenig I believe I have addressed all your requests.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -347,12 +347,19 @@ package experimental {
...
     // This is dangerous because it can be called before the module is closed and thus there could
     // be more ports and names have not yet been finalized.
     // This should only to be used during the process of closing when it is safe to do so.
-    private[chisel3] def findPort(name: String): Option[Data] = _ports.find(_.seedOpt.contains(name))
+    private[chisel3] def findPort(name: String): Option[Data] = {
+      _ports.find(_._1.seedOpt.contains(name)) match {
+        case Some((data, _)) => Some(data)
+        case _               => None
+      }
+    }
```
jackkoenig, line 362:
```suggestion
    private[chisel3] def findPort(name: String): Option[Data] =
      _ports.collectFirst { case (data, _) => data.seedOpt.contains(name) => data }
```
We can simplify the logic here a little bit

jackkoenig (approved):
Looks great! 1 suggested improvement.

PR #2729, opened 2022-09-14
Change SInt unary negation (expanding width) note

moriglia:
Fix chipsalliance/chisel3#2728

### Contributor Checklist

  No function/method added

#### Type of Improvement

- documentation

#### API Impact

No API impact

No backend code generation impact

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Bits.scala
+++ b/core/src/main/scala/chisel3/Bits.scala
@@ -896,7 +896,7 @@ sealed class SInt private[chisel3] (width: Width) extends Bits(width) with Num[S
   private[chisel3] override def cloneTypeWidth(w: Width): this.type =
     new SInt(w).asInstanceOf[this.type]
 
-  /** Unary negation (expanding width)
+  /** Unary negation (constant width)
     *
     * @return a hardware $coll equal to zero minus this $coll
     * $constantWidth


aswaterman:
Isn't it the other way around: the comment three lines earlier is incorrect, whereas the one you changed was already correct?  I see that `unary_-` is defined to be `0.S - this`, and `-` is not expanding.

jackkoenig:
I think actually both `unary_-` and `unary_-%` are non-expanding, see Scastie: https://scastie.scala-lang.org/NLnK5LqGSeKJOHSdqoY1HQ

Furthermore, `unary_-%` isn't a real unary operator (you can only call it as `mySInt.unary_-%`, see the [Scala language spec](https://scala-lang.org/files/archive/spec/2.12/06-expressions.html#prefix-operations).

moriglia:
> Isn't it the other way around: the comment three lines earlier is incorrect, whereas the one you changed was already correct? I see that `unary_-` is defined to be `0.S - this`, and `-` is not expanding.
I see. However in the [documentation note](https://www.chisel-lang.org/api/latest/chisel3/SInt.html#-(that:T):T): "The width of the returned hardware type is max(width of this, width of that) + 1." Should this be also fixed?

What I found confusing is that for the "normal" subtraction we have
- Subtraction with no "&" or "%" is non-expanding, but its [note](https://www.chisel-lang.org/api/latest/chisel3/SInt.html#-(that:T):T) says that the size increases by 1
- Subtraction with "&" is expanding too ([note](https://www.chisel-lang.org/api/latest/chisel3/SInt.html#-&(that:chisel3.SInt):chisel3.SInt))
- Subtraction with "%" is non-expanding ([note](https://www.chisel-lang.org/api/latest/chisel3/SInt.html#-%(that:chisel3.SInt):chisel3.SInt))

So what I expected was a similar behaviour for the unary negation, i.e.:
- `unary_-`  expanding
- `unary_-%` non expanding

I've just noticed that there's no expanding width unary negation

> I think actually both `unary_-` and `unary_-%` are non-expanding, see Scastie: https://scastie.scala-lang.org/NLnK5LqGSeKJOHSdqoY1HQ

Thank you for the example. Doesn't it just show that I can represent both the expanding and the non-expanding results in +1 bits?
Anyway, I [modified the example](https://scastie.scala-lang.org/0DCnMhVWQlqMiZpmbVEEAw) and convinced myself that both of you are right

> Furthermore, `unary_-%` isn't a real unary operator (you can only call it as `mySInt.unary_-%`, see the [Scala language spec](https://scala-lang.org/files/archive/spec/2.12/06-expressions.html#prefix-operations).
Ok, true, I realized that because I had to use it in one of my modules an I had to use the postfix notation. Since I now realize that the `unary_-` is non-expanding, I'll put it back.

Shall I fix my commit and also change the note in the documentation for the normal subtraction?

Thanks for taking time to check into such a tiny and meaningless commit :)

aswaterman:
Not a meaningless commit; it’s good to get this stuff correct :) we should change any incorrect documentation and keep the behavior as currently implemented to avoid creating any backwards incompatibility.

jackkoenig (approved):
Thanks for fixing the documentation!

moriglia:
> Shall I fix my commit and also change the note in the documentation for the normal subtraction?

In the end I didn't touch the description for the normal subtraction. Is it somehow inherited from the Num trait?

PR #2731, opened 2022-09-15
Always show BlackBox parameters in alphabetical key order

chick:
Always show BlackBox parameters in alphabetical key order
Includes test

### Contributor Checklist

#### Type of Improvement

- backend code generation

#### API Impact

No API change

Verilog is changed
Verilog generation of blackboxes will have the parameter map ordered alphabetically by key

#### Desired Merge Strategy

#### Release Notes

Verilog is changed
Verilog generation of blackboxes will have the parameter map ordered alphabetically by key

- [ ] Did you review?


src/test/scala/chiselTests/BlackBox.scala
```diff
@@ -222,4 +230,48 @@ class BlackBoxSpec extends ChiselFlatSpec {
...
   }
+
+  "BlackBoxes" should "sort the verilog output of their param map by param key" in {
+    val rand = scala.util.Random
+    rand.setSeed(0)
+
+    class Top(m: Map[String, Param]) extends Module {
+      val io = IO(new Bundle {})
+      val ibufds = Module(new ParameterizedBlackBox(m))
+      ibufds.io.I := clock
```
jackkoenig, line 241: I'm a little confused why FIRRTL is letting you only drive one of the inputs but not the other.

src/test/scala/chiselTests/BlackBox.scala
```diff
@@ -222,4 +230,48 @@ class BlackBoxSpec extends ChiselFlatSpec {
...
+      val paramsShowInOrder = verilogString.split("\n").exists { line =>
+        val offset = line.indexOf(".key_0")
+        val result = if (offset >= 0) {
+          paramsSeen = true
+          keysInOrder(line.drop(offset))
+        } else {
+          false
+        }
+        result
+      } || !paramsSeen
```
jackkoenig, line 272: I find this logic pretty hard to follow. This is a test so it doesn't matter _that_ much but I think it would be more clear and still sufficient to just take a static list of like 6 parameters like parameter named `A` has value `0`, `B` has value `1`, etc., randomize there order, and then just check the exactly expected line in the Verilog, something like:

```scala
verilogString should include ("#(.A(0), .B(1), .C(2)...")
```

jackkoenig (commented):
Solution looks great but I find the test a little hard to follow.

src/test/scala/chiselTests/BlackBox.scala
chick: Makes sense, I've simplified the test

mwachs5:
looks good! Should we backport this even though it changes the verilog, as it is basically a bug?

jackkoenig:
> looks good! Should we backport this even though it changes the verilog, as it is basically a bug?

This is a tough one, because even though the previous output is not _ordered_ and the order will scramble if you add another parameter, it was at least stable between runs. I'm inclined to not backport this but am open to being convinced.

PR #2747, opened 2022-09-23
Added Defaulting type to experimental

azidar:
## Background

In order to remove `Chisel._`, we need to somehow tackle the difference in semantics between `chisel3.<>` and `Chisel.<>`. Our approach has been to revisit connections from first principles, and arrived at the conclusion that the `FixChisel3` operators from RocketChip are an excellent starting place, where we can likely describe the behaviors of `chisel3.<>` and `Chisel.<>` as different combinations of those operators with erroring in cases which were buggy or unintuitive behavior.

The PR https://github.com/chipsalliance/chisel3/pull/2713 adds these operators with the following changes:
- Any field being assigned to must have a corresponding field to assign from
- No support for `CustomBulkAssignable`
- The datatype used for recursing (left-of-connect or right-of-connect) is clarified

## The final blocking issue: `CustomBulkAssignable`

We believe we can handle `Chisel.<>` and `chisel3.<>` now, but the `FixChisel3` operators have one additional behavior which we have not figured out how to support in chisel3: `BundleMap`'s `CustomBulkAssignable`.

When at first glance this appears to be a catch-all loophole to support weird connection behavior, it is only ever used for the single use-case of assigning default values to unassigned BundleMapFields (which is always associated with the field definition).

## Proposals

We considered two major proposals:

Proposal 1: Relies on last-connect-semantics to give the default value, removes the strictness of the previously referenced chisel3 PR, and relies on initialization errors to give back the strictness.

Proposal 2: Enable letting Chisel types define default values, and these are used rather than erroring for the bulk operators (similar to BundleMap behavior).

After some playing around, we realized that BundleMaps would often be nested within other type hierarchies, making last-connect-semantics more painful to rely on for the default behavior, as you would need to only assign defaults to the subfields of bundlemap-kind of things. In addition, the new support for opaque types provided an excellent mechanism to create a special defaulting wrapping type. Hence, we selected Proposal 2.

## Specific Reviewer Requests

1. Whether mdoc should be written
2. Whether you have any ideas for the name besides `Defaulting`

### Contributor Checklist

#### Type of Improvement

 - new feature/API 

#### API Impact

Adds a new experimental datatype, `Defaulting` which enables associating a default value with a type. In a future PR, the new connection operators will assign this default value for missing fields of `Defaulting` type, rather than erroring.

#### Release Notes

Adds a new experimental datatype, `Defaulting` which enables associating a default value with a type.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/Defaulting.scala
+++ b/core/src/main/scala/chisel3/experimental/Defaulting.scala
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental
+
+import chisel3.{chiselTypeOf, Record, Data, Element}
+
+import scala.collection.immutable.SeqMap
+
+/** Data type for representing any type who has a default value
+  *
+  * Uses the underlying opaque type support.
+  *
+  * @note This API is experimental and subject to change
+  */
+final class Defaulting[T <: Data] private (private[chisel3] val tpe: T, val defaultValue: T) extends Record {
+  requireIsChiselType(tpe, s"Chisel hardware type $tpe must be a pure type, not bound to hardware.")
+  requireIsHardware(defaultValue, s"Default value $defaultValue must be bound to a hardware component")
+
+  /** The underlying hardware component, is either the Chisel data type or hardware component */
+  val underlying: T = tpe
+
+  /** The default value for this Defaulting */
+  val default: T = defaultValue
+
+  val elements = SeqMap("" -> underlying)
+  override def opaqueType = elements.size == 1
+  override def cloneType: this.type = (new Defaulting[T](chiselTypeOf(tpe), defaultValue)).asInstanceOf[this.type]
+}
+
+
+/** Object that provides factory methods for [[Analog]] objects
+  *
+  * @note This API is experimental and subject to change
+  */
+object Defaulting {
+  /** Build a Defaulting[T <: Data]
+    *
+    * @param tpe the Chisel data type
+    * @param default the Chisel default value, must be bound to a hardware value
+    */
+  def apply[T <: Data](tpe: T, default: T): Defaulting[T] = new Defaulting(tpe, default)
+
+  /** Build a Defaulting[T <: Data]
+    *
+    * @param default the Chisel default value, must be bound to a hardware value. The underlying type is pulled from the default value.
+    */
+  def apply[T <: Data](default: T): Defaulting[T] = new Defaulting(chiselTypeOf(default), default)
+}
--- a/src/test/scala/chiselTests/experimental/DefaultingSpec.scala
+++ b/src/test/scala/chiselTests/experimental/DefaultingSpec.scala
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests
+package experimental
+
+import chisel3._
+import chisel3.stage.ChiselStage
+import chisel3.experimental.Defaulting
+import chisel3.experimental.BundleLiterals._
+import chisel3.experimental.VecLiterals._
+
+
+class DefaultingSpec extends ChiselFunSpec with Utils {
+  describe("(0): Defaulting type creation") {
+    it("(0.a): Creating a default type of UInt, and then accessing the value") {
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(UInt(32.W), 1.U)))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io <= UInt<1>(\"h1\")"))
+    }
+    it("(0.b): Creating a defaulting type of Bundle") {
+      class MyBundle(width: Int) extends Bundle {
+        val a = UInt(width.W)
+        val b = UInt(width.W)
+      }
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(new MyBundle(32).Lit(_.a -> 1.U, _.b -> 2.U))))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io.a <= UInt<1>(\"h1\")"))
+      assert(firrtl.contains("io.b <= UInt<2>(\"h2\")"))
+    }
+    it("(0.b): Creating a defaulting type of Vec") {
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(Vec(2, UInt(32.W)).Lit(0 -> 1.U, 1 -> 2.U))))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io[0] <= UInt<32>(\"h1\")"))
+      assert(firrtl.contains("io[1] <= UInt<32>(\"h2\")"))
+    }
+  }
+}


src/test/scala/chiselTests/experimental/DefaultingSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+import chisel3.stage.ChiselStage
+import chisel3.experimental.Defaulting
+import chisel3.experimental.BundleLiterals._
+import chisel3.experimental.VecLiterals._
+
+
+class DefaultingSpec extends ChiselFunSpec with Utils {
+  describe("(0): Defaulting type creation") {
+    it("(0.a): Creating a default type of UInt, and then accessing the value") {
+      val firrtl = ChiselStage.emitFirrtl { new Module {
```
seldridge, line 16: Nit: 
```suggestion
      val firrtl = ChiselStage.emitChirrtl { new Module {
```

mwachs5:
> Whether mdoc should be written

I think we can add the mdoc when we PR the :<>= with defaulting, as this type seems pointless without that context, yes?

> Whether you have any ideas for the name besides Defaulting

I am wondering whether we want the name to make this clear that this is `DefaultingInAssignment`, as in Wire(new BundleWithADefaulting()) is *not* going to assign the default value, nor is it assigned to its default on IOs.

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental
+
+import chisel3.{chiselTypeOf, Record, Data, Element}
+
+import scala.collection.immutable.SeqMap
+
+/** Data type for representing any type who has a default value
```
mwachs5, line 9:
```suggestion
/** Data type for representing any type who has a defined default value
```

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
...
+  *
+  * Uses the underlying opaque type support.
+  *
+  * @note This API is experimental and subject to change
+  */
+final class Defaulting[T <: Data] private (private[chisel3] val tpe: T, val defaultValue: T) extends Record {
+  requireIsChiselType(tpe, s"Chisel hardware type $tpe must be a pure type, not bound to hardware.")
+  requireIsHardware(defaultValue, s"Default value $defaultValue must be bound to a hardware component")
+
+  /** The underlying hardware component, is either the Chisel data type or hardware component */
```
mwachs5, line 19: isn't this comment wrong? Isn't the default value the underlying hardware component?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
...
+  val underlying: T = tpe
+
+  /** The default value for this Defaulting */
+  val default: T = defaultValue
+
+  val elements = SeqMap("" -> underlying)
+  override def opaqueType = elements.size == 1
```
mwachs5, line 26: why wouldn't this just be "True"

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
...
+  override def cloneType: this.type = (new Defaulting[T](chiselTypeOf(tpe), defaultValue)).asInstanceOf[this.type]
+}
+
+
+/** Object that provides factory methods for [[Analog]] objects
```
mwachs5, line 31:
```suggestion
/** Object that provides factory methods for [[Defaulting]] objects
```

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
...
+  *
+  * @note This API is experimental and subject to change
+  */
+object Defaulting {
```
mwachs5, line 35: can we get any source locator stuff right from the beginning? What do we have to do for Data etc?

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+        val b = UInt(width.W)
+      }
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(new MyBundle(32).Lit(_.a -> 1.U, _.b -> 2.U))))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io.a <= UInt<1>(\"h1\")"))
+      assert(firrtl.contains("io.b <= UInt<2>(\"h2\")"))
+    }
+    it("(0.b): Creating a defaulting type of Vec") {
```
mwachs5, line 34: can we create a defaulting type of Record

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(Vec(2, UInt(32.W)).Lit(0 -> 1.U, 1 -> 2.U))))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io[0] <= UInt<32>(\"h1\")"))
+      assert(firrtl.contains("io[1] <= UInt<32>(\"h2\")"))
+    }
+  }
+}
```
mwachs5, line 43: can we show test of annotating a Defaulting and/or DataView of a Defaulting (unless we think the Opaque types already cover this)?

core/src/main/scala/chisel3/experimental/Defaulting.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import scala.collection.immutable.SeqMap
+
+/** Data type for representing any type who has a default value
+  *
+  * Uses the underlying opaque type support.
+  *
+  * @note This API is experimental and subject to change
+  */
+final class Defaulting[T <: Data] private (private[chisel3] val tpe: T, val defaultValue: T) extends Record {
```
terpstra, line 15: Because these things get defined at the top-level (outside of a Module), you probably want the defaultValue to be a thunk, not a value.

mwachs5:
Does this compose w things that would be opaque for other reasons (should there be a defaulting trait...?)

ekiwi:
This feature might be pretty useful for the AXI Bundle definitions in order to provide good defaults for rarely used fields: https://github.com/chipsalliance/chisel3/pull/2441

(Currently many developers will write a "initialization" function that needs to be called in order to assign default values)

core/src/main/scala/chisel3/experimental/Defaulting.scala
azidar: Thanks!

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
azidar: Thanks!

core/src/main/scala/chisel3/experimental/Defaulting.scala
azidar: I don't think so? The underlying hardware component, (e.g. a field in a bundle) is the `underlying`, and the defined default value is `default`. I'm not sure if I'm misunderstanding your question.. I clarified in the next PR.

core/src/main/scala/chisel3/experimental/Defaulting.scala
azidar: Thanks!

core/src/main/scala/chisel3/experimental/Defaulting.scala
azidar: Yup, added in the next PR.

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
azidar: As an example? Sure!

core/src/main/scala/chisel3/experimental/Defaulting.scala
azidar: Good point. I'll change it, thanks!

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+  describe("(0): Defaulting type creation") {
+    it("(0.a): Creating a default type of UInt, and then accessing the value") {
+      val firrtl = ChiselStage.emitFirrtl { new Module {
+        val io = IO(Output(Defaulting(UInt(32.W), 1.U)))
+        io.underlying := io.default
+      }}
+      assert(firrtl.contains("io <= UInt<1>(\"h1\")"))
+    }
+    it("(0.b): Creating a defaulting type of Bundle") {
+      class MyBundle(width: Int) extends Bundle {
```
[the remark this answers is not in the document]
azidar: I'm not sure I understand. Are you saying to not make `Defaulting` a wrapping type, but rather a property of Bundle/Record?

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
azidar: I think the Opaque types test cover this.

azidar:
> Does this compose w things that would be opaque for other reasons (should there be a defaulting trait...?)

The composition module would be to wrap a Data multiple times, e.g. `Defaulting[OtherOpaqueType[T]]`. Inheritance should be discouraged, because if you have a known default value for an extensible type, that default value will not be valid for subtypes.

azidar:
Closed in favor of adding it directly to #2713

src/test/scala/chiselTests/experimental/DefaultingSpec.scala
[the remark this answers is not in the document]
ekiwi: Yes. Allow users to write something like:
```.scala
val inputA = Input(UInt(8.W)) withDefault 0.U
```

Similar to a case class:
```.scala
case class MyBundle(a: Int = 0)
```

PR #2751, opened 2022-09-27
Error on S-interpolator usage for assert, assume and printf

adkian-sifive:
This is to be backported to 3.5.x as a warning instead of an error

### Contributor Checklist

#### Type of Improvement

 - updated feature/API      

#### API Impact

Current `assert`, `assume` and `printf` APIs allow the Scala s-interpolator for formatting messages. This often leads to user errors where the Scala object is printed instead of the Chisel object, which leads to the messages being generally unhelpful.

After this update, using s-interpolators will throw an error about deprecation.

N/A
#### Desired Merge Strategy

#### Release Notes

Using Scala s-interpolators now throws an error recommending a shift to the cf interpolator.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printf.scala
+++ b/core/src/main/scala/chisel3/Printf.scala
@@ -92,8 +92,8 @@ object printf {
         c.error(
           c.enclosingPosition,
           "The s-interpolator prints the Scala .toString of Data objects rather than the value " +
-          "of the hardware wire during simulation. Use the cf-interpolator instead. If you want " +
-          "an elaboration time print, use println."
+            "of the hardware wire during simulation. Use the cf-interpolator instead. If you want " +
+            "an elaboration time print, use println."
         )
       case _ =>
     }
--- a/core/src/main/scala/chisel3/Printf.scala
+++ b/core/src/main/scala/chisel3/Printf.scala
@@ -2,10 +2,11 @@
 
 package chisel3
 
-import scala.language.experimental.macros
 import chisel3.internal._
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal.sourceinfo.SourceInfo
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox
 
 /** Prints a message in simulation
   *
@@ -75,8 +76,28 @@ object printf {
     * @param fmt printf format string
     * @param data format string varargs containing data to print
     */
+  def _applyMacroWithInterpolatorCheck(
+    c:              blackbox.Context
+  )(fmt:            c.Tree,
+    data:           c.Tree*
+  )(sourceInfo:     c.Tree,
+    compileOptions: c.Tree
+  ): c.Tree = {
+    import c.universe._
+    fmt match {
+      case q"scala.StringContext.apply(..$_).s(..$_)" =>
+        c.warning(
+          c.enclosingPosition,
+          "s-interpolators for Chisel assert, assume and printf statements will be deprecated in 3.5; use p or cf interpolators instead"
+        )
+      case _ =>
+    }
+    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithReset"))
+    q"$apply_impl_do(_root_.chisel3.Printable.pack($fmt, ..$data))($sourceInfo, $compileOptions)"
+  }
+
   def apply(fmt: String, data: Bits*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
-    apply(Printable.pack(fmt, data: _*))
+    macro _applyMacroWithInterpolatorCheck
 
   /** Prints a message in simulation
     *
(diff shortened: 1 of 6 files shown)


src/test/scala/chiselTests/Vec.scala
```diff
@@ -235,7 +235,7 @@ class IterateTester(start: Int, len: Int)(f: UInt => UInt) extends BasicTester {
   val testVec = VecInit.iterate(start.U, len)(f)
   assert(
     controlVec.asUInt() === testVec.asUInt(),
-    s"Expected Vec to be filled like $controlVec, instead creaeted $testVec\n"
+    cf"Expected Vec to be filled like $controlVec, instead creaeted $testVec\n"
```
mwachs5, line 238:
```suggestion
    cf"Expected Vec to be filled like $controlVec, instead created $testVec\n"
```

pull a weed

src/test/scala/chiselTests/VecLiteralSpec.scala
```diff
@@ -205,7 +205,7 @@ class VecLiteralSpec extends ChiselFreeSpec with Utils {
 
     assertTesterPasses {
       new BasicTester {
-        chisel3.assert(outsideVecLit(0) === 0xdd.U, s"v(0)")
+        chisel3.assert(outsideVecLit(0) === 0xdd.U, cf"v(0)")
```
mwachs5, line 208: is this supposed to be ${v(0)}???

src/test/scala/chiselTests/VecLiteralSpec.scala
```diff
@@ -216,7 +216,7 @@ class VecLiteralSpec extends ChiselFreeSpec with Utils {
...
```
mwachs5, line 219: again is this supposed to be ${v(0)}

mwachs5:
I think this requires some fixes to the printable docs? Or can we say something there to show people the warning/error they will get if they try to printf(s"...")?

src/test/scala/chiselTests/VecLiteralSpec.scala
mwachs5: or why do we need any interpolator at all?

core/src/main/scala/chisel3/Printf.scala
```diff
@@ -100,6 +121,14 @@ object printf {
     printfId
   }
 
+  def _applyWithReset(pable: Printable)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf = {
+    var printfId: Printf = null
+    when(!Module.reset.asBool) {
+      printfId = printfWithoutReset(pable)
+    }
+    printfId
```
jackkoenig, line 129: Why is this here? This feels like an unrelated change.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -43,12 +43,12 @@ object assert extends VerifPrintMacrosDoc {
   // Macros currently can't take default arguments, so we need two functions to emulate defaults.
   def apply(
     cond:    Bool,
-    message: String,
+    message: => String,
```
jackkoenig, line 46: In the Pull Request description you said:

> This is to be backported to 3.5.x as an error instead of a warning

I think you got that backwards that it should be backported as a warning, not an error.

We have a problem though, this change to call by-name (which is probably required for the macro to work, but I'm not 100% sure) breaks binary compatibility so cannot be backported.

adkian-sifive: yeah makes sense, i suspected i did it the other way round. fixed

adkian-sifive: i think the call by name change was needed by something that i've since deleted. i have reverted it

jackkoenig: > yeah makes sense, i suspected i did it the other way round. fixed

The first line of the PR message still has the same phrasing.

core/src/main/scala/chisel3/Printf.scala
adkian-sifive: I've moved this around so that this logic is part of the function `printfWIthReset` so that i can call it from the macro

docs/src/explanations/printing.md
```diff
@@ -103,6 +103,8 @@ Message:
 
 Notice the use of `+` between `cf` interpolated "strings". The results of `cf` interpolation can be concatenated by using the `+` operator.
 
+> :warning: The Scala s-interpolator is deprecated since 3.5 and will output the warning: "s-interpolators for Chisel assert, assume and printf statements are deprecated (since 3.5); use p or cf interpolators instead"
```
jackkoenig, line 106: Can you do an actual execution driven example rather than just a text warning?

Since it fails in this PR, you can do the following:
````
```scala mdoc:fail
  val in = IO(Input(UInt(8.W)))
  printf(s"in = $in\n")
}
```
````
This should fail to compile and render with the exact error when you run `sbt docs/mdoc`

**Review by jackkoenig — APPROVED** [review]

**Comment by jackkoenig** [comment]
This is not marked with Milestone `3.5.x` because we do not want it to be backported and automerged, instead after merging we will do a manual backport and change the behavior to be just a warning.

File: docs/src/explanations/printing.md [hunk 7]
```diff
@@ -13,11 +13,23 @@ Chisel provides the `printf` function for debugging purposes. It comes in two fl
...
 ```
+
+```scala mdoc:fail
+class MyModule extends Module {
+  printf(s"in = $in\n")
+}
+```
+
+Instead, use Chisel's `cf` interpolator as in the following examples:
```
mwachs5, line 31: Nice :)

mwachs5 (approved):

PR #2757, opened 2022-10-03
 Warn on S-interpolator usage for assert, assume and printf (backport #2751)

adkian-sifive:
Backport of #2751, change errors to warnings for usages of s-interpolators

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printf.scala
+++ b/core/src/main/scala/chisel3/Printf.scala
@@ -103,10 +103,12 @@ object printf {
 
   // Private internal methods that serve to maintain binary
   // compatibility after interpolator check updates
-  private[chisel3] def apply(fmt: String, sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
+  @deprecated("This Printf.apply method has been deprecated and will be removed in Chisel 3.6")
+  def apply(fmt: String, sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
     apply(fmt, Nil, sourceInfo, compileOptions)
 
-  private[chisel3] def apply(
+  @deprecated("This Printf.apply method has been deprecated and will be removed in Chisel 3.6")
+  def apply(
     fmt:            String,
     data:           Seq[Bits],
     sourceInfo:     SourceInfo,
--- a/core/src/main/scala/chisel3/Printf.scala
+++ b/core/src/main/scala/chisel3/Printf.scala
@@ -2,10 +2,11 @@
 
 package chisel3
 
-import scala.language.experimental.macros
 import chisel3.internal._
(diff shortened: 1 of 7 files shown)


core/src/main/scala/chisel3/Printf.scala
```diff
@@ -76,7 +77,29 @@ object printf {
     * @param data format string varargs containing data to print
     */
   def apply(fmt: String, data: Bits*)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
-    apply(Printable.pack(fmt, data: _*))
+    macro _applyMacroWithInterpolatorCheck
```
jackkoenig, line 80: The binary compatibility check is failing with
```
[error]  * static method apply(java.lang.String,scala.collection.immutable.Seq,chisel3.internal.sourceinfo.SourceInfo,chisel3.CompileOptions)chisel3.printf#Printf in class chisel3.printf does not have a correspondent in current version
[19](https://github.com/chipsalliance/chisel3/actions/runs/3178232805/jobs/5179516226#step:9:20)
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("chisel3.printf.apply")
[20](https://github.com/chipsalliance/chisel3/actions/runs/3178232805/jobs/5179516226#step:9:21)
[error]  * method apply(java.lang.String,scala.collection.immutable.Seq,chisel3.internal.sourceinfo.SourceInfo,chisel3.CompileOptions)chisel3.printf#Printf in object chisel3.printf does not have a correspondent in current version
[21](https://github.com/chipsalliance/chisel3/actions/runs/3178232805/jobs/5179516226#step:9:22)
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("chisel3.printf.apply")
```
This is due to the fact that we're changing this method to a macro invocation, which means the actual method does not exist in the emitted Java bytecode.

We should be able to restore the methods in the bytecode by adding package private methods that match those exact signatures. We make them package private so that users cannot call them from Scala, but because Java doesn't have a notion of package private, the methods are actually public in the bytecode.

Something like:

```scala
private[chisel3] def apply(fmt: String, sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf = apply(fmt, Nil, sourceInfo, compileOptions)

private[chisel3] def apply(fmt: String, data: Seq[Bits], sourceInfo: SourceInfo, compileOptions: CompileOptions): Printf =
     apply(Printable.pack(fmt, data: _*))(sourceInfo, compileOptions)
```
These functions will not be callable from Scala user code in Chisel 3.5.5, but any `chisel3.printf` in user code compiled against 3.5.0-3.5.4 will still be invoking these functions at runtime.

jackkoenig (approved):

Issue #959, opened 2018-12-10
[RFC] Rename WireInit to WireDefault


schoeberl:
When describing combinational circuits a common error is not covering all paths with an assignment. This can happen in Verilog and in VHDL, not in Chisel (which is very good!). To avoid this error common practice is to start with a *default* assignment for the signal. Chisel WireInit does the same. However, the name Init is misleading (relative to RegInit) as it sounds like something is set at initialization time (at reset). A default assignment "executed" alsways. Therefore, I would propose to rename WireInit to WireDefault for better readability.

**Type of issue**: feature request

**Impact**:  API modification

aswaterman:
I have to agree that WireDefault is a clearer name, as unlike RegInit it is unrelated to the concept of reset.  Obviously for compatibility we'd need to keep WireInit, though.

schoeberl:
WireInit can immediately be marked as depreciated. Google found just a single file where it is used: https://github.com/ucb-bar/dsptools/blob/master/rocket/src/main/scala/dspblocks/MemToRegmap.scala

Cheers,
Martin

ducky64:
This was discussed at the dev meeting today. There's agreement that WireDefault is better terminology, though it's unclear whether the impact of even a deprecation phase is worth the gain - especially if hundreds of deprecated warnings start showing up. If we don't decide to change WireInit to WireDefault, one alternative is to give WireInit better documentation, as a wire with an initial assignment.

A code rewriter may mitigate the impact of deprecations.

Note: VecInit may be in a similar situation.

Historical note: WireInit was a refactoring of the Wire(init=...) construct.

aswaterman:
I know my opinion on the matter isn't shared globally, but we don't actually have to deprecate old interfaces.  Could just add WireDefault as an alias for WireInit.  Having only one way to do things is a guideline, not a dogma.

ducky64:
I think the disadvantage to having multiple interfaces that do the same thing is that there's more to learn. Obviously not a problem for you, me, or those who follow these issues - we know what it does - but for someone new who might be learning Chisel from examples, it could be detrimental.

I don't think the benefit of the change outweighs the drawbacks of having two ways to do the same thing.

That being said, I'm not entirely opposed to keeping both around. The main concern would be that at least new users should be aware of the recommended way - so that means some kind of warning (or maybe something more clever?) - though it'd have to be opt-out instead of opt-in because I don't think it's practical to rely on people finding and adhering to style guides.

PR #986, opened 2019-01-17
WireDefault instead of WireInit, keep WireInit around

schoeberl:
**Type of change**: feature request

**Impact**:  API modification

**Development Phase**: implementation

**Release Notes**

Rename WireInit to WireDefault as its semantic is not an initialization (like a register on a reset), but a default assignment for a combinational circuit.

WireInit is kept around as an alias to WireDefault.

Closes #959

PR #2764, opened 2022-10-07
deprecate WireInit

sequencer:
`WireInit` and `WireDefault` stayed together for almost 3 years, see #959
It's reasonable to deprecate one of which and remove `WireInit` in the next Chisel major version.

### Contributor Checklist

#### Type of Improvement

- code cleanup

#### Release Notes
deprecate `WireInit`

- [ ] Did you review?


ekiwi:
How does this relate to `VecInit`?

jerryz123:
What is the incentive for removing WireInit? Many existing projects use WireInit extensively.

aswaterman:
In a SiFive repo, both WireInit and WireDefault are used extensively, but WireInit is by far the more popular choice.

jerryz123:
> In a SiFive repo, both WireInit and WireDefault are used extensively, but WireInit is by far the more popular choice.

Now I less bad about using `WireInit` exclusively

mwachs5:
i feel like we bikeshedded this already, but we already have all the *Init why introduce a new word Default?

mwachs5:
i have no problem if we agree WireDefault is the right name to globally find and replace

jerryz123:
I don't want to debate whether `WireInit` or `WireDefault` is the better name, it seems `WireDefault` will be the recommended API going forwards, and that's fine.

I just want to point out that maintaining the `WireInit` API is zero burden, so I don't see the motivation for removing it (or adding the deprecation flag).

seldridge:
> I just want to point out that maintaining the WireInit API is zero burden, so I don't see the motivation for removing it (or adding the deprecation flag).

It's not zero burden for Chisel in testing, for onboarding Chisel users, and readability of existing Chisel libraries, though. If there's one function for creating a wire with an initial (or default) value that seems like a good language design principle to me. I hope that we can all agree that having one way of doing this would benefit Chisel as a language, even if it requires users to update their code.

Analogously, the LLVM project has always had `llvm::Optional` because the code base didn't support C++ 17 (which added `std::optional`). `llvm::Optional` had differently named member functions that `std::optional`, e.g., `hasValue`/`has_value`, `getValue`/`value`, and `getValueOr`/`value_or`. Recently LLVM moved to compiling with C++17. It was then viewed as bad to deviate from standard C++ conventions for names of member functions that new users were used to seeing. A C++ user expects to see `has_value` and not `hasValue`. New methods were added, the old were deprecated, and everybody migrated (at who knows what cost inside Google and other companies!). See: https://discourse.llvm.org/t/deprecating-llvm-optional-x-hasvalue-getvalue-getvalueor/63716.

> What is the incentive for removing WireInit? Many existing projects use WireInit extensively.

I think these types of questions should be evaluated in terms of the language: Does having one way of initializing a wire make _Chisel_ better? I'm not saying backwards compatibility is bad to aim for, but we need a cost metric other than "anything that requires an existing project to change anything is bad". Other, much larger and much more widely used projects (e.g. LLVM) go through this kind of healthy activity and continuous improvement/refinement at a surprising pace.

aswaterman:
Given that we accept the premise that it is justifiable to deprecate things, why are we choosing to deprecate the version that people have spoken up in favor of using (and is easier to type)?

sequencer:
> How does this relate to VecInit?

I don't think this API relate to `VecInit`, the reason I'm proposing this deprecation is there are two API works exactly same, while it's tough to explain these to new users: "which one should be used, and why two APIs". `VecInit` doesn't have this issue.
 
> What is the incentive for removing WireInit? Many existing projects use WireInit extensively.

The incentive is keeping two same APIs is misleading to new users, since `WireDefault` is merged later, and have a clear semantic, so I'm proposing deprecating the `WireInit` API.

> Given that we accept the premise that it is justifiable to deprecate things, why are we choosing to deprecate the version that people have spoken up in favor of using (and is easier to type)?

Agree, but after #986 was in, I switch to `WireDefualt`, I think for both these two APIs, they are good to have a semantic: "define a hardware Wire, and set the default value with some constant value", however to keep the codebase tidy, I'm proposing keeping Chisel it self as small as possible. Deprecating any one of these two API helps making things clean, like clean the compatibility layer did.

aswaterman (changes requested):
A SiFive-internal discussion considered the merits of `Init` vs. `Default` in this context.  There was sympathy for the notion that `WireInit` and `RegInit` are not actually symmetric.  But there was also some agreement that `VecInit` is consistent with `WireInit`, and it would be even more inconsistent to deprecate `WireInit` without also creating `VecDefault` and deprecating `VecInit`.

My opinion is that we should make no changes until we've thought through how to name _all_ of these constructs, weighing `Init` vs. `Default` (vs. `Reset` vs `...`) for each API.  Then, we can do the new-API creation and old-API deprecation in bulk.  I suspect we'll cause less churn that way.

(Other maintainers should feel free to dismiss this review if there's sufficient agreement to the contrary.)

jerryz123:
> VecInit doesn't have this issue.

Given that `VecInit` internally constructs a `Wire` with a default value, should `VecInit` also be renamed to `VecDefault`?  The line of reasoning here opens the door to such questions. There is a a lot of code which uses `VecInit` and `WireInit` analogously, so `VecInit`/`WireDefault` could also be confusing to new users. 

> like cleaning the compatibility layer did.

The `WireInit` alias seems to be a much smaller maintenance burden than the Chisel2 compatibility layer

> Analogously, the LLVM project has always had llvm::Optional

Fair point. Although in that case, LLVM changed to conform to existing patterns in the STL and C++ community. I don't think the benefit is quite as clear-cut here.

> we need a cost metric other than "anything that requires an existing project to change anything is bad"

I agree, but that does not imply the cost metric shouldn't consider backwards compatibility as a factor at all. It ought to be considered that there is a growing collection of chisel generators without active maintainers, or maintainers who are interested in keeping up with the latest chisel. Preserving backwards compatibility such that these packages can be used in projects with newer chisel versions is good for the chisel community in general.

schoeberl:
It is hard to fix namings after a product is in use. I still strongly support the "default" name, as we say when describing combinational circuits: "we assign a default value to a combinational circuit, especially to avoid describing latches" (not in Chisel but in V). We "initialize" a register on reset.

Yes, ```VecInit``` then does not fit as well. But then it becomes a slippery slope, as we can have a ```RegInit(VecInit(...))```, which is now a reset value for a register.

When we talk about naming things: I don't like ```bits``` in the ```DecoupledIO```. It should be called ```data```. And I don't like ```DecoupledIO```, as it should be called ```ReadyValidIO```. OK, now I stop ;-)

sequencer:
My opinion is: bad names doesn't matter, `WireDefault`/`WireInit` are both OK name, users will get used to it, but two names+same function is creepy... Maybe at the first time we shouldn't merge `WireDefault`...

seldridge:
> It ought to be considered that there is a growing collection of chisel generators without active maintainers, or maintainers who are interested in keeping up with the latest chisel. Preserving backwards compatibility such that these packages can be used in projects with newer chisel versions is good for the chisel community in general.

This is much stronger than considering the cost of changes to code. This is a very strong premise of "I will be able to _compile_ a long dead project and an active, live project _together_ with the newest version of Chisel." It then follows that _any_ backwards-incompatible change to Chisel, however trivial, will violate this premise.

I am arguing that this premise is wrong. A long dead project stuck on an old version of Chisel should be able to be _used_ in a new project with a newer version of Chisel, but they should be _separately compiled_ and linked at the FIRRTL/MLIR or Verilog levels. The version of Chisel used to generate FIRRTL/Verilog is then per-project as opposed to the lowest common denominator (or the newest version, but requiring forever backwards compatibility). Admittedly, Chisel has never provided a good answer to how to do this and, in consequence, projects have commonly chosen monolithic compilation and assemblage via submodules (e.g., Rocket Chip, Chipyard, etc.). Additionally, libraries have been built on top of Chisel that assume monolithic compilation.

Definition/Instance was an approach to help with this, ideas like [`seldridge/chisel-separable`](https://github.com/seldridge/chisel-separable) are other experiments in this space.

WDYT?

(Thanks for this comment, though! I've had a very hard time articulating why seemingly innocuous changes like this are easy for other projects and hard for Chisel. I think your comment sheds a lot of light on the problem.)

PR #2772, opened 2022-10-12
Remove "don't touch" behavior from Trace API

seldridge:
Change the behavior of the Trace API to not have "don't touch" behavior unconditionally.  Removing "don't touch" enables a finer granularity of functionality: if a user wants "don't touch" they can add a "don't touch".  However, they are not locked optimization blocking behavior which may be undesirable in certain situations.


sequencer (approved):
LGTM

PR #2776, opened 2022-10-13
Fix scalaVersion in README

jackkoenig:
Chisel 3.5.4 predates Scala 2.13.10 so the plugin is not published for 2.13.10.


chick (approved):
LGTM

PR #2784, opened 2022-10-17
Add traceNameV2 for backwards compat. of traceName

seldridge:
Add utilities to enable backwards compatibility of the Trace.traceName API to Chisel 3.5.x.  This adds a Trace.traceNameV2 utility which aliases to Trace.traceName.  This also removes the TraceNameAnnotation and renames it TraceAnnotation.  In 3.5.x, traceName will point at TraceNameAnnotation (which has don't touch behavior) and will be deprecated telling people to use traceNameV2 which will point at TraceAnnotation (which does not have don't touch behavior).  This will require fixups to the backport associated with this PR.


jackkoenig (approved):
LGTM! Thanks!

PR #2787, opened 2022-10-18
Enable AOP for CIRCT

sequencer:
I'm using AOP to split verification and rtl in our designs, by appending different `InjectingAspect` to add different coverage and different signals to trace.
This PR enables AOP for MFC with the handover API, however it only add one transform `chisel3.aop.injecting.InjectingTransform`, see: 
```
  firrtl.stage.transforms.Compiler
  ├── chisel3.aop.injecting.InjectingTransform
  └── firrtl.ChirrtlEmitter
```
In the further development. Since `InjectingTransform` doesn't depend on any thing, including `CheckChirrtl` and `CInferTypes`, thus I'm proposing move `InjectingTransform` from SFC directly to ChiselStage, which is straightforward to change `Transform`(`TransformLike[CircuitState]`) to `Phase`(`TransformLike[AnnotationSeq]`).

### Contributor Checklist

- [ ] Did you review?


ekiwi (dismissed):
This looks great, but could you also add a test for this?
Maybe you could change the existing aspect tests so that they get run with both, the `firrtl` and the `firtool` compiler.

src/main/scala/circt/stage/phases/MaybeSFC.scala
```diff
@@ -18,7 +19,16 @@ class MaybeSFC extends Phase {
 
   override def transform(annotations: AnnotationSeq) = {
     val sfcAnnotations: Option[AnnotationSeq] = view[CIRCTOptions](annotations).handover.get match {
-      case CIRCTHandover.CHIRRTL      => None
+      case CIRCTHandover.CHIRRTL      => if (annotations.exists {
+        // TODO[jiuyang]: This is a hotfix to ask SFC to resolve AOP.
+        //                Eventually, I think AOP should be moved to an early stage which out a custom firrl transform
```
mwachs5, line 24: I'm not sure what this comment is saying. Do you mean "without a custom firrtl transform"?

seldridge:
I would prefer to not support _injecting_ aspects from Chisel in an MLIR-based FIRRTL Compiler (MFC) Chisel flow. There are huge benefits towards getting on a single, better FIRRTL compiler and continuing to use the Scala-based FIRRTL Compiler (SFC), even in limited circumstances, places huge constraints on the evolution of Chisel.  E.g., we cannot evolve the FIRRTL spec aggressively and we have to continue to maintain two FIRRTL compilers.

More technically, the main use case for injecting AOPs is to keep some collateral (typically verification collateral) separate from the design. There are alternative approaches that are now available in the MFC, though without Chisel APIs, yet, to enable this type of thing. Using these, we've successfully purged all uses of the injecting aspects in favor of better approaches.

An example, here is a use of a Grand Central Tap to separate verification code into a separate module, use taps to grab values via XMR, and then extract everything out of the design:

```
circuit TestHarness: %[[
  {"class": "sifive.enterprise.grandcentral.DataTapsAnnotation",
   "keys": [
     {
       "class": "sifive.enterprise.grandcentral.ReferenceDataTapKey",
       "source": "~TestHarness|TestHarness/foo:Foo>x",
       "sink": "~TestHarness|TestHarness/foo_assert:Foo_verif>w"
     }
   ]
  }
]]
  module Foo_verif:
    input clock: Clock

    wire w: UInt<1>
    w is invalid

    assert(clock, w, UInt<1>(1), "hello")
    cover(clock, not(w), UInt<1>(1), "world")

  module Foo:
    input a: UInt<1>
    output b: UInt<1>

    node x = a

    b <= x

  module TestHarness:
    input clock: Clock
    input a: UInt<1>
    output b: UInt<1>

    inst foo of Foo
    foo.a <= a
    b <= foo.b

    inst foo_assert of Foo_verif
    foo_assert.clock <= clock
```

Compiling this with `firtool Foo.fir -extract-test-code`:

```verilog
// Generated by CIRCT unknown git version
// VCS coverage exclude_file
module Foo_verif_assert(
  input clock);

  always @(posedge clock)
    assert(TestHarness.foo.a) else $error("hello");
endmodule

// VCS coverage exclude_file
module Foo_verif_cover(
  input clock);

  always @(posedge clock)
    cover(~(TestHarness.foo.a));
endmodule

module Foo(
  input  a,
  output b);

  assign b = a;
endmodule

module TestHarness(
  input  clock,
         a,
  output b);

  Foo foo (
    .a (a),
    .b (b)
  );
  /* This instance is elsewhere emitted as a bind statement.
    Foo_verif_assert Foo_verif_assert (
      .clock (clock)
    );
  */
  /* This instance is elsewhere emitted as a bind statement.
    Foo_verif_cover Foo_verif_cover (
      .clock (clock)
    );
  */
endmodule


// ----- 8< ----- FILE "bindfile" ----- 8< -----

bind TestHarness Foo_verif_assert Foo_verif_assert (
  .clock (clock)
);
bind TestHarness Foo_verif_cover Foo_verif_cover (
  .clock (clock)
);
```

We should totally keep _selecting_ aspects. However, it is a very questionable design pattern to continue to support arbitrary amendment of FIRRTL statements. This makes designs extremely difficult to reason about and dramatically complicates any extremely important goals like enabling separable compilation.

ekiwi:
> There are huge benefits towards getting on a single, better FIRRTL compiler and discontinuing usage of the Scala-based FIRRTL Compiler (SFC), even in limited circumstances, places huge constraints on the evolution of Chisel. E.g., we cannot evolve the FIRRTL spec aggressively and we have to continue to maintain two FIRRTL compilers

How does an Injecting Aspect place restrictions on the firrtl compiler used? I thought it all happened at the Chirrtl level, before any of the compilation.

ekiwi:
> It is calling `RunFirrtlTransform`. This is only working because all the infrastructure in the `firrtl` codebase to "run a firrtl transform" would still exist. Just because we are doing it in one process or another doesn't mean that it is "in chisel".

It is using that in the background though and the CIRCT Stage is already using the `RunFirrtlTransform` mechanism (see the code below the proposed addition). I assume that this will be replaced with an alternative mechanism in the future.

sequencer:
Basically, this is how I use the AOP:
The target is: RTL designers only force on RTL designing, without any attention on the verification, and tapeout. Thus there are no options like `debug = true`, `fpga = true`, `tape = true` in that pure RTL generator, even not `printf`, no `cover`, no `assert` and `assume`.
This helps the code review flow and git flow, we don't want a three-way merge because the verification engineer and RTL engineer both changed the same file.

The AOP is the feature I used to achieve this goal:
1. I use mill as build system, split the Scala module into `rtl` and `verif`. of cause the `verif` depends on the `rtl`, the `rtl` module only contains the Chisel RTL code.(not test, no verification, no elaboration), it is used for distribution and depend by other projects.
2. in the `verif` module, we use this `Jvm.runSubprocess` to elaborate the SV from the main class of `verif`, e.g.
```
    def rtls = T {
      mill.modules.Jvm.runSubprocess(
        finalMainClass(),
        runClasspath().map(_.path),
        forkArgs(),
        forkEnv(),
        Seq("--dir", T.dest.toString),
        workingDir = forkWorkingDir(),
      )
      val artifacts = os.walk(T.dest)
      T.log.info(s"RTL generated:\n${artifacts.mkString("\n")}")
      artifacts.map(PathRef(_))
    }
```
The artifacts(trace annotations and verilog) will be put in the the `T.dest`, the `main` class will consume the destination folder( provided by the build system) RTL config(provided by the build system as well).
The verification flow, `debug`, `cover`, `ci` are inserted by different AOP, verification engineers can directly use them.

mwachs5:
In traditional verilog workflow AOP is achieved by bind mechanism which is basically remote instantiating a submodule into an existing module (vs injecting verilog statements directly into an existing module). MFC already has support for this version so it could be better to focus on firrtl spec and good chisel APIs to link up to it. Fundamentally, if we are doing manipulations in scala on firrtl circuit (post convert) we are using SFC, regardless of which process or stage we run the pass in.

sequencer:
> MFC already has support for this version so it could be better to focus on firrtl spec and good chisel APIs to link up to it.

Yes, I admit this, I have two argues here:
1. Before we reach into good chisel APIs for traditional Verilog, we should still keep this feature, since it does work with MFC without any maintaining burden.
2. The remote instantiation sounds good and suitable for providing a good API, but before verilator supporting it, or any other MFC-based simulator support this, we don't have a good way to test these good features in the open-source world, thus we don't cannot have CI for them.

So if we can do this without any burden, why not?

mwachs5:
>  since it does work with MFC without any maintaining burden.

I'm a little confused by this statement, because my understanding of the current API is that it emits snippets of firrtl, then more snippets of firrtl to be appended to the end of an existing module as an annotation. SFC is then responsible for linking appropriately, and MFC does NOT support that linking. If we can do the "linking" in chisel IR instead of firrtl IR, then I agree there is no burden on MFC / dependence on SFC. But if we are doing the linking in firrtl IR,  I'm not sure how you are suggesting to do this?

seldridge:
My alternative proposal is to add direct support for Verilog `bind`. This is already understood by the internal pipeline of CIRCT via a `lowerToBind` attribute that can be added to instances and is used to support Grand Central and Extract Test Code.

While I think this is something that should be put in the FIRRTL spec (`bound_instance("bindfile.sv") foo of Foo`), it can be initially bridged using an MFC-only annotation. 

Consider something like the following. This moves all the verification logic into a `Foo_verif` module. This module is then instantiated inside `Foo` and hooked up to anything it needs.

```
circuit Foo: %[[
  {
    "class": "jiuyang.bind",
    "target": "~Foo|Foo/verif:Foo_verif"
  }
]]
  module Foo_verif:
    input clock: Clock
    input reset: UInt<1>
    input a: UInt<8>
    input b: UInt<8>

    cover(clock, eq(a, UInt<8>(0)), reset, "a is zero")
    cover(clock, eq(b, UInt<8>(0)), reset, "b is zero")

  module Foo:
    input clock: Clock
    input reset: UInt<1>
    input a: UInt<8>
    input b: UInt<8>
    output c: UInt<9>

    c <= add(a, b)

    inst verif of Foo_verif
    verif.clock <= clock
    verif.reset <= reset
    verif.a <= a
    verif.b <= b
```

With a [tiny change to CIRCT](https://github.com/llvm/circt/compare/dev/seldridge/bind-demo), this compiles to:

```verilog
// Generated by CIRCT unknown git version
module Foo_verif(
  input       clock,
              reset,
  input [7:0] a,
              b);

  always @(posedge clock) begin
    if (reset) begin
      cover(a == 8'h0);
      cover(b == 8'h0);
    end
  end // always @(posedge)
endmodule

module Foo(
  input        clock,
               reset,
  input  [7:0] a,
               b,
  output [8:0] c);

  /* This instance is elsewhere emitted as a bind statement.
    Foo_verif verif (
      .clock (clock),
      .reset (reset),
      .a     (a),
      .b     (b)
    );
  */
  assign c = {1'h0, a} + {1'h0, b};
endmodule


// ----- 8< ----- FILE "bindfile" ----- 8< -----

bind Foo Foo_verif verif (
  .clock (clock),
  .reset (reset),
  .a     (a),
  .b     (b)
);
```

@sequencer: is something like this workable?

sequencer:
> @sequencer: is something like this workable?

Thanks @seldridge! The reason I'm using AOP is that I don't want to make verification thing visible to RTL designer, while verification enginner is still able to modify RTL.
I think `lowerToBind` does help splitting the elaboration result of SystemVerilog. My main concern is where is the place I should add the binding attribute, and how do I write the `Foo_verif`.
I think my final solution to this is something like the grand central annotation, but implemented with the Trace API to split the DUT and TB in two phases. The reason I'm sticking on this is I have observed some tricky design flows in rocket-chip, for example the `TLMonitor` and diplomacy hacks for it.

and another `gray box` usage is my way to replace the `fpga-shell` that hooks the blackbox inside RTL and linking device specific modules, it works with FPGA DRAM island generation, but also is a convenient way to hook clock gen module(PLL), IO Cell(including IO Mux), ICG and other process/flow specific designs.

sequencer:
> If we can do the "linking" in chisel IR instead of firrtl IR, then I agree there is no burden on MFC / dependence on SFC

Yes, this is true, I thought about it today: basically the technical issue Chisel IR doesn't support Namespace, which means when injecting, we need to go though each module to collect their names. 
I think there are three ways implementing this purely in Chisel:
1. add new feature to compute a module Namespace from ChiselIR, this need a traverse the module for each time of injection, it's simple but may lead unnecessary traverse to modules if there are multiple injections(personally I'd like to arrange test points in different `InjectStatement`, and inject to DUT when necessary).
1. consuming `InjectStatement` in the `Convert` phase, this changes the behavior of **injection** from "after elaborating" to "before elaborating", this will block some usage: copy a `CircuitCircuitAnnotation` and have different injections.
1. consuming `InjectStatement` in the `InjectingPhase`, but add new `NamespaceAnnotation(module: ModuleTarget, namesapce: Namespace)` to passthough the namespace.

Personally I like the third option.

sequencer:
Here are two issues to be discussed of this PR:
1. @seldridge should AOP still exist in the Chisel?  
2. @mwachs5 the InjectingPhase is using SFC features, which won't be maintained anymore.

For the first issue, AOP is considering a bad programming paradigm, I think this is exactly true, since injecting RTL from the out side of the `Module` will lead to a non-deterministic result, like what diplomacy gave us: a flexible user experience but everything is a mess: "- why here is a register? - Injected via `InModuleBody`"

On the other hand, when we start discuss about the verification, AOP is still a good(or maybe the best) practice for design. verification engineers are able to inject any verification logic to Module, bind signals from the module without drilling signals from the module hierarchy, while for RTL engineers, those verification codes are totally agnostic.
@seldridge gave a better idea about a verification specific `bind-in`(remote instantiate) in like this:
```
class Foo_verif extends RawModule
class Foo extends RawModule

class Top extends RawModule {
  val foo = Module(new Foo)
  bind(Module(new Foo_verif)).in(foo)
}
```
This feature depends on the `RefOp` and `BindOp` in the future FIRRTL. Rather than injecting arbitrary IRs after elaboration. the `Foo_verif` forced the remote instantiated code to be to verification specific, and it should never have side effects. It is awesome to provide a real SystemVerilog verification experience.

The second tricky usage is that: maybe AOP is the only easy way to linking graybox:
1. replace the arithmetic modules like adder, multiplier to hard macros.
2. replace the process/fpga specific blackbox island like PLL, PCIe, DDR.

I think the AOP is still need after we migrate to MFC:
1. It's the only API in Chisel 3.6+ to let user being able to split verification and RTL codes. 
2. It's the only elegant way in Chisel 3.6+ to let user being able to replace graybox in Modules.

Basically, it's the only "official" method that Chisel provides to users to view the RTL after elaboration. We should limit its functionality, since abusing it will let the compiling flow too mutable and difficult to maintain. But there are still reasonable needs for this API: add `tracename`, mark `dont touch` to signals, flatten instance for post synthesis designs, replace design specific blackboxes like `ICG`, `PLL`, `DDR Phy`.
So I may suggest that we should adopt this PR to Chisel for and wait for `RefOp` and `bind-in` API. After these API lands to Chisel code base, we can start to limit the feature of AOP by migrating the `InjectingPhase` to MFC pass, and in that pass, we can provide a strict mode to force only some safe IRs injected via AOP.


For the second issue, by @mwachs5, "the InjectingPhase is using SFC features, which won't be maintained anymore."
I don't think so: AOP only needs the `Convert` pass, while it only uses the IR from the SFC. So does MFC: MFC depends on the `Convert` as well, the injection doesn't depends on any transforms or states or pass from SFC, e.g `WIR`, `InferType` etc, it just simply injecting `firrtl.ir` to another `firrtl.ir`, after injection finished, it will be emitted directly.
So does this usage really doesn't matter? I think the only burden is when we finally start to use JNA/JNI to access MFC: get rid of the `Converter`, In that time, the AOP is really a new feature that should be re-developed in MFC.

ekiwi:
> @seldridge gave a better idea about a verification specific `bind-in`(remote instantiate) in like this:

That solution only binds to a particular instance. However, how would you instantiate verification code for all instances of a particular module? For example, let's say that you have some code that generates assertions for all Fifos in the design. How can you make sure that it runs on every instance in order to generate the right assertions taking into account the Fifo depth and other design parameters.

mwachs5:
> However, how would you instantiate verification code for all instances of a particular module? 

We are NOT suggesting to get rid of the Select APIs that would allow you to do this, e.g. from a testharness crawling over the elaborated DUT and attaching verif code.
You would just not be able to do it with an *injection* -- this code would live in the testbench, bound individually to each module instance.

ekiwi:
> We are NOT suggesting to get rid of the Select APIs that would allow you to do this, e.g. from a testharness crawling over the elaborated DUT and attaching verif code.
> You would just not be able to do it with an _injection_ -- this code would live in the testbench, bound individually to each module instance e.g. with cross-module references.

Do you have a code snippet that shows how that would work?

jackkoenig (approved):
We discussed this last week in Chisel dev. The long term plan remains to get rid of InjectingAspects, but this PR is a reasonable bridge for Chisel 3.6 and at least pulls the InjectingAspect API out of the FIRRTL compiler and in to just Chisel itself.

We will continue to discuss use cases and possible better future APIs, but this is good for the time being.

PR #2791, opened 2022-10-19
Added distinguishing directory names to source locators

azidar:
We currently only emit the filename with each source locator. The issue with this approach is if you have multiple files with the same name (e.g. `package.scala`), it is unclear which file it originated from.

If more parent directory information is included, you can distinguish between these files. However, knowing how much information to include (a path relative to.. what?) is difficult to answer without user input.

As a heuristic, we strip the absolute path of a filename relative to the user directory that the JVM is invoked in. We also provide a chisel option to disable this behavior and just give the filename (as was done previously).

TODO: Write tests, and solicit community feedback.

### Contributor Checklist

- [ ] Did you review?


jackkoenig (changes requested):
I think the goal is good but see the note about how filtering the filename at Chisel runtime is not great because it breaks reproducibility and hermetic build flows. It is very important that the actual class or jar files that come out of running Scala compilation are identical irrespective of where they are compiled. Thus I think the filtering of absolute paths _must_ happen at macro time, not at Chisel runtime.

I think there is a simpler solution that does not have these negative properties:

Have the user provide a Java system property that sets the base directory of the build and strip that from the absolute path. I think the implementation of this is actually really simple, the only real complexity is in choosing what to do if the user does not provide this argument (my suggestion would be to default to just the local file name and print a deprecation warning that will later become an error).

ekiwi:
I like the idea as a good and simple fix, much simpler to implement than trying to engineer the perfect source locator format.

The line coverage pass that I have implemented for Chisel will benefit from this since I was struggling with files of the same name before.

One thing I am curious about: how does this interact with precompiled libraries (JARs) the sources of which are not part of the current project.

jackkoenig:
> One thing I am curious about: how does this interact with precompiled libraries (JARs) the sources of which are not part of the current project.

Also good point, this approach does not work with published Chisel binaries. What result are you going to get building your project in `/work/azidar/cool-project/` when source locators from the Chisel published jar are of the form `/Users/jackk/work/chisel3/src/main/scala/chisel3/util/Decoupled.scala`?

core/src/main/scala/chisel3/internal/Trie.scala
```diff
@@ -0,0 +1,139 @@
...
+import scala.collection.mutable
+
+/** A prefix tree. Prefixes represented by a list of `Token` values. For each
+  * valid path the tree will contain a value of type `Value`.
+  * 
+  * For more information on Trie's, visit https://en.wikipedia.org/wiki/Trie
+  *
+  * To use this, do not extend this trait (it is sealed). Instead, create an empty trie with
+zidar marked this conversation as resolved.
+how resolved
```
seldridge, line 15: Stray comment.

core/src/main/scala/chisel3/internal/SourceInfo.scala
```diff
@@ -43,17 +44,38 @@ case object DeprecatedSourceInfo extends NoSourceInfo
...
+
+
 /** Provides a macro that returns the source information at the invocation point.
   */
 object SourceInfoMacro {
   def generate_source_info(c: Context): c.Tree = {
     import c.universe._
     val p = c.enclosingPosition
-    q"_root_.chisel3.internal.sourceinfo.SourceLine(${p.source.file.name}, ${p.line}, ${p.column})"
+    q"_root_.chisel3.internal.sourceinfo.SourceLine(${p.source.file.canonicalPath}, ${p.line}, ${p.column})"
```
jackkoenig, line 78: A sketch of the solution I'm proposing:

```suggestion
    val root = sys.env.get("chisel_project_root").getOrElse(/* Figure out what to do if not provided*/
    val path = p.source.file.canonicalPath.stripPrefix(root) // TODO warn/error if nothing is stripped
    q"_root_.chisel3.internal.sourceinfo.SourceLine($path, ${p.line}, ${p.column})"
```

Then the user in their `build.sbt` just needs something like:
```scala
Compile / javaOptions += "-Dchisel_project_root=" + baseDirectory.value
```
And I think that's basically all we need.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -424,9 +424,34 @@ private[chisel3] class DynamicContext(
...
+    val paths = SourceFiles.getSourceFilePaths()
+    paths.foreach { p => computor.addPath(p) }
+    computor
+  }
+}
+
+object SourceFiles {
+  import java.io.File
+  def getSourceFilePaths(): List[Seq[String]] = {
+    def rec(file: File): List[Seq[String]] = {
```
ekiwi, line 438: Have you considered using `os-lib` instead of the `java.io` package? Some of the functions in the `os` package are much nicer to work with and most new code in `firrtl` and `chiseltest` has made use of them.

ekiwi: In this case you would work with a `os.Path` instead of a `File`

azidar:
Thanks for the feedback folks, I'm in agreement that the java option is a better solution. I'll work on that and get back to y'all. Thanks!

core/src/main/scala/chisel3/internal/Builder.scala
azidar: This isn't needed anymore with relying on the "user-dir" system property.

azidar:
Ok, updated the PR. Turns out there is a `user-dir` system property that effectively does the `-Dchisel_project_dir` option automatically. If there are weird things where the user-dir is not where we'd expect or the file size is too big, you can give chisel the option `--omit-source-locator-paths` to omit the path and just emit the filename.

docs/src/explanations/source-locators.md
```diff
@@ -0,0 +1,27 @@
...
+wire w : UInt<3> @[src/main/scala/MyProject/MyFile.scala 1210:21]
+```
+
+In a Verilog file, it looks like this:
+
+```verilog
+wire [2:0] w; // @[src/main/scala/MyProject/MyFile.scala 1210:21]
+```
+
+By default, the file's relative path to where the JVM is invoked is included. To change where the relative path is computed, set the Java system property `-Dchisel_project_root=/absolute/path/to/root`. This option can be directly passed to sbt (`sbt -Dchisel_project_root=/absolute/path/to/root`). Setting the value in the `build.sbt` file won't work because it needs to be passed to the JVM that invokes sbt (not the other way around). We expect this only relecant for publishing versions which may want more customization.
```
mwachs5, line 24:
```suggestion
By default, the file's relative path to where the JVM is invoked is included. To change where the relative path is computed, set the Java system property `-Dchisel_project_root=/absolute/path/to/root`. This option can be directly passed to sbt (`sbt -Dchisel_project_root=/absolute/path/to/root`). Setting the value in the `build.sbt` file won't work because it needs to be passed to the JVM that invokes sbt (not the other way around). We expect this is only relevant for publishing versions which may want more customization.
```

docs/src/explanations/source-locators.md
```diff
@@ -0,0 +1,27 @@
...
+
+Alternatively, you can omit the relative path entirely (pre-3.6 release) by passing the Chisel option
```
mwachs5, line 26: not sure what you mean by "pre-3.6 release" do you mean to get behavior same as 3.5.4 and earlier?

jackkoenig (commented):
This looks awesome, nice work 🙂. I have been trying to "break" the use of `user.dir` (to prove a point) and I have found it pretty hard to do. It seems that both SBT and mill are pretty sensitive to where Java is run and basically be construction require that it is run in a parent directory to the source. This is good for users of Chisel in that this default behavior should work well for most everyone.

I do have 2 questions:
1. If the source file is someone not actually present in the root path, should we print a warning?
2. Should we provide an option to just use the filename via a System property? I don't know how often library writers would want this, but it could potentially be a performance improvement if someone did want it.

These are both a bit of theoretical concerns so maybe we should just deal with them if/when they come up.

docs/src/explanations/source-locators.md
```diff
@@ -0,0 +1,24 @@
...
+wire w : UInt<3> @[src/main/scala/MyProject/MyFile.scala 1210:21]
+```
+
+In a Verilog file, it looks like this:
+
+```verilog
+wire [2:0] w; // @[src/main/scala/MyProject/MyFile.scala 1210:21]
+```
+
+By default, the file's relative path to where the JVM is invoked is included. To change where the relative path is computed, set the Java system property `-Dchisel.project.root=/absolute/path/to/root`. This option can be directly passed to sbt (`sbt -Dchisel.project.root=/absolute/path/to/root`). Setting the value in the `build.sbt` file won't work because it needs to be passed to the JVM that invokes sbt (not the other way around). We expect this only relevant for publishing versions which may want more customization.
```
jackkoenig, line 24:
```suggestion
By default, the file's relative path to where the JVM is invoked is included.
To change where the relative path is computed, set the Java system property `-Dchisel.project.root=/absolute/path/to/root`.
This option can be directly passed to sbt (`sbt -Dchisel.project.root=/absolute/path/to/root`).
Setting the value in the `build.sbt` file won't work because it needs to be passed to the JVM that invokes sbt (not the other way around).
We expect this only relevant for publishing versions which may want more customization.
```
Split into 1 sentence per line which is generally better for future diffs and a lot better for folks using editors without line wrapping support

jackkoenig (approved):

PR #2797, opened 2022-10-21
Fixes #2796 for assigning to BlackBoxes with Chisel._ fields in their IOs

### LINKED ISSUE #2796: Issues connecting to BlackBox with Chisel._ fields in its IO
**Type of issue**: Bug Report

Scastie:

https://scastie.scala-lang.org/dg1ZsSbfQUiWLnEYkj7Ehg

Code is emitted incorrectly (the direction of the `<-` operator is incorrect an fir compiler complains about source/sink flow.

See the firrtl below:

```
circuit Top :
  module FooModule :
    input clock : Clock
    input reset : Reset
    output io : { flip quz : { q : { valid : UInt<1>, bits : { foo : UInt<1>, bar : UInt<1>}}}, foo : UInt<1>, flip bar : UInt<1>}

    io.bar is invalid @[main.scala 35:12]
    io.foo is invalid @[main.scala 35:12]
    io.quz.q.bits is invalid @[main.scala 35:12]
    io.quz.q.valid is invalid @[main.scala 35:12]

  extmodule FooMirrorBlackBox :
    output quz : { q : { valid : UInt<1>, bits : { foo : UInt<1>, bar : UInt<1>}}}
    input foo : UInt<1>
    output bar : UInt<1>
    defname = FooMirrorBlackBox

  module Top :
    input clock : Clock
    input reset : UInt<1>

    inst foo of FooModule @[main.scala 44:25]
    foo.clock <= clock
    foo.reset <= reset
    inst mirror of FooMirrorBlackBox @[main.scala 45:28]
    mirror.bar is invalid
    mirror.foo is invalid
    mirror.quz is invalid
    foo.io.bar <= mirror.bar @[main.scala 46:16]
    mirror.foo <= foo.io.foo @[main.scala 46:16]
    mirror.quz.q.bits <- foo.io.quz.q.bits @[main.scala 46:16].     <--- backwards
    foo.io.quz.q.valid <= mirror.quz.q.valid @[main.scala 46:16]
```

Code would be emitted correctly with the right flippedness on the `<-` if necessary.

3.5.4
  
**Other Information**

I believe I traced this down to the traceFlow function here:

https://github.com/chipsalliance/chisel3/blob/77a6c93592d5766d66f199720fc6d69478005091/core/src/main/scala/chisel3/internal/MonoConnect.scala#L326

For the purposes of determining what can be on the LHS of a `<-` or `<=` operator, shouldn't this be looking at `direction`, not `specifiedDirection`?

Having blackboxes connect properly.

mwachs5:
### Contributor Checklist

#### Type of Improvement

   - bug fix                           

#### API Impact

This fixes #2796, but it also greatly changes how many `<=` vs `<-` we will emit. Previously when connecting two records if we decided we couldn't emit a `<=`, if we were allowed to we would just emit a `<-`. This code path is hittable when connecting a BlackBox IO becasue we are not "allowed" to emit `<=` for blackboxes, but the legacyConnect does not do proper analysis about the directionality before issuing the connect.

The solution here is to stop treating compatibility mode so specially in this case, and just go into the `recordConnect` function and within *that* do/waive the check about mismatching fields, which should be the only reason we can't connect at this point.

This would create a substantial difference in generated firrtl if there are other cases where we are relying on the legacy connect to allow us to connect mismatched Chisel bundles, because now we will emit them as individual `<=` where applicable instead of relying on `<-` semantics to do the same in the firrtl compiler. This *might* not be correct / equivalent (may be more or less strict than `<-` would be) so I would defer to the reviewers if this is acceptable change.

Many more blasted-out `<=` instead of `<-` in cases where fields don't match up in compatbility bundle assignments.

#### Desired Merge Strategy

#### Release Notes

Change the emission of records with Compatibility fields that connect to blackboxes or have mismatched fields to be blasted out to individual fields vs relying on emitting `<-`. Fixes #2796.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -231,9 +231,10 @@ private[chisel3] object BiConnect {
     val recordConnectFieldsMustMatch =
       left_r.compileOptions.connectFieldsMustMatch || right_r.compileOptions.connectFieldsMustMatch
 
+    val connectFieldsMustMatch = /*connectCompileOptions.connectFieldsMustMatch &&*/ recordConnectFieldsMustMatch
     // For each field in left, descend with right.
     // Don't bother doing this check if we don't expect it to necessarily pass.
-    if (connectCompileOptions.connectFieldsMustMatch || recordConnectFieldsMustMatch) {
+    if (connectFieldsMustMatch) {
       for ((field, right_sub) <- right_r.elements) {
         if (!left_r.elements.isDefinedAt(field)) {
           throw MissingLeftFieldException(field)
@@ -246,7 +247,7 @@ private[chisel3] object BiConnect {
         right_r.elements.get(field) match {
           case Some(right_sub) => connect(sourceInfo, connectCompileOptions, left_sub, right_sub, context_mod)
           case None => {
-            if (connectCompileOptions.connectFieldsMustMatch || recordConnectFieldsMustMatch) {
+            if (connectFieldsMustMatch) {
               throw MissingRightFieldException(field)
             }
           }
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -179,9 +179,10 @@ private[chisel3] object BiConnect {
           )
         ) {
           pushCommand(Connect(sourceInfo, leftReified.get.lref, rightReified.get.lref))
-        } else if (!emitStrictConnects) {
-          newLeft.legacyConnect(newRight)(sourceInfo)
         } else {
+          // May be connecting to a blackbox or have missing fields in compatibility mode, etc
+          // Emitting a legacyConnect here may do the wrong thing w.r.t directionality, so
+          // blast out the connection so we can perform the proper analysis.
           recordConnect(sourceInfo, connectCompileOptions, left_r, right_r, context_mod)
         }
 
@@ -227,9 +228,14 @@ private[chisel3] object BiConnect {
     context_mod:           RawModule
   ): Unit = {
     // Verify right has no extra fields that left doesn't have
-    for ((field, right_sub) <- right_r.elements) {
-      if (!left_r.elements.isDefinedAt(field)) {
-        if (connectCompileOptions.connectFieldsMustMatch) {
+    val recordConnectFieldsMustMatch =
+      left_r.compileOptions.connectFieldsMustMatch || right_r.compileOptions.connectFieldsMustMatch
+
+    // For each field in left, descend with right.
+    // Don't bother doing this check if we don't expect it to necessarily pass.
+    if (connectCompileOptions.connectFieldsMustMatch || recordConnectFieldsMustMatch) {
+      for ((field, right_sub) <- right_r.elements) {
+        if (!left_r.elements.isDefinedAt(field)) {
           throw MissingLeftFieldException(field)
         }
       }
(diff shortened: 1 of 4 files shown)


mwachs5:
oops, i committed the wrong version of the test, cleaning it up...

azidar:
So if I'm understanding this change correctly, does it mean we will never call the legacy bulk connect from a `chisel3.<>` ?

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -179,9 +179,10 @@ private[chisel3] object BiConnect {
           )
         ) {
           pushCommand(Connect(sourceInfo, leftReified.get.lref, rightReified.get.lref))
-        } else if (!emitStrictConnects) {
-          newLeft.legacyConnect(newRight)(sourceInfo)
```
azidar, line 183: I'm confused as to why you are not emitting it, but rather correctly determining the `newLeft`/`newRight` ?

mwachs5: how should i correctly determine the newLeft/newRight? And how different will it be from just using the algorithms we already have?

src/test/scala/chiselTests/CompatibilityInteroperabilitySpec.scala
```diff
@@ -390,4 +390,137 @@ class CompatibilityInteroperabilitySpec extends ChiselFlatSpec {
...
+        val in = IO(Input(if (swap) new Compat.FooBundle else new FooBarBundle))
+        val out = IO(Output(if (swap) new FooBarBundle else new Compat.FooBundle))
+        out <> in
+      }
+    }
+    val thrown0 = the[chisel3.internal.ChiselException] thrownBy (compile(new Chisel3.MyModule(true)))
+    thrown0.getMessage should include("Right Record missing field (bar)")
+    val thrown1 = the[chisel3.internal.ChiselException] thrownBy (compile(new Chisel3.MyModule(false)))
+    thrown1.getMessage should include("Left Record missing field (bar)")
+  }
```
jackkoenig, line 497: I'm not against this stricter behavior in principle (we are try to get rid of compatibility mode after all), but this change makes this PR much more than a bug fix. This exact test (at least with `false` as the argument) works on Chisel 3.5.4: https://scastie.scala-lang.org/tdeqjGZqTMWXt3Swi4Nozg

mwachs5: i can easily make it work if that is the behavior we want. I wasn't sure what the behavior was supposed to be

mwachs5: it also works with true (as long as you initialize) https://scastie.scala-lang.org/jQaddfnrQaaE5Ar4TP9pzQ

mwachs5: so I will relax the check so that as long as one of the sides of the connection is Chisel you are allowed to have mismatched fields

jackkoenig: Yeah that sounds reasonable. We'll be tightening up the behavior here when we delete compatibility mode so keeping this zombie alive a little bit longer is the right move for now (esp. for a bug fix).

mwachs5: relaxed and updated the test

mwachs5:
This is still more strict if you were using a compile options with emitStrictConnects = false and connectFieldsMustMatch = true, but presumably if you set that for the latter you really meant it even for compatibility mode.

core/src/main/scala/chisel3/internal/BiConnect.scala
mwachs5: Took your suggestion, #2801 is the version of this that just fixes the canBeSink/canBeSource logic and keeps the legacy connects here.

mwachs5:
replaced by #2801

PR #2804, opened 2022-10-24
Publish an Initial Roadmap

mwachs5:
### Contributor Checklist

#### Type of Improvement

   - documentation      

#### API Impact

No impact

No impact

#### Desired Merge Strategy

#### Release Notes

Publish an Initial Roadmap

- [ ] Did you review?

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,89 @@
+# Chisel Roadmap
+
+This roadmap captures the current plans of the Chisel maintainers and can be
+changed through the usual PR process.
+Not every small feature add or planned bugfix is logged here, but the high-level
+changes and planned direction that are relevant for those using and contributing
+to the project.
+Two especially notable directions on the roadmap below are:
+
+* We are migrating Chisel to rely on, exclusively, the
+  [MLIR-based FIRRTL Compiler](https://github.com/llvm/circt)
+* Accelerating improvements and changes to Chisel and the FIRRTL specification
+
+
+## 3.5.x (and Earlier)
+
+- Continue to maintain binary compatibility on these releases
+- Continue to backport bug fixes, new features if possible, documentation
+  updates, etc.
+- Deprecations are backported to ease in users migrating to next major release
+  (3.6).
+
+## 3.6 (Target 2022 November)
+
+This release is intended to provide a smooth path for the Chisel codebase to no
+longer depend on the 
+[Scala FIRRTL Compiler (SFC)](https://github.com/chipsalliance/firrtl)
+and use the [MLIR Firrtl Compiler (MFC)](https://github.com/llvm/circt) as the
+main backend. While the SFC may continue to work to consume the `.fir` produced
+by Chisel, it will not be a requirement for adding new features to the FIRRTL
+spec and the Chisel APIs. Thus, the 3.6 release will:
+
+- Deprecate `Chisel._`
+- Turn the deprecations from previous releases into Errors
+- Support integration with [MFC](https://github.com/llvm/circt) by bringing
+  `chisel-circt` in-repo
+- Stop using [SFC](https://github.com/chipsalliance/firrtl) except to test
+  deprecated features
+- Deprecate all Chisel features that are not implemented in MFC. A nonexhaustive
+  list includes:
+  - Interval Types
+  - Fixed Point Types
+  - Injecting Aspects
+  - EnumAnnotations
+  - RunFirrtlTransformAnnotation
+  - LoadMemoryFromFile
+  - Trace API
+  - ChiselAnnotation/custom annotations
+- Publish Roadmap
+
+
+## 5.0.0 (Target Q1 2023)
+
+As we look forward to Chisel development, we plan to accelerate coordinated
+changes with the
+[FIRRTL specification](https://github.com/chipsalliance/firrtl-spec) to add new
+functionality. The bump to version 5.0.0 will be a major change in the overall
+versioning scheme of the Chisel project. We will switch to
+[Semantic Versioning 2.0.0](https://semver.org/), which means we expect the
+MAJOR version number to increment much more frequently. To make this change more
+apparent, these changes will be made in the Chisel 5.0.0 release:
+
+- The build artifact will be renamed to `org.chipsalliance.chisel`
+- The repository will be renamed to github.com/chipsalliance/chisel (the
+  `chisel3` URL will still work as an alias).
+- All of Chisel’s internal tests will rely on MFC, not SFC, for lowering from
+  firrtl
+- Delete Chisel._
+- The repo will be flattened with its dependencies/dependents as much as
+  possible
+  - Bring all things in from SFC into chisel repo that are needed(FIRRTL IR,
+    Stage, etc)
+  - chisel-test (or a subset of it that works with MFC) will be brought into the
+    chisel repo
+  - The chisel-lang website will be brought in-repo
+- Only the chisel repo will be released as an artifact (as all supported
+  dependents and dependencies will be flattened in)
+
+## Beyond 5.0.0
+
+Faster co-improvements to Chisel, MFC, the FIRRTL IR and Chisel.
+
+## Q & A
+
+* What happened to Chisel 4?
+
+We want to make it clear that there is a change in development versioning and
+philosophy (moving faster), so we are skipping 4 and moving straight from
+chisel3 to Chisel 5.0.0.
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -49,6 +49,10 @@ spec and the Chisel APIs. Thus, the 3.6 release will:
 - Publish Roadmap
 
 
+## 4.0
+
+See the Q&A below.
+
 ## 5.0.0 (Target Q1 2023)
 
 As we look forward to Chisel development, we plan to accelerate coordinated
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -44,7 +44,6 @@ spec and the Chisel APIs. Thus, the 3.6 release will:
   - EnumAnnotations
   - RunFirrtlTransformAnnotation
   - LoadMemoryFromFile
-  - Trace API
   - ChiselAnnotation/custom annotations
 - Publish Roadmap
 
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Chisel is powered by [FIRRTL (Flexible Intermediate Representation for RTL)](htt
   * [Chisel3 Architecture Overview](#chisel3-architecture-overview)
   * [Chisel Sub-Projects](#chisel-sub-projects)
   * [Which version should I use?](#which-version-should-i-use)
+  * [Chisel Roadmap](#roadmap)
 
 ---
 
@@ -388,3 +389,7 @@ If you're developing a Chisel library (or `chisel3` itself), you'll probably wan
 By default, the master branches of the chisel repositories are configured to build and publish their version of the code as `Z.Y-SNAPSHOT`.
 Updated SNAPSHOTs are publised on every push to master.
 You are encouraged to do your development against the latest SNAPSHOT, but note that neither API nor ABI compatibility is guaranteed so your code may break at any time.
+
+### Roadmap
+
+See [ROADMAP.md].


ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+  deprecated features
+- Deprecate all Chisel features that are not implemented in MFC. A nonexhaustive
+  list includes:
+  - Interval Types
+  - Fixed Point Types
+  - Injecting Aspects
+  - EnumAnnotations
+  - RunFirrtlTransformAnnotation
+  - LoadMemoryFromFile
+  - Trace API
```
seldridge, line 47: This is now supported!

seldridge (approved):
LGTM!

Let's get this train rolling.

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+
+- The build artifact will be renamed to `org.chipsalliance.chisel`
+- The repository will be renamed to github.com/chipsalliance/chisel (the
+  `chisel3` URL will still work as an alias).
+- All of Chisel’s internal tests will rely on MFC, not SFC, for lowering from
+  firrtl
+- Delete Chisel._
+- The repo will be flattened with its dependencies/dependents as much as
+  possible
+  - Bring all things in from SFC into chisel repo that are needed(FIRRTL IR,
```
davidbiancolin, line 71: Are these things enumerated somewhere?

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+- Turn the deprecations from previous releases into Errors
+- Support integration with [MFC](https://github.com/llvm/circt) by bringing
+  `chisel-circt` in-repo
+- Stop using [SFC](https://github.com/chipsalliance/firrtl) except to test
+  deprecated features
+- Deprecate all Chisel features that are not implemented in MFC. A nonexhaustive
+  list includes:
+  - Interval Types
+  - Fixed Point Types
+  - Injecting Aspects
```
sequencer, line 43: AOP doesn't heavily depend on SFC. Maybe we can still provide this.

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+  - EnumAnnotations
+  - RunFirrtlTransformAnnotation
+  - LoadMemoryFromFile
```
ekiwi, line 46: Could you provide a replacement for this? I know @seldridge wanted to move things like this into the IR which seems like a good idea.

ROADMAP.md
schoeberl: Is Chisel 5 then Scala 3?
Can we get the bit width as part of the Scala type and err on the assignment of different widths?

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+- Support integration with [MFC](https://github.com/llvm/circt) by bringing
+  `chisel-circt` in-repo
+- Stop using [SFC](https://github.com/chipsalliance/firrtl) except to test
+  deprecated features
+- Deprecate all Chisel features that are not implemented in MFC. A nonexhaustive
+  list includes:
+  - Interval Types
+  - Fixed Point Types
+  - Injecting Aspects
+  - EnumAnnotations
```
ekiwi, line 44: Would you consider keeping these around until you add enums to the firrtl IR?
Essentially in the same PR where you switch chisel to emitting the native firrtl enums would be the appropriate place for removing the old annotation based system. This way we do not loose any information in the meantime.

ekiwi:
> I think what I really miss is the formal framework by @ekiwi

I talked to @jackkoenig and @seldridge and what is probably going to happen is that I will copy the sources to the `firrtl` compiler into the `chiseltest` repo to try and keep `chiseltest` going for a little while.

ekiwi:
Thanks for clarifying the chisel roadmap. Could you also elaborate on which associated project you are going to port to `firtool` and which you are going to discontinue? I am thinking about:

- FireSim
- treadle
- the chisel bootcamp
- chiseltest (won't be ported to `firtool`, already talked to @jackkoenig and @seldridge about this)
- dsp-tools

ROADMAP.md
seldridge: Chisel has a high, initial level of complication arising from generator semantics. Injecting Aspects exacerbate this with a further level of complexity by enabling arbitrary FIRRTL to be appended after the Chisel generator runs in an order not captured in the Chisel.

This has been used as a workaround for long-missing Chisel/FIRRTL features, e.g., separation of design/verif code which is better served by an extract test code pass (which is available in CIRCT). Now that we are supporting, or have infrastructure to support, long missing features and the evolution of FIRRTL and other IRs to support user needs, we can remove the reliance on a very bad programming paradigm.

Put differently, are injecting aspects something we want? Are they a good design point we want to support in Chisel or were they patching an underlying problem in Chisel/SFC arising from missing features? If they're not the former or are the latter, then we should get rid of them. I'm soundly in the camp that they were a bad experiment and there time has come. 😅 

That said, this list is not final. If anyone is willing to put in the work to get these features into CIRCT and keep them maintained, I don't think anybody is against that!

ekiwi: > Put differently, are injecting aspects something we want? Are they a good design point we want to support in Chisel or were they patching an underlying problem in Chisel/SFC arising from missing features? If they're not the former or are the latter, then we should get rid of them. I'm soundly in the camp that they were a bad experiment and there time has come.

If you can already generate assertions outside of the main chisel modules already, then maybe you can show @sequencer how to do that without aspects. If you can cover his use case, which is also what I am interested in, then we might not need injecting aspects.

ROADMAP.md
mwachs5: Chisel 5.0.0 is not expected to use scala 3. But the point is we may quickly end up at a Chisel 52.0.0 that does use scala3 (though not on the roadmap yet)

schoeberl: I am not a big fan of version number inflation like Java did in recent years.

schoeberl: Any chance that we merge with SpinlHDL?

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+  - RunFirrtlTransformAnnotation
+  - LoadMemoryFromFile
+  - Trace API
+  - ChiselAnnotation/custom annotations
```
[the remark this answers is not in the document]
ekiwi: if there is no support for the decoder API, does that mean that rocket-chip won't be able to update?

seldridge:
> chiseltest (won't be ported to firtool, already talked to @jackkoenig and @seldridge about this)

We never said this. We've been running ChiselTest and plan to support it's core testing features with standard backends.

What we did say is that we don't see a way forward for Treadle and the formal work, i.e., technologies which rely heavily on the internals of the SFC or on custom SFC backends.

This hits the formal backend hard, unfortunately. It goes without saying that there is a ton of interest in getting tighter integration between CIRCT and formal flows. There was a Google Summer of Code project working on formal equivalence of CIRCT dialects. Similarly, we've seen a number of things related to scheduling dialects that can be formulated as SMT problems, which starts to motivate an SMT dialect in CIRCT.

Treadle will eventually get replaced with native CIRCT simulation support. We have internal experiments/demos doing this on non-trivial designs. We're jumping through open-source release hoops for this now.

How I would view these changes is the following: a number of Chisel/SFC features are extremely closely coupled to the SFC implementation. As it stands, this only benefits a very narrow set of Chisel/SFC users. There's wide demand for this stuff in CIRCT. Taking a step back, learning lessons from some of these SFC efforts, and seeing how we can enable this for core CIRCT dialects (e.g., how do we simulate any HW Dialect circuit or how do we extend simulation to support Verilog?) means huge wins for the broader hardware community going forward.

ROADMAP.md
mwachs5: Ah, so are you suggesting I remove it from the list?

ekiwi:
> We never said this. We've been running ChiselTest and plan to support it's core testing features with standard backends.

There will be a `chiseltest` like library, but about 80% of current `chiseltest` features won't work with `firtool` and thus making a new library for `firtool` is in my opinion preferable to removing 80% of the `chiseltest` functionality.

ROADMAP.md
seldridge: > Are these things enumerated somewhere?

Not enumerated, but will be obvious once we go through the exercise of deprecating/removing things. What is expected to stay is anything related to infrastructure, e.g., options parsing, stage/phase infrastructure (even though I'd like to replace this with something simpler), annotation infra.

> Is Chisel 5 then Scala 3? Can we get the bit width as part of the Scala type and err on the assignment of different widths?

No. Though a goal that this migration will help support. Bit width support, not yet. However, there are two interesting avenues for this: (1) moving bit width into the type as we've discussed and (2) relying on CIRCT for doing the checking and providing reports back to the user. CIRCT will do some magical stuff where it reads Scala source to show error messages (as long as it can find the files, which is what https://github.com/chipsalliance/chisel3/pull/2791 is enabling).

Regardless, a predominant goal of this roadmap is to enable faster development of new features and changes to existing Chisel features that have long been difficult/impossible to implement. What you're asking for is more possible in a Chisel 5 world.

ROADMAP.md
```diff
@@ -0,0 +1,89 @@
...
+- Publish Roadmap
+
+
+## 5.0.0 (Target Q1 2023)
+
```
edwardcwang, line 53:
```suggestion
## 4.0

See the Q&A below.

## 5.0.0 (Target Q1 2023)

```

ROADMAP.md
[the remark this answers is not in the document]
sequencer: The decoder is resolved in the front end, they are pure chisel api, which has been masked in the mfc. 
the reason I added them is: if users are sticking on qmc decoder, they can use this api to cache the decoder results to speed up elaboration among different runs.

ROADMAP.md
sequencer: Agree, skipping a large version is strange, I think 4.0 with a clear release note would be better. 

py2 -> py3 takes 10+ years, without directly give py4 to community. I don’t think the migration from SFC to MFC is bigger problem like this.

sequencer:
I think chiseltest is not a big issue, after trace api + MLIR querying framework is given, test frameworks can be more powerful than current chiseltest version.

ROADMAP.md
ekiwi: > relying on CIRCT for doing the checking and providing reports back to the user. CIRCT will do some magical stuff where it reads Scala source to show error messages

Is there a companion PR for CIRCT that I could look at? This sound like a pretty cool feature.

ROADMAP.md
seldridge: Yes! This shouldn't be on the list because we're supporting it.

ROADMAP.md
```diff
@@ -0,0 +1,93 @@
...
+  deprecated features
+- Deprecate all Chisel features that are not implemented in MFC. A nonexhaustive
+  list includes:
+  - Interval Types
+  - Fixed Point Types
+  - Injecting Aspects
+  - EnumAnnotations
+  - RunFirrtlTransformAnnotation
+  - LoadMemoryFromFile
+  - Trace API
```
mwachs5, line 47:
```suggestion
```

ROADMAP.md
seldridge: > Is there a companion PR for CIRCT that I could look at? This sound like a pretty cool feature.

No PR. This surprisingly just works:

```scala
//> using scala 2.13
//> using lib "edu.berkeley.cs::chisel3::3.5.4"
//> using lib "com.sifive::chisel-circt::0.6.0"

import chisel3._

class Foo extends RawModule {
  /** This wire is uninitialized! */
  val w = Wire(Bool())
}

object Main extends App {
  println(ChiselStage.emitSystemVerilog(new Foo))
}
```

`scala-cli Foo.scala` produces:

```
Running CIRCT: 'firtool -format=fir -warn-on-unprocessed-annotations -verify-each=false -disable-infer-rw -dedup -annotation-file Foo.anno.json < $input'
Error: firtool failed.
ExitCode:
1
STDOUT:

STDERR:
Foo.scala:10:15: error: sink "w" not fully initialized
  val w = Wire(Bool())
              ^
Foo.scala:10:15: note: see current operation: %0 = "firrtl.wire"() {annotations = [], name = "w", nameKind = #firrtl<name_kind droppable_name>} : () -> !firrtl.uint<1>

```

ekiwi: good stuff, now this just needs a IntelliJ plugin

ROADMAP.md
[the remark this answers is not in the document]
seldridge: The plan, and current implementation of MFC, is to reject any annotations that it does not explicitly understand. (This may be disabled by a user, but almost always without the correct effect on their design.) A `chisel-circt` based flow can output a Chisel `.anno.json` which includes things like decoder results. This doesn't have to be ingested/understood by a FIRRTL compiler.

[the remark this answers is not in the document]
mwachs5: i think the question about EnumAnnotations  would match this as well

ROADMAP.md
ekiwi: I have a FSM coverage pass that relies on the enum annotations to discover state machine registers and possible state transitions.

ROADMAP.md
seldridge: We would probably want a replacement for this API. That would seem to require an evolution of the FIRRTL spec to talk about memory initialization.

ROADMAP.md
seldridge: I think it is reasonable to allow Chisel to emit whatever annotations it wants. However, an `EnumAnnotation` is not updated or understood by a FIRRTL compiler. Put differently, this is a move towards FIRRTL compilers having an allow-list of known annotations and `EnumAnnotation` would not be allowed or, more weakly, any `EnumAnnotation` is not output from a FIRRTL compiler. If a user wants this style of behavior, they should use `chisel3.experimental.Trace`.

ekiwi: So firrtl compilers won't be required by the spec to support user defined annotations, but if a specific firrtl compiler choose to do so, it would still be able to get the annotation out of Chisel, correct? That would sound reasonable to me.

ROADMAP.md
seldridge: This is the dream. 🥺

ekiwi: (the cheap version is to just format the file path correctly, then IntelliJ should at least turn it into a hyperlink that will let you jump to the file and line)

ROADMAP.md
seldridge: I started the prerequisite work to support this in https://github.com/llvm/circt/pull/4212. That adds modeling of `$readmemb` and `$readmemh` to the SystemVerilog Dialect. It is then just necessary to: (1) update FIRRTL Dialect to include information about a memory loading file to memory ops and (2) lower these ops to use the new SystemVerilog dialect operations for `$readmemb` / `$readmemh`.

ROADMAP.md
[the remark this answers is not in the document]
seldridge: Clarifying: `ChiselAnnotation` will be made a `private[chsiel3]` API and is thereby deprecated for Chisel users. Any custom annotation, i.e., anything not in-tree with Chisel3, will not be supported. That said, we're very, very open to improvements to Chisel, the FIRRTL spec, and the MFC to meet user needs, e.g., enumerated types in FIRRTL to replace enum annotations, as well as continuing to seamlessly support Chisel APIs that are currently emitted using annotations, e.g., trace, inline/flatten, and dontTouch (_these examples are non-exhaustive_). We'll address FireSim usages in the near future.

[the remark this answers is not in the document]
seldridge: Simply put, we are totally missing a good pass injection flow for the MFC. That needs to be added. You should be able to, LLVM-style, inject a custom pass built as a shared library into MFC and have it work.

The main question is then what to do with custom annotations. Thinking about this a bit more... what we could support is a true "custom annotation" which requires a user to opt-in and only supports a `SingleTargetAnnotation` format (all `MultiTargetAnnotations` can be converted to a formulation that uses multiple `SingleTargetAnnotations`). There are already kind-of hooks for this in CIRCT that exist for testing only with fake annotations:

1) `circt.test` is a `SingleTargetAnnotation` that can be non-local
2) `circt.testLocalOnly` is a `SingleTargetAnnotation` that _must_ be local (effectively point at a `Named`)
3) `circt.testNT` is a `NoTargetAnnotation`

These _could_ be enshrined as a special annotation format called `CustomAnnotation.TargetLocal` and `CustomAnnotation.TargetNonLocal` (where a `NoTargetAnnotation` has to be an annotation that targets the circuit). This then enables arbitrary custom annotations to be shoved into CIRCT. We currently have an optional warning that annotations must be removed by the time `LowerToHW` (conversion of FIRRTL Dialect to HW Dialects happens). If we enable custom annotations, then this should be an error by default and demotable to a warning.

An example of the above (using existing infra) is:

```
circuit Foo: %[[
  {
    "class": "circt.test",
    "target": "~Foo|Foo>w",
    "payload": {
      "class": "gcd.tools.ResilientRegisterAnnotation"
    }
  }
]]
  module Foo:
    input a: UInt<1>
    output b: UInt<1>

    wire w: UInt<1>
    w <= a
    b <= w
```

Producing:

```mlir
module {
  firrtl.circuit "Foo"  {
    firrtl.module @Foo(in %a: !firrtl.uint<1>, out %b: !firrtl.uint<1>) {
      %w = firrtl.wire interesting_name  {annotations = [{class = "circt.test", payload = {class = "gcd.tools.ResilientRegisterAnnotation"}}]} : !firrtl.uint<1>
      firrtl.strictconnect %w, %a : !firrtl.uint<1>
      firrtl.strictconnect %b, %w : !firrtl.uint<1>
    }
  }
}
```

We then just need at least two hooks for custom passes: 

1) You need a pass right after `LowerAnnotations` that can resolve the annotations to something else.
2) You need a pass somewhere else in the FIRRTL pipeline to remove the annotations.

My main concern about this approach is that mid-term, almost all annotations (which are not representable using the `TraceAPI`) are really custom operations and _not_ existing operations with discarable metadata. There is nothing in CIRCT that can know what is the behavior of a custom annotation is nor should it know. Put differently, a TMR Register is better modeled as `firrtl.reg.tmr` as opposed to `firrtl.reg annotation {"tmr"}`. It would be better, mid-term, if we worked towards enabling Chisel to construct MLIR directly which could enable novel, library-linked FIRRTL or other dialect operations to be constructed. This seems possible, but I don't have line-of-sight on how to do it.

For your second point of scan chain connection. This is really something that should be handled on the Chisel side. @jackkoenig has some experiments with providing a resource, having resources add information about what they need to do, and then having the resource make modifications to the module. More concretely, a scan chain could work like:

```scala
class ScanChainResource {
  def addToScanChain[A <: Data](a: Data): ???

  def buildScanChain(): Unit
}

class Bar(scanChain: ScanChainResource) extends Module {
  val r = Reg(Bool())
  scanChain.addToScanChain(r)
}

class Foo extends Module {
  val scanChain = new ScanChainResource

  val bar1, bar2 = Module(new Bar(scanChain))

  scanChain.buildScanChain()
}
```

Or: what Albert was doing with the scan chain stuff (and [what I've wrote passes for in the past](https://github.com/IBM/chiffre/blob/master/src/main/scala/chiffre/passes/ScanChainTransform.scala) 😬 ) is really a missing first-class Chisel ability. We didn't have this, so we always fell back to using the FIRRTL compiler in somewhat inappropriate ways.

WDYT?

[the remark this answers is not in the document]
tymcauley: This sounds like a great path forward. FWIW the TMR and scan-chain passes were just examples, but I'm interested in supporting other transforms that we use internally, and I'm sure other companies have plenty of transforms that they'd like to trigger with custom annotations.

Regarding the annotations that we pass on from Chisel to the MFC, it seems like two pieces of information are generally important:

- What is the target of the annotation (some circuit element/module, either local or non-local)
- What is the "type" of the annotation

It looks like from your example, that "payload.class" field gives that second piece of information. I haven't seen that sort of pattern in Chisel/CIRCT yet, but it looks like a great way to get this information across.

And agreed, in the not-immediate term I really like the idea of extending the FIRRTL MLIR dialect in some library-linked manner, but I have no idea how that sort of thing would work out-of-tree from the CIRCT repo.

PR #2817, opened 2022-10-31
Deprecate Fixed Point and Intervals

seldridge:
Deprecate Fixed Point types and Interval types.

This is the first of several PRs that are implementing the [Chisel Roadmap](https://github.com/chipsalliance/chisel3/blob/master/ROADMAP.md) plan to migrate to the MLIR-based FIRRTL Compiler. For more information, please see the referenced roadmap.

### DIFF
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -358,4 +358,7 @@ package object chisel3 {
   // Connection exceptions.
   case class BiConnectException(message: String) extends ChiselException(message)
   case class MonoConnectException(message: String) extends ChiselException(message)
+
+  final val deprecatedMFCMessage =
+    "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
 }
--- a/core/src/main/scala/chisel3/Bits.scala
+++ b/core/src/main/scala/chisel3/Bits.scala
@@ -1852,11 +1852,13 @@ package experimental {
     *
     * @param range       a range specifies min, max and binary point
     */
+  @deprecated(deprecatedMFCMessage, "Chisel 3.6")
   sealed class Interval private[chisel3] (val range: chisel3.internal.firrtl.IntervalRange)
       extends Bits(range.getWidth)
       with Num[Interval]
       with HasBinaryPoint {
 
+    @deprecated(deprecatedMFCMessage, "Chisel 3.6")
     override def toString: String = {
       litOption match {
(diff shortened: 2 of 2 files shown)


core/src/main/scala/chisel3/package.scala
```diff
@@ -385,4 +385,7 @@ package object chisel3 {
   // Connection exceptions.
   case class BiConnectException(message: String) extends ChiselException(message)
   case class MonoConnectException(message: String) extends ChiselException(message)
+
+  final val deprecatedMFCMessage =
+    "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
```
mwachs5, line 390:
```suggestion
    "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
```

nit: why aren't we capitalizing `this`

mwachs5 (approved):

core/src/main/scala/chisel3/package.scala
```diff
@@ -385,4 +385,7 @@ package object chisel3 {
   // Connection exceptions.
   case class BiConnectException(message: String) extends ChiselException(message)
   case class MonoConnectException(message: String) extends ChiselException(message)
+
+  final val deprecatedMFCMessage =
```
jackkoenig, line 389: Can/Should this be package private?

seldridge: I was trying to hide this, but `@deprecated` seemed to elide the message if it was package private. I'll check.

core/src/main/scala/chisel3/package.scala
seldridge: Well, I can actually see this going either way. The default sentence isn't capitalized in the error message. However, this is two complete sentences. I will update it. Thanks!

```
[warn] /Users/schuylere/repos/github.com/chipsalliance/chisel3/src/test/scala/chiselTests/AnnotatingDiamondSpec.scala:34:5: method apply in object annotate is deprecated (since Chisel 3.6): this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md.
[warn]     annotate(anno)
[warn]     ^
```

LLVM/MLIR warnings often include a partial sentence in the message, e.g., `foo->emitError() << "is badf"` gets emitted as `firrtl.circuit op 'HerpDerp' is bad`. I think I got used to this and wrote the same thing here. 😕

PR #2819, opened 2022-10-31
Deprecate ChiselAnnotation

seldridge:
Deprecate ChiselAnnotation as custom FIRRTL annotations are not supported as part of the migration to the MILR-based FIRRTL Compiler.


mwachs5:
I see people using ChiselAnnotation as a quick-and-dirty at-end-of-elaboration callback since you can do arbitrary things in the `toFirrtl` conversion. Should we provide a better API for that?

PR #2825, opened 2022-11-04
Add DataMirror.getParent for getting parents of Modules

jackkoenig:
### Contributor Checklist

#### Type of Improvement

  - new feature/API 

#### API Impact

Adds a new experimental API, useful for generators and metadata.

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Add DataMirror.getParent for getting parents of Modules

- [ ] Did you review?


src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
```diff
@@ -55,4 +55,21 @@ class DataMirrorSpec extends ChiselFlatSpec {
     }
     ChiselStage.elaborate(new MyModule)
   }
+
+  it should "support getParent for normal modules" in {
+    class GrandChild(parent: RawModule) extends Module {
+      DataMirror.getParent(this) should be(Some(parent))
```
mwachs5, line 61: oh nice, you can call this even before elaboration completes? that was more than i was hoping for...

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
```diff
@@ -55,4 +55,21 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
+      DataMirror.getParent(inst) should be(Some(this))
+      DataMirror.getParent(this) should be(Some(parent))
+    }
+    class Parent extends Module {
+      val inst = Module(new Child(this))
+      DataMirror.getParent(inst) should be(Some(this))
+      DataMirror.getParent(this) should be(None)
+    }
+    ChiselStage.elaborate(new Parent)
+  }
```
mwachs5, line 74: can we show a D/I test. What happens if you call this from within a Definition(...) (sjhould be None I expect) and `getParent(Instance(...))` should be `this`?

seldridge (commented):
I'm slightly nervous about this API and how this will compose with separable compilation. E.g., simply considering D/I a module may have zero or more parents and this number may change during elaboration.

That said, this API makes total sense for things within a separable compilation unit using a normal depth-first elaboration flow (I either have one parent or I don't).

jackkoenig:
> I'm slightly nervous about this API and how this will compose with separable compilation. E.g., simply considering D/I a module may have zero or more parents and this number may change during elaboration.
>
> That said, this API makes total sense for things within a separable compilation unit using a normal depth-first elaboration flow (I either have one parent or I don't).

I share your concerns about separable compilation. I will note that this API does not support D/I at all. It can be used within a D/I context, but has the same behavior as outside of one because the use of `Definition(...)` creates a separate Chisel elaboration context. As requested by @mwachs5, I will show this with a test, but the point is that you cannot call this API on an `Instance[_]` nor a `Definition[_]`.

What a similar API should look like for `Instance[_]` and `Definition[_]` objects I have left for future debate because you're right that a `Definition` at least can have zero or more parents. An related observation is that the top-level module can also have zero or more parents since you don't know in what context it will be instantiated.

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
jackkoenig: So this does not support `Definition[_]` nor `Instance[_]`, as I commented below, I'm sort of punting on that possible API extension. I have added a test though.

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
```diff
@@ -55,4 +74,18 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
+
+  it should "support getParent for normal modules" in {
+    ChiselStage.elaborate(new Parent)
+  }
+
+  it should "support getParent for normal modules even when used in a D/I context" in {
+    import chisel3.experimental.hierarchy._
+    class Top extends Module {
+      val defn = Definition(new Parent)
+      val inst = Instance(defn)
```
mwachs5, line 86: what happens in `DataMirror.getParent(inst) should be(this)`?

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
mwachs5: i dont quite get the test, will it just not compile if you try to call it on Definition or Instance?

It does show that it works as expected from within the Definition code though so that's good.

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
```diff
@@ -55,4 +74,18 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
   }
+
+  it should "support getParent for normal modules" in {
+    ChiselStage.elaborate(new Parent)
+  }
+
+  it should "support getParent for normal modules even when used in a D/I context" in {
+    import chisel3.experimental.hierarchy._
+    class Top extends Module {
+      val defn = Definition(new Parent)
```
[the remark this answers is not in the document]
jackkoenig: `getParent` accepts `BaseModule`, so you cannot pass a `Definition[_]` nor an `Instance[_]`.

src/test/scala/chiselTests/experimental/DataMirrorSpec.scala
jackkoenig: Compiler error 😇, see https://github.com/chipsalliance/chisel3/pull/2825#discussion_r1014729172.

mwachs5 (approved):

PR #2839, opened 2022-11-10
Change RawModule._commands to a VectorBuilder

jackkoenig:
Use the resulting Vector to build the underlying Component's commands and then use those instead of copying the original ArrayBuffer when iterating on commands. Previously, the Component was using a List to hold the commands which is particularly memory inefficient, especially for large modules.

### Contributor Checklist

#### Type of Improvement

  - performance improvement 

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Optimize Module internals

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -2,7 +2,6 @@
 
 package chisel3
 
-import scala.collection.mutable.{ArrayBuffer, HashMap}
 import scala.util.Try
 import scala.language.experimental.macros
 import scala.annotation.nowarn
@@ -13,6 +12,7 @@ import chisel3.internal.Builder._
 import chisel3.internal.firrtl._
 import chisel3.internal.sourceinfo.UnlocatableSourceInfo
 import _root_.firrtl.annotations.{IsModule, ModuleTarget}
+import scala.collection.immutable.VectorBuilder
 
 /** Abstract base class for Modules that contain Chisel RTL.
   * This abstract base class is a user-defined module which does not include implicit clock and reset and supports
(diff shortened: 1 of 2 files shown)


jackkoenig:
The commit just changing `_commands` to a `VectorBuilder` appears to reduce memory for a large design by 5%, but has no impact on performance. I'm betting this is because we call `.toList` on the commands before conversion which, without that commit is a no-op but with the commit does the same copy the commit avoids. Using `VectorBuilder` still reduces memory use since that List can be garbage collected immediately after conversion, but it keeps the total amount of compute basically the same.

I'm trying to change the converter to not use the List, but it definitely assumes List in how it traverses the commands so that code needs to change to see any performance benefit.

edwardcwang:
?
```
getVerilogString(new DetectTwoOnes)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
java.lang.StackOverflowError
	at scala.collection.Iterator$ConcatIterator.hasNext(Iterator.scala:224)
	at scala.collection.Iterator$$anon$22.hasNext(Iterator.scala:1089)
	at scala.collection.Iterator$ConcatIterator.hasNext(Iterator.scala:224)
	at scala.collection.Iterator$$anon$22.hasNext(Iterator.scala:1089)
...
```

jackkoenig:
This is not yet functionally correct

jackkoenig:
Okay this works now and it seems to be _slightly_ faster than before. I also tried simpler things of clearing the `ArrayBuffer` after turning it into a `List` (that `.toSeq` in `getCommands` makes a `List`), and I tried using a `ListBuffer`. This PR uses less memory and is marginally faster than those approaches (~1-2% for a typical large design).

edwardcwang (approved):

PR #2856, opened 2022-11-20
Implement compressed Namespace

jackkoenig:
The namespace disambiguates requests for the same name with _<idx>. Rather than storing every disambiguated name in the underlying HashMap, it now only stores the base along with the "next available" index. This makes the logic for checking if a name is already contained in the namespace slightly more sophisticated because users can name things in a way that will collide with disambiguated names from a common substring.

For example, in naming the sequence "foo", "foo", "foo_1", the 2nd "foo" takes the name "foo_1" so the following "foo_1" gets disambiguated to "foo_1_1". But since we compressed that original "foo_1" into the same HashMap entry as just "foo", we have to do a form of "prefix checking" whenever naming something that ends in `_<idx>`.

In practice, the saved memory allocations more than make up for the more complicated logic to disambiguate names because the common case is still fast.

To benchmark this, I took a large internal design a picked a module that had a large Namespace. I then dumped every requested name to a file for that module and used that set of 243126 names to measure how much memory the version on master uses vs. this version.

I found that this change reduces the amount of time to name those 243126 names from 66ms to 34ms (about 2x) and reduces the memory used from 80 MiB to a mere 1.2 MiB (a staggering **66x** reduction). Of course these results depend on how many name collisions there are, and the fewer there are the smaller the savings, but I believe this module to be fairly representative.

Namespace operations are not a huge part of overall Chisel elaboration, so it's a bit hard to measure the benefit for full elaboration, but I suspect it reduces memory use and pressure by a few percent, maybe 3-5%. I'm trying to get some good measurements here.

### Contributor Checklist

#### Type of Improvement

 - performance improvement  

#### API Impact

No impact

~This does perturb naming of a handful of signals (`~200` in a design that has millions). I am not too worried but can spend more time understanding the issue if anyone finds this concerning.~

Nevermind I figured it out. It was the case of `_0` being a "false collision" because the namespace starts disambiguating at `_1`. There's extra logic now to preserve the old behavior for `_0` that just results in 1 additional boolean check in an uncommon (but hit at least 200 times!) code path.

This has no impact on the generated FIRRTL/Verilog.

#### Desired Merge Strategy

  - Squash

#### Release Notes

Reduce the memory use of internal Namespace datastructure.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -19,18 +19,29 @@ import logger.LazyLogging
 
 import scala.collection.mutable
 import chisel3.internal.sourceinfo.UnlocatableSourceInfo
+import scala.annotation.tailrec
 
 private[chisel3] class Namespace(keywords: Set[String]) {
+  // This HashMap is compressed, not every name in the namespace is present here.
+  // If the same name is requested multiple times, it only takes 1 entry in the HashMap and the
+  // value is incremented for each time the name is requested.
+  // Names can be requested that collide with compressed sets of names, thus the algorithm for
+  // checking if a name is present in the Namespace is more complex than just checking the HashMap,
+  // see getIndex below.
   private val names = collection.mutable.HashMap[String, Long]()
   def copyTo(other: Namespace): Unit = names.foreach { case (s: String, l: Long) => other.names(s) = l }
   for (keyword <- keywords)
     names(keyword) = 1
 
-  private def rename(n: String): String = {
-    val index = names(n)
+  @tailrec
+  private def rename(n: String, index: Long): String = {
     val tryName = s"${n}_${index}"
-    names(n) = index + 1
-    if (this contains tryName) rename(n) else tryName
+    if (names.contains(tryName)) {
+      rename(n, index + 1)
+    } else {
+      names(n) = index + 1
+      tryName
+    }
   }
 
   private def sanitize(s: String, leadingDigitOk: Boolean = false): String = {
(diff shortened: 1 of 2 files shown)


aswaterman:
Nice space reduction!

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -42,14 +53,55 @@ private[chisel3] class Namespace(keywords: Set[String]) {
...
+      case Some(idx) => rename(sanitized, idx)
+      case None =>
+        names(sanitized) = 1
+        sanitized
+    }
+  }
+
+  def print(): Unit = {
+    names.foreach(println)
   }
```
jackkoenig, line 105:
```suggestion
```

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -42,14 +53,55 @@ private[chisel3] class Namespace(keywords: Set[String]) {
...
+      else {
+        // If we get a prefix collision and our index is taken, we start disambiguating with _<idx>_1
+        names
+          .get(elem.take(maybePrefix))
+          .filter { prefixIdx =>
+            val ourIdx = elem.drop(maybePrefix + 1).toInt
+            // The namespace starts disambiguating at _1 so _0 is a false collision case
+            ourIdx != 0 && prefixIdx > ourIdx
+          }
+          .map(_ => 1) // If we have a prefix match and our index is take
```
jackkoenig, line 88:
```suggestion
          .map(_ => 1)
```
I ended up not finishing writing the commend here and instead putting it above

jackkoenig:
I finally got some measurements and this decreases peak memory use for a fairly large design by about ~3%. Not huge, but still nice.

azidar (approved):

PR #2857, opened 2022-11-21
SerializableModuleGenerator

sequencer:
- add upickle to dependency
- implement SerializableModuleGenerator to serialize a parameterized module
- add simple test

### Contributor Checklist

- [ ] Did you review?


mwachs5:
is there an overall "rationale doc" or RFC for the approach you are pursuing here for serialization / upickling / auto-blackbox?

sequencer:
doc added, 
however in https://github.com/chipsalliance/chisel3/actions/runs/3587493060/jobs/6037901924
it seems that `mdoc` make all class inner class, and `SerializableModuleGenerator` forbid this.

sequencer:
> is there an overall "rationale doc" or RFC for the approach you are pursuing here for serialization / upickling / auto-blackbox?

Serialization documentation is added in this PR.

docs/src/cookbooks/serialization.md
```diff
@@ -0,0 +1,177 @@
...
+```
+using `write` function in `upickle`:
+```scala
+val j = upickle.default.write(
+  SerializableModuleGenerator(
+    classOf[GCDSerializableModule],
+    GCDSerializableModuleParameter(32)
+  )
+)
+println(j)
```
jackkoenig, line 81:
````suggestion
```scala mdoc:silent
val j = upickle.default.write(
  SerializableModuleGenerator(
    classOf[GCDSerializableModule],
    GCDSerializableModuleParameter(32)
  )
)
````

Is the constraint on no inner modules only at runtime or also at compile time? If you can still compile it you could do `mdoc:silent` (also please remove the println, especially becuase it doesn't execute).

File: docs/src/cookbooks/serialization.md [hunk 2]
```diff
@@ -0,0 +1,177 @@
...
+  z = _RAND_2[31:0];
+`endif
+  `endif
+end
+`ifdef FIRRTL_AFTER_INITIAL
+`FIRRTL_AFTER_INITIAL
+`endif
+`endif
+endmodule
+```
```
jackkoenig, line 177:
```suggestion
You can then read from json string and elaborate the Module:
```scala mdoc:silent
chisel3.stage.ChiselStage.emitVerilog(
  upickle.default.read[SerializableModuleGenerator[GCDSerializableModule, GCDSerializableModuleParameter]](
    ujson.read(j)
  ).module()
)
```
````
Assuming it will still compile (and just can't be run), you can use `mdoc:silent` here. Also I don't see value in showing the emitted Verilog, especially since it will get stale and we'll forget to update it. I think it's fine to just show the API.

**Review by jackkoenig — APPROVED** [review]
I think this is fine, although I have suggested some tweaks to the mdoc.

File: docs/src/cookbooks/serialization.md (hunk 1 above)
Comment (sequencer) [reply docs/src/cookbooks/serialization.md:81]: It's runtime. I just took a look at the doc of mdoc, thanks Jack.

File: docs/src/cookbooks/serialization.md (hunk 2 above)
Comment (sequencer) [reply docs/src/cookbooks/serialization.md]: I change `mdoc:silent` into `mdoc:compile-only`, see: https://scalameta.org/mdoc/docs/modifiers.html#compile-only

File: docs/src/cookbooks/serialization.md (hunk 1 above)
Comment (sequencer) [reply docs/src/cookbooks/serialization.md]: I restore the silent in json dump command, since it's a one-line result.

**Comment by sequencer** [comment]
Recently I encounter a type issue of this API:
since `Seq[SerializableModuleGenerator]` doesn't consume covariant type,
I cannot serialize `case class TopParameter(modules: Seq[SerializableModuleGenerator[_, _]])`,
since scala cannot prove this constraint:
```
SerializableModuleGenerator[M <: SerializableModule[P], P <: SerializableModuleParameter]
```

PR #2866, opened 2022-12-03
Optimize _ids

jackkoenig:
Only add Data to _ids when the binding is constrained. ConstrainedBinding means that the bound Data is scoped to a module. Since _ids are used for naming of signals within a module, there is no reason to add Unconstrained Data to the _ids of the Module they were constructed during. Unconstrained Data include things like literals and children of Aggregates (which sort of are constrained to a Module but only via their parent, they are not directly named by the module).

Note that this cannot be backported because it breaks the naming of unnamed things which is deprecated in 3.5 (but still works).

I measured this for a balanced, large design and it reduced minimum memory use from ~8000 MiB to ~7500 MiB (~6.25%). It did not have a measurable impact on wall-clock performance, but does reduce total user time which does reduce overall system compute. I don't currently measure user time very scientifically, but it seems to reduce user time by ~6% which would make sense. So definitely a win.
 
### Contributor Checklist

#### Type of Improvement

   - performance improvement

#### API Impact

There is a slight order change to the ordering of the `Data` in `DataProduct[BaseModule]`. This is due to the fact that `DataProduct[Data]` iterates on the elements of `Aggregates` in a slightly different order than the order those elements were added to the `_ids` of the parent `Module`. I think this is a fine tradeoff but it may be worth carefully considering the orders from `DataProduct` in case users may depend on them in the future.

#### Desired Merge Strategy

  - Squash

#### Release Notes

Reduce memory use of some internal data structures. Reduces memory use by ~6%. Slight change to ordering of `Data` in `DataProduct[BaseModule]`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -184,7 +184,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
   }
 
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
-    _parent.foreach(_.addId(this))
+    this.maybeAddToParentIds(target)
     binding = target
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
@@ -987,7 +987,7 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
   }
 
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
-    _parent.foreach(_.addId(this))
+    this.maybeAddToParentIds(target)
     binding = target
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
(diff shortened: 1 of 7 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -184,7 +184,10 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
   }
 
   private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
-    _parent.foreach(_.addId(this))
+    // ConstrainedBinding means this is scoped to a Module, no need to add to _ids otherwise
+    if (target.isInstanceOf[ConstrainedBinding]) {
+      _parent.foreach(_.addId(this))
+    }
```
seldridge, line 190: This snippet is repeated multiple times. Is there a factoring that avoids the repetition?

seldridge (approved):
LGTM

Nice savings in user time.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: Good point, I'll factor it out.

PR #2891, opened 2022-12-14
Enrich ShiftRegister with SyncReadMem-based implementation.

milovanovic:
### Contributor Checklist

#### Type of Improvement
* backend code generation
* new feature/API
* performance improvement

#### API Impact

This feature extends the current API preserving full backward compatibility.

If SyncReadMem-based Shift Register implementation is selected this will impact the generated Verilog code.

#### Desired Merge Strategy

No preference.

#### Release Notes

Supplement ShiftRegister with SyncReadMem-based implementation.

- [ ] Did you review?


macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala
```diff
@@ -251,6 +251,10 @@ class SourceInfoTransform(val c: Context) extends AutoSourceTransform {
   def inNResetDataEnArg(in: c.Tree, n: c.Tree, resetData: c.Tree, en: c.Tree): c.Tree = {
     q"$thisObj.$doFuncTerm($in, $n, $resetData, $en)($implicitSourceInfo, $implicitCompileOptions)"
   }
+
+  def inNEnUseSpNameArg(in: c.Tree, n: c.Tree, en: c.Tree, use_sp_mem: c.Tree, name: c.Tree): c.Tree = {
+    q"$thisObj.$doFuncTerm($in, $n, $en, $use_sp_mem, $name)($implicitSourceInfo, $implicitCompileOptions)"
+  }
```
sequencer, line 257:
```suggestion

  def inNEnUseSpNameArg(in: c.Tree, n: c.Tree, en: c.Tree, useSinglePortMem: c.Tree, name: c.Tree): c.Tree = {
    q"$thisObj.$doFuncTerm($in, $n, $en, $useSinglePortMem, $name)($implicitSourceInfo, $implicitCompileOptions)"
  }
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -123,6 +123,91 @@ object ShiftRegister {
...
+
+  /** Returns the n-cycle delayed version of the input signal (SyncReadMem-based ShiftRegister implementation).
+    *
+    * @param in input to delay
+    * @param n  number of cycles to delay
+    * @param en enable the shift
+    * @param use_sp_mem single port or dual port SRAM based implementation
+    * @param name name of SyncReadMem object
+    */
+  def apply[T <: Data](in: T, n: Int, en: Bool, use_sp_mem: Boolean, name: String): T =
```
sequencer, line 135:
```suggestion
  def apply[T <: Data](in: T, n: Int, en: Bool, useSinglePortMem: Boolean, name: String): T =
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -123,6 +123,91 @@ object ShiftRegister {
     compileOptions:      CompileOptions
   ): T =
     ShiftRegisters(in, n, resetData, en).lastOption.getOrElse(in)
+
+  /** Returns the n-cycle delayed version of the input signal (SyncReadMem-based ShiftRegister implementation).
+    *
+    * @param in input to delay
+    * @param n  number of cycles to delay
+    * @param en enable the shift
+    * @param use_sp_mem single port or dual port SRAM based implementation
```
sequencer, line 132:
```suggestion
    * @param useSinglePortMem single port or dual port SRAM based implementation
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -123,6 +123,91 @@ object ShiftRegister {
...
+    */
+  def apply[T <: Data](in: T, n: Int, en: Bool, use_sp_mem: Boolean, name: String): T =
+    macro SourceInfoTransform.inNEnUseSpNameArg
+
+  /** @group SourceInfoTransformMacro */
+  def do_apply[T <: Data](
+    in:         T,
+    n:          Int,
+    en:         Bool,
+    use_sp_mem: Boolean,
```
sequencer, line 143:
```suggestion
    useSinglePortMem: Boolean,
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -123,6 +123,91 @@ object ShiftRegister {
...
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): T = _apply_impl_sram(in, n, en, use_sp_mem, name)
+
+  private def _apply_impl_sram[T <: Data](
+    in:         T,
+    n:          Int,
+    en:         Bool = true.B,
+    use_sp_mem: Boolean = false,
```
sequencer, line 154:
```suggestion
    useSinglePortMem: Boolean = false,
```

src/main/scala/chisel3/util/Reg.scala
sequencer, line 135: How about using `name: Option[String]`? `null` in Scala is a bad design choice.

sequencer (changes requested):
After taking a look at the implementation of new ShiftRegister with `use_sp_mem = true`, I think this is simply adding a new instance of `mem` to increase `ShiftRegister` to multiple bank?

Besides the nitpick on names and the usage of `null`, I think this PR is actually "adding multiple banks support to `ShiftRegister`."? if so, I think the new banks to add should be `2,4,8...`, rather than only being `2`.

src/main/scala/chisel3/util/Reg.scala
milovanovic: Actually the whole idea is summarized in a [recent issue](https://github.com/chipsalliance/chisel3/issues/2889). In other words, please have a look at [slide 34](http://pages.hmc.edu/harris/class/e158/lect15-sram.pdf): for larger (i.e., wider and deeper) shift registers it is much more power and area efficient to implement them using SRAM than with flip-flops. If dual-port SRAM is available (i.e., if such memory compiler exists) that is a straightforward path. If you want to be even more power and area efficient, or if dual-port memory compiler isn't available in the target technology, you can alternatively use two half-size single-port SRAMs in ping-pong configuration. This is well explained in, say, [Chapter 9](https://onlinelibrary.wiley.com/doi/epdf/10.1002/9780470822500.ch9) (pp. 208-211) of the textbook [OFDM Baseband Receiver Design for Wireless Communications](https://onlinelibrary.wiley.com/doi/book/10.1002/9780470822500) by [Tzi-Dar Chiueh](http://cc.ee.ntu.edu.tw/~chiueh) and [Pei-Yun Tsai](https://scholars.ncu.edu.tw/en/persons/pei-yun-tsai). Even though it is certainly possible to have any number of banks, just as you've said, there is no real gain in power and area for anything larger than two due to (control) logic overhead.

By the way, thanks on the suggestion how to eliminate `null`. You can review it again.

sequencer: Thanks for the clear explanation! Would you mind renaming the arg to `useDualSRAMPort` and default to `false` or something similar and provide a documentation to which? That will be much clear to users.

sequencer (approved):

src/test/scala/chiselTests/Reg.scala
```diff
@@ -51,6 +51,16 @@ class ShiftTester(n: Int) extends BasicTester {
   }
 }
 
+class ShiftMemTester(n: Int, dp_mem: Boolean) extends BasicTester {
+  val (cntVal, done) = Counter(true.B, n)
+  val start = 23.U
+  val sr = ShiftRegister(cntVal + start, n, true.B, dp_mem, Some("simple_sr"))
```
mwachs5, line 57: looking a this test, should we call this flavor of `ShiftRegister` `ShiftMem`?

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -123,6 +123,96 @@ object ShiftRegister {
...
+
+  /** Returns the n-cycle delayed version of the input signal (SyncReadMem-based ShiftRegister implementation).
+    *
+    * @param in input to delay
+    * @param n  number of cycles to delay
+    * @param en enable the shift
+    * @param useDualSRAMPort dual port or single port SRAM based implementation
+    * @param name name of SyncReadMem object
+    */
+  def apply[T <: Data](in: T, n: Int, en: Bool, useDualSRAMPort: Boolean, name: Option[String]): T =
```
mwachs5, line 135: should this .apply have a more clear name, like `mem` or something?

`ShiftReg.mem(...)` ? Or as your test called it, `ShiftMem`?

milovanovic: I would rather leave it more general as it is now, nevertheless if you believe `.mem` or something else is more aligned to your code guidelines there's no problem to change it.

src/test/scala/chiselTests/Reg.scala
milovanovic: Believe that this is in connection to the subsequent comment.

src/main/scala/chisel3/util/Reg.scala
milovanovic: All right, please have a look one possible implementation that's renaming the `apply` method to `mem` so that the call is now `ShiftRegister.mem`. How does it seem to you?

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -124,6 +124,96 @@ object ShiftRegister {
     compileOptions:      CompileOptions
   ): T =
     ShiftRegisters(in, n, resetData, en).lastOption.getOrElse(in)
+
+  /** Returns the n-cycle delayed version of the input signal (SyncReadMem-based ShiftRegister implementation).
+    *
+    * @param in input to delay
+    * @param n  number of cycles to delay
+    * @param en enable the shift
+    * @param useDualSRAMPort dual port or single port SRAM based implementation
```
mwachs5, line 133:
```suggestion
    * @param useDualPortSram dual port or single port SRAM based implementation
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -124,6 +124,96 @@ object ShiftRegister {
...
+    * @param name name of SyncReadMem object
+    */
+  def mem[T <: Data](in: T, n: Int, en: Bool, useDualSRAMPort: Boolean, name: Option[String]): T =
```
mwachs5, line 136:
```suggestion
  def mem[T <: Data](in: T, n: Int, en: Bool, useDualPortSram: Boolean, name: Option[String]): T =
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -124,6 +124,96 @@ object ShiftRegister {
...
+    macro SourceInfoTransform.inNEnUseDualSRAMpNameArg
+
+  /** @group SourceInfoTransformMacro */
+  def do_mem[T <: Data](
+    in:              T,
+    n:               Int,
+    en:              Bool,
+    useDualSRAMPort: Boolean,
```
mwachs5, line 144:
```suggestion
    useDualPortSram: Boolean,
```

src/main/scala/chisel3/util/Reg.scala
```diff
@@ -124,6 +124,96 @@ object ShiftRegister {
...
+  )(
+    implicit sourceInfo: SourceInfo,
+    compileOptions:      CompileOptions
+  ): T = _apply_impl_mem(in, n, en, useDualSRAMPort, name)
+
+  private def _apply_impl_mem[T <: Data](
+    in:              T,
+    n:               Int,
+    en:              Bool = true.B,
+    useDualSRAMPort: Boolean = false,
```
mwachs5, line 155: and so on...

mwachs5 (commented):
Looking good! Please consider my suggestion on the parameter name

src/main/scala/chisel3/util/Reg.scala
milovanovic: Changed all occurrences of `useDualSRAMPort` to `useDualPortSram`, as you've suggested. I also think it looks a bit better with the new parameter name.

PR #2892, opened 2022-12-16
Port docs to use CIRCT

seldridge:
Change all docs to use CIRCT in their mdoc regions.  Everything worked without issue except for one weird mdoc region in hierarchy.md which I am not sure how it is even legal.  This mdoc region has been disabled for now.


docs/src/cookbooks/hierarchy.md
```diff
@@ -161,8 +166,10 @@ class Top extends Module {
   println(s"Width is: ${definition.width}")
 }
 ```
-```scala mdoc:verilog
-chisel3.stage.ChiselStage.emitVerilog(new Top())
+
+<!-- TODO: This currently does not work. -->
+```scala
+circt.stage.ChiselStage.emitSystemVerilog(new Top())
```
seldridge, line 172: This test errors in CIRCT with a StageError and little additional information. I'm actually very confused by this because the above `AddOne` doesn't even appear to be legal Scala. It's doing something like:

```scala
class Foo(val a: Int) {
  val a = a
}
```

docs/src/cookbooks/hierarchy.md
```diff
@@ -135,7 +135,12 @@ class Top extends Module {
 ```
 ```scala mdoc:passthrough
 println("```")
-chisel3.stage.ChiselStage.elaborate(new Top)
+val chiselCircuit = (new chisel3.stage.phases.Elaborate)
```
mwachs5, line 138: why did this need to change?

docs/src/cookbooks/hierarchy.md
mwachs5: what do you mean? It's doing out := in + 1

?

Oh, you mean the val width = width? The compiler plugin should be doing things with the @public that make that legal (yes, it's weird). I'm not sure how that would be different for circt stage though?

docs/src/explanations/chisel-type-vs-scala-type.md
```diff
@@ -47,28 +47,36 @@ class MyModule(gen: () => MyBundle) extends Module {
...
+
     // Need to initialize all hardware values
     xReg := DontCare
 }
 ```
 
 ```scala mdoc:invisible
 // Just here to compile check the above
-ChiselStage.elaborate(new MyModule(() => new MyBundle(3)))
+def elaborate(module: => chisel3.RawModule) = {
```
mwachs5, line 63: again this one could have just stayed as chisel3.stage.ChiselStage.elaborate(...)?

In the end I assume our plan is that chisel3.stage.ChiselStage will NOT have an emitSystemVerilog, and/or circt.stage will have all these same functions? If the latter is the plan, this PR could have just added a .elaborate() to circt.stage?

mwachs5 (commented):
thanks! looks great.

The comments about elaborate() aren't blocking. But we should figure out what's going on with the hierarchy example.

mwachs5:
also, did you do a diff of the generated mdoc before-and-after? I'm especially interested in any "crash"-es, are they reasonable with circt-stage...

docs/src/cookbooks/hierarchy.md
seldridge: `circt.stage.ChiselStage` has no `elaborate` method. This avoids any usage of `chisel3.stage.ChiselStage` by doing this manually.

docs/src/cookbooks/hierarchy.md
seldridge: It seems really concerning to me that we're rewriting illegal Scala code...

Can this instead be written as:

```scala
class Foo(_a: Int) {
  val a = _a
}
```

Can the `@public` rewriter cause _new_ name collisions with its rewriting?

There's the secondary problem in the above where `val a = a`, ignoring the double definition, ambiguously typed as it appears recursive.

seldridge: I'll dig into what's going on with CIRCT here. Part of what was flummoxing me was that I couldn't run this snippet in `Scala-cli` because I was forgot to add whatever the compiler plugin is that does the D/I magic.

jackkoenig: It's not compiler plugin, it's just the `@instantiable` macro. Doesn't really make a difference though, it's very weird that this doesn't work with CIRCT as there should be no difference...

seldridge: The issue is an abstract reset in the definition module which is then included in the compile.

Jack pointed me at `VerilogMdocModifier.scala` which needs to be fixed up before this PR lands.

seldridge:
Crashes look reasonable with #2896: 

```diff
-ChiselStage.emitVerilog(new BadRegConnect)
-// firrtl.passes.CheckHighFormLike$RegWithFlipException:  @[cookbook.md 821:20]: [module BadRegConnect] Register monitor cannot be a bundle type with flips.
+ChiselStage.emitSystemVerilog(new BadRegConnect)
+// circt.stage.phases.Exceptions$FirtoolNonZeroExitCode: firtool returned a non-zero exit code
+// ------------------------------------------------------------------------------
+// ExitCode:
+// 1
+// STDOUT:
+// 
+// STDERR:
+// cookbook.md:821:20: error: 'firrtl.reg' op result #0 must be a passive base type that does not contain analog, but got '!firrtl.bundle<ready flip: uint<1>, valid: uint<1>, bits: uint<8>>'
+// cookbook.md:821:20: note: see current operation: %5 = "firrtl.reg"(%arg0) {annotations = [], name = "monitor", nameKind = #firrtl<name_kind interesting_name>} : (!firrtl.clock) -> !firrtl.bundle<ready flip: uint<1>, valid: uint<1>, bits: uint<8>>
+// 
+// ------------------------------------------------------------------------------
+//     at circt.stage.phases.CIRCT.transform(CIRCT.scala:233)
+//     at circt.stage.phases.CIRCT.transform(CIRCT.scala:101)
+//     at firrtl.options.DependencyManager.$anonfun$transform$5(DependencyManager.scala:280)
+//     at firrtl.Utils$.time(Utils.scala:181)
+//     at firrtl.options.DependencyManager.$anonfun$transform$3(DependencyManager.scala:280)
+//     at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
+//     at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
+//     at scala.collection.immutable.List.foldLeft(List.scala:91)
+//     at firrtl.options.DependencyManager.transform(DependencyManager.scala:269)
+//     at firrtl.options.DependencyManager.transform$(DependencyManager.scala:255)
+//     at firrtl.options.PhaseManager.transform(DependencyManager.scala:443)
+//     at circt.stage.ChiselStage$.emitSystemVerilog(ChiselStage.scala:108)
+//     at repl.MdocSession$MdocApp44$$anonfun$158.apply(cookbook.md:830)
+//     at repl.MdocSession$MdocApp44$$anonfun$158.apply(cookbook.md:830)
```

docs/src/cookbooks/hierarchy.md
seldridge: This method should be added, but that would be a follow-up PR.

mwachs5 (approved):

seldridge:
> also, did you do a diff of the generated mdoc before-and-after? I'm especially interested in any "crash"-es, are they reasonable with circt-stage...

I realize my comment above wasn't clear. I did do this diff and everything looked reasonable. The only actual difference was related to that above comment when `firtool` exits with a non-zero exit code. This didn't look good before the rebase, but looks great now.

PR #2894, opened 2022-12-16
Bugfix waiveEach not casting. Add waiveAllAs

azidar:
### Contributor Checklist

#### Type of Improvement

  - bug fix

 - new feature/API 

#### API Impact

Adds a new API `Connectable.waiveAllAs`, which enables a more succinct syntax for doing "partial-connect-like" connections. Previously, you had to do `foo.waiveAll.asInstanceOf[Connectable[Data]]`.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/connectable/Connectable.scala
+++ b/core/src/main/scala/chisel3/connectable/Connectable.scala
@@ -214,7 +214,7 @@ object Connectable {
       * @group connection
       * @param producer the right-hand-side of the connection; will always drive leaf connections, and never get driven by leaf connections ("aligned connection")
       */
-    final def :<=[S <: Data](lProducer: => S)(implicit evidence: S =:= T, sourceInfo: SourceInfo): Unit = {
+    final def :<=[S <: Data](lProducer: => S)(implicit evidence: T =:= S, sourceInfo: SourceInfo): Unit = {
       val producer = prefix(consumer.base) { lProducer }
       connect(consumer, producer, ColonLessEq)
     }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/connectable/Connectable.scala
```diff
@@ -54,6 +54,9 @@ final class Connectable[+T <: Data] private (
     this.copy(waived = waivedMembers.toSet) // not appending waived because we are collecting all members
   }
 
+  /** Waive all members of base and static cast to a new type */
+  def waiveAllAs[S <: Data]: Connectable[S] = waiveAll.asInstanceOf[Connectable[S]]
```
jackkoenig, line 58: I think you may want some proof that `T <:< S` to let people call these functions. What happens if I do something like:
```scala
val foo = IO(Decoupled(UInt(8.W)))
foo.waiveAllAs[UInt]
```

azidar: You'd get a runtime error, but that's the same as if you used `.asInstanceOf[UInt]` right?

jackkoenig: Can you add a test showing the behavior? I'm not convinced you get a runtime error (because the cast is about a type parameter and those are erased at runtime), you might just get silently dropped connections

azidar: Oh, you are right. There is no error. Ok, I'll add the super type constraint implicit.

azidar: Actually, in my experience, having the constraint actually isn't a good idea. For example, you may want to connected a Decoupled to a Valid. Since DecoupledIO is not a subtype of Valid, you still may want to do:
```
val foo: DecoupledIO[UInt] = ???
val bar: Valid[UInt] = ???
foo.waiveAs[Valid[UInt]](_.ready) :<>= bar
```

jackkoenig (approved):

azidar:
@jackkoenig did you see my comment about connecting a DecoupledIO to a ValidIO ?

core/src/main/scala/chisel3/connectable/Connectable.scala
jackkoenig: You could try our shiny `ChiselSubtypeOf` https://github.com/chipsalliance/chisel3/blob/b90a803849724662c115e8bbb537587bdcec9f32/core/src/main/scala/chisel3/experimental/dataview/ChiselSubtypeOf.scala#L28

🙂

PR #2896, opened 2022-12-16
[circt] Cleanup CIRCT error behavior

seldridge:
Change the behavior of what happens if firtool returns a non-zero exit
code from printing information to stdout to throwing an exception with
this information in the body.  This is done primarily to ease
interoperability with mdoc.  However, this is generally better
principled than the previous behavior of scattering information in
different mediums.  Add a test of this behavior.

Stop intercepting a file not found error if CIRCT fails to print this to
stdout and wrap this in a StageError.  It is better to just let the file
not found exception through as this already includes all the information
necessary to tell the user what the problem is.  Do not test this
behavior (as it is tedious to test).


jackkoenig (approved):

seldridge:
I'll hold off on merging this until I can verify that this is what `mdoc` wants... This gets the message to display when building the doc, but not to show up in the `mdoc` output. Digging...

mwachs5:
what's the mdoc modifiers on the section you have? If it's `:silent` it won't show the output anyway... if it's `:crash` it expects an exception to be thrown and will show only the stack trace. You probably want `:passthrough` if you want to see the stderr/stdout output in your mdoc (if you aren't putting your dramaticError into an actual thrown exception). I am not sure if we often do `:crash:passthrough` but you could try that...? Or change this PR to put your message into the thrown exception.

seldridge:
As far as I can tell, there is no way to do this with how `mdoc` works unless maybe using a `PreModifier`. The `PostModifers` only give you access to what happens after the fact and usage of `:crash` seems to be wrapping certain statements with a `doc.crash` object so that this is specially handled. Basically, we're in a really weird place where `mdoc` seems to expect one of two use cases:

1. Watch for an exception.
2. Post-process the output.

For this to work as is, we need both.

Jack had a suggestion that we could instead use an exception here instead of printing to `stderr`. This would fix it and would likely be cleaner. The exceptions that come up here don't suffer from the problem of Chisel runtime exceptions that are buried in the stack trace. Anything CIRCT related would show up right at the top.

I'll rework this to change to CIRCT flavored exceptions with verbose text packed in their message.

WDYT?

mwachs5:
Having the exception message with all the useful info sounds good to me.

src/main/scala/circt/stage/phases/CIRCT.scala
```diff
@@ -185,55 +217,47 @@ class CIRCT extends Phase {
...
+    val stdoutWriter = new java.io.PrintWriter(stdoutStream)
+    val stderrWriter = new java.io.PrintWriter(stderrStream)
+    val exitValue = (cmd #< new java.io.ByteArrayInputStream(input.getBytes))
+      .!(ProcessLogger(stdoutWriter.println, stderrWriter.println))
+    stdoutWriter.close()
+    stderrWriter.close()
+    val result = stdoutStream.toString
+    logger.info(result)
+    val errors = stderrStream.toString
+    if (exitValue != 0)
```
mwachs5, line 232: In some environments a missing firtool binary led to an exception here instead of a nonzero exit code

Have you tested what this does if firtool is not on the path?

mwachs5 (commented):
Can you test (manually or with a test) how this fails if firtool is missing from the path? In some environments it failed differently than others... see the pr on chisel-circt where I changed this and let's make sure we aren't regressing?

src/main/scala/circt/stage/phases/CIRCT.scala
seldridge: This appears to not have been working correctly before either. I added a commit to do the right thing here.

This is the current output with the new `FirtoolNotFound` exception. The `Process` that gets created seems to do its own thing with exception printing that I can't sort out how to control. This will now narrowly try/catch the `Process` to sort out what happened if it returns a `RunTimeException` with a message indicating that firtool wasn't found:

```
Running CIRCT: 'firtool -format=fir -warn-on-unprocessed-annotations -verify-each=false -dedup -output-annotation-file circt.anno.json -disable-infer-rw -annotation-file Foo.anno.json < $input'
Exception in thread "Thread-9668" java.io.IOException: Cannot run program "firtool": error=2, No such file or directory
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1128)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
	at scala.sys.process.ProcessBuilderImpl$Simple.run(ProcessBuilderImpl.scala:75)
	at scala.sys.process.ProcessImpl$PipedProcesses.runAndExitValue(ProcessImpl.scala:142)
	at scala.sys.process.ProcessImpl$PipedProcesses.runAndExitValue(ProcessImpl.scala:121)
	at scala.sys.process.ProcessImpl$CompoundProcess.$anonfun$x$4$1(ProcessImpl.scala:100)
	at scala.sys.process.ProcessImpl$Spawn$$anon$1.run(ProcessImpl.scala:27)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.base/java.lang.ProcessImpl.forkAndExec(Native Method)
	at java.base/java.lang.ProcessImpl.<init>(ProcessImpl.java:340)
	at java.base/java.lang.ProcessImpl.start(ProcessImpl.java:271)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1107)
	... 6 more
circt.stage.phases.Exceptions$FirtoolNotFound: firtool not found
Chisel requires that firtool, the MLIR-based FIRRTL Compiler (MFC), is installed
and available on your $PATH.  (Did you forget to install it?)  You can download
a binary release of firtool from the CIRCT releases webpage:
  https://github.com/llvm/circt/releases
  at circt.stage.phases.CIRCT.transform(CIRCT.scala:247)
  at circt.stage.phases.CIRCT.transform(CIRCT.scala:117)
  at firrtl.options.DependencyManager.$anonfun$transform$5(DependencyManager.scala:280)
  at firrtl.Utils$.time(Utils.scala:181)
  at firrtl.options.DependencyManager.$anonfun$transform$3(DependencyManager.scala:280)
  at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:126)
  at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:122)
  at scala.collection.immutable.List.foldLeft(List.scala:91)
  at firrtl.options.DependencyManager.transform(DependencyManager.scala:269)
  at firrtl.options.DependencyManager.transform$(DependencyManager.scala:255)
  at firrtl.options.PhaseManager.transform(DependencyManager.scala:443)
  at circt.stage.ChiselStage$.emitSystemVerilog(ChiselStage.scala:108)
  ... 55 elided
```

mwachs5: Thanks!

mwachs5 (approved):

PR #2900, opened 2022-12-19
Demonstrate Circt Intrinsic

darthscsi:
### Contributor Checklist

#### Type of Improvement

- backend code generation
- new feature/API

#### API Impact

This a self-contained utility function.  This makes some minimal infrastructure to produce circt intrinsics.  The particular one used as an example is `sizeof` which allows querying (from the design) of the size of a type after width inference runs.

This targets specific Circt intrinsics allowing new code gen opportunities, but only when explicitly used in a design.

#### Desired Merge Strategy

Squash: The PR will be squashed and merged (choose this if you have no preference.

- [ ] Did you review?


src/main/scala/chisel3/util/SizeOf.scala
```diff
@@ -0,0 +1,30 @@
...
+  annotate(new ChiselAnnotation {
+    def toFirrtl =
+      intrinsic(toTarget, "circt.sizeof")
+  })
+  override val desiredName = "SizeOf" + _id.toString()
+}
+
+object SizeOf {
+  def apply[T <: Data](gen: T): Data = {
+    val sizeOfInst  = Module(new SizeOfIntrinsic(chiselTypeOf(gen)));
```
ekiwi, line 26: Could this use D/I in order to avoid duplicate modules? All we would need is a way to store the module cache in the Builder context in order to allow parallel elaboration of multiple designs.

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,33 @@
+package chiselTests.util
+
+import circt.stage.CIRCTStage
+
+import java.io.File
+import java.io.{PrintWriter, Writer}
```
ekiwi, line 6: Please use `os-lib` instead of `java.io`: https://github.com/com-lihaoyi/os-lib

ekiwi (changes requested):
Thanks. I added some suggestions.

src/main/scala/chisel3/util/SizeOf.scala
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util
```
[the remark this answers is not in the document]
seldridge: There is a PR open for intrinsics: https://github.com/chipsalliance/firrtl-spec/pull/43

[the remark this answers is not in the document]
ekiwi: The intrinsic itself would still be `circt` specific though, right? Then it should live under the `circt` namespace.
We could also just add a `sizeOf` primitive op in firrtl instead of a `circit` specific intrinsic.

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+    val w = Input(UInt(65.W))
+    val outw = UInt(32.W)
+  })
+  io.outw  := SizeOf(io.w)
+}
+
+/** A test for intrinsics */
+class SizeOfSpec extends AnyFlatSpec with Matchers {
+  it should "Should work for types" in {
+    val fir = circt.stage.ChiselStage.emitFIRRTLDialect(new SizeOfTop)
```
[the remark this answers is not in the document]
ekiwi: What if the way this is generated changes but the behavior stays compliant? What if the syntax of the CIRCT firrtl IR changes? What about non-CIRCT firrtl compilers?
If you generate a `BasicTester` then this has the advantage that the test will work with any firrtl compliant compiler.

src/main/scala/chisel3/util/SizeOf.scala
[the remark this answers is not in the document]
darthscsi: Planning wise, this is one of many intriniscs.  The point of inttrinsics is to separate features which need compiler knowledge, and hence cannot be libraries, from language specification.   This PR is infrastructure for a capability, which will eventually be in the spec as an op.  It is for now done with legal firrtl.  Which is why I put this in chisel3: the output of using the intrinsic objects is valid firrtl now.  (just as SSE intriniscs are valid C functions, they aren't a new kind of thing, just a specific interpretation of an existing thing)

src/main/scala/chisel3/util/SizeOf.scala
darthscsi: Probably, but it seems a lot of work for avoiding a few redundant external modules.

src/test/scala/chiselTests/util/SizeOfSpec.scala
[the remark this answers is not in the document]
darthscsi: Chisel generates firrtl, so what matters is that the firrtl + annotations are what they should be.  There is no reason to repeat tests which already exist in firrtl compilers for the actual implementation of the intrinsic.

The test is updated to just check the firrtl is the expected form.

When the spec has explicit intrinisics,  we'll check those are generated correctly.

src/main/scala/chisel3/util/SizeOf.scala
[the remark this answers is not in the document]
mwachs5: I'm all for adding intrinsic emission support to chisel3. But i'm also confused about this specific sizeOf intrinsic. What's an expected use case for this particular one? Is this best as a test case that uses the chisel3-support-for-intrinsics?

src/main/scala/chisel3/util/SizeOf.scala
mwachs5: Because using D/I limits the use of APIs like Select, I don't think it's appropriate to unconditionally use in library functions. It's also still quite experimental.

src/main/scala/circt/Intrinisic.scala
```diff
@@ -0,0 +1,15 @@
...
+/** This has to be in a separate file and package to generate the correct class
+  *  in the annotation.
+  */
+
+package circt
+
+import firrtl.annotations.SingleTargetAnnotation
+import firrtl.annotations.ModuleTarget
+
+case class intrinsic(target: ModuleTarget, intrinsic: String) extends SingleTargetAnnotation[ModuleTarget] {
```
mwachs5, line 13:
```suggestion
case class Intrinsic(target: ModuleTarget, intrinsic: String) extends SingleTargetAnnotation[ModuleTarget] {
```

Is there a reason to deviate from Scala style here to make this lowercase?

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  })
+  io.outw  := SizeOf(io.w)
+}
+
+/** A test for intrinsics.  Since chisel is producing intrinsics as tagged 
+ * extmodules (for now), we explicitly test the chirrtl and annotations rather
+ * than the processed firrtl or verilog.  It is worth noting that annotations 
+ * are implemented (for now) in a way which makes the output valid for all
+ * firrtl compilers, hence we write a localized, not end-to-end test */
+class SizeOfSpec extends AnyFlatSpec with Matchers {
```
mwachs5, line 26: my comment above this could just be IntrinsicSpec and we define the Sizeof intrinsic here... I'm unclear why anyone would want the sizeof functionality in hardware design

src/test/scala/chiselTests/util/SizeOfSpec.scala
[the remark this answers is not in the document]
ekiwi: > Chisel generates firrtl, so what matters is that the firrtl + annotations are what they should be.

This test does not check standard firrtl, but some internal CIRCT dialect instead.
Also, you are not checking the firrtl generated by Chisel (which would not be as brittle), but the firrtl generated after running some CIRCT compiler passes, which means that if CIRCT ever decides to represent the result differently, i.e. a different constant encoding or somehow adding an intermediate signal, then this test would break.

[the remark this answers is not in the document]
ekiwi: Not sure how many Chisel tests you have so far ported to CIRCT, however, I assume that all the execution tests will be easy to port - you just need to change the way the Verilog is generated - while any test that matches on the exact Verilog strings being generated will require manual effort to port. Thus executable tests are much more robust wrt. future compiler changes.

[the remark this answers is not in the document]
mwachs5: We have lots of tests that check emitted .fir or .v. There is already a PR open by @chick starting to port them over.

> The test is updated to just check the firrtl is the expected form.

@ekiwi are you saying this is still not the case, since you last reviewed it?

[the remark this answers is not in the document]
ekiwi: > are you saying this is still not the case, since you last reviewed it?

Ah sorry, somehow github was showing me the old code. I think the new version is an improvement. I am still a big fan of executable tests where possible, but the new version is fine since it is a lot more robust wrt. firrtl compiler improvements.

src/main/scala/chisel3/util/SizeOf.scala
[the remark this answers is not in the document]
darthscsi: It is generally useful and a very standard meta-programming function when you have inferred widths or other cases of non-statically knowable types to correctly handle the result of inference.  Chisel has width inference, so being able to react (statically in the generated code, since you can't react to it in Chisel land) to the result of that inference is useful.  (for example, computing values to put on a bus based on the width in a type-safe way)

This is also a simple example of an intrinsic to get the ball rolling on the infrastructure.  It is a capability which requires compile support and cannot be done at chisel time or as a library (well, with width inference, I can think of a few ways, but they are super ugly).

src/test/scala/chiselTests/util/SizeOfSpec.scala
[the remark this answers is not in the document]
darthscsi: >  Thus executable tests are much more robust wrt. future compiler changes.

Not here, since not all compilers or compiler versions are expected to support an intrinsic.  Executable tests are testing both the generation which is compiler and version independent and the implementation which is compiler and version dependent.  Just testing that the correct firrtl is generated is more specific and more robust.

src/main/scala/circt/Intrinisic.scala
darthscsi: For the not great reason that the class name appears in the output (there doesn't seem to be a way to make the annotation json have "class=whatever_you_want_which_isnt_based_on_scala_name").

Probably easy to fix.

src/main/scala/chisel3/util/SizeOf.scala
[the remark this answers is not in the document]
mwachs5: It makes sense in a near-future world where we can react to information after width inference Ints and Strings to potentially build hardware with widths that react to that value, but I don't think one could emit verilog that reacts to a sizeof value being driven on a wire, so I'm still not sure what the immediate use case is here.

mwachs5:
> The specific sizeOf intrinsic however is circt specific and thus should live under the circt package.

And I'm still questioning why it belong anywhere except in the tests of intrinsics 🤷‍♀️

ekiwi:
> And I'm still questioning why it belong anywhere except in the tests of intrinsics

That would also be fine with me. I just think that if something is going to be part of the firrtl spec, it should be supported by any hypothetical firrtl compiler and thus does not need to be prefixed by `circt`.

seldridge:
> And I'm still questioning why it belong anywhere except in the tests of intrinsics 🤷‍♀️ 

Because it's already supported in CIRCT. 😉  I believe there was some interest in using this for parameterization experiments and/or to enforce checks on separably compiled units. (I.e., this was intended to be used, not just a toy demonstration.)

> However, I still think the packages are backwards. Intrinsics are going to be a standard firrtl feature, thus they should live under the standard chisel3 package. The specific sizeOf intrinsic however is circt specific and thus should live under the circt package.

Your point about putting the annotation under a chisel3 package is reasonable. I'd suggest `chisel3.annotations.Intrinsic` or `firrtl.annotations` (as long as there are no conflicts). Using `firrtl.annotations` is easier once SFC infra is mono-repoed under Chisel.

For the `sizeOf` intrinsic, the question that has to be answered is: Does this go into the FIRRTL spec?

If this does, then it's fine to put this under some `chisel3` area (again, a `firrtl` area would be better). If it does not, then there is some question of where to put it. Using a `circt` package seems fine.

ekiwi:
> For the `sizeOf` intrinsic, the question that has to be answered is: Does this go into the FIRRTL spec?

We can always move it as soon as a PR to add it to the FIRRTL spec is accepted.

ekiwi:
> We can always move it as soon as a PR to add it to the FIRRTL spec is accepted.

The question though is: If it is part of the FIRRTL spec, shouldn't it be a Primitive Operation instead of an intrinsic?

seldridge:
> We can always move it as soon as a PR to add it to the FIRRTL spec is accepted.

Both of these are living projects and one doesn't have to change before the other. There have been tons of features that went into Chisel first which made there way into the FIRRTL spec later. Considering this is effectively just another annotation, I don't see a big problem with moving Chisel first and considering changes to package names later.

> The question though is: If it is part of the FIRRTL spec, shouldn't it be a Primitive Operation instead of an intrinsic?

It depends.

The point of intrinsics is that they serve specialized use cases without having to move the FIRRTL spec to be the set of all possible operations that anybody could want. It is also reasonable to add primitive operations when they clearly make sense within the bounds of digital logic.

Something like `sizeOf` is complicated as it is outside the bounds of normal digital logic and may not have a well-defined lowering for all backends, e.g. VHDL. (My Google-fu is indicating that VHDL has attributes for encoding size and does have system functions for doing this like Verilog.)

Because of this, it is reasonable for the spec to define certain intrinsics as opposed to leaving them entirely to each compiler to define.

That said, all this can change in the future. The spec and Chisel are flexible and can be updated rapidly. There is not an assumption that we will get this right the first (or second, or third!) time. (This is part of the Chisel 5 push in the roadmap is to enable faster development with the expectation that we may make mistakes.)

darthscsi:
> We can always move it as soon as a PR to add it to the FIRRTL spec is accepted.

The fixation on the firrtl spec status of `sizeof` is misplaced.  The point of intrinsics is to provide a type-safe, defined extension mechanism to the language for features which cannot be implemented by libraries.  This requires a standard way to define them.  Right now, this is encoded with extmodule and an annotation.  Intrinsics will come to firrtl as first class things at which point the chisel implementation can switch to using them.  Intrinsics are not yet part of firrtl and it behooves everyone to use a light weight, backwards compatible implementation, like this, to get experience with them before pushing them into the standard.

As for what the set of intrinsics which can be generated are, that is up to the implementations.  Right now this set for Circt is 4, and for SFC is 0.  Regardless, the output of chisel for an intrinsic is well defined, even if the backend doesn't support it.  It is also fine if multiple backends pick up support for the same intrinsic, even if it was originally defined by a different backend.  Clang has extensive support for gcc intrinsics, for example.  `sizeof` is useful, and probably will be folded into a primitive operation at some point, but let's see how the meta-programming capabilities play out before committing to that as a part of the language.

darthscsi:
> Intrinsics are going to be a standard firrtl feature, thus they should live under the standard `chisel3` package.

And when they do, the support should go there.  For now, they are implemented in a circt-specific annotation which is not supported by SFC.  Long term they should be part of the spec, but while they are target specific, it seems sensible to put the annotation string in that target.

When intrinsics are part of the spec, chisel will port over and generate them directly and not depend on this annotation at all.  At that point, this code will be removed.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,30 @@
...
+  * results of type inference.
+  */
+private class SizeOfIntrinsic [T <: Data](gen: T) extends ExtModule {
+  val i    = IO(Input(gen));
+  val size = IO(Output(UInt(32.W)));
+  annotate(new ChiselAnnotation {
+    def toFirrtl =
+      Intrinsic(toTarget, "circt.sizeof")
+  })
+  override val desiredName = "SizeOf" + _id.toString()
```
seldridge, line 21: This `_id` to string is going to be a numeric Builder ID which isn't what is actually wanted. What you can do is just not give this a `desiredName` and every instance will get a unique `_<n>` suffix. This isn't ideal, but there won't be any conflicts in the defname.

darthscsi: I don't think it will for desiredName, I want each module instance to have a unique one, I think the suffix only applies (by design) to the module name.

darthscsi: It's a dead string but required since I can't emit an extModule without it.

seldridge: Ack... otherwise the desired name is the module name...

darthscsi: Used the hash of the string of the type.  Should be properly unique and stable.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,32 @@
...
+  val size = IO(Output(UInt(32.W)));
+  annotate(new ChiselAnnotation {
+    def toFirrtl =
+      Intrinsic(toTarget, "circt.sizeof")
+  })
+  override val desiredName = "SizeOf_" + MurmurHash3.stringHash(gen.toString()).toHexString
+}
+
+object SizeOf {
+  def apply[T <: Data](gen: T): Data = {
```
mwachs5, line 27:
```suggestion
  /** can we put scaladoc here, since it won't appear in docs on the private class?
   *
   * Can we include this motivating example:
   *
   * @example
   * ```
   * // sets the top bit of a possibly width-inferred bus
   *
   * a := 1 << (SizeOf(a) - 1)
   * ```
   */
  def apply[T <: Data](gen: T): Data = {
```

src/main/scala/chisel3/stage/ChiselStage.scala
```diff
@@ -195,6 +194,28 @@ object ChiselStage {
...
+        Dependency[chisel3.stage.phases.Convert],
+        Dependency[chisel3.stage.phases.MaybeInjectingPhase]
+      )
+    }
+
+    phase
+      .transform(Seq(ChiselGeneratorAnnotation(() => gen)))
+      .map(a => a.serialize)
+      .reduceLeft(_ + "\n" + _)
+  }
```
seldridge, line 217: This isn't necessary to add. I'll add an example of doing this in the test.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,32 @@
...
+/** Create a module with a parameterized type which returns the size of the type
+  * as a compile-time constant.  This lets you write code which depends on the
+  * results of type inference.
+  */
+private class SizeOfIntrinsic[T <: Data](gen: T) extends ExtModule {
+  val i = IO(Input(gen));
+  val size = IO(Output(UInt(32.W)));
+  annotate(new ChiselAnnotation {
+    def toFirrtl =
+      Intrinsic(toTarget, "circt.sizeof")
```
seldridge, line 21: Mega-nit: Using `override def` is preferable when implementing abstract methods. Extremely annoyingly, Scala has no way to force this.

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+    (fir should include).regex(b1)
+    val a2 = """extmodule SizeOf_e""".r
+    (fir should include).regex(a2)
+    val b2 = """defname = SizeOf_e""".r
+    (fir should include).regex(b2)
+
+    // The second elaboration uses a unique name since the Builder is reused (?)
+    val anno = ChiselStage.emitAnnotations(new SizeOfTop)
+    val c = """Intrinsic\(~SizeOfTop\|SizeOf.*,circt.sizeof\)"""
+    (anno should include).regex(c)
```
seldridge, line 46: This can use patten matching to clean this up somewhat. Consider:

```scala
  ((new chisel3.stage.ChiselStage)
    .execute(
      args = Array("--no-run-firrtl"),
      annotations = Seq(chisel3.stage.ChiselGeneratorAnnotation(() => new SizeOfTop))
    )
    .mkString("\n") should include).regex(c)
```

Alternatively, if you make the Intrinsic annotation package private `private[chisel3]` and place this test in the `chisel3` package, you can then test this more directly without using strings. I.e., the test is in the `chisel3` package, has access to the annotation, and can then pattern match on it directly.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,32 @@
...
+  })
+  override val desiredName = "SizeOf_" + MurmurHash3.stringHash(gen.toString()).toHexString
```
seldridge, line 23: I'm still slightly worried about this hash approach, especially given that `MurmurHash3` could realistically have collisions. I think this is a reasonable place to start with this, though.

seldridge (approved):
Nice. I think this is a fine place to start playing with intrinsics.

I have a couple of nitty comments throughout and an alternative to avoid adding the `emitAnnotations` method. We probably eventually want something like this, but it's not necessary in this patch.

src/main/scala/chisel3/util/circt/SizeOf.scala
darthscsi: My concern is that the type's string won't always be well formed, but I don't know if that is justified.  If it is, then it could just be suitably escaped.  If it's not, then a hash doesn't help.  Hashing just seemed easier than escaping the string to be firrtl compatible.

seldridge: 🤔 The reasoning makes sense. It may be problematic to use the type as this is only guaranteed to be unique for ground types (with the restriction that external modules cannot have uninformed widths). E.g., could a user could cause a collision if they have two bundles with the same name?

seldridge: Using the global object avoids the problem.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,55 @@
...
+import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule}
+
+import circt.Intrinsic
+
+// We have to unique designedName per type, be we can't due to name conflicts
+// on bundles.  Thus we use a globally unique id.
+private object SizeOfGlobalIDGen {
+  private var id: Int = 0
+  def getID() = {
+    this.synchronized {
```
seldridge, line 17: Nice use of synchronized.

seldridge (approved):
LGTM

src/main/scala/chisel3/util/circt/SizeOf.scala
ekiwi: @jackkoenig  This is another example, where a Builder context cache would be useful. `synchronized` works here, but only because gaps in the IDs are acceptable.

ekiwi (changes requested):
I still would like to have `Intrinsic.scala` moved into a `chisel3` (sub)package. Since intrinsics are not a `circt` specific thing, but something that has been standardized. So I see no reason for it to reside in a special `circt` package.

src/main/scala/chisel3/util/circt/SizeOf.scala
jackkoenig: Agreed, can this instead use the same pattern as `Instantiate` which has [a cache that is part of the Builder](https://github.com/chipsalliance/chisel3/blob/df203ea8a7331ead3ed7a6e3ef2d0a5a8bb1a9f5/core/src/main/scala/chisel3/internal/Builder.scala#L553)? See https://github.com/chipsalliance/chisel3/pull/2717 for an example of what needs to be added.

Then at least we will have all examples that need the pattern grouped up so that we'll remember to fix them when we add the Builder context cache.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,55 @@
...
+    * is after width inference, so you can use this to compute expressions based
+    * on the inferred size of types.
+    *
+    * @example {{{
+    * val a = Wire(UInt())
+    * a := 1 << (SizeOf(a) - 1)
+    * }}}
+    */
+  def apply[T <: Data](gen: T): Data = {
+    val sizeOfInst = Module(new SizeOfIntrinsic(chiselTypeOf(gen)));
```
jackkoenig, line 51:
```suggestion
    val sizeOfInst = Module(new SizeOfIntrinsic(chiselTypeOf(gen)))
```
No need for semicolons unless you are doing multiple statements on a single line.

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.circt
```
jackkoenig, line 3: Should we call this experimental? Especially because it doesn't have firrtl compiler support yet?

src/main/scala/circt/Intrinisic.scala
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/** This has to be in a separate file and package to generate the correct class
+  *  in the annotation.  This is supper annoying.
+  */
```
jackkoenig, line 5: Unnecessary comment and sort of weird. Yes, in Scala you set the package on a per file basis.

src/test/scala/chiselTests/MultiAssign.scala
```diff
@@ -5,7 +5,7 @@ package chiselTests
 import chisel3._
 import chisel3.testers.BasicTester
 import chisel3.util._
-import circt.stage.ChiselStage
+import _root_.circt.stage.ChiselStage
```
jackkoenig, line 8: Obviously it's fine to specify starting at the root, but alternatively you could put this import before the `import chisel3.util._` to avoid ambiguity with the new package you're adding.

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+    val b2 = """defname = SizeOf_1""".r
+    (fir should include).regex(b2)
+
+    // The second elaboration uses a unique name since the Builder is reused (?)
+    val c = """Intrinsic\(~SizeOfTop\|SizeOf.*,circt.sizeof\)"""
+    ((new chisel3.stage.ChiselStage)
+      .execute(
+        args = Array("--no-run-firrtl"),
+        annotations = Seq(chisel3.stage.ChiselGeneratorAnnotation(() => new SizeOfTop))
+      )
```
jackkoenig, line 53:
```suggestion
    chisel3.stage.ChiselStage.emitChirrtl(new SizeOfTop)
```
This can be terser but is fine as is, doesn't really matter.

src/main/scala/chisel3/util/circt/SizeOf.scala
seldridge: We have MFC support for this and three other intrinsics: https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp#L198

darthscsi: It's been supported in circt since Dec 18 or so.

src/main/scala/circt/Intrinisic.scala
darthscsi: No, it's that I can't control the class string in the json annotation.  It's weird that class structure implementation details are reflected in the output.

seldridge: Specifically, this needs to line up with "circt.Intrinsic` in the JSON class name so that it works here: https://github.com/llvm/circt/blob/cd0cae2fd1b0c765b90bc51f87025bb11196c6c4/lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp#L317

src/test/scala/chiselTests/util/SizeOfSpec.scala
darthscsi: emitChirrtl (at least.toString()) doesn't give the annotations.

darthscsi:
> I still would like to have `Intrinsic.scala` moved into a `chisel3` (sub)package. Since intrinsics are not a `circt` specific thing, but something that has been standardized. So I see no reason for it to reside in a special `circt` package.

Please provide a link to the section of the firrtl spec in which they appear.

docs/src/explanations/chisel-enum.md
```diff
@@ -13,9 +13,9 @@ In contrast with `Chisel.util.Enum`, `ChiselEnum` are subclasses of `Data`, whic
 
 ```scala mdoc
 // Imports used in the following examples
+import circt.stage.ChiselStage
```
mwachs5, line 16: any particular reason you moved this line?

src/main/scala/chisel3/util/circt/SizeOf.scala
mwachs5: is synchronized deterministic?

docs/src/explanations/chisel-enum.md
darthscsi: To get rid of `_root_.` per Jack's fine suggestion.

darthscsi: But perhaps this one just got caught up in the refactoring.

src/main/scala/chisel3/util/circt/SizeOf.scala
ekiwi: > is synchronized deterministic?

IDs may be non-deterministic if you are elaborating more than one circuit in a single process

src/main/scala/circt/Intrinisic.scala
ekiwi: Why not change the name in CIRCT to match the one in chisel3? Seems like an unnecessary restriction if we have to stick to the one currently used in CIRCT.

mwachs5:
```suggestion
/** This has to have the class name `circt.Intrinsic` to match the class name 
  * for the annotation that CIRCT is expecting.
  */
```

docs/src/explanations/chisel-enum.md
darthscsi: I thought it was required because the build was failing.  Figuring out how to build the doc examples locally.

src/main/scala/circt/Intrinisic.scala
mwachs5: in other words, can we file a bug / feature request for the "annoying" part

ekiwi:
> Please provide a link to the section of the firrtl spec in which they appear.

Oh, I did not realize that intrinsic hasn't made it in yet. Thanks.

src/test/scala/chiselTests/BlackBox.scala
```diff
@@ -2,11 +2,11 @@
 
 package chiselTests
 
+import circt.stage.ChiselStage
 import chisel3._
 import chisel3.experimental._
 import chisel3.testers.{BasicTester, TesterDriver}
 import chisel3.util._
-import circt.stage.ChiselStage
```
mwachs5, line 9: curious why you moved the imports around here?

ekiwi (approved):
LGTM once @jackkoenig and @mwachs5 's comments have been addressed.

src/test/scala/chiselTests/util/SizeOfSpec.scala
```diff
@@ -0,0 +1,56 @@
+package chiselTests.util
+
+import chisel3._
+import chisel3.stage.{ChiselGeneratorAnnotation, ChiselStage}
```
mwachs5, line 4: any reason to not use the circt.stage one HERE?

src/test/scala/chiselTests/util/SizeOfSpec.scala
mwachs5: i thought we were moving everythign to circt stage?

src/test/scala/chiselTests/util/SizeOfSpec.scala
seldridge: `circt.stage.ChiselStage` doesn't have a `--no-run-firrtl` option to enable just getting the annotations. The main problem isn't use of `chisel3.stage.ChiselStage`, but use of the SFC. (We should add `--no-run-firrtl` to the circt variant, but that's out of scope for this PR. Adam was going to add it after Chick also ran into the same issue.

src/test/scala/chiselTests/BlackBox.scala
seldridge: There are comments buried above. `chisel3.util.circt` (added by this PR) confuses `import circt.stage.ChiselStage`. Andrew fixed this with `import _root_.circt.stage.ChiselStage`. Jack suggested moving the imports instead.

src/test/scala/chiselTests/util/SizeOfSpec.scala
darthscsi: Because I want the interchange output, not the compiled output.

docs/src/explanations/chisel-enum.md
darthscsi: no, it's needed too.

src/test/scala/chiselTests/BlackBox.scala
darthscsi: Because apparently scala doesn't search from the root first for new imports, and adding `_root_.` was kind of ugly.

docs/src/explanations/chisel-enum.md
jackkoenig: 😂

src/main/scala/circt/Intrinisic.scala
darthscsi: > Why not change the name in CIRCT to match the one in chisel3? Seems like an unnecessary restriction if we have to stick to the one currently used in CIRCT.

We did once already, in response to a capitalization comment above.  But the annoying part is that annotation text is tied to Scala class name.

docs/src/explanations/chisel-enum.md
mwachs5: yah, well I guess it is good if we are consistent the order.

src/main/scala/chisel3/util/circt/SizeOf.scala
mwachs5: hm, i do not like that. That is something we do, e.g. in testing code. Can we do what Jack suggested?

PR #2907, opened 2022-12-22
Check for Vec subaccess in NamedComponent and throw a nicer error.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- code cleanup

#### API Impact

The only effect on the current API should be that existing usages that threw errors should now throw a more precise and helpful error.

This has no impact on code generation.

#### Desired Merge Strategy

- Squash

#### Release Notes

Improved error message when targeting Vec subaccess

- [x] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -292,17 +292,21 @@ private[chisel3] trait HasId extends InstanceId {
   private def refName(c: Component): String = _ref match {
     case Some(arg) => arg.fullName(c)
     case None => {
-      val nameGuess = _computeName(None) match {
-        case Some(name) => s": '$name'"
-        case None       => ""
-      }
-      val parentGuess = _parent match {
-        case Some(ViewParent) => s", in module '${reifyParent.pathName}'"
-        case Some(p)          => s", in module '${p.pathName}'"
-        case None             => ""
-      }
-      throwException("You cannot access the .instanceName or .toTarget of non-hardware Data" + nameGuess + parentGuess)
+      throwException(
+        "You cannot access the .instanceName or .toTarget of non-hardware Data" + _nameGuess + _parentGuess
+      )
     }
+  }
+
+  private[chisel3] def _nameGuess: String = _computeName(None) match {
+    case Some(name) => s": '$name'"
+    case None       => ""
+  }
+
+  private[chisel3] def _parentGuess: String = _parent match {
+    case Some(ViewParent) => s", in module '${reifyParent.pathName}'"
+    case Some(p)          => s", in module '${p.pathName}'"
+    case None             => ""
   }
 
   // Helper for reifying views if they map to a single Target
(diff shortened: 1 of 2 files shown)


mikeurbach (commented):
Some self-review... I had a couple questions while working on this. I'll also need to run scalafmt I'm sure.

This should subsume https://github.com/chipsalliance/chisel3/pull/2779 and close https://github.com/chipsalliance/chisel3/issues/2778.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -292,19 +292,27 @@ private[chisel3] trait HasId extends InstanceId {
...
-        case Some(name) => s": '$name'"
-        case None       => ""
-      }
-      val parentGuess = _parent match {
-        case Some(ViewParent) => s", in module '${reifyParent.pathName}'"
-        case Some(p)          => s", in module '${p.pathName}'"
-        case None             => ""
-      }
-      throwException("You cannot access the .instanceName or .toTarget of non-hardware Data" + nameGuess + parentGuess)
+      throwException(
```
mikeurbach, line 295: I'm not sure why scalafmt changed the breaking on this line... I ran it as described in CONTRIBUTING.md.

src/test/scala/chiselTests/VecToTargetSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    val vecWithTmp = WireInit(vecNodeIdx)
+  }
+
+  var foo: Foo = null
+
+  ChiselStage.elaborate { foo = new Foo; foo }
+
+  val expectedError = "You cannot target Vec subaccess:"
+}
+
```
jackkoenig, line 34: You really don't want to run things like Chisel elaboration during the constructor of a Spec class. It means this code will run before ScalaTest knows what the tests there are to run (Think about each of those `<string> should <string> in { ... }` as test definitions that run before we actually decide which tests to run). Instead, you should use a [Fixture](https://www.scalatest.org/user_guide/sharing_fixtures), or use a testing stile that allows nesting like `FreeSpec` or `FunSpec`, then you can do the elaboration in the outermost block, eg.
```scala
class VecToTargetSpec extends ChiselFunSpec with VecToTargetSpecUtils with Utils {
  describe("VecToTargets") {
    var foo: Foo = null
    ChiselStage.elaborate { foo = new Foo; foo }
    it("Vec subaccess with Scala literal should convert to ReferenceTarget") {
      // Use foo here
    }
  }
}

File: core/src/main/scala/chisel3/internal/Builder.scala (hunk 1 above)
Comment (jackkoenig) [reply core/src/main/scala/chisel3/internal/Builder.scala:295]: It is weird but whatever, we let the tool do what it wants 🤷‍♂️

File: core/src/main/scala/chisel3/internal/Builder.scala [hunk 3]
```diff
@@ -390,6 +393,14 @@ private[chisel3] trait NamedComponent extends HasId {
       case None             => localTarget
     }
   }
+
+  private def isVecSubaccess(): Boolean = {
```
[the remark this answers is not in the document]
jackkoenig: It might be useful as a helper, but I think it's right to just have it as a private method to start.

src/test/scala/chiselTests/VecToTargetSpec.scala
mikeurbach: Ok that makes sense. I looked into AnyFunSpec a bit and have reworked this a bit.

mikeurbach:
@jackkoenig just a friendly ping after the holidays. Do you mind taking another look? I tried to internalize your previous comment about using AnyFunSpec, and added a little more spec re-use while I was at it.

jackkoenig (approved):

mikeurbach:
Thanks @jackkoenig !

PR #2912, opened 2022-12-27
Add message to cover

sequencer:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -429,7 +429,7 @@ object cover extends VerifPrintMacrosDoc {
   ): Cover = {
     val id = new Cover()
     when(!Module.reset.asBool) {
-      Builder.pushCommand(Verification(id, Formal.Cover, sourceInfo, Module.clock.ref, cond.ref, ""))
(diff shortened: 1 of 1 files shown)


ekiwi:
How is this message supposed to be displayed by different backends like SymbiYosys or Jasper Gold?
The reason `cover` does not have a message was that there did not seem to be a way to display it with the existing tools. If you want to give your cover points a name, you should be able to just use `cover(...).suggestName("name")`

sequencer:
i see. In this case, I think we should remove the message field in spec to avoid misleading to users.

seldridge:
It looks like the spec had a message for covers since they were originally added: https://github.com/chipsalliance/firrtl/pull/1653

I can see this going one of two ways:

1. FIRRTL (and Chisel) are free to define a message for covers. It doesn't matter if formal tools can show these or not as a valid lowering (for some backend) is to just drop the message. We could invent a lowering for these. Extracting a name from the message seems wrong, though, as this creates divergence between how assert/assume and cover treat their message.
2. The assert/assume/cover optional name need to be accessible from Chisel. `suggestName` is probably the closest thing to that right now. Does this work today?

MFC should pull the label from the optional name if it is provided. I think the issue right now is that that isn't accessible from Chisel.

I think (2) makes sense and we should not have a message for cover while also making the label accessible from Chisel.

PR #2923, opened 2023-01-08
Improve performance of ChiselEnum annotations

jackkoenig:
ChiselEnums check if they should create annotations every time an instance of them is bound. Because so many annotations would be created, they check to see if an equivalent annotation has already been added to the annotations. Previously, this used a linear search of the annotations, now it uses a HashSet.

I noticed this inefficiency glancing at a Flamegraph of a very large design. In that case, just eyeballing where this is obviously showing up, it's at least 18% of the runtime. As with all things that are fundamentally `O(n*m)`, the benefits from this change will be greatest in large designs with lots of annotations (`n`) and lots of `ChiselEnum`s (`m`).

### Contributor Checklist

#### Type of Improvement

  - performance improvement

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Improve performance of ChiselEnum

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/StrongEnum.scala
+++ b/core/src/main/scala/chisel3/StrongEnum.scala
@@ -240,11 +240,13 @@ abstract class EnumType(private[chisel3] val factory: EnumFactory, selfAnnotatin
       case None    => EnumComponentChiselAnnotation(this, enumTypeName)
     }
 
-    if (!Builder.annotations.contains(anno)) {
+    if (!Builder.enumAnnos.contains(anno)) {
+      Builder.enumAnnos += anno
       annotate(anno)
     }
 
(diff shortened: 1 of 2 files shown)


mwachs5 (approved):

ekiwi:
@jackkoenig This would be a good use case for providing a `Builder` local cache for libraries. Maybe with Keys similar to how the CDE works. There are many different reasons why a library might want to cache things local to a `Builder` context, e.g., Diplomacy fails to elaborate in parallel because it has its own global cache.

jackkoenig:
@ekiwi I agree, and Instantiate being another good sample: https://github.com/chipsalliance/chisel3/pull/2717

PR #2929, opened 2023-01-10
Promote ChiselEnum user APIs from experimental

jackkoenig:
They are commonly used and if we were ever to change them, we will need
to go through a deprecation process anyway.

Note that the EnumAnnotations remain in chisel3.experimental because,
like all Chisel annotation support, they are slated to be deprecated and removed.

In the backport, the types won't actually move but will be deprecated while the aliases will not be deprecated.
Automated backport is safe because this will fail binary compatibility checking and I'll fix it as described.

This also has the nice property of adding `ChiselEnum` to the ScalaDoc rather than it being called `EnumFactory`.
I also renamed the tests and changed all `StrongEnum` to `ChiselEnum`.

### Contributor Checklist

#### Type of Improvement

 - code refactoring

#### API Impact

This moves the ChiselEnum APIs into package `chisel3` and provides deprecated aliases in `chisel3.experimental`.

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Move `ChiselEnum` and related types from package `chisel3.experimental` to package `chisel3`.

- [ ] Did you review?


seldridge (approved):
LGTM

If we aren't already doing it, feel free to stop emitting the annotations in a follow-on!

ekiwi:
> If we aren't already doing it, feel free to stop emitting the annotations in a follow-on!

I would prefer if you could hold on with removing the annotation until enums are part of the IR. I have a pass that relies on the annotation and I am actually planning on integrating that into `chiseltest` in the near future. In general it would be nice to only remove features once the replacement is ready (and enums are something where I think we all agree on the replacement).

jackkoenig:
> > If we aren't already doing it, feel free to stop emitting the annotations in a follow-on!
> 
> I would prefer if you could hold on with removing the annotation until enums are part of the IR. I have a pass that relies on the annotation and I am actually planning on integrating that into `chiseltest` in the near future. In general it would be nice to only remove features once the replacement is ready (and enums are something where I think we all agree on the replacement).

Fair enough, I think we should still deprecate with the same "not supported by MFC" message and folks not using MFC should suppress the warning, but we will not remove.

PR #2932, opened 2023-01-10
Support zero-width integer literals.

dtzSiFive:
These are in the FIRRTL language specification now and supported by both SFC + MFC.

If integer literal width is specified, support zero-width (`0.U(0.W)`), otherwise at least 1-bit (`0.U`), for compatibility with existing code.

Modified `assertKnownWidth`/`assertInferredWidth` to allow checking for width of zero, avoid attempting to use `-1S(0.W)`.  Used these to check basic zero-width behavior of wires/regs being initialized with zero-width integer literals.

Add a few tests to ensure zero-width literals and a few operations/constructs work as expected, and that their behavior is preserved.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### Desired Merge Strategy

#### Release Notes

Add support for zero-width integer literals, which must be explicitly requested (`0.U(0.W)`), for compatibility reasons `0.U` will still be (minimum) 1-bit.

- [x] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/WidthSpec.scala
+++ b/src/test/scala/chiselTests/WidthSpec.scala
@@ -98,18 +98,18 @@ abstract class WireRegWidthSpecImpl extends ChiselFlatSpec {
     }
   }
 
-  it should "infer width as 1-bit if the template type has no width and is initialized to zero-width literal" in {
-    assertInferredWidth(1) {
+  it should "infer width as zero if the template type has no width and is initialized to zero-width literal" in {
+    assertInferredWidth(0) {
       val w = builder(UInt())
       w := 0.U(0.W)
       w
     }
-    assertInferredWidth(1) {
+    assertInferredWidth(0) {
       val w = builder(new ZeroWidthBundle)
       w := ZeroWidthBundle.intoWire()
       w.y
     }
-    assertInferredWidth(1) {
+    assertInferredWidth(0) {
       val w = builder(Vec(1, UInt()))
       w(0) := 0.U(0.W)
       w(0)
--- a/src/test/scala/chiselTests/WidthSpec.scala
+++ b/src/test/scala/chiselTests/WidthSpec.scala
@@ -17,15 +17,30 @@ object SimpleBundle {
   }
 }
 
+class ZeroWidthBundle extends Bundle {
+  val x = UInt(0.W)
+  val y = UInt()
+}
+object ZeroWidthBundle {
+  def intoWire(): ZeroWidthBundle = {
+    val w = Wire(new ZeroWidthBundle)
+    w.x := 0.U(0.W)
+    w.y := 0.U(0.W)
+    w
+  }
+}
+
 class WidthSpec extends ChiselFlatSpec {
   "Literals without specified widths" should "get the minimum legal width" in {
     "hdeadbeef".U.getWidth should be(32)
     "h_dead_beef".U.getWidth should be(32)
     "h0a".U.getWidth should be(4)
     "h1a".U.getWidth should be(5)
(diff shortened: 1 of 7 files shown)


dtzSiFive:
At least some of the CI failures are because current firtool used by CI is not new enough, PR bumping to latest submitted: #2933 .

Looks like really want `0.U` to be 1-bit, will look into making that the case next!

src/test/scala/chiselTests/WidthSpec.scala
```diff
@@ -67,6 +97,24 @@ abstract class WireRegWidthSpecImpl extends ChiselFlatSpec {
...
+      val w = builder(new ZeroWidthBundle)
+      w := ZeroWidthBundle.intoWire()
+      w.y
+    }
+    assertInferredWidth(1) {
+      val w = builder(Vec(1, UInt()))
+      w(0) := 0.U(0.W)
+      w(0)
+    }
+  }
```
jackkoenig, line 117: This seems wrong to me (and super kudos for writing these tests because I would not have noticed this behavior without them).

I would expect `Cat(1.U, 0.U(0.W), 1.U)` to be isomorphic to `Cat(1.U, WireInit(UInt(), 0.U(0.W)), 1.U)`, but this suggests  otherwise, why?
This indeed seems to be the case when I run that actual Chisel to Verilog with both MFC and SFC so I'm confused why there's inferred width == 1?

Is it perhaps a test setup issue? Determining inferred width from Chisel is tricky and maybe it has a bug around zero-width.

jackkoenig (commented):
This looks really good and I especially appreciate all of the tests. There is 1 test that isn't making sense to me but otherwise this LGTM.

src/test/scala/chiselTests/WidthSpec.scala
dtzSiFive: > This seems wrong to me

Me too, honestly, thanks for bringing it up.  I agree with your suggestion it may be a test setup issue (definitely was not 100% on those changes when I made them) and indeed it seems to have been!  Will push a fix soon, happily it was actually really dumb with fresh eyes -- was using a `0.U` which this PR went out of the way to make NOT zero-width (min 1-bit) which was why it wouldn't pass things that actually were zero-width.  Making it `0.U(0.W)` fixes that :+1: .

> I would expect `Cat(1.U, 0.U(0.W), 1.U)` to be isomorphic to `Cat(1.U, WireInit(UInt(), 0.U(0.W)), 1.U)`, but this suggests otherwise, why?

Fantastic test case, thanks!

Is this something suitable for a test as well? Not sure where / how that might work, especially without assuming constant propagation in the compiler and such, but for I really liked that (immediately seemed clear what you were trying to check/determine/say) and so would be great to check moving forward.  If that's not really suitable, the width checks already present are probably enough.  Who tests the testers, indeed.

( I've been testing various behavior using an increasingly modified blink example, so I can replicate your findings re:`2'h3` for both (no surprise, for my own sanity) )

jackkoenig: > Is this something suitable for a test as well? Not sure where / how that might work, especially without assuming constant propagation in the compiler and such, but for I really liked that (immediately seemed clear what you were trying to check/determine/say) and so would be great to check moving forward. If that's not really suitable, the width checks already present are probably enough.

Yeah I have the same thoughts/questions/concerns. I think it would be really handy to be able to express things like this, and perhaps a suite of formal checks for certain properties we expect to hold would be good. I think such a thing is beyond the scope of this PR though--I think the tests you've written are more than sufficient. Nice work 🙂

jackkoenig (approved):
LGTM! Nice work and great tests!

PR #2937, opened 2023-01-12
Deprecate public internals and fix warning suppression 

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

  - code cleanup   

#### API Impact

Public APIs in `chisel3.internal` that never should have been public are being deprecated.

N/A
#### Desired Merge Strategy

#### Release Notes

Internal APIs in package `chisel3.internal` which never should have been public are deprecated.

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,9 +17,9 @@ lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
   version := "3.6-SNAPSHOT",
   autoAPIMappings := true,
+  scalacOptions := Seq("-deprecation", "-feature"),
   scalaVersion := "2.12.17",
   crossScalaVersions := Seq("2.13.10", "2.12.17"),
-  scalacOptions := Seq("-deprecation", "-feature"),
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
   // Macros paradise is integrated into 2.13 but requires a scalacOption
   scalacOptions ++= {
@@ -35,6 +35,8 @@ lazy val commonSettings = Seq(
     }
   }
 )
+
+lazy val warningSuppression = Seq(scalacOptions += "-Wconf:msg=APIs in chisel3.internal:s,msg=Importing from firrtl:s,msg=migration to the MLIR:s")
 
 lazy val publishSettings = Seq(
   versionScheme := Some("pvp"),
@@ -171,6 +173,7 @@ lazy val core = (project in file("core"))
   )
   .settings(publishSettings: _*)
   .settings(mimaPreviousArtifacts := Set())
+  .settings(warningSuppression: _*)
   .settings(
     name := "chisel3-core",
     libraryDependencies ++= Seq(
@@ -178,7 +181,6 @@ lazy val core = (project in file("core"))
       "com.lihaoyi" %% "os-lib" % "0.8.1"
     ),
     scalacOptions := scalacOptions.value ++ Seq(
-      "-deprecation",
       "-explaintypes",
       "-feature",
       "-language:reflectiveCalls",
@@ -202,6 +204,7 @@ lazy val chisel = (project in file("."))
   .dependsOn(macros)
   .dependsOn(core)
   .aggregate(macros, core, plugin)
+  .settings(warningSuppression: _*)
   .settings(
     mimaPreviousArtifacts := Set(),
     libraryDependencies += defaultVersions("treadle") % "test",
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,13 @@ lazy val commonSettings = Seq(
   }
 )
 
-lazy val warningSuppression = Seq(scalacOptions += "-Wconf:msg=APIs in chisel3.internal:s,msg=Importing from firrtl:s,msg=migration to the MLIR:s")
+lazy val warningSuppression = Seq(
+  scalacOptions += "-Wconf:" + Seq(
+    "msg=APIs in chisel3.internal:s",
+    "msg=Importing from firrtl:s",
+    "msg=migration to the MLIR:s"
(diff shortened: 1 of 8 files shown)


core/src/main/scala/chisel3/package.scala
```diff
@@ -361,4 +361,6 @@ package object chisel3 {
 
   final val deprecatedMFCMessage =
     "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
+
+  final val deprecatedPublicAPIMsg = "APIs in chisel3.internal will no longer be public as of 3.6"
```
jackkoenig, line 365:
```suggestion
  final val deprecatedPublicAPIMsg = "APIs in chisel3.internal are not intended to be public"
```
They'll just be deprecated in 3.6, not private until 5 so I think we just warn and don't mention a version for when they'll be removed.

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -103,42 +112,51 @@ sealed trait ConditionalDeclarable extends TopBinding {
...
 }
 // A DontCare element has a specific Binding, somewhat like a literal.
 // It is a source (RHS). It may only be connected/applied to sinks.
+@deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 case class DontCareBinding() extends UnconstrainedBinding
 
 // Views currently only support 1:1 Element-level mappings
-private[chisel3] case class ViewBinding(target: Element) extends UnconstrainedBinding
+@deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
+case class ViewBinding(target: Element) extends UnconstrainedBinding
```
jackkoenig, line 159:
```suggestion
private[chisel3] case class ViewBinding(target: Element) extends UnconstrainedBinding
```
This one is already private, why are you making it public?

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -147,20 +165,25 @@ private[chisel3] case class ViewBinding(target: Element) extends UnconstrainedBi
   * @note The types of key and value need not match for the top Data in a total view of type
   *       Aggregate
   */
-private[chisel3] case class AggregateViewBinding(childMap: Map[Data, Data]) extends UnconstrainedBinding {
+@deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
+case class AggregateViewBinding(childMap: Map[Data, Data]) extends UnconstrainedBinding {
```
jackkoenig, line 169:
```suggestion
private[chisel3] case class AggregateViewBinding(childMap: Map[Data, Data]) extends UnconstrainedBinding {
```

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -28,7 +33,11 @@ object requireIsHardware {
...
 /** Requires that a node is a chisel type (not hardware, "unbound")
   */
+@deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 object requireIsChiselType {
+  def apply(node: Data, msg: String = ""): Unit = _requireIsChiselType.apply(node, msg)
+}
+private[chisel3] object _requireIsChiselType {
   def apply(node: Data, msg: String = ""): Unit = if (node.isSynthesizable) {
     val prefix = if (msg.nonEmpty) s"$msg " else ""
     throw ExpectedChiselTypeException(s"$prefix'$node' must be a Chisel type, not hardware")
```
jackkoenig, line 43: This, and `requireIsHardware` above should just be moved to chisel3.experimental (where there is currently a val alias). Then, instead of objects that replicate all of the arguments, you should just put a deprecated val alias in package object chisel3.internal.

build.sbt
```diff
@@ -36,6 +36,8 @@ lazy val commonSettings = Seq(
   }
 )
 
+lazy val warningSuppression = Seq(scalacOptions += "-Wconf:msg=APIs in chisel3.internal:s,msg=Importing from firrtl:s,msg=migration to the MLIR:s")
```
jackkoenig, line 39:
```suggestion
lazy val warningSuppression = Seq(
  scalacOptions += "-Wconf:" + Seq(
    "msg=APIs in chisel3.internal:s",
    "msg=Importing from firrtl:s",
    "msg=migration to the MLIR:s"
  ).mkString(",")
)
```
I think we should write this in a way that highlights each rule being applied.

adkian-sifive:
Note: SourceInfo will be moved to package experimental in a follow up PR. This is because moving around packages causes the chiseltest integration test CI checks to fail, and we don't currently have the ability to skip this check in CI. The follow up PR will happen after the ability to skip chiseltest is added in a separate PR.

build.sbt
```diff
@@ -17,9 +17,9 @@ lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
   version := "3.6-SNAPSHOT",
   autoAPIMappings := true,
+  scalacOptions := Seq("-deprecation", "-feature"),
   scalaVersion := "2.12.17",
   crossScalaVersions := Seq("2.13.10", "2.12.17"),
-  scalacOptions := Seq("-deprecation", "-feature"),
```
jackkoenig, line 22: Can you remove this unnecessary diff?

jackkoenig (approved):
1 minor request but LGTM otherwise!

build.sbt
jackkoenig: I pushed a commit that fixes it.

jackkoenig:
Also thank you very very much for cleaning up the warnings, it reduced the total number of warnings for a Chisel3 compile from 188 to 59.

PR #2945, opened 2023-01-18
Require AutoCloneType

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

- code cleanup 

#### API Impact

Deprecate `cloneType` for bundles and let the compiler plugin automatically clone bundle fields. Better error message for cases where the plugin is not able to clone types.

#### Desired Merge Strategy

#### Release Notes

Records now have auto cloneType always enabled. `chisel3.experimental.AutoCloneType` is now deprecated.

- [ ] Did you review?


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1246,15 +1268,10 @@ package experimental {
...
-    * All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
+  /** Indicates that the compiler plugin should generate [[clonType]] for this typll user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
     */
-  trait AutoCloneType { self: Record =>
-
-    override def cloneType: this.type = _cloneTypeImpl.asInstanceOf[this.type]
-
-  }
+  @deprecated("AutoCloneType is handled by the plugin", "3.6")
+  trait AutoCloneType
```
jackkoenig, line 1274:
```suggestion
  trait AutoCloneType { self: Record => }
```
No reason to remove the self type

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -937,6 +937,28 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
     case _ => false
   }
 
+  trait AutoCloneType { self: Record =>
+    override def cloneType: this.type = _cloneTypeImpl.asInstanceOf[this.type]
+  }
+
```
jackkoenig, line 943:
```suggestion
```
Unnecessary copy-paste.

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -247,7 +246,7 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
         val isAutoCloneType: Boolean = isAnAutoCloneType(record.symbol)
 
         if (!isAutoCloneType) {
-          warnOnCloneType(record.impl.body)
+          errorOnCloneType(record.impl.body)
           // Other than warning, there is nothing to do on Records that don't mixin AutoCloneType
           return super.transform(record)
         }
```
jackkoenig, line 252: All of this should be deleted because users no longer should be mixing in `AutoCloneType`, it needs to be done automatically even when they don't mix it in.

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -83,12 +83,10 @@ class RecordAutoCloneType[T <: Data](gen: T) extends Record with AutoCloneType {
 
 // Records that don't mixin AutoCloneType should still be able to implement the related methods
 // NOTE: This is a very weird thing to do, don't do it.
-class RecordWithVerbotenMethods(w: Int) extends Record {
+class RecordWithVerbotenMethods(w: Int) extends Record with AutoCloneType {
```
jackkoenig, line 86: See this comment, this test no longer applies with autoclonetype being mandatory, just delete it.

src/test/scala/chiselTests/CompatibilityInteroperabilitySpec.scala
```diff
@@ -17,9 +18,8 @@ object CompatibilityComponents {
     val a = UInt(width = 32)
     val b = UInt(width = 32).flip
   }
-  class ChiselRecord extends Record {
+  class ChiselRecord extends Record with AutoCloneType {
```
jackkoenig, line 21:
```suggestion
  class ChiselRecord extends Record {
```
The idea is to make `AutoCloneType` happen automatically, NOT to require everyone to mix it in.

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -118,13 +118,12 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
+    def errorOnCloneType(body: List[Tree]): Unit = {
       body.foreach {
         case d: DefDef if isNullaryMethodNamed("cloneType", d) =>
-          val msg = "It is no longer necessary to implement cloneType. " +
-            "Mix in chisel3.experimental.AutoCloneType to let the compiler plugin generate it. " +
-            "This will become an error in Chisel 3.6."
-          global.reporter.warning(d.pos, msg)
+          val msg = "cloneType is unsupported starting Chisel 3.6. " +
+            "Mix in chisel3.experimental.AutoCloneType to let the compiler plugin generate it. "
+          global.reporter.error(d.pos, msg)
```
jackkoenig, line 126: This function is no longer needed, the error is up in `getConstructorAndParams`, just use the one there.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -937,6 +937,28 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
...
+  private def checkClone(clone: Record): Unit = {
+    for ((name, field) <- elements) {
+      if (clone.elements(name) eq field) {
+        throw new AutoClonetypeException(
+          s"The bundle plugin was unable to clone $clone that has field '$name' aliased with base $this." +
+            " Try wrapping the field(s) in Input(...), Output(...), or Flipped(...) if appropriate." +
+            " As a last resort, you can override cloneType manually."
+        )
```
jackkoenig, line 951: This error message needs to be updated since users cannot override cloneType manually.

It also needs to give the user some hint for the case when AutoCloneType does not realize there are Data in the arguments to the Bundle or Record which in 3.5.6 gives rebinding exceptions but should be caught by this error. eg. see https://scastie.scala-lang.org/uMPihS2TSxy3sPCYfUb7AA for an example.

jackkoenig: You have a test below for this case, but this error needs to help the user figure out what went wrong.

jackkoenig (changes requested):
This is close but I think there was a slight miscommunication on the idea--it's not that users should be required to mixin `AutoCloneType`, it's that `AutoCloneType` applies to all Records automatically and manually implementing `cloneType` is now banned.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -937,6 +937,28 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
...
+    override def cloneType: this.type = _cloneTypeImpl.asInstanceOf[this.type]
+  }
+
+  private def checkClone(clone: Record): Unit = {
+    for ((name, field) <- elements) {
+      if (clone.elements(name) eq field) {
+        throw new AutoClonetypeException(
+          s"The bundle plugin was unable to clone $clone that has field '$name' aliased with base $this." +
+            " Try wrapping the field(s) in Input(...), Output(...), or Flipped(...) if appropriate." +
+            " As a last resort, you can override cloneType manually."
```
jackkoenig, line 950:
```suggestion
          s"The compiler plugin was unable to clone $clone that has field '$name' aliased with base $this." +
            " Try wrapping the field(s) in Input(...), Output(...), or Flipped(...) if appropriate."
```
It might also be worth suggesting something like "It is best to avoid nesting any Data arguments inside of other data structures" and something like "as a last resort, you can call chisel3.reflect.DataMirror.internal.chiselTypeClone on any nested Data arguments". 

Something like that, but please massage the phrasing 🙂

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1246,15 +1268,10 @@ package experimental {
   class BundleLiteralException(message: String) extends ChiselException(message)
   class VecLiteralException(message: String) extends ChiselException(message)
 
-  /** Indicates that the compiler plugin should generate [[cloneType]] for this type
-    *
-    * All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
+  /** Indicates that the compiler plugin should generate [[clonType]] for this typll user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
     */
```
jackkoenig, line 1272:
```suggestion
```
Since it's deprecated, I think we can just delete the doc (since the doc is also wrong).

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1246,16 +1266,6 @@ package experimental {
 
   class BundleLiteralException(message: String) extends ChiselException(message)
   class VecLiteralException(message: String) extends ChiselException(message)
-
-  /** Indicates that the compiler plugin should generate [[cloneType]] for this type
-    *
-    * All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
-    */
-  trait AutoCloneType { self: Record =>
```
mwachs5, line 1254: are we going to do a deprecating backport of this?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -938,6 +938,26 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
...
 
+  private def checkClone(clone: Record): Unit = {
+    for ((name, field) <- elements) {
+      if (clone.elements(name) eq field) {
+        throw new AutoClonetypeException(
+          s"The bundle plugin was unable to clone $clone that has field '$name' aliased with base $this." +
+            "This likely happened because you tried nesting Data arguments inside of other data structures." +
+            " Try wrapping the field(s) in Input(...), Output(...), or Flipped(...) if appropriate." +
+            " As a last resort, you can call chisel3.reflect.DataMirror.internal.chiselTypeClone on any nested Data arguments." +
+            "See the cookbok entry on this issue for more details."
```
mwachs5, line 949:
```suggestion
            "See the cookbook entry on 'How do I deal with the \"unable to clone\" error' for more details."
```

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
 getVerilogString(new Top(new UsingCloneTypeBundle(UInt(8.W))))
 ```
 
+### <a name="bundle-unable-to-clone"></a> How do I deal with the "unable to clone" error?
+
+Most Chisel objects need to be cloned in order to differentiate between the software representation of the bundle field from its "bound" hardware representation, where "binding" is the process of generating a hardware component. For Bundle fields, this cloning is supposed to happen automatically with a compiler plugin.
```
mwachs5, line 449: nit: can we have more line breaks to help maintaining this text? Break at 80 chars or at punctuation?

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+
+In some cases though, the plugin may not be able to clone the Bundle fields. The most common case for when this happens is when the `chisel3.Data` part of the Bundle field is nested inside some other data structure and the compiler plugin is unable to figure out how to clone the entire structure. It is best to avoid such nested structures.
```
mwachs5, line 451: Isn't this not just about Bundle fields? Isn't this for any Record?

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+ import chisel3._
+
+ class CustomBundleBroken(elts: (String, Data)*) extends Record {
+ 	val elements = ListMap(elts.map {
+ 		case (field, elt) =>
+ 		field -> elt
+ 	}: _*)
+ 	def apply(elt: String): Data = elements(elt)
+ }
+```
```
mwachs5, line 465: can you show a usage that shows the error with scala mdoc:crash?

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+You can use `chiselTypeClone` to clone the elements as:
+
+
+```scala mdoc
+ import chisel3._
+ import chisel3.reflect.DataMirror
+ 
+ class CustomBundleBroken(elts: (String, Data)*) extends Record {
+ 	val elements = ListMap(elts.map {
+ 		case (field, elt) =>
```
mwachs5, line 476: not sure if we want to show requireIsChiselType(elt) here, to show *even if it was already a chisel type, still clone it* ?

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+		field -> DataMirror.internal.chiselTypeClone(elt)
```
mwachs5, line 477: is this what we want vs chiselTypeOf(...) ? I wish we could just settle on one API here. 

In other words, I don't want to actively encourage some weird pattern that people should never be doing in the first place.

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,42 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+```
+
+You can use `chiselTypeClone` to clone the elements as:
+
+
+```scala mdoc
+ import chisel3._
+ import chisel3.reflect.DataMirror
+ 
+ class CustomBundleBroken(elts: (String, Data)*) extends Record {
```
mwachs5, line 474: let's call this CustomBundleBrokenWithWorkaround or CustomBundleFixed

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -118,17 +115,6 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
       (primaryConstructor, paramAccessors.toList)
     }
 
-    def warnOnCloneType(body: List[Tree]): Unit = {
-      body.foreach {
-        case d: DefDef if isNullaryMethodNamed("cloneType", d) =>
-          val msg = "It is no longer necessary to implement cloneType. " +
-            "Mix in chisel3.experimental.AutoCloneType to let the compiler plugin generate it. " +
```
mwachs5, line 125: seems like the backport should change this message if they *do* mix in AutoCloneType?

src/test/scala/chiselTests/AutoClonetypeSpec.scala
```diff
@@ -85,10 +84,8 @@ class RecordAutoCloneType[T <: Data](gen: T) extends Record with AutoCloneType {
 // NOTE: This is a very weird thing to do, don't do it.
 class RecordWithVerbotenMethods(w: Int) extends Record {
```
mwachs5, line 85: does this test still make any sense?

docs/src/cookbooks/cookbook.md
adkian-sifive: I believe this is the right one, I'm not sure if chiselTypeOf clones the type

core/src/main/scala/chisel3/Aggregate.scala
adkian-sifive: I think so, @jackkoenig thoughts?

src/test/scala/chiselTests/AutoClonetypeSpec.scala
adkian-sifive: nice catch, yeah i don't think so

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: Please do not delete the trait, leave it here but deprecate it, something like
```scala
@deprecated("Auto-clonetype is now always on, you no longer need to mix it in, "Chisel 3.6")
trait AutoCloneType { self: Record =>
}

The Chisel 3.5 work is already done (adding autoclonetype to 3.5 sort of is the "backport")

This trait was only added in 3.5.5 and is step 1 in auto clonetype for Records.
Step 2 is upgrading to 3.6.0 where autoclonetype is always enabled and the trait is deprecated.
Step 3 we'll delete the trait in Chisel 5.

File: docs/src/cookbooks/cookbook.md [hunk 20]
```diff
@@ -443,6 +444,68 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+```
+
+```scala mdoc:crash
+ class CustomBundleBroken(elts: (String, Data)*) extends Record {
+ 	val elements = ListMap(elts.map {
+ 		case (field, elt) =>
+ 		field -> elt
+ 	}: _*)
+ 	def apply(elt: String): Data = elements(elt)
+ }
```
jackkoenig, line 479: Please replace the tabs with spaces, 2 spaces per tab.

docs/src/cookbooks/cookbook.md
jackkoenig: chiselTypeOf clones the type, it's the same as chiselTypeClone except it also checks that the type is bound as hardware. In this case, the elts will not be bound hardware so we need chiselTypeClone.

I agree we should settle on one API but they have different uses and figuring out how to not need both uses is subtle and beyond this PR.

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
jackkoenig: Yeah agreed. We should do that by keeping the trait around but deprecating it as mentioned elsewhere.

src/test/scala/chiselTests/AutoClonetypeSpec.scala
jackkoenig: It still does `_usingPlugin` and `_elementsImpl` which are checked for Bundles but not Records so it's exercising that code path.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1249,8 +1269,9 @@ package experimental {
 
   /** Indicates that the compiler plugin should generate [[cloneType]] for this type
     *
-    * All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
+    *  All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
```
jackkoenig, line 1272:
```suggestion
    * All user-defined [[Record]]s should mix this trait in as it will be required for upgrading to Chisel 3.6.
```

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -938,6 +938,26 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
...
 
+  private def checkClone(clone: Record): Unit = {
+    for ((name, field) <- elements) {
+      if (clone.elements(name) eq field) {
+        throw new AutoClonetypeException(
+          s"The bundle plugin was unable to clone $clone that has field '$name' aliased with base $this." +
+            "This likely happened because you tried nesting Data arguments inside of other data structures." +
+            " Try wrapping the field(s) in Input(...), Output(...), or Flipped(...) if appropriate." +
+            " As a last resort, you can call chisel3.reflect.DataMirror.internal.chiselTypeClone on any nested Data arguments." +
+            " See the cookbook entry 'How do I deal with the \"unable to clone\" error' for more details."
```
jackkoenig, line 949:
```suggestion
            " See the cookbook entry 'How do I deal with the \"unable to clone\" error?' for more details."
```

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,67 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+ 
+ class NewModule extends Module {
+   val out = Output(UInt(8.W))
+   val recordType = new CustomBundleBroken("fizz" -> UInt(16.W), "buzz" -> UInt(16.W))
+   val record = Wire(recordType)
+   val uint = record.asUInt
+   val record2 = uint.asTypeOf(recordType)
+   out := record
+ }
+ getVerilogString(new NewModule)
```
jackkoenig, line 487:
```suggestion
class CustomBundleBroken(elts: (String, Data)*) extends Record {
  val elements = ListMap(elts: _*)

  def apply(elt: String): Data = elements(elt)
}

class NewModule extends Module {
  val out = Output(UInt(8.W))
  val recordType = new CustomBundleBroken("fizz" -> UInt(16.W), "buzz" -> UInt(16.W))
  val record = Wire(recordType)
  val uint = record.asUInt
  val record2 = uint.asTypeOf(recordType)
  out := record
}
getVerilogString(new NewModule)
```

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,67 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+
+There are a few ways around this issue - you can try wrapping the problematic
+fields in Input(...), Output(...), or Flipped(...) if appropriate. You can also
+try manually cloning each field in the Bundle using the `chiselTypeClone` method
+in `chisel3.reflect.DataMirror`. Here's an example with the Bundle whose fields
+won't get cloned:
+
+```scala mdoc:invisible
+ import chisel3._
+ import scala.collection.immutable.ListMap
```
jackkoenig, line 469:
```suggestion
import chisel3._
import scala.collection.immutable.ListMap
```

docs/src/cookbooks/cookbook.md
```diff
@@ -443,6 +444,67 @@ class UsingCloneTypeBundle[T <: Data](gen: T) extends Bundle {
...
+
+ class CustomBundleFixed(elts: (String, Data)*) extends Record {
+   val elements = ListMap(elts.map {
+     case (field, elt) =>
+       requireIsChiselType(elt)
+       field -> DataMirror.internal.chiselTypeClone(elt)
+   }: _*)
+
+   def apply(elt: String): Data = elements(elt)
+ }
```
jackkoenig, line 505:
```suggestion
import chisel3.reflect.DataMirror
import chisel3.experimental.requireIsChiselType

class CustomBundleFixed(elts: (String, Data)*) extends Record {
  val elements = ListMap(elts.map {
    case (field, elt) =>
      requireIsChiselType(elt)
      field -> DataMirror.internal.chiselTypeClone(elt)
  }: _*)

  def apply(elt: String): Data = elements(elt)
}
```

jackkoenig (approved):
LGTM! I have a few very nitty suggestions but I will just merge them. Thanks!

PR #2948, opened 2023-01-24
Added deprecation for non-Undefined read-under-write mems

azidar:
It turns out that because there is no syntax for declaring non-undefined read-under-write mems in FIRRTL/CHIRRTL, going through that serialization path actually reverts the behavior to Undefined.

This is an opportunity to revisit adding CHIRRTL to the FIRRTL spec and adding this back in, but until then, we should not be generating this from Chisel.

### Contributor Checklist

- [N/A] Did you add at least one test demonstrating the PR?

#### Type of Improvement

 - removed feature/API

#### API Impact

Users of non-undefined memory behavior will get a warning, and in Chisel 5.0, it will get removed. Likely in the future, we will add back the feature once the backing compiler infra supports it more fully.

None.

#### Release Notes

Deprecated generating memories with non-Undefined read-under-write collision semantics, as they would silently get changed to Undefined r-u-w behavior if going through a CHIRRTL serialization step.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Mem.scala
+++ b/core/src/main/scala/chisel3/Mem.scala
@@ -318,6 +318,9 @@ object SyncReadMem {
     if (compileOptions.declaredTypeMustBeUnbound) {
       requireIsChiselType(t, "memory type")
     }
+    if (ruw != Undefined) {
+      Builder.warning(s"specifying read-under-write behavior other than Undefined is deprecated 3.6 and removed in 5.0")
+    }
     val mt = t.cloneTypeFull
     val mem = new SyncReadMem(mt, size, ruw)
     mt.bind(MemTypeBinding(mem))


core/src/main/scala/chisel3/Mem.scala
```diff
@@ -318,6 +318,9 @@ object SyncReadMem {
     if (compileOptions.declaredTypeMustBeUnbound) {
       requireIsChiselType(t, "memory type")
     }
+    if (ruw != Undefined) {
+      Builder.warning(s"specifying read-under-write behavior other than Undefined is deprecated 3.6 and removed in 5.0")
+    }
```
seldridge, line 323: Leaving this and also deprecating the `apply(Int, T, ReadUnderWrite)` methods would provide a compile-time warning.

seldridge (approved):
LGTM

Can we also deprecate the apply methods that accept a ReadUnderWrite argument?

core/src/main/scala/chisel3/Mem.scala
azidar: Not sure.. if we are planning on adding it back, I'm not sure its worth going through a round of removing the API then adding it back. Do you feel strongly?

seldridge: I don't feel strongly.

azidar: @jackkoenig ?

ekiwi:
It would be great if we could somehow preserve this since some circuits actually are broken if read-under write is undefined. See: https://github.com/freechipsproject/ip-contributions/pull/27

If you want to get rid of the options, then it would be good to at least make the memories WriteFirst or ReadFirst instead of undefined. Whatever matches the behavior of the default Verilog that gets generated by firrtl, since many have come to rely on that and if we simulate true undefined read-under-write (e.g. the way the formal backend does it) many circuits will probably be broken.

azidar:
That's a good point @ekiwi. Making it not undefined but the default behavior seems a sensible non-blocking step forwards.

I think we will need to add this back to Chisel post the chisel5 release because I feel like specifying the r-u-w behavior will be necessary, and I suppose from the perspective of the vast majority of users who used the standard ChiselStage, this feature worked for them previously.

@seldridge do you know what the default emitted r-u-w behavior is?

seldridge:
Testing for FIRRTL memory constructs, not CHIRRTL: SFC emits "new" the same as "undefined". MFC emits everything the same (This is a bug).

It's correct to lock in undefined as the only option and immediately makes SFC, MFC, and the FIRRTL serialization correct.  My primary concern is that this is broken for the serialization format (SFC is broken here, too).  Locking in "new" is problematic as it's not reflected in the serialization and FIRRTL compilers can legally miscompile it.

Mid-term multi-part fix:

1. Define a serialization format for read-under-write for CHIRRTL.
2. Have MFC reject "old".
3. Modify MFC to support "old".
4. Modify Chisel to emit CHIRRTL with read-under-write specified.
5. Add CHIRRTL mems to the spec.
6. Change MFC to emit actual undefined read-under-write.

Tl;dr: I think this will come back.  There's just a number of steps to bring it back all centered around serializing this in the IR.

ekiwi:
I fixed the serialization: https://github.com/chipsalliance/firrtl/pull/2595

Turns out that the `firrtl` parser already supported a serialization format. It was just missing from the Serializer.

seldridge:
Awesome. This should be enough to proceed without deprecating this.

seldridge:
Tracking issue to add this to CIRCT: https://github.com/llvm/circt/issues/4588

azidar:
Thanks @ekiwi and @seldridge , I'll close the ticket.

PR #2951, opened 2023-01-25
Pin firrtl version to 1.6.0-M2

mwachs5:
Using -SNAPSHOT does not make sense, because that means breaking changes in firrtl can break our C/I. At the moment, an upstream change has broken our documentation generation flow.

### Contributor Checklist

#### Type of Improvement

 - bug fix           

yes, it does, but only because upstream broke its generation

#### Desired Merge Strategy

#### Release Notes

Pin firrtl version to 1.6.0-M2

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@
 enablePlugins(SiteScaladocPlugin)
 
 val defaultVersions = Map(
-  "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6-SNAPSHOT",
+  "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6.0-M2",
   "treadle" -> "edu.berkeley.cs" %% "treadle" % "1.6-SNAPSHOT"
   // chiseltest intentionally excluded so that release automation does not try to set its version
   // The projects using chiseltest are not published, but SBT resolves dependencies for all projects
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ enablePlugins(SiteScaladocPlugin)
 
 val defaultVersions = Map(
   "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6.0-M2",
-  "treadle" -> "edu.berkeley.cs" %% "treadle" % "1.6-SNAPSHOT"
+  "treadle" -> "edu.berkeley.cs" %% "treadle" % "1.6.0-M2"
   // chiseltest intentionally excluded so that release automation does not try to set its version
   // The projects using chiseltest are not published, but SBT resolves dependencies for all projects
   // when doing publishing and will not find a chiseltest release since chiseltest depends on


jackkoenig (changes requested):
> Using -SNAPSHOT does not make sense, because that means breaking changes in firrtl can break our C/I. 

On the contrary, that is the point. It is a negative consequence of having closely coupled projects in different repositories, but it is intentional.

mwachs5:
I'm confused. How is this helpful ...? We can always change this back to SNAPSHOT when that's what we want it to be. Right now, nothing will make it through chisel3 CI.

mwachs5:
though this PR is failing C/I in the same way anyway so I'm confused what's going on here.

jackkoenig:
> I'm confused. How is this helpful ...? We can always change this back to SNAPSHOT when that's what we want it to be. Right now, nothing will make it through chisel3 CI.

You noticed the breakage immediately and now we should fix it. It's a way of catching compatibility breaking changes immediately rather than letting them build up until you decide to bump. This is because they are such closely coupled projects, obviously this is not necessarily a development process that every project should follow.

mwachs5:
ok, right, again we can immediately put this back when the issue is resolved. Are we OK blocking on both chisel and CIRCT fixes to do any chisel3 work until this is resolved?

build.sbt
```diff
@@ -3,7 +3,7 @@
 enablePlugins(SiteScaladocPlugin)
 
 val defaultVersions = Map(
-  "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6-SNAPSHOT",
+  "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6.0-M2",
   "treadle" -> "edu.berkeley.cs" %% "treadle" % "1.6-SNAPSHOT"
```
mwachs5, line 7:
```suggestion
  "firrtl" -> "edu.berkeley.cs" %% "firrtl" % "1.6.0-M2",
  "treadle" -> "edu.berkeley.cs" %% "treadle" % "1.6.0-M2"
```

jackkoenig:
It depends on how hard the breakage is to fix but my suspicion is that it's super easy so why not just fix it?

mwachs5:
> is that it's super easy so why not just fix it?

I was told this needs a fix to llvm/CIRCT, so we need to wait for CIRCT to fix the issue then do a release then bump it here

ekiwi:
> I was told this needs a fix to llvm/CIRCT, so we need to wait for CIRCT to fix the issue then do a release then bump it here

We can also do a small fix in firrtl.

I was initially going to only add the RUW type if it is not undefined, but @seldridge asked that I add it unconditionally: https://github.com/chipsalliance/firrtl/pull/2595#discussion_r1086797340

We can make a PR to only conditionally emit it which should fix most of the breakage since using RUW explicitly is rare.

mwachs5:
> We can also do a small fix in firrtl.

I mean, the goal is to move forward towards correctness, so if the current SFC behavior is where we want it to be, let's leave it there and keep pushing forward towards correctness. If no one else is blocked by this CI being broken we can leave this, if we are we can figure out a way to use "the last working version" until we have more pieces that work together...

ekiwi:
here you go: https://github.com/chipsalliance/firrtl/pull/2597

ekiwi:
> I mean, the goal is to move forward towards correctness, so if the current SFC behavior is where we want it to be, let's leave it there and keep pushing forward towards correctness.

Either behavior is correct. Just one is always explicit and the other leaves out the default when possible.

jackkoenig:
I see, I agree with @ekiwi's change because breaking Chisel with all existing releases of firtool and SFC is ill-advised unless absolutely necessary. In cases where it is necessary, then I agree with @mwachs5's proposal to temporarily pin the version to keep making progress in the meantime. https://github.com/chipsalliance/firrtl/pull/2591 would also help for such things 😇

mwachs5:
closing since SFC fix should get -SNAPSHOT working well again

PR #2956, opened 2023-01-27
Fix FIRRTL direction calculations for Vec and OpaqueType

jackkoenig:
I noticed that `OpaqueTypes` do not get the proper direction in the output FIRRTL. When debugging why this was the case, I realized that it's the same fundamental issue as my nemesis bug...

The bug that will never die, see https://github.com/chipsalliance/chisel3/pull/945, https://github.com/chipsalliance/chisel3/issues/1063, https://github.com/chipsalliance/chisel3/issues/1192, and https://github.com/chipsalliance/chisel3/pull/1196 (I would not be surprised if there are more that my quick search missed).

For years, Chisel always seemed to have bugs around whether you put direction operators inside or outside of a Vec, `Output(Vec(2, ...))` vs. `Vec(2, Output(...))`. Our fixes to this in the past were piecemeal, never addressing the true root cause, and it turns out... **it's still broken**.

The solution for top-level Vec ports was not properly recursive, so for nested Vecs, it's still broken: https://scastie.scala-lang.org/rdiDRQIqRrqL2OZbdRsRIw
We also never applied that solution to BlackBoxes, so even the simple case is broken there: https://scastie.scala-lang.org/vuMe2HxNSJCbkXWcglOQ0A

It also turns out, that the correct recursive solution already exists and was being used for calculating flippedness of Vec elements _inside_ of Records, so the fix is to just use the same function for ports. This function is then the right place to handle OpaqueTypes as it's basically the same issue of having more than 1 Chisel direction that need to map to a single direction (or flip) in FIRRTL.

Given my history with this bug, I'm far too humbled to think this is the last time we'll deal with it, but a boy can dream.

### Contributor Checklist

#### Type of Improvement

 - bug fix 

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix calculation of direction and Bundle field flippedness for Vecs and OpaqueTypes.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -351,7 +351,7 @@ private[chisel3] object Converter {
   }
 
   def convert(port: Port, topDir: SpecifiedDirection = SpecifiedDirection.Unspecified): fir.Port = {
-    val resolvedDir = SpecifiedDirection.fromParent(topDir, port.dir)
+    val resolvedDir = SpecifiedDirection.fromParent(topDir, firrtlUserDirOf(port.id))
     val dir = resolvedDir match {
       case SpecifiedDirection.Unspecified | SpecifiedDirection.Output => fir.Output
       case SpecifiedDirection.Flip | SpecifiedDirection.Input         => fir.Input
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -307,6 +307,8 @@ private[chisel3] object Converter {
   private def firrtlUserDirOf(d: Data): SpecifiedDirection = d match {
     case d: Vec[_] =>
       SpecifiedDirection.fromParent(d.specifiedDirection, firrtlUserDirOf(d.sample_element))
+    case d: Record if d._isOpaqueType =>
+      SpecifiedDirection.fromParent(d.specifiedDirection, firrtlUserDirOf(d.elementsIterator.next))
     case d => d.specifiedDirection
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -351,7 +353,7 @@ private[chisel3] object Converter {
   }
 
   def convert(port: Port, topDir: SpecifiedDirection = SpecifiedDirection.Unspecified): fir.Port = {
-    val resolvedDir = SpecifiedDirection.fromParent(topDir, port.dir)
+    val resolvedDir = SpecifiedDirection.fromParent(topDir, firrtlUserDirOf(port.id))
     val dir = resolvedDir match {
```
jackkoenig, line 357: For reviewers, this slight change is the entire fix for Vecs, deleting the logic in RawModule is just a cleanup since that code is no longer necessary (it was not recursive and thus incomplete anyway).

The change to `firrtlUserDirOf` above combined with this change fixes the bug for `OpaqueTypes`.

mwachs5 (commented):
this looks good, can you add some tests on the tests for hierarchy where i had issues before with the cloning to the Module clone?

seldridge (approved):
LGTM

I'm glad this was already solved elsewhere.

mwachs5:
> this looks good, can you add some tests on the tests for hierarchy where i had issues before with the cloning to the Module clone?

Sorry for the drive-by comment. I was thinking of these: https://github.com/chipsalliance/chisel3/blob/master/src/test/scala/chiselTests/CompatibilityInteroperabilitySpec.scala#L428 but these are more interoperability, so maybe not relevant comment.

mwachs5 (approved):

PR #2958, opened 2023-01-27
More Circt intrinsic wrappers (IsX, PlusArgsTest, PlusArgsValue)

darthscsi:
### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This extends the set of circt intrinsics with Chisel wrappers.  These have no impact on code not using them.

N/A

#### Desired Merge Strategy

#### Release Notes

Add support for Circt intrinsics.

- [ ] Did you review?


src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,49 @@
...
+package chisel3.util.circt
+
+import chisel3._
+import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule}
+
+import circt.Intrinsic
+
+// We have to unique designedName per type, be we can't due to name conflicts
+// on bundles.  Thus we use a globally unique id.
+private object IsXGlobalIDGen {
```
ekiwi, line 12: @jackkoenig What happened to putting this cache into the builder?

jackkoenig: @darthscsi I agree with Kevin here, please do what I suggested here: https://github.com/chipsalliance/chisel3/pull/2900#discussion_r1085901366

All of these ids could live in a single mutable HashMap that keys by their "id namespace"

src/test/scala/chiselTests/util/IsXSpec.scala
```diff
@@ -0,0 +1,56 @@
+package chiselTests.util
```
mwachs5, line 1: license header plz

src/main/scala/chisel3/util/circt/IsX.scala
darthscsi: I've reused an existing mechanism now.  It would be nice if one could use instantiate directly, but that is limited to RawModule derived types.

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,37 @@
...
+  * is a verilog 'x'.
+  */
+private class IsXIntrinsic[T <: Data](gen: T) extends ExtModule {
+  val i = IO(Input(gen))
+  val found = IO(Output(UInt(1.W)))
+  annotate(new ChiselAnnotation {
+    override def toFirrtl =
+      Intrinsic(toTarget, "circt.isX")
+  })
+  override val desiredName = "IsX_" + Builder.idGen.next
```
jackkoenig, line 21: Don't use `idGen`, as you note in the tests below, it is shared between Chisel runs. You should do basically exactly what https://github.com/chipsalliance/chisel3/pull/2717 does in it's changes to the Builder. Add a new `mutable.HashMap[String, Long]` which would let you have a separate namespace for each of these ExtModules.

Then in the function you put in `Builder`, you can have that return the fresh id for the particular namespace and update the underlying mutable HashMap.

src/main/scala/chisel3/util/circt/IsX.scala
ekiwi: @jackkoenig is talking about something different. This is only about how unique IDs are generated in the case where multiple Chisel circuits are elaborated in parallel in the same process. It would probably best if the two of you had a short chat to sort this out. As @mwachs5 mentioned, the current approach could lead to non-stable Chirrtl output in the case of parallel elaborations.

jackkoenig:
I've opened a PR that I think will make it much easier to cleanup the 1 issue with this PR: https://github.com/chipsalliance/chisel3/pull/3009

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,37 @@
...
+import chisel3._
+import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule}
+import chisel3.internal.Builder
+
+import circt.Intrinsic
+
+/** Create a module with a parameterized type which returns whether the input
+  * is a verilog 'x'.
+  */
+private class IsXIntrinsic[T <: Data](gen: T) extends ExtModule {
```
mwachs5, line 14: should this use the new `IntrinsicModule` now instead of `ExtModule`?

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,37 @@
...
+  })
+  override val desiredName = "IsX_" + Builder.idGen.next
+}
+
+object IsX {
+
+  /** Creates an intrinsic which returns whether the input is a verilog 'x'.
+    *
+    * @example {{{
+    * b := isX(a)
```
mwachs5, line 29:
```suggestion
    * b := IsX(a)
```

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
```diff
@@ -0,0 +1,41 @@
...
+package chisel3.util.circt
+
+import chisel3._
+import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule}
+import chisel3.internal.Builder
+
+import circt.Intrinsic
+
+/** Create a module with a parameterized type which calls the verilog function
+  * $test$plusargs to test for the existance of the string str in the
```
mwachs5, line 12:
```suggestion
  * $test$plusargs to test for the existence of the string str in the
```

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
```diff
@@ -0,0 +1,41 @@
...
+  * simulator command line.
+  */
+private class PlusArgsTestIntrinsic[T <: Data](gen: T, str: String) extends ExtModule(Map("FORMAT" -> str)) {
```
mwachs5, line 15: why is gen: T , where T is any data, needed here? Should it just be a Bool type? can you show an `@example` of something where it fills out a useful Data of some arbitrary type here?

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
```diff
@@ -0,0 +1,41 @@
...
+  val found = IO(Output(UInt(1.W)))
```
mwachs5, line 16:
```suggestion
  val found = IO(Output(Bool))
```
? is there a reason to have this as 1-bit uint vs just a Bool?

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,46 @@
...
+import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule, FlatIO}
+import chisel3.internal.Builder
+
+import circt.Intrinsic
+
+/** Create a module which generates a verilog $plusargs$value.  This returns a
+  * value as indicated by the format string and a flag for whether the value
+  * was found.
+  */
+private class PlusArgsValueIntrinsic[T <: Data](gen: T, str: String) extends ExtModule(Map("FORMAT" -> str)) {
```
mwachs5, line 15: does it make sense to take any Data here, vs just a UInt?

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,46 @@
...
+  })
+  override val desiredName = "PlusArgsValue_" + Builder.idGen.next
+}
+
+object PlusArgsValue {
+
+  /** Creates an intrinsic which calls $test$plusargs.
+    *
+    * @example {{{
+    * b := PlusArgsValue(UInt<32.W>, "FOO=%d")
```
mwachs5, line 32:
```suggestion
    * b := PlusArgsValue(UInt(32.W), "FOO=%d")
```

mwachs5: what is the point of exposing to the user the `Foo=%d` vs just the string foo? is it to allow different string formats?

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,46 @@
...
+    * b.found
+    * b.value
+    * }}}
+    */
+  def apply[T <: Data](gen: T, str: String) = {
```
mwachs5, line 37: so is the intrinsic handling the unpacking into some complicated data type from the `%d` which is going to just be an integer? Is this useful vs making that happen explicitly in chisel and make this intrinsic just return a UInt?

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+private class PlusArgsValueTop extends Module {
+  val io = IO(new Bundle {
+    val wf = Output(UInt(1.W))
+    val wv = Output(UInt(32.W))
+    val xf = Output(UInt(1.W))
+    val xv = Output(UInt(32.W))
+  })
+  val tmpw = PlusArgsValue(UInt(32.W), "FOO=%d")
+  val tmpx = PlusArgsValue(io.xv, "BAR=%d")
```
mwachs5, line 23: if we are going to support complicated bundle types then please show a test of that here... but I'm not sure the advantage of having it take any Data vs just a UInt (or SInt possibly through some API...?)

mwachs5, line 23: and if the format string is user controlled please show some tests/applications of things that are not %d

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
darthscsi: Is there a particular reason to?  I've heard rumors of Bool v.s. UInt(1.W) fights, but I haven't payed them much heed.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: yes, you can have fancy string formats in verilog.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: The intrinsic returns any verilog type which can be parsed with format strings (and is supported from firrtl).  The format string can be used in verilog to pick apart complex specification strings for subfields you are interested in.

darthscsi: The intrinsic is nothing more than a wrapper for verilog-spec functionality.

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
darthscsi: To what end?  The string is passed by chisel verbatim to firrtl.  These are not end-to-end verilog generation tests.

darthscsi: Here, for example is parsing a non-integer substring.  We can't express this in firrtl since we don't have real types, but in verilog: `$value$plusargs("FREQ+%0F",frequency)`

you can also control integer encoding in parsing, parse strings, etc.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: you could be reading a string or binary representation of a struct.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,42 @@
...
+/** Create a module which generates a verilog \$plusargs\$value.  This returns a
+  * value as indicated by the format string and a flag for whether the value
+  * was found.
+  */
+private class PlusArgsValueIntrinsic[T <: Data](gen: T, str: String)
+    extends IntrinsicModule("circt.plusargs.value", Map("FORMAT" -> str)) {
+  val io = FlatIO(new Bundle {
+    val found = Output(UInt(1.W))
+    val result = Output(gen)
+  })
```
seldridge, line 20: You can directly use `val found = Output(UInt(1.W))` instead of passing this through `FlatIO`. That really only becomes necessary when dealing with a `BlackBox` which dynamically checks that there is a `val io`.

Same nit as above, `Bool` is likely better for `found`.

For `result`, is this appropriate to take a type parameter? It seems like we wouldn't know what to do with this if it was anything other than a `UInt`? Would this be better expressed as a `UInt` of fixed width (for now) or parametric width (later)?

src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -2,37 +2,17 @@
...
-}
-
 /** Create a module with a parameterized type which returns the size of the type
   * as a compile-time constant.  This lets you write code which depends on the
   * results of type inference.
   */
-private class SizeOfIntrinsic[T <: Data](gen: T) extends ExtModule {
-  val i = IO(Input(gen));
+private class SizeOfIntrinsic[T <: Data](gen: T) extends IntrinsicModule("circt.sizeof") {
+  val i = IO(Input(gen))
```
seldridge, line 14: Way cleaner than the explicit annotation. 👍

src/test/scala/chiselTests/util/IsXSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+  io.outx := IsX(io.x)
+  io.outy := IsX(io.y)
+}
+
+class IsXSpec extends AnyFlatSpec with Matchers {
+  it should "Should work for types" in {
+    val fir = ChiselStage.emitCHIRRTL(new IsXTop)
+    println(fir)
+    (
+      (fir.split('\n').map(x => x.trim) should contain).allOf(
```
seldridge, line 40: Mega-nit / FYI. You can use a `_` here if you so choose:

```suggestion
      (fir.split('\n').map(_.trim) should contain).allOf(
```

seldridge (commented):
Nice to see this getting filled out!

My only major question is on the use of a type parameter inside the plus arg intrinsic. Otherwise, looks good.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
mwachs5: I think @seldridge meant `val found = IO(Output(Bool())` (edit mine on the Bool, but the point is you do still need the `IO`, but you're allowed to have as many `IO` calls as you want.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
mwachs5: right, I am echoing Schuyler's comment above that I'm unclear why this can return anything other than a UInt at this level. Users can easily cast from UInt to their appropriate data type in a safer way than this happening automatically?

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
mwachs5: can you show that being used in the test? 

how would we cast something with format string "FREQ+%0F" to a `gen: DecoupledIO(UInt(32.W))`, something the API will happily let us do right now?

mwachs5: This is a user API we are designing. I am trying to understand the utility of exposing this string to the user, as well as the utility of allowing arbitrary gen data types. If we have no use for it, and users are only ever going to reasonably get UInts back, then let's remove the ambiguity int he API. If we are saying "sure, you can format your results as a float, or signed number" then what do we expect the users to assign those to?

mwachs5: I guess we are trying to formalize https://github.com/chipsalliance/rocket-chip/blob/master/src/main/resources/vsrc/plusarg_reader.v, which does expose the format string, but not the output data type (which is hard-coded to a UInt of a user-specified width).

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,32 @@
...
+import chisel3.internal.Builder
+
+import circt.Intrinsic
+
+/** Create a module with a parameterized type which returns whether the input
+  * is a verilog 'x'.
+  */
+private class IsXIntrinsic[T <: Data](gen: T) extends IntrinsicModule("circt.isX") {
+  val i = IO(Input(gen))
+  val found = IO(Output(UInt(1.W)))
```
mwachs5, line 16:
```suggestion
  val in = IO(Input(gen))
  val isx = IO(Output(Bool))
```

or are these `i` and `found` tied to the intrinsic as coded in `circt.isX`?

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
```diff
@@ -0,0 +1,37 @@
...
+    * }}}
+    */
+  def apply[T <: Data](gen: T, str: String): Data = {
+    if (gen.isSynthesizable) {
+      val inst = Module(new PlusArgsTestIntrinsic(chiselTypeOf(gen), str))
+      inst.found
+    } else {
+      val inst = Module(new PlusArgsTestIntrinsic(gen, str))
+      inst.found
+    }
```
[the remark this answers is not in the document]
mwachs5: yeah we've generally leaned towards making it the user's problem to pass the right thing in (type vs hw)

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
mwachs5: and we could still take a gen: UInt, and the user controls the width. Alternatively the old plus arg reader just forced you to specify the width and then it made the UInt with that width, no gen.

src/test/scala/chiselTests/util/IsXSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+
+private class IsXTop extends Module {
+  val io = IO(new Bundle {
+    val w = Input(UInt(65.W))
+    val x = Input(new IsXBundle)
+    val y = Input(UInt(65.W))
+    val outw = UInt(1.W)
+    val outx = UInt(1.W)
+    val outy = UInt(1.W)
+  })
```
mwachs5, line 29: super nit but mixing the `Input` and "missing Output means 'Aligned'` is making my teeth hurt a bit:

```suggestion
  val io = IO(new Bundle {
    val w = Flipped(UInt(65.W))
    val x = Flipped(new IsXBundle)
    val y = Flipped(UInt(65.W))
    val outw = UInt(1.W)
    val outx = UInt(1.W)
    val outy = UInt(1.W)
  })
```

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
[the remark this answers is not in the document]
darthscsi: It seemed nicer to be able to pass in the kind of thing you are storing the result in rather than having to get it's type yourself.  That way you can do a:
```
Wire w : UInt<4>
w <= PlusArgsTest(w, "FOO")
```

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: FlatIO was an old artifact of extmodule.  Fixed.

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
darthscsi: We are not trying to formalize that.  We are trying to provide access to functionality described in SV 2012 section 21.6.  That that blackbox will be directly implementable is a (intended) side effect.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: I take that back, this gives me a bundle for scala purposes but  flat io in the module.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: For example:
```
// Generated by CIRCT sifive/1/25/0-469-ge8de64f62-dirty
module PlusArgsValueTop(
  input         clock,
                reset,
  output        io_wf,
  output [31:0] io_wv,
  output        io_xf,
  output [15:0] io_xv_a,
                io_xv_b
);

  logic [31:0]                                     _pargs;
  struct packed {logic [15:0] a; logic [15:0] b; } _pargs_0;
  logic _pf, _pf_0;
  initial begin
    _pf = $value$plusargs("FOO=%d", _pargs);
    _pf_0 = $value$plusargs("BAR=%d", _pargs_0);
  end
  assign io_wf = _pf;
  assign io_wv = _pargs;
  assign io_xf = _pf_0;
  assign io_xv_a = _pargs_0.a;
  assign io_xv_b = _pargs_0.b;
endmodule

```

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
darthscsi: I care about exposing verilog functionality via an intrinsic, not about doing exactly what one code bases blackbox does.

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
darthscsi: A missing point is it is not the case that the format has to be "STR=FORMAT", it can be any pattern match.

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
mwachs5: Now that I understand that this is meaningful in verilog, I am OK with it. We might want to add something to the scaladoc that in terms of order / unpacking it's equivalent to using a UInt and then .asTypeOf(gen).

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object IsX {
+
+  /** Creates an intrinsic which returns whether the input is a verilog 'x'.
+    *
+    * @example {{{
+    * b := IsX(a)
+    * }}}
+    */
+  def apply[T <: Data](gen: T): Data = {
```
mwachs5, line 27:
```suggestion
  def apply[T <: Data](gen: T): Bool = {
```

src/main/scala/chisel3/util/circt/IsX.scala
```diff
@@ -0,0 +1,32 @@
...
+    val inst = Module(new IsXIntrinsic(chiselTypeOf(gen)))
+    inst.i := gen
+    inst.found
```
mwachs5, line 30:
```suggestion
    inst.found(0)
```

If it's important to keep the intrinsic module type aligned with firrtl as UInt(1.W), that's fine, but we can Bool-ify it here. At least we should not be returning `Data` from this apply method.

src/main/scala/chisel3/util/circt/PlusArgsTest.scala
```diff
@@ -0,0 +1,35 @@
...
+    * }}}
+    */
+  def apply[T <: Data](gen: T, str: String): Data = {
+    Module(if (gen.isSynthesizable) {
+      new PlusArgsTestIntrinsic(chiselTypeOf(gen), str)
+    } else {
+      new PlusArgsTestIntrinsic(gen, str)
+    }).found
+  }
+}
```
mwachs5, line 35:
```suggestion
  def apply[T <: Data](gen: T, str: String): Bool = {
    Module(if (gen.isSynthesizable) {
      new PlusArgsTestIntrinsic(chiselTypeOf(gen), str)
    } else {
      new PlusArgsTestIntrinsic(gen, str)
    }).found(0)
  }
}
```

same comment as above

mwachs5: ah didnt realize you already changed found to a Bool, editing suggestion above

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Creates an intrinsic which calls \$test\$plusargs.
+    *
+    * @example {{{
+    * b := PlusArgsValue(UInt(32.W), "FOO=%d")
+    * b.found
+    * b.value
+    * }}}
+    */
+  def apply[T <: Data](gen: T, str: String) = {
```
mwachs5, line 33: can we add this API?

```suggestion
  /** Return the value of a plus arg if set, otherwise return the default value.
    */
  def apply[T <: Data](gen: T, str: String, default: T): T = {
    val result = apply(gen, str)
    Mux(result.found, result.value, default)
  }
```

src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -0,0 +1,40 @@
...
+    Module(if (gen.isSynthesizable) {
+      new PlusArgsValueIntrinsic(chiselTypeOf(gen), str)
+    } else {
+      new PlusArgsValueIntrinsic(gen, str)
+    }).io
+  }
```
mwachs5, line 39: Can we add this API:

```suggestion
  }
  
  /** Return the value of a plus arg if set, otherwise return the default value.
    */
  def apply[T <: Data](gen: T, str: String, default: T): T = {
    val result = apply(gen, str)
    Mux(result.found, result.value, default)
  }
```

src/test/scala/chiselTests/util/PlusArgsValueSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+  wv := tmpw.result
+  xf := tmpx.found
+  xv := tmpx.result
+}
+
+class PlusArgsValueSpec extends AnyFlatSpec with Matchers {
+  it should "Should work for types" in {
+    val fir = ChiselStage.emitCHIRRTL(new PlusArgsValueTop)
+    println(fir)
+    (fir.split('\n').map(_.trim) should contain).inOrder(
```
seldridge, line 34: The PR can do a `takeWhile(_ != '@')` to strip the source locators before the trim. (Generally it's bad to match on the source locator.)

seldridge (approved):
LGTM

This will help immensely with writing assertions without compiler escape hatches.

PR #2964, opened 2023-01-31
[website] Make Contributors fetching more robust

jackkoenig:
Things seem to randomly fail when looking up the contributors. This will provide more information when that occurs and also fix what I suspect to be the cause (we had the token environment variable wrong).

### Contributor Checklist

#### Type of Improvement

  - bug fix 

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

- [ ] Did you review?

### DIFF
--- a/website/project/Contributors.scala
+++ b/website/project/Contributors.scala
@@ -22,7 +22,7 @@ object Contributors {
     JavaNetClientBuilder[IO](blocker).create // use BlazeClientBuilder for production use
   }
 
-  val token: Option[String] = sys.env.get("GITHUB4S_ACCESS_TOKEN")
+  val token: Option[String] = sys.env.get("GITHUB_TOKEN")
 
   def contributors(repo: GitHubRepository): List[User] =
     Github[IO](httpClient, token)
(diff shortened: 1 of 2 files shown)


jackkoenig:
Well the good news is that this showed the purpose of the change because this failed on the step in question with 
```
Error:  java.lang.Exception: Unable to fetch contributors for github.com:chipsalliance/chisel3. Did you misspell it? Did the repository move? Is access token defined: false? Original exception: API rate limit exceeded for 172.177.213.82. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)```

So now the problem is the "fix" for the token didn't work, need to figure that out!

File: website/project/Contributors.scala [hunk 1]
```diff
@@ -22,15 +22,18 @@ object Contributors {
...
+  val token: Option[String] = sys.env.get("GITHUB_TOKEN")
 
   def contributors(repo: GitHubRepository): List[User] =
     Github[IO](httpClient, token)
       .repos
       .listContributors(repo.owner, repo.repo)
       .unsafeRunSync()
       .result match {
-        case Left(e) => throw new Exception(s"Unable to fetch contributors for ${repo.serialize}. Did you misspell it? Did the repository move?")
+        case Left(e) => throw new Exception(
```
mwachs5, line 33: do we really need this to be a hard error? Could we just stick something in like "failed to grab the contributors for <...>, will try again on next website rebuild!"

jackkoenig: That shouldn't be necessary now that we will properly be using the token. Besides, if we did make it silently succeed and we broke it, we would never notice because odds are we'll never look and see that warning.

mwachs5: yeah your fix is better

mwachs5 (approved):

PR #2980, opened 2023-02-07
Move prefix and noprefix from internal to experimental

jackkoenig:
Previously, they were aliases into experimental but this does not work very well for ScalaDoc.

### Contributor Checklist

#### Type of Improvement

   - code refactoring

#### API Impact

This is a binary API breakage, but it is source compatible.

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Move implementation of `prefix` and `noprefix` from `chisel3.internal` to `chisel3.experimental`. They were previously aliased in experimental which does not work very well in the generated API docs.

- [ ] Did you review?


mwachs5:
I think there are some docs with this in the naming cookbook? Let's make sure they don't show the now-deprecated version

jackkoenig:
> I think there are some docs with this in the naming cookbook? Let's make sure they don't show the now-deprecated version

The docs already import from `chisel3.experimental.prefix` because we were aliasing from there back to internal--now the alias is reversed.

This is failing CI because it is a binary incompatible change and thus the published chiseltest artifact does not work. We have the label to skip chiseltest CI but the problem is it will still fail on the push to master and thus new SNAPSHOTs will not be published. I could change this to not use an alias but rather use a deprecated object in package internal that has it's own apply method that just invokes the version in experimental. I'll do this tomorrow.

chick (approved):
lgtm

jackkoenig:
I cannot figure out how to avoid the binary compatibility issue. Keeping around the old object doesn't seem to do it so I'm at a loss. I'll just stick with the deprecated alias and manually publish SNAPSHOTs to deal with the integration test crash.

mwachs5:
I'm confused about the binary compatibility issue, doesn't that check only exist on backport branches?

PR #2987, opened 2023-02-09
Deprecate forceName Component APIs, Transform

seldridge:
Deprecate forceName APIs on components.  These are only supported on modules/instances in the MFC.  (This could be revived in the future if desired.)

Deprecate the ForceNamesTransform.  This doesn't need to be public given that the forceName Chisel API is available to users.

This PR is stacked on #2986.


jackkoenig (approved):

mwachs5:
oh now i see this can dismiss my question on #2988

PR #2988, opened 2023-02-09
Convert ForceNamesSpec to use MFC

seldridge:
Change the ForceNamesSpec, a test of the forceName Chisel API, to use the MFC instead of the SFC.  This requires deleting all tests forcing component names as this is not supported by the MFC.

This PR is stacked on https://github.com/chipsalliance/chisel3/pull/2987 which deprecates the APIs whose tests are removed here.

### DIFF
--- a/src/main/scala/chisel3/util/experimental/ForceNames.scala
+++ b/src/main/scala/chisel3/util/experimental/ForceNames.scala
@@ -2,6 +2,7 @@
 
 package chisel3.util.experimental
 
+import chisel3.deprecatedMFCMessage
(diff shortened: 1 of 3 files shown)


jackkoenig (approved):

mwachs5 (commented):
have we deprecated / give a runtime deprecation error for the functionalities that don't work?

seldridge:
> have we deprecated / give a runtime deprecation error for the functionalities that don't work?

Yep! Added in the previous PR in this stack.

PR #2991, opened 2023-02-09
Convert ExtModuleImplSpec to use MFC

seldridge:
Modify the ExtModuleImplSpec test to use the MFC and not the SFC.  This includes one disabled test until firtool 1.30.0 is released.  There is a silly bug where a supported Annotation (BlackBoxPathAnno) is supported but rejected during parsing. See:

  - https://github.com/llvm/circt/commit/0285a98d96b8df898e02c5ed9528f869bff80dcf


jackkoenig (approved):

seldridge:
I will open a separate PR to re-enable the test.

PR #2997, opened 2023-02-10
Publish chisel3.std

ZenithalHourlyRate:
### Contributor Checklist

#### Type of Improvement

bug fix

chisel3.std was not published, and downstream users want to use this, e.g. [rocket-chip](https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/util/BarrelShifter.scala).

#### API Impact

Publish chisel3.std

#### Desired Merge Strategy

Squash

#### Release Notes

Publish chisel3.std

- [ ] Did you review?


sequencer (approved):
Nice catch. I was always compiling from source, and never detected this before.

jackkoenig (changes requested):
This does not correctly publish stdlib.

1. You need to set the `name` key of this project or this just clobbers the normal chisel3 publishing (it thinks it is the project named `chisel3` and thus publishes to the exact same artifact name as `chisel3`). 
2. It is not included when running `+publishLocal`, please figure out how to make it publish the standard way, you currently have to run `standardLibrary / publishLocal` to get it to publish.

ZenithalHourlyRate:
`standardLibrary` depending on `chisel` is troublesome (more specifically, it uses `chisel3.util`) and could result in cyclic dependency if we want to publish it to `stardardLibrary` in `chisel` by a single `+publish` (aggregate standardLibrary)

There are three ways to resolve the issue

1. Create a `chisel3-util` package and make standardLibrary depending on this, like this commit. However, `chisel` package is not the root package anymore and we need `sbt chisel/xxx`, which is unacceptable.
2. Move the current `src` into `util/src` or some other path and then the (virtual) chisel package is the root package. This is a big change in directory structure.
3. Publish chisel3-std as a standalone package

ZenithalHourlyRate:
Yet another easier way to resolve the issue is to move `stdlib/src` into `src`.

ZenithalHourlyRate:
Since caf0518617401 we can remove the CI check on standardLibrary/test

sequencer:
I think split a stdlib to `chisel-stdlib` is good choice, since we never released which(only those build from source(like me))

hakan-demirli:
Any update on this? I need the [BarrelShifter](https://github.com/chipsalliance/chisel/blob/b6c0411dd2f1f65c7da84a74282285ab91c532e2/stdlib/src/main/scala/chisel3/std/BarrelShifter.scala#L27). Preferably without duplicating the code in my repo.
```scala
import chisel3.std //  `object std is not a member of package chisel`
```

PR #3008, opened 2023-02-14
Deprecate chisel3.stage.ChiselStage

seldridge:
Deprecate existing ChiselStage class and object as this will be removed in Chisel 5.  Add a message telling people to switch to the equivalent circt class/object.

I'm preparing to remove this on the `chisel5` branch, hence the deprecation. This would be ideal to get this into Chisel 3.6.

### DIFF
--- a/src/main/scala/chisel3/stage/package.scala
+++ b/src/main/scala/chisel3/stage/package.scala
@@ -12,6 +12,8 @@ import scala.annotation.nowarn
 
 package object stage {
 
+  final val pleaseSwitchToCIRCT = deprecatedMFCMessage + " Please switch to circt.stage.ChiselStage."
+
   @nowarn("cat=deprecation&msg=WarnReflectiveNamingAnnotation")
   implicit object ChiselOptionsView extends OptionsView[ChiselOptions] {
 
--- a/docs/src/cookbooks/hierarchy.md
+++ b/docs/src/cookbooks/hierarchy.md
@@ -85,7 +85,7 @@ class AddTwoInstantiate(width: Int) extends Module {
 }
 ```
 ```scala mdoc:verilog
-chisel3.stage.ChiselStage.emitVerilog(new AddTwoInstantiate(16))
+circt.stage.ChiselStage.emitSystemVerilog(new AddTwoInstantiate(16))
 ```
 
 ## How do I access internal fields of an instance?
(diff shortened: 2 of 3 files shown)


seldridge:
Ack. This deprecates the wrong ChiselStage. 😂

src/main/scala/circt/stage/ChiselStage.scala
```diff
@@ -13,6 +13,7 @@ import firrtl.{AnnotationSeq, EmittedVerilogCircuitAnnotation}
   *
   * @note The companion object, [[ChiselStage$]], has a cleaner API for compiling and returning a string.
   */
+@deprecated(pleaseSwitchToCIRCT, "Chisel 5")
```
jackkoenig, line 16:
```suggestion
@deprecated(pleaseSwitchToCIRCT, "Chisel 3.6")
```
Deprecated version is "since"

src/main/scala/circt/stage/package.scala
```diff
@@ -12,6 +12,9 @@ import java.io.File
 
 package object stage {
 
+  final val pleaseSwitchToCIRCT =
+    "this way of running Chisel is going away as part of the migration to the MLIR-based FIRRTL Compiler (MFC). Please switch to circt.stage.ChiselStage."
+
```
jackkoenig, line 17: I think having the more precise message is good, but can this use or share a base message with `deprecatedMFCMessage` so that users who plan to stick with Chisel 3.6 and SFC can easily suppress all of these messages with a single suppression?

src/main/scala/chisel3/stage/package.scala
```diff
@@ -12,6 +12,8 @@ import scala.annotation.nowarn
 
 package object stage {
 
+  final val pleaseSwitchToCIRCT = deprecatedMFCMessage + "Please switch to circt.stage.ChiselStage."
```
jackkoenig, line 15:
```suggestion
  final val pleaseSwitchToCIRCT = deprecatedMFCMessage + " Please switch to circt.stage.ChiselStage."
```
Warning currently runs together:
```
object ChiselStage in package stage is deprecated (since Chisel 3.6): this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md.Please switch to circt.stage.ChiselStage.
```

seldridge: Thanks! Good catch. Corrected in the last `!fixup` commit.

seldridge:
I'm combining this to a single commit for ease of cherry-picking on the `chisel5-slice-n-dice` branch.

jackkoenig (approved):

PR #3009, opened 2023-02-14
Builder ContextCache

jackkoenig:
This is sort of a first attempt at an API discussed https://github.com/chipsalliance/chisel3/pull/2923#issuecomment-1376432828 and https://github.com/chipsalliance/chisel3/pull/2900#discussion_r1085901366, and as such I think it's very useful for https://github.com/chipsalliance/chisel3/pull/2958

See each commit, 1st adds the API, next 3 use it for 3 different use cases. I'm actually not that keen on using it for ChiselEnum just because it turns a static function call into a dynamic map lookup which is extra compute for dubious benefit. I think this is definitely useful for `Instantiate` and the intrinsics (more compute overhead but relatively sparsely used and more functional since it's cleared between elaborations).

This is currently private as I'm just trying to get the ball rolling on this API, would appreciate feedback but we can also have a lower barrier to merging this cleanup than we do to adding the public API.

### Contributor Checklist

#### Type of Improvement

 - code refactoring 
 - code cleanup  

#### API Impact

No impact, this API is currently private

This will change the numbering of SizeOf intrinsics when you have multiple separate Chisel elaborations in a single JVM.

#### Desired Merge Strategy

- Rebase

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/circt/SizeOf.scala
+++ b/src/main/scala/chisel3/util/circt/SizeOf.scala
@@ -6,19 +6,18 @@ import scala.util.hashing.MurmurHash3
 
 import chisel3._
 import chisel3.experimental.{annotate, ChiselAnnotation, ExtModule}
+import chisel3.internal.{Builder, BuilderContextCache}
 
 import circt.Intrinsic
 
 // We have to unique designedName per type, be we can't due to name conflicts
 // on bundles.  Thus we use a globally unique id.
 private object SizeOfGlobalIDGen {
-  private var id: Int = 0
+  private case object CacheKey extends BuilderContextCache.Key[Int]
   def getID() = {
-    this.synchronized {
-      val oldID = id
-      id = id + 1
-      oldID
-    }
+    val oldID = Builder.contextCache.getOrElse(CacheKey, 0)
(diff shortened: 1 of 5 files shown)


src/main/scala/chisel3/util/circt/SizeOf.scala
```diff
@@ -6,19 +6,18 @@ import scala.util.hashing.MurmurHash3
...
 private object SizeOfGlobalIDGen {
-  private var id: Int = 0
+  private case object CacheKey extends BuilderContextCache.Key[Int]
   def getID() = {
-    this.synchronized {
-      val oldID = id
-      id = id + 1
-      oldID
-    }
+    val oldID = Builder.contextCache.getOrElseUpdate(CacheKey, 0)
```
ekiwi, line 18: A `getOrElse` should be enough here, since you do an unconditional `put` right after

jackkoenig: contextCache isn't a `Map` and I didn't originally add `getOrElse`, but I went ahead and added it 🙂

seldridge (approved):

PR #3012, opened 2023-02-14
Improve Builder.forceName collision errors

jared-barocsi:
Changes the existing and somewhat unhelpful message about `forceName` duplicate names, and now correctly errors for three specific duplicated name scenarios

1. `forceName` attempts to name a `HasId` with an already existing name (the default/existing behavior)
2. `forceName` attempts to sanitize the name of a `HasId`, resulting in a new name that collides with an existing name
3. `forceName` attempts to name a nameless `HasId`, which is also required to not have a default name (e.g. `IO`s)

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

- no change to API

- no impact to Verilog generation

#### Desired Merge Strategy
- Squash and merge

#### Release Notes
- Improve name collision error messages when `Builder.forceName` attempts to name a `HasId` and encounters a duplicated name

- [ ] Did you review?


src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+          foo.suggestName("unsanitary")
+          // Seed an additional name which, when sanitized, collides with the previous name
+          bar.suggestName("unsanitary-")
+
+          out := foo + bar
+        },
+        
+        Array("--throw-on-first-error")
+      )
+    }).getMessage should include("Attempted to name NameCollisionSpec_Anon.unsanitary-: IO[UInt<8>] with an unsanitary name 'unsanitary-'")
```
mwachs5, line 48: does this need a collision? should we have a case without the collision where it still yells at you for the unsanitary name?

mwachs5: oh wait that was the next one.

So shouldn't this error message not just tell you had an unsanitary name, but "after sanitizing, it collides with `unsanitary`?

it would be nice if you got a source locator for the port you were using and the port you are colliding with...

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+        new Module {
+          // Write to an output port that isn't assigned to a val, and so doesn't get prefixed
+          IO(Output(UInt(8.W))) := 123.U
+        },
+        
+        Array("--throw-on-first-error")
+      )
+
+      println(chirrtl)
+    }).getMessage should include("Default names are not allowed for IO ports.")
```
mwachs5, line 78: does this have a source locator?

As a user API I'm not sure what a "Default name" is. Maybe add something to the message like "Please assign to a `val`, use `.suggestName`, or use the `prefix` API?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -260,10 +251,18 @@ private[chisel3] trait HasId extends InstanceId {
     refBuilder: String => Arg = Ref(_)
   ): Unit =
     if (_ref.isEmpty) {
-      val candidate_name = _computeName(Some(default).filterNot(_ => errorIfDup)).get
+      val candidate_name = _computeName(Some(default).filterNot(_ => errorIfDup)).getOrElse {
+        throwException(s"Default names are not allowed for IO ports. Use suggestName to seed a unique name for $this instead")
```
mwachs5, line 255:
```suggestion
        throwException(s"Default names are not allowed for IO ports. Use suggestName to seed a unique name for $this instead, or assign it to a `val`") ?
```

src/test/scala/chisel3/internal/NameCollisionSpec.scala
mwachs5: ok i looked at the full error message (i was just reading the tests) and you cover this

jackkoenig: > it would be nice if you got a source locator for the port you were using and the port you are colliding with...

I agree but that requires more extensive internals work, this PR at least makes the error messages better in the meantime.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -183,6 +183,16 @@ package object internal {
     }
   }
 
+  // Sanitizes a name, e.g. from a `HasId`, by stripping all non ANSI-C characters
+  def sanitize(s: String, leadingDigitOk: Boolean = false): String = {
```
jackkoenig, line 187:
```suggestion
  private[chisel3] def sanitize(s: String, leadingDigitOk: Boolean = false): String = {
```
package object internal is public so need to mark the method package private. Also, package object internal moved to it's own file a couple of days ago, you need to rebase.

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -63,6 +63,16 @@ package object internal {
     }
   }
 
+  // Sanitizes a name, e.g. from a `HasId`, by stripping all non ANSI-C characters
+  private[chisel3] def sanitize(s: String, leadingDigitOk: Boolean = false): String = {
+    // TODO what character set does FIRRTL truly support? using ANSI C for now
+    def legalStart(c: Char) = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_'
+    def legal(c:      Char) = legalStart(c) || (c >= '1' && c <= '9')
```
jackkoenig, line 70:
```suggestion
    def legal(c:      Char) = legalStart(c) || (c >= '0' && c <= '9')
```
Very subtle copy-paste mistake 😂 Seems like an accidental `ctrl-a` in vim.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -239,10 +230,24 @@ private[chisel3] trait HasId extends InstanceId {
...
+      if (errorIfDup && (candidate_name != sanitized) && (available_name != sanitized)) {
+        Builder.error(
+          s"Attempted to name $this with an unsanitary name '$candidate_name': sanitization results in a duplicated name '$sanitized'. Please seed a more unique name"
+        )(UnlocatableSourceInfo)
+      }
+      if (errorIfDup && (available_name != sanitized)) {
+        Builder.error(
+          s"Attempted to name $this with a duplicated name '$candidate_name'. Use suggestName to seed a unique name"
+        )(UnlocatableSourceInfo)
       }
```
jackkoenig, line 251: Are these errors intended to be exclusive? Because the conditions overlap and these are recoverable errors, if the first 1 executes the 2nd one always will. Also there's a lot of duplication in these conditions, can we make the logic a little tighter?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -239,11 +230,30 @@ private[chisel3] trait HasId extends InstanceId {
     refBuilder: String => Arg = Ref(_)
   ): Unit =
     if (_ref.isEmpty) {
-      val candidate_name = _computeName(Some(default).filterNot(_ => errorIfDup)).get
+      val candidate_name = _computeName(Some(default).filterNot(_ => errorIfDup)).getOrElse {
+        throwException(
+          s"Attempted to name a nameless IO port ($this): this is usually caused by instantiating an IO but not assigning it to a val.\nAssign $this to a val, or explicitly call suggestName to seed a unique name"
```
jackkoenig, line 235:
```suggestion
          s"Attempted to name a nameless IO port ($this): this is usually caused by instantiating an IO but not assigning it to a val.\n" +
          s"Assign $this to a val, or explicitly call suggestName to seed a unique name"
```
It's best to break up crazy long lines

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -239,11 +230,30 @@ private[chisel3] trait HasId extends InstanceId {
...
+
+      // Check for both cases of name duplication
+      if (errorIfDup && (available_name != sanitized)) {
+        // If sanitization occurred, then the sanitized name duplicate an existing name
+        if ((candidate_name != sanitized)) {
+          Builder.error(
+            s"Attempted to name $this with an unsanitary name '$candidate_name': sanitization results in a duplicated name '$sanitized'. Please seed a more unique name"
+          )(UnlocatableSourceInfo)
+          // Otherwise the candidate name duplicates an existing name
+        } else {
```
jackkoenig, line 250:
```suggestion
        } else {
         // Otherwise the candidate name duplicates an existing name
```
Nit of nits but I think the comment makes more sense inside the else curly braces

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests
+
+import chisel3._
+import chisel3.stage.ChiselStage
```
jackkoenig, line 6:
```suggestion
import circt.stage.ChiselStage
```
We're trying to avoid `chisel3.stage.ChiselStage`

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+
+class NameCollisionSpec extends ChiselFlatSpec with Utils {
+  behavior.of("Builder")
+
+  it should "error on duplicated names with a correct message" in {
+    (the[ChiselException] thrownBy extractCause[ChiselException] {
+      (new ChiselStage).emitChirrtl(
```
jackkoenig, line 13:
```suggestion
      (new ChiselStage).elaborate(
```
These error so no need for chirrtl, can just elaborate. Also the method on `circt.stage.ChiselStage` is `emitCHIRRTL` for some reason

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+      )
+    }).getMessage should include(
+      "Attempted to name NameCollisionSpec_Anon.same: IO[UInt<8>] with a duplicated name 'same'."
+    )
+  }
+
+  it should "error on sanitization resulting in duplicated names with a helpful message" in {
+    // Case one: An unsanitary name that results in a collision with an existing name once sanitized
+    (the[ChiselException] thrownBy extractCause[ChiselException] {
+      (new ChiselStage).emitChirrtl(
```
jackkoenig, line 34:
```suggestion
      (new ChiselStage).elaborate(
```

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+      val chirrtl = (new ChiselStage).emitChirrtl(
+        new Module {
+          // Write to an output port that isn't assigned to a val, and so doesn't get prefixed
+          IO(Output(UInt(8.W))) := 123.U
+        },
+        Array("--throw-on-first-error")
+      )
+
+      println(chirrtl)
+    }).getMessage should include("Default names are not allowed for IO ports.")
```
jackkoenig, line 78: This seems to be hitting a different error:
```
Unable to name port NameCollisionSpec_Anon.?: IO[UInt<8>] in chiselTests.NameCollisionSpec$$anon$4@ed3f708, try making it a public field of the Module SourceLine(src/test/scala/chisel3/internal/NameCollisionSpec.scala,72,13)"
```
Can you fix that error to call `.makeMessage(x => x)` instead of using the default `.toString` of the `SourceInfo`?

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+          out := foo + bar
+        },
+        Array("--throw-on-first-error")
+      )
+    }).getMessage should include(
+      "Attempted to name NameCollisionSpec_Anon.unsanitary-: IO[UInt<8>] with an unsanitary name 'unsanitary-'"
+    )
+
+    // Case two: An unsanitary name which does not collide with any names once sanitized. No error is raised
+    (new ChiselStage).emitChirrtl(
```
jackkoenig, line 53:
```suggestion
    (new ChiselStage).elaborate(
```

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+    (the[ChiselException] thrownBy extractCause[ChiselException] {
+      val chirrtl = (new ChiselStage).emitChirrtl(
+        new Module {
+          // Write to an output port that isn't assigned to a val, and so doesn't get prefixed
+          IO(Output(UInt(8.W))) := 123.U
+        },
+        Array("--throw-on-first-error")
+      )
+
+      println(chirrtl)
```
jackkoenig, line 77:
```suggestion
      (new ChiselStage).elaborate(
        new Module {
          // Write to an output port that isn't assigned to a val, and so doesn't get prefixed
          IO(Output(UInt(8.W))) := 123.U
        },
        Array("--throw-on-first-error")
      )
```
Chirrtl is fine for debugging, but we should get rid of it in the final test.

jackkoenig (approved):

PR #3013, opened 2023-02-15
Cleanup warnings 4

chick:
This PR takes chisel3 to no warnings using the default scala 2.13.10


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -98,7 +99,7 @@ class BoringUtilsException(message: String) extends Exception(message)
   */
 object BoringUtils {
   /* A global namespace for boring ids */
-  private val namespace: SyncVar[Namespace] = new SyncVar
+  private val namespace: LinkedBlockingDeque[Namespace] = new LinkedBlockingDeque[Namespace](1)
```
ekiwi, line 102: This should probably use the new Builder ContextCache: https://github.com/chipsalliance/chisel3/pull/3009

jackkoenig: I agree, I reverted this change so as to unblock the rest of these changes and we can fix that separately.

jackkoenig (approved):

PR #3014, opened 2023-02-15
Decouple {circt, chisel3}.stage.ChiselStage

seldridge:
Remove the dependency of circt.stage.ChiselStage on chisel3.stage.ChiselStage.  Remove the internal call to ChiselStage and instead directly construct a PhaseManager that uses the target phases of ChiselStage directly.

Add a new argument type to the existing "--target" option of circt.stage.ChiselStage, "chirrtl", to indicate that specification FIRRTL (in the specification textual format and not MLIR) should be produced.

Use this new option to migrate two methods of ChiselStage$, emitCHIRRTL and convert, away from using a call to chisel3.stage.ChiselStage.

Add tests of the new option, emitCHIRRTL, and convert.

### DIFF
--- a/src/main/scala/circt/stage/Annotations.scala
+++ b/src/main/scala/circt/stage/Annotations.scala
@@ -42,6 +42,9 @@ object CIRCTTarget {
   /** The parent type of all CIRCT targets */
   sealed trait Type
 
+  /** Specification FIRRTL */
+  case object CHIRRTL extends Type
+
   /** The FIRRTL dialect */
   case object FIRRTL extends Type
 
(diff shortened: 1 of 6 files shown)


jackkoenig (approved):
LGTM!

seldridge:
There was a weird failure stemming from a test that was doing elaboration inside a `BasicTester` and then not catching the error. I think what was going on was that the exception was getting wrapped by the "pretty stdout" exception logic and then not showing up as an actual exception. I just rewrote the tests to avoid nested elaboration.

PR #3021, opened 2023-02-16
Cleanup warnings final

chick:
Fixed up two final warnings.
In BoringUtils replace deprecated SyncVar with BuilderContextCache
In stage.phases.CIRCT fixed possible not exhaustive match
Number of warnings in 2.13 == 0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ lazy val commonSettings = Seq(
   // Macros paradise is integrated into 2.13 but requires a scalacOption
   scalacOptions ++= {
     CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, n)) if n >= 13 => "-Ymacro-annotations" :: "-Werror" :: Nil
+      case Some((2, n)) if n >= 13 => "-Ymacro-annotations" :: Nil
       case _                       => Nil
     }
   },
@@ -31,6 +31,15 @@ lazy val commonSettings = Seq(
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, n)) if n >= 13 => Nil
       case _                       => compilerPlugin(("org.scalamacros" % "paradise" % "2.1.1").cross(CrossVersion.full)) :: Nil
+    }
+  }
+)
+
+lazy val fatalWarningsSettings = Seq(
+  scalacOptions ++= {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, n)) if n >= 13 => "-Werror" :: Nil
+      case _                       => Nil
     }
   }
 )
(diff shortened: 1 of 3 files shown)


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -98,19 +98,18 @@ class BoringUtilsException(message: String) extends Exception(message)
   */
 object BoringUtils {
   /* A global namespace for boring ids */
-  private val namespace: SyncVar[Namespace] = new SyncVar
-  namespace.put(Namespace.empty)
+  private[chisel3] case object CacheKey extends BuilderContextCache.Key[Namespace]
+  private val boringNamespace = Builder.contextCache.getOrElseUpdate(CacheKey, Namespace.empty)
```
jackkoenig, line 102:
```suggestion
  private def boringNamespace = Builder.contextCache.getOrElseUpdate(CacheKey, Namespace.empty)
```
This cannot be a `val`, the point is that multiple Chisel elaborations need to be able to run in the same JVM at the same time.

build.sbt
```diff
@@ -35,6 +35,15 @@ lazy val commonSettings = Seq(
   }
 )
 
+lazy val fatalWarningsSettings = Seq(
+  scalacOptions ++= {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, n)) if n >= 13 => "-Werror" :: Nil
```
ekiwi, line 41: I would recommend only enabling this in CI. While you are developing, you might want to leave some warnings unaddressed until you have solved the problem you are working on.
In `chiseltest` we check for warnings like this:
https://github.com/ucb-bar/chiseltest/blob/fed4822cc562293e373ed568d6a12139b631e94c/.github/workflows/test.yml#L159

jackkoenig: Good idea @ekiwi, I agree that we should do it this way.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -98,19 +98,18 @@ class BoringUtilsException(message: String) extends Exception(message)
...
-    namespace.put(ns)
-    valuex
+    boringNamespace.name(value)
   }
+  /* True if the requested name (value) exists in the namespace */
+  private def checkName(value: String): Boolean = boringNamespace.contains(value)
 
   /* True if the requested name (value) exists in the namespace */
-  private def checkName(value: String): Boolean = namespace.get.contains(value)
+  //  private def checkName(value: String): Boolean = namespace.get.contains(value)
```
jackkoenig, line 112: I think this and the comment above need to be deleted.

build.sbt
chick: That seems like a good idea. I will try and get this working in chisel3

src/main/scala/chisel3/util/experimental/BoringUtils.scala
chick: Done

jackkoenig (approved):
LGTM, we'll address the idea of having warnings as errors only in CI in a follow on PR.

PR #3026, opened 2023-02-17
Add "--split-verilog" circt.*.ChiselStage option

seldridge:
Add an option to "circt.stage.ChiselStage", "--split-verilog", which replaces the janky "firrt.EmitAllModulesAnnotation" method of generating one-file-per-module output and output file splitting in "firtool". Leave in the old behavior as a run-time warning until this gets removed sometime after the Chisel 5 release.

### DIFF
--- a/src/main/scala/circt/stage/CIRCTStage.scala
+++ b/src/main/scala/circt/stage/CIRCTStage.scala
@@ -22,7 +22,8 @@ trait CLI { this: Shell =>
     ChiselGeneratorAnnotation,
     PrintFullStackTraceAnnotation,
     ThrowOnFirstErrorAnnotation,
-    WarningsAsErrorsAnnotation
+    WarningsAsErrorsAnnotation,
+    SplitVerilog
   ).foreach(_.addOptions(parser))
 }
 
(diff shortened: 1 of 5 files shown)


seldridge:
This is primarily motivated by the fact that the `EmitAllModules(classOf[firrtl.Emitter])` mechanism of running `circt.stage.ChiselStage` was never very ergonomic and confusing as you can give it any `Emitter` and it will do the same thing. (Also, `firtool` almost always needs to run in this mode for any sane flow that produces auxiliary metadata files.) This also enables deprecation of the `EmitAllModules` annotation which, with SFC removed, only needs to hang around for split-file emission.

src/main/scala/circt/stage/CIRCTStage.scala
```diff
@@ -22,7 +22,8 @@ trait CLI { this: Shell =>
     ChiselGeneratorAnnotation,
     PrintFullStackTraceAnnotation,
     ThrowOnFirstErrorAnnotation,
-    WarningsAsErrorsAnnotation
+    WarningsAsErrorsAnnotation,
+    ChiselStage.scala
```
jackkoenig, line 26: 🧐

jackkoenig (approved):
Approved once you fix the compile error

mwachs5 (commented):
Fix dangling comment plz

src/main/scala/circt/stage/CIRCTStage.scala
seldridge: I likely botched the cherry-pick. This one doesn't apply cleanly between here and the chisel5 branch. 😅

PR #3028, opened 2023-02-17
Deprecate Protobuf Emission

seldridge:
Deprecate ProtoBufFileFormat and warn if a user is doing Protobuf emission.  This will be ripped out in Chisel 5.


jackkoenig (approved):
💪👍

PR #3030, opened 2023-02-17
Add an annotation for specifying module port conventions

rwy7:
This annotation is being handled in mfc via:
  https://github.com/llvm/circt/pull/4665

The convention lets users define how ports with aggregate types are lowered to verilog.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/circt/Convention.scala
+++ b/src/main/scala/circt/Convention.scala
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package circt
+
+import chisel3.experimental.{annotate, BaseModule, ChiselAnnotation}
+import firrtl.annotations.{ModuleTarget, SingleTargetAnnotation}
+import firrtl.FirrtlUserException
+
+/** Annotation to specify a module's port convention.
+  * The port convention defines how the ports of a module are transformed while
+  * lowering to verilog.
+  */
+case class ConventionAnnotation(target: ModuleTarget, convention: String) extends SingleTargetAnnotation[ModuleTarget] {
+  override def duplicate(target: ModuleTarget): ConventionAnnotation =
+    ConventionAnnotation(target, convention)
+}
+
+/** Utilities for annotating modules with a port convention.
+  * The port convention defines how the ports of a module are transformed while
+  * lowering to verilog.
+  *
+  * @example {{{
+  * class Inner extends Module {
+  *   val io = IO(new Bundle{})
+  * }
+  *
+  * class Top extends Module {
+  *   val inner = Module(new Inner)
+  *   convention.scalarized(inner)
+  * }
+  * }}}
+  */
+object convention {
+  private def apply[T <: BaseModule](data: T, convention: String): Unit =
+    annotate(new ChiselAnnotation {
+      def toFirrtl = ConventionAnnotation(data.toTarget, convention)
+    })
+
+  /** Annotate a module as having the "scalarized" port convention.
+    * With this port convention, the aggregate ports of the module will be
+    * fully scalarized.
+    */
+  def scalarized[T <: BaseModule](data: T): Unit =
+    apply(data, "scalarized")
+}
--- a/src/test/scala/circtTests/ConventionSpec.scala
+++ b/src/test/scala/circtTests/ConventionSpec.scala
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package circtTests
+
+import chisel3._
+import circt.convention
+import circt.stage.ChiselStage
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+
+class ConventionSpec extends AnyFunSpec with Matchers {
+  describe("convention.scalarized") {
+    it("should scalarize the ports of a module") {
+      class PassThrough extends RawModule {
+        val i = IO(Input(new Bundle {
+          val x = UInt(16.W)
+          val y = UInt(16.W)
+        }))
+        val o = IO(Output(new Bundle {
+          val x = UInt(16.W)
+          val y = UInt(16.W)
+        }))
+        o := i
+      }
+
+      def makePassThrough() = {
+        val p = new PassThrough
+        convention.scalarized(p)
+        p
+      }
+
+      val expected = Array(
+        "input  [15:0] i_x,",
+        "input  [15:0] i_y,",
+        "output [15:0] o_x,",
+        "output [15:0] o_y",
+        "assign o_x = i_x;",
+        "assign o_y = i_y;"
+      )
+
+      val options = Array(
+        "--preserve-aggregate=all",
+        "--strip-debug-info",
+        "--scalarize-top-module=false",
+        "--lowering-options=disallowPortDeclSharing"
+      )
+
+      val actual = ChiselStage.emitSystemVerilog(makePassThrough(), Array(), options).split("\n").map(_.trim)
+      expected.foreach { e => actual should contain(e) }
+    }
+  }
+}


src/main/scala/chisel3/annotations/annotations.scala
```diff
@@ -0,0 +1,44 @@
...
+    case "internal" => Internal
+    case "scalarized" => Scalarized
+    case _ => throw new FirrtlUserException(s"Unrecognized Convention: $string")
+  }
+}
+
+/** Annotation to force the compiler to use a specified port convention.
+  * May only be applied to modules.
+  */
+case class ConventionAnnotation(target: ModuleName, convention: String)
```
rwy7, line 32: Can't store the convention as the enum version, has to be a string, because we cannot extend the anno serialization code, which is buried in firrtl.

src/main/scala/chisel3/annotations/annotations.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import firrtl.annotations._
+import firrtl.FirrtlUserException
+
+/** The lowering convention of a module
+  * Controls how a module's ports are lowered to verilog.
+  */
+sealed class Convention(val string: String) {
+  def serialize: String = string
+}
```
rwy7, line 13: implementation based on another annotation in firrtl:
https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/annotations/LoadMemoryAnnotation.scala#L11-L26

src/main/scala/chisel3/annotations/annotations.scala
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.annotations
```
ekiwi, line 3: This seems like it probably should go into the `circt` package unless these different lowering conventions are added to the firrtl spec.

src/main/scala/chisel3/annotations/annotations.scala
```diff
@@ -0,0 +1,45 @@
...
+}
+
+object ConventionAnnotation {
+
+  /** Create a ConventionAnnotation
+    * Helper to convert create an annotation from the Convention enum.
+    */
+  def apply(target: ModuleName, convention: Convention): ConventionAnnotation =
+    ConventionAnnotation(target, convention.serialize)
+}
```
jackkoenig, line 45: This all looks reasonable, but this does need some Chisel user-API, some function-like API that accepts a Chisel module as an argument and annotates it. See `dontTouch` for an example (although that operates on `Data` instead of `RawModule`): https://github.com/chipsalliance/chisel3/blob/3c14563743971842069679c263adbc1aa74adc3c/core/src/main/scala/chisel3/dontTouch.scala

rwy7: Done, thanks!

src/main/scala/chisel3/annotations/annotations.scala
rwy7: Done, thanks!

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import chisel3.RawModule
+import chisel3.experimental.{annotate, ChiselAnnotation}
+import firrtl.annotations.{ModuleName, SingleTargetAnnotation}
+import firrtl.FirrtlUserException
+
+/** Annotation to force the compiler to use a specified port convention.
+  * May only be applied to modules.
+  */
+case class ConventionAnnotation(target: ModuleName, convention: String) extends SingleTargetAnnotation[ModuleName] {
```
rwy7, line 11: Not sure if the annotation should go into a package dedicated to annotations, or if it's good here.

src/main/scala/chisel3/annotations/annotations.scala
seldridge: I'm not sure I agree with this. Currently there are _no_ Annotations enumerated in the spec, yet we have Annotations scattered all over.

Starting to put these in a common area would greatly simplify the structure of the code base. (And I would like to move all annotations here in the future.)

src/main/scala/circt/Convention.scala
seldridge, line 11: `ModuleTarget` is a slightly better API than `ModuleName` (though they are basically the same thing). This requires using `toTarget` below as opposed to `toNamed` in the Chisel API.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+  def duplicate(target: ModuleName): ConventionAnnotation =
```
seldridge, line 12: Nit: `override def` is better Scala style as it indicates that you are overriding something/implementing an abstract member. Unfortunately there is no way to strictly enforce this. 🥲

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+    case _            => throw new FirrtlUserException(s"Unrecognized Convention: $string")
+  }
+}
+
+object convention {
+  def apply[T <: RawModule](data: T, convention: Convention) {
+    annotate(new ChiselAnnotation {
+      def toFirrtl = ConventionAnnotation(data.toNamed, convention.serialize)
+    })
+  }
```
seldridge, line 43: I'd make this `private` for now. The `scalarized` and `internal` methods are sufficient to expose.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+  def duplicate(target: ModuleName): ConventionAnnotation =
+    ConventionAnnotation(target, convention)
+}
+
+/** The lowering convention of a module
+  * Controls how a module's ports are lowered to verilog.
+  */
+sealed class Convention(val string: String) {
+  def serialize: String = string
+}
```
seldridge, line 21: Given that this is declared as `val string`, the PR can actually skip the `def serialize` and just just `convention.string`. The `val` will create a public member.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+  case object Scalarized extends Convention("scalarized")
+
+  def deserialize(string: String): Convention = string match {
+    case "internal"   => Internal
+    case "scalarized" => Scalarized
+    case _            => throw new FirrtlUserException(s"Unrecognized Convention: $string")
+  }
+}
+
+object convention {
```
seldridge, line 38: Including some Scaladoc on this. Something like what is on `dontTouch` or `BoringUtils` are good examples.

seldridge (commented):
Looking great! Thanks for your first Chisel PR.

This needs tests and I think it's good to go.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
+package chisel3.circt
+
+import chisel3.RawModule
+import chisel3.experimental.{annotate, ChiselAnnotation}
+import firrtl.annotations.{ModuleName, SingleTargetAnnotation}
+import firrtl.FirrtlUserException
+
+/** Annotation to force the compiler to use a specified port convention.
+  * May only be applied to modules.
```
mwachs5, line 9: can you be a little more precise on this comment? any subclass of `BaseModule`? (can I put this annotation on a BlackBox)?

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
+package chisel3.circt
```
mwachs5, line 1: I think this is the first time we have a `package chisel3.circt`, everything else is just `package circt`?

I'm fine with a `chisel3.annotations` or `circt.annotations` or just `circt`. If we decide to put it in a subpackage we should (proper scala style) mirror that in the directory structure.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def deserialize(string: String): Convention = string match {
+    case "internal"   => Internal
+    case "scalarized" => Scalarized
+    case _            => throw new FirrtlUserException(s"Unrecognized Convention: $string")
+  }
+}
+
+object convention {
+  def apply[T <: RawModule](data: T, convention: Convention) {
```
mwachs5, line 39: so you CANNOT put this on an ExtModule?

rwy7: I think it makes sense to support ext modules, this is just a mistake on my part, It seems like `BaseModule` would be the right thing?

src/main/scala/chisel3/annotations/annotations.scala
ekiwi: I guess this is a nudge from me for the SiFive team to start putting interface constraints, i.e., the "ABI" of the circuit, into the firrtl spec.

mwachs5: there is a PR open here for the current version: https://github.com/chipsalliance/firrtl-spec/pull/70

src/test/scala/circtTests/ConventionSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+       |              reset,
+       | input  [15:0] i_x,
+       |               i_y,
+       | output [15:0] o_x,
+       |               o_y
+       |);
+       |  assign o_x = i_x;
+       |  assign o_y = i_y;
+       |endmodule
+    """.stripMargin.split("\n").map(_.trim)
```
seldridge, line 41: This test can be tightened up to only check the scalarized ports as opposed to everything.

Changing the `ChiselStage` invocation to use `--loweringOptions=disallowPortDeclSharing` will allow more exact checking of the ports.

src/main/scala/circt/Convention.scala
```diff
@@ -0,0 +1,45 @@
...
+import firrtl.FirrtlUserException
+
+/** Annotation to specify a module's port convention.
+  * The port convention defines how the ports of a module are transformed while
+  * lowering to verilog.
+  */
+case class ConventionAnnotation(target: ModuleTarget, convention: String) extends SingleTargetAnnotation[ModuleTarget] {
+  override def duplicate(target: ModuleTarget): ConventionAnnotation =
+    ConventionAnnotation(target, convention)
+}
```
seldridge, line 16: Two thoughts:

1. This can be made private. We only want people accessing it through the `convention` object.
2. This can actually be a `NoTargetAnnotation` instead of a `SingleTargetAnnotation[ModuleTarget]`. This avoids having to provide an implementation of the abstract `duplicate` method.

src/test/scala/circtTests/ConventionSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+package circtTests
+
+import chisel3._
+import circt.convention
+import circt.stage.ChiselStage
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+
+class ConventionSpec extends AnyFunSpec with Matchers {
+  private class PassThrough extends Module {
```
seldridge, line 12: Nit:
```suggestion
  private class PassThrough extends RawModule {
```

Using `RawModule` avoids creating the "implicit" clock and reset ports which are unused here.

seldridge (commented):
A few comments. This is looking great.

src/main/scala/circt/Convention.scala
seldridge: Talked to @jackkoenig: there may be some issues with converting this to `NoTargetAnnotation` due to the injecting aspect path possibly relying on this. Let's keep this as `SingleTargetAnnotation[ModuleTarget]` for now.

rwy7: OK sounds good, leaving as a single target anno (this means I can't make `duplicate` private, because it's an overriden method :)) Thanks.

src/test/scala/circtTests/ConventionSpec.scala
rwy7: done, thanks

src/test/scala/circtTests/ConventionSpec.scala
rwy7: ah this is much better, thanks.

seldridge (approved):
LGTM

jackkoenig:
This needs to be formatted (`sbt fmt`) and it probably also needs https://github.com/chipsalliance/chisel/pull/3179 to be merged first.

jackkoenig (approved):

PR #3036, opened 2023-02-23
Add `const` type support

debs-sifive:
### Contributor Checklist

- [ ] Did you review?


firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -984,6 +984,15 @@ abstract class AggregateType extends Type {
...
 }
+case class ConstGroundType(underlying: GroundType) extends GroundType with UseSerializer{
+  val width = underlying.width
+  def mapWidth(f:     Width => Width) = underlying.mapWidth(f)
+  def foreachWidth(f: Width => Unit) = underlying.foreachWidth(f)
+}
+case class ConstAggregateType(underlying: AggregateType) extends AggregateType with UseSerializer{
+  def mapType(f: Type => Type): Type = underlying.mapType(f)
+  def foreachType(f: Type => Unit): Unit = underlying.foreachType(f)
+}
```
jackkoenig, line 995: I know I originally suggested having both, but can you make `ConstType` just extend `Type`? Do we need both?

debs-sifive: done, I don't think we need both.

firrtl/src/test/scala/firrtlTests/SerializerSpec.scala
```diff
@@ -98,6 +98,30 @@ object SerializerSpec {
...
+          DefWire(NoInfo, "constInt", ConstType(UIntType(IntWidth(3)))),
+          DefWire(NoInfo, "constAsyncReset", ConstType(AsyncResetType)),
+          DefWire(NoInfo, "constBundle", ConstType(BundleType(Seq(
+            Field("real", Default, UIntType(IntWidth(32))),
+            Field("imag", Default, UIntType(IntWidth(32))),
+            Field("other", Default, ConstType(SIntType(IntWidth(1)))),
+          )))),
+        )
+      )
+    )
```
seldridge, line 124: Nit: this may be slightly cleaner as simpler tests of each `Serializer.serialize(DefWire(...)) should be "wire ..."` instead of making it a complete module. This may then be short enough to inline all the tests into one-line-per test.

core/src/main/scala/chisel3/experimental/Const.scala
```diff
@@ -0,0 +1,15 @@
...
+
+/** Create a constant type in FIRRTL, which is guaranteed to take a single
+  * constant value.
+  */
+object Const {
+  def apply[A <: Data](a: A): A = {
+    a.isConst = true
+    a
+  }
+}
```
jackkoenig, line 15: Chisel APIs need to give the illusion of immutability even if they actually use mutation under the hood. We typically do this by cloning the argument, mutating the value on the clone, and then returning the clone although we have optimized this a bit to only clone when we need to. Take a look at how `Input`, `Output`, and `Flipped` work (which are setting the direction while pretending to not mutate anything).

debs-sifive: Are you referring to the code here? https://github.com/chipsalliance/chisel3/blob/4194e10697ce501a79da2ddad323f2e3dced830a/core/src/main/scala/chisel3/Data.scala#L59

Is it ok to use `cloneType` for this? I am getting the following when using cloneType in the `Bundle` case:
```
chisel3.package$ExpectedHardwareException: 'FooBundle' must be hardware, not a bare Chisel type. Perhaps you forgot to wrap it in Wire(_) or IO(_)?
```

mwachs5 (commented):
High level comment, do we plan to bump the version of the firrtl spec in this pr to one that contains const?

debs-sifive:
@mwachs5 can you explain more? Isn't `firrtl-spec` going to remain a separate repo? Or is there some link in `chisel3` to `firrtl-spec`?

jackkoenig:
> @mwachs5 can you explain more? Isn't firrtl-spec going to remain a separate repo? Or is there some link in chisel3 to firrtl-spec?

The version of FIRRTL we emit is encoded in the `Serializer` and needs to be bumped to match whatever version of the spec includes `const`: https://github.com/chipsalliance/firrtl/blob/6a999e306c81bb6a3276311022349523a576b552/src/main/scala/firrtl/ir/Serializer.scala#L20

core/src/main/scala/chisel3/experimental/Const.scala
jackkoenig: Yes something like that. Yes you in fact _need_ to clone the thing at times, that's the point. Instead of `cloneType` though you should use `cloneTypeFull` which maintains set specified direction. I don't know why you're hitting that exception, but that must be a bug in your code, the point stands that `Const` must clone things (when it needs to as defined by `mustClone` as used in the logic you linked).

debs-sifive:
OK sounds good r.e. spec. I don't think `const` has been merged to the FIRRTL spec yet, but updating the version sounds like something we should do in this PR.

core/src/main/scala/chisel3/experimental/Const.scala
debs-sifive: Ah yes, I see my bug, thanks Jack!

azidar:
Overall, approach and implementation look great!

PR #3037, opened 2023-02-23
Detect bound hardware when processing record elements

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

   - code cleanup                     

#### API Impact

Using bound hardware as an Element for a Record will be caught earlier and throw a better error message 

N/A
#### Desired Merge Strategy

#### Release Notes

Using bound hardware as an Element for a Record will now throw a better error message 

- [ ] Did you review?


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1023,6 +1023,12 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
           s"${this.className} does not return the same objects when calling .elements multiple times. Did you make it a def by mistake?"
         )
       }
+      if (child.binding.isDefined) {
+        throw RebindingException(
+          s"Cannot create Record ${this.className}; element ${child} of record is already a bound hardware."
```
ekiwi, line 1028: This PR is maybe not the right place to address this. But do we really expect users to make anything of the term `bound hardware`? Like if you ask someone at SiFive using Chisel, would they be able to tell you what it means?

adkian-sifive: But yeah that's a good point, I think something about what it means for the thing to be bound would be helpful here

ekiwi: I was commenting on the error message. I wonder if it would be better to frame this issue as `Chisel Type` vs. `Chisel Value` or something like that. `a bound hardware` seems very implementation specific. But please let these musings not hold up your PR. If we were to improve the language around this issue, we would have to update this in many places, so this would require a separate PR anyways.

jackkoenig: I agree with Kevin, we should have better terminology but it's a much larger issue than this PR.

jackkoenig:
I'm marking this 3.6 because, while this is just an improved error message, it plausibly will cause code that currently works to throw an exception. It is a tightening up of intended semantics, but ones we clearly weren't enforcing for a long time.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -951,6 +951,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
             " See the cookbook entry 'How do I deal with the \"unable to clone\" error?' for more details."
         )
       }
+      if (field.binding.isDefined) {
+        throw RebindingException(
+          s"Cannot create Record ${this.className}; element ${field} of record is already a bound hardware."
```
mwachs5, line 956: we can use terminology here and/or refer people to this link: https://www.chisel-lang.org/chisel3/docs/explanations/chisel-type-vs-scala-type.html

mwachs5: and that is the sort of error messages you usually get:

```
// chisel3.package$ExpectedChiselTypeException: wire type '_44_Anon.hardware: Wire[MyBundle]' must be a Chisel type, not hardware
```

mwachs5, line 956:
```suggestion
          s"Cannot create Record ${this.className}; element ${field} of record must be a Chisel type, not hardware."
```

to be consistent with that message. maybe we should pull that "must be a Chisel type, not hardware" into some common string so we're consistent with these sorts of errors?

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -951,6 +951,11 @@ abstract class Record(private[chisel3] implicit val compileOptions: CompileOptio
             " See the cookbook entry 'How do I deal with the \"unable to clone\" error?' for more details."
         )
       }
+      if (field.binding.isDefined) {
```
jackkoenig, line 954:
```suggestion
      if (clone.elements(name).binding.isDefined) {
```
This was checking the `field` of the parent, not the `clone`.

Although please assign `clone.elements(name)` to a `val` above rather than doing the lookup twice... it will probably be optimized by the compiler but I'd rather not risk it.

adkian-sifive: Hmm I'm seeing that with `clone.elements(name)`, the `binding.isDefined` is always false and the test I added is failing with no exceptions

adkian-sifive: did we wanna check the clone or the field itself? i thought the field was the thing that we shouldn't be cloning if it's bound already, not the clone

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1009,13 +1009,18 @@ abstract class Record extends Aggregate {
...
 
-    for ((child, sameChild) <- this.elementsIterator.zip(this.elementsIterator)) {
-      if (child != sameChild) {
+    // This check is for making sure that elements always returns the
+    // same object, which will not be the case if the user makes it a
+    // def inside the Record. Checking elementsIterator against itself
+    // is not useful for this check because it's a lazy val which will
+    // always return the same thing.
+    for ((child, sameChild) <- this.elements.iterator.zip(this.elementsIterator)) {
+      if (child._2 != sameChild) {
```
jackkoenig, line 1018:
```suggestion
    for (((_, child), sameChild) <- this.elements.iterator.zip(this.elementsIterator)) {
      if (child != sameChild) {
```

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1009,13 +1009,18 @@ abstract class Record extends Aggregate {
...
         throwException(
           s"${this.className} does not return the same objects when calling .elements multiple times. Did you make it a def by mistake?"
         )
       }
-      child.bind(ChildBinding(this), resolvedDirection)
+      child._2.bind(ChildBinding(this), resolvedDirection)
```
jackkoenig, line 1023:
```suggestion
      child.bind(ChildBinding(this), resolvedDirection)
```
With the above

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1180,6 +1185,17 @@ abstract class Record extends Aggregate {
 
   def elements: SeqMap[String, Data]
 
+  private[chisel3] lazy val _elements: SeqMap[String, Data] = {
```
jackkoenig, line 1188: It would be good to have a comment describing why we use this now instead of `elements`.

jackkoenig (changes requested):
A few nits, and then a larger thought.

There are about ~35 remaining places where we call `.elements` on a `Record` in `core`. IMO all of these should be changed to be calls to `_elements` since it will sure we do the check at the earliest possible point.

This does lead to a slightly awkward issue that while we have always used `elements`, it's somewhat better to [internally] use `_elements`, but it's not incorrect to keep using `elements` so even if we remove uses, it'll creep back in. That being said, I still think we should try to use `_elements` everywhere we can.

core/src/main/scala/chisel3/Aggregate.scala
adkian-sifive: nice

jackkoenig (approved):

PR #3038, opened 2023-02-23
Convert Scaladoc comment to normal comment, NFC

seldridge:
Change a comment to workaround an unmoored Scaladoc comment error.

This is done to workaround an error I started seeing here: https://github.com/chipsalliance/chisel3/actions/runs/4256264963/jobs/7404920411#step:8:1225


seldridge:
This doesn't fix the issue in the PR description. The problem is that `firrtl` has `-Xfatal-warnings` set for the doc build. This arguably should be on for Chisel. However, this is a lot of cleanup to get it to that point and should form part of a larger cleanup.

jackkoenig:
@chick is working on cleaning up the Chisel3 ScalaDoc to enable fatal warnings

seldridge:
This should already be fixed via other mechanisms.

PR #3039, opened 2023-02-24
Give a source line and caret for errors when possible

jackkoenig:
This includes new command-line option --source-root which allows the user of Chisel to point to the root directory for source files. The default value is "." which should work for typical SBT and Mill projects.

### Contributor Checklist

#### Type of Improvement

 - new feature/API 

#### API Impact

New command-line, annotation API into Chisel, but no other API impact.

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Improve error reporting with source line and caret pointing to column. This requires knowing the root directory of source files which can be provided with `--source-root <dir>`. It will default to `.` which is the directory in which the Chisel elaboration process is invoked which should be correct for typical projects.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Error.scala
+++ b/core/src/main/scala/chisel3/internal/Error.scala
@@ -6,6 +6,8 @@ import scala.annotation.tailrec
 import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, LinkedHashSet}
 import scala.util.control.NoStackTrace
 import _root_.logger.Logger
+import java.io.File
+import scala.io.Source
 
 import chisel3.experimental.{NoSourceInfo, SourceInfo, SourceLine, UnlocatableSourceInfo}
 
@@ -98,7 +100,29 @@ private[chisel3] object ErrorLog {
   val errTag = s"[${Console.RED}error${Console.RESET}]"
 }
 
-private[chisel3] class ErrorLog(warningsAsErrors: Boolean) {
+private[chisel3] class ErrorLog(warningsAsErrors: Boolean, sourceRoots: Seq[File]) {
+
+  private def getErrorLineInFile(sl: SourceLine): List[String] = {
+    def tryFileInSourceRoot(sourceRoot: File): Option[List[String]] = {
+      try {
+        val file = new File(sourceRoot, sl.filename)
+        val lines = Source.fromFile(file).getLines()
+        var i = 0
+        while (i < (sl.line - 1) && lines.hasNext) {
+          lines.next()
+          i += 1
+        }
+        val line = lines.next()
+        val caretLine = (" " * (sl.col - 1)) + "^"
+        Some(line :: caretLine :: Nil)
+      } catch {
+        case scala.util.control.NonFatal(_) => None
+      }
+    }
+    val sourceRootsWithDefault = if (sourceRoots.nonEmpty) sourceRoots else Seq(new File("."))
+    // View allows us to search the directories one at a time and early out
+    sourceRootsWithDefault.view.map(tryFileInSourceRoot(_)).collectFirst { case Some(value) => value }.getOrElse(Nil)
+  }
 
   /** Returns an appropriate location string for the provided source info.
     * If the source info is of `NoSourceInfo` type, the source location is looked up via stack trace.
@@ -119,8 +143,9 @@ private[chisel3] class ErrorLog(warningsAsErrors: Boolean) {
 
   private def errorEntry(msg: String, si: Option[SourceInfo], isFatal: Boolean): ErrorEntry = {
     val location = errorLocationString(si)
+    val sourceLineAndCaret = si.collect { case sl: SourceLine => getErrorLineInFile(sl) }.getOrElse(Nil)
     val fullMessage = if (location.isEmpty) msg else s"$location: $msg"
-    ErrorEntry(fullMessage, isFatal)
+    ErrorEntry(fullMessage :: sourceLineAndCaret, isFatal)
   }
 
   /** Log an error message */
@@ -158,7 +183,7 @@ private[chisel3] class ErrorLog(warningsAsErrors: Boolean) {
       case ((message, sourceLoc), count) =>
         logger.warn(s"${ErrorLog.depTag} $sourceLoc ($count calls): $message")
     }
-    errors.foreach(e => logger.error(s"${e.tag} ${e.msg}"))
+    errors.foreach(e => logger.error(e.serialize))
 
     if (!deprecations.isEmpty) {
       logger.warn(
@@ -237,6 +262,8 @@ private[chisel3] class ErrorLog(warningsAsErrors: Boolean) {
   private def elapsedTime: Long = System.currentTimeMillis - startTime
 }
 
-private case class ErrorEntry(msg: String, isFatal: Boolean) {
+private case class ErrorEntry(lines: Seq[String], isFatal: Boolean) {
   def tag = if (isFatal) ErrorLog.errTag else ErrorLog.warnTag
+
+  def serialize: String = lines.map(s"$tag " + _).mkString("\n")
 }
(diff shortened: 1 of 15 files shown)


seldridge:
Can you bridge that to the firtool option https://github.com/llvm/circt/pull/4576 and consider keeping the option consistently named? I care less about the latter point. I can understand that include-dir may be confusingly named so only offering the alternative name for consideration. (Not blocking.)

jackkoenig:
> Can you bridge that to the firtool option https://github.com/llvm/circt/pull/4576

Absolutely, will do

> and consider keeping the option consistently named? I care less about the latter point. I can understand that include-dir may be confusingly named so only offering the alternative name for consideration.

It is an important configuration. I will admit I don't love `--include-dir` since it implies to me something like `#include` or `gcc -I` which would affect compilation when it's really just helping with error reporting. I am willing to be convinced but my preference is to do something else. I'm not sure `--source-root` is the greatest either. Perhaps we should go with something like `--project-root` or `--chisel-project-root` (to be consistent with [the Scala compile-time system property used for SourceInfos](https://github.com/chipsalliance/chisel3/blob/4cb44bc046efdd1cf64d17dcd313305affabf3d9/core/src/main/scala/chisel3/internal/SourceInfoMacro.scala#LL17C37-L17C37)). That being said, I expect this Chisel runtime argument to be used infinitely more than the Scala compile-time one.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -99,6 +99,25 @@ case object WarningsAsErrorsAnnotation
...
+      toAnnotationSeq = { dir =>
+        val f = new File(dir)
+        if (!f.isDirectory()) {
+          throw new OptionsException(s"Must be directory that exists!")
+        }
+        Seq(SourceRootAnnotation(f))
+      },
+      helpText = "Root directory for source files, used for enhanced error reporting"
+    )
+  )
```
seldridge, line 118: Is it reasonable that this could be a `Seq[String]` instead of a single value? I.e., it's not just a single project root, but could be a bunch of directories where sources live. (The option on the `firtool` side accepts a list.)

jackkoenig: I should be able to make that work!

mwachs5:
> and consider keeping the option consistently named?

I'm not sure I love the `--include-dir` name, so I'd be fine with these names diverging

src/main/scala/circt/stage/CIRCTStage.scala
```diff
@@ -23,6 +24,7 @@ trait CLI { this: Shell =>
     PrintFullStackTraceAnnotation,
     ThrowOnFirstErrorAnnotation,
     WarningsAsErrorsAnnotation,
+    SourceRootAnnotation,
```
mwachs5, line 27: doesn't this still need to do a translation step into the invocation to `firtool`? (sorry if you're still working on this!)

jackkoenig:
@seldridge now accepting multiple `--source-root` arguments and propagating them as `--include-dir` to firtool

src/main/scala/circt/stage/CIRCTStage.scala
jackkoenig: I am now propagating it through but still need to write a test that it propagates properly, any suggestions @seldridge? Also any thoughts on if I should migrate all of the tests out of ChiselMainSpec pre-emptively since it goes away on the chisel5 branch?

seldridge: This may be difficult to test. I'm fine with _not_ testing it. There may be some magic with putting some Chisel in an alternative directory and then compiling that. I admit it's complicated. This actual feature is pretty well tested on the CIRCT side: https://github.com/llvm/circt/blob/main/test/firtool/include-dirs.fir

I moved everything that I thought made sense here: https://github.com/chipsalliance/chisel3/commit/15214885eb4354938b7310771cc90254e153cb64 Is there anything that you see that's missing?

core/src/main/scala/chisel3/internal/Error.scala
```diff
@@ -98,7 +100,29 @@ private[chisel3] object ErrorLog {
...
+  private def getErrorLineInFile(sl: SourceLine): List[String] = {
+    def tryFileInSourceRoot(sourceRoot: File): Option[List[String]] = {
+      try {
+        val file = new File(sourceRoot, sl.filename)
+        val lines = Source.fromFile(file).getLines()
+        var i = 0
+        while (i < (sl.line - 1) && lines.hasNext) {
+          lines.next()
+          i += 1
+        }
```
seldridge, line 114: I was trying to see if there was a way to do this with `Positioner` or some way which doesn't involve chugging through line-by-line. I couldn't figure it out, though. Future work.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -93,12 +93,36 @@ case object WarningsAsErrorsAnnotation
     new ShellOption[Unit](
       longOption = "warnings-as-errors",
       toAnnotationSeq = _ => Seq(WarningsAsErrorsAnnotation),
-      helpText = "Treat warnings as errors"
+      helpText = "Treat warnings as errors",
+      helpValueName = Some("<file>")
```
seldridge, line 97: Nit: unrelated change.

src/main/scala/circt/stage/CIRCTStage.scala
jackkoenig: https://github.com/chipsalliance/chisel3/commit/15214885eb4354938b7310771cc90254e153cb64 doesn't seem to actually check that the source line and caret are there, I also added a couple of a new tests showing that multiple `--source-root` arguments are allowed and that the first wins when a similarly named source file exists in both. See the stuff I added to `src/test/resources` to help

seldridge (approved):
LGTM

It would be nice to duplicate the tests into `circtTests.stage.ChiselStageSpec` to avoid me having to do this in the `chisel5-slice-n-dice` branch.

src/test/scala/chiselTests/stage/ChiselMainSpec.scala
```diff
@@ -275,8 +283,70 @@ class ChiselMainSpec extends AnyFeatureSpec with GivenWhenThen with Matchers wit
...
+        generator = Some(classOf[BuilderErrorModuleFakeSourceInfo]),
+        message = Seq(Right("Fatal errors during hardware elaboration")),
+        stdout = Seq(
+          Right(
+            "Foo:3:10: Invalid bit range (3,-1)"
+          ),
+          // This spacing is important to make sure the caret is correct
+          // Keep the source-root test below in sync
+          Right("I am the file in sourceroot2"),
+          Right("         ^")
```
mwachs5, line 349: as I said, can't you write a test that will hit a firrtl -> .v error (an uninitialized Wire(Bool()) should do it) which goes through emitVerilog and reflects the error in the Exception message?

src/main/scala/chisel3/stage/ChiselAnnotations.scala
jackkoenig: Lol I put that in the wrong place, that was supposed to be on `SourceRootAnnotation`, will fix

src/test/scala/chiselTests/stage/ChiselMainSpec.scala
jackkoenig: I added the same suite of tests to circt ChiselStageSpec which also includes checking that the --source-root to --include-dir propagation is correct

core/src/main/scala/chisel3/internal/Error.scala
jackkoenig: There's probably a better way, but hell it works

PR #3040, opened 2023-02-25
Bump firtool to 1.32.0

jackkoenig:


seldridge (approved):

seldridge:
This seems to be failing for unrelated reasons. I can't quite figure out what is going on as it appears that the `Queue` tests are somehow returning a `-1` from `Gen.choose(0, 15)` which seems impossible. E.g., test failure [here](https://github.com/chipsalliance/chisel3/actions/runs/4267415687/jobs/7429046100#step:7:197) which is showing a `tap` of `-1` coming from [here](https://github.com/chipsalliance/chisel3/blob/master/src/test/scala/chiselTests/QueueSpec.scala#L213).

jackkoenig:
Yeah that is weird, I'm taking a look. We did just delete the stuff telling it not to shrink the input, but that shouldn't have affected the initial test.

jackkoenig:
Looks like we cannot use firtool 1.32.0. The basic issue is that `chisel3.util.Queue` has an option to use a `SyncReadMem` instead of a combinationally read `Mem` as it's underlying data store, but it requires `WriteFirst` read-under-write behavior when using a `SyncReadMem`. While firtool officially does not support read-under-write for CHIRRTL mems (https://github.com/llvm/circt/issues/4588), in practice it implements them as "write first", giving the correct behavior for `Queue` with a `SyncReadMem`. [A recent change](https://github.com/llvm/circt/pull/4687) included in firtool 1.32.0 changed the lowering for `SyncReadMems` that only have a single entry, and now it no longer is implemented with write-first behavior. Considering that CIRCT change is replacing single-address memories with registers, I'm very surprised it's not maintaining "write first" behavior--perhaps there is simply a bug. Regardless, we're going to stick with firtool 1.31.0 for now.

There is a secondary issue that ScalaCheck just ignores the constraints on your inputs when shrinking inputs and thus the CI failures give very confusing (and incorrect) failing test cases. We used to disable input shrinking but the ScalaCheck API is weird on this in 2.13 and @chick and I were [unable to come up with a way to fix it](https://github.com/chipsalliance/chisel3/pull/2977#discussion_r1102295386), so we just re-enabled shrinking leading to these terrible CI failure messages. Ideally we could just remove ScalaCheck because it's not really doing much for us (that wouldn't also be caught with targeted tests hitting the corner cases), but in lieu of figuring out what the targeted tests should be, it did catch a bug here.


Lots of credit to @seldridge for helping me debug this.

PR #3041, opened 2023-02-27
Get rid of warnings in / doc build

chick:
Get rid of warnings in / doc build
Many `[[...]]` fixed but a large of class of these warnings
did not work no matter what I did. So changed these cases to just replace with back-ticks
Printable docs are very problematic, use Warning suppression for those 2

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,6 +56,19 @@ lazy val warningSuppression = Seq(
     "msg=Importing from firrtl:s",
     "msg=migration to the MLIR:s",
     "msg=method hasDefiniteSize in trait IterableOnceOps is deprecated:s",  // replacement `knownSize` is not in 2.12
+    "msg=object JavaConverters in package collection is deprecated:s",
+    "msg=INTERNALskipFile:s",
+    "msg=undefined in comment for method cf in class PrintableHelper:s",
+    "msg=method checkLengths in class StringContext is deprecated:s"
+  ).mkString(",")
+)
+
+lazy val docWarningSuppression = Seq(
+  scalacOptions += "-Wconf:" + Seq(
+    "msg=APIs in chisel3.internal:s",
+    "msg=Importing from firrtl:s",
+    "msg=migration to the MLIR:s",
+    "msg=method hasDefiniteSize in trait IterableOnceOps is deprecated:s",  // replacement `knownSize` is not in 2.12
     "msg=object JavaConverters in package collection is deprecated:s"
   ).mkString(",")
 )
@@ -110,6 +123,8 @@ lazy val chiselSettings = Seq(
 )
 
 autoCompilerPlugins := true
+autoAPIMappings := true
+autoAPIMappings := true
 
 // Plugin must be fully cross-versioned (published for Scala minor version)
 // The plugin only works in Scala 2.12+
@@ -318,7 +333,3 @@ lazy val docs = project // new documentation project
     )
   )
   .settings(fatalWarningsSettings: _*)
-
-addCommandAlias("com", "all compile")
-addCommandAlias("lint", "; compile:scalafix --check ; test:scalafix --check")
-addCommandAlias("fix", "all compile:scalafix test:scalafix")
--- a/build.sbt
+++ b/build.sbt
@@ -57,19 +57,7 @@ lazy val warningSuppression = Seq(
     "msg=migration to the MLIR:s",
     "msg=method hasDefiniteSize in trait IterableOnceOps is deprecated:s",  // replacement `knownSize` is not in 2.12
     "msg=object JavaConverters in package collection is deprecated:s",
-    "msg=INTERNALskipFile:s",
-    "msg=undefined in comment for method cf in class PrintableHelper:s",
-    "msg=method checkLengths in class StringContext is deprecated:s"
-  ).mkString(",")
-)
-
-lazy val docWarningSuppression = Seq(
-  scalacOptions += "-Wconf:" + Seq(
-    "msg=APIs in chisel3.internal:s",
-    "msg=Importing from firrtl:s",
-    "msg=migration to the MLIR:s",
-    "msg=method hasDefiniteSize in trait IterableOnceOps is deprecated:s",  // replacement `knownSize` is not in 2.12
-    "msg=object JavaConverters in package collection is deprecated:s"
+    "msg=undefined in comment for method cf in class PrintableHelper:s"
   ).mkString(",")
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -124,7 +124,6 @@ lazy val chiselSettings = Seq(
 
 autoCompilerPlugins := true
 autoAPIMappings := true
-autoAPIMappings := true
 
 // Plugin must be fully cross-versioned (published for Scala minor version)
 // The plugin only works in Scala 2.12+
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -28,11 +28,7 @@ trait ChiselMultiAnnotation {
   def toFirrtl: Seq[Annotation]
 }
 
-/** Mixin for [[ChiselAnnotation]] that instantiates an associated FIRRTL Transform when this Annotation is present
-  * during a run of
-  * [[Driver$.execute(args:Array[String],dut:()=>chisel3\.RawModule)* Driver.execute]].
-  * Automatic Transform instantiation is *not* supported when the Circuit and Annotations are serialized before invoking
-  * FIRRTL.
+/** Mixin for [[ChiselAnnotation]] that instantiates an associated Transform when this Annotation is present
   */
 trait RunFirrtlTransform extends ChiselAnnotation {
   def transformClass: Class[_ <: Transform]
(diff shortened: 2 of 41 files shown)


build.sbt
```diff
@@ -51,6 +51,19 @@ lazy val fatalWarningsSettings = Seq(
...
+    "msg=migration to the MLIR:s",
+    "msg=method hasDefiniteSize in trait IterableOnceOps is deprecated:s",  // replacement `knownSize` is not in 2.12
+    "msg=object JavaConverters in package collection is deprecated:s",
+    "msg=INTERNALskipFile:s",
+    "msg=undefined in comment for method cf in class PrintableHelper:s",
+    "msg=method checkLengths in class StringContext is deprecated:s"
+  ).mkString(",")
+)
+
+lazy val docWarningSuppression = Seq(
```
jackkoenig, line 66: `docWarningSuppression` isn't used, did you mean to just add a few things rather than fully duplicate?

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -81,7 +77,7 @@ object annotate {
   *
   * @note Calling this on [[Data]] creates an annotation that Chisel emits to a separate annotations
   * file. This file must be passed to FIRRTL independently of the `.fir` file. The execute methods
-  * in [[chisel3.Driver]] will pass the annotations to FIRRTL automatically.
+  * in `chisel3.Driver` will pass the annotations to FIRRTL automatically.
   */
```
jackkoenig, line 81: We should just delete references to `chisel3.Driver` (or perhaps replace them at times with a reference to `ChiselStage`).

In this case, we should just delete this whole `@note` since this API cannot be called on `Data`, it only applies to modules.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -64,7 +64,7 @@ object assert extends VerifPrintMacrosDoc {
     * @param cond condition, assertion fires (simulation fails) on a rising clock edge when false and reset is not asserted
     * @param message optional chisel Printable type message
     *
-    * @note See [[printf.apply(fmt:Printable)]] for documentation on printf using Printables
+    * @note See `printf.apply(fmt:Printable)` for documentation on printf using Printables
```
jackkoenig, line 67:
```suggestion
    * @note See [[printf.apply(pable:chisel3\.Printable)*]] for documentation on printf using Printables
```
I don't love using backticks for a specific function, so I figured out how to make it work, it seems you need no spaces, but correct argument name, and definitely need a `*` at the end. Also need to use `\.` if it seems to want a package included in the type of the argument.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -239,7 +239,7 @@ object assume extends VerifPrintMacrosDoc {
     * @param cond condition, assertion fires (simulation fails) when false
     * @param message optional Printable type message when the assertion fires
     *
-    * @note See [[printf.apply(fmt:Printable]] for documentation on printf using Printables
+    * @note See ``printf.apply(fmt:Printable`` for documentation on printf using Printables
```
jackkoenig, line 242:
```suggestion
    * @note See [[printf.apply(pable:chisel3\.Printable)*]] for documentation on printf using Printables
```

core/src/main/scala/chisel3/connectable/package.scala
```diff
@@ -5,6 +5,9 @@ package chisel3
 import chisel3.internal.{BiConnect, Builder}
 import chisel3.experimental.{prefix, SourceInfo}
 
+/*
+ * * @define colonGreaterEq The "flipped connection operator", or the "backpressure connection operator" between a producer and consumer.
+ */
```
jackkoenig, line 10:
```suggestion
```
I don't think this one is doing anything.

core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -19,11 +19,6 @@ import firrtl.transforms.DontTouchAnnotation
   *   dontTouch(dead) // Marking it as such will preserve it
   * }
   * }}}
-  * @note Calling this on [[Data]] creates an annotation that Chisel emits to a separate annotations
-  * file. This file must be passed to FIRRTL independently of the `.fir` file. The execute methods
-  * in [[chisel3.Driver]] will pass the annotations to FIRRTL automatically.
-  * @note Because this is an optimization barrier, constants will not be propagated through a signal marked as
-  * dontTouch.
   */
```
jackkoenig, line 22:
```suggestion
  * @note Because this is an optimization barrier, constants will not be propagated through a signal marked as
  * dontTouch.
  */
  I agree with deleting that first note but please don't delete the 2nd one.
```

core/src/main/scala/chisel3/experimental/dataview/DataView.scala
```diff
@@ -170,7 +170,7 @@ object DataView {
     )
   }
 
-  /** Provides implementations of [[DataView]] for [[Tuple2]] to [[HWTuple2]] */
+  /** Provides implementations of [[DataView]] for `Tuple2`  to [[HWTuple2]] */
```
jackkoenig, line 173:
```suggestion
  /** Provides implementations of [[DataView]] for [[scala.Tuple2]]]  to [[HWTuple2]] */
```
Can't you do the same thing here you did in `DataProduct.scala`?

core/src/main/scala/chisel3/package.scala
```diff
@@ -215,7 +215,7 @@ package object chisel3 {
       StringContext(t: _*).cf(args: _*)
     }
 
-    /** Custom string interpolator for generating formatted Printables : cf"..."
+    /** Custom string interpolator for generating formatted Printables : the cf interpolator
```
jackkoenig, line 218: This one seems unnecessary but I don't really care either way on it.

plugin/src/main/scala/chisel3/internal/plugin/ChiselPlugin.scala
```diff
@@ -69,8 +69,6 @@ class ChiselPlugin(val global: Global) extends Plugin {
         val filename = option.stripPrefix(arguments.skipFilePluginOpt)
         arguments.skipFiles += filename
         // Be annoying and warn because users are not supposed to use this
```
jackkoenig, line 71:
```suggestion
```
Gotta delete the comment too

core/src/main/scala/chisel3/package.scala
```diff
@@ -250,23 +250,22 @@ package object chisel3 {
...
       *         if a `parts` string contains a backslash (`\`) character
       *         that does not start a valid escape sequence.
       *
-      * @throws IllegalArgumentException
+      * throws IllegalArgumentException
       *         if the number of `parts` in the enclosing `StringContext` does not exceed
       *         the number of arguments `arg` by exactly 1.
       */
-    @nowarn("msg=checkLengths in class StringContext is deprecated")
+    @nowarn("msg=checkLengths in class StringContext is deprecated:s")
```
jackkoenig, line 268: Does adding the `:s` actually do anything here? I thought `nowarn` is always `:s`?

src/main/scala/chisel3/stage/phases/Checks.scala
```diff
@@ -8,8 +8,8 @@ import firrtl.AnnotationSeq
 import firrtl.annotations.Annotation
 import firrtl.options.{Dependency, OptionsException, Phase}
 
-/** Sanity checks an [[firrtl.AnnotationSeq]] before running the main [[firrtl.options.Phase]]s of
-  * [[chisel3.stage.ChiselStage]].
+/** Sanity checks an [[firrtl.AnnotationSeq!]] before running the main [[firrtl.options.Phase]]s of
```
jackkoenig, line 11:
```suggestion
/** Sanity checks an [[firrtl.AnnotationSeq]] before running the main [[firrtl.options.Phase]]s of
```
Weirdly the `!` doesn't seem to break it but it also doesn't seem right.

src/main/scala/chisel3/stage/phases/MaybeInjectingPhase.scala
```diff
@@ -6,7 +6,7 @@ import chisel3.aop.injecting.{InjectStatement, InjectingPhase}
 import firrtl.AnnotationSeq
 import firrtl.options.Phase
 
-/** Run [[InjectingPhase]] if a [[InjectStatement]] is present.
+/** Run `InjectingPhase` if a `InjectStatement`` is present.
```
jackkoenig, line 9:
```suggestion
/** Run `InjectingPhase` if a `InjectStatement` is present.
```

src/main/scala/chisel3/util/Valid.scala
```diff
@@ -55,8 +55,8 @@ class Valid[+T <: Data](gen: T) extends Bundle {
   *   }
   * }}}
   *
-  * To convert this to a "valid" interface, you wrap it with a call to the [[Valid$.apply `Valid` companion object's
-  * apply method]]:
+  * To convert this to a `valid`` interface, you wrap it with a call to the `Valid`.apply `Valid` companion object's
+  * apply method:
```
jackkoenig, line 59:
```suggestion
  * To convert this to a `valid` interface, you wrap it with a call to the `Valid` companion object's
  * apply method:
```

jackkoenig (changes requested):
Generally looks good but I did some sanity checking.

jackkoenig (approved):
I went ahead and applied my feedback so we can get this in and I can crack RC2 tonight

PR #3052, opened 2023-03-02
Remove/make private Deprecated APIs

seldridge:
This takes care of all the easy/straightforward deprecations I could find.  This does not touch `compatibility` mode as @azidar said that he would tackle that.  (Removing `compatibility` mode should enable two internal FIRRTL IR nodes from being made private. Due to their usage their, both `Width` and `Circuit` need to remain public.)


jackkoenig:
I want to release 5.0.0-M1 before we merge this if that's okay, it will serve as an artifact that has a Chisel API near to 3.6.0 but without SFC. I should be able to get it out by tomorrow.

jackkoenig:
I created the v5.0.0-M1 branch but I think we should merge https://github.com/chipsalliance/chisel/pull/3064 for unique SNAPSHOTs before we merge anything major like this.

PR #3044, opened 2023-03-01
Chisel5 publishing

jackkoenig:
Setup initial Chisel 5 SNAPSHOT publishing

~This _almost works_, but I think `sbt-ci-release` only publishes the plugin and not the new `unipublish` project so I need to get that working before we can merge.~ This is now ready to go.

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [NA] Did you add appropriate documentation in `docs/src`?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

  - new feature/API

#### API Impact

No impact

#### Backend Code Generation Impact

No impact

#### Desired Merge Strategy

- Squash

#### Release Notes

Change published artifact from edu.berkeley.cs::chisel3 to org.chipsalliance::chisel. Set up automated publishing for 5.0-SNAPSHOT (to s01.oss.sonatype.org because Chips Alliance is a relatively new Sonatype organization). Merge old `firrtl`, `chisel3-macros`, `chisel3-core`, and `chisel3` artifacts into a single artifact: `chisel`.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (Bug fix: `3.4.x`, [small] API extension: `3.5.x`, API modification or big change: `3.6.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3054, opened 2023-03-02
Fix website contributors code

seldridge:
The contributors seems to want `chisel` and not `chisel3`.


jackkoenig:
Same change was included in https://github.com/chipsalliance/chisel/pull/3044

Issue #2435, opened 2022-03-07
Add Bundle.Init (analogous to VecInit / Bundle.Lit)


jackkoenig:
I sketched out a version of this here: https://scastie.scala-lang.org/D0948BD1SgeZEqLGkVBoJg

Ideally this (and VecInit for that matter) would:
1) compose
2) utilize DataView -- (we probably cannot change VecInit but views here so that you can connect to the underlying fields would be nice)

Regardless, this sketch is useful on its own even without the more robust solution

**Type of issue**: bug report | feature request | documentation | other enhancement

**Impact**: API addition (no impact on existing code)

**Development Phase**: proposal

**What is the use case for changing the behavior?**

Useful API for users

ekiwi:
Bundle expressions is a feature that I recently would have liked to use. Do you know how this is different from Bundle literals though?

jackkoenig:
> Do you know how this is different from Bundle literals though?

Bundle literals require that everything field is initialized to a literal whereas `Bundle.Init` would allow anything. Do the names of the APIs need to be different? Unclear but they are very similar.

ekiwi:
> Do the names of the APIs need to be different? Unclear but they are very similar.

I think `Bundle.Init` seems like a more consistent API and it would be cool if it could replace Bundle literal as long as you make sure to assign all elements.

poddar92:
@jackkoenig anyone working on this issue? is it okay if I pick it up? I would need some info on where we want to add this API and how to integrate it with the whole system.

jackkoenig:
Hi @poddar92 and welcome! Feel free to pick this one up, the sketch in the Scastie should serve as a good starting point. You should also feel welcome to join Chisel dev meetings which are on Mondays at 11 am Pacific Time (https://github.com/chipsalliance/chisel3#chisel-dev-meeting).

The basic idea is to take that sketch and add the implicit class inside of `package object chisel3`: https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala/chisel3/package.scala

You should then add some tests that show that it works similar to the Vec literal tests, a good test to start with is one that does something like this test: https://github.com/chipsalliance/chisel3/blob/bc55ed16bdcc7fcc08d29f8bcbdf89cdefced96a/src/test/scala/chiselTests/VecLiteralSpec.scala#L79

Please add it to a new file though called `BundleInitSpec` in the same directory as that `VecLiteralSpec`.

[...]

PR #3060, opened 2023-03-02
added the BundleInit function in package and BundleInitSpec for testi…

poddar92:
…ng it with one basic test.(doesnt work right now)

### Contributor Checklist

- [y ] Did you add at least one test demonstrating the PR?
- [y ] Did you delete any extraneous printlns/debugging code?
- [y ] Did you specify the type of improvement?

#### Type of Improvement

 

- new feature/API

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -412,4 +412,19 @@ package object chisel3 {
     "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
 
   final val deprecatedPublicAPIMsg = "APIs in chisel3.internal are not intended to be public"
+  
+  object BundleInit {
+    implicit class AddInitConstruct[T <: Record](bun: T) {
+      def Init(fs: (T => (Data, Data))*): T = {
+        chisel3.experimental.requireIsChiselType(bun, "Can only init from clean types")
+        val init = Wire(bun)
+        for (f <- fs) {
+          val (field, value) = f(init)
+          // TODO check that field is a field of init
+          field := value
+        }
+        init
+      }
+    }
+  }
 }
--- a/src/test/scala/chiselTests/BundleInitSpec.scala
+++ b/src/test/scala/chiselTests/BundleInitSpec.scala
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests
+
+import chisel3._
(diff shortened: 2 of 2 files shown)


poddar92:
@jackkoenig I have just created a draft PR; so that we can discuss the issue I was seeing. This is also mentioned in the issue comment. #2435

jackkoenig:
@poddar92 I apologize for not responding to your question in that comment, I must've missed the notification about it. Pasting your question below so we can continue discussion here:

> @jackkoenig do we want to recursively go though the bundle and then initialize them? what if one of the elements is a Vec? 
Should we follow how Bundle.Lit does it?
Also; i see the below error with above sketch:
>
> ```
> [error] /Users/radhikapoddar/chisel3/core/src/main/scala/chisel3/package.scala:395:24: macro implementation not found: materialize
> [error] (the most common reason for that is that you cannot use macro implementations in the same compilation run that defines them)
> [error]         val init = Wire(bun)
> ```
>
> is it because we are trying to assign the bundle directly without defining its type in Wire first? is this related to sourceInfo?
 
This issue is a bit subtle, and for better understanding, I suggest you watch my video describing some of the details of how Chisel's internals work: [Chisel Breakdown 3](https://www.youtube.com/watch?v=4vgSrW7Cdkk).

Basically, yes this is related to `SourceInfo` and the macro that materializes source locators. Both `Wire` and `:=` require a source locator but since you are in Chisel's internals, you're not allowed to materialize source locators pointing to the internals. The solution is to have the function you are adding take a `SourceInfo` implicit argument. Then, both `Wire` and `:=` will use _that_ source locator which will be materialized in the caller's source code. I'll leave a suggestion on your code showing what I mean.

core/src/main/scala/chisel3/package.scala
```diff
@@ -412,4 +412,19 @@ package object chisel3 {
     "this feature will not be supported as part of the migration to the MLIR-based FIRRTL Compiler (MFC). For more information about this migration, please see the Chisel ROADMAP.md."
 
   final val deprecatedPublicAPIMsg = "APIs in chisel3.internal are not intended to be public"
+  
+  object BundleInit {
+    implicit class AddInitConstruct[T <: Record](bun: T) {
+      def Init(fs: (T => (Data, Data))*): T = {
```
jackkoenig, line 418:
```suggestion
      def Init(fs: (T => (Data, Data))*)(implicit sourceInfo: SourceInfo): T = {
```
As I alluded to in my comment, https://github.com/chipsalliance/chisel/pull/3060#issuecomment-1452793826, you need to thread a source locator through so that it can be materialized in the caller's code.

PR #3063, opened 2023-03-02
Disallow --target-dir in ChiselStage$

seldridge:
Refactor firrtl.options.Shell into BareShell and Shell.  This is done to
enable the leaner BareShell to _not_ provide no command line options
other than "--help".  This enables use cases like the ChiselStage$
object which should not have options that are available to a Shell,
e.g., "--target-directory".

Change the ChiselStage$ object to use BareShell which enables this to
now reject options like "--target-directory"/"-td" which have no
effect.  Add a test that "-td" is rejected by emitCHIRRTL.

ht/ @trilorez for the report


jackkoenig:
Backport will require some manual work to firrtl but I think this would be good for 3.6

Github Actions are having trouble so we might have to wait on that.

jackkoenig (approved):
LGTM! Thanks!

jackkoenig:
Disabling automerge because I realized there is something wrong with the publishing, it isn't finding the v5.0.0-M1 tag, would like to fix that really fast

PR #3069, opened 2023-03-03
[CI] Do full unshallow clone for publish job

jackkoenig:
We accidentally published such SNAPSHOTs as `0.0.0+1-2ebb8670-SNAPSHOT` 😬

PR #3070, opened 2023-03-03
Require unshallow clone to publish in SBT

jackkoenig:
Now if you try to publish from a shallow clone, it will error. This check would've helped detect https://github.com/chipsalliance/chisel/pull/3069.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,8 @@ lazy val warningSuppression = Seq(
   ).mkString(",")
 )
 
+// This should only be mixed in by projects that are published
+// See 'unipublish' project below
 lazy val publishSettings = Seq(
   versionScheme := Some("semver-spec"),
   publishMavenStyle := true,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -67,6 +67,13 @@ lazy val publishSettings = Seq(
     </licenses>,
   sonatypeCredentialHost := "s01.oss.sonatype.org",
   sonatypeRepository := "https://s01.oss.sonatype.org/service/local",
+  publish / skip := {
```
seldridge, line 70: It seems a little weird to use `skip` as a hook to error. You may want to add a comment indicating that this is what is going on here.

seldridge (approved):

build.sbt
jackkoenig: Oh yeah I meant to add a comment and totally forgot. Yeah there isn't like a "publishing hook" key so I arbitrarily picked this, definitely needs to be explained.

PR #3062, opened 2023-03-02
Move the Cat object to chisel3 package

chick:
Move Cat function to chisel3 package from chisel3.util
Supercedes PR #3051 

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you add appropriate documentation in `docs/src`?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- code cleanup
- 
#### API Impact

Cat import has moved to chisel3. util.Cat is deprecated

#### Backend Code Generation Impact

No changes

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference. 

#### Release Notes
The Cat operator has been moved to the chisel3 package. The util.Cat operator has been deprecated
 
### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (Bug fix: `3.4.x`, [small] API extension: `3.5.x`, API modification or big change: `3.6.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3075, opened 2023-03-09
Remove Cat from util

sequencer:
This bug(not so sure if is a bug) was introduced by #3062.  
There are a lot of projects importing `chisel.util._`, when user have this importing, `Cat` will become ambiguous, for example:
```
[error] /tmp/playground/dependencies/berkeley-hardfloat/src/main/scala/AddRecFN.scala:94:9: reference to Cat is ambiguous;
[error] it is imported twice in the same scope by
[error] import chisel3.util._
[error] and import chisel3._
```
However if merging this, any codes importing `chisel3.util.Cat` will break.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact
Remove `chisel3.util.Cat` 

None

#### Desired Merge Strategy

#### Release Notes
chisel3.util.Cat is removed

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/util.scala
+++ b/src/main/scala/chisel3/util/util.scala
@@ -11,7 +11,4 @@ package object util {
   type ValidIO[+T <: Data] = chisel3.util.Valid[T]
   val ValidIO = chisel3.util.Valid
   val DecoupledIO = chisel3.util.Decoupled
-
-  @deprecated("Use chisel3.Cat instead", since = "Chisel 3.6.x")
-  val Cat = chisel3.Cat
 }


seldridge (approved):
Nice. 

Given that this has been deprecated since Chisel 3.6, this is fine to remove on `main`.

jackkoenig:
While this is fine to do on main, moving with a deprecated alias in the old place is never a bug. It is true that imports will become ambiguous but that can easily be fixed with a specific import, eg. 
```scala
import chisel3._
import chisel3.util._
import chisel3.Cat
```
Or with a compound import:
```scala
import chisel3.{Cat, _}
import chisel3.util._
```
Or it can be fixed with a suppression:
```scala
import chisel3._
import chisel3.util.{Cat => _, _}
```
It's important that it's possible for code to work across 2 versions so the deprecated alias is necessary. That being said, that's only the case in Chisel 3.6 and it's fine to have it removed here.

PR #3090, opened 2023-03-15
update README publishVersion

jensengrey:
publishVersion is now defined in build.sc

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ sbt test
 ### Running Projects Against Local Chisel
 
 To use the development version of Chisel (`master` branch), you will need to build from source and `publishLocal`.
-The repository version can be found in the [build.sbt](build.sbt) file.
+The repository version can be found in the [build.sc](build.sc) file.
 As of the time of writing it was:
 
 ```scala


README.md
```diff
@@ -303,7 +303,7 @@ sbt test
 ### Running Projects Against Local Chisel
 
 To use the development version of Chisel (`master` branch), you will need to build from source and `publishLocal`.
-The repository version can be found in the [build.sbt](build.sbt) file.
+The repository version can be found in the [build.sc](build.sc) file.
```
jackkoenig, line 306: `build.sbt` is still the main build and what is used for publishing. These instructions are wrong and do need to be updated though, one should instead use:
```
sbt unipublish/publishLocal
```

Also default branch is `main`.

These instructions need work 😬

jensengrey: Ok, that makes sense. I saw that the version number moved from `build.sbt` to `build.sc`. Was trying to clean the readme as I went through checking out and running the code. 

I'd be open to updating and refactoring the readme if you gave me some pointers.

I'll close this pull request.

Issue #2780, opened 2022-10-14
Improve MuxLookup's type declaration to get better type errors


jackkoenig:
**Type of issue**: Feature Request

**Is your feature request related to a problem? Please describe.**

When users misuse [MuxLookup](https://javadoc.io/doc/edu.berkeley.cs/chisel3_2.13/3.5.4/chisel3/util/MuxLookup$.html), they can sub-optimal error messages. Consider:
```scala
val sel = IO(Input(Bool()))
val packet0 = IO(Input(Vec(4, UInt(8.W))))
val packet1 = IO(Input(Vec(4, UInt(8.W))))
val out      = IO(Output(Vec(4, UInt(8.W))))

// Call some operation on our Vecs that turn them into Seqs
val choices = Seq(
  0.U -> packet0.map(_ + 1.U),
  1.U -> packet1.map(_ - 1.U)
)

out := MuxLookup(sel, VecInit.fill(4)(0.U), choices)
```
(Scastie: https://scastie.scala-lang.org/ZuJdv3XLQz61C0wmLy57vw)

This gives the error message:
```
Example.scala:16:10: inferred type arguments [chisel3.UInt,IndexedSeq[chisel3.UInt]] do not conform to method apply's type parameter bounds [S <: chisel3.UInt,T <: chisel3.Data]
  out := MuxLookup(sel, VecInit.fill(4)(0.U), choices)
         ^
Example.scala:16:20: type mismatch;
 found   : chisel3.Bool
 required: S
  out := MuxLookup(sel, VecInit.fill(4)(0.U), choices)
                   ^
Example.scala:16:40: type mismatch;
 found   : chisel3.Vec[chisel3.UInt]
 required: T
  out := MuxLookup(sel, VecInit.fill(4)(0.U), choices)
                                       ^
Example.scala:16:47: type mismatch;
 found   : Seq[(chisel3.UInt, IndexedSeq[chisel3.UInt])]
 required: Seq[(S, T)]
  out := MuxLookup(sel, VecInit.fill(4)(0.U), choices)
                                              ^
four errors found
```
Now the real issue is that `choices` is a `Seq[(Bool, IndexedSeq[UInt])]` instead of `Seq[(Bool, Vec[UInt])]`, but the error message is pretty bad. It is even misleading in that it suggests the first 2 arguments are wrong.

**Describe the solution you'd like**

We could make some minor changes to MuxLookup to improve the error messages here:

[...]

PR #3071, opened 2023-03-06
add MuxLookup.fromEnum

albertchen-sifive:
### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you add appropriate documentation in `docs/src`?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### API Impact

adds `chisel3.util.MuxLookup.fromEnum`

#### Backend Code Generation Impact

none

#### Desired Merge Strategy

Squash

#### Release Notes

add `chisel3.util.MuxLookup.fromEnum`

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (Bug fix: `3.5.x` or `3.6.x` depending on impact, API modification or big change: `5.0.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3095, opened 2023-03-16
add curried MuxLookup.apply, deprecate old apply

albertchen-sifive:
fixes https://github.com/chipsalliance/chisel/issues/2780

### Contributor Checklist

#### Type of Improvement

new feature/API

#### API Impact

deprecates the current `MuxLookup.apply` and adds two overloads which behave the same as the old `MuxLookup.apply` but take two parameter lists instead. `MuxLookup.fromEnum` is replaced with the new `MuxLookup.apply`.

none, but should generate better source locators for `MuxLookup.apply`

#### Desired Merge Strategy

Squash

#### Release Notes

Add a new version of `MuxLookup.apply` that takes two parameter lists instead of one. This helps the scala compiler report better type errors.

- [ ] Did you review?


src/main/scala/chisel3/util/Mux.scala
```diff
@@ -49,21 +52,47 @@ object PriorityMux {
...
+  def apply[T <: Data](key: UInt, default: T)(mapping: Seq[(UInt, T)]): T =
+    macro MuxLookupTransform.applyCurried[UInt, T]
+
   /** @param key a key to search for
     * @param default a default value if nothing is found
     * @param mapping a sequence to search of keys and values
     * @return the value found or the default if not
     */
-  def apply[S <: UInt, T <: Data](key: S, default: T, mapping: Seq[(S, T)]): T = {
+  def apply[T <: Data](key: EnumType, default: T)(mapping: Seq[(EnumType, T)]): T =
```
mwachs5, line 92: Shouldn't this still have some type E that extends an enumtype? Like I should not be able to mix days of the week with months

src/test/scala/chiselTests/MuxSpec.scala
```diff
@@ -43,14 +43,14 @@ class MuxLookupEnumTester extends BasicTester {
...
+  assert(MuxLookup(TestEnum.b, 3.U)(mapping) === 1.U)
+  assert(MuxLookup(TestEnum.c, 3.U)(mapping) === 2.U)
 
   val incompleteMapping = Seq(TestEnum.a -> 0.U, TestEnum.c -> 2.U)
-  assert(MuxLookup.fromEnum(TestEnum.a, 3.U)(incompleteMapping) === 0.U)
-  assert(MuxLookup.fromEnum(TestEnum.b, 3.U)(incompleteMapping) === 3.U)
-  assert(MuxLookup.fromEnum(TestEnum.c, 3.U)(incompleteMapping) === 2.U)
+  assert(MuxLookup(TestEnum.a, 3.U)(incompleteMapping) === 0.U)
+  assert(MuxLookup(TestEnum.b, 3.U)(incompleteMapping) === 3.U)
+  assert(MuxLookup(TestEnum.c, 3.U)(incompleteMapping) === 2.U)
```
mwachs5, line 53: Like a case where I define anotherEnum I should not be able to put AnotherEnum.c in for the default or any of the cases, I think.

mwachs5 (commented):
Are you intentionally relaxing the constraint that the enumerated should all be of the same type?

src/main/scala/chisel3/util/Mux.scala
albertchen-sifive: ah yeah that's still a useful constraint to have

src/test/scala/chiselTests/MuxSpec.scala
jackkoenig: Agreed--we really need a testing flow for negative compile tests...

I have a prototype of the build flow to make this work, but we need some pithy way to express what we expect. Either a standalone `FileCheck`, a simpler custom similar thing, or perhaps Linux `expect`

src/main/scala/chisel3/util/Mux.scala
```diff
@@ -49,21 +52,47 @@ object PriorityMux {
...
+    macro MuxLookupTransform.applyCurried[UInt, T]
+
   /** @param key a key to search for
     * @param default a default value if nothing is found
     * @param mapping a sequence to search of keys and values
     * @return the value found or the default if not
     */
-  def apply[S <: UInt, T <: Data](key: S, default: T, mapping: Seq[(S, T)]): T = {
+  def apply[S <: EnumType, T <: Data](key: S, default: T)(mapping: Seq[(S, T)]): T =
+    apply[T](key.asUInt, default)(mapping.map { case (s, t) => (s.asUInt, t) })
```
jackkoenig, line 93: Can we have a source locator on this one too?

src/test/scala/chiselTests/MuxSpec.scala
albertchen-sifive: we could use a macro like the one shapeless uses https://github.com/milessabin/shapeless/blob/main/core/shared/src/main/scala/shapeless/test/typechecking.scala

jackkoenig: ScalaTest has something similar, but macros don't work super well because they aren't full separate invocations of the compiler and thus can't take into account differing compile options

jackkoenig:
Since this deletes an API add in https://github.com/chipsalliance/chisel/pull/3071 and that was backported to 3.6.x and 3.5.x, we need to backport this. As is, this won't be binary compatible because of the change to the deprecated apply method, but that can be fixed in the 3.5.x backport (we can just remove the use of the macro and it won't have a good source locator--it's fine).

jackkoenig (approved):
LGTM! Nice!

jackkoenig:
I pushed a commit that makes sure the `do_` methods are properly segregated in the ScalaDoc to make it cleaner.

PR #3102, opened 2023-03-16
[nfc] Remove two unused SFC annotations

seldridge:
These annotations are no longer used now that they cannot affect CIRCT execution.


seldridge:
This is currently stacked on #3101.

jackkoenig (approved):
LGTM

Note mostly to myself: no need or ability to backport deprecation because these are in package `firrtl`.

PR #3107, opened 2023-03-17
Unified className

azidar:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -1180,14 +1180,6 @@ abstract class Record extends Aggregate {
 
   def elements: SeqMap[String, Data]
 
-  /** Name for Pretty Printing */
-  def className: String = try {
-    this.getClass.getSimpleName
-  } catch {
-    // This happens if your class is defined in an object and is anonymous
-    case e: java.lang.InternalError if e.getMessage == "Malformed class name" => this.getClass.toString
-  }
-
   private[chisel3] override def typeEquivalent(that: Data): Boolean = that match {
     case that: Record =>
       this.getClass == that.getClass &&
@@ -1322,15 +1314,10 @@ abstract class Bundle extends Record {
       "Please see https://github.com/chipsalliance/chisel3#build-your-own-chisel-projects."
   assert(_usingPlugin, mustUsePluginMsg)
 
-  override def className: String = try {
-    this.getClass.getSimpleName match {
-      case name if name.startsWith("$anon$") => "AnonymousBundle" // fallback for anonymous Bundle case
-      case ""                                => "AnonymousBundle" // ditto, but on other platforms
-      case name                              => name
-    }
-  } catch {
-    // This happens if you have nested objects which your class is defined in
-    case e: java.lang.InternalError if e.getMessage == "Malformed class name" => this.getClass.toString
+  override def className: String = super.className match {
+    case name if name.startsWith("$anon$") => "AnonymousBundle" // fallback for anonymous Bundle case
+    case ""                                => "AnonymousBundle" // ditto, but on other platforms
+    case name                              => name
   }
 
   /** The collection of [[Data]]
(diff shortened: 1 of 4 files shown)


mwachs5 (commented):
Should we add some test cases for this?
And cover something like:

* Anonymous Class
* Class inside an Object
* Anonymous extended class val foo = new Module(){...}
* ...? Anything else?

src/test/scala/chiselTests/naming/ClassNameSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+
+class ClassNameSpec extends ChiselFunSpec with Utils {
+  object DynamicObject {
+    class MyDynamicRecord extends Record { val elements = SeqMap.empty }
+    class MyDynamicBundle extends Bundle
+    class MyDynamicModule extends Module
+  }
+
+  describe("(0) Dynamic declarations") {
+    it("(0.a): name of an dynamic class of Module is anonymous") {
```
mwachs5, line 27: how is this a test of anonymous? you provided the name MyDynamicModule...?

mwachs5: to me a test of anonymous would be something like:

val foo = new Module { val io = IO(...)} 
assert(foo.chiselClassName = ???)

src/test/scala/chiselTests/naming/ClassNameSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+        val x = new StaticObject.MyStaticModule {}
+        assert(x.chiselClassName == "ClassNameSpec_Anon")
+        x
+      }
+    }
+    it("(2.b): name of an dynamic class of Record") {
+      ChiselStage.elaborate(
+        new Module {
+          val x = new StaticObject.MyStaticRecord {}
+          assert(x.chiselClassName == "")
```
mwachs5, line 90: why not AnonymousRecord?

src/test/scala/chiselTests/naming/ClassNameSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+        }
+      )
+    }
+    it("(2.c): name of an dynamic class of Bundle") {
+      ChiselStage.elaborate(
+        new Module {
+          val x = new StaticObject.MyStaticBundle {}
+          assert(x.chiselClassName == "AnonymousBundle")
+        }
+      )
```
mwachs5, line 100: do we need any cases where the user deliberately overrides the name with the expected API(s)?

src/test/scala/chiselTests/naming/ClassNameSpec.scala
azidar: That's the current behavior. I just wanted to consolidate all the places we do class name reflection into one function, then we can change its behavior in a followup PR.

src/test/scala/chiselTests/naming/ClassNameSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+          val x = new StaticObject.MyStaticRecord {}
+          assert(x.chiselClassName == "")
+        }
+      )
+    }
+    it("(2.c): name of an dynamic class of Bundle") {
+      ChiselStage.elaborate(
+        new Module {
+          val x = new StaticObject.MyStaticBundle {}
+          assert(x.chiselClassName == "AnonymousBundle")
```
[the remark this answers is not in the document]
azidar: Because that's the current behavior. I think a separate PR can propose a new, standard way of naming everything consistently.

src/test/scala/chiselTests/naming/ClassNameSpec.scala
azidar: No, because this is only querying the reflection name result. Other APIs can then use this to customize the name themselves (e.g. naming conflict resolution will append stuff).

PR #3130, opened 2023-03-22
Implement typeName API for stable Module names

jared-barocsi:
`typeName` is a new member of `Data` types which specifies a default, non-ambiguous name of its type. This is useful for problems like generating a descriptive name for a Chisel type, as it allows modules (e.g. `Queue`) to use this naming information for their own `desiredName`. 

As a motivating use case, `Queue` modules now have a default name based on the depth parameter of the `Queue` as well as the generated `typeName` of the Chisel type passed to it. For example, `Queue4_UInt4` describes a `Queue` of size 4 which takes a `UInt(4.W)`, and would be the default `desiredName` of such a queue.

User-defined types like `Bundles` should override this feature if needed to diversify the names of `Modules` that use them, preventing situations like multiple `Queues` all sharing the same name and having unstable names as a result.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

Adds a new public-facing feature to `Data`

Module names in FIRRTL will now include the parameter information of their respective Chisel modules in their names (via overriding `desiredName`), and consequently Verilog names will differ as well.

#### Desired Merge Strategy

Squash and merge

#### Release Notes

Implement `typeName API` for Chisel `Data` types
- This gives a flexible way to generate a stable name for a Chisel type, which is useful for problems like generating stable names for `Modules` and `Queues`

- [ ] Did you review?


core/src/main/scala/chisel3/Bits.scala
```diff
@@ -1054,6 +1059,11 @@ object AsyncReset {
 sealed class AsyncReset(private[chisel3] val width: Width = Width(1)) extends Element with Reset {
   override def toString: String = stringAccessor("AsyncReset")
 
+  /** A non-ambiguous name of this `AsyncReset` for use in generated Verilog names
+    * Inserts the width directly after the typeName, e.g. AsyncReset1
+    */
+  override def typeName = s"${this.getClass.getSimpleName}$width"
```
mwachs5, line 1065: i dont think async resets can have any width that is not 1?

core/src/main/scala/chisel3/Clock.scala
```diff
@@ -17,6 +17,11 @@ object Clock {
 sealed class Clock(private[chisel3] val width: Width = Width(1)) extends Element {
   override def toString: String = stringAccessor("Clock")
 
+  /** A non-ambiguous name of this `Clock` for use in generated Verilog names
+    * Inserts the width directly after the typeName, e.g. Clock1
+    */
+  override def typeName = s"Clock$width"
```
mwachs5, line 23: i dont think clocks can have any width that isn't 1

src/main/scala/chisel3/util/Valid.scala
```diff
@@ -173,6 +178,15 @@ object Pipe {
   */
 class Pipe[T <: Data](val gen: T, val latency: Int = 1) extends Module {
 
+  /** A non-ambiguous name of this `Pipe` for use in generated Verilog names.
+   * Includes the cycle count in the name as well as the parameterized generator's
+   * `typeName`, e.g. `Pipe_4cycles_UInt4`
+   */
+  override def desiredName = {
+    val latencyStr = s"${latency}cycle${if (latency > 1) "s" else ""}"
+    s"${this.getClass.getSimpleName}_${latencyStr}_${gen.typeName}"
```
mwachs5, line 187: why does this one have an underscore before latency, whereas queue depth does not have an underscore? In general what is the suggested pattern for naming things?

src/test/scala/chiselTests/naming/TypenameSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+
+      val barEnq = Wire(Decoupled(Bool()))
+      val barDeq = Queue(barEnq, 5) // Queue5_Bool
+      barEnq <> io.bar
+      
+      val fizzbuzzEnq = Wire(Decoupled(Vec(3, UInt(8.W))))
+      val fizzbuzzDeq = Queue(fizzbuzzEnq, 32) // Queue32_Vec3_UInt8
+      fizzbuzzEnq <> io.fizzbuzz
+    }
+
```
mwachs5, line 29: can we show Clock, Reset, AsyncReset, Analog

mwachs5 (commented):
Should you update the naming cookbook, especially https://github.com/chipsalliance/chisel/blob/main/docs/src/cookbooks/naming.md#my-module-names-are-super-unstable---i-change-one-thing-and-queue_1-becomes-queue_42-help after this change?

jared-barocsi:
> Should you update the naming cookbook, especially https://github.com/chipsalliance/chisel/blob/main/docs/src/cookbooks/naming.md#my-module-names-are-super-unstable---i-change-one-thing-and-queue_1-becomes-queue_42-help after this change?

I think I will, but also clarify that queue names will still have some measure of instability based on the specificity of the `typeName` that's being used, and that users should add enough relevant parameter information to a `typeName` as necessary.

src/main/scala/chisel3/util/Valid.scala
mwachs5: could even hash the parameters or something

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,43 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
-### My module names are super unstable - I change one thing and Queue_1 becomes Queue_42. Help!
+### My module names are super unstable - I change one thing and Module_1 becomes Module_42. Help!
 
-This is the infamous `Queue` instability problem. In general, these cases are best solved at the source - the module
-itself! If you overwrite `desiredName` to include parameter information (see the
-[explanation](../explanations/naming#set-a-module-name) for more info), then this can avoid this problem permanantly.
-We've done this with some Chisel utilities with great results!
+This is an example of the module instability problem, which results from several modules all sharing the exact same name. To fix this, you must add more specificity to your `Module`'s name to avoid these name collisions.
+
+This can be done by leveraging the `typeName` API, which provides parameter information from a `Data` object for use in a Module's `desiredName` -- thus reducing name collisions and fixing this problem permanently. For instance, suppose your module looks like the following:
```
mwachs5, line 79:
```suggestion
This can be done by leveraging the `desiredName` and `typeName` APIs. 
`desiredName` is for indicating the names of `Modules` (e.g. influenced by the parameters passed in), and `typeName` is useful for modules which are type-parameterized by subclasses of `Data`. Overriding `desiredName` can reduce or even eliminate name collisions. For instance, suppose your module looks like the following:
```

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,43 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+```scala mdoc
+val foo = Module(new MyModule(UInt(4.W))) // MyModule_UInt4
+val foo = Module(new MyModule(Vec(3, UInt(4.W)))) // MyModule_Vec3_UInt4
+```
+
+
+All default Chisel modules, like `Queue` and `Pipe`, already have their `desiredName` overrided in this manner, which solves the infamous `Queue` stability problem.
+
+### I have already overriden `desiredName` to use a `typeName` but my module names are still conflicting!
+
```
mwachs5, line 106: this is a cookbook and should have more on code less on explanation :) show them the before-and-after of a Queue using their user defined type with them overriding the type name Bundle's type name and how it will influence the Queue.

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,43 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+You either must add additional information to the `desiredName`, or if you're using your own user-defined `Bundle`, increase the specificity of its own `typeName`. All `Data` types have a simple default implementation of `typeName` (which is simply their own name), but you can override this yourself, of course!
+
+In general, the suggested pattern for `typeName`, and subsequently `desiredName`, is to fold single integer-like parameters with the name itself (for example, `Queue4`, `UInt3`, `MyBundle9`) and separate these with underscores (`Queue4_UInt3`, `FooBundle_BarType4`).
+
+Integers should not occur with an underscore before it at the very end of the name (`MyBundle_1`) because this is the _same_ syntax used for duplicates, and so would cause confusion. Having to disambiguate modules all named `FooModule_MyBundle_4_1`, `FooModule_MyBundle_4_2`, `FooModule_MyBundle_4_3`, and so on would be undesirable, indeed!
```
mwachs5, line 111: so what *should* you do for multiple integer parameters? We haven't had to address this, so you can either say "No Chisel library components have to deal with it, it's an excercise left to the reader" or say something like "No Chisel library components have to deal with it, but if you have multiple integer parameters suggest you prefix them with a letter, like `FooModule_MyBundle4_p3_b3_p3` ...?

i think for this PR i'd rather we just have no stance on it.

src/test/scala/chiselTests/naming/TypenameSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+      val clockEnq = Wire(Decoupled(Clock()))
+      val clockDeq = Queue(clockEnq, 20) // Queue20_Clock
+      clockEnq <> io.clock
+
+      val analogTest = Module(new AnalogTest(Analog(32.W)))
+      analogTest.bus <> io.analog
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(new Test)
+    chirrtl should include("module Queue16_UInt4")
```
mwachs5, line 57: can we add an SInt

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,44 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+class MyModule[T <: Data](gen: T) extends Module {
+  val in = IO(Input(gen))
+  val out = IO(Output(gen))
+  out := in
+}
+```
+
+We can override `desiredName` to include the type name of the `gen` parameter like so:
+
+```scala mdoc
```
mwachs5, line 92: I'd probably show this as a subclass of MyModule, or just write it out again (you can use a bunch of tags to reset the compilation context, or just give it a new name) , so that the mdoc will actually compile. If this is how you want it, you can just omit the mdoc tag.

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,44 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+override def desiredName = s"MyModule_${gen.typeName}"
+```
+
+Any instances of your `MyModule` will now have Verilog module names containing the type parameter.
+
+```scala mdoc
```
mwachs5, line 98: This won't execute work because you aren't in a builder context, you can still do `mdoc:compileOnly`

docs/src/cookbooks/naming.md
jared-barocsi: > you can use a bunch of tags to reset the compilation context

How would I do this? I'd like to show that the `MyModule` example can be supplemented with overriding `desiredName` to demonstrate use of the `typeName` API

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,88 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+We can override `desiredName` of the module to include the type name of the `gen` parameter like so:
+
+```scala
+override def desiredName = s"MyModule_${gen.typeName}
+}
+```
+
+Any instances of your `MyModule` will now have Verilog module names containing the type parameter.
+
+```scala mdoc:compileOnly
```
mwachs5, line 97: should be:
`scala mdoc:compile-only`

sorry I had the wrong suggestion

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,91 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
 `_GEN` signals are usually generated from the FIRRTL compiler, rather than the Chisel library. We are working on
 renaming these signals with more context-dependent names, but it is a work in progress. Thanks for caring!
 
-### My module names are super unstable - I change one thing and Queue_1 becomes Queue_42. Help!
```
mwachs5, line 75: lets make sure to add these topics to the ToC at the top.

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,91 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+
+Previously, the recommended solution was to manually override `desiredName` in a `Queue` for extra specificity, but now, both the `typeName` of a `Queue`'s inner type and the Queue's depth are automatically included in its `desiredName`, with *almost* no changes:
+
+```scala
+val fooQueue = Queue(fooEnq, 8) // Verilog module would be named 'Queue8_UInt8'
+val barQueue = Queue(barEnq, 8) // ... and 'Queue8_SInt8'
+val bazQueue = Queue(bazEnq, 8) // ... and 'Queue8_Bool'
+```
+
+### I have already overriden `desiredName` to use a `typeName` but my module names are still conflicting!
```
mwachs5, line 129:
```suggestion
### How can I avoid name conflicts on Chisel library modules like `Queue`?
```

docs/src/cookbooks/naming.md
```diff
@@ -72,12 +72,91 @@ ChiselStage.emitSystemVerilog(new ExampleWhenPrefix)
...
+```
+
+Any instances of your `MyModule` will now have Verilog module names containing the type parameter.
+
+```scala mdoc:compile-only
+val foo = Module(new MyModule(UInt(4.W))) // MyModule_UInt4
+val bar = Module(new MyModule(Vec(3, UInt(4.W)))) // MyModule_Vec3_UInt4
+```
+
+### Does the `typeName` apply for built-in module names too, most notably `Queue`?
```
mwachs5, line 105: Sinc this isn't really a "recipe", I'd probably just move all this underneath the next topic maybe after the quick 'recipe' answer as a deeper explanation

docs/src/cookbooks/naming.md
mwachs5: correction: there is no ToC :) we could add in a follow-up

src/test/scala/chiselTests/naming/TypenameSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+    chirrtl should include("module Queue8_SInt8")
+    chirrtl should include("module Queue5_Bool")
+    chirrtl should include("module Queue32_Vec3_UInt8")
+    chirrtl should include("module Queue17_AsyncReset")
+    chirrtl should include("module Queue3_Reset")
+    chirrtl should include("module Queue20_Clock")
+    chirrtl should include("module AnalogTest_Analog32")
```
mwachs5, line 69: one more clarifying question -- what happens if you do a inferred-width thing, like `val queue = Module(new Queue(UInt(), 3)); queue.enq.bits := 0.U(4.W)`? Is the name just `Queue3_UInt` (which is fine, but i dont see that case)

mwachs5 (approved):
LGTM!

PR #3131, opened 2023-03-23
SRAM Intrinsic

sequencer:
### Contributor Checklist

- [ ] Did you review?


ekiwi:
I agree with the need for a new Chisel interface to create memories. However, why do we need an intrinsic for that instead of using the standard `firrtl` memories with their well-defined behavior in the backend?

sequencer:
> I agree with the need for a new Chisel interface to create memories. However, why do we need an intrinsic for that instead of using the standard firrtl memories with their well-defined behavior in the backend?

Sorry for the late reply, I was in the hospital these days.

I just read the spec, I think this should be a good idea, but I think there are some issues in spec that we need concern:
- spec should take the additional port into consideration for BIST or DFT designs. CIRCT currently has a dirty way for it, see [AddSeqMemPortAnnotation](https://github.com/llvm/circt/blob/aceaba0442967a9630fd05276ad4dd6bd0aab83b/docs/Dialects/FIRRTL/FIRRTLAnnotations.md?plain=1#L344) and [AddSeqMemPortsFileAnnotation](https://github.com/llvm/circt/blob/aceaba0442967a9630fd05276ad4dd6bd0aab83b/docs/Dialects/FIRRTL/FIRRTLAnnotations.md?plain=1#L369). Those are necessary for real world designs.
- The current firrtl memory model doesn't support memory init, which means it has a bad FPGA support.
- personally, I prefer treat memory as a instance, rather a memory type, since in the PnR, it's a hard marco, which should be treat as an instance.

I think this issue should follow #3162, in the future, we should add more metadata to firrtl spec, and switch to it. and then deprecate and remove the memory behavior model get rid of firrtl mem.

sequencer:
I thought about today, here is my solution to this:
- I still think we should use `IntrinsicModule` to instantiate a SRAM, rather than maintaining a `Mem/SyncMem` type in chisel, since unlike register, memory is always a hard macro and has corresponding IOs, it should be a Module/Instance, I think we should waive this 10y~ old design fault.
- For `firrtl.mem`, I think it should still be supported in firrtl spec and firtool: IMO, the correct behavior is: if not replacing macro, CIRCT should use the `firrtl.mem` to replace the intrinsic, and lower in the current way, I think this can solve your problem. In the `repl-seq-mem` flow(Or I prefer `repl-mem`) flow, this Module will be replace by the correct memory macros(even XPM for Xilinx FPGA).

PR #3135, opened 2023-03-24
Added definition identifiers

azidar:
Added compiler plugin to for eagerly computed module definition identifiers. Due to unknown reasons, the plugin fails to determine the identifier for traits (which should be a rare corner case that its used), but I added a macro that users, if they so choose, can use instead.

Right now, nothing uses the identifiers.

By-name-parameters are ignored by the plugin in name generation.

### Contributor Checklist

- [ ] Did you review?


core/src/main/scala/chisel3/internal/Identifier.scala
```diff
@@ -0,0 +1 @@
+
```
jackkoenig, line 1: Is this file supposed to be here?

macros/src/main/scala/chisel3/internal/naming/Identifier.scala
```diff
@@ -0,0 +1,58 @@
...
+            stat match {
+              case aDef: DefDef if aDef.name.toString == "_traitModuleDefinitionIdentifierProposal" =>
+                c.error(aDef.pos, s"Custom implementations of _traitModuleDefinitionIdentifierProposal are not allowed")
+              case _ =>
+            }
+          }
+          val newMethod = q"override protected def _traitModuleDefinitionIdentifierProposal = Some(${tpname.toString})"
+          val newStats = newMethod +: stats
+          (
+            q"$mods trait $tpname[..$tparams] extends { ..$earlydefns } with ..$parents with chisel3.experimental.hierarchy.IsInstantiable { $self => ..$newStats }",
```
jackkoenig, line 36:
```suggestion
            q"$mods trait $tpname[..$tparams] extends { ..$earlydefns } with ..$parents { $self => ..$newStats }",
```
Left over from copy-paste I think.

macros/src/main/scala/chisel3/internal/naming/Identifier.scala
```diff
@@ -0,0 +1,58 @@
...
+          clz
+      }
+      objOpt match {
+        case None => newClz
+        case Some(o) =>
+          q"""
+            $newClz
+
+            $o
+          """
```
jackkoenig, line 49: Do we need to deal with the companion object in this macro? We aren't transforming it right?

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
```diff
@@ -0,0 +1,122 @@
...
+          primaryConstructor = Some(con)
+        case d: DefDef if isNullaryMethodNamed("_moduleDefinitionIdentifierProposal", d) =>
+          val msg = "Users cannot override _moduleDefinitionIdentifierProposal. Let the compiler plugin generate it."
+          global.reporter.error(d.pos, msg)
+        case _ =>
+      }
+      (primaryConstructor, paramAccessors.toList)
+    }
+
+    def generateIdentifierMethod(module: ClassDef, thiz: global.This, baseClass: global.Symbol): Tree = {
```
jackkoenig, line 63: I don't love how much code we've duplicated between the other components and this one, maybe we can make a utils object?

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
```diff
@@ -0,0 +1,122 @@
...
+      localTyper.typed(DefDef(identifierSym, body)).asInstanceOf[DefDef]
+    }
+
+    override def transform(tree: Tree): Tree = tree match {
+
+      case module: ClassDef
+          if isAModule(module.symbol)
+            && !isExactBaseModule(module.symbol)
+            && !module.mods.hasFlag(Flags.TRAIT)
+            && !module.name.decode.contains("$anon") =>
```
jackkoenig, line 102: Does this not work on anonymous functions?

core/src/main/scala/chisel3/naming/Identifier.scala
```diff
@@ -0,0 +1,49 @@
+package chisel3.naming
+
+trait IdentifierProposer[T] {
+  def propose(obj: T): String
+}
```
jackkoenig, line 5:
```suggestion
sealed trait IdentifierProposer[T] {
  def propose(obj: T): String
}
```

Since it doesn't seem to be useful for users to extend this, I think it should be private. Some day I hope we can make it user implementable, but I don't want to give the appearance of doing so when we don't yet.

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import chisel3._
+import chisel3.naming.IdentifierProposer
+
+case class Blah(i: Int)
+class Foo(val i: Int)
+object Foo {
+  implicit val fooProposer = new IdentifierProposer[Foo] {
+    def propose(foo: Foo): String = s"Foo(${foo.i})"
+  }
```
jackkoenig, line 14: This appears to be unused, related to my comments about the "Typeclass-y API not being implemented [yet]".

jackkoenig (commented):
Generally looks good, lots of little suggestions.

I hope we can ultimately figure out how to make this work as a user-implementable typeclass, but it seems you went halfway down that route but the actually implementation is basically a hardcoded implementation. That's fine, but we should make sure not to expose the typeclass-y implementation if it isn't actually supported (yet).

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+}
+class Baz(val i: Int)
+
+class IdentifierProposerSpec extends ChiselFunSpec with Utils {
+  it("(0): getProposal works on a variety of object types") {
+    assert(IdentifierProposer.getProposal("Hi") == "Hi")
+    assert(IdentifierProposer.getProposal(1) == "1")
+    assert(IdentifierProposer.getProposal(Blah(1)) == "Blah_1")
+    assert(IdentifierProposer.getProposal(new Bundle {}) == "AnonymousBundle")
+    assert(IdentifierProposer.getProposal(new Foo(1)) == "Foo_1")
```
mwachs5, line 24: what does it do for a Foo with multiple parameters?

mwachs5, line 24: shouldn't this be `Foo(1)` given the above proposer?

src/test/scala/chiselTests/naming/IdentifierSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+class HasByNameArg(g: => Module) extends Module {
+  val x = Module(g)
+}
+
+class IdentifierSpec extends ChiselFunSpec with Utils {
+  it("(1): definitionIdentifier works on classes, abstract classes, but not traits") {
+    class TopA extends Module {
+      assert(Module(new ConcreteClass(0)).definitionIdentifier == "ConcreteClass_0")
+      assert(Module(new ConcreteClass(0) {}).definitionIdentifier == "ConcreteClass_0_1")
```
mwachs5, line 34: what is the `_1` coming from? is this the auto-namespace-collision-avoidance? Can we just have an error if there is a namespace collision in the space of `definitionIdentifier`?

src/test/scala/chiselTests/naming/IdentifierSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+      assert(Module(new NormalTrait {}).definitionIdentifier == "_1_Anon") // Traits don't work, Scala compiler bug?!?
+      assert(Module(new AbstractClass {}).definitionIdentifier == "AbstractClass")
+      assert(Module(new AbstArgsClass(1) {}).definitionIdentifier == "AbstArgsClass_1")
+      assert(Module(new NestedAbtrClz(1) {}).definitionIdentifier == "NestedAbtrClz_1")
```
mwachs5, line 38: is this `_1` due to the argument being passed or some sort of namespace dedup? Can we make these things "look different" (see the type name API docs) or make the second one illegal?

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
azidar: One issue is all the types are imported from `import global._`, so its hard to actually do this well.

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
azidar: It gets filtered to be a clean name without "(", see `def filterProposal` on `object IdentiferProposer`

src/test/scala/chiselTests/naming/IdentifierSpec.scala
azidar: We could but I'd prefer to do that in a future version, right now its not used anywhere, so I don't want to just cause additional errors.

src/test/scala/chiselTests/naming/IdentifierSpec.scala
azidar: Sure

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
jackkoenig: In the utils, you can take `global` as an argument and use it to derive types, see https://github.com/chipsalliance/chisel/blob/3a59415380a75085286e5cbfd006323ea04ea02d/core/src/main/scala/chisel3/Printf.scala#L39

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
jackkoenig: Are you sure it's being used? Because `Blah` is getting the exact same style of name below. I'd like to see an example that does an actually interesting proposal to prove it's being used. I'm skeptical because I had an extremely difficult time getting implicit resolution to work when generating the methods in the compiler plugin. I suspect it _is_ possible, but I don't see it here yet. And even if this case works (because it's defined in implicit scope), we need to show that the implicit resolution follows the rules, try putting the type class definition in a different object and importing it into lexical scope--that was the case I could not get to work.

macros/src/main/scala/chisel3/internal/naming/Identifier.scala
azidar: I thought I still needed to return it, if it exists?

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
azidar: Anonymous classes. There doesn't exist a new name for `new MyModule {}` that is better that whatever was derived for `MyModule`, so there is no point in overriding the `_moduleDefinitionIdentifierProposal` for this anonymous class.

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
azidar: Yeah it works, any other cases I should test?

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
azidar: Just appends it, like the tuple or list.

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
azidar: Sounds good.

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
jackkoenig: Is there a better way to detect it than String matching the name?

src/test/scala/chiselTests/naming/IdentifierProposerSpec.scala
jackkoenig: I realize I commented in the wrong place, it of course works when you call `IdentifierProposer.getProposal`, but it does **not** work for actually customizing the name of a Module, the test that I'm concerned about it:
```scala
class MyModule(i: Int) extends Module
object MyModule {
  implicit val proposer = new IdentifierProposer[MyModule] {
    def propose(x: MyModule): String = "SomeOtherName"                                                                                                                                                         
  }
}

Module(new MyModule(0)).definitionIdentifier // This should be "SomeOtherName" but is "MyModule_0"
```
I want to do the typeclass API, but if it doesn't work yet, I'm not sure that we should expose it as public.

azidar: After offline conversation w/ @jackkoenig , we decided the typeclass customization would be too onerous to do because you'd need to pepper `: IdentiferProposer` everywhere, so instead we will just use inheritance.

plugin/src/main/scala/chisel3/internal/plugin/IdentifierComponent.scala
azidar: Not that I can figure out, sadly.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -261,6 +262,30 @@ package experimental {
...
+        .split("\\.|\\$") // 2. Split string at '.' or '$'
+        .filterNot(_.forall(_.isDigit)) // 3. Drop purely numeric names
+        .last // 4. Use the last name
+    }
+    // Needed this to override identifier for DefinitionClone
+    private[chisel3] def _definitionIdentifier = {
+      val madeProposal = chisel3.naming.IdentifierProposer.makeProposal(this._moduleDefinitionIdentifierProposal)
+      Builder.globalIdentifierNamespace.name(madeProposal)
+    }
+    // Represents an eagerly-determined unique and descriptive identifier for this module
```
jackkoenig, line 287:
```suggestion
    /** Represents an eagerly-determined unique and descriptive identifier for this module */
```
Gotta be ScalaDoc yo

plugin/src/main/scala/chisel3/internal/plugin/ChiselUtils.scala
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.internal.plugin
+
+import scala.tools.nsc
+import scala.tools.nsc.{Global, Phase}
+import scala.tools.nsc.transform.TypingTransformers
+
+trait ChiselOuterUtils { outerSelf: TypingTransformers =>
```
jackkoenig, line 9:
```suggestion
private[plugin] trait ChiselOuterUtils { outerSelf: TypingTransformers =>
```

jackkoenig (approved):
LGTM with a couple of nits

PR #3137, opened 2023-03-27
Remove var _parent, add indirection through Builder

azidar:
### Contributor Checklist

- [ ] Did you review?


jackkoenig:
I am generally in favor of removing fields from `Data` where possible. That being said, Maps are more expensive on a per-member basis than a field, so if a high percentage of `Data` use this field, a Map may be more expensive. In particular, bound Data have it set (does this include children of Aggregates? I am not sure), and bound Data cannot be garbage collected, thus they make up a disproportionate percentage of the `Data` objects impacting overall memory use.

This is a very long-winded way of saying that we need to benchmark this change to see if it's worthwhile.

PR #3140, opened 2023-03-28
Added Context, with some tests

azidar:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Context.scala
+++ b/core/src/main/scala/chisel3/internal/Context.scala
@@ -1,0 +1,189 @@
+package chisel3.internal
+
+private[internal] trait CloneToContext {
+  def cloneTo(c: Context): CloneToContext
+}
+
+// Datastructure to represent contexts (with back pointers to "provenance", or version of this context with less context information)
+private[chisel3] class Context(val key: String, val parent: Option[Context], val provenance: Option[Context]) {
+  private val localChildrenKeys = new collection.mutable.LinkedHashSet[String]()
+  private val children = new collection.mutable.LinkedHashMap[String, Context]()
+
+  /** At this moment in time, all known children keys of this or any provenance Contexts */
+  def childrenKeys: List[String] = if (isOrigin) localChildrenKeys.toList else provenance.map(_.childrenKeys).get
+  def isChild(childKey: String): Boolean = childrenKeys.contains(childKey)
+
+  /* Accessors to important contexts, relative to this */
+
+  // Top-most context
+  lazy val top: Context = parent.map(_.top).getOrElse(this)
+  // Provenance origin, could have a different top context
+  lazy val origin: Context = provenance.map(_.origin).getOrElse(this)
+  // Provenance origin, but who shares same top context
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/Context.scala
```diff
@@ -0,0 +1,189 @@
...
+  // entire provenance context path
+  lazy val provenanceContextPath: Seq[Context] = (this +: provenance.toSeq.flatMap(_.provenanceContextPath))
+
+  /* Helpers to determine whether this Context has these properties */
+
+  lazy val isTop = parent.isEmpty
+  lazy val isOrigin = origin == this
+  lazy val isTemplate = template == this
+
+  /* Adding children to origin new Contexts */
```
mwachs5, line 37: what does this comment mean?

azidar: Clarified.

src/test/scala/chisel3/internal/ContextSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      List(
+        "Connection between sink (Bottom.in: IO[UInt<3>]) and source (Bottom_1.out: IO[UInt<3>]) failed",
+        "Top/m0=Middle/b0=Bottom/in", // Generated from Context.target
+        "Top/m0=Middle/b1=Bottom_1/out", // Generated from Context.target
+        "unavailable to current module",
+        "Top" // Generated from Context.target
+      ).forall(s => e.getMessage.contains(s))
+    }
```
jackkoenig, line 79: FYI this test is vacuously passing because there's no actual assertion, that `forall` is returning `false` (because we're obviously not using `Context` yet).

src/test/scala/chisel3/internal/ContextSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+
+      // Right subsumes left
+      assert(R_b0.copyTo(B_l0) == R_b0)
+      assert(R_b0_l1.copyTo(B_l1) == R_b0_l1)
+      assert(R.copyTo(B_l1) == R)
+
+      // Left matches right
+      assert(R_b0.copyTo(R_b0) == R_b0)
+      assert(R_b0_l1.copyTo(R_b0_l1) == R_b0_l1)
+      assert(R.copyTo(R_b0_l1) == R)
```
jackkoenig, line 50: I don't completely understand these tests. You don't override `equals` in `Context` so we're using referential equality. I think I would prefer `eq` to call that out to the reader. Also, these `copyTo` methods mutate the underlying datastructures, I would like to see what mutation is expected captured in the tests.

jackkoenig (approved):
I've got some suggestions, I'd especially like to understand the tests a little better. I think we should merge to make forward progress, but I maintain some reservations about how this is going to be used. If we're going to have one of these instances _per Data_ we're going to have to figure out how to have some structural sharing otherwise the memory cost is going to be astronomical.

PR #3141, opened 2023-03-29
Use %[[]] format for ChsielAnnotations

seldridge:
Change FIRRTL Circuit conversion/serialization to place ChiselAnnotations (converted to FIRRTL Annotations) inside the FIRRTL Circuit and to serialize this using the "%[[]]" inline JSON format that is present in the FIRRTL spec.  This does not remove _all_ annotations passed via the annotation file, but it is the first step towards that.

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -366,11 +366,11 @@ private[chisel3] object Converter {
   }
 
   def convert(circuit: Circuit): fir.Circuit =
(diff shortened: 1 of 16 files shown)


jackkoenig (commented):
Generally LGTM. I would like to see a unit test at least sanity checking the serialization (eg. looking for a dontTouch in the `.fir`) and I think it would be good to control this via a command-line option, WDYT?

seldridge:
> LGTM, I still want to see a unit test

Added a test of both `ChiselStage` and `ChiselStage$`.

seldridge:
Also, the inline Annotation format is still FIRRTL spec 1.2.0, so no bump to the emitted version needed. This is the first hard-break with the SFC.

PR #3160, opened 2023-04-04
Deprecate ChiselStage$.elaborate

seldridge:
Deprecate ChiselStage$.elaborate as this exposes internal Chisel APIs, namely "chisel3.internal.firrtl.Circuit".  Replace usages of this in tests with emitCHIRRTL.

### DIFF
--- a/src/main/scala/circt/stage/ChiselStage.scala
+++ b/src/main/scala/circt/stage/ChiselStage.scala
@@ -211,6 +211,10 @@ object ChiselStage {
     *
     * @param gen a call-by-name Chisel module
     */
+  @deprecated(
+    "this exposes the internal Chisel circuit which was not supposed to be public---use either ChiselStage.convert or ChiselStage.emitCHIRRTL instead",
+    "Chisel 5.0"
+  )
   def elaborate(
     gen:  => RawModule,
     args: Array[String] = Array.empty
--- a/src/test/scala/chisel3/internal/NameCollisionSpec.scala
+++ b/src/test/scala/chisel3/internal/NameCollisionSpec.scala
@@ -10,7 +10,7 @@ class NameCollisionSpec extends ChiselFlatSpec with Utils {
 
   it should "error on duplicated names with a correct message" in {
     (the[ChiselException] thrownBy extractCause[ChiselException] {
-      ChiselStage.elaborate(
+      ChiselStage.emitCHIRRTL(
         new Module {
           val foo, bar = IO(Input(UInt(8.W)))
           val out = IO(Output(UInt(8.W)))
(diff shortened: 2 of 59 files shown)


jackkoenig (approved):

src/main/scala/circt/stage/ChiselStage.scala
```diff
@@ -211,6 +211,10 @@ object ChiselStage {
     *
     * @param gen a call-by-name Chisel module
     */
+  @deprecated(
+    "this exposes the internal Chisel circuit which was not supposed to be public---use either ChiselStage.convert or ChiselStage.emitCHIRRTL instead ",
+    "Chisel 5.0"
+  )
```
jackkoenig, line 217:
```suggestion
  @deprecated(
    "this exposes the internal Chisel circuit which was not supposed to be public--use either ChiselStage.convert or ChiselStage.emitCHIRRTL instead",
    "Chisel 5.0"
  )
```
Minor nit on the message, feel free to ignore

src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -10,7 +10,7 @@ class NameCollisionSpec extends ChiselFlatSpec with Utils {
 
   it should "error on duplicated names with a correct message" in {
     (the[ChiselException] thrownBy extractCause[ChiselException] {
-      ChiselStage.elaborate(
+      ChiselStage.emitCHIRRTL(
```
mwachs5, line 13: wouldn't you want to keep the elaborate but just kill the return value? We don't actually care about emitChirrtl?

mwachs5:
seems like we have lots of tests where we don't actually care about the emission, and just want to see if we can successfully elaborate. Could we make something like `testElaborate` just return Unit? Does `emitChirrtl` always emit a file, which could be annoying...

seldridge:
`ChiselStage$.emitCHIRRTL` or `ChiselStage$.convert` do not write any files. I considered adding a test-only `elaborate` method to do this, but it seemed unnecessary given that there was an existing method which we could re-use here.

src/test/scala/chisel3/internal/NameCollisionSpec.scala
seldridge: I mean, we could. However, it didn't seem useful to have a public `ChiselStage$.elaborate` method which would return nothing. This could be a test-only method. However, it didn't seem worthwhile given that we could re-use `emitCHIRRTL` here.

Use of `emitCHIRRTL` does do slightly more work for situations which are not catching exceptions. Now, FIRRTL text is generated which would not be (though immediately discarded). For tests that are catching exceptions, the runtime should be the same.

src/main/scala/circt/stage/ChiselStage.scala
seldridge: The use of a fixed-width em-dash was intentional as opposed to a fixed-width en-dash which would not be appropriate. I could use alternative punctuation, though. The space was unintentional. I'll remove that.

src/test/scala/chisel3/internal/NameCollisionSpec.scala
mwachs5: oh yeah good point that most of these tests are expecting an exception anyway

PR #3163, opened 2023-04-06
Fix naming for RHS of named unapply expressions

jackkoenig:
Previously, the naming plugin would not recurse for the RHS of an unapply, meaning that vals declared in blocks that ultimately return some result that is unapplied into some named Data would not be named.

The actual fix here is like 10 characters, but also see the tests and improvements to docs--the examples I reworked were because either they give bad results with MFC or are optimized in such a way that they don't show the purpose of the example (registers are useful for this)

Also, audit and improve the naming mdocs.

I've marked this for 3.6.x. I think we should get this into 3.6.0, the question is if we should do a 3.6.0-RC4 or if this is fine to sneak in to 3.6.0.

### Contributor Checklist

#### Type of Improvement

- bug fix
- documentation

#### API Impact

No impact

Fixes a bug in the intended behavior of the compiler plugin. I am evaluating the impact on SiFive Verilog. It might be substantial.

#### Desired Merge Strategy

- Squash

#### Release Notes

This results in previously unnamed signals receiving names from the compiler plugin.

- [ ] Did you review?


jackkoenig:
I've reviewed the impact on Verilog in some internal code and from my random sampling, lots of signals are now getting named and overall, the churn isn't too bad. There's a bit of churn because unnamed signals now have lower indices, eg. `_T_657` became `_T_124`. This obviously a good thing because more things are being named which improves overall stability.

azidar (approved):

PR #3167, opened 2023-04-07
BoringUtils in Chisel

seldridge:
Rework Chisel internals to delay generation of Modules until all modules
are known.  This enables modules to make modifications to each other
because the modules have not closed yet.

Change instantiation to not populate a DefInstance with ports as the
ports are not known yet.

Add a new BoringUtils method, also called "bore", that will go grab a
Data in a child and bring it up to the current module drilling any
necessary ports.

---

This is an experiment to see how far I could get with this. Due to certain APIs wanting to eagerly inspect the ports of modules, I opted for this lazy approach. The module is closed if the user tries to look at it. I don't particularly like this, but it gets most tests to pass.

Currently the following three tests fail:

1. chiselTests.experimental.ProgrammaticPortsSpec due to ports being named after components.
2. chiselTests.experimental.hierarchy.InstanceSpec due to unknown reasons
3. chiselTests.experimental.hierarchy.DefinitionSpec due to unknown reasons

### DIFF
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -193,4 +193,39 @@ object BoringUtils {
     sinks.foreach(addSink(_, genName, true, true))
     genName
   }
+
+  /** Access a source [[Data]] that may or may not be in the current module.  If
+    * this is in a child module, then create ports to allow access the
+    * requrested source.
+    */
+  def bore(source: Data): Data = {
+    val context = Builder.captureContext()
+    val thisModule = context.currentModule.get
+
+    var rhs = source
+    var module = source.topBinding.location.get
+
+    while (module != thisModule) {
+
+      /** Change the context to the current module. */
+      context.currentModule = Some(module)
+      Builder.restoreContext(context)
+
+      val bore = source.cloneTypeFull
+
+      /** Create a port. */
+      module.bindIoInPlace(bore)
+
+      bore :<>= rhs
+
+      rhs = bore
+      module = module._parent.get
+    }
+
+    /** Reset the context to the current module. */
+    context.currentModule = Some(thisModule)
+    Builder.restoreContext(context)
+
+    rhs
+  }
 }
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -67,19 +67,14 @@ object Module extends SourceInfoDoc {
     Builder.currentClock = saveClock // Back to clock and reset scope
     Builder.currentReset = saveReset
 
-    // Only add the component if the module generates one
-    val componentOpt = module.generateComponent()
-    for (component <- componentOpt) {
-      Builder.components += component
-    }
+    Builder.componentsToGenerate += module
 
     Builder.setPrefix(savePrefix)
 
     // Handle connections at enclosing scope
     // We use _component because Modules that don't generate them may still have one
-    if (Builder.currentModule.isDefined && module._component.isDefined) {
-      val component = module._component.get
-      pushCommand(DefInstance(sourceInfo, module, component.ports))
+    if (Builder.currentModule.isDefined) {
+      pushCommand(DefInstance(sourceInfo, module, Seq.empty))
       module.initializeInParent()
     }
     module
@@ -212,12 +207,11 @@ package internal {
     )(
       implicit sourceInfo: SourceInfo
     ): ClonePorts = {
-      require(proto.isClosed, "Can't clone a module before module close")
+      proto.tryClose()
       // Fake Module to serve as the _parent of the cloned ports
       // We make this before clonePorts because we want it to come up first in naming in
       // currentModule
       val cloneParent = Module(new ModuleClone(proto))
-      require(proto.isClosed, "Can't clone a module before module close")
       require(cloneParent.getOptionRef.isEmpty, "Can't have ref set already!")
       // Fake Module to serve as the _parent of the cloned ports
       // We don't create this inside the ModuleClone because we need the ref to be set by the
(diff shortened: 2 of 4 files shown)


azidar:
I'd recommend a different approach. Instead, still close the module, but add a chisel3 private function that adds a "hidden port" IR thing (i'd even make it a different IR command), which is just added to the module's component's ArrayBuffer of chisel IR commands (despite the module being closed).

mwachs5:
> function that adds a "hidden port" IR thing (i'd even make it a different IR command), which is just added to the module's component's ArrayBuffer of chisel IR commands (despite the module being closed).

I think this also addresses my concern in formalizing what these reflective APIs get to do and see -- it says that these APIs can *never* see these ports, and thus users should never expect to be able to do anything with them.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +193,39 @@ object BoringUtils {
...
     genName
   }
+
+  /** Access a source [[Data]] that may or may not be in the current module.  If
+    * this is in a child module, then create ports to allow access the
+    * requrested source.
+    */
+  def bore(source: Data): Data = {
+    val context = Builder.captureContext()
+    val thisModule = context.currentModule.get
```
mwachs5, line 203: Conceptually, do you intend this to work on an instance or a definition? if I call val copy = Instance(thisModule.toDefinition) after this boring should i expect copy to have bored ports or not?

seldridge: Good question. I put a longer response to this in the main comment area as that should make discussion easier.

seldridge:
I think this is hitting at the same problem as if you have a multiply-instantiated module and you bore ports on an instance. The `toDefinition` API brings up some specific interesting angles, but I don't think it's fundamentally different. I'll come back to that.

If I have a module that is multiply-instantiated and I try to add a port to an instance, one of two things can happen:

1. Only that instance gets ports. This causes that instance to be duplicated.
2. All instances get the port.
3. This is disallowed.

(1) is something that may not fundamentally compose with any solution for multiple instantiation. If the user says both this module must be multiply instantiated and this instance has different ports, then these are fundamentally incompatible. _It can be automatically reconciled by duplicating_, however, it's not clear that this is better behavior than erroring.

I would note that this is _the exact same problem_ that a FIRRTL compiler has to deal with today and will error on. If you try to run `BoringUtils` on a multiply-instantiated module, then MFC will immediately error and SFC will do a bit more work to try to figure out what you mean, but usually error.

(2) is problematic for anything which is adding inputs to instances. For outputs you could argue that these are just left unconnected. This doesn't seem great, but might be okay. Unconnected inputs violate FIRRTL initialization checking.

(3) seems like a totally reasonable place to start as this is almost the same situation today as where we are with FIRRTL compilers. If we can make it obvious to a user that they are doing something illegal earlier, we could potentially avoid a whole class of user errors that we are patching over with `MustDeduplicateAnnotation`. Put differently, I think we should never be deduplicating or duplicating once a Chisel generator has produced FIRRTL.

The `toDefinition` API is interesting because it brings up the situation of deriving a `Definition` from the output of a generator. (The generator has to `@public` things that the later created `Definition` can use, but the underlying FIRRTL type can be entirely arbitrary unless we move to restrict calls to `IO`.) What happens if I add ports to this module before creating the `Definition` via `toDefinition`? One of several things along two axes can happen:

4. The port is added to the new `Definition`.
  a. The port is accessible via D/I APIs.
  b. The port is only accessible through reflective, `DataMirror` APIs 
  c. The port is not accessible except in FIRRTL.
5. The port is not added to the new `Definition`.
6. You get an error.

(4) seems pretty reasonable here. However, there are difficulties with any of the sub-bullets. The user clearly cause an `IO` to be added. The current proposal that Adam brought up is (4.c). This is problematic for reasons outlined in (3). This is likely going to be an initialization error in a FIRRTL compiler. I don't know how (4.a) is realizable given with how `D/I` works. You fundamentally cannot pre-declare an `@public` `IO` that you don't know is going to be added. Granted, I think this is reasonable to error on. Only allowing access through `DataMirror`-style APIs seems reasonable, however, this may not be the best user-experience. It does provide a way for a user to avoid FIRRTL compiler errors.

(5) would mean that the `Definition` is effectively _not_ a true clone of the original module. This is essentially the original case of (1). When creating the `Definition` the Chisel runtime knows that this module has been "tainted" and it needs to create a copy of the original module without the extra ports. This composes with Adam's suggestion that the late-added ports are treated specially. However, this does mean that we need to keep track of every time that ports are added between new calls to `toDefinition`. This works, but may have surprising duplication behavior which should probably be avoided.

(6) is always reasonable for similar reasons as stated earlier. (4.b) is likely going to be a FIRRTL compiler error and (5) is likely going to violate something related to `MustDeduplicateAnnotation` restrictions.

There are then questions of what happens if a user adds `IO` _after_ a user calls `toDefinition`. It would be safest to error here. Duplication or adding the ports everywhere is possible, but likely extremely confusing.

mwachs5:
> I would note that this is the exact same problem that a FIRRTL compiler has to deal with today and will error on. If you try to run BoringUtils on a multiply-instantiated module, then MFC will immediately error and SFC will do a bit more work to try to figure out what you mean, but usually error.

Right. I agree this is a problem. Which is why I want to work it out before just making it Chisel's problem too :)

azidar:
Here's another way to think about it.

Punching ports downwards is only legal across a `Module(new MyModule)` boundary. For a D/I usecase, we can make `@public val myWire = Wire(..)` auto-punch the `myWire` probe upwards.

This avoids the top-down-punching on multi-instantiated modules problem, while giving equivalent capability for multi-instantiated modules.

mwachs5:
> For a D/I usecase, we can make @public val myWire = Wire(..) auto-punch the myWire probe upwards. 

I find this very appealing. Similar @public val myChild = Module(new SomeChild) or @public val myChildren: Seq[Instance[SomeChildPublicIfc]] = Seq.fill(3)(Instantiate(new SomeChild)) would forward the IOs, but they would be punched unconditionally upwards regardless if someone wanted to tap them.

But the Probe APIs are still in flux anyway and this could be built on top.

seldridge:
Closing in favor of: https://github.com/chipsalliance/chisel/pull/3187

PR #3171, opened 2023-04-10
Patch VecInit.fill(0) invocation to successfully compile and yield a zero-width Vec

jared-barocsi:
Fixes common workarounds for instantiating a 0-width Vec, like:
```
if(n < 1) Wire(Vec(0, Bool())) else VecInit.fill(n)(false.B)
```
by making a direct `VecInit.fill(0)` call possible.

Some connection semantics issues were discovered however which allow users to connect zero `Vecs` of one type to another zero `Vec` of a different type, because the existing logic only operates on a per-child-element basis and thus doesn't execute at all for `Aggregates` containing no child elements. A temporary fix is applied for this with the additional note that the connection logic needs to be looked at for future unification.
### Contributor Checklist

#### Type of Improvement

- feature
- bug fix

#### API Impact

Modifies `VecInit.fill`

No impact

#### Desired Merge Strategy

Squash and merge

#### Release Notes

Fix `VecInit.fill(0)` calls so that they compile and yield 0-width `Vecs`

- [ ] Did you review?


mwachs5:
@jackkoenig i think we'd apply the `feature` label here if we can pick only one? The "it's now a bug to do something you should not have been doing" seems OK right?

jackkoenig:
@mwachs5 I agree with `Feature`, `BugFix` is also reasonable but I think `Feature` makes the most sense.

src/test/scala/chiselTests/Vec.scala
```diff
@@ -276,6 +276,56 @@ class VecSpec extends ChiselPropSpec with Utils {
...
+      val chirrtl = emitCHIRRTL(new Module {
+        val io = IO(new Bundle {
+          val out = Vec(0, UInt(8.W))
+        })
+
+        val zeroWidthVec = VecInit.fill(0)(gen)
+
+        connect(io.out, zeroWidthVec)
+      })
+      println(chirrtl)
```
jackkoenig, line 306:
```suggestion
```

jackkoenig, line 306: Can these tests be combined, basically the bottom seems to be testing the same thing and you could probably do the same chirrtl check on every iteration of the bottom test?

jackkoenig (approved):

PR #3173, opened 2023-04-11
Auto probe derived from @public

azidar:
Early work demonstrating how this could work. This falls into the 'bottom-up' declare your interface early camp.

### Contributor Checklist

- [ ] Did you review?


src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -413,6 +413,33 @@ class InstanceSpec extends ChiselFunSpec with Utils {
...
     }
+    it("(3.r): @public should create probes") {
+      @instantiable
+      class Leaf extends Module {
+        @public val w = WireInit(true.B)
+      }
+      class Top() extends Module {
+        val a = IO(Output(Probe(Bool())))
+        val i = Instance(Definition(new Leaf))
+        a := i.w
```
mwachs5, line 424: isn't this supposed to have to use `Probe.define`?

mwachs5:
What should I look at in this PR to see the ideas that are independent from @debs-sifive base PR?

src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -413,6 +413,33 @@ class InstanceSpec extends ChiselFunSpec with Utils {
       val (chirrtl, _) = getFirrtlAndAnnos(new AddTwoNestedInstantiableDataWrapper(4))
       exactly(3, chirrtl.serialize.split('\n')) should include("i1.in <= i0.out")
     }
+    it("(3.r): @probe should create probes") {
```
mwachs5, line 416: probe IOs right? so the idea is the `@probe` macro is equivalent to 

- create the probe IO
- create the probed-value
- define the probe IO to the probed-value?

is any @probe automatically @public, which lookup of returns the handle to the probe IO?

src/test/scala/chiselTests/experimental/hierarchy/InstanceSpec.scala
```diff
@@ -413,6 +413,33 @@ class InstanceSpec extends ChiselFunSpec with Utils {
...
+        @probe final val w = WireInit(true.B)
+      }
+      @instantiable
+      class Top() extends Module {
+        val i = Instance(Definition(new Leaf))
+        @probe final val x = i.w
+      }
+      val (chirrtl, _) = getFirrtlAndAnnos(new Top())
+      println(chirrtl.serialize) // TODO: _ref_i_ref_w should be named as 'x' or '_ref_x'?
+      exactly(1, chirrtl.serialize.split('\n')) should include("define _ref_i_ref_w = i._ref_w")
```
mwachs5, line 441: shouldn't this also check the creation of a probed-IO and the `ProbeValue`? this seems to just check for the connection?

Can you show the behavior if @probe is also @public, what happens if I try to access a probed variable in an Instance? Must I use probe.define/read and not `:=`?

mwachs5: i guess this test *is* showing what I am asking for, in a way.

PR #3174, opened 2023-04-12
Report firtool version when firtool invocation errors

jackkoenig:
* the firtool version is now included in the BuildInfo
* firtool now must be on the path when publishing, but is not required for building or testing
* FirtoolNonZeroExitCode now no longer includes a stack trace

Note that this means any firtool error when using `circt.stage.ChiselStage` will emit this error. This means that even if the user has the same version of firtool installed, or a newer version, and they get some other error (eg. initialization error), they the message will tell them which version of firtool Chisel was built against. One the one hand, this could be a good thing because some unpredictable errors may come up from using too new of versions of firtool. On the other hand, it could nudge users toward using older versions of firtool even when a newer one will do, and it could be confusing if they do already have the exact version installed.

I'm not sure how to thread this needle so would appreciate ideas from others. We can also accept this imperfect solution in the meantime (for releases of Chisel 3.6.0 and v5.0.0) and come up with a better solution in the future.

### Contributor Checklist

#### Type of Improvement

- Bugfix

(I'm calling this a bugfix because it gives a better error message, maybe not the best category but not sure)

#### Desired Merge Strategy

- Squash

- [ ] Did you review?


seldridge (approved):
LGTM

Is a test possible?

PR #3180, opened 2023-04-13
Emit annotations in the .fir file

jackkoenig:
This commit reimplements the previous work adding emission of annotations via the FIRRTL specification v2.0.0 %[[...]] style. Now all annotations that were previously emitted to the .anno.json file are emitted in the .fir file. This uses a slightly hacky new internal mechanism in firrtl that is already marked deprecated to discourage users from touching.

Revamping of https://github.com/chipsalliance/chisel/pull/3141

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

#### Release Notes

Annotations are now emitted in the `.fir` file instead of in an auxiliary `.anno.json` file

- [ ] Did you review?


firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -543,7 +545,6 @@ case class IntModule(
     extends DefModule
     with UseSerializer
 
-case class Circuit(info: Info, modules: Seq[DefModule], main: String, annotations: AnnotationSeq)
-    extends FirrtlNode
-    with HasInfo
-    with UseSerializer
+case class Circuit(info: Info, modules: Seq[DefModule], main: String) extends FirrtlNode with HasInfo with UseSerializer
+
+case class CircuitWithAnnos(circuit: Circuit, annotations: Seq[Annotation]) extends FirrtlNode with UseSerializer
```
seldridge, line 550: Any reason to break this up as opposed to keeping it unified?

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -192,12 +195,39 @@ case class CircuitSerializationAnnotation(circuit: Circuit, filename: String, fo
...
   protected def suffix: Option[String] = Some(format.extension)
 
-  override def getBytesBuffered: Iterable[Array[Byte]] = format match {
-    case FirrtlFileFormat =>
-      OldEmitter
-        .emitLazily(circuit)
-        .map(_.getBytes)
+  /** Write the circuit and annotations to the .fir file
+    */
+  protected def writeToFileImpl(file: File, annos: Seq[Annotation]): Unit = {
```
seldridge, line 200: Nit: 
```suggestion
  override protected def writeToFileImpl(file: File, annos: Seq[Annotation]): Unit = {
```

firrtl/src/main/scala/firrtl/options/internal/WritableCircuitAnnotation.scala
```diff
@@ -0,0 +1,17 @@
...
+package firrtl.options.internal
+
+import firrtl.options.CustomFileEmission
+import firrtl.annotations.Annotation
+
+import java.io.File
+
+// Hack for enabling special emission of the ChiselCircuitAnnotation in WriteOutputAnnotations
+@deprecated("This trait is for internal use only. Do not use it.", "Chisel 5.0")
+trait WriteableCircuitAnnotation extends CustomFileEmission { self: Annotation =>
```
seldridge, line 12: We should soon move the `firrtl` package under `chisel3` to enable making this private. 🥲

src/main/scala/circt/stage/ChiselStage.scala
```diff
@@ -96,14 +106,18 @@ object ChiselStage {
...
-      .transform(annos)
-      .collectFirst {
-        case a: ChiselCircuitAnnotation => CircuitSerializationAnnotation(a.circuit, "", FirrtlFileFormat).getBytes
-      }
-      .get
-      .map(_.toChar)
-      .mkString
+    val resultAnnos = phase.transform(annos)
+
+    var circuitAnno: Option[CircuitSerializationAnnotation] = None
```
seldridge, line 111: For your consideration: This is the type of thing that the [package view/options](https://github.com/chipsalliance/chisel/blob/main/src/main/scala/circt/stage/package.scala) is supposed to handle for you. It could be used here or not. I want to gut that and the rest of stage/phase so I can see not wanting to use it here.

seldridge (approved):
This is pretty janky, but I'm okay with it.

Is the output annotation file _not_ printed now if there are no annotations? Can the `OutputAnnotationFileAnnotation` be fully removed?

jackkoenig:
> Is the output annotation file not printed now if there are no annotations? Can the OutputAnnotationFileAnnotation be fully removed?

Possibly, I'll do an audit and at least deprecate it. There may be some remaining uses of Stage/Phase that don't result in a `.fir` file or want to emit to `.anno.json`. I'll leave actual removal for a Chisel 6 change (which can happen very soon once I create the `5.x` release branch).

firrtl/src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
```diff
@@ -41,52 +42,79 @@ class WriteOutputAnnotations extends Phase {
...
+          filesToWrite(canonical) = a
+        case _ =>
+      }
+      // Write files with CustomFileEmission and filter out Unserializable ones
+      anno match {
+        case _:   Unserializable => None
+        case wca: WriteableCircuitAnnotation =>
+          val filename = wca.filename(annotations)
+          circuitAnnoOpt = Some(filename -> wca)
+          None
```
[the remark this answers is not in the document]
jackkoenig: Possibly, I didn't mess with it because Chisel or Stage/Phase are probably adding the "default one". I'll play around with it a little and see.

src/main/scala/circt/stage/ChiselStage.scala
jackkoenig: I could move it in there if you think it's better. I don't have a strong opinion as this logic is somewhat duplicative of what's in WriteOutputAnnotations but it would be very difficult to share the code between the two. I figure neither this code nor WriteOutputAnnotations are long for this world so maybe it's not worth worrying about since we'll just rewrite it soon 🤷‍♀️

firrtl/src/main/scala/firrtl/options/internal/WritableCircuitAnnotation.scala
jackkoenig: Agreed, good Chisel 6 change

firrtl/src/main/scala/firrtl/ir/IR.scala
jackkoenig: Yeah because the Converter creates a firrtl Circuit but returns Seq[Annotation]. I'm totally down with refactoring/replacing/nuking the Stage/Phase approach but this is a bit more of a min-cut approach that keeps all Stages and Phases working as they have in the past (including custom user code filtering out annotations after running the Converter which we're doing at SiFive) while only imposing the new "CircuitWithAnnos" upon emission at the end of the Stage. In any case, let's rewrite it all but I think for 5.0 this is an appropriately modest way to do it.

firrtl/src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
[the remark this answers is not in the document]
jackkoenig: Oh sorry I misread the suggestion, yes we should do that assertion, I thought this was referring to something later.

PR #3181, opened 2023-04-13
Allow DataView of Reset <=> [UInt<1>, AsyncReset]

seldridge:
Change what DataView allows in its type checking to enable Reset to be viewed as either a UInt<1> or an AsyncReset.  This enables more flexible DataViews which match already allowable Chisel and FIRRTL connection semantics.

#### Release Notes

Change what DataView allows in its type checking to enable Reset to be viewed as either a UInt<1> or an AsyncReset.  This enables more flexible DataViews which match already allowable Chisel and FIRRTL connection semantics.

### DIFF
--- a/core/src/main/scala/chisel3/experimental/dataview/package.scala
+++ b/core/src/main/scala/chisel3/experimental/dataview/package.scala
@@ -103,10 +103,21 @@ package object dataview {
       val tex = unfoldView(te).find(targetContains).getOrElse(err("Target", te))
       val vex = unfoldView(ve).find(viewFieldLookup.contains).getOrElse(err("View", ve))
 
-      if (tex.getClass != vex.getClass) {
-        val fieldName = viewFieldName(vex)
-        throw InvalidViewException(s"Field $fieldName specified as view of non-type-equivalent value $tex")
+      (tex, vex) match {
+        /* Allow views where the types are equal. */
+        case (a, b) if a.getClass == b.getClass =>
+        /* Allow UInt <=> Reset views. */
+        case (a: UInt, _: Reset) if a.isWidthKnown && a.getWidth == 1 => true
+        case (_: Reset, a: UInt) if a.isWidthKnown && a.getWidth == 1 => true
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/experimental/DataView.scala
```diff
@@ -546,6 +546,40 @@ class DataViewSpec extends ChiselFlatSpec {
     chirrtl should include("dataOut <= vec[addrReg]")
   }
 
+  it should "allow views between reset types" in {
+    class A extends Bundle {
+      val uint = AsyncReset()
```
jackkoenig, line 551:
```suggestion
      val uint = Bool()
```
I think you meant to make this a Bool

core/src/main/scala/chisel3/experimental/dataview/package.scala
```diff
@@ -103,10 +103,21 @@ package object dataview {
...
+        /* Allow views where the types are equal. */
+        case (a, b) if a.getClass == b.getClass =>
+        /* Allow UInt <=> Reset views. */
+        case (a: UInt, _: Reset) if a.isWidthKnown && a.getWidth == 1 => true
+        case (_: Reset, a: UInt) if a.isWidthKnown && a.getWidth == 1 => true
+        /* Allow AsyncReset <=> Reset views. */
+        case (a: AsyncReset, _: Reset) => true
+        case (_: Reset, a: AsyncReset) => true
+        /* All other views produce a runtime error. */
+        case _ =>
```
jackkoenig, line 116: The return values are unnecessary, also should this be UInts with checked widths or just Bool? Generally we say that synchronous reset is Bool, not 1-bit UInt.

seldridge: I didn't realize that it was restricted to `Bool` only as opposed to `UInt(1.W)`. Good catch. I'll change this.

seldridge:
Review bump @jackkoenig. 😉

jackkoenig (approved):

PR #3187, opened 2023-04-14
Handle BoringUtils in Chisel, rather than in FIRRTL compiler

azidar:
This is a more complete implementation of #3167 , but rather than delaying closing of a module, adds a special area where ports and connections can be added to a module, post-closing. This second phase is closed when calling `.toDefinition` or reflecting on the ports (e.g. with `DataMirror`). 

A longer term plan is to fix some internals (e.g. when closing happens) so that we can unify these phases more gracefully.

For handling the D/I case, we can consider https://github.com/chipsalliance/chisel/pull/3173 as one approach.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference.)

#### Release Notes

Added a new BoringUtils.bore that is performed during Chisel elaboration, not via Annotations+CIRCT. Punched ports are accessible to the user via DataMirror. However, using these reflection APIs or calling .toDefinition will fully close a module, to ensure that subsequent boring fails (and thus getting all ports is never stale).

Added a `skipPrefix` to enable ignoring the last prefix value in the prefix name stack.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -321,7 +321,7 @@ package experimental {
     private[chisel3] var modulePortsAskedFor: Option[SourceInfo] = None
 
     /** Where a Module becomes fully closed (no secret ports drilled afterwards) */
-    private[chisel3] def fullyClosedLocation = toDefinitionCalled.orElse(modulePortsAskedFor)
+    private[chisel3] def isFullyClosed = fullyClosedErrorMessages.nonEmpty
     private[chisel3] def fullyClosedErrorMessages: Iterable[(SourceInfo, String)] = {
       toDefinitionCalled.map(si =>
         (si, s"Calling .toDefinition fully closes ${name}, but it is later bored through!")
@@ -561,17 +561,14 @@ package experimental {
     // data must be a fresh Chisel type
     private[chisel3] def createSecretIO(data: => Data)(implicit sourceInfo: SourceInfo): Data = {
       val iodef = data
-      // forcename
       internal.requireIsChiselType(iodef, "io type")
-
-      if (toDefinitionCalled.nonEmpty)
-        Builder.error("can only create secret ports when a module has not been used as a Definition")
-      if (modulePortsAskedFor.nonEmpty)
-        Builder.error("can only create secret ports when a module has not been asked for all its IO")
+      require(!isFullyClosed, "Cannot create secret ports if module is fully closed")
 
       iodef.bind(internal.SecretPortBinding(this))
       iodef
     }
+
+    private[chisel3] val secretPorts: ArrayBuffer[Port] = ArrayBuffer.empty
 
     // Must have separate createSecretIO from addSecretIO to get plugin to name it
     private[chisel3] def addSecretIO(iodef: Data)(implicit sourceInfo: SourceInfo): Data = {
@@ -583,8 +580,6 @@ package experimental {
       } else secretPorts += newPort
       iodef
     }
-
-    private[chisel3] val secretPorts: ArrayBuffer[Port] = ArrayBuffer.empty
 
     /**
       * This must wrap the datatype used to set the io field of any Module.
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -247,18 +247,23 @@ package internal {
 package experimental {
 
   import chisel3.experimental.hierarchy.core.{IsInstantiable, Proto}
+  import scala.annotation.nowarn
 
   object BaseModule {
-    implicit class BaseModuleExtensions[T <: BaseModule](b: T) {
+    implicit class BaseModuleExtensions[T <: BaseModule](b: T)(implicit si: SourceInfo) {
       import chisel3.experimental.hierarchy.core.{Definition, Instance}
-      def toInstance:   Instance[T] = new Instance(Proto(b))
-      def toDefinition: Definition[T] = new Definition(Proto(b))
+      def toInstance: Instance[T] = new Instance(Proto(b))
+      def toDefinition: Definition[T] = {
+        b.toDefinitionCalled = Some(si)
+        new Definition(Proto(b))
+      }
     }
   }
 
   /** Abstract base class for Modules, an instantiable organizational unit for RTL.
     */
   // TODO: seal this?
+  @nowarn("msg=class Port") // delete when Port becomes private
   abstract class BaseModule extends HasId with IsInstantiable {
     _parent.foreach(_.addId(this))
 
@@ -309,8 +314,22 @@ package experimental {
     //
     protected var _closed = false
 
-    /** Internal check if a Module is closed */
+    /** Internal check if a Module's constructor has finished executing */
     private[chisel3] def isClosed = _closed
+
+    private[chisel3] var toDefinitionCalled:  Option[SourceInfo] = None
+    private[chisel3] var modulePortsAskedFor: Option[SourceInfo] = None
+
+    /** Where a Module becomes fully closed (no secret ports drilled afterwards) */
+    private[chisel3] def fullyClosedLocation = toDefinitionCalled.orElse(modulePortsAskedFor)
+    private[chisel3] def fullyClosedErrorMessages: Iterable[(SourceInfo, String)] = {
+      toDefinitionCalled.map(si =>
+        (si, s"Calling .toDefinition fully closes ${name}, but it is later bored through!")
+      ) ++
+        modulePortsAskedFor.map(si =>
+          (si, s"Reflecting on all io's fully closes ${name}, but it is later bored through!")
+        )
+    }
 
     // Fresh Namespace because in Firrtl, Modules namespaces are disjoint with the global namespace
     private[chisel3] val _namespace = Namespace.empty
@@ -511,9 +530,10 @@ package experimental {
       *
       * TODO: Use SeqMap/VectorMap when those data structures become available.
       */
-    private[chisel3] def getChiselPorts: Seq[(String, Data)] = {
+    private[chisel3] def getChiselPorts(implicit si: SourceInfo): Seq[(String, Data)] = {
       require(_closed, "Can't get ports before module close")
-      _component.get.ports.map { port =>
+      modulePortsAskedFor = Some(si) // super-lock down the module
+      (_component.get.ports ++ _component.get.secretPorts).map { port =>
         (port.id.getRef.asInstanceOf[ModuleIO].name, port.id)
       }
     }
@@ -536,6 +556,35 @@ package experimental {
     )(
       implicit sourceInfo: SourceInfo
     ): Unit = _bindIoInPlace(iodef)
+
+    // Must have separate createSecretIO from addSecretIO to get plugin to name it
+    // data must be a fresh Chisel type
+    private[chisel3] def createSecretIO(data: => Data)(implicit sourceInfo: SourceInfo): Data = {
+      val iodef = data
+      // forcename
+      internal.requireIsChiselType(iodef, "io type")
+
+      if (toDefinitionCalled.nonEmpty)
+        Builder.error("can only create secret ports when a module has not been used as a Definition")
+      if (modulePortsAskedFor.nonEmpty)
+        Builder.error("can only create secret ports when a module has not been asked for all its IO")
+
+      iodef.bind(internal.SecretPortBinding(this))
+      iodef
+    }
+
+    // Must have separate createSecretIO from addSecretIO to get plugin to name it
+    private[chisel3] def addSecretIO(iodef: Data)(implicit sourceInfo: SourceInfo): Data = {
+      val name = iodef._computeName(None).getOrElse("secret")
+      iodef.setRef(ModuleIO(this, _namespace.name(name)))
+      val newPort = new Port(iodef, iodef.specifiedDirection, sourceInfo)
+      if (_closed) {
+        _component.get.secretPorts += newPort
+      } else secretPorts += newPort
+      iodef
+    }
+
+    private[chisel3] val secretPorts: ArrayBuffer[Port] = ArrayBuffer.empty
 
     /**
       * This must wrap the datatype used to set the io field of any Module.
--- a/core/src/main/scala/chisel3/internal/Binding.scala
+++ b/core/src/main/scala/chisel3/internal/Binding.scala
@@ -38,7 +38,7 @@ private[chisel3] object BindingDirection {
     */
   def from(binding: TopBinding, direction: ActualDirection): BindingDirection = {
     binding match {
-      case PortBinding(_) =>
+      case _: PortBinding | _: SecretPortBinding =>
         direction match {
           case ActualDirection.Output => Output
           case ActualDirection.Input  => Input
--- a/core/src/main/scala/chisel3/internal/Binding.scala
+++ b/core/src/main/scala/chisel3/internal/Binding.scala
@@ -89,6 +89,9 @@ sealed trait ConditionalDeclarable extends TopBinding {
 @deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 case class PortBinding(enclosure: BaseModule) extends ConstrainedBinding
 
+// Added to handle BoringUtils in Chisel
+private[chisel3] case class SecretPortBinding(enclosure: BaseModule) extends ConstrainedBinding
+
 @deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 case class OpBinding(enclosure: RawModule, visibility: Option[WhenContext])
     extends ConstrainedBinding
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -5,7 +5,7 @@ package chisel3.reflect
 import chisel3._
 import chisel3.internal._
 import chisel3.internal.firrtl._
-import chisel3.experimental.BaseModule
+import chisel3.experimental.{BaseModule, SourceInfo}
 import scala.reflect.ClassTag
 
 object DataMirror {
@@ -14,6 +14,15 @@ object DataMirror {
   def directionOf(target: Data): ActualDirection = {
     requireIsHardware(target, "node requested directionality on")
     target.direction
+  }
+
+  /** Returns true if target has been `Flipped` or `Input` directly */
+  def hasOuterFlip(target: Data): Boolean = {
+    import chisel3.SpecifiedDirection.{Flip, Input}
+    target.specifiedDirection match {
+      case Flip | Input => true
+      case _            => false
+    }
   }
 
   private def hasBinding[B <: ConstrainedBinding: ClassTag](target: Data) = {
@@ -114,7 +123,7 @@ object DataMirror {
     * // )
     * }}}
     */
-  def modulePorts(target: BaseModule): Seq[(String, Data)] = target.getChiselPorts
+  def modulePorts(target: BaseModule)(implicit si: SourceInfo): Seq[(String, Data)] = target.getChiselPorts
 
   /** Returns a recursive representation of a module's ports with underscore-qualified names
     * {{{
@@ -145,7 +154,7 @@ object DataMirror {
     *       of its children.
     * @see [[DataMirror.modulePorts]] for a non-recursive representation of the ports.
     */
-  def fullModulePorts(target: BaseModule): Seq[(String, Data)] = {
+  def fullModulePorts(target: BaseModule)(implicit si: SourceInfo): Seq[(String, Data)] = {
     def getPortNames(name: String, data: Data): Seq[(String, Data)] = Seq(name -> data) ++ (data match {
       case _: Element => Seq()
       case r: Record =>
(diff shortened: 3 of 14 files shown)


core/src/main/scala/chisel3/IO.scala
```diff
@@ -44,3 +44,42 @@ object IO {
     iodefClone
   }
 }
+
+private[chisel3] object SecretIO {
+
+  /** Constructs a secret port for the current Module
```
mwachs5, line 50: is a SecretIO visible in DataMirror getPorts and friends? if not why not?

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -89,6 +89,10 @@ sealed trait ConditionalDeclarable extends TopBinding {
 @deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 case class PortBinding(enclosure: BaseModule) extends ConstrainedBinding
 
+// Added to handle BoringUtils in Chisel
+@deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
```
mwachs5, line 93: why not make it private right away?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +125,86 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
     runTester(new InternalBoreTester, annotations = TesterDriver.verilatorOnly) should be(true)
   }
 
+  it should "work using new API" in {
```
mwachs5, line 128: havent read the test in detail, but can you show boring up into a parent and back into a sibling...? or how would that use case be handled?

core/src/main/scala/chisel3/IO.scala
azidar: It isn't, and that's to avoid the issue of mutability of a definition. To expand further, if I have code like:
```
val child = Module(new Child)
...
functionFoo(DataMirror.getPorts(child))
...
val myWire = BoringUtils.bore(child.wire)
...
functionBar(DataMirror.getPorts(child))
```

`functionFoo` and `functionBar` would get different ports, which would be difficult to debug because the order of execution matters. This is in general why Chisel has a module "close", so that once I have a reference to `child`, all its Chisel-y things are known.

Another reason supporting not making these present in DataMirror APIs is that all existing uses of boring utils would never show up in a DataMirror getPorts because they were added by CIRCT, so this is backwards compatible behavior.

src/test/scala/chiselTests/BoringUtilsSpec.scala
azidar: The later test shows this is illegal - for right now I'm only letting a user bore downwards.

azidar: Do I need to support that?

core/src/main/scala/chisel3/IO.scala
```diff
@@ -44,3 +44,42 @@ object IO {
...
+      else
+        try {
+          data.cloneTypeFull
+        } catch {
+          // For now this is going to be just a deprecation so we don't suddenly break everyone's code
+          case e: AutoClonetypeException =>
+            Builder.deprecated(e.getMessage, Some(s"${data.getClass}"))
+            data
+        }
+    module.bindSecretIo(iodefClone)
```
jackkoenig, line 82: I know this is from original IO, but I think this chunk of code is old compatibility mode stuff that is no longer necessary.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -259,6 +260,7 @@ package experimental {
   /** Abstract base class for Modules, an instantiable organizational unit for RTL.
     */
   // TODO: seal this?
+  @nowarn("msg=class Port") // delete when Port becomes private
```
jackkoenig, line 263: We should probably just promote this warning suppression to build.sbt `-Wconf` level, or go ahead and make the class package private and remove the deprecation.

src/test/scala/chiselTests/BoringUtilsSpec.scala
seldridge: It would be good to allow it to enable the existing Grand Central View API.

This has taps that reach into the parent. Anything in the parent's visible `Data` can be captured. An example use is approximately:

```scala
class Bar extends RawModule {
  val a = Wire(Bool())
}

class Foo extends RawModule {
  val bar = Module(new Bar)
  View {
    val b = Tap(bar.a)
  }
}
```

The result of this is something like the following FIRRTL:

```
module Bar:
  output a_probe: Probe<UInt<1>>
  wire a: UInt<1>

  define(a_probe, probe(a))

module View_companion:
  input a: UInt<1>

  node b = a

module Foo:

  inst bar of Bar
  inst view of View_companion

  view.a <= read(bar.a_probe)
```

core/src/main/scala/chisel3/IO.scala
mwachs5: sure, but is this a silent failure of the DataMirror APIs that they would return something different than the user might expect?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,54 @@ object BoringUtils {
...
+            Builder.error(s"Can only bore into modules that have finished building, but $source is in an open module")
+            module = thisModule // exit early
+          } else {
+
+            /** Create a port. */
+            val bore = SecretIO(chiselTypeOf(source), module)
+            module.addSecretIo(bore)
+            module.asInstanceOf[RawModule].secretConnection(bore, rhs)
+
+            rhs = bore
```
mwachs5, line 239: stupid question -- is this actually doing a connection through the bored IOs? where does that happen?

In other words, its not ust a secret IO that we need, its also a secret connection

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,54 @@ object BoringUtils {
...
+
+        if (source.topBinding.location.get == thisModule) {
+
+          /** If no boring actually happened */
+          source
+        } else {
+
+          /** Creating an intermediate wire so secret stuff never escapes */
+          val bore = Wire(chiselTypeOf(source))
+          thisModule.asInstanceOf[RawModule].secretConnection(bore, rhs)
```
mwachs5, line 252: oh i see here is the secret connection

src/test/scala/chiselTests/BoringUtilsSpec.scala
mwachs5: right, we don't need a "secret IO" for that use case since the modules in question (the View and the Parent) are still open, but the Bore API should support at least one level of up-boring (TBD whether the bore API is supposed to work arbitrarily for some sort of 
"i promise this is accessible from some heirarchy if you keep going up both mine and the boring-top-point you'll find something in common")

core/src/main/scala/chisel3/IO.scala
azidar: I can update their documentation to report that they return all non-secret ports?

core/src/main/scala/chisel3/Module.scala
azidar: Do you want me to do it in this PR?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +126,145 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
...
+      "input q_bore : UInt<1>",
+      "q <= q_bore_1", // Do normal connection before secret ones
+      "q_bore_1 <= q_bore",
+      "module Foo :",
+      "input a : UInt<1>",
+      "bar.q_bore <= a"
+    )
+  }
+
+  it should "not affect DataMirror.getPorts, as they are secret (shhh)" in {
```
mwachs5, line 221: this seems a little inconsistent that I am allowed to copy bored ports with toDefinition but I can't reflect on them even before .toDefinitioning? How am I supposed to interact with bored ports on a definition?

core/src/main/scala/chisel3/IO.scala
azidar: I think i'll leave it for a future cleanup PR.

azidar: Nvm, just inlined it.

core/src/main/scala/chisel3/IO.scala
azidar: Another way we could do this is if `DataMirror.getPorts` calls `.toDefinition` to super-close a module prior to returning all the ports? That way you still get all the ports, but you never get a partial list (e.g. if someone adds ports later, it will error)?

core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -346,4 +349,23 @@ object DataMirror {
...
   }
+
+  // Function to find the least common ancestor of two nodes
+  private[chisel3] def leastCommonAncestorModule(left: HasId, right: HasId): Option[BaseModule] = {
+    val leftPathSet = modulePath(left, None).toSet
+    val rightPath = modulePath(right, None)
+    rightPath.collectFirst { case p if leftPathSet.contains(p) => p }
+  }
+  // Function to path upwards, stopping if reaching including
+  private[chisel3] def modulePath(h: HasId, until: Option[BaseModule]): Seq[BaseModule] = {
```
seldridge, line 360: Mega-nit: could you re-order `def modulePath` to occur before it's usage in the LCA method?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+    }
+
+    val lca = DataMirror.leastCommonAncestorModule(source, thisModule)
+    if (lca.isEmpty) {
+      Builder.error(s"Cannot bore from $source to ${thisModule.name}, as they do not share a least common ancestor")
+      return Wire(purePortType) //return empty wire to continue elaboration to collect other errors
+    }
+
+    val upPath = DataMirror.modulePath(source, lca)
+    val downPath = DataMirror.modulePath(thisModule, lca)
```
seldridge, line 244: This organization is a bit odd. The LCA method is used to find the LCA using `modulePath`. Then `modulePath` is re-used to find the path. Would a single method like: `findLCAPaths(foo, bar): (Seq[BaseModule], Seq[BaseModule]) = ???` work better?

core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -346,4 +349,23 @@ object DataMirror {
...
+    val me = h match {
+      case m: BaseModule => Seq(m)
+      case d: Data       => d.topBinding.location.toSeq
+      case m: MemBase[_] => m._parent.toSeq
+    }
+    if (me == until.toSeq) Nil
+    else {
+      me ++ me.flatMap(x => x._parent.toSeq.flatMap(p => modulePath(p, until)))
+    }
+  }
```
seldridge, line 370: This method is a bit weird with it's use of `Option` and casts to `Seq` only to `flatMap`. Also, this is using recursion, but is not tail recursive. This would likely be better if rewritten to use iteration.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+
+  /** Access a source [[Data]] that may or may not be in the current module.  If
+    * this is in a child module, then create ports to allow access the
+    * requrested source.
+    */
+  def bore(source: Data)(implicit si: SourceInfo): Data = {
+    import reflect.DataMirror
+    def parent(d: Data): experimental.BaseModule = d.topBinding.location.get
+    def purePortType = if (DataMirror.hasOuterFlip(source)) Flipped(chiselTypeOf(source)) else chiselTypeOf(source)
+    def emptyWireOnError = Wire(purePortType) //return empty wire to continue elaboration to collect other errors
```
seldridge, line 216: Can this return `DontCare` or is that problematic due to that not working with zero-width?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+        return emptyWireOnError
+      case Some(internal.CrossModuleBinding) =>
+        Builder.error(
+          s"Cannot bore across a Definition/Instance boundary:${thisModule._errorContext} cannot access ${source}"
+        )
+        return emptyWireOnError
+      case _ => // Actually bore
+    }
+    if (parent(source) == thisModule) {
+
```
seldridge, line 232: Nit: weird spacing which scalaFmt would clean up.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+    }
+    val sink = if (downPath.isEmpty) { lcaSource }
+    else {
+      downPath.reverse.foldLeft(lcaSource) {
+        case (rhs, module) =>
+          chisel3.experimental.skipPrefix { // so `lcaSource` isn't in the name of the secret port
+            if (module.toDefinitionCalled) {
+              Builder.error(
+                s"Can only bore into modules that have not been converted to a Definition: ${module._errorContext}"
+              )
```
seldridge, line 273: Improving this to track what line did the `toDefinition` would help with debugability. That would make a nice separate PR!

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+            }
+
+            /** create a port. */
+            val bore = SecretIO(Flipped(purePortType), module)
+            module.addSecretIo(bore)
+            parent(rhs).asInstanceOf[RawModule].secretConnection(bore, rhs)
+            bore
+          }
+      }
+    }
```
seldridge, line 284: Almost all of these two code paths is the same. Can they be combined?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -193,4 +204,88 @@ object BoringUtils {
...
+
+    /** Creating an intermediate wire so secret stuff never escapes */
+    val bore = Wire(purePortType)
+    thisModule.asInstanceOf[RawModule].secretConnection(bore, sink)
```
seldridge, line 288: What is this doing?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +126,145 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
...
+      a := BoringUtils.bore(bar.baz.a_wire)
+      b := BoringUtils.bore(bar.b_wire)
+      c := BoringUtils.bore(c_wire)
+    }
+    //println(circt.stage.ChiselStage.emitCHIRRTL(new Foo, args))
+    (circt.stage.ChiselStage
+      .emitCHIRRTL(new Foo)
+      .split('\n')
+      .map(_.takeWhile(_ != '@'))
+      .map(_.trim) should contain).inOrder(
```
seldridge, line 159: This is getting thrown around so much. It would make a nice utility.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +126,145 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
...
+      "b_bore <= bar.b_bore"
+    )
+  }
+
+  it should "not work over a Definition/Instance boundary" in {
+    import chisel3.experimental.hierarchy._
+    @instantiable
+    class Bar extends RawModule {
+      @public val a_wire = WireInit(UInt(1.W), DontCare)
+      dontTouch(a_wire)
```
seldridge, line 181: Nit: dontTouch is unnecessary here.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +126,145 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
...
+    }
+    class Foo extends RawModule {
+      val a = IO(Output(UInt()))
```
seldridge, line 184: Nit: `val a` and the connection to it are not necessary.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -125,4 +126,145 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners {
...
+      val bar = Module(new Bar)
+      bar.toDefinition
+      a := BoringUtils.bore(bar.a_wire)
+    }
+
+    val e = intercept[Exception] {
+      circt.stage.ChiselStage.emitCHIRRTL(new Foo, args)
+    }
+    e.getMessage should include("Can only bore into modules that have not been converted to a Definition")
+  }
```
seldridge, line 269: Nice error test case. This could also be cleaned up by removing the dontTouch and port `a`.

seldridge (commented):
This is shaping up great!

Minor comments throughout. I do think that the second-class nature of `DataMirror` relative to `toDefinition` may not be a great API. I was always okay with `DataMirror` and `toDefinition` having the "effect" of making the ports visible (which is roughly the same as causing the module to close).

core/src/main/scala/chisel3/Module.scala
```diff
@@ -312,6 +317,8 @@ package experimental {
     /** Internal check if a Module is closed */
     private[chisel3] def isClosed = _closed
 
+    private[chisel3] var toDefinitionCalled = false
+
```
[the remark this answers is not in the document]
azidar: The reason I'm making the distinction is that the `_closed` check means that the normal phase is done, so creating `Wires` etc in a module should be illegal. It has implications for when the `component` is generated, as well as other kinds of checks internally.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
azidar: Hmm. This is only returned on the case when we are definitely throwing an error, but probably `DontCare` will work, I'll try it.

src/test/scala/chiselTests/BoringUtilsSpec.scala
azidar: Agreed, I'm fixing it to work the same way, that reflecting "fully closes the modules"

src/main/scala/chisel3/util/experimental/BoringUtils.scala
azidar: I'll add it to this PR, it's not too hard.

core/src/main/scala/chisel3/IO.scala
azidar: Added clear error messages, good discussion, thanks!

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -360,3 +360,22 @@ trait Utils {
...
+    val lines = output.split("\n").toList
+    val unmatched = matchList.flatMap { m =>
+      if (matches(lines, m).nonEmpty) None else Some(m)
+    }.map(x => s"  > $x was unmatched")
+    val unomitted = omitList.flatMap { o => omits(lines, o) }.map {
+      case (o, l) => s"  > $o was not omitted in ($l)"
+    }
+    val results = unmatched ++ unomitted
+    assert(results.isEmpty, results.mkString("\n"))
+  }
```
seldridge, line 380: Mega-nit: isn't this `matchesAndOmits`?

Non-blocking feedback: the multiple parameter lists can be hard to grok, but are necessary to have multiple variadic arguments. Using Arrays and named arguments when using this method may make it more obvious.

Alternatively, just having a FIRRTL trimming/canonicalizing utility and letting users do their own ScalaTest matching is more what I was thinking here.

src/test/scala/chiselTests/naming/PrefixSpec.scala
```diff
@@ -67,6 +67,38 @@ class PrefixSpec extends ChiselPropSpec with Utils {
...
+          "x1_wire1",
+          "x1",
+          "wire1",
+          "x2",
+          "wire1_1",
+          "wire2"
+        )
+      )
+    }
+  }
```
seldridge, line 100: An `emitCHIRRTL` test would be cleaner, IMO.

seldridge (approved):
LGTM

core/src/main/scala/chisel3/Module.scala
```diff
@@ -537,6 +557,35 @@ package experimental {
...
+    // data must be a fresh Chisel type
+    private[chisel3] def createSecretIO(data: => Data)(implicit sourceInfo: SourceInfo): Data = {
+      val iodef = data
+      // forcename
+      internal.requireIsChiselType(iodef, "io type")
+
+      if (toDefinitionCalled.nonEmpty)
+        Builder.error("can only create secret ports when a module has not been used as a Definition")
+      if (modulePortsAskedFor.nonEmpty)
+        Builder.error("can only create secret ports when a module has not been asked for all its IO")
```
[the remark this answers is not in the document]
azidar: My mistake, this is an internal error, i'll switch to require.

sequencer:
I wonder if it’s possible to use RefOp + XMR to replace the BoringUtils?

seldridge:
> I wonder if it’s possible to use RefOp + XMR to replace the BoringUtils?

We want both! This `bore` API is going to do normal port boring. We'll add a second API, `tap`, that will use the same algorithm, but do ref type ports on the upwards connection once https://github.com/chipsalliance/chisel/pull/3088 lands.

I.e., this is the exact same behavior as FIRRTL Dialect's `WiringProblem`. The awesome work Adam is doing here enables us to do all this work in Chisel.

sequencer:
Sounds great!

PR #3190, opened 2023-04-17
Implement read-write memory accessors for SyncReadMem

jared-barocsi:
Implements `SyncReadMem.readWrite(address, writeData, enabled, isWrite)` as a way to explicitly instantiate a readwrite port for a memory. This addresses the problem of unintentionally creating multiple read or write ports through multiple calls of `SyncReadMem.read()` and/or `SyncReadMem.write()`. Additionally, implements clocked and masked versions of `readWrite` for parity with the existing `read` and `write` functions.

Compare:
```
val myMem = SyncReadMem(100, UInt(32.W))
val readValue = myMem.read(someAddress)
when (timeToWrite)
  myMem.write(someAddress, valueToWrite)
when (timeToWrite) 
  myMem.write(someAddress, valueToWrite)
```
which instantiates a 1R2W memory, as opposed to
```
val myMem = SyncReadMem(100, UInt(32.W))
val rwPort = myMem.readWrite(someAddress, writeData, portEnabled, timeToWrite)

val readValue = rwPort
when (timeToWrite)
  writeData = valueToWrite
when (timeToWrite)
  writeData = differentValueToWrite
```
which instantiates a 1RW memory.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

squash and merge

#### Release Notes

`SyncReadMem.readWrite(address, writeData, enabled, isWrite)` explicitly generates a read-write port that supports both read and write access to the memory. 

- [ ] Did you review?


mwachs5:
Can we add this to the Memories explanations doc

mwachs5:
Or add a new cookbook "how to I make a 1readwrite port memory, how do I make a 1r1w port memory"?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,153 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
   }
   // note: we implement do_read(addr) for SyncReadMem in terms of do_read(addr, en) in order to ensure that
   //       `mem.read(addr)` will always behave the same as `mem.read(addr, true.B)`
+
+  /** Creates a read-write accessor into this SyncReadMem.
+    *
+    * @param idx memory element index to write into
+    * @param writeData new data to write
+    * @param enable enables access to the memory
+    * @param isWrite when enable is true, if this access is a write
```
mwachs5, line 401:
```suggestion
    * @param isWrite perform a write (vs a read) when enable is true. Read data is undefined when this is true.
```

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,153 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *
+    * @return the memory port accessor
```
mwachs5, line 403: what is "the memory port accessor"? Doesn't this just return the read data...? So should this be @return the read data, defined only when en is true and isWrite is false?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,153 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+
+  /** Creates a read-write accessor into this SyncReadMem.
+    *
+    * @param idx memory element index to write into
+    * @param writeData new data to write
+    * @param enable enables access to the memory
+    * @param isWrite when enable is true, if this access is a write
+    * @param clock clock to bind to this accessor
+    *
+    * @return the memory port accessor
```
mwachs5, line 419: similar comments as above

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,153 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+  /** Creates a masked read-write accessor into this SyncReadMem.
+    *
+    * @param idx memory element index to write into
+    * @param writeData new data to write
+    * @param mask the write mask as a Seq of Bool: a write to the Vec element in
+    * memory is only performed if the corresponding mask index is true.
+    * @param enable enables access to the memory
+    * @param isWrite when enable is true, if this access is a write
+    *
+    * @return the memory port accessor
```
mwachs5, line 471: similar comments as above

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,153 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    * @param idx memory element index to write into
+    * @param writeData new data to write
+    * @param mask the write mask as a Seq of Bool: a write to the Vec element in
+    * memory is only performed if the corresponding mask index is true.
+    * @param enable enables access to the memory
+    * @param isWrite when enable is true, if this access is a write
+    * @param clock clock to bind to this accessor
+    *
+    * @return the memory port accessor
+    * @note this is only allowed if the memory's element data type is a Vec
```
mwachs5, line 495: can we add an @example on how people are supposed to use this? Are they supposed to embed it in a `when` or are they expected to compute the `enable` and `isWrite` outside of the call to `readwrite`?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *   val mem = SyncReadMem(2, UInt(width.W))
+    *   val rwPort = mem.readWrite(io.address, io.wdata, io.enable, io.isWrite)
+    *   io.rdata := rwPort
+    * }
+    *
+    * class Top extends Module {
+    *   val mem = Module(new MyMemWrapper)
+    *   mem.io := DontCare
+    *
+    *   val rdata = Wire(UInt(2.W))
```
mwachs5, line 431: this code would not quite compile

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *   readData := DontCare
```
mwachs5, line 432:
```suggestion
    *   val rdata = Wire(UInt(2.W))
    *   rdata := DontCare
```

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *
+    *   // Read from the memory port
+    *   when(doRead) {
```
mwachs5, line 435: what is `doRead`? an Input?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *     mem.io.enable := true.B
+    *     mem.io.isWrite := false.B
+    *     mem.io.address := someAddress
+    *   }
+    *   // Read data will show up on the next cycle
+    *   when(timeToRead) {
+    *     readData := mem.io.rdata
```
mwachs5, line 442: this is a little confusing, the user is going to have to understand that the rddata is available.

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    * @param idx memory element index to write into
+    * @param writeData new data to write
+    * @param enable enables access to the memory
+    * @param isWrite performs a write instead of a read when enable is true; the return
+    * value becomes undefined when this parameter is true
+    * @param clock clock to bind to this read-write port
+    *
+    * @return The value of the memory at idx when enable is true and isWrite is false,
+    * or an undefined value otherwise.
+    */
```
mwachs5, line 473:
```suggestion
    * @return The read port of the memory, which will return the value of the memory at idx when enable is true and isWrite is false,
    * or an undefined value otherwise, on the next rising clock edge
    */
```

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,272 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *   readData := DontCare
+    *
+    *   // Read from the memory port
+    *   when(doRead) {
+    *     mem.io.enable := true.B
+    *     mem.io.isWrite := false.B
+    *     mem.io.address := someAddress
+    *   }
+    *   // Read data will show up on the next cycle
+    *   when(timeToRead) {
```
mwachs5, line 562:
```suggestion
    *   when(ShiftReg(doRead) {
```

maybe? 

I think this might be too fancy an example, I think the `MyMaskedMemWrapper` is probably fine?

docs/src/explanations/memories.md
```diff
@@ -114,6 +114,28 @@ Here is an example single read/write port waveform, with [masks](#masks) (again,
...
+    val addr = Input(UInt(10.W))
+    val dataIn = Input(UInt(width.W))
+    val dataOut = Output(UInt(width.W))
+  })
+
+  val mem = SyncReadMem(1024, UInt(width.W))
+  val rwPort = mem.readWrite(io.addr, io.dataIn, io.enable, io.write)
+
+  when(io.enable && !io.write) {
+    io.dataOut := rwPort
```
mwachs5, line 134: this seems wrong. Isn't it off by a cycle? I think we should just assign io.dataOut := rwPort without any when

docs/src/explanations/memories.md
```diff
@@ -114,6 +114,28 @@ Here is an example single read/write port waveform, with [masks](#masks) (again,
...
+  val io = IO(new Bundle {
+    val enable = Input(Bool())
+    val write = Input(Bool())
+    val addr = Input(UInt(10.W))
+    val dataIn = Input(UInt(width.W))
+    val dataOut = Output(UInt(width.W))
+  })
+
+  val mem = SyncReadMem(1024, UInt(width.W))
+  val rwPort = mem.readWrite(io.addr, io.dataIn, io.enable, io.write)
```
mwachs5, line 131:
```suggestion
  io.dataOut := mem.readWrite(io.addr, io.dataIn, io.enable, io.write)
```

I think it's just this?

mwachs5:
This LGTM. I am not sure if there are any issues on the CIRCT side in interpreting this. Can @jackkoenig @seldridge give a review?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,207 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    data:    T,
+    enable:  Bool,
+    isWrite: Bool,
+    clock:   Clock,
+    warn:    Boolean
+  )(
+    implicit sourceInfo: SourceInfo
+  ): T = {
+    val a = Wire(UInt())
+    a := DontCare
```
seldridge, line 474: Using `val _a` indicates better that this is a temporary. Same for `port`. Basically, I would use `_` for _any_ hardware component created in a library/utility.

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,207 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    *     val isWrite = Input(Bool())
+    *     val rdata = Output(Vec(2, UInt(width.W)))
+    *   })
+    *
+    *   val mem = SyncReadMem(2, Vec(2, UInt(width.W)))
+    *   io.rdata := mem.readWrite(io.address, io.wdata, io.mask, io.enable, io.isWrite)
+    * }
+    * }}}
+    *
+    * @note this is only allowed if the memory's element data type is a Vec
```
seldridge, line 521: At least in FIRRTL this is allowed so long as the type of the mask is the same type as the underlying data type with the leaf types as all 1-bit UInts. I.e., this is supported on any aggregate including bundles.

This is a relatively niche feature and pretty rare. I am fine to start this as only working for `Vec` (which makes sense).

core/src/main/scala/chisel3/Mem.scala
seldridge: Nits/golf:
```suggestion
    val _a = WireDefault(chiselTypeOf(addr), DontCare)
```

seldridge: Same comments for the masked version.

src/test/scala/chiselTests/Mem.scala
```diff
@@ -188,6 +188,156 @@ private class TrueDualPortMemory(addrW: Int, dataW: Int) extends RawModule {
   }
 }
 
+class MemReadWriteTester extends BasicTester {
```
seldridge, line 191: I'm not a huge fan of big integration tests like this. However, this is entirely consistent with the rest of this existing file.

Can you open an issue to move all refactor this to move the integration tests into the integration test area and slim these down to just checking the CHIRRTL?

seldridge (approved):
A long missing API. Great work!

Minor nits throughout. Good to land whenever you're ready.

seldridge:
I didn't notice you had auto-merge on. 😂  Any cleanup/feedback can happen post-merge. 😉

core/src/main/scala/chisel3/Mem.scala
jackkoenig: Is the declaration of `a` separate from the connection of `addr` to it required for the `enable` to infer correctly @jared-barocsi? Or why else is is split?

Also are we sure returning the port works? Such value capturing is often problematic for lexical scoping, what happens if someone uses the return value of this function? Perhaps we should be assigning the read data to a wire? The integration test is passing so that suggests the lexical scoping is okay or at least supported by firtool but I'd be curious to see what the FIRRTL actually looks like.

src/test/scala/chiselTests/Mem.scala
```diff
@@ -188,6 +188,156 @@ private class TrueDualPortMemory(addrW: Int, dataW: Int) extends RawModule {
...
+    is(1.U) { // Cycle 2: Write 2.U to address 1
+      address := 1.U
+      enable := true.B
+      isWrite := true.B
+      wdata := 2.U
+    }
+    is(2.U) { // Cycle 3: Read from address 0 (data returned next cycle)
+      address := 0.U;
+      enable := true.B;
+      isWrite := false.B;
```
jackkoenig, line 229: Can you please remove all of the semicolons in this file?

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -392,4 +392,207 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
...
+    mask:    Seq[Bool],
+    enable:  Bool,
+    isWrite: Bool,
+    clock:   Clock,
+    warn:    Boolean
+  )(
+    implicit evidence: T <:< Vec[_]
+  ): T = {
+    implicit val sourceInfo = UnlocatableSourceInfo
+    val a = Wire(UInt())
```
jackkoenig, line 574: It would also be good to have a source locator here, since we're already taking an implicit argument we can just take the sourceInfo argument. Users won't be able to immediately call `.apply` on the return value, but due to the existing implicit argument, they already can't.

PR #3196, opened 2023-04-19
Add --module-json option to ChiselStage

seldridge:
Add a new option, "--module-json", to construct a
ChiselGeneratorAnnotation, from a class name and some JSON text.  The JSON text, using "class" full type hints, will be deserialized to an object and passed as the single parameter to a Chisel generator.  This is intended to be used by tools which want to do "inline generation" (to call out to a Chisel generator to fill in a module body, e.g., by Firesim).

This is available by default by ChiselStage and can be used by any non-ChiselStage tool that supports ChiselGeneratorAnnotation.

CC: @nandor

#### Release Notes

Add a "--module-json" option to ChiselGeneratorAnnotation and ChiselStage.

### DIFF
--- a/src/main/scala/chisel3/stage/ChiselAnnotations.scala
+++ b/src/main/scala/chisel3/stage/ChiselAnnotations.scala
@@ -218,6 +218,19 @@ object ChiselGeneratorAnnotation extends HasShellOptions {
     )
   )
 
+  /** Utility to write an arbitrary parameter to a JSON format that --module-json will understand.
+    * @param parameter the parameter to serialize to a JSON string
+    * @param classes a list of all classes that may exist recursively inside the parameter
+    */
+  def serializeJSON(parameter: Any, classes: List[Class[_]]): String = {
+    import org.json4s._
+    import org.json4s.native.JsonMethods._
+    import org.json4s.native.Serialization
+
+    implicit val defaultFormat = Serialization.formats(FullTypeHints(classes, "class"))
+    Serialization.write(parameter)
+  }
+
 }
 
 /** Stores a Chisel Circuit
--- a/src/main/scala/chisel3/stage/ChiselAnnotations.scala
+++ b/src/main/scala/chisel3/stage/ChiselAnnotations.scala
@@ -145,12 +145,76 @@ object ChiselGeneratorAnnotation extends HasShellOptions {
     ChiselGeneratorAnnotation(gen)
   }
 
+  /** Construct a [[ChiselGeneratorAnnotation]] with a generator function that
+    * will try to construct a Chisel Module from using that Module's name and a
+    * single parameters object. The Module must both exist in the class path and
+    * take exactly the one parameter of the provided type.
+    * @param name a module name
+    * @throws firrtl.options.OptionsException if the module name is not found or if no parameterless constructor for
+    * that Module is found.
+    */
+  def apply(name: String, parameters: Any): ChiselGeneratorAnnotation = {
+    val gen = () =>
+      try {
+        Class
+          .forName(name)
+          .asInstanceOf[Class[_ <: RawModule]]
+          .getDeclaredConstructor(parameters.getClass())
+          .newInstance(parameters)
+      } catch {
+        // The reflective instantiation will box any exceptions thrown, unbox them here.
+        // Note that this does *not* need to chain with the catches below which are triggered by an
+        // invalid name or a constructor that takes arguments rather than by the code being run
+        // itself.
+        case e: InvocationTargetException =>
+          throw e.getCause
+        case e: ClassNotFoundException =>
+          throw new OptionsException(s"Unable to locate module '$name'! (Did you misspell it?)", e)
+        case e: NoSuchMethodException =>
+          throw new OptionsException(
+            s"Unable to create instance of module '$name' with parameter type '${parameters.getClass().getName()}'!",
+            e
+          )
+      }
+    ChiselGeneratorAnnotation(gen)
+  }
+
   val options = Seq(
     new ShellOption[String](
       longOption = "module",
       toAnnotationSeq = (a: String) => Seq(ChiselGeneratorAnnotation(a)),
       helpText = "The name of a Chisel module to elaborate (module must be in the classpath)",
       helpValueName = Some("<package>.<module>")
+    ),
+    new ShellOption[String](
+      longOption = "module-json",
+      toAnnotationSeq = (a: String) => {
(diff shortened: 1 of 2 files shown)


nandor:
To put my mind at ease, it would be great if a test included an example of `CustomFileEmission` still working.

seldridge:
@nandor: `CustomFileEmission` works as expected. 👍 Commit https://github.com/chipsalliance/chisel/pull/3196/commits/7ee7931ff3ea7007ffe0406c84a4921802694b5d shows this. (This could be backed out before merging as `CustomFileEmission` is a very separate code path that is already tested. Basically, as long as annotations are created properly, any that extend `CustomFileEmission` will definitely work.)

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -145,15 +145,92 @@ object ChiselGeneratorAnnotation extends HasShellOptions {
     ChiselGeneratorAnnotation(gen)
   }
 
+  /** Construct a [[ChiselGeneratorAnnotation]] with a generator function that
+    * will try to construct a Chisel Module from using that Module's name and a
+    * single parameters object. The Module must both exist in the class path and
```
mwachs5, line 150: can you clarify the commends, so a parameters object is something that is like a case class that is trivially deserialized from JSON? Or it is a class name that points to a specific Object?

src/test/scala/circtTests/stage/ChiselStageSpec.scala
```diff
@@ -334,6 +364,60 @@ class ChiselStageSpec extends AnyFunSpec with Matchers with chiselTests.Utils {
...
       os.read(targetDir / "HasUnserializableAnnotation.fir") shouldNot include("DummyAnnotation")
     }
+
+    it("should allow building a module from JSON") {
+      import ChiselStageSpec.{Corge, Parameters}
+
+      val targetDir = baseDir / "should-allow-building-a-module-from-JSON"
+
+      val clazz = classOf[Corge].getName()
+      val json = ChiselGeneratorAnnotation.serializeJSON(Parameters(42), classOf[Parameters] :: Nil)
```
mwachs5, line 374: what is in this json file? Would it be worth having this test show it / write one out explicitly? is the JSON supposed to stand on its own or only make sense if we call it from something ? What is the relevance of evaluating `clazz` could I just make up a string here and everything just works?

seldridge:
@sequencer added a very, very similar API here: https://github.com/chipsalliance/chisel/pull/2857

A minimal example of using this is here:

```scala
//> using scala "2.13.10"
//> using lib "org.chipsalliance::chisel::5.0.0-M1+32-baed66d4-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::5.0.0-M1+32-baed66d4-SNAPSHOT"
//> using options "-Ymacro-annotations"

import chisel3._
import chisel3.experimental.{SerializableModule, SerializableModuleGenerator, SerializableModuleParameter}
import upickle.default.{macroRW, ReadWriter}

case class FooParameter(width: Int) extends SerializableModuleParameter

object FooParameter {
  implicit def rwP: ReadWriter[FooParameter] = macroRW
}

class Foo(val parameter: FooParameter) extends Module with SerializableModule[FooParameter] {
  val a = IO(Input(UInt(parameter.width.W)))
  val b = IO(Output(UInt(parameter.width.W)))

  b := a
}

object Main extends App {
  val generator = SerializableModuleGenerator(classOf[Foo], FooParameter(16))

  val json = upickle.default.write(generator)

  println(
    ChiselStage.emitSystemVerilog(
      gen = upickle.default.read[SerializableModuleGenerator[Foo, FooParameter]](json).module(),
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

The main difference from this PR is that the existing API doesn't rely on reflection to figure out what to build. E.g., you need to do `read[SerializableModuleGenerator[Foo, FooParameter]](json)` and can't just do a `read(json)` and have it figure out what to build. This could be removed with more information passed to `--module-json` or with some custom pickling.

mwachs5:
I guess my question is whether this JSON format is intended to be human understandable / generatable, or is this just a generic "Serialize and deserialize a scala object through JSON, don't worry about how  and why"?

src/test/scala/circtTests/stage/ChiselStageSpec.scala
seldridge: This is the exact output:

```json
{"class":"circtTests.stage.ChiselStageSpec$Parameters","width":42}
```

The API is `--module-json <class-name-string>,<json-string>`. `clazz` isn't being evaluated. I'm programmatically calling the `getName` method in order to not hard-code `chiselTests.whatever.Corge`.

seldridge:
It's intended for machine consumption. It is descriptive enough that a human could write it if they wanted to.

seldridge:
@nandor has a use case where he wants to be able to do reflective reading. upickle, because it targets ScalaJS, cannot support reflective reading/writing. I'll open an issue about how to change/improve this to support this use case.

sequencer:
@seldridge 
I'm drafting a new API to avoid the type parameter, since we found it might be impossible for upickle to deal with class with type parameters.
```
trait SerializableModule { this: BaseModule =>
  type SerializableModuleParameter
  val parameter: SerializableModuleParameter
}

trait SerializableModuleGenerator {
  type M <: SerializableModule
  val parameter: M#SerializableModuleParameter
  // user should implement the serialization for M#SerializableModuleParameter 
  implicit def parameterRW: ReadWriter[M#SerializableModuleParameter]

  // use tt to find the class name for serialization
  protected[chisel3] val mTypeTag: TypeTag[M] = typeTag[M]
  protected[chisel3] val moduleClassName = mTypeTag.tpe.typeSymbol.name.toString
  protected[chisel3] val serializedParameter = upickle.default.writeJs(parameter)
}
```
In this case, it won't exist any type parameter and the generator can live in the collection, and the reflective reading/writing can be resolved as well.

seldridge:
@sequencer: I was playing around with a library for building arbitrary Scala objects and calling methods on them here: https://github.com/seldridge/reflective-builder

See if this, or pieces of this, would work here. ChiselStage already has the --module option. It's only really missing a variadic --argument option. I got the latter working in the reflective builder repo.

sequencer:
Sounds interesting, I'm working on another idea of taking full advantage of macro + reflection, so the user API will still happy without the problem of type parameter.

PR #3199, opened 2023-04-24
[Feature] Add Interface, ConformsTo for Separate Compilation

seldridge:
Migrate work from [seldridge/chisel-separable](https://github.com/seldridge/chisel-separable) on separate compilation into upstream Chisel.  This defines an Interface and a Conformance.  An Interface is a Singleton Object that maps to a FIRRTL external module with pre-defined IO.  A Conformance is a type-class whose implementations define how a Chisel module aligns itself to the Interface.  Used together, these enable separate compilation of a client that uses an Interface from a component that implements the Interface.

This originally came from https://github.com/seldridge/chisel-separable/. We've used this internally in non-load-bearing ways and it's becoming unwieldy to continue to maintain out-of-tree.

This API may change rapidly in the future.

#### Release Notes

Add Interface/ConformsTo APIs for separate compilation.

### DIFF
--- a/src/main/scala/chisel3/interface/Interface.scala
+++ b/src/main/scala/chisel3/interface/Interface.scala
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: Apache-2.0
+package chisel3.interface
+
+import chisel3.{BlackBox => _, Module => _, _}
+import chisel3.experimental.{BaseModule, FlatIO}
+import chisel3.experimental.dataview._
+import scala.annotation.implicitNotFound
+
+@implicitNotFound(
+  "this method requires information from the separable compilation implementation, please bring one into scope as an `implicit val`. You can also consult the team that owns the implementation to refer to which one you should use!"
+)
+trait ConformsTo[Intf <: Interface, Mod <: BaseModule] {
+
+  /** Return the module that conforms to a port-level interface. */
+  private[interface] def genModule(): Mod
+
+  /** Define how this module hooks up to the port-level interface. */
+  private[interface] def portMap: Seq[(Mod, Intf#Ports) => (Data, Data)]
+
+}
+
+/** Functionality which is common */
+sealed trait InterfaceCommon {
+
+  private[interface] type Ports <: Record
+
+  /** Returns the Record that is the port-level interface. */
+  private[interface] def ports(): Ports
+
+}
+
+/** A generator of different Interfaces. Currently, this is just an Interface
+  * that is not a Singleton.
+  */
+trait InterfaceGenerator extends InterfaceCommon
+
+/** An interface between hardware units. Any module that implements this
+  * interface may be separately compiled from any module that instantiates this
+  * interface.
+  */
+trait Interface extends InterfaceCommon { self: Singleton =>
+
+  /** This types represents the type of a valid conformance to this Interface.
+    */
+  private type Conformance[Mod <: RawModule] = ConformsTo[this.type, Mod]
+
+  /** The name of this interface. This will be used as the name of any module
+    * that implements this interface.
+    * I.e., this is the name of the `BlackBox` and `Module` that are provided
+    * below. The implementation of this method is just coming up with a good
+    * name derived from the class name. (The name of the Interface in FIRRTL
+    * will be the name of the Scala class that extends the Interface.)
+    */
+  private[interface] def interfaceName: String = {
+    val className = getClass().getName()
+    className
+      .drop(className.lastIndexOf('.') + 1)
+      .split('$')
+      .toSeq
+      .filter {
+        case str if str.forall(_.isDigit) => false
+        case str                          => true
+      }
+      .last
+  }
+
+  sealed trait Entity { this: BaseModule =>
+    val io: Ports
+  }
+
+  object Wrapper {
+
+    /** The black box that has the same ports as this interface. This is what is
+      * instantiated by any user of this interface, i.e., a test harness.
+      */
+    final class BlackBox extends chisel3.BlackBox with Entity {
+      final val io = IO(ports())
+
+      override final def desiredName = interfaceName
+    }
+
+    /** The module that wraps any module which conforms to this Interface.
+      */
+    final class Module[B <: RawModule](
+    )(
+      implicit conformance: Conformance[B])
+        extends RawModule
+        with Entity {
+      final val io = FlatIO(ports())
+
+      // Use a dummy clock and reset connection when constructing the module.
+      // This is fine as we rely on DataView to catch missing connections to
+      // clock and reset.  The dummy clock and reset will never be used.
+      private val internal = withClockAndReset(
+        WireInit(Clock(), DontCare),
+        WireInit(Reset(), DontCare)
+      ) {
+        chisel3.Module(conformance.genModule())
+      }
+
+      private implicit val pm = PartialDataView[B, Ports](
+        _ => ports(),
+        conformance.portMap: _*
+      )
+      io :<>= internal.viewAs[Ports]
+
+      override def desiredName = interfaceName
+    }
+
+    /** A stub module that implements the interface. All IO of this module are
+      * just tied off.
+      */
+    final class Stub extends RawModule with Entity {
+      final val io = FlatIO(ports())
+      io := DontCare
+      dontTouch(io)
+    }
+
+  }
+
+}
--- a/src/main/scala/chisel3/interface/Interface.scala
+++ b/src/main/scala/chisel3/interface/Interface.scala
@@ -34,6 +34,20 @@ sealed trait InterfaceCommon {
   */
 trait InterfaceGenerator extends InterfaceCommon
 
+/** Utilities related to Interface infastructure */
+object Interface {
+
+  /** An exception that indicates that the component-side conformance failed. This
+    * is used to wrap other exceptions which may hide the cause of the invalid
+    * conformance, e.g., a view was non-total.
+    * @param message the message for this exception
+    * @param cause an optional exception that caused this exception to occur
+    */
+  private case class InvalidConformance(message: String, cause: Throwable = null)
+      extends ChiselException(message, cause)
+
+}
+
 /** An interface between hardware units. Any module that implements this
   * interface may be separately compiled from any module that instantiates this
   * interface.
@@ -102,7 +116,17 @@ trait Interface extends InterfaceCommon { self: Singleton =>
         _ => ports(),
         conformance.portMap: _*
       )
-      io :<>= internal.viewAs[Ports]
+
+      // If the view fails, report this witha  slightly better error message.
+      try {
+        io :<>= internal.viewAs[Ports]
+      } catch {
+        case e: InvalidViewException =>
+          throw Interface.InvalidConformance(
+            s"unable to conform module '${internal.name}' to interface '$interfaceName' (see InvalidViewException below for more information)",
+            e
+          )
+      }
 
       override def desiredName = interfaceName
     }
--- a/src/main/scala/chisel3/interface/Interface.scala
+++ b/src/main/scala/chisel3/interface/Interface.scala
@@ -56,7 +56,7 @@ trait Interface extends InterfaceCommon { self: Singleton =>
 
   /** This types represents the type of a valid conformance to this Interface.
     */
-  private type Conformance[Mod <: RawModule] = ConformsTo[this.type, Mod]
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,121 @@
...
+        WireInit(Reset(), DontCare)
+      ) {
+        chisel3.Module(conformance.genModule())
+      }
+
+      private implicit val pm = PartialDataView[B, Ports](
+        _ => ports(),
+        conformance.portMap: _*
+      )
+      io :<>= internal.viewAs[Ports]
```
mwachs5, line 105: curious will this error if any ports of the Module were missed in this?

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,121 @@
...
+      // This is fine as we rely on DataView to catch missing connections to
+      // clock and reset.  The dummy clock and reset will never be used.
+      private val internal = withClockAndReset(
+        WireInit(Clock(), DontCare),
+        WireInit(Reset(), DontCare)
+      ) {
+        chisel3.Module(conformance.genModule())
+      }
+
+      private implicit val pm = PartialDataView[B, Ports](
```
trilorez, line 101: Can we have some sort of less-strict "do your own wiring" alternative to at least support the Circt Chisel interface stub approach? Something like the `useStructuralConnect` option I prototyped but maybe more flexible?

mwachs5: agree that this portMap style should be more a  convenience than a requirement...?

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,121 @@
+// SPDX-License-Identifier: Apache-2.0
+package chisel3.interface
```
jackkoenig, line 2: I would prefer putting this in `chisel3.experimental.interface` at least while we iterate on the API

src/main/scala/chisel3/interface/Interface.scala
seldridge: I'm looking at a way that this can work. The specific problem (thanks for more side-band info, Daniel) is that we'd like to be able to connect an anonymous `Bundle` or multiple, nested anonymous `Bundle`s and make this pattern still usable. `DataView` does a check that classes match which makes this funky.

Daniel added a `useStructuralConnect` to this internally which does a weaker pairwise connection of the pairs defined in the `PortMap` that does not use `DataView`.

seldridge: This does work today, but it requires that the portMap is blown out into all the constituent connections. E.g., if you have an anonymous `Bundle` in the `Component`:

```scala
class Qux extends RawModule {
  val aa = IO(Input(new Bundle {
    val bb = Bool()
  }))
  val cc = IO(Output(new Bundle {
    val dd = Bool()
  }))

  cc.dd := aa.bb
}

class Foo extends RawModule {

  val a = IO(Input(Bool()))
  val b = IO(Output(Bool()))

  val qux1, qux2 = chisel3.Module(new BundleInterface.Wrapper.BlackBox)

  qux1.io.a.b := a
  qux2.io.a.b := qux1.io.c.d
  b := qux2.io.c.d
}
```

You need to blow out the connections in the `ConformsTo`:

```scala
implicit val quxConformance =
  new ConformsTo[BundleInterface.type, Qux] {

    override def genModule() = new Qux

    override def portMap = Seq(
      _.aa.bb -> _.a.b,
      _.cc.dd -> _.c.d
    )

  }
```

You cannot do the following:

```scala
```

seldridge: Instead of doing the blown-out connection, a second `DataView` could be used that defines the mapping between the structural, anonymous `Bundle` and the named `Bundle`. However, this is doing essentially the same work as blowing out the connections in the conformance.

@trilorez: Are you fine with keeping this more restrictive for now?

trilorez: Does this second `DataView` approach only handle a single level of bundles? In the "simple" example I'm working with, there are up to 4 levels of nested bundles.

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,121 @@
...
+
+/** An interface between hardware units. Any module that implements this
+  * interface may be separately compiled from any module that instantiates this
+  * interface.
+  */
+trait Interface extends InterfaceCommon { self: Singleton =>
+
+  /** This types represents the type of a valid conformance to this Interface.
+    */
+  private type Conformance[Mod <: RawModule] = ConformsTo[this.type, Mod]
```
trilorez, line 45: For working with `ExtModule`, I needed to change `RawModule` to `BaseModule` here and below

src/main/scala/chisel3/interface/Interface.scala
seldridge: The actual `io :<>= internal.viewAs[Ports]` connection should always be correct as this is connection `Ports :<>= Ports`. If the view is non-total, then an exception is thrown. I added some checking and a test in https://github.com/chipsalliance/chisel/pull/3199/commits/34cf05bd7c3a7707c92cb528eda166d941d88eff that makes it more obvious that the user has an invalid conformance.

src/main/scala/chisel3/interface/Interface.scala
seldridge: Good catch! Fixed in https://github.com/chipsalliance/chisel/pull/3199/commits/b8a5fc0a2a8c0e455147143388fdb96d33593d0e.

src/main/scala/chisel3/interface/Interface.scala
seldridge: It will nest as far as you want. However, the information content is still going to be the same. The blown out connections are still there, just distributed one level at a time. Consider:

```scala
//> using scala "2.13.10"
//> using lib "org.chipsalliance::chisel::5.0.0-M2"
//> using plugin "org.chipsalliance:::chisel-plugin::5.0.0-M2"
//> using options "-Ymacro-annotations"

import chisel3._
import java.awt.MenuBar

class Baz extends Bundle {
  val a = Bool()
}

class Bar extends Bundle {
  val a = new Baz
}

class Foo extends Module {

  private val x = Wire(new Bar)
  private val y = Wire(new Bundle {
    val b = new Bundle {
      val b = Bool()
    }
  })

  implicit val view4 = PartialDataView[Bundle { val b: Bool }, Baz](
    _ => new Baz,
    _.b -> _.a
  )

  implicit val view3 =
    PartialDataView[Bundle { val b: Bundle { val b: Bool } }, Bar](
      _ => new Bar,
      _.b.viewAs[Baz] -> _.a
    )

  y.viewAs[Bar] :<>= x

}

object Main extends App {
  println(
    ChiselStage.emitCHIRRTL(
      new Foo,
      Array("--full-stacktrace")
    )
  )
}
```

src/main/scala/chisel3/interface/Interface.scala
seldridge: Discussed with you offline. I'd like to kill `experimental` for new features going forward. We should also promote everything out of there that's currently in there.

src/main/scala/chisel3/interface/Interface.scala
trilorez: Ah, I see. Yeah, this works but it's cumbersome. It seems like more flexibility would be generally useful, at least to facilitate experimentation. But we can start here and add flexibility later.

seldridge: We can change this in the future either by relaxing how the connections work or by improvements to `DataView`. `DataView` already needs improvements around bit select and padding. We'll likely need even more power here. We can start with the strict approach and expand it as needed.

jackkoenig (approved):
A few nits but generally LGTM

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,145 @@
...
+    val className = getClass().getName()
+    className
+      .drop(className.lastIndexOf('.') + 1)
+      .split('$')
+      .toSeq
+      .filter {
+        case str if str.forall(_.isDigit) => false
+        case str                          => true
+      }
+      .last
```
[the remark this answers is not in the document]
seldridge: Nice one! ⛳

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,145 @@
...
+/** Utilities related to Interface infastructure */
+object Interface {
+
+  /** An exception that indicates that the component-side conformance failed. This
+    * is used to wrap other exceptions which may hide the cause of the invalid
+    * conformance, e.g., a view was non-total.
+    * @param message the message for this exception
+    * @param cause an optional exception that caused this exception to occur
+    */
+  private case class InvalidConformance(message: String, cause: Throwable = null)
```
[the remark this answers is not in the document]
seldridge: I intentionally made this uncatchable as I didn't want users to do arbitrary things with it. I'll keep it private and open it up in the future if there is a good use case.

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,145 @@
...
+      ) {
+        chisel3.Module(conformance.genModule())
+      }
+
+      private implicit val pm = PartialDataView[B, Ports](
+        _ => ports(),
+        conformance.portMap: _*
+      )
+
+      // If the view fails, report this witha  slightly better error message.
```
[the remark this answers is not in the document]
seldridge: 🙏

src/main/scala/chisel3/interface/Interface.scala
```diff
@@ -0,0 +1,145 @@
...
+      try {
+        io :<>= internal.viewAs[Ports]
+      } catch {
+        case e: InvalidViewException =>
+          throw Interface.InvalidConformance(
+            s"unable to conform module '${internal.name}' to interface '$interfaceName' (see InvalidViewException below for more information)",
+            e
+          )
```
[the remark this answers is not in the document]
seldridge: The intent is to provide a better error message than just the invalid view. The view is something that is super internal and is supposed to be hidden from the user. However, if it fails, the user likely needs both the information that: (1) the conformance failed and (2) the internal view failed because some elements were unmatched.

At least in the test it looks okay-ish. The stack trace is trimmed to user code for the first exception and then shows everything for the later one:

```
[info] - should error if an Interface is not connected to *** FAILED ***
[info]   chisel3.interface.Interface$InvalidConformance: unable to conform module 'Qux' to interface 'BarInterface' (see InvalidViewException below for more information)
[info]   at ... ()
[info]   at chiselTests.interface.InterfaceSpec.$anonfun$new$15(InterfaceSpec.scala:170)
[info]   at ... ()
[info]   at ... (Stack trace trimmed to user code only. Rerun with --full-stacktrace to see the full stack trace)
[info]   ...
[info]   Cause: chisel3.experimental.dataview.package$InvalidViewException: Viewing chiselTests.interface.InterfaceSpec$Qux$1@2453f3e3 as BarBundle is non-Total!
[info]   View fields '_.a, _.b' are missing.
[info]   DataView used is PartialDataView(defined @[src/main/scala/chisel3/interface/Interface.scala 111:58]).
[info]   at chisel3.experimental.dataview.package$.nonTotalViewException(package.scala:70)
[info]   at chisel3.experimental.dataview.package$.chisel3$experimental$dataview$package$$doBind(package.scala:169)
[info]   at chisel3.experimental.dataview.package$DataViewable.viewAs(package.scala:28)
[info]   at chisel3.interface.Interface$Wrapper$Module.$anonfun$new$1(Interface.scala:118)
[info]   at chisel3.experimental.prefix$.apply(prefix.scala:33)
[info]   at chisel3.connectable.Connectable$ConnectableOpExtension.$colon$less$greater$eq(Connectable.scala:319)
[info]   at chisel3.interface.Interface$Wrapper$Module.<init>(Interface.scala:118)
[info]   at chiselTests.interface.InterfaceSpec.$anonfun$new$15(InterfaceSpec.scala:170)
[info]   at chisel3.Module$.do_apply(Module.scala:53)
[info]   at chisel3.stage.phases.Elaborate.$anonfun$transform$2(Elaborate.scala:40)
[info]   ...
```

It looks like the stack trace trimming is not applied to nested exceptions.

PR #3222, opened 2023-04-26
Enable push CI on all *.x branches

jackkoenig:


seldridge:
Nice future-proofing!

PR #3228, opened 2023-05-01
Implement a SyncReadMem wrapper with explicit read, write, and read/write ports

jared-barocsi:
This PR implements a `SRAM` object, which wraps a `SyncReadMem` and instantiates a desired number of read, write, and read/write ports.

For example:
```
val memory = SRAM(UInt(8.W), 1000, 2, 2, 1)
```
generates a 1R, 2W, 1RW `SyncReadMem` with a size of 1000, and is controlled like so:
```
memory.rd(0).enable := timeToRead
memory.rd(0).addr := 3.U
// one cycle later after timeToRead is driven
fooData := memory.rd(0).readValue

memory.wr(1).enable := timeToWrite
memory.rw(1).addr := 100.U
memory.wr(1).writeValue := barData

memory.rw(0).enable := timeToReadWrite
memory.rw(0).isWrite := timeToWrite
memory.rw(0).writeData := bazData
memory.rw(0).addr := 50.U
// one cycle later when timeToReadWrite is driven but not timeToWrite
fizzData := memory.rw(0).readValue
```

### Contributor Checklist

#### Desired Merge Strategy

Squash and merge

#### Release Notes
This `SyncReadMem` wrapper is instantiated using a new object, `SRAM.apply`, and invokes `.write`, `.read`, and `.readWrite` to generate a desired number of read, write, and read/write ports. This function returns a new `Bundle` wire containing the control signals for each requested port.

- [ ] Did you review?


src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,187 @@
...
+import chisel3.internal.sourceinfo.{MemTransform, SourceInfoTransform}
+import scala.language.reflectiveCalls
+import scala.language.experimental.macros
+
+/** A bundle of signals representing a write memory port.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param width The width of the address signal
+  */
+class RDPortInterface[T <: Data](tpe: T, width: Int) extends Bundle {
```
mwachs5, line 16:
```suggestion
class MemRdPortInterface[T <: Data](tpe: T, width: Int) extends Bundle {
```

WDYT?

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,187 @@
...
+
+import chisel3._
+
+import chisel3.internal.Builder
+import chisel3.experimental.SourceInfo
+import chisel3.internal.sourceinfo.{MemTransform, SourceInfoTransform}
+import scala.language.reflectiveCalls
+import scala.language.experimental.macros
+
+/** A bundle of signals representing a write memory port.
```
mwachs5, line 11:
```suggestion
/** A bundle of signals representing a memory read port.
```

src/main/scala/chisel3/util/MemInterface.scala
mwachs5, line 16:
```suggestion
class RDPortInterface[T <: Data](tpe: T, addrWidth: Int) extends Bundle {
```

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,187 @@
...
+  val enable = Input(Bool())
+  val writeValue = Input(tpe)
+}
+
+/** A bundle of signals representing a read/write memory port.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param width The width of the address signal
+  */
+class RWPortInterface[T <: Data](tpe: T, width: Int) extends Bundle {
```
mwachs5, line 38:
```suggestion
class MemRdWrPortInterface[T <: Data](tpe: T, width: Int) extends Bundle {
```

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/** A IO bundle of signals connecting to the ports of a wrapped `SyncReadMem`, as requested by
+  * `SyncReadMem.interface`.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param width The width of the address wires of each port
+  * @param numRD The number of read ports
+  * @param numWR The number of write ports
+  * @param numRW The number of read/write ports
```
mwachs5, line 53:
```suggestion
  * @param numRd The number of read ports
  * @param numWr The number of write ports
  * @param numRdWr The number of read/write ports
```

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,187 @@
...
+  * @tparam tpe The data type of the memory port
+  * @param width The width of the address signal
+  */
+class RDPortInterface[T <: Data](tpe: T, width: Int) extends Bundle {
+  val addr = Input(UInt(width.W))
+  val enable = Input(Bool())
+  val readValue = Output(tpe)
+}
+
+/** A bundle of signals representing a write memory port.
```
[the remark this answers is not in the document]
jared-barocsi: Write masks seem to be a niche case, so I'm not sure if these should support the capability off the bat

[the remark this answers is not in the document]
jackkoenig: Write masks are pretty important. They aren't used in examples that often but they are quite common in actual uses of SyncReadMems as SRAMs in real designs.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+/** A bundle of signals representing a read memory port.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param addrWidth The width of the address signal
+  */
+class MemRdPortInterface[T <: Data](tpe: T, addrWidth: Int) extends Bundle {
+  val addr = Input(UInt(addrWidth.W))
+  val enable = Input(Bool())
+  val readValue = Output(tpe)
+}
```
jackkoenig, line 20: This going to be very bikesheddy but I feel like we should spend the time to make sure we pick APIs we like here. We are doing a lot of mixing of abbreviations and not abbreviations so I think we should try to be self-consistent, at least within categories of things. Also, the word `Interface` is suuuuper overloaded and not particularly useful in this circumstance IMO but others may disagree.

As such, I will propose code changes for my preference on each of these bundles to get this bikeshedding party started.

```suggestion
class MemReadPort[T <: Data](tpe: T, addrWidth: Int) extends Bundle {
  val addr = Input(UInt(addrWidth.W))
  val en = Input(Bool())
  val data = Output(tpe)
}
```

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+  val maskSize = tpe match {
+    case vec: Vec[_] => vec.size
+    case _ => 0
+  }
+
+  val addr = Input(UInt(addrWidth.W))
+  val enable = Input(Bool())
+  val writeValue = Input(tpe)
+  val mask: Option[Vec[Bool]] = if (withMask && maskSize > 0) Some(Input(Vec(maskSize, Bool()))) else None
+}
```
jackkoenig, line 41: Related to discussion above

```suggestion
class MemWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) extends Bundle {
  val addr = Input(UInt(addrWidth.W))
  val enable = Input(Bool())
  val data = Input(tpe)
  val mask: Option[Vec[Bool]] = if (masked) {
    val maskSize = tpe match {
      case vec: Vec[_] => vec.size
      case bad => require(false, s"Write ports can only be masked with Vec data types! Got $bad")
    }
    Some(Input(Vec(maskSize, Bool())))
  } else {
    None
  }
}
```
It's also not useful to have `maskSize` as a field when there is no mask, so I think it's better to hide it as a local variable.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+    case _ => 0
+  }
+
+  val addr = Input(UInt(addrWidth.W))
+  val enable = Input(Bool())
+  val isWrite = Input(Bool())
+  val readValue = Output(tpe)
+  val writeValue = Input(tpe)
+  val mask: Option[Vec[Bool]] = if (withMask && maskSize > 0) Some(Input(Vec(maskSize, Bool()))) else None
+}
```
jackkoenig, line 64:
```suggestion
class MemReadWritePort[T <: Data](tpe: T, addrWidth: Int, mask: Boolean) extends Bundle {

  val addr = Input(UInt(addrWidth.W))
  val en = Input(Bool())
  val wen = Input(Bool())
  val readData = Output(tpe)
  val writeData = Input(tpe)
  val mask: Option[Vec[Bool]] = if (masked) {
    val maskSize = tpe match {
      case vec: Vec[_] => vec.size
      case bad => require(false, s"Readwrite ports can only be masked with Vec data types! Got $bad")
    }
    Some(Input(Vec(maskSize, Bool())))
  } else {
    None
}
```

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+  * @param numRdWr The number of read/write ports
+  */
+class MemInterface[T <: Data](tpe: T, addrWidth: Int, numRd: Int, numWr: Int, numRdWr: Int, withMask: Boolean)
+    extends Bundle {
+  if (withMask) {
+    require(
+      tpe.isInstanceOf[Vec[_]],
+      s"masked writes require that MemInterface is instantiated with a data type of Vec (got $tpe instead)"
+    )
+  }
```
jackkoenig, line 82: I see you have the check here rather than in each port. That is fine, but if we want to do that, we need to make the other constructors package private so that you can only construct them through this type safety.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+  override def typeName: String = s"MemInterface_${MemInterface.portedness(numRd, numWr, numRdWr)}"
+
+  val rd: Vec[MemRdPortInterface[T]] = Vec(numRd, new MemRdPortInterface(tpe, addrWidth))
+  val wr: Vec[MemWrPortInterface[T]] = Vec(numWr, new MemWrPortInterface(tpe, addrWidth, withMask))
+  val rw: Vec[MemRdWrPortInterface[T]] = Vec(numRdWr, new MemRdWrPortInterface(tpe, addrWidth, withMask))
+}
+
+object MemInterface {
```
jackkoenig, line 90: I'm not sure what we should call this, but it's not `MemInterface`. When someone uses these APIs they are creating an actual memory, not just the interface. This is basically an alternative API for `SyncReadMem` so I think we should lean into that, but I'm really not sure what we should call it. `SRAM` may be too specific but may also be the desired behavior.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+    numRdWr: Int,
+    clock:   Clock
+  )(
+    implicit sourceInfo: SourceInfo
+  ): MemInterface[T] = {
+    val addrWidth = log2Up(size + 1)
+
+    val _wrappedMem = Module(new Module {
+      override def desiredName: String =
+        s"SyncReadMemWrapper_${MemInterface.portedness(numRd, numWr, numRdWr)}_${tpe.typeName}"
```
jackkoenig, line 247: Do we really want a level of module hierarchy here? If we do, this Module should at least be a named class, if we're trying to hide it from the user it can be private but I think it's better to avoid anonymous classes for this sort of thing.

I'm less sure if we want the level of hierarchy here. Part of the stated goal of an explicit memory API is that it may ultimately just directly instantiate FIRRTL mems rather than Chirrtl mems. Would we still want the layer of hierarchy in that case? Having this as an anonymous Module rather than a public class suggests to me we may be removing the level of hierarchy, but levels of module hierarchy are the canonical example of [Hyrum's Law](https://www.hyrumslaw.com) in Chisel--users shouldn't rely on the specific module hierarchy but they usually do--especially for memories. So I'm wondering if we should just not have this level of hierarchy here.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+
+      val io = IO(new MemInterface(tpe, addrWidth, numRd, numWr, numRdWr, false))
+
+      val _innerMem = SyncReadMem(size, tpe)
```
jackkoenig, line 251:
```suggestion
      val mem = SyncReadMem(size, tpe)
```
We should definitely name the mem construct, this is an important name.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+
+      for (i <- 0 until numRd) {
+        io.rd(i).readValue := _innerMem.read(io.rd(i).addr, io.rd(i).enable, clock)
```
jackkoenig, line 254: You're passing a clock as an argument but also injecting a level of module hierarchy. The passed clock is not in the right module so this cannot be right and I'm shocked there is no error being thrown. Please figure out why this isn't erroring (because it should be) and then remove this clock argument. If we're going to pass a clock as an argument, you can only use it by driving ports of the wrapping Module, but probably should just be doing a `withClock` around the module instantiation (if we even want the Module).

src/test/scala/chiselTests/Mem.scala
```diff
@@ -447,3 +447,86 @@ class MemorySpec extends ChiselPropSpec {
...
 }
+
+class MemInterfaceSpec extends ChiselFunSpec {
+  describe("MemInterface") {
+    val portCombos: Seq[(Int, Int, Int)] =
+      for {
+        numRD <- 0 until 3;
+        numWR <- 0 until 3;
+        numRW <- 0 until 3 if (numRD + numWR + numRW) > 0
+      } yield (numRD, numWR, numRW)
```
jackkoenig, line 458:
```suggestion
        numRD <- 0 until 3
        numWR <- 0 until 3
        numRW <- 0 until 3
        if (numRD + numWR + numRW) > 0
      } yield (numRD, numWR, numRW)
```

jackkoenig (commented):
The overall code here is really good with a couple of minor issues. My main comments are bikeshedding about names but as this is probably going to become the main SRAM API for a lot of users, I think we should spend the time to get the names right.

jackkoenig:
This also needs to consider https://github.com/chipsalliance/chisel/issues/3278, it should probably be illegal to mix masked and unmasked write ports.

schoeberl:
Comments on names: simply call it ```Memory```. That it is an interface to ```SyncReadMem``` is an implementation detail.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+          io.rw(i).mask.get,
+          io.rw(i).enable,
+          io.rw(i).isWrite,
+          clock
+        )
+      }
+    })
+
+    val _out = Wire(new MemInterface(tpe, addrWidth, numRd, numWr, numRdWr, true))
+    _wrappedMem.io <> _out
```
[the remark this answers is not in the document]
schoeberl: Is it address width or should it be number of elements?

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+  def apply[T <: Data](
+    size:    BigInt,
+    tpe:     T,
+    numRd:   Int,
+    numWr:   Int,
+    numRdWr: Int,
+    clock:   Clock
+  ): MemInterface[T] =
+    macro MemTransform.apply_memInterfaceClk[T]
+
```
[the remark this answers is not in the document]
schoeberl: Do we want to support ports with different clocks? Memories are sometimes used for clock domain crossing? Maybe add this later.

schoeberl:
As a inspiration for names the Altera/Intel on-chip memory documentation, see around page 4-2
https://www.intel.com/programmable/technical-pdfs/654378.pdf

mwachs5:
Yeah this should definitely be called SyncReadMem, and if we want we could do something like `SyncReadMem.interface(...): MemInterface`

One thing I want to agree on is whether we want to distinguish a `MemInterface` and `SyncReadMemInterface` in the `Bundle` subclass types... you could imagine wrapping a combinational mem with the same MemInterface and someone getting confused because the bits are the same but the protocol is different. If we only care about this for now for SyncReadMem let's call the bundles SyncReadMemInterface (or whatever jack suggested) too

src/main/scala/chisel3/util/MemInterface.scala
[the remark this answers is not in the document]
mwachs5: i think this comment can be resolved?

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,254 @@
...
+import scala.language.reflectiveCalls
+import scala.language.experimental.macros
+
+/** A bundle of signals representing a memory read port.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param addrWidth The width of the address signal
+  */
+class MemoryReadPort[T <: Data](tpe: T, addrWidth: Int) extends Bundle {
+  val addr = Input(UInt(addrWidth.W))
```
sequencer, line 17: Personally prefer `address` here.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,254 @@
...
+  * @param masked Whether this read/write port should have an optional mask.
+  *
+  * @note `masked` is only valid if tpe is a Vec; if this is not the case no mask will be initialized
+  *       regardless of the value of `masked`.
+  */
+class MemoryWritePort[T <: Data] private[chisel3] (tpe: T, addrWidth: Int, masked: Boolean) extends Bundle {
+  val addr = Input(UInt(addrWidth.W))
+  val enable = Input(Bool())
+  val data = Input(tpe)
+  val mask: Option[Vec[Bool]] = if (masked) {
```
sequencer, line 35: IMO, chisel type might be too high level for memories. Even for `tpe: T`(I personally don't like it since it cannot be serialized).
For all memories I have encountered in ASIC designs, they are all bit level, so I think dataWidth mask is enough for it.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,254 @@
...
+/** A IO bundle of signals connecting to the ports of a wrapped `SyncReadMem`, as requested by
+  * `MemInterface.apply`.
+  *
+  * @tparam tpe The data type of the memory port
+  * @param width The width of the address wires of each port
+  * @param numRd The number of read ports
+  * @param numWr The number of write ports
+  * @param numRdWr The number of read/write ports
+  */
+class MemInterface[T <: Data](tpe: T, addrWidth: Int, numRd: Int, numWr: Int, numRdWr: Int, masked: Boolean)
```
sequencer, line 81: I'm still wondering, where are clocks.
Basically for each port they should have their own `clock`, otherwise, how do you implement a Async Queue with dual port memory?

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,254 @@
...
+    require(
+      tpe.isInstanceOf[Vec[_]],
+      s"masked writes require that MemInterface is instantiated with a data type of Vec (got $tpe instead)"
+    )
+  }
+  override def typeName: String = s"MemInterface_${MemInterface.portedness(numRd, numWr, numRdWr)}"
+
+  val rd: Vec[MemoryReadPort[T]] = Vec(numRd, new MemoryReadPort(tpe, addrWidth))
+  val wr: Vec[MemoryWritePort[T]] = Vec(numWr, new MemoryWritePort(tpe, addrWidth, masked))
+  val rw: Vec[MemoryReadWritePort[T]] = Vec(numRdWr, new MemoryReadWritePort(tpe, addrWidth, masked))
```
sequencer, line 93: I think besides these ports, chisel should provide an optional port for MBIST and DFT usage.

sequencer (changes requested):
For the memory wrapper, I have these questions:
- what IR it will be lowered to in the future?(hopefully a memory intrinsic).
- why not use a Module to wrap it like #3131 does.
- for DFT flow, how to meet the MBIST needs.

I personally suggest using the 
```
case class SRAMPortInfo(hasRead: Boolean, hasWrite: Boolean, hasMask: Boolean, portDelay: Int = 1)
class SRAMPort(portInfo: SRAMPortInfo, addressWidth: Int, dataWidth: Int) extends Bundle
class SRAM(width: Int, depth: Int, memoryPorts: Seq[SRAMPortInfo], extraIO: Option[Record] = None, contents: Option[os.Path] = None, macroName: Option[String] = None) extends Module 
```
from #3131, since I think:
- unlike register/wires, memory are always hard-marco from IP vendors, post-syn flow need it to be module and instance.
- API in this PR(chisel types for data, `object` user interfaces) is less intuitive than a single `Module` and `Bundle` I suggest using that and let users handle `Data` to `UInt` conversion and byte-level granularity mask.
- a custom port should be provided in the future PR, it's too much suffer for MBIST design experience with Chisel :(

src/main/scala/chisel3/util/MemInterface.scala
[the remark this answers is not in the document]
jackkoenig: We probably should support multiple clocks, but that will need to be specified per port and the API I was commenting on was a single clock for the full memory. We should discuss how to support per-port clocks though.

jackkoenig:
@sequencer these are all great questions, I think we should discuss at Chisel dev on Monday!

core/src/main/scala/chisel3/Mem.scala
```diff
@@ -484,7 +484,7 @@ sealed class SyncReadMem[T <: Data] private[chisel3] (
   )(
     implicit sourceInfo: SourceInfo
   ): T =
-    _readWrite_impl(idx, data, en, isWrite, clock, true)
+    _readWrite_impl(idx, data, en, isWrite, clock, false)
```
jackkoenig, line 487: I'm confused by this change, is this not a change in behavior for regular SyncReadMems? What does this have to do with the rest of this PR?

src/main/scala/chisel3/util/MemInterface.scala
jackkoenig: Regular Chisel type is fine, if you don't want to use it, you can just use `UInt` or `Vec[UInt]` but theres no reason to define it away. Per the memories being bit-level, we can of course just pack the data type into bitvectors in Verilog.

src/main/scala/chisel3/util/MemInterface.scala
jackkoenig: Agreed, treating these ports as future work.

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,337 @@
...
+    })
+
+    val _out = Wire(new MemInterface(tpe, addrWidth, numRd, numWr, numRdWr, true))
+    _wrappedMem.io <> _out
+    _out
+  }
+
+  // Helper util to generate portedness descriptors based on the input parameters
+  // supplied to MemInterface.apply
+  def portedness(rd: Int, wr: Int, rw: Int): String = {
```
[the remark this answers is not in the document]
jackkoenig: @jared-barocsi Is there any reason this method is public?

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,263 @@
...
+    require(
+      tpe.isInstanceOf[Vec[_]],
+      s"masked writes require that SRAMInterface is instantiated with a data type of Vec (got $tpe instead)"
+    )
+  }
+  override def typeName: String = s"SRAMInterface_${SRAM.portedness(numRd, numWr, numRdWr)}"
+
+  val rd: Vec[MemoryReadPort[T]] = Vec(numRd, new MemoryReadPort(tpe, addrWidth))
+  val wr: Vec[MemoryWritePort[T]] = Vec(numWr, new MemoryWritePort(tpe, addrWidth, masked))
+  val rw: Vec[MemoryReadWritePort[T]] = Vec(numRdWr, new MemoryReadWritePort(tpe, addrWidth, masked))
```
jackkoenig, line 96: I still don't like these names since they are extremely abbreviated whereas we are not using abbreviations in the fields of those actual Bundles. My two proposals are:
* `reads`, `writes`, and `readwrites`
* `readPorts`, `writePorts`, and `readwritePorts`.

hcook: > readPorts, writePorts, and readwritePorts.
+1

src/main/scala/chisel3/util/MemInterface.scala
```diff
@@ -0,0 +1,263 @@
...
+object SRAM {
+
+  /** Generates a [[SyncReadMem]] within the current module, connected to an explicit number
+    * of read, write, and read/write ports
+    *
+    * @param size The desired size of the inner `SyncReadMem`
+    * @tparam T The data type of the memory element
+    * @param numRd The number of desired read ports, >= 0
+    * @param numWr The number of desired write ports, >= 0
+    * @param numRdWr The number of desired read/write ports, >= 0
```
hcook, line 108: Same as above, `numReadPorts` etc

core/src/main/scala/chisel3/Mem.scala
jared-barocsi: I forgot to make this its own PR, but the SyncReadMem erroneously raises implicit clock warnings when an explicit clock is being used, and this change fixes that. I'm going to revert it since the warnings don't affect the SRAM itself

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,277 @@
...
+  val mask: Option[Vec[Bool]] = if (masked) {
+    val maskSize = tpe match {
+      case vec: Vec[_] => vec.size
+      case _ => 0
+    }
+    Some(Input(Vec(maskSize, Bool())))
+  } else {
+    None
+  }
+  val clock = Input(Clock())
```
jackkoenig, line 72:
```suggestion
  val address = UInt(addrWidth.W)
  val enable = Bool()
  val isWrite = Bool()
  val readData = Flipped(tpe) // Probably need to check that tpe is passive or otherwise coerce it
  val writeData = tpe // should we .cloneType?
  val mask: Option[Vec[Bool]] = if (masked) {
    val maskSize = tpe match {
      case vec: Vec[_] => vec.size
      case _ => 0
    }
    Some(Vec(maskSize, Bool()))
  } else {
    None
  }
  val clock = Clock()
```

It feels weird to me that give these fields directions rather than treating them more as data types with some flipped fields. Do you have any thoughts on this @azidar? Up above we have `MemoryWritePort` which has all fields going in the same direction marked `Input` which feels like an antipattern to me.

azidar: Using `Input` and `Output` will coerce the data types to be passive, which I think is what we want; the other option is to detect and error, but IMO that will be more confusing than helpful to the user.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,255 @@
...
+    numReadPorts:      Int,
+    numWritePorts:     Int,
+    numReadwritePorts: Int
+  )(
+    implicit evidence: T <:< Vec[_],
+    sourceInfo:        SourceInfo
+  ): SRAMInterface[T] =
+    memInterface_impl(size, tpe)(numReadPorts, numWritePorts, numReadwritePorts, Builder.forcedClock)
+
+  /** @group SourceInfoTransformMacro */
```
azidar, line 155: I don't think you need this comment anymore.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,255 @@
...
+        _out.readwritePorts(i).isWrite,
+        _out.readwritePorts(i).clock
+      )
+      _out.readwritePorts(i).clock :#= clock
+    }
+
+    _out
+  }
+
+  /** @group SourceInfoTransformMacro */
```
azidar, line 197: Same, unnecessary comment.

azidar (commented):
Looks good, but we should update the mdoc to include examples.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,255 @@
...
+    memInterface_impl(size, tpe)(numReadPorts, numWritePorts, numReadwritePorts, Builder.forcedClock)
+
+  /** @group SourceInfoTransformMacro */
+  private def memInterface_impl[T <: Data](
+    size:              BigInt,
+    tpe:               T
+  )(numReadPorts:      Int,
+    numWritePorts:     Int,
+    numReadwritePorts: Int,
+    clock:             Clock
```
sequencer, line 162: I have been thinking: what's this `clock` used for.
I think `clock` should always be bind to port rather than a SRAM instance?

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,255 @@
...
+object SRAM {
+
+  /** Generates a [[SyncReadMem]] within the current module, connected to an explicit number
+    * of read, write, and read/write ports
+    *
+    * @param size The desired size of the inner `SyncReadMem`
+    * @tparam T The data type of the memory element
+    * @param numReadPorts The number of desired read ports, >= 0
+    * @param numWritePorts The number of desired write ports, >= 0
+    * @param numReadwritePorts The number of desired read/write ports, >= 0
```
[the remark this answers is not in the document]
mwachs5: I'm not sure about the constraint part, but could we at least add a test that everything works as expected for those 2 cases?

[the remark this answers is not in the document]
hcook: The write-only case seems pretty bad...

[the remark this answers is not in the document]
aswaterman: I agree that we should probably require (R + RW >= 1) and (W + RW >= 1)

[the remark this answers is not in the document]
sequencer: This API is for SRAM, for read-only case that is supposed to be a ROM, and I think it should be resolved in another API.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,255 @@
...
+  val data = Input(tpe)
+  val mask: Option[Vec[Bool]] = if (masked) {
+    val maskSize = tpe match {
+      case vec: Vec[_] => vec.size
+      case _ => 0
+    }
+    Some(Input(Vec(maskSize, Bool())))
+  } else {
+    None
+  }
```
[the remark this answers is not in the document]
jackkoenig: We can and should drop 2.12 support on `main` at any time, but we may want to backport this PR so we definitely don’t want to use Scala 2.13-only features.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -0,0 +1,248 @@
...
+  * @tparam tpe The data type of the memory port
+  * @param addrWidth The width of the address signal
+  */
+class MemoryReadPort[T <: Data](tpe: T, addrWidth: Int) extends Bundle {
+  val address = Input(UInt(addrWidth.W))
+  val enable = Input(Bool())
+  val data = Output(tpe)
+  val clock = Input(Clock())
+}
+
```
jackkoenig, line 22: I do not think that clocks should be part of the memory port Bundles. Yes, we may want to associate different clocks with different memory ports for the actual memory, but I don't think that putting them in these Bundles is the right API. If a user wants to bulk connect the memory port to a wire or something, they almost certainly are NOT going to want to be dealing with the clock, so I think it's just going to be an API burden in that case.

The API I have in mind (discussed with @aswaterman), is at the point of declaring the SRAM, you can describe the clocks, ie. whereas the normal API is:
```scala
  def apply[T <: Data](
     size:              BigInt,
     tpe:               T,
     numReadPorts:      Int,
     numWritePorts:     Int,
     numReadwritePorts: Int
   )
```
We can add one where instead of a number of ports, you give the clocks you want for each port:
```scala
  def apply[T <: Data](
     size:              BigInt,
     tpe:               T,
     readPortClocks:      Seq[Clock],
     writePortClocks:     Seq[Clock],
     readwritePortClocks: Seq[Clock]
   )
```
We can just have this PR be single clock and add that API in a follow on PR though.

src/main/scala/chisel3/util/SRAM.scala
jackkoenig: I agree we don't need this particular clock, but I have a different opinion on what the clock API should look like: https://github.com/chipsalliance/chisel/pull/3228/files#r1220340613

src/main/scala/chisel3/util/SRAM.scala
sequencer: Sounds great.

docs/src/explanations/memories.md
```diff
@@ -200,3 +200,53 @@ Chisel memories can be initialized from an external `binary` or `hex` file emitt
 
 For more information, check the experimental docs on [Loading Memories](../appendix/experimental-features#loading-memories) feature.
 
+## SRAM
+
+Chisel provides an API to generate `SRAM` memories, which are currently alternative APIs for `SyncReadMem`. The key difference between the `SRAM` and `SyncReadMem` APIs is the former's capability to declare a specific number of read, write, and read-write memory ports, which are interacted with using explicit bundles.
```
jackkoenig, line 205:
```suggestion
Chisel provides an API to generate `SRAM` memories, which are currently alternative APIs for `SyncReadMem`.
The key difference between the `SRAM` and `SyncReadMem` APIs is the former's capability to declare a specific number of read, write, and read-write memory ports, which are interacted with using explicit bundles.
```
As a general style when writing markdown, it's best to do 1 sentence per line--it makes diffs easier to read and merge conflicts easier to resolve.

docs/src/explanations/memories.md
```diff
@@ -200,3 +200,53 @@ Chisel memories can be initialized from an external `binary` or `hex` file emitt
...
+  val width: Int = 8
+
+  val io = IO(new SRAMInterface(1024, UInt(width.W), numReadPorts, numWritePorts, numReadwritePorts))
+
+  // Generate a SyncReadMem representing an SRAM with an explicit number of read, write, and read-write ports
+  io :<>= SRAM(1024, UInt(width.W), numReadPorts, numWritePorts, numReadwritePorts)
+}
+```
+
+To interact with a desired port, index into the `numReadPorts`, `numWritePorts`, and `numReadwritePorts` signals:
```
jackkoenig, line 220:
```suggestion
To interact with a desired port, use the `readPorts`, `writePorts`, and `readwritePorts` fields:
```

docs/src/explanations/memories.md
```diff
@@ -200,3 +200,53 @@ Chisel memories can be initialized from an external `binary` or `hex` file emitt
...
+  memModule.io.readwritePorts(2).enable := true.B
+  memModule.io.readwritePorts(2).isWrite := true.B
+  memModule.io.readwritePorts(2).writeData := 100.U
+
+  // Read:
+  memModule.io.readwritePorts(2).address := 5.U
+  memModule.io.readwritePorts(2).enable := true.B
+  memModule.io.readwritePorts(2).isWrite := false.B
+  val bar = WireInit(UInt(8.W), memModule.io.readwritePorts(2).readData)
+}
```
jackkoenig, line 251: I think this example is reasonable, but I don't see the point in wrapping the call to `SRAM` in a `ModuleWithSRAM`, I would just instantiate it directly.

jackkoenig (commented):
Looks good, would like a couple of changes to the explanation, then let's get this merged!

docs/src/explanations/memories.md
```diff
@@ -200,3 +200,55 @@ Chisel memories can be initialized from an external `binary` or `hex` file emitt
 
 For more information, check the experimental docs on [Loading Memories](../appendix/experimental-features#loading-memories) feature.
 
+## SRAM
+
+Chisel provides an API to generate `SRAM` memories, which are currently alternative APIs for `SyncReadMem`.
```
jackkoenig, line 205:
```suggestion
Chisel provides an API to generate `SRAMs`, an alternative APIs for `SyncReadMem`.
```

jackkoenig (approved):

PR #3237, opened 2023-05-04
Create "tap" API for boring probe-type (non-synthesizable) ports

debs-sifive:
### Contributor Checklist

#### Type of Improvement
- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Create "tap" API for boring probe-type (non-synthesizable) ports.

- [ ] Did you review?


seldridge (commented):
This isn't doing the right thing for the upwards tap case. However, part of this is due to https://github.com/chipsalliance/chisel/issues/3239

Here's a small example:

```scala
import chisel3._

class Bar extends RawModule {
  val a = WireInit(Bool(), DontCare)
  dontTouch(a)
}

class Baz(_a: Bool) extends RawModule {
  val b = WireInit(Bool(), BoringUtils.tapAndRead(_a))
  dontTouch(b)
}

class Foo extends RawModule {
  val bar = Module(new Bar)
  val baz = Module(new Baz(bar.a))
}

println(circt.stage.ChiselStage.emitSystemVerilog(new Foo))
```

This currently produces:

```
circuit Foo :
  module Bar :
    output bore : Probe<UInt<1>>

    wire a : UInt<1>
    bore <= a
    baz.bore <= bore

  module Baz :
    input bore : Probe<UInt<1>>

    wire b : UInt<1>

  module Foo :

    inst bar of Bar
    inst baz of Baz
```

It needs to produce:

```
circuit Foo:
  module Bar :
    output bore : Probe<UInt<1>>

    wire a : UInt<1>
    a is invalid
    define bore = probe(a)

  module Baz :
    input bore : UInt<1>

    wire b : UInt<1>
    b <= bore

  module Foo :

    inst bar of Bar
    inst baz of Baz

    baz.bore <= read(bar.bore)
```

There are a couple of things to note here:

1. The tapping needs to do probes on the way up and real ports on the way down.
2. The behavior of (1) complicates the cleanness of the `tap` and `tapAndRead` APIs. The latter is fine, but I don't know what the former should do. Should it re-probe it to get a probe? That's kind of weird, but at least consistent.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -205,12 +206,13 @@ object BoringUtils {
 
   /** Access a source [[Data]] that may or may not be in the current module.  If
     * this is in a child module, then create ports to allow access the
-    * requrested source.
+    * requested source.
     */
-  def bore(source: Data)(implicit si: SourceInfo): Data = {
+  def bore(source: Data, createProbe: Boolean = false)(implicit si: SourceInfo): Data = {
```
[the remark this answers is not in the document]
mwachs5: I'm not sure why you are saying it has to do:

> The tapping needs to do probes on the way up and real ports on the way down.

Why...? isn't this an option for the lowering, but as far as what we are trying to express we view it as ref ports all the way utnil we say it is safe to read?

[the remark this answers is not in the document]
seldridge: This is how the internal Tap API works today. We never create input probe types as that causes problems for dedup.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -263,11 +265,35 @@ object BoringUtils {
     val (upPath, downPath) = lcaResult.get
     val lcaSource = drill(source, upPath.tail, true)
 
-    /** Creating a wire to assign the result to.  We will return this. */
-    val bore = Wire(purePortType)
-    val sink = drill(bore, downPath.tail, false)
+    if (createProbe) {
+      drill(lcaSource, downPath.tail, false)
```
debs-sifive, line 269: @seldridge does this have the same up-down bug that you were trying to fix? I don't want to create the `bore` wire here, so not sure exactly what I should be drilling here.

seldridge:
This isn't quite right. For the example that I showed, this is producing:

```
circuit Foo :
  module Bar :
    output bore : Probe<UInt<1>>

    wire a : UInt<1>
    ; Below is missing `probe(a)`
    bore <= a

  module Baz :
    input bore : Probe<UInt<1>>

    wire b : UInt<1>
    ; This connection is in the wrong module
    ; There is a missing `read(bore)`.
    bar.bore <= bore

  module Foo :

    inst bar of Bar
    inst baz of Baz
```

You can check for some of these errors by running the result through `firtool` using `ChiselStage.emitFIRRTL`. This will cause `firtool` to try to read the produced CHIRRTL and put it through it's verification/checking. 

This is currently failing as this is creating input probes which are not allowed:

```
circt.stage.phases.Exceptions$FirtoolNonZeroExitCode: firtool returned a non-zero exit code. Note that this version of Chisel (5.0.0-RC1+26-24b98255-SNAPSHOT) was published against firtool version <unknown>.
ExitCode:
1
STDOUT:

STDERR:
<stdin>:10:11: error: input probes not yet supported
    input bore : Probe<UInt<1>>
```

seldridge:
The only thing I'm seeing here is that the `tapAndRead` when doing an up--down is emitting an extra `read(probe(...))` when it could avoid drop the `read(probe(...))`. CIRCT should be canonicalizing this away, but it is not and this creates a second XMR. Otherwise, I think this is doing the right thing. 👍

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -106,13 +106,34 @@ abstract class RawModule extends BaseModule {
...
+          _component.get.asInstanceOf[DefModule].secretConnects += Connect(si, left.lref, ProbeRead(Node(right)))
+        } else {
+          stagedSecretConnects += Connect(si, left.lref, ProbeRead(Node(right)))
+        }
+      case (None, None) =>
+        if (_closed) {
+          _component.get.asInstanceOf[DefModule].secretConnects += Connect(si, left.lref, Node(right))
+        } else {
+          stagedSecretConnects += Connect(si, left.lref, Node(right))
+        }
```
seldridge, line 136: The PR can avoid the duplication with a factoring like:

```scala
val rhs = (left.probeInfo.nonEmpty, right.probeInfo.nonEmpty) match {
  case (true, true)   => ProbeDefine(si, left.lref, Node(right))
  case (true, false)  => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
  case (false, true)  => Connect(si, left.lref, ProbeRead(Node(right)))
  case (false, false) => Connect(si, left.lref, Node(right))
}
val secretConnects = if (_closed) {
  _component.get.asInstanceOf[DefModule].secretConnects
} else {
  stagedSecretConnects
} 
secretConnects += rhs

File: core/src/main/scala/chisel3/internal/firrtl/IR.scala [hunk 4]
```diff
@@ -382,7 +382,7 @@ abstract class Component extends Arg {
 @nowarn("msg=class Port") // delete when Port becomes private
 @deprecated(deprecatedPublicAPIMsg, "Chisel 3.6")
 case class DefModule(id: RawModule, name: String, ports: Seq[Port], commands: Seq[Command]) extends Component {
-  val secretConnects: mutable.ArrayBuffer[Connect] = mutable.ArrayBuffer[Connect]()
+  val secretConnects: mutable.ArrayBuffer[Command] = mutable.ArrayBuffer[Command]()
```
seldridge, line 385: If this is now allowing for arbitrary Chisel IR to be added, doing the name change to `secretIR` or `secretCommands` would be better.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -302,4 +303,108 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      "module Baz :",
+      "input b_bore : UInt<1>",
+      "wire b : UInt<1>",
+      "b <= read(probe(b_bore))",
+      "module BoringUtilsSpec_Anon :",
+      "inst bar of Bar",
+      "inst baz of Baz",
+      "baz.b_bore <= read(bar.b_bore)"
+    )()
+  }
```
seldridge, line 409: These tests look good. Can you also add integration tests in the integration test area?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -302,4 +303,108 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+          val b = Wire(Bool())
+          b := BoringUtils.tapAndRead(_a)
+        }
+
+        val bar = Module(new Bar)
+        val baz = Module(new Baz(bar.a))
+      },
+      Array("--full-stacktrace", "--throw-on-first-error")
+    )
+    println(chirrtl)
```
seldridge, line 393: The printlns can be removed.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -302,4 +303,108 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      "a is invalid ",
+      "define b_bore = probe(a)",
+      "module Baz :",
+      "input b_bore : UInt<1>",
+      "wire b : UInt<1>",
+      "b <= read(probe(b_bore))",
+      "module BoringUtilsSpec_Anon :",
+      "inst bar of Bar",
+      "inst baz of Baz",
+      "baz.b_bore <= read(bar.b_bore)"
```
seldridge, line 407: These can be slimmed down to test the bare minimum. E.g., it's not necessary to test the existence of `wire a` or `a is invalid`. However, it is important to check `baz.b_bore <= read(bar.b_bore)`. Can the tests be changed to only test the connections/probes/defines?

seldridge (commented):
This is looking good. A few comments.

integration-tests/src/test/scala/chiselTest/util/experimental/TapSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+  *       LeftUpper   RightUpper
+  *          /           \
+  *    LeftLower       RightLower
+  * 
+  * Tap a signal in RightLower from LeftLower and check that it contains the
+  * correct value.
+  */
+class TapSpec extends ChiselFlatSpec with Utils {
+  assertTesterPasses(new UpDownTest)
+}
```
seldridge, line 52: Mega-nit: newline at end-of-file missing 🙃

integration-tests/src/test/scala/chiselTest/util/experimental/TapSpec.scala
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
```
seldridge, line 1: This test is looking great. One other idea is to test at every level. I.e., make sure that this thing definitely works. You could do something like:

```scala
object TapAndCheck {
  def apply(tapSignal: UInt) = {
    val readTap = Wire(UInt(8.W))
    readTap := tapAndRead(tapSignal)
    chisel3.assert(readTap === 123.U)
  }
}
```

Then use that in every module to check that it works, including in the same module as the tap. It would then also be good to check one or two levels down, i.e., to add two levels below leftlower and rightlower. This is clearly overkill.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -302,4 +303,170 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      },
+      Array("--full-stacktrace", "--throw-on-first-error")
+    )
+    matchesAndOmits(chirrtl)(
+      "output bore : Probe<UInt<1>>",
+      "output out_bore : Probe<UInt<1>>",
+      "define bore = probe(internalWire)",
+      "define out_bore = probe(internalWire)",
+      "define outProbe = foo.bore",
+      "out <= read(foo.out_bore)"
```
seldridge, line 330: I know I said to tighten this up. However, continuing to include the `module` will make this more readable. This may require not using anonymous modules or also using `suggestName` on the anonymous modules to give them names that will show up in the FIRRTL.

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -302,4 +303,170 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      new RawModule {
+        class Bar(grandParentData: Data) extends RawModule {
+          val out = IO(Bool())
+          out := BoringUtils.tapAndRead(grandParentData)
+        }
+
+        class Foo(parentData: Data) extends RawModule {
+          val out = IO(Bool())
+          val bar = Module(new Bar(parentData))
+          out := bar.out
```
seldridge, line 397: Nit: the IO on the parent are unneeded:
```suggestion
          val bar = Module(new Bar(parentData))
```

seldridge (approved):
Looks good to me. Please clean up the comments and this is ready to go. Great work!

Something to consider for future improvement: currently the boring will generate duplicate wires if the same hardware thing is tapped/bored. It would be interesting to see if there is a way to improve this. Consider the first example. This produces the following FIRRTL:

```
circuit BoringUtilsSpec_Anon :
  module Foo :
    output bore : Probe<UInt<1>>
    output out_bore : Probe<UInt<1>>

    wire internalWire : UInt<1>
    define bore = probe(internalWire)
    define out_bore = probe(internalWire)

  module BoringUtilsSpec_Anon :
    output outProbe : Probe<UInt<1>>
    output out : UInt<1>

    inst foo of Foo
    define outProbe = foo.bore
    out <= read(foo.out_bore)
```

However, the second bore is unnecessary. This could instead be emitted as:
```
circuit BoringUtilsSpec_Anon :
  module Foo :
    output bore : Probe<UInt<1>>

    wire internalWire : UInt<1>
    define bore = probe(internalWire)

  module BoringUtilsSpec_Anon :
    output outProbe : Probe<UInt<1>>
    output out : UInt<1>

    inst foo of Foo
    define outProbe = foo.bore
    out <= read(foo.bore)
```

Can you open a tracking issue about this?

debs-sifive:
@seldridge good point, this affects regular `bore`s as well. I've opened an issue here: https://github.com/chipsalliance/chisel/issues/3275

PR #3238, opened 2023-05-04
[feature] Add properties to Interface

seldridge:
Enable a user to define Scala-time properties on an Interface.  These
properties require a conformance to examine.  This enables patterns where
a client wants to configure itself based on knowledge of the component,
e.g., when the client is a testbench.

#### Release Notes

Add properties to Interface.

### DIFF
--- a/src/test/scala/chiselTests/interface/InterfaceSpec.scala
+++ b/src/test/scala/chiselTests/interface/InterfaceSpec.scala
@@ -24,6 +24,8 @@ class InterfaceSpec extends AnyFunSpec with Matchers {
 
     override type Ports = BarBundle
 
+    override type Properties = Unit
+
     /** Generate the ports given the parameters. */
     override val ports = new Ports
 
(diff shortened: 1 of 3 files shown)


src/test/scala/chiselTests/interface/InterfaceSpec.scala
```diff
@@ -167,6 +173,52 @@ class InterfaceSpec extends AnyFunSpec with Matchers {
       Drivers.link(dir, "compile-0/Foo.sv")
 
     }
+  }
+
+  describe("Basic behavior of properties") {
+
+    // A container of the types of properties that are on this Interface.
+    case class SomeProperties(a: Int, b: String, c: Int => Int)
```
seldridge, line 181: Both `b` and `c` are not used at all here and could be removed. I included them to show that arbitrary things can be properties, including functions.

seldridge:
The original way that I wanted to do stubs is bugged because modules error if they have `Unit` parameters: https://github.com/chipsalliance/chisel/issues/3245

mwachs5 (approved):

PR #3248, opened 2023-05-05
[5.x] Enable MiMa for v5.0.0-RC2

chiselbot:
Enable MiMa for v5.0.0-RC2


jackkoenig (approved):

jackkoenig:
In general we should enable binary compatibility checking as soon as we have a release candidate. We reserve the right to break binary compatibility by the time of the `.0` release, but it's good to check and be intentional about doing so.

PR #3256, opened 2023-05-05
Make release-notes.yml more robust to quotes

jackkoenig:


jackkoenig:
Confirmed that this fixes an issue, see https://github.com/chipsalliance/chisel/actions/runs/4897631745 and it is fixed using this branch in https://github.com/chipsalliance/chisel/actions/runs/4897986613

PR #3262, opened 2023-05-06
[bugfix] Fix BoringUtils up--down bug

### LINKED ISSUE #3239: [bug] BoringUtils Creating Tap in Wrong Module
The new `BoringUtils.bore` API can create U-turn connections in the wrong module. This can happen for cases where there is up--down wiring. Consider the following which you can run with `scala-cli Foo.scala`:

```scala
//> using scala "2.13.10"
//> using lib "org.chipsalliance::chisel::5.0.0-RC1+19-4a5cbfd0-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::5.0.0-RC1+19-4a5cbfd0-SNAPSHOT"
//> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"

import chisel3._

class Bar extends RawModule {
  val a = Wire(Bool())
}

class Baz(_a: Bool) extends RawModule {
  val b = Wire(Bool())
  BoringUtils.bore(_a)
}

class Foo extends RawModule {
  val bar = Module(new Bar)
  val baz = Module(new Baz(bar.a))
}

object Main extends App {
  ChiselStage.emitFIRRTLDialect(new Foo)
}
```

This produces the following FIRRTL:

```
circuit Foo :
  module Bar :
    output bore : UInt<1>

    wire a : UInt<1>
    bore <= a
    baz.bore <= bore ; <-- This is in the wrong module

  module Baz :
    input bore : UInt<1>

    wire b : UInt<1>
    wire bore_1 : UInt<1>
    bore_1 <= bore

  module Foo :

    inst bar of Bar
    inst baz of Baz
```

As commented, `baz.bor <= bore` is in the wrong module and this should be `baz.bore <= bar.bore`

This will error if compiled with CIRCT:

```
Exception in thread "main" circt.stage.phases.Exceptions$FirtoolNonZeroExitCode: firtool returned a non-zero exit code. Note that this version of Chisel (5.0.0-RC1+19-4a5cbfd0-SNAPSHOT) was published against firtool version 1.40.0.
ExitCode:
1
STDOUT:

STDERR:
<stdin>:8:5: error: use of unknown declaration 'baz'
    baz.bore <= bore @[Users/schuylere/repos/github.com/llvm/circt/build/scala-cli/BoringUtilsBugs.scala 17:19]
    ^
```

seldridge:
Fix a bug in the new BoringUtils.bore API where this would generate invalid FIRRTL if the user did boring which created a path that went up to an LCA and then down.  Change the internal leastCommonAncestor utility to also return the LCA module as this avoids a second call to compute the LCA.

Fixes #3239.

Skipping release notes on this as the new BoringUtils API is not in a release.

### DIFF
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -381,10 +381,10 @@ object DataMirror {
     val rightPath = modulePath(right, None)
     rightPath.collectFirst { case p if leftPathSet.contains(p) => p }
   }
-  // Returns LCA paths if a common ancestor exists
+  // Returns LCA paths if a common ancestor exists.  The returned paths includes the LCA.
   private[chisel3] def findLCAPaths(left: HasId, right: HasId): Option[(Seq[BaseModule], Seq[BaseModule])] = {
     leastCommonAncestorModule(left, right).map { lca =>
-      (modulePath(left, Some(lca)), modulePath(right, Some(lca)))
+      (Seq(lca) ++ modulePath(left, Some(lca)), Seq(lca) ++ modulePath(right, Some(lca)))
     }
   }
 }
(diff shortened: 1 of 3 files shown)


azidar:
Sweet, thanks!!!

PR #3263, opened 2023-05-06
[bugfix] Fix naming module with Unit parameter

### LINKED ISSUE #3245: [bug] Unit Arguments to Module Crashes Naming
The following module results in an exception in naming:

```scala
//> using scala "2.13.10"
//> using lib "org.chipsalliance::chisel::5.0.0-RC1"
//> using plugin "org.chipsalliance:::chisel-plugin::5.0.0-RC1"
//> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"

import chisel3._

class Foo(a: Unit) extends RawModule

object Main extends App {
  println(ChiselStage.emitCHIRRTL(new Foo({})))
}
```

This produces:

```
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: begin -1, end 1, length 2
	at ... ()
	at Foo._moduleDefinitionIdentifierProposal(Foo.scala:9)
	at chisel3.experimental.BaseModule._definitionIdentifier(Module.scala:284)
	at chisel3.experimental.BaseModule.<init>(Module.scala:289)
	at chisel3.RawModule.<init>(RawModule.scala:21)
	at Foo.<init>(Foo.scala:9)
	at Main$.$anonfun$new$1(Foo.scala:12)
	at ... ()
	at ... (Stack trace trimmed to user code only. Rerun with --full-stacktrace to see the full stack trace)
```

This is a bit of a weird thing to do, but this comes up if you have a module which is polymorphic over a type and in certain situations that type can be `Unit`.

seldridge:
Fix a bug in naming where a Module that took a Unit parameter could cause naming to crash.  This is, admittedly, a very weird pattern, but can occur if you have a type-parametric module where certain type parameterizations can be Unit.

Fix #3245.

#### Release Notes

Fix bag where a Module with a Unit parameter crashes.

### DIFF
--- a/core/src/main/scala/chisel3/naming/Identifier.scala
+++ b/core/src/main/scala/chisel3/naming/Identifier.scala
@@ -33,7 +33,7 @@ object IdentifierProposer {
         }
       }
     }
-    if (s == "") s
+    if (s == "" || firstOkChar == -1) ""
     else
       s.substring(firstOkChar, if (finalChar == -1) lastOkChar + 1 else finalChar).map { x =>
         if (!legal(x)) '_' else x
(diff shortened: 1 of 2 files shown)


seldridge:
This also fixes other issues where Modules that take String parameters which don't have "legal" characters will also cause crashes. E.g.:

```scala
class Foo(a: String) extends RawModule
circt.stage.ChiselStage.emitCHIRRTL(new Foo("((((((("))
```

It seems reasonable to try to do something more intelligent with the `Module` above than calling it `Foo`. However, this naming is intended to be best effort, so `Foo` is better than not crashing.

PR #3273, opened 2023-05-10
[main] Enable MiMa for v6.0.0-M1

chiselbot:
Enable MiMa for v6.0.0-M1


jackkoenig:
Generally only enable binary compatibility checking once we've released a release candidate.

PR #3276, opened 2023-05-12
Add Top-level parameterized reset type

mwachs5:
### Contributor Checklist

- [TODO] Did you add appropriate documentation in `docs/src`?

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Add a HasParameterizedResetType to mix into Modules for their top level `.reset`

- [ ] Did you review?


core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -126,3 +126,8 @@ trait RequireAsyncReset extends Module {
 trait RequireSyncReset extends Module {
   override private[chisel3] def mkReset: Bool = Bool()
 }
+
+trait HasParameterizedResetType extends Module {
+  val hasAsyncNotSyncReset: Boolean
```
jackkoenig, line 131:
```suggestion
trait HasParameterizedResetType extends Module {
  def hasAsyncNotSyncReset: Boolean
```
virtual methods should be defs

Also, I don't think "HasParameterizedResetType" is the right name for this. HasParameterizedResetType implies to me an abstract reset rather than saying "you must now pick with this val. I think this is a good thing to add, I just would like a different name.

Another option is to just make `mkReset` public (or protected) and then users can pick the reset type directly.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -140,6 +140,17 @@ object Module extends SourceInfoDoc {
...
+/** Allowed values for the types of Module.reset */
+sealed trait ModuleResetType
+/* The default reset type. This is Uninferred, unless it is the top Module, in which case it is Bool */
+case object ModuleResetTypeDefault extends ModuleResetType
+/* Explicitly Uninferred Reset, even if this is the top Module */
+case object ModuleResetTypeUninferred extends ModuleResetType
+/* Explicitly Bool (Synchronous) Reset */
+case object ModuleResetTypeBool extends ModuleResetType
+/* Explicitly Asynchronous Reset */
+case object ModuleResetTypeAsync extends ModuleResetType
```
seldridge, line 152: Style nit: I'm not a big fan of the repeated names. What about:

```scala
object ModuleResetType { // This could just be "ResetType"
  sealed trait Type
  case object Default extends Type
  case object Uninferred extends Type
  case object Synchronous extends Type
}
```

This still gets the full naming, but done via a package, e.g., `ModuleReset.Asynchronous`.

mwachs5: agree, and I'll ask if you'd accept one step further... 

```
object Module {
  object ResetType {
    sealed trait Type
    case object Default extends Type
    ...
  }
}
```

thoughts?

mwachs5: i have implemented my version

src/test/scala/chiselTests/experimental/hierarchy/InstantiateSpec.scala
```diff
@@ -144,6 +145,17 @@ object InstantiateSpec {
...
+@instantiable
+class ParameterizedReset(hasAsyncNotSyncReset: Boolean) extends Module {
+  override def resetType = if (hasAsyncNotSyncReset) Module.ResetType.Asynchronous else Module.ResetType.Synchronous
+  @public val in = IO(Input(UInt(8.W)))
+  @public val out = IO(Output(UInt(8.W)))
+  @public val reset = reset
+  val reg = RegInit(0.U)
+  reg := in
+  out := reg
+}
```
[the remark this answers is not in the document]
mwachs5: yeah, i was eybeballign the verilog in another much noisier version, was also worried that it would constant prop everything away, but i guess Convert does not do that anyway

core/src/main/scala/chisel3/Module.scala
```diff
@@ -138,6 +138,18 @@ object Module extends SourceInfoDoc {
...
+  object ResetType {
+    sealed trait Type
+    /* The default reset type. This is Uninferred, unless it is the top Module, in which case it is Bool */
+    case object Default extends Type
+    /* Explicitly Uninferred Reset, even if this is the top Module */
+    case object Uninferred extends Type
+    /* Explicitly Bool (Synchronous) Reset */
+    case object Synchronous extends Type
+    /* Explicitly Asynchronous Reset */
+    case object Asynchronous extends Type
```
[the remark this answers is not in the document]
mwachs5: oh yeah not sure how I missed that

core/src/main/scala/chisel3/Module.scala
```diff
@@ -138,6 +138,24 @@ object Module extends SourceInfoDoc {
     }
   }
 
+  /** Allowed values for the types of Module.reset */
+  object ResetType {
+
+    /** Allowed values for the types of Module.reset */
+    sealed trait Type
```
jackkoenig, line 145:
```scala
   /** Allowed values for the types of Module.reset */
  sealed trait ResetType
  /** Allowed values for the types of Module.reset */
  object ResetType {

    /** The default reset type. This is Uninferred, unless it is the top Module, in which case it is Bool */
    case object Default extends ResetType
    ...
  }
```

I'm a little late here, but the more canonical way to write this would be having the object be the companion object to tye type.

zhutmost:
Hi. I am writing some components which should generate various circuits for different reset types. So is there any APIs to get the reset type of a Component from its context? 
I guess  the reset type is inferred at the CIRCT stage? (so it is impossible?)

PR #3294, opened 2023-05-23
[bugfix] Don't return Data from BoringUtils APIs

seldridge:
Fix a bug in BoringUtils (and secret IO APIs) where APIs would return a Data.  This could create problems if the BoringUtils APIs were used directly in operations, e.g., the following could cause a problem:

    val a = b & BoringUtils.bore(c)

This requires some casting of DontCare as it is unfortunately not a proper "bottom" type.

No test is added as the test is that Scala passes type checking.

This has no release notes as this is a bugfix for a feature that is in 6.0.

### DIFF
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -205,7 +205,7 @@ object BoringUtils {
     genName
   }
 
-  private def boreOrTap(source: Data, createProbe: Option[ProbeInfo] = None)(implicit si: SourceInfo): Data = {
+  private def boreOrTap[A <: Data](source: A, createProbe: Option[ProbeInfo] = None)(implicit si: SourceInfo): A = {
     import reflect.DataMirror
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -262,7 +262,7 @@ object BoringUtils {
     val lcaResult = DataMirror.findLCAPaths(source, thisModule)
     if (lcaResult.isEmpty) {
       Builder.error(s"Cannot bore from $source to ${thisModule.name}, as they do not share a least common ancestor")
-      return DontCare
+      return DontCare.asInstanceOf[A]
```
jackkoenig, line 265: Is this not a runtime error?

seldridge: This one can be removed. I removed all the `DontCare`'s which are actually not necessary.

jackkoenig (approved):

PR #3337, opened 2023-06-07
Add basic linear temporal logic assertions

fabianschuiki:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/ltl/LTL.scala
+++ b/src/main/scala/chisel3/ltl/LTL.scala
@@ -1,0 +1,391 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.ltl
+
+import chisel3._
+import chisel3.util.circt._
+
+/** An opaque sequence returned by an intrinsic.
+  *
+  * Due to the lack of opaque user-defined types in FIRRTL, the Linear Temporal
+  * Logic (LTL) intrinsics operate on `Bool` values instead of a more desirable
+  * `Sequence` type. To prevent abuse, the Chisel API here wraps these `Bool`s
+  * in an opaque type over which we can provide a safe interface that cannot be
+  * used to construct invalid IR.
+  */
+private case class OpaqueSequence(private[ltl] val inner: Bool) extends Sequence
+
+/** An opaque property returned by an intrinsic.
+  *
+  * See `OpaqueSequence` for details.
+  */
+private case class OpaqueProperty(private[ltl] val inner: Bool) extends Property
+
+/** A single item that may be used in the `Sequence(...)` convenience
+  * constructor for sequences. These atoms may either be `Sequence`s themselves,
+  * like `a` or `a and b`, or a `DelayAtom`, like `Delay`. Together they enable
+  * a shorthand notation for sequences like:
+  *
+  * `Sequence(a, Delay(), b, Delay(2), c, Delay(3, 9), d, Delay(4, None), e)`.
+  */
+sealed trait SequenceAtom
+
+/** A delay atom for the `Sequence(...)` convenience constructor. * */
+private case class DelayAtom(val min: Int, val max: Option[Int]) extends SequenceAtom
+
+/** The delay atoms available to users. Can be interleaved with actual sequences
+  * in `Sequence(...)`. See `SequenceAtom` for details.
+  */
+object Delay {
+  def apply(): SequenceAtom = DelayAtom(1, Some(1))
+  def apply(delay: Int): SequenceAtom = DelayAtom(delay, Some(delay))
+  def apply(min:   Int, max: Int): SequenceAtom = DelayAtom(min, Some(max))
+  def apply(min:   Int, max: Option[Int]): SequenceAtom = DelayAtom(min, max)
+}
+
+/** A Linear Temporal Logic (LTL) sequence. */
+sealed trait Sequence extends Property {
+
+  /** See `Sequence.delay`. */
+  def delay(delay: Int = 1): Sequence = Sequence.delay(this, delay)
+
+  /** See `Sequence.delayRange`. */
+  def delayRange(min: Int, max: Int): Sequence = Sequence.delayRange(this, min, max)
+
+  /** See `Sequence.delayAtLeast`. */
+  def delayAtLeast(delay: Int): Sequence = Sequence.delayAtLeast(this, delay)
+
+  /** See `Sequence.concat`. */
+  def concat(other: Sequence): Sequence = Sequence.concat(this, other)
+
+  /** See `Sequence.and`. */
+  def and(other: Sequence): Sequence = Sequence.and(this, other)
+
+  /** See `Sequence.or`. */
+  def or(other: Sequence): Sequence = Sequence.or(this, other)
+
+  /** See `Sequence.clock`. */
+  override def clock(clock: Clock): Sequence = Sequence.clock(this, clock)
+
+  /** See `Property.implication`. */
+  def implication(prop: Property): Property = Property.implication(this, prop)
+
+  /** See `Property.implication`. */
+  def implicationNonOverlapping(prop: Property): Property = Property.implicationNonOverlapping(this, prop)
+
+  // Convenience operators.
+
+  /** Equivalent to `|->` in SVA. */
+  def |->(prop: Property): Property = this.implication(prop)
+
+  /** Equivalent to `|=>` in SVA. */
+  def |=>(prop: Property): Property = this.implicationNonOverlapping(prop)
+
+  /** Equivalent to `a ## b` in SVA. */
+  def ###(other: Sequence): Sequence = this.concat(other.delay())
+
+  /** Equivalent to `a ##[*] b` in SVA. */
+  def ##*(other: Sequence): Sequence = this.concat(other.delayAtLeast(0))
+
+  /** Equivalent to `a ##[+] b` in SVA. */
+  def ##+(other: Sequence): Sequence = this.concat(other.delayAtLeast(1))
+}
+
+/** Prefix-style utilities to work with sequences.
+  *
+  * This object exposes the primary API to create and compose sequences from
+  * booleans and shorter sequences.
+  */
+object Sequence {
+
+  /** Implicitly wraps a `Bool` and allows it to be used as a sequence or
+    * property. Use via `import chisel3.util.ltl.Sequence.BoolSequence`.
+    */
+  implicit class BoolSequence(val inner: Bool) extends Sequence with SequenceAtom
+
+  /** Delay a sequence by a fixed number of cycles. Equivalent to `##delay` in
+    * SVA.
+    */
+  def delay(seq: Sequence, delay: Int = 1): Sequence = {
+    val ltl_delay = Module(new LTLDelayIntrinsic(delay, Some(0)))
+    ltl_delay.in := seq.inner
+    OpaqueSequence(ltl_delay.out)
+  }
+
+  /** Delay a sequence by a bounded range of cycles. Equivalent to `##[min:max]`
+    * in SVA.
+    */
+  def delayRange(seq: Sequence, min: Int, max: Int): Sequence = {
+    require(min <= max)
+    val ltl_delay = Module(new LTLDelayIntrinsic(min, Some(max - min)))
+    ltl_delay.in := seq.inner
+    OpaqueSequence(ltl_delay.out)
+  }
+
+  /** Delay a sequence by an unbounded range of cycles. Equivalent to
+    * `##[delay:$]` in SVA.
+    */
+  def delayAtLeast(seq: Sequence, delay: Int): Sequence = {
+    val ltl_delay = Module(new LTLDelayIntrinsic(delay, None))
+    ltl_delay.in := seq.inner
+    OpaqueSequence(ltl_delay.out)
+  }
+
+  /** Concatenate multiple sequences. Equivalent to
+    * `arg0 ##0 arg1 ##0 ... ##0 argN` in SVA.
+    */
+  def concat(arg0: Sequence, argN: Sequence*): Sequence = {
+    var lhs = arg0
+    for (rhs <- argN) {
+      val ltl_concat = Module(new LTLConcatIntrinsic)
+      ltl_concat.lhs := lhs.inner
+      ltl_concat.rhs := rhs.inner
+      lhs = OpaqueSequence(ltl_concat.out)
+    }
+    lhs
+  }
+
+  /** Form the conjunction of two sequences. Equivalent to
+    * `arg0 and arg1 and ... and argN` in SVA.
+    */
+  def and(arg0: Sequence, argN: Sequence*): Sequence = {
+    var lhs = arg0
+    for (rhs <- argN) {
+      val ltl_and = Module(new LTLAndIntrinsic)
+      ltl_and.lhs := lhs.inner
+      ltl_and.rhs := rhs.inner
+      lhs = OpaqueSequence(ltl_and.out)
+    }
+    lhs
+  }
+
+  /** Form the disjunction of two sequences. Equivalent to
+    * `arg0 or arg1 or ... or argN` in SVA.
+    */
+  def or(arg0: Sequence, argN: Sequence*): Sequence = {
+    var lhs = arg0
+    for (rhs <- argN) {
+      val ltl_or = Module(new LTLOrIntrinsic)
+      ltl_or.lhs := lhs.inner
+      ltl_or.rhs := rhs.inner
+      lhs = OpaqueSequence(ltl_or.out)
+    }
+    lhs
+  }
+
+  /** Specify a `clock` relative to which all cycle delays within `seq` are
+    * specified. Equivalent to `@(posedge clock) seq` in SVA.
+    */
+  def clock(seq: Sequence, clock: Clock): Sequence = {
+    val ltl_clock = Module(new LTLClockIntrinsic)
+    ltl_clock.in := seq.inner
+    ltl_clock.clock := clock
+    OpaqueSequence(ltl_clock.out)
+  }
+
+  /** Convenience constructor for sequences. Allows for the following syntax:
+    *
+    * `Sequence(a, Delay(), b, Delay(2), c, Delay(3, 9), d, Delay(4, None), e)`
+    *
+    * See `SequenceAtom` for more details.
+    */
+  def apply(atoms: SequenceAtom*): Sequence = {
+    require(atoms.nonEmpty)
+    def needDelayTail = {
+      require(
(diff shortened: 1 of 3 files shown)


src/test/scala/chiselTests/util/LTLSpec.scala
```diff
@@ -0,0 +1,350 @@
...
+import chisel3.testers.BasicTester
+import chisel3.util.ltl._
+import _root_.circt.stage.ChiselStage
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import Sequence._
+
+class LTLSpec extends AnyFlatSpec with Matchers {
```
mwachs5, line 15: can you add a scaladoc this

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+
+package chisel3.util.ltl
+
+import chisel3._
+import chisel3.util.circt._
+import chisel3.experimental.hierarchy.{Instance, Instantiate}
+
+/** An opaque sequence returned by an intrinsic.
+  *
+  * Due to the lack of opaque user-defined types in FIRRTL, the LTL intrinsics
```
mwachs5, line 11:
```suggestion
  * Due to the lack of opaque user-defined types in FIRRTL, the Logic Transfer Level (?) (LTL) intrinsics
```

mwachs5: L... Temporal Logic?

jackkoenig: Linear Temporal Logic

src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
```diff
@@ -0,0 +1,102 @@
...
+
+/** Base class for all binary intrinsics with `lhs`, `rhs`, and `out` ports. */
+@instantiable
+private[util] class BinaryLTLIntrinsic(
+  intrinsicName: String,
+  params:        Map[String, Option[Param]] = Map.empty[String, Option[Param]])
+    extends BaseIntrinsic(f"ltl_$intrinsicName", params) {
+  @public val lhs = IO(Input(Bool()))
+  @public val rhs = IO(Input(Bool()))
+  @public val out = IO(Output(Bool()))
```
jackkoenig, line 47: Generally `lhs` vs. `rhs` is for referring to the two sides of an assignment operator, not the two arguments to a binary function, I would not use that terminology for things like `and`.

src/test/scala/chiselTests/util/LTLSpec.scala
jackkoenig: I don't think we need ScalaDoc on the spec, but definitely on the actual APIs

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+/** The delay atoms available to users. Can be interleaved with actual sequences
+  * in `Sequence(...)`. See `SequenceAtom` for details.
+  */
+object ### extends SequenceAtom {
+  def apply(delay: Int): SequenceAtom = DelayAtom(delay, Some(delay))
+  def apply(min:   Int, max: Int): SequenceAtom = DelayAtom(min, Some(max))
+  def apply(min:   Int, max: Option[Int]): SequenceAtom = DelayAtom(min, max)
+}
+
+/** An LTL sequence. */
```
mwachs5, line 46:
```suggestion
/** A Linear Temporal Logic (LTL) sequence. */
```

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+      case seq: Sequence if atoms.tail.nonEmpty => seq.concat(Sequence(atoms.tail: _*))
+      case seq: Sequence                        => seq
+      case ###                       => needDelayTail.delay(1)
+      case DelayAtom(min, None)      => needDelayTail.delayAtLeast(min)
+      case DelayAtom(min, Some(max)) => needDelayTail.delayRange(min, max)
+    }
+  }
+}
+
+/** An LTL property. */
```
mwachs5, line 211:
```suggestion
/** A Linear Temporal Logic (LTL) property. */
```

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+  * over which we can provide a safe interface that cannot be used to construct
+  * invalid IR.
+  */
+case class OpaqueSequence(private[ltl] val inner: Bool) extends Sequence
+
+/** An opaque property returned by an intrinsic.
+  *
+  * See `OpaqueSequence` for details.
+  */
+case class OpaqueProperty(private[ltl] val inner: Bool) extends Property
```
jackkoenig, line 23: This doc is interesting and all, but do these really need to be part of the public API? These could be private and then it is truly opaque what actual Sequence or Property is returned to the user by various APIs.

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: 👍 Adjusted

src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
fabianschuiki: Yeah good point. These are part of the CIRCT-side interface of the intrinsics though, which I have to stick with here to emit the correct `intmodule`.

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: 👍

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: 👍

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+    val ltl_implication = Instantiate(new LTLImplicationIntrinsic)
+    ltl_implication.lhs := seq.inner
+    ltl_implication.rhs := prop.inner
+    OpaqueProperty(ltl_implication.out)
+  }
+
+  /** Non-overlapping variant of `Property.implication`. Equivalent to
+    * `seq ##1 true |-> prop` and `seq |=> prop` in SVA.
+    */
+  def implication_nonoverlapping(seq: Sequence, prop: Property): Property = {
```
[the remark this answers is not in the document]
fabianschuiki: 👍

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+  * See `OpaqueSequence` for details.
+  */
+private case class OpaqueProperty(private[ltl] val inner: Bool) extends Property
+
+/** A single item that may be used in the `Sequence(...)` convenience
+  * constructor for sequences. These atoms may either be `Sequence`s themselves,
+  * like `a` or `a and b`, or a `DelayAtom`, like `###`. Together they enable a
+  * shorthand notation for sequences like:
+  *
+  * `Sequence(a, ###, b, ###(2), c, ###(3, 9), d, ###(4, None))`.
```
mmaloney-sf, line 30: This syntax is likely to be confusing to anyone the first time they see it.

I thought that `###` was intended to be an infix operator or something when I saw this at first glance. But if it's being used like in this example, what compelling reason is there to use a symbol instead of just calling it `Delay`.

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+  // Convenience operators.
+
+  /** Equivalent to `|->` in SVA. */
+  def |->(prop: Property): Property = this.implication(prop)
+
+  /** Equivalent to `|=>` in SVA. */
+  def |=>(prop: Property): Property = this.implicationNonOverlapping(prop)
+
+  /** Equivalent to `a ## b` in SVA. */
+  def ###(other: Sequence): Sequence = this.concat(other.delay())
```
mmaloney-sf, line 85: With my above comment, this definition also means that `###` is being used in two different ways in the code.

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: The syntax tries to stay close to SVA sequences, where it would read as:
```
a ## b ##2 c ##[3:9] d ###[4:$] e
```
(`e` is missing in the comment)

So `###` actually is an infix operator. It's just that `b ###(2) c` doesn't work in scala. So the closest I could get to providing infix notation is to fake it by accepting sequences and the delay infix operators as individual atoms in the `Sequence(...)` argument list.

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: Conceptually they are both infix operators. The `###` here is a true infix operator in the Scala sense, but it has the delay hardwired to 1, since you can't have additional parameters on the infix operator. For the more complicated case you need `###` in `Sequence(...)`, which technically isn't an infix operator from Scala's point of view, but just an individual object; however the way it's used inside of `Sequence(...)` is still with infix notation, just with `,` sprinkled in.

fabianschuiki:
Thanks for the reviews! Applied the suggested changes 👍

src/main/scala/chisel3/util/LTL.scala
jackkoenig: Another option is to make the pure Scala API just use a named thing, like `Delay`, so it could instead be:
```scala
Sequence(a, Delay(), b, Delay(2), c, Delay(3, 9), d, Delay(4, None))
```
And then use a custom String interpolator to provide a compile-time checked SVA DSL:
```
sva"$a ## $b ##2 $c ##[3:9] $d ###[4:$] e"
```
It may even be possible to elide the `$`, that I'm unsure of, but I know the above can work and even give pretty Scala compile-time errors when used incorrectly.

jackkoenig:
> Thanks for the reviews! Applied the suggested changes 👍
> 
> (Not sure how to fix the failing [Scala-CLI Template](https://github.com/chipsalliance/chisel/actions/runs/5214212858/jobs/9410194243?pr=3337#logs) test 🤔)

I am not sure what the fix is because it's a ScalaDoc-time failure it seems, but you can reproduce it with:
```
sbt "unipublish / publishLocal"
```

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+    ltl_disable.in := prop.inner
+    ltl_disable.condition := cond
+    OpaqueProperty(ltl_disable.out)
+  }
+}
+
+/** The base class for the `AssertProperty`, `AssumeProperty`, and
+  * `CoverProperty` verification constructs.
+  */
+abstract class AssertPropertyLike {
```
jackkoenig, line 337:
```suggestion
sealed abstract class AssertPropertyLike {
```
We don't want users extending this right?

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+      case seq: Sequence                        => seq
+      case ###                       => needDelayTail.delay(1)
+      case DelayAtom(min, None)      => needDelayTail.delayAtLeast(min)
+      case DelayAtom(min, Some(max)) => needDelayTail.delayRange(min, max)
+    }
+  }
+}
+
+/** A Linear Temporal Logic (LTL) property. */
+trait Property {
```
jackkoenig, line 212:
```suggestion
sealed trait Property {
```

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+  * in `Sequence(...)`. See `SequenceAtom` for details.
+  */
+object ### extends SequenceAtom {
+  def apply(delay: Int): SequenceAtom = DelayAtom(delay, Some(delay))
+  def apply(min:   Int, max: Int): SequenceAtom = DelayAtom(min, Some(max))
+  def apply(min:   Int, max: Option[Int]): SequenceAtom = DelayAtom(min, max)
+}
+
+/** A Linear Temporal Logic (LTL) sequence. */
+trait Sequence extends Property {
```
jackkoenig, line 47:
```suggestion
sealed trait Sequence extends Property {
```

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
...
+
+  /** See `Sequence.delay`. */
+  def delay(delay: Int = 1): Sequence = Sequence.delay(this, delay)
+
+  /** See `Sequence.delayRange`. */
+  def delayRange(min: Int, max: Int): Sequence = Sequence.delayRange(this, min, max)
+
+  /** See `Sequence.delayAtLeast`. */
+  def delayAtLeast(delay: Int): Sequence = Sequence.delayAtLeast(this, delay)
+
```
jackkoenig, line 57: Are the values returned here actually valid `Sequences`? `a.delay(2)` doesn't seem like a valid Sequence, I think you need a "that" argument for these to make them valid or they should be returning a different type (maybe `SequenceFragment`?, that might be worse than just requiring another argument to complete the `Sequence`.

src/main/scala/chisel3/util/LTL.scala
```diff
@@ -0,0 +1,392 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.ltl
```
jackkoenig, line 3: This is very bikesheddy but I feel like this package should just be `chisel3.ltl` or perhaps `chisel3.formal`. What do people think @mwachs5, @azidar, @seldridge?

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: Yeah I could totally switch this over to a named thing, and then we can add the string interpolator as a follow-up!

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: Yeah, `a.delay(2)` is a valid sequence. It delays `a` by two cycles, corresponding to `##2 a` in SVA.

fabianschuiki: SVA conflates concatenation and delay. The CIRCT LTL intrinsics separate them properly: concat is a variadic operation that just concatenates sequences, and delay is a unary operation on a single sequence. That's why the API here has them separate, but there is a convenience `###` operator (and the `Sequence(...)`) stuff that offers a combined concat-and-delay as you'd expect to see them in SVA, to make it easier to write sequences.

src/main/scala/chisel3/util/LTL.scala
fabianschuiki: Yeah, great point!

fabianschuiki:
> I am not sure what the fix is because it's a ScalaDoc-time failure it seems, but you can reproduce it with:
> 
> ```
> sbt "unipublish / publishLocal"
> ```

Does the `Instance` API need some sort of compiler plugin that is not enabled during Scala API generation? All errors are related to `Instance[T]` not exposing a port of the underlying `T`. 🤔

src/main/scala/chisel3/util/LTL.scala
jackkoenig: `chisel3.ltl` sounds good to me!

src/main/scala/chisel3/util/LTL.scala
jackkoenig: Makes sense, thanks for the explanation!

fabianschuiki:
I backed out the use of `Instantiate` to work around that ScalaDoc issue. It's ugly and causes a lot of noise in the output, but at least it unblocks this PR.

src/test/scala/chiselTests/LTLSpec.scala
```diff
@@ -0,0 +1,336 @@
...
+      val s2 = c.delayAtLeast(5)
+      val s3 = a ### b
+      val s4 = a ##* b
+      val s5 = a ##+ b
+      require(s0.isInstanceOf[Sequence])
+      require(s1.isInstanceOf[Sequence])
+      require(s2.isInstanceOf[Sequence])
+      require(s3.isInstanceOf[Sequence])
+      require(s4.isInstanceOf[Sequence])
+      require(s5.isInstanceOf[Sequence])
```
jackkoenig, line 62:
```suggestion
      val s0: Sequence = a.delay(1)
      val s1: Sequence = b.delayRange(2, 4)
      val s2: Sequence = c.delayAtLeast(5)
      val s3: Sequence = a ### b
      val s4: Sequence = a ##* b
      val s5: Sequence = a ##+ b
```
It's not really clear to me what these `requires` are trying to check, but if you're just trying to make sure the correct methods are being called, I think you'd be better of just with a static type annotation than the ugly runtime check.

jackkoenig: Same suggestion for all `requires` below.

src/main/scala/chisel3/LTL.scala
```diff
@@ -0,0 +1,391 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.ltl
```
jackkoenig, line 3: The file path should reflect the package, please put this in a new directory called `ltl`

jackkoenig (approved):
LGTM once final comments addressed!

src/test/scala/chiselTests/LTLSpec.scala
fabianschuiki: I was trying to be clever and not influence type inference with a static type annotation :grimacing: I'll drop the `require`s :+1:

fabianschuiki:
Done :+1:

src/main/scala/chisel3/ltl/LTL.scala
```diff
@@ -0,0 +1,391 @@
...
+    val ltl_delay = Module(new LTLDelayIntrinsic(delay, Some(0)))
+    ltl_delay.in := seq.inner
+    OpaqueSequence(ltl_delay.out)
+  }
+
+  /** Delay a sequence by a bounded range of cycles. Equivalent to `##[min:max]`
+    * in SVA.
+    */
+  def delayRange(seq: Sequence, min: Int, max: Int): Sequence = {
+    require(min <= max)
```
jackkoenig, line 119: I'm going to merge the PR anyway but in a future PR, any `require` really needs to have a message, otherwise it errors as just `Requirement Failed`.

PR #3344, opened 2023-06-08
Make it possible to illegalize .asUInt on OpaqueTypes

jackkoenig:
I also split the `OpaqueType` tests out of `RecordSpec` since we would like to make OpaqueTypes "opaquer" which ultimately means making them no longer be Records.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Subclasses of OpaqueType can override `errorOnAsUInt` to make it an elaboration time error if .asUInt is called on an instance of the particular type (including when nested inside of an Aggregate). This closes a large loophole in the OpaqueType API.

- [ ] Did you review?


core/src/main/scala/chisel3/experimental/OpaqueType.scala
```diff
@@ -16,7 +18,28 @@ import chisel3._
...
+      Builder.error(s"${coordinates}${this.typeName} does not support .asUInt.")
+    }
+    super.do_asUInt
+  }
+
+  /** If set to true, calling .asUInt on instances of this type will throw an Exception
+    *
+    * Users can override this to increase the "opacity" of their type.
+    */
+  protected def errorOnAsUInt: Boolean = false
```
mwachs5, line 42: perhaps they could override it with a string message...? so you could say something like

mwachs5 (approved):

core/src/main/scala/chisel3/experimental/OpaqueType.scala
jackkoenig: I think you lost the second half of that sentence.

We could make the String message customizable, I'm curious what API you have in mind. I think the current error message is pretty good, eg. from the tests (admittedly hard to see since I also moved a bunch, but see the end of `OpaqueTypeSpec.scala`): `"Field '_.bits' of type MaybeNoAsUInt does not support .asUInt."`. That `Arg.earlyLocalName` trick is very much a private API and IMO you wouldn't want to lose that "coordinate" information when overriding the error message.

src/test/scala/chiselTests/WidthSpec.scala
```diff
@@ -42,6 +44,30 @@ class WidthSpec extends ChiselFlatSpec {
...
+  they should "result in a 1-bit UInt when calling .asUInt" in {
+    assertInferredWidth(1) {
+      val x = Wire(Vec(0, UInt(8.W))).asUInt
+      WireInit(x)
+    }
+    assertInferredWidth(1) {
+      val x = Wire(new EmptyBundle).asUInt
+      WireInit(x)
+    }
+  }
```
jackkoenig, line 70: My original change resulted in some surprising test failures and exposed some concerning buggy behavior. As shown here, if you call `.asUInt` on an empty Aggregate, you get `0.U` which is `0.U(1.W)`. Fortunately, when these empty Aggregates are part of larger Aggregates, they are flattened out (as zero-width wires), but in the special case of just `.asUInt` on a empty Aggregate, this is what happens.

This is definitely a bug, obviously it should hold that `x.getWidth == x.asUInt.getWidth`, but such width issues are tricky to change. This one derives from calling `SeqUtils.asUInt` which is the same reason why `Cat(Seq())` returns `0.U` (which again is `0.U(1.W)`.

I think it's reasonable to fix this particular bug while leaving the `Cat` and other similar cases alone, but such changes are a little sketchy because it can result in width changes for downstream things and thus silently change behavior. In particular, it's scary if generator code is concatenating the result of `.asUInt` on a `Vec` that can be parameterized all the way down to size 0, or similarly with a Record or Bundle that could be empty.

Thoughts @aswaterman @azidar @seldridge @mwachs5?

jackkoenig: In any case, this PR is not really trying to be concerned with this change and just documents and maintains the existing behavior.

aswaterman: > obviously it should hold that `x.getWidth == x.asUInt.getWidth`

I think this is the overriding concern and fixing it is worth the pain.

src/test/scala/chiselTests/MixedVecSpec.scala
```diff
@@ -62,7 +62,7 @@ class MixedVecZeroEntryTester extends BasicTester {
     val bar = zeroEntryMixedVec
   }
   require(0.U.asTypeOf(bundleWithZeroEntryVec).getWidth == 1)
-  require(bundleWithZeroEntryVec.asUInt.getWidth == 1)
+  require(WireInit(bundleWithZeroEntryVec, DontCare).asUInt.getWidth == 1)
```
mwachs5, line 65: why is this different than the one in Vec.scala? 
```
require(bundleWithZeroEntryVec.asUInt.getWidth == 1)
```

jackkoenig: I changed them several days apart and forgot 🙂. I'll make this one consistent with the one in `Vec.scala`.

jackkoenig: Fixed.

mwachs5 (approved):

PR #3359, opened 2023-06-14
Fix empty aggregate width

jackkoenig:
Split out focused fix from https://github.com/chipsalliance/chisel/pull/3344#discussion_r1227398218

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

#### Release Notes

Previously, calling `.asUInt` on an empty Aggregate would return `0.U(1.W)`, now it returns `0.U(0.W)`. This can affect width propagation and inference.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -82,7 +82,9 @@ sealed abstract class Aggregate extends Data {
   }
 
   override def do_asUInt(implicit sourceInfo: SourceInfo): UInt = {
-    SeqUtils.do_asUInt(flatten.map(_.asUInt))
+    val flattened = flatten.map(_.do_asUInt)
+    // Special-case empty because SeqUtils.do_asUInt returns 0.U (which is 0.U(0.W))
+    if (flattened.isEmpty) 0.U(0.W) else SeqUtils.do_asUInt(flattened)
   }
 
   private[chisel3] override def connectFromBits(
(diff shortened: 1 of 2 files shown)


jackkoenig:
I'm a bit unsure how we should proceed here. This is the minimal fix for the width inconsistency discovered here: https://github.com/chipsalliance/chisel/pull/3344#discussion_r1227398218.

I intentionally did **not** change the behavior in `SeqUtils.asUInt`, but I could. The reason I didn't is because that would also change the behavior of `Cat` for empty Seqs. Arguably we should change it there as well though, because at the moment there are inconsistent semantics there as well, specifically, it should hold that `x == Cat(x.asBools.reverse)`, yet `0.U(0.W) != Cat(0.U(0.W).asBools) == 0.U(1.W)`.

Now I marked this change `API Modification` because this fix is something that can silently change behavior and thus users should take care. I'm wondering if instead we should just bite the bullet and change `0.U` to become `0.U(0.W)` instead of `0.U(1.W)`, but provide some help to users to check that it has not impacted their designs. What I have in mind is a command-line option to restore the old behavior, and users are instructed to generate Verilog with and without the command-line option and diff it. Even better would be if firtool could detect and warn when this change would impact designs, but that is potentially a lot of work to implement.

jackkoenig:
I've marked this `Do Not Merge` and as a draft PR because I think we need to just bite the bullet and change `0.U` to be zero-width, but to do that we need to build a decent warning system to help users migrate.

PR #3364, opened 2023-06-15
SRAM API: Add a parameter to initialize the memory 

jared-barocsi:
This PR adds a new String parameter to `SRAM.apply` which loads the contents of a binary file from the filesystem into the memory (invokes `loadMemoryFromFileInline` on the inner `SyncReadMem`.)
### Contributor Checklist

#### Type of Improvement

- Feature

#### Desired Merge Strategy

- Squash and merge

#### Release Notes

`SRAM.apply` and `SRAM.masked` now take a `contents` parameter, by default a `None`, which is a string path to a binary file on the filesystem which the SRAM should be initialized with.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -117,34 +117,6 @@ object SRAM {
     * @param numReadPorts The number of desired read ports >= 0, and (numReadPorts + numReadwritePorts) > 0
     * @param numWritePorts The number of desired write ports >= 0, and (numWritePorts + numReadwritePorts) > 0
     * @param numReadwritePorts The number of desired read/write ports >= 0, and the above two conditions must hold
-    * @param contents A filesystem path to a binary file to preload this SRAM's contents with
-    *
-    * @return A new `SRAMInterface` wire containing the control signals for each instantiated port
-    * @note This does *not* return the `SyncReadMem` itself, you must interact with it using the returned bundle
-    * @note Read-only memories (R >= 1, W === 0, RW === 0) and write-only memories (R === 0, W >= 1, RW === 0) are not supported by this API, and will result in an error if declared.
-    */
-  def apply[T <: Data](
-    size:              BigInt,
-    tpe:               T,
-    numReadPorts:      Int,
-    numWritePorts:     Int,
-    numReadwritePorts: Int,
-    contents:          Option[String] = None
-  )(
-    implicit sourceInfo: SourceInfo
-  ): SRAMInterface[T] =
-    memInterface_impl(size, tpe)(numReadPorts, numWritePorts, numReadwritePorts, Builder.forcedClock, contents)
-
-  /** Generates a [[SyncReadMem]] within the current module, connected to an explicit number
-    * of read, write, and read/write ports. This SRAM abstraction has both read and write capabilities: that is,
-    * it contains at least one read accessor (a read-only or read-write port), and at least one write accessor
-    * (a write-only or read-write port).
-    *
-    * @param size The desired size of the inner `SyncReadMem`
-    * @tparam T The data type of the memory element
-    * @param numReadPorts The number of desired read ports >= 0, and (numReadPorts + numReadwritePorts) > 0
-    * @param numWritePorts The number of desired write ports >= 0, and (numWritePorts + numReadwritePorts) > 0
-    * @param numReadwritePorts The number of desired read/write ports >= 0, and the above two conditions must hold
     *
     * @return A new `SRAMInterface` wire containing the control signals for each instantiated port
     * @note This does *not* return the `SyncReadMem` itself, you must interact with it using the returned bundle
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -106,6 +107,34 @@ class SRAMInterface[T <: Data](
...
+    * @note This does *not* return the `SyncReadMem` itself, you must interact with it using the returned bundle
+    * @note Read-only memories (R >= 1, W === 0, RW === 0) and write-only memories (R === 0, W >= 1, RW === 0) are not supported by this API, and will result in an error if declared.
+    */
+  def apply[T <: Data](
+    size:              BigInt,
+    tpe:               T,
+    numReadPorts:      Int,
+    numWritePorts:     Int,
+    numReadwritePorts: Int,
+    contents:          Option[String] = None
```
jackkoenig, line 132:
```suggestion
    loadMemoryFile:          String
```

The name `contents` is a bit misleading to me, I think making it more obvious that this is a filename is helpful.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -106,6 +107,34 @@ class SRAMInterface[T <: Data](
...
+    * of read, write, and read/write ports. This SRAM abstraction has both read and write capabilities: that is,
+    * it contains at least one read accessor (a read-only or read-write port), and at least one write accessor
+    * (a write-only or read-write port).
+    *
+    * @param size The desired size of the inner `SyncReadMem`
+    * @tparam T The data type of the memory element
+    * @param numReadPorts The number of desired read ports >= 0, and (numReadPorts + numReadwritePorts) > 0
+    * @param numWritePorts The number of desired write ports >= 0, and (numWritePorts + numReadwritePorts) > 0
+    * @param numReadwritePorts The number of desired read/write ports >= 0, and the above two conditions must hold
+    * @param contents A filesystem path to a binary file to preload this SRAM's contents with
```
jackkoenig, line 120: Given that `loadMemoryFromFileInline` supports both Hex and Binary, why are we only exposing support for binary? I think either we should define a new type to represent the filename and file format, or we should at least add the file format as another argument to this function

jackkoenig: This API doc should also mention that the file path will be emitted into the Verilog (as opposed to being read during elaboration time).

src/main/scala/chisel3/util/SRAM.scala
jackkoenig: Also, this should not have a default value because otherwise it appears to be ambiguous with the version of `apply` that does not have this extra argument. I'm surprised the compiler doesn't complain about it, but in any case, the version where `contents` is not set is just the other version of this method.

jackkoenig (changes requested):
Some comments about the API, also this needs a test, it can just check that the annotation is properly created.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -104,6 +106,24 @@ class SRAMInterface[T <: Data](
     Vec(numReadwritePorts, new MemoryReadWritePort(tpe, addrWidth, masked))
 }
 
+private[chisel3] abstract class MemoryFile(private[chisel3] val fileType: MemoryLoadFileType) {
+  val path: String
+}
```
jackkoenig, line 111:
```suggestion
/** A memory file with which to preload an [[SRAM]]
  *
  * See concrete subclasses [[BinaryMemoryFile]] and [[HexMemoryFile]]
  */
sealed abstract class MemoryFile(private[chisel3] val fileType: MemoryLoadFileType) {
  /** The path to the memory contents file */
  val path: String
}
```

Since this type is part of the public API of SRAM, it should be public.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -205,6 +289,9 @@ object SRAM {
       )
     }
 
+    // Emit Verilog for preloading the memory from a file if requested
+    memoryFile.map { file: MemoryFile => loadMemoryFromFileInline(mem, file.path, file.fileType) }
```
jackkoenig, line 293:
```suggestion
    memoryFile.foreach { file: MemoryFile => loadMemoryFromFileInline(mem, file.path, file.fileType) }
```
Use foreach when return value is `Unit` or we otherwise don't care about the return value

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -261,6 +349,9 @@ object SRAM {
...
```
jackkoenig, line 353:
```suggestion
    memoryFile.foreach { file: MemoryFile => loadMemoryFromFileInline(mem, file.path, file.fileType) }
```
Use foreach when return value is `Unit` or we otherwise don't care about the return value

jackkoenig (commented):
A few tweaks then LGTM

jackkoenig (approved):

PR #3372, opened 2023-06-21
Add MuxCell intrinsics

uenoku:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This commit adds new intrinsics (MuxCell2 and MuxCell4) to provide users a way to directly use 2-to-1 and 4-to-1 MUX cells in synthesis tools (currently synopsys and cadence are supported) from Chisel level. Appropriate pragmas will be annotated in output verilog. FIRRTL implementation is https://github.com/llvm/circt/pull/5428 

- [x] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/circt/Synthesis.scala
+++ b/src/main/scala/chisel3/util/circt/Synthesis.scala
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.circt
+
+import chisel3._
+import chisel3.experimental.IntrinsicModule
+import chisel3.internal.Builder
+
+import circt.Intrinsic
+
+/** A 2-to-1 mux cell intrinsic.
+  */
+private class MuxCell2Intrinsic[T <: Data](gen: T) extends IntrinsicModule("circt_muxcell2") {
+  val sel = IO(Input(UInt(1.W)))
+  val high = IO(Input(gen))
+  val low = IO(Input(gen))
+  val out = IO(Output(gen))
+}
+
+object MuxCell2 {
+
+  /** Creates an intrinsic which is lowered to a 2-to-1 MUX cell in synthesis tools.
+    * @example {{{
+    * v := MuxCell2(sel, high, low)
+    * }}}
+    */
+
+  def apply[T <: Data](sel: UInt, high: T, low: T): T = {
+    val inst = Module(new MuxCell2Intrinsic(chisel3.chiselTypeOf(high)))
+    inst.sel := sel
+    inst.high := high
+    inst.low := low
+    inst.out
+  }
+}
+
+/** A 4-to-1 mux cell intrinsic.
+  */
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/circt/Synthesis.scala
```diff
@@ -0,0 +1,67 @@
...
+  val high = IO(Input(gen))
+  val low = IO(Input(gen))
+  val out = IO(Output(gen))
+}
+
+object MuxCell2 {
+
+  /** Creates an intrinsic which is lowered to a 2-to-1 MUX cell in synthesis tools.
+    * @example {{{
+    * v := MuxCell2(sel, high, low)
```
ekiwi, line 24: It might be interesting to include the Chisel code that would simulate this cell in the documentation. This way a user who knows Chisel could immediately understand the exact behavior.
In this case you could just say that `MuxCell2(a,b,c) === Mux(a,b,c)` for all `a,b,c`.

src/main/scala/chisel3/util/circt/Synthesis.scala
```diff
@@ -0,0 +1,67 @@
...
+    * }}}
+    */
+
+  def apply[T <: Data](sel: UInt, high: T, low: T): T = {
+    val inst = Module(new MuxCell2Intrinsic(chisel3.chiselTypeOf(high)))
```
jackkoenig, line 29: This, and Mux4 below, should check the inputs to make sure they're valid (and the same type), see the 4 lines 37-40 here: https://github.com/chipsalliance/chisel/blob/b0f9fe0f24a43fc7a434167807eee8f5733a65bf/core/src/main/scala/chisel3/Mux.scala#L37

src/main/scala/chisel3/util/circt/Synthesis.scala
```diff
@@ -0,0 +1,67 @@
...
+/** A 2-to-1 mux cell intrinsic.
+  */
+private class MuxCell2Intrinsic[T <: Data](gen: T) extends IntrinsicModule("circt_muxcell2") {
+  val sel = IO(Input(UInt(1.W)))
+  val high = IO(Input(gen))
+  val low = IO(Input(gen))
+  val out = IO(Output(gen))
+}
+
+object MuxCell2 {
```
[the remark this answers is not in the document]
uenoku: Good point, I didn't think that much. I agree that `Mux2Cell` is more standard naming so I'll change (renamed CIRCT side as well https://github.com/llvm/circt/commit/fba3b2f5e78702c049791d11944396c4baf33fec)

src/main/scala/chisel3/util/circt/Synthesis.scala
uenoku: Thanks, do we need to add special handling for `DontCare` as well?

jackkoenig: I don't think so, if it comes up we can add it.

jackkoenig (approved):

PR #3374, opened 2023-06-21
[codegen] Emit literal identifiers for numeric ids

seldridge:
Change the FIRRTL serializer to use literal identifiers if any identifiers begin with a leading digit.  This is a FIRRTL 3.0.0 feature that simplifies parsing.

This enables literal identifier emission anywhere.  However, this code path is only reachable via MixedVec due to mangling of numeric names to add a leading underscore in all other circumstances.  In the future, this commit will enable removing this restriction and switching to literal identifiers when this happens.

No changes are made to annotation targets.  E.g., the literal identifier, "wire `42`" in circuit "Foo" and module "Bar" is still referred to by a local target "~Foo|Bar>42".  This is intentional (for now).  Targets don't have parsing ambiguity like FIRRTL text and the name of this wire is still "42" not "`42`".  It follows that the storage of this name in FIRRTL IR is not changed.  This is the same way that this is handled in CIRCT.

#### Release Notes

Use FIRRTL 3.0.0 emission of literal identifiers when a Chisel name begins with a digit.

#### Notes

This PR is stacked on #3188 (and should be merged after it).

#### Example

```scala
import chisel3._
import chisel3.util.MixedVec

class `9000` extends RawModule {
  val `0` = IO(Input(MixedVec(Bool())))
  val `1` = IO(Output(MixedVec(Bool())))

  val `2` = Wire(MixedVec(Bool()))
  dontTouch(`2`(0))
  `2`(0) := `0`(0)

  `1` := `2`
}

println(ChiselStage.emitCHIRRTL(new `9000`))
println(ChiselStage.emitSystemVerilog(new `9000`))
```

With this commit, this will produce the following FIRRTL and Verilog:

```
FIRRTL version 3.0.0
circuit iw :%[[
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~iw|iw>_2"
  }
]]
  module iw :
    input _0 : { `0` : UInt<1>}
    output _1 : { `0` : UInt<1>}

    wire _2 : { `0` : UInt<1>}
    connect _2.`0`, _0.`0`
    connect _1, _2
```

```verilog
module iw(
  input  _0_0,
  output _1_0
);

  wire _2_0 = _0_0;
  assign _1_0 = _2_0;
endmodule
```

With further changes that go beyond this PR< this should emit as:

```
FIRRTL version 3.0.0
circuit iw :%[[
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~iw|iw>2"
  }
]]
  module iw :
    input `0` : { `0` : UInt<1>}
    output `1` : { `0` : UInt<1>}

    wire `2` : { `0` : UInt<1>}
    connect `2`.`0`, `0`.`0`
    connect `1`,`2`
```

### DIFF
--- a/firrtl/src/main/scala/firrtl/ir/Serializer.scala
+++ b/firrtl/src/main/scala/firrtl/ir/Serializer.scala
@@ -82,15 +82,25 @@ object Serializer {
     case other  => b ++= other.serialize // Handle user-defined nodes
   }
 
+  /** Hash map containing names that were changed due to legalization. */
+  private val legalizedNames = scala.collection.mutable.HashMap.empty[String, String]
+
+  /** Generate a legal FIRRTL name. */
+  private def legalize(name: String): String = name match {
+    // If the name starts with a digit, then escape it with backticks.
+    case _ if name.head.isDigit => legalizedNames.getOrElseUpdate(name, s"`$name`")
+    case _                      => name
+  }
+
   private def s(str: StringLit)(implicit b: StringBuilder, indent: Int): Unit = b ++= str.serialize
 
   private def s(node: Expression)(implicit b: StringBuilder, indent: Int): Unit = node match {
-    case Reference(name, _) => b ++= name
+    case Reference(name, _) => b ++= legalize(name)
     case DoPrim(op, args, consts, _) =>
       b ++= op.toString; b += '('; s(args, ", ", consts.isEmpty); s(consts, ", "); b += ')'
     case UIntLiteral(value, width) =>
       b ++= "UInt"; s(width); b ++= "(0h"; b ++= value.toString(16); b ++= ")"
-    case SubField(expr, name, _)   => s(expr); b += '.'; b ++= name
+    case SubField(expr, name, _)   => s(expr); b += '.'; b ++= legalize(name)
     case SubIndex(expr, value, _)  => s(expr); b += '['; b ++= value.toString; b += ']'
     case SubAccess(expr, index, _) => s(expr); b += '['; s(index); b += ']'
     case Mux(cond, tval, fval, _) =>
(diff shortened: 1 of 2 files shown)


firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -259,14 +271,14 @@ object Serializer {
...
       b ++= "write-latency => "; b ++= writeLatency.toString; newLineAndIndent(1)
-      readers.foreach { r => b ++= "reader => "; b ++= r; newLineAndIndent(1) }
-      writers.foreach { w => b ++= "writer => "; b ++= w; newLineAndIndent(1) }
-      readwriters.foreach { r => b ++= "readwriter => "; b ++= r; newLineAndIndent(1) }
+      readers.foreach { r => b ++= "reader => "; b ++= legalize(r); newLineAndIndent(1) }
+      writers.foreach { w => b ++= "writer => "; b ++= legalize(w); newLineAndIndent(1) }
+      readwriters.foreach { r => b ++= "readwriter => "; b ++= legalize(r); newLineAndIndent(1) }
       b ++= "read-under-write => "; b ++= readUnderWrite.toString
     case Attach(info, exprs) =>
       // exprs should never be empty since the attach statement takes *at least* two signals according to the spec
```
jackkoenig, line 284: What about verification Statements? They have labels which are names in the sense that they can be annotated and they should correspond to labels in the Verilog.

jackkoenig (approved):
Generally LGTM but question about verification statements

firrtl/src/main/scala/firrtl/ir/Serializer.scala
seldridge: Good catch. I had considered this and thought it wasn't necessary. However, it is. I added a commit that makes these literal identifiers and adds checks that it works here: https://github.com/chipsalliance/chisel/pull/3374/commits/9587c6c0925851220d1b4a2f9254fd6849331a4e

PR #3389, opened 2023-06-27
Support literals and DontCare in DataView targets

jackkoenig:
The best example of why this is useful is the test showing some possible uses with `Valid`:
```scala
object ValidExtensions {
  implicit def view[T <: Data] = DataView[T, Valid[T]](
    x => Valid(x.cloneType), // Valid will strip direction with `Output(...)` anyway
    _ -> _.bits,
    (_, v) => true.B -> v.valid
  )
}
```
This enables you to box any `T` as a `Valid[T]` where the `valid` bit is the literal true:
```scala
import ValidExtensions._
  val in = IO(Input(UInt(8.W)))
  val out= IO(Output(Valid(UInt(8.W))))
  out := in.viewAs[Valid[UInt]]
}
```

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/experimental/DataView.scala
+++ b/src/test/scala/chiselTests/experimental/DataView.scala
@@ -62,6 +62,18 @@ object ValidExtensions {
   )
 }
 
+object DontCareDataView {
+  class Foo extends Bundle {
+    val a = UInt(8.W)
+    val b = UInt(8.W)
+  }
+  implicit def view = DataView[UInt, Foo](
+    _ => new Foo,
+    _ -> _.a,
+    (_, f) => DontCare -> f.b
+  )
+}
+
 class DataViewSpec extends ChiselFlatSpec {
 
   behavior.of("DataView")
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/experimental/DataView.scala
```diff
@@ -54,6 +54,26 @@ object FlatDecoupledDataView {
   implicit val view2 = view.invert(_ => new FlatDecoupled)
 }
 
+object ValidExtensions {
+  implicit def view[T <: Data] = DataView[T, Valid[T]](
+    x => Valid(x.cloneType), // Valid will strip direction with `Output(...)` anyway
+    _ -> _.bits,
+    (_, v) => true.B -> v.valid
+  )
+}
```
seldridge, line 63: Should this be a public utility?

seldridge (approved):
Nice feature improvement. The valid view is pretty slick.

src/test/scala/chiselTests/experimental/DataView.scala
jackkoenig: Probably, we need to figure out a general coding style for such methods. I think it should probably be a factory method on `object Valid` rather than making people call `.viewAs[Valid[CorrectType]]`. The problem is that the natural answer would be `Valid.apply`, but that is obviously taken by the factory for making types rather than values. It's sort of the "VecInit" pattern (but obviously that name is not the best), and if we can come up with a decent "standard", I think this would be a great example of it.

PR #3392, opened 2023-06-28
Fix issue where reify would do the same lookup twice

jackkoenig:
This is not a functional bug. The recursive form of reify for Element would use the .topBinding from the wrong argument, so it would recurse an extra time, but it would ultimately converge to the right value.

This is really subtle and I don't think it's even possible to write a test for it (other than by adding a side effect to reify but that would be insane).

It is a very slight, probably unmeasurable, performance improvement.

The way I convinced myself that the problem happened and was fixed was with the following example code:
```scala
import chisel3._

class Foo(val foo: UInt) extends Bundle
class Bar(val bar: UInt) extends Bundle

object Views {
  implicit val foo2bar = DataView[Foo, Bar](f => new Bar(f.foo.cloneType), _.foo -> _.bar)
}
import Views._

  val a = IO(Input(new Foo(UInt(8.W))))
  val z = IO(Output(new Bar(UInt(8.W))))
  z := a.viewAs[Bar]
}

object Main extends App {
  val chirrtl = ChiselStage.emitCHIRRTL(new MyModule)
  println(chirrtl)
}
```

then add `println(s"reify on $elt")` as the first statement in the fixed function you'll see:
```
[info] reify on MyModule.z.bar: IO[UInt<8>]
[info] reify on _$$View$$_.z_view.bar: [UInt<8>]
[info] reify on MyModule.a.foo: IO[UInt<8>]
[info] reify on MyModule.a.foo: IO[UInt<8>]
```
You can see a little double reify there, with this fix you get:
```
[info] reify on MyModule.z.bar: IO[UInt<8>]
[info] reify on _$$View$$_.z_view.bar: [UInt<8>]
[info] reify on MyModule.a.foo: IO[UInt<8>]
```

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/dataview/package.scala
+++ b/core/src/main/scala/chisel3/experimental/dataview/package.scala
@@ -237,7 +237,7 @@ package object dataview {
     */
   @tailrec private[chisel3] def reify(elt: Element, topBinding: TopBinding): Element =
     topBinding match {
-      case ViewBinding(target) => reify(target, elt.topBinding)
+      case ViewBinding(target) => reify(target, target.topBinding)
       case _                   => elt
     }
 


seldridge (approved):
I don't 💯 follow this, but the difference in the test case LGTM. Nice find.

PR #3393, opened 2023-06-28
Avoid emitting 'const const' for const vecs of const elements

trilorez:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Multiple consecutive 'const' modifiers are no longer emitted when emitting a const vector of const elements.

- [x] Did you review?

### DIFF
--- a/firrtl/src/main/scala/firrtl/ir/Serializer.scala
+++ b/firrtl/src/main/scala/firrtl/ir/Serializer.scala
@@ -337,22 +337,23 @@ object Serializer {
     case Field(name, flip, tpe) => s(flip); b ++= legalize(name); b ++= " : "; s(tpe)
   }
 
-  private def s(node: Type)(implicit b: StringBuilder, indent: Int): Unit = node match {
+  private def s(node: Type)(implicit b: StringBuilder, indent: Int): Unit = s(node, false)
+
+  private def s(node: Type, lastEmittedConst: Boolean)(implicit b: StringBuilder, indent: Int): Unit = node match {
     // Types
     case ProbeType(underlying: Type) => b ++= "Probe<"; s(underlying); b += '>'
     case RWProbeType(underlying: Type) => b ++= "RWProbe<"; s(underlying); b += '>'
     case ConstType(underlying: Type) => {
-      // Avoid emitting 'const const' for const vectors of const elements
-      underlying match {
-        case VectorType(ConstType(_), _) =>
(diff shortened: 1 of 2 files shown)


trilorez:
I realize now I need to handle vecs of vecs

seldridge (approved):
LGTM

The PR needs a label for the release generation flow. Could you do two things:

1. Add the label. Probably "Codegen"
2. Populate the "Release Notes" section of the PR template.

Jack wrote some cool automation to use these two things to automatically generate the release notes.

trilorez:
> The PR needs a label for the release generation flow. Could you do two things:
> 
>     1. Add the label. Probably "Codegen"
> 
>     2. Populate the "Release Notes" section of the PR template.

I added the release notes. I don't believe I'm able to add a label (I'm not a member of this repo) unless I'm missing something.

firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -337,15 +337,23 @@ object Serializer {
...
     case ProbeType(underlying: Type) => b ++= "Probe<"; s(underlying); b += '>'
     case RWProbeType(underlying: Type) => b ++= "RWProbe<"; s(underlying); b += '>'
-    case ConstType(underlying: Type) => b ++= "const "; s(underlying)
+    case ConstType(underlying: Type) => {
+      // Avoid emitting multiple consecurive 'const', which can otherwise occur for const vectors of const elements
+      if (!lastEmittedConst) {
+        b ++= "const "
+      }
+      s(underlying, true)(b, indent)
+    }
```
jackkoenig, line 352: I'm of two minds about this solution. I'm a bit concerned because this same issue is solved for directions in the Converter (https://github.com/chipsalliance/chisel/blob/a6728aa2e81fd96824415693d9710165c50f90cb/core/src/main/scala/chisel3/internal/firrtl/Converter.scala#L321), and presumably we don't have it solved for probes. I'm worried that when we fold the Converter and Serializer together, we will lose this fix. I grant there is some elegance to the spot in that it ensures that any FIRRTL IR will be emitted properly whereas the Converter fix doesn't guarantee that. But we don't create FIRRTL IR any other way so I think I prefer the further upstream fix.

Better yet would be if we could pull the fix further up into the actual construction of the Chisel types such that these fixups aren't necessary.

PR #3410, opened 2023-07-06
Add FileCheck for unit testing

jackkoenig:
Draft to see if the process works end-to-end

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/.github/workflows/install-filecheck/action.yml
+++ b/.github/workflows/install-filecheck/action.yml
@@ -1,0 +1,27 @@
+name: Install FileCheck
+
+inputs:
+  version:
+    description: 'version to install'
+    required: false
+    default: 'FileCheck-16.0.6-test'
+
+runs:
+  using: composite
+  steps:
+    - id: cache-filecheck
+      uses: actions/cache@v3
+      with:
+        path: FileCheck
+        key: filecheck-${{ runner.os }}-${{ inputs.version }}
+
+    - shell: bash
+      if: steps.cache-filecheck.outputs.cache-hit != 'true'
+      run: |
+        mkdir -p filecheck/bin
+        wget -q https://github.com/jackkoenig/FileCheck/releases/download/${{ inputs.version }}/FileCheck-linux-x64
+        chmod +x FileCheck-linux-x64
+        mv FileCheck-linux-x64 filecheck/bin/FileCheck
+
+    - shell: bash
+      run: echo "$(pwd)/filecheck/bin" >> $GITHUB_PATH
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,13 @@ jobs:
           version: ${{ inputs.slang }}
       - name: Setup Java
         uses: actions/setup-java@v4
+      - name: Install FileCheck
+        run: |
+          sudo apt update
+          sudo apt install llvm-12-tools
+          echo "/usr/lib/llvm-12/bin" >> $GITHUB_PATH
+      - name: Setup Scala
+        uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
           java-version: ${{ inputs.jvm }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,15 +58,13 @@ jobs:
         uses: ./.github/workflows/build-slang
         with:
           version: ${{ inputs.slang }}
-      - name: Setup Java
-        uses: actions/setup-java@v4
       - name: Install FileCheck
         run: |
           sudo apt update
           sudo apt install llvm-12-tools
           echo "/usr/lib/llvm-12/bin" >> $GITHUB_PATH
-      - name: Setup Scala
-        uses: actions/setup-java@v3
+      - name: Setup Java
+        uses: actions/setup-java@v4
         with:
           distribution: 'adopt'
           java-version: ${{ inputs.jvm }}
--- a/src/test/scala/chiselTests/ChiselSpec.scala
+++ b/src/test/scala/chiselTests/ChiselSpec.scala
@@ -211,6 +211,50 @@ trait ChiselRunners extends Assertions {
   }
 }
 
+trait FileCheck extends BeforeAndAfterEachTestData { this: Suite =>
+  import scala.Console.{withErr, withOut}
+
+  private def sanitize(n: String): String = n.replaceAll(" ", "_").replaceAll("\\W+", "")
+
+  private val testRunDir: os.Path = os.pwd / os.RelPath(BackendCompilationUtilities.TestDirectory)
+  private val suiteDir:   os.Path = testRunDir / sanitize(suiteName)
+  private var checkFile:  Option[os.Path] = None
+
+  override def beforeEach(testData: TestData): Unit = {
+    // TODO check that these are always available
+    val nameDir = suiteDir / sanitize(testData.name)
+    os.makeDir.all(nameDir)
+    checkFile = Some(nameDir / s"${sanitize(testData.text)}.check")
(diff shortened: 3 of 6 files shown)


.github/workflows/install-filecheck/action.yml
```diff
@@ -0,0 +1,27 @@
...
+        path: FileCheck
+        key: filecheck-${{ runner.os }}-${{ inputs.version }}
+
+    - shell: bash
+      if: steps.cache-filecheck.outputs.cache-hit != 'true'
+      run: |
+        mkdir -p filecheck/bin
+        wget -q https://github.com/jackkoenig/FileCheck/releases/download/${{ inputs.version }}/FileCheck-linux-x64
+        chmod +x FileCheck-linux-x64
+        mv FileCheck-linux-x64 filecheck/bin/FileCheck
```
seldridge, line 24: Instead of always assuming Linux, this can use `{{ runner.os }}` to download the right version.

I think you can use:

```
OS=$(echo {{runner.os}} | | awk '{print tolower($0)}')
wget -q https://github.com/jackkoenig/FileCheck/releases/download/${{ inputs.version }}/FileCheck-$OS-x64
```

.github/workflows/test.yml
```diff
@@ -41,6 +41,8 @@ jobs:
         uses: ./.github/workflows/install-circt
         with:
           version: ${{ matrix.circt }}
+      - name: Install FileCheck
+        uses: ./.github/workflows/install-filecheck
```
seldridge, line 45: As a complete alternative, is there some Ubuntu package which provides FileCheck? Is there any way that we can avoid packaging FileCheck ourselves? (Is this a requirement because of Windows developers?)

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -195,6 +195,39 @@ trait ChiselRunners extends Assertions {
...
+  }
+
+  def generateFirrtlAndFileCheck(t: => RawModule, fileCheckArgs: String*)(check: String): Unit = {
+    // Filecheck needs the thing to check in a file
+    os.write.over(checkFile.get, check)
+    val outStream = new ByteArrayOutputStream()
+    val chirrtl = withOut(outStream)(withErr(outStream)(ChiselStage.emitCHIRRTL(t)))
+    val result = outStream.toString + chirrtl
+    val extraArgs = os.Shellable(fileCheckArgs)
+    os.proc("FileCheck", checkFile.get, extraArgs).call(stdin = result)
```
seldridge, line 227: Good thinking to add `fileCheckArgs`/`extraArgs`. 👍

src/test/scala/chiselTests/DecoupledSpec.scala
```diff
@@ -19,21 +19,17 @@ class DecoupledSpec extends ChiselFlatSpec {
...
+      val enq = IO(Flipped(Decoupled(UInt(8.W))))
+      val deq = IO(Decoupled(UInt(8.W)))
+      deq <> enq.map(_ + 1.U)
+    })("""|CHECK: node [[node1:[a-zA-Z0-9_]+]] = add(enq.bits, UInt<1>(0h1)
+          |CHECK: node [[node2:[a-zA-Z0-9_]+]] = tail([[node1]], 1)
+          |CHECK: connect [[result:[a-zA-Z0-9_]+]].bits, [[node2]]
+          |# Check for back-pressure (ready signal is driven in the opposite direction of bits + valid)
+          |CHECK: connect enq.ready, [[result]].ready
+          |CHECK:  connect deq, [[result]]
+          |""".stripMargin)
```
seldridge, line 32: This is sooooooooo much better. 💯

seldridge (approved):
This looks amazing. The tests can really be bulletproof now.

Two thoughts:

1. Is there any way to avoid needing to package FileCheck ourselves?
2. This follows a pattern of FileCheck usage where the checks are all after the code. I tend to use this frequently. However, I will sometimes also write in-line tests. Is there a way that we could support this as well, e.g., something like:

```scala
val fileCheck = new FileChecker()

class Foo extends RawModule {
  fileCheck("CHECK-LABEL: module Foo")
  val a = IO(Input(Bool()))
  fileCheck("CHECK-NEXT: input a: UInt<1>")
}

fileCheck.check(ChiselStage.emitCHIRRTL(new Foo))
```

This can be handled in a follow-on.

src/test/scala/chiselTests/DecoupledSpec.scala
mwachs5: can we add something to the readme before this is merged with the link to primer/docs for those not familiar with file check syntax?

sequencer:
Just wonder if possible to share the cases among firrtl, verilog.

seldridge:
> Just wonder if possible to share the cases among firrtl, verilog.

FileCheck has support for checking multiple prefixes which is commonly used to share across separate invocations of some test run. This may work for sharing between firrtl and verilog, but is unlikely.

.github/workflows/test.yml
jackkoenig: > As a complete alternative, is there some Ubuntu package which provides FileCheck?

Yeah we can just use llvm-15-tools, I'll switch to that for CI

> Is there any way that we can avoid packaging FileCheck ourselves? (Is this a requirement because of Windows developers?)

I'm not sure that there's a strict requirement, but AFAICT there is no project packaging FileCheck in a standalone and platform independent way which is why I created https://github.com/jackkoenig/FileCheck. We can point Windows developers to it if there are any, but no need to use it here! It's also nice for local development on my Mac although I of course can just build FileCheck myself.

src/test/scala/chiselTests/DecoupledSpec.scala
jackkoenig: @mwachs5 I added instructions on how to install and a very brief description of FileCheck and where to look for more information. How much of a tutorial do you think we need to include?

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -195,6 +195,50 @@ trait ChiselRunners extends Assertions {
...
+  }
+
+  /** Elaborate a Module to FIRRTL and check the FIRRTL with FileCheck */
+  def generateFirrtlAndFileCheck(t: => RawModule, fileCheckArgs: String*)(check: String): Unit = {
+    // Filecheck needs the thing to check in a file
+    os.write.over(checkFile.get, check)
+    val result = ChiselStage.emitCHIRRTL(t)
+    val extraArgs = os.Shellable(fileCheckArgs)
+    os.proc("FileCheck", checkFile.get, extraArgs).call(stdin = result)
+  }
```
seldridge, line 227: For this and the other method, just a `t: => String` is probably more useful as I can see wanting to use this to check SystemVerilog emission. It isn't too onerous to have a user do `fileCheck(ChiselStage.emitCHIRRTL(new Foo))`.

seldridge (approved):

src/test/scala/chiselTests/ChiselSpec.scala
jackkoenig: Capturing stdout and stderr is kind of annoying though, maybe I'll provide a 3rd method that just takes a String, and have the other 2 call it?

seldridge: That works!

seldridge:
@jackkoenig: I rebased this and let it merge. I'm running into situations where I'd like to be using this and it would be immensely helpful.

jackkoenig:
@seldridge Thanks! I have had the same thought recently.

PR #3412, opened 2023-07-07
Rework on SerializableModuleGenerator for abstract type.

sequencer:
The reason this API need to be rewritten is because `SerializableModuleGenerator[_, _]` cannot live in a collection(e.g. Seq, Map), since the types of `SerializableModule` and `SerializableModuleParameter` are unbounded.

Basically I replaced the type bound between `SerializableModuleParameter` and `SerializableModule` with concrete inner type: `SerializableModule#SerializableModuleParameter`. And type bound is not required anymore.

But I'm still struggling with the serialization version of `SerializableModuleGenerator[_]`:
```
FIXME: could not find implicit value for parameter e: upickle.default.Reader[Seq[chisel3.experimental.SerializableModuleGenerator[_ <: chisel3.experimental.SerializableModule]]]
```

This is a type gymnastics that I have been struggled for my evening... 

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/SerializableModuleGenerator.scala
+++ b/core/src/main/scala/chisel3/experimental/SerializableModuleGenerator.scala
@@ -1,83 +1,75 @@
 package chisel3.experimental
 
+import ujson.Obj
 import upickle.default._
 
-import scala.reflect.runtime.universe
+import scala.reflect.macros.whitebox
+import scala.reflect.runtime.universe._
+import scala.language.experimental.macros
 
-/** Parameter for SerializableModule, it should be serializable via upickle API.
-  * For more information, please refer to [[https://com-lihaoyi.github.io/upickle/]]
-  */
-trait SerializableModuleParameter
-
-/** Mixin this trait to let chisel auto serialize module, it has these constraints:
-  * 1. Module should not be any inner class of other class, since serializing outer class is impossible.
-  * 2. Module should have and only have one parameter with type `T`:
-  * {{{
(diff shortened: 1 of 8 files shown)


sequencer:
I encountered the deserialization problem in this PR:
`module` and `parameter` are read from json via `ujson` at runtime.

which means we cannot use whichbox(macro) to construct it at compile time.
So the only solution is runtime reflection.

I spent a whole day to find if there are any solution to instantiate a `abstract class` with `type` and `implicit val` in it. But found nothing...

I comes to doubt that there is no possibility for constructing the `SerializableModuleGenerator` at Scala runtime.

:melt:

sequencer:
Maybe use scala plugin to traverse code and find all `SerializableModule` and create a companion concrete class `SerializableGenerator`, so that the scala reflection will have concrete class to construct.

sequencer:
Ok this is still impossible even with the scala compiler. 
Since for Seq[SerModule], elements of which are from runtime, and there seems to be no way to pass the implicit parameter during the runtime...

sequencer:
I wish Chisel could support Scala3 in the future, so that we can have the dependent type to resolve such issues....
We may need to considerswitching to runtime json serdes lib like json4s

sequencer:
@jackkoenig This was the draft PR which I mentioned in the dev meeting.

PR #3414, opened 2023-07-11
Add support for configurable warnings

jackkoenig:
Currently opened as a draft but I would like to get the ball rolling on this.

This is heavily inspired by [configurable warnings in Scala](https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html), but is a little more limited.

#### TODO
- [x] Audit warning numbering (want to backport to 3.6, so make sure warnings existing there are first in order)
- [x] Add warning number to warning messages
- [x] ~Make new warnings more clear that they are deprecations~ I'm going to do this in a follow on PR
- [x] Write a docs page
- [x] Write release notes
- [x] Add support for `--warn-conf-file` which is the same syntax but with `\n` separated pairs in a file rather than `,` separated pairs in a String.
- [x] Add tests mixing `--warn-conf-file` with `--warn-conf`

The commit text gives a pretty good overview:
```
Add support for configurable warnings

Warnings have been assigned unique integers to use as identifiers for
purposes of documentation and configuration.

New CLI option --warn-conf accepts a comma-separated sequence of
<filter>:<action> pairs for customizing warnings.

Supported filters are:
* any          - matches all warnings
* id=<integer> - matches warnings with the integer id
* src=<glob>   - matches warnings where <glob> matches the warning
                 source file

id and src filters can be combined with &.

Supported actions are:
* :s - suppress the warning
* :w - report the warning as a warning (default behavior)
* :e - error on the warning

For example, the new Vec dynamic width matching warning can be
suppressed for source files in src/main/scala while all other warnings
elevated to errors with:
  --warn-conf "id=4&src=src/main/scala/**:s,any:e"

The is also --warn-conf-file which works similarly to --warn-conf except
the <filter>:<action> pairs are newline separated in a file. # is used
for line comments.
```

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Add support for configurable warnings, see https://www.chisel-lang.org/chisel3/docs/explanations/warnings.html

- [x] Did you mark the proper milestone (Bug fix: `3.5.x`, `3.6.x`, or `5.x` depending on impact, API modification or big change: `6.0`)?
- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -308,16 +308,18 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     }
 
     if (length == 0) {
-      Builder.warning(s"Cannot extract from Vec of size 0.")
+      Builder.warning(Warning(WarningID.ExtractFromVecSizeZero, s"Cannot extra from Vec of size 0."))
     } else {
       p.widthOption.foreach { pWidth =>
         val correctWidth = BigInt(length - 1).bitLength
-        def warn(msg: String) =
-          Builder.warning(
-            s"Dynamic index with width $pWidth is too $msg for Vec of size $length (expected index width $correctWidth)."
-          )
-        if (pWidth > correctWidth) warn("wide")
-        else if (pWidth < correctWidth) warn("narrow")
+        def mkMsg(msg: String): String =
+          s"Dynamic index with width $pWidth is too $msg for Vec of size $length (expected index width $correctWidth)."
+
+        if (pWidth > correctWidth) {
+          Builder.warning(Warning(WarningID.DynamicIndexTooWide, mkMsg("wide")))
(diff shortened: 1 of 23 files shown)


jackkoenig:
We could also possibly implement `suppressEnumCastWarning` with this as one of these filters... This does open the door to suppressing warnings locally with something similar to Scala's `@nowarn` although it would apply recursively to a call stack rather than just affect a specific block of Scala code.

jackkoenig:
After chatting offline with @aswaterman and @seldridge, I'm going to move of regular expressions in favor of globs for the filenames and assigned unique ids for each warning.

azidar (approved):
I did a once-over, looks like a very useful feature and I don't see any issues with the implementation.

mwachs5 (approved):
i am basically approving on principle as described, have not dug deep into the code and the tests seem pretty straightforward :)

I appreciate the file input format!

PR #3418, opened 2023-07-12
Extend constants in `probe.force` and `probe.forceInitial` methods.

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Properly extend forced values in `probe.force` and `probe.forceInitial` methods. Error out on unknown widths.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -79,7 +79,7 @@ package object probe extends SourceInfoDoc {
     data.widthOption match {
       case None => Builder.error("Data width unknown.")
       case Some(w) =>
-        if (w > probeWidth) Builder.error(s"Data width $w is larger than $probeWidth.")
+        if (probe.widthOption.exists(w > _)) Builder.error(s"Data width $w is larger than $probeWidth.")
     }
 
     data match {
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -70,8 +70,11 @@ package object probe extends SourceInfoDoc {
 
   /** Pad Data if it supports padding */
   private def padData[T <: Data](data: T, width: Int)(implicit sourceInfo: SourceInfo): T = {
-    if (!data.width.known) Builder.error("Data width unknown")
-    if (data.width.get > width) Builder.error(s"Data width ${data.width.get} is larger than $width.")
+    data.widthOption match {
+      case None => Builder.error("Data width unknown.")
+      case Some(w) =>
+        if (w > width) Builder.error(s"Data width $w is larger than $width.")
+    }
     data match {
       case d: Bits => d.pad(width).asInstanceOf[T]
       case d => d
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -70,9 +70,8 @@ package object probe extends SourceInfoDoc {
 
   /** Pad Data if it supports padding */
   private def padData[T <: Data](data: T, width: Int)(implicit sourceInfo: SourceInfo): T = {
-    if (data.width.get > width) {
-      Builder.error(s"Data width ${data.width.get} is larger than $width.")
-    }
+    if (!data.width.known) Builder.error("Data width unknown")
+    if (data.width.get > width) Builder.error(s"Data width ${data.width.get} is larger than $width.")
     data match {
       case d: Bits => d.pad(width).asInstanceOf[T]
       case d => d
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -68,10 +68,21 @@ package object probe extends SourceInfoDoc {
     }
   }
 
+  /** Pad Data if it supports padding */
+  private def padData[T <: Data](data: T, width: Int)(implicit sourceInfo: SourceInfo): T = {
+    if (data.width.get > width) {
+      Builder.error(s"Data width ${data.width.get} is larger than $width.")
+    }
```
jackkoenig, line 75: What if the width of the data is unknown? Does firtool accept forcing from a Wire that may have inferred width?

If that's illegal, then there needs to be a specific check and error rather than just a `.get`.

debs-sifive: i added in another check for a known width before this check, does that suffice?

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -68,10 +68,20 @@ package object probe extends SourceInfoDoc {
     }
   }
 
+  /** Pad Data if it supports padding */
+  private def padData[T <: Data](data: T, width: Int)(implicit sourceInfo: SourceInfo): T = {
+    if (!data.width.known) Builder.error("Data width unknown")
+    if (data.width.get > width) Builder.error(s"Data width ${data.width.get} is larger than $width.")
```
jackkoenig, line 74:
```suggestion
    data.widthOption match {
      case None => Builder.error("Data width unknown")
      case Some(w) =>
        if (w > width) Builder.error(s"Data width $w is larger than $width.")
    }
```
As written, an unknown width will still result in throwing an exception on that `.get`. That wouldn't be the worst thing in the world but it will immediately kill elaboration whereas these are recoverable errors so it is better to continue elaboration so that we can aggregate any warnings or errors from later elaboration.

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -424,4 +424,34 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
...
+        new RawModule {
+          val a = IO(Output(RWProbe(UInt(2.W))))
+          forceInitial(a, 123.U)
+        },
+        Array("--throw-on-first-error")
+      )
+    }
+    exc.getMessage should be("Data width 7 is larger than 2.")
+  }
+
```
jackkoenig, line 456: It would be good to show a test padding a non-constant, and one showing the error for using a Wire of unknown width.

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -83,7 +93,7 @@ package object probe extends SourceInfoDoc {
   /** Override existing driver of a writable probe. */
   def force(clock: Clock, cond: Bool, probe: Data, value: Data)(implicit sourceInfo: SourceInfo): Unit = {
     requireHasWritableProbeTypeModifier(probe, "Cannot force a non-writable Probe.")
-    pushCommand(ProbeForce(sourceInfo, clock.ref, cond.ref, probe.ref, value.ref))
+    pushCommand(ProbeForce(sourceInfo, clock.ref, cond.ref, probe.ref, padData(value, probe.width.get).ref))
```
jackkoenig, line 96: What happens if the width of a `Probe` itself is unknown?

debs-sifive: Repeating from the conversation on slack, firtool errors out if the probe width does not eventually compute to the force value width. So I have decided for now to just error if either of these widths are unknown, though for future purposes this may be too strict.

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -68,10 +68,30 @@ package object probe extends SourceInfoDoc {
...
+    // check probe width is known
+    requireHasProbeTypeModifier(probe, s"Expected $probe to be a probe.")
+    if (!probe.isWidthKnown) Builder.error("Probe width unknown.")
+    val probeWidth = probe.widthOption.getOrElse(0)
+
+    // check data width is known
+    data.widthOption match {
+      case None => Builder.error("Data width unknown.")
+      case Some(w) =>
+        if (w > probeWidth) Builder.error(s"Data width $w is larger than $probeWidth.")
```
jackkoenig, line 82: If probe width is unknown, but data width is known, there's an extra error being generated that is unnecessary. I think you should tweak this like so:
```suggestion
      case Some(w) =>
        if (probe.widthOption.exists(w > _)) Builder.error(s"Data width $w is larger than $probeWidth.")
```

jackkoenig (approved):
Looks good but there is 1 minor issue where 1 error will cause an extra one to be reported and the 2nd one is wrong, please fix then feel free to merge!

PR #3425, opened 2023-07-14
Initial properties package and Property type.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.properties
+
+import chisel3.{ActualDirection, BaseType, SpecifiedDirection}
+import chisel3.internal.Binding
+import scala.reflect.runtime.universe.{typeOf, TypeTag}
+import scala.annotation.implicitNotFound
+
+/** PropertyType defines a typeclass for valid Property types.
+  *
+  * Typeclass instances will be defined for Scala types that can be used as
+  * properties. This includes builtin Scala types as well as types defined in
+  * Chisel.
+  */
+@implicitNotFound("unsupported Property type ${T}")
+private[chisel3] class PropertyType[T]
+
+/** Companion object for PropertyType.
+  *
+  * Typeclass instances for valid Property types are defined here, so they will
+  * be in the implicit scope and available for users.
+  */
+private[chisel3] object PropertyType {}
+
+/** Property is the base type for all properties.
+  *
+  * Properties are similar to normal Data types in that they can be used in
+  * ports, connected to other properties, etc. However, they are used to
+  * describe a set of non-hardware types, so they have no width, cannot be used
+  * in aggregate Data types, and cannot be connected to Data types.
+  */
+class Property[T: PropertyType: TypeTag] extends BaseType {
+
+  /** A TypeTag for the type T of this Property.
+    */
+  val tpe = typeOf[T]
+
+  /** Bind this node to the in-memory graph.
+    */
+  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
+    this.maybeAddToParentIds(target)
+    binding = target
+    val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
+    direction = ActualDirection.fromSpecified(resolvedDirection)
+  }
+
+  /** Clone type by simply constructing a new Property[T].
+    */
+  override def cloneType: this.type = new Property[T].asInstanceOf[this.type]
+
+  /** Clone type with extra information preserved.
+    *
+    * The only extra information present on a Property type is directionality.
+    */
+  private[chisel3] override def cloneTypeFull: this.type = {
+    val clone = this.cloneType
+    clone.specifiedDirection = specifiedDirection
+    clone
+  }
+}
+
+/** Companion object for Property.
+  */
+object Property {
+
+  /** Create a new Property based on the type T.
+    */
+  def apply[T: PropertyType: TypeTag](): Property[T] = {
+    new Property[T]
+  }
+}
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -34,7 +34,7 @@ class Property[T: PropertyType: TypeTag] extends BaseType {
 
   /** A TypeTag for the type T of this Property.
     */
-  val tpe = typeOf[T]
+  private[chisel3] val tpe = typeOf[T]
 
   /** Bind this node to the in-memory graph.
     */
--- a/core/src/main/scala/chisel3/properties/package.scala
+++ b/core/src/main/scala/chisel3/properties/package.scala
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+/** The properties package includes functionality related to non-hardware data.
+  *
+  * This entire package is currently very experimental, so expect some rough
+  * edges and rapid API evolution.
+  */
+package object properties {}
--- a/src/test/scala/chiselTests/properties/PropertySpec.scala
+++ b/src/test/scala/chiselTests/properties/PropertySpec.scala
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests.properties
+
+import chisel3._
+import chisel3.properties.Property
+import chiselTests.ChiselFlatSpec
+
+class PropertySpec extends ChiselFlatSpec {
+  behavior.of("Property")
+
+  it should "fail to compile with unsupported Property types" in {
+    assertTypeError("""
+      class MyThing
+      val badProp = Property[MyThing]()
+    """)
+  }
+}
--- a/docs/src/explanations/explanations.md
+++ b/docs/src/explanations/explanations.md
@@ -45,4 +45,5 @@ read these documents in the following order:
 * [Source Locators](source-locators)
 * [Warnings](warnings)
 * [Deep Dive into Legacy Connection Operators](connection-operators)
+* [Properties](properties)
 
--- a/docs/src/explanations/properties.md
+++ b/docs/src/explanations/properties.md
@@ -1,0 +1,30 @@
+---
+layout: docs
+title:  "Properties"
+section: "chisel3"
+---
+
+# Properties
+
+Chisel *properties* represent information about the design that is not hardware.
+This is useful to capture domain-specific knowledge and design intent alongside
+the hardware description within the same generator.
+
+## Property Types
+
+The core primitive for using properties is the `Property` type.
+
+`Property` types work similarly to the other Chisel
+[Data Types](../explanations/data-types), but rather than specifying the type of
+values held in state elements or flowing through wires in the circuit,
+properties never flow through or affect the generated hardware. Instead, they
+flow through the hierarchy as ports that can be connected.
+
+What makes `Property` types useful is their ability to express non-hardware
+information that is present in the generated hierarchy, and can be composed to
+create domain-specific data models that are tightly coupled to the design. An
+input port with `Property` type represents a part of the data model that must be
+supplied when its module is instantiated. An output port with `Property` type
+represents a part of the data model that may be accessed when its module is
+instantiated. As the complete design is generated, an arbitrary data model can
+be generated alongside it.
--- a/website/docs/src/main/resources/microsite/data/menu.yml
+++ b/website/docs/src/main/resources/microsite/data/menu.yml
@@ -168,6 +168,10 @@ options:
         url: chisel3/docs/explanations/connection-operators.html
         menu_type: chisel3
         menu_section: deep-dive-into-connection-operators
+      - title: Properties
+        url: chisel3/docs/explanations/properties.html
+        menu_type: chisel3
+        menu_section: properties
   - title: Appendix
     url: chisel3/docs/appendix/appendix.html
     menu_type: chisel3


mikeurbach:
This just sets the stage for properties. I'll send a PR next that will show the first Property types.

Should I add a page under src/docs for properties?

mikeurbach:
I filed a PR against this branch on my fork to show what the first properties additions look like: https://github.com/mikeurbach/chisel3/pull/1

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,79 @@
...
+import scala.reflect.runtime.universe.{typeOf, TypeTag}
+import scala.annotation.implicitNotFound
+
+/** PropertyType defines a typeclass for valid Property types.
+  *
+  * Typeclass instances will be defined for Scala types that can be used as
+  * properties. This includes builtin Scala types as well as types defined in
+  * Chisel.
+  */
+@implicitNotFound("unsupported Property type ${T}")
```
mwachs5, line 18: since i dont see it in your test, does this actually print with a substituted string without the `s"..."`? or does it just print `${T}`?

jackkoenig: It properly substitutes it, see the ScalaDoc for implicitNotFound! https://www.scala-lang.org/api/2.13.11/scala/annotation/implicitNotFound.html

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,79 @@
...
+private[chisel3] object PropertyType {}
+
+/** Property is the base type for all properties.
+  *
+  * Properties are like normal Elements in that they can be used in ports,
+  * connected to other properties, etc. However, they are used to describe a set
+  * of non-hardware types, so they have no width, cannot be used in hardware
+  * aggregates, and cannot be connected to hardware types.
+  */
+class Property[T: PropertyType: TypeTag] extends Element {
```
jackkoenig, line 35: I don't think that `Property` should extend `Element` because other than the use in ports, they are very much not sharing much in common with the other `Elements`, (as evidenced by all of the overridden methods that just throw exceptions). I get that you need to figure out how to make them be allowed in Ports, but I think maybe we need a split hierarchy here, I don't think Properties are `Data`, they are a new thing.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Yeah, I was testing it manually, but I wasn't sure how to test a compile error message in a scalatest.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Thanks, this was a question I had, and your reasoning makes sense to me. I think cordoning these off will be great. The only real functionality I do want from `Data` was `bind` and the nice `:=` syntax, which I guess we will just implement this separately for properties when we get to being able to assign them to outputs, etc. Looking a bit ahead, there will be some other laundry list items to make work with properties, like BoringUtils. But I don't think making them their own hierarchy makes that any harder.

mikeurbach: I am taking a look at what this might look like. It seems like we could almost update `Port` so it's `id` field is `HasId` instead of Data. There are still some things to figure out how to update, but then we could make `Property` extend `HasId`. Or do you think there is a better way to attack this?

mikeurbach: As I pull on this thread, I guess what we want is roughly a trait that extends `HasId`, has a specified direction, and has some of the clonetype machinery. Maybe some other stuff I haven't uncovered. And both Data and Property would extend this trait. Does that seem about right?

core/src/main/scala/chisel3/properties/Property.scala
jackkoenig: >  I wasn't sure how to test a compile error message in a scalatest.

ScalaTest doesn't work very well for that. We should probably set up a separate lit-like flow for checking compiler errors and warnings. I prototyped one a bit a while ago but just don't worry about it here.

core/src/main/scala/chisel3/properties/Property.scala
mwachs5: since you mentioned it, do you expect Connectables to work on Property aggregates? Like propertyIn :<>= propertyOut? It seems like we'd need to figure out if that is really the same thing at all or just something that happens to look the same but is defined on your property type.

By the way, i know you don't want properties inside aggregates in firrtl, but we might consider whether it makes sense to have them inside chisel3.Bundle, and thus doing :<>= on a bundle that contains properties would also assign the properties...

mikeurbach: > just something that happens to look the same but is defined on your property type.

I haven't gotten to aggregates in my prototyping, but i guess in this world that is what it would look like: similar surface syntax to Chisel users, but implemented separately for Properties.

In my original prototype that this PR is split from, when I went with the notion that a Property is Data, it was fairly straightforward to add support for Property elements in MonoConnect, and I guess it would be similar for BulkConnect.

But I think the suggestion that Property is not Data makes sense, and we can set up a convenient user experience that feels like you're connecting Properties like you'd connect Data.

jackkoenig: I think it's easy enough to just implement the methods we want rather than trying to reuse the machinery of Data, largely because the machinery of Data is not implemented very well and has a lot of issues. We have an opportunity here to create similar feeling APIs that work and compose better and then perhaps apply some of those learnings back to Data.

jackkoenig: I *think* `cloneIORecord` is only used for `CloneModuleAsRecord` which I think we just forgot to deprecate as it was sort of a proto-D/I API that we really don't need to support. For purposes of Properties, just ignore it and separately I think we should deprecate it.

mikeurbach:
This now includes https://github.com/chipsalliance/chisel/pull/3429. I'd base this PR on that one for initial review, but this PR is on upstream chisel not my fork. I can open a PR within my fork for early review here if it would be helpful.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,72 @@
...
+  */
+@implicitNotFound("unsupported Property type ${T}")
+private[chisel3] class PropertyType[T]
+
+/** Companion object for PropertyType.
+  *
+  * Typeclass instances for valid Property types are defined here, so they will
+  * be in the implicit scope and available for users.
+  */
+private[chisel3] object PropertyType {}
```
jackkoenig, line 24: I don't know that these can really be package private and yet used in a public interface (on `Property` below), I think we will need to make these public.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Ok, the refactoring to add a new BaseType has merged, so we can resolve this discussion.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: In a later PR, where I actually add some typeclass instances here, it seems to work fine from the unit tests. Maybe I'm missing something--is there something I should test that would require this object to be public?

core/src/main/scala/chisel3/properties/package.scala
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+/** The properties package includes functionality related to non-hardware data.
+  *
+  * This entire package is currently very experimental, so expect some rough
+  * edges and rapid API evolution.
+  */
+package object properties {}
```
mwachs5, line 10: what is this for?

src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -0,0 +1,18 @@
...
+import chisel3.properties.Property
+import chiselTests.ChiselFlatSpec
+
+class PropertySpec extends ChiselFlatSpec {
+  behavior.of("Property")
+
+  it should "fail to compile with unsupported Property types" in {
+    assertTypeError("""
+      class MyThing
+      val badProp = Property[MyThing]()
```
mwachs5, line 15: Can you show the failing case where we try to `propOut := propIn` and you get the error message that its not legal? Such that in a subsequent PR it will be shown to now be legal?

mwachs5: But that same test would show that you can indeed have IO ports of property types?

mikeurbach: Well on this PR `:=` is not even defined, so it will fail for that reason. Two PRs down the road there will be an example with both success and failure for `:=`.

core/src/main/scala/chisel3/properties/package.scala
mikeurbach: I'm not sure if it's needed... I was just adding this boilerplate in case we want to define any functions in here later.

src/test/scala/chiselTests/properties/PropertySpec.scala
mwachs5: ok. i guess i thought this one was slated to land after #3442

mwachs5: but in this PR i should be able to have IO of Input, Output properties? Can that be shown?

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,72 @@
...
+  /** A TypeTag for the type T of this Property.
+    */
+  val tpe = typeOf[T]
+
+  /** Bind this node to the in-memory graph.
+    */
+  private[chisel3] override def bind(target: Binding, parentDirection: SpecifiedDirection): Unit = {
+    this.maybeAddToParentIds(target)
+    binding = target
+    val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
```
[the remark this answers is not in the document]
mikeurbach: Hmm I mainly care about properties having Input/Output for their SpecifiedDirection and ActualDirection. When we get to property aggregates, I'm hoping to avoid having flips and have aggregate properties all point the same direction.

src/test/scala/chiselTests/properties/PropertySpec.scala
mikeurbach: > i guess i thought this one was slated to land after

These are currently independent, there will be a future change to actually make properties connectable.

> i should be able to have IO of Input, Output properties? Can that be shown?

Not in the current PR, since no actual types are being defined to work with Property (there are no typeclass instances for PropertyType). I'm trying to just stage these in small pieces with this piece being the initial boilerplate. In the next PR, you will be able to have an IO of Input or Output properties. If it would help, I can make draft PRs for the next two changes.

mwachs5: ok, its fine. sounds like you are being cognizant of having the tests come in with the right change

mikeurbach: yep trying to break this up into small pieces with tests for what is supported in each piece. This PR is mostly boilerplate, but we can at least test one error.

core/src/main/scala/chisel3/properties/Property.scala
jackkoenig: Perhaps it's changed in the last few days, but I don't see any unit tests calling this, only 1 that checks that it fails to compile. I think if you actually try to use this code from outside of the `chisel3` package, you will get compile errors.

mikeurbach: This draft PR on my fork is no longer in sync, but I think it indicates what I mean: https://github.com/mikeurbach/chisel3/pull/1/files#diff-4b64ebc257d840c404c27e41af276a1ae760940bdf80acf7f0cac225100fc944R20

There are tests in the chiselTests package that successfully create properties, and PropertyType is still package private. Maybe I'm missing something, but it seems like just `Property` needs to be public, and Scala implicit resolution did the rest.

mikeurbach:
@jackkoenig @mwachs5 any other feedback on this initial PR? I have been using it in later patches in this series and it's working for me. For a preview, here's a draft PR of the next patch, based on this one: https://github.com/mikeurbach/chisel3/pull/1

mikeurbach:
I just force pushed this--no changes, but rebasing on main. That makes it easier for me to stack up the next three PRs, now that the required refactors are in main.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,72 @@
...
+  * Typeclass instances will be defined for Scala types that can be used as
+  * properties. This includes builtin Scala types as well as types defined in
+  * Chisel.
+  */
+@implicitNotFound("unsupported Property type ${T}")
+private[chisel3] class PropertyType[T]
+
+/** Companion object for PropertyType.
+  *
+  * Typeclass instances for valid Property types are defined here, so they will
```
trilorez, line 21: Will this approach be flexible enough for user-defined struct properties? It seems like everything needs to be declared here to be supported. 
I suppose if the plan is to have some sort of property aggregate base class this would work. Or just only supporting a Scala dictionary for that use case.

mikeurbach: Yeah it's a good question. I haven't thought about user-defined struct properties yet. We will have some kinds of aggregates, like a Scala Map as a property. Fairly soon, we will also add class/object to Chisel, which is sort of like a Module but only with Property typed ports. And an instance of a class (an object) will itself have a property type. So for user-defined "structs", I think we can support them that way.

In general, I think we will have to have a known set of types that can be used with Properties, if we want to take care of defining how they serialize to FIRRTL. So hopefully the above is enough extensibility, and we can still keep the set of possible Property types a sealed thing.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,72 @@
...
+  * Properties are similar to normal Data types in that they can be used in
+  * ports, connected to other properties, etc. However, they are used to
+  * describe a set of non-hardware types, so they have no width, cannot be used
+  * in aggregate Data types, and cannot be connected to Data types.
+  */
+class Property[T: PropertyType: TypeTag] extends BaseType {
+
+  /** A TypeTag for the type T of this Property.
+    */
+  val tpe = typeOf[T]
```
jackkoenig, line 37: Is this something we want to be public or should it be package private?

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  /** Clone type with extra information preserved.
+    *
+    * The only extra information present on a Property type is directionality.
+    */
+  private[chisel3] override def cloneTypeFull: this.type = {
+    val clone = this.cloneType
+    clone.specifiedDirection = specifiedDirection
+    clone
+  }
```
jackkoenig, line 60: If we go the more type-classy route, I'm wondering if `.cloneType` is something we're actually going to need. I know it's needed for now so not really a comment on this PR, but something we should keep in mind as we continue to flesh this API out.

jackkoenig (approved):
I feel like the next PR is going to be the more important one so let's just get this one in and continue fleshing this API out

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Hmm good point, package private is probably sufficient. In the next PR we will see how it's used internally... we might not even want this in the end.

mikeurbach: Made this package private for now, and we can asses if we even want this on the next PR.

mikeurbach:
Thanks @jackkoenig !

PR #3434, opened 2023-07-20
Add FirtoolBinaryPath option

GeorgeLyon:
In certain environments, such as when running tests via `bloop`, it may be difficult to influence which `firtool` CIRCT uses via the PATH environment variable (the only method currently available). This allows us to replace the version of `firtool` that is used at runtime (or point Chisel to the correct firtool binary in scenarios where it isn't on your PATH).

This will also be helpful if we want to package firtool in the Chisel jar @jackkoenig, since we will probably be extracting it to a temporary directory and not in the PATH.

#### Type of Improvement

- Feature (or new API)

#### Release Notes

- Add `FirtoolBinaryPathOption` to select a different `firtool` binary at runtime.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/circt/stage/Annotations.scala
+++ b/src/main/scala/circt/stage/Annotations.scala
@@ -103,15 +103,15 @@ case class EmittedMLIR(
 
 case object FirtoolBinaryPath extends HasShellOptions {
   override def options = Seq(
-    new ShellOption[Unit](
+    new ShellOption(
       longOption = "firtool-binary-path",
-      toAnnotationSeq = path => Seq(FirtoolBinaryPath(path)),
-      helpText =
-        """Specifies the path to the "firtool" binary Chisel should use.""",
-      helpValueName = "path"
+      toAnnotationSeq = (path: String) => Seq(FirtoolBinaryPath(path)),
+      helpText = """Specifies the path to the "firtool" binary Chisel should use.""",
+      helpValueName = Some("path")
     )
   )
 }
+
 /** Annotation that tells [[circt.stage.phases.CIRCT CIRCT]] what firtool executable to use */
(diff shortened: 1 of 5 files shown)


src/main/scala/circt/stage/Annotations.scala
```diff
@@ -101,6 +101,8 @@ case class EmittedMLIR(
 
 }
 
+case class FirtoolBinaryPath(option: String) extends NoTargetAnnotation with CIRCTOption
```
jackkoenig, line 104: This should probably have a ScalaDoc as well as a shell option form

GeorgeLyon: I added a scaladoc... couldn't quite figure out how to add a shell option but if you are invoking from a shell you can always do `PATH=path/to/firtool:$PATH <command>` so not sure if it is 100% necessary.

GeorgeLyon:
@jackkoenig I just updated this to add a `HasShellOptions` case object for the FirtoolBinaryPath annotation. I'm not sure I did this right as I'm not sure how to test, and it seems `HasShellOptions` uses that registeredlibrary magic and isn't actually connected to the annotation case class, but declared on a separate object that then creates the appropriate annotation.

Is this correct?

src/main/scala/circt/stage/Annotations.scala
```diff
@@ -101,6 +101,20 @@ case class EmittedMLIR(
 
 }
 
+case object FirtoolBinaryPath extends HasShellOptions {
```
seldridge, line 104: Nit:

```suggestion
private[stage] case object FirtoolBinaryPath extends HasShellOptions {
```

jackkoenig (approved):

PR #3436, opened 2023-07-24
update ChiselPlugin to name tuples

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

The chisel plugin will now name wires that are part of tuples e.g. `val x = (Wire(Bool()), Wire(Bool()))` will generate wires with names `x_1` and `x_2`

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -768,6 +768,12 @@ private[chisel3] object Builder extends LazyLogging {
       for ((elt, i) <- iter.zipWithIndex) {
         nameRecursively(s"${prefix}_${i}", elt, namer)
       }
+    case product: Product =>
+      product.productIterator.zipWithIndex.foreach {
+        case (elt, i) =>
+          // tuple field names are 1-based (_1, _2, _3, ...)
+          nameRecursively(s"${prefix}_${i + 1}", elt, namer)
+      }
     case _ => // Do nothing
   }
 
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -772,7 +772,6 @@ private[chisel3] object Builder extends LazyLogging {
       product.productIterator.zip(product.productElementNames).foreach {
         case (elt, fullName) =>
           val name = fullName.stripPrefix("_")
-          // tuple field names are 1-based (_1, _2, _3, ...)
           nameRecursively(s"${prefix}_${name}", elt, namer)
       }
     case _ => // Do nothing
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -769,10 +769,11 @@ private[chisel3] object Builder extends LazyLogging {
         nameRecursively(s"${prefix}_${i}", elt, namer)
       }
     case product: Product =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -768,6 +768,12 @@ private[chisel3] object Builder extends LazyLogging {
       for ((elt, i) <- iter.zipWithIndex) {
         nameRecursively(s"${prefix}_${i}", elt, namer)
       }
+    case product: Product =>
+      product.productIterator.zipWithIndex.foreach {
```
jackkoenig, line 772: Rather than using `.zipWithIndex`, I think it would be safer to `.zip(product.productElementNames)`, then this will do sensible things for non-Tuple products (eg. every case class)

src/test/scala/chiselTests/naming/NamePluginSpec.scala
```diff
@@ -375,4 +375,37 @@ class NamePluginSpec extends ChiselFlatSpec with Utils {
...
+
+  "tuples containing non-Data" should "be named" in {
+    class Test extends Module {
+      val x = (Wire(UInt(3.W)), "foobar", Wire(UInt(3.W)))
+    }
+
+    aspectTest(() => new Test) { top: Test =>
+      Select.wires(top).map(_.instanceName) should be(List("x_1", "x_3"))
+    }
+  }
```
jackkoenig, line 410: What happens when the return value is an `Option[(UInt, UInt)]`?

albertchen-sifive: the names are the same `x_1`, `x_2`. I added a test case

mwachs5 (commented):
this looks great!

Can you show what happens if there are temporaries? Something like:

def myFunc: (UInt, String) = { val out = IO(Output(...), in = IO(Input(...) , io := Mux(in(0), RegNext(in + 2), in << 3); (out, "Hi!")}

val foo = myFunc;

what are the IOs called, and the temporaries inside? I would have assumed they all have the prefix of `foo`?

src/test/scala/chiselTests/naming/NamePluginSpec.scala
```diff
@@ -375,4 +375,66 @@ class NamePluginSpec extends ChiselFlatSpec with Utils {
...
+        (out, "Hi!")
+      }
+
+      val foo = myFunc()
+    }
+
+    aspectTest(() => new Test) { top: Test =>
+      Select.ios(top).map(_.instanceName) should be(List("clock", "reset", "foo_1", "foo_in"))
+      Select.registers(top).map(_.instanceName) should be(List("foo_out_REG"))
+      Select.wires(top).map(_.instanceName) should be(List())
```
albertchen-sifive, line 437: > what are the IOs called, and the temporaries inside? I would have assumed they all have the prefix of foo?

yes they are all prefixed by foo. Since the output is assigned to the first element of `foo` it gets the name `foo_1` while the input gets `foo_in`. I think in this example the prefixed name `foo_out` would be better.

mwachs5: yeah the name `foo_out` would be better, but i think foo_1 is sensical at least and someone could wrap their head around why it is called that.

mwachs5 (approved):

PR #3437, opened 2023-07-24
ChiselPlugin: prefix destructuring assignments

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

ChiselPlugin will now prefix temporaries generated by the right-hand side of a destructuring assignment e.g. `val (a, b, _) = (Wire(Bool()), Wire(Bool()), Wire(Bool()))` will generate name `_a_b_WIRE` instead of `_WIRE` for the third element.

- [ ] Did you review?

### DIFF
--- a/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
+++ b/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
@@ -232,7 +232,11 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
               val onames: List[Option[String]] =
                 fieldsOfInterest.zip(names).map { case (ok, name) => if (ok) Some(name) else None }
               val newRHS = transform(rhs)
-              val named = q"chisel3.internal.plugin.autoNameRecursivelyProduct($onames)($newRHS)"
+              // for tuple destructuring assignments, prefix using the concatenation of all names
+              // e.g. for `val (a, b, c) = ...` use prefix "a_b_c"
+              val prefix = names.map(_.stripPrefix("_")).mkString("_")
+              val prefixed = q"chisel3.experimental.prefix.apply[$tpt](name=$prefix)(f=$newRHS)"
+              val named = q"chisel3.internal.plugin.autoNameRecursivelyProduct($onames)($prefixed)"
               treeCopy.ValDef(dd, mods, name, tpt, localTyper.typed(named))
             case None => // It's not clear how this could happen but we don't want to crash
               super.transform(tree)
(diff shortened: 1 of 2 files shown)


jackkoenig (changes requested):
I'm not certain that this is what we want to do in the case of an unapply. I think we should chat with some designers to figure out what the desired behavior is before we move forward with this.

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -232,7 +232,11 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
               val onames: List[Option[String]] =
                 fieldsOfInterest.zip(names).map { case (ok, name) => if (ok) Some(name) else None }
               val newRHS = transform(rhs)
-              val named = q"chisel3.internal.plugin.autoNameRecursivelyProduct($onames)($newRHS)"
+              // for tuple destructuring assignments, prefix using the concatenation of all names
+              // e.g. for `val (a, b, c) = ...` use prefix "a_b_c"
```
mwachs5, line 236: why would designers want the prefix on all of them? Usually this tupling is just done for convenience that a code block is returning 3 things, wouldn't we want to prefix each with just its corresponding variable name?

jackkoenig: The returned things aren't prefixed, but any local values in the RHS callstack are, for example:
```scala
val (a, b) = {
  val x, y = Wire(UInt(3.W))
  val sum = WireInit(x + y)
  (x, y)
}
```
This will generate (simplified a bit):
```
wire a : UInt<3>
wire b : UInt<3>
wire a_b_sum : UInt<3>
a_b_sum <= tail(add(a, b,), 1)
```
Whereas currently it generates
```
wire a : UInt<3>
wire b : UInt<3>
wire sum : UInt<3>
sum <= tail(add(a, b,), 1)
```

jackkoenig: I'm not certain if this is a change we really want, or if we do want to add prefixes here, what they should be, the 4 options are:
1. No prefix, the user can call prefix if they want (current behavior)
2. Pick the first val as a prefix (`a`)
3. Pick the last val as a prefix (`b`)
4. Use all vals as prefixes (`a_b` would be the result)

This PR is (4).

sequencer:
Why not you sing CIRCT transform to handle this?
If the code inside WireInit is a block, what will happen in this situation?

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
mwachs5: IMHO the only thing that makes sense here is 1

PR #3452, opened 2023-08-03
[ci] Add "nightly" option to install-circt action

seldridge:
Add support for grabbing the latest nightly version of `firtool` to the
install-circt action.  This requires passing the "github.token" so that
the GitHub API can be used.  Note that the token is marked as "required",
however, it is only actually needed when using a nightly build.

### DIFF
--- a/.github/workflows/build-scala-cli-template/action.yml
+++ b/.github/workflows/build-scala-cli-template/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Install CIRCT
       uses: ./.github/workflows/install-circt
       with:
-        github-token: ${{ gitub.token }}
+        github-token: ${{ github.token }}
     - name: Cache Scala-CLI
       uses: coursier/cache-action@v6
     - name: Setup Scala-CLI
--- a/.github/workflows/build-scala-cli-template/action.yml
+++ b/.github/workflows/build-scala-cli-template/action.yml
@@ -5,6 +5,8 @@ runs:
   steps:
     - name: Install CIRCT
       uses: ./.github/workflows/install-circt
+      with:
+        github-token: ${{ gitub.token }}
     - name: Cache Scala-CLI
       uses: coursier/cache-action@v6
     - name: Setup Scala-CLI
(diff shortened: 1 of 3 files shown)


.github/workflows/install-circt/action.yml
```diff
@@ -2,24 +2,54 @@ name: Install CIRCT
...
 inputs:
   version:
-    description: 'version to install'
+    description: 'version to install ("nightly" installs latest nightly)'
     required: false
     default: 'firtool-1.48.0'
+  github-token:
+    description: 'The GitHub token used to download CIRCT nightly'
+    required: true
+    default: ${{ github.token }}
```
jackkoenig, line 11: I'm guessing it's required that this be threaded through and you can't just always use `${{ github.token }}`? If so, what's the point of the default? If it need not be threaded through then maybe don't bother passing it at every call site?

jackkoenig (approved):
LGTM although see the question/suggestion on the github-token argument.

.github/workflows/build-scala-cli-template/action.yml
```diff
@@ -5,6 +5,8 @@ runs:
   steps:
     - name: Install CIRCT
       uses: ./.github/workflows/install-circt
+      with:
+        github-token: ${{ gitub.token }}
```
jackkoenig, line 9:
```suggestion
        github-token: ${{ github.token }}
```
This typo is breaking CI from running 😂

.github/workflows/install-circt/action.yml
seldridge: Empirically, it had to be threaded through. Supporting evidence is that you see this with other actions. E.g., the docs mention passing the token to the labeler action (https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input) and they seem to do it with the `default` (https://github.com/actions/labeler/blob/main/action.yml#L6-L8). I cargo-culted this from there.

If this isn't an error, then it may be so that you can test the action in isolation (where you have a `github.token`). However, I couldn't get this to just grab the token from the other action without passing it.

jackkoenig: I think if it's required, there's no point in having a default right?

PR #3453, opened 2023-08-03
Add --dump-fir option to ChiselStage

jackkoenig:
As usual, a 20 min hack had a bit of scope creep but I think this is a nice improvement

I'm open to a better name for the argument, but I think the functionality is pretty handy

@fabianschuiki FYI

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Rebase

#### Release Notes
This option will dump the .fir before invoking firtool.

* Additional changes:
    * Use os.lib for invoking firtool
    * Use lazy serialization to avoid holding the entire FIRRTL in memory.
    * Mix NoStackTrace into FirtoolNotFound
    * Fix detection of no firtool

- [ ] Did you review?

### DIFF
--- a/src/main/scala/circt/stage/Annotations.scala
+++ b/src/main/scala/circt/stage/Annotations.scala
@@ -137,3 +137,17 @@ private[circt] case object SplitVerilog extends NoTargetAnnotation with CIRCTOpt
   )
 
 }
+
+/** Write the intermediate `.fir` file in [[circt.stage.ChiselStage]]
+  */
+private[circt] case object DumpFir extends NoTargetAnnotation with CIRCTOption with HasShellOptions {
+  override def options = Seq(
+    new ShellOption[Unit](
+      longOption = "dump-fir",
+      toAnnotationSeq = _ => Seq(this),
+      helpText = "Write the intermediate .fir file",
+      helpValueName = None
+    )
+  )
+
+}
--- a/src/main/scala/circt/stage/phases/CIRCT.scala
+++ b/src/main/scala/circt/stage/phases/CIRCT.scala
@@ -107,6 +107,7 @@ private[this] object Exceptions {
                     |  https://github.com/llvm/circt/releases""".stripMargin
         )
       )
+      with NoStackTrace
 
 }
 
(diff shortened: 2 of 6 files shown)


src/main/scala/circt/stage/Annotations.scala
```diff
@@ -137,3 +137,17 @@ private[circt] case object SplitVerilog extends NoTargetAnnotation with CIRCTOpt
   )
 
 }
+
+/** Write the intermediate `.fir` file in [[circt.stage.ChiselStage]]
+  */
+case object DumpFir extends NoTargetAnnotation with CIRCTOption with HasShellOptions {
```
seldridge, line 143: For truly new annotations, consider making these package private and force users to go through the CLI-level APIs.

src/main/scala/circt/stage/phases/CIRCT.scala
```diff
@@ -175,20 +176,34 @@ class CIRCT extends Phase {
...
-      case Some(circuit) => CircuitWithAnnos(circuit = circuit, annotations = filteredAnnotations).serialize
+    val serialization: Iterable[String] = firrtlOptions.firrtlCircuit match {
+      case None => throw new OptionsException("No input file specified!")
+      // TODO can we avoid converting, how else would we include filteredAnnos?
+      case Some(circuit) =>
+        val cwa = CircuitWithAnnos(circuit = circuit, annotations = filteredAnnotations)
+        firrtl.ir.Serializer.lazily(cwa)
     }
 
+    val (input: Option[Iterable[String]], firFile: Option[os.Path]) =
```
seldridge, line 187: The None, None and Some, Some aren't reachable. Consider Either or a sum type.

src/test/scala/circtTests/stage/ChiselStageSpec.scala
```diff
@@ -1077,5 +1105,19 @@ class ChiselStageSpec extends AnyFunSpec with Matchers with chiselTests.Utils {
...
+      }
+
+      info("The exception includes a useful error message")
+      val message = exception.getMessage
+      message should include("potato not found")
+      message should include("Chisel requires that firtool, the MLIR-based FIRRTL Compiler (MFC), is installed")
+
+      info("The exception should not contain a stack trace")
+      exception.getStackTrace should be(Array())
+    }
```
seldridge, line 1120: Nice addition.

seldridge (approved):

src/main/scala/circt/stage/Annotations.scala
jackkoenig: I've kind of waffled on this, for users who have their own command line parsers, they need the `HasShellOptions` objects to be public to build their own CLI. That being said, I'm not sure what use case there would be to use this option on its own rather than just using all of `circt.stage.CLI` so I think making this one private is reasonable. Options to Chisel generation probably need to remain public but that is not what this is!

PR #3454, opened 2023-08-03
[docs] Update naming explanation for newer firtool

jackkoenig:
Firtool improvements have optimized many examples to the point where they do not show anything. Using registers and assigning the same value to multiple outputs helps make the examples show interesting naming behavior.

h/t @lsteveol for noticing and pointing this out on Gitter

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?


docs/src/explanations/naming.md
```diff
@@ -153,35 +171,8 @@ Users who desire a prefix are encouraged to provide one as [described below](#pr
...
-
-  val add = in + in + in
-
-  out := prefix("ECO") { add + 1.U + in }
-}
-```
-```scala mdoc:verilog
-emitSystemVerilog(new Example4)
-emitSystemVerilog(new Example5)
-```
```
jackkoenig, line 184: I deleted the ECO example only because I literally cannot figure out a reliable way to make temporaries show up in the Verilog (other than naming the signals myself but that sort of defeats the purpose).

I do think we should have an ECO example but we need to make sure it captures enough complexity to actually be interesting. I'm also worried that firtool's much more aggressive optimizations will make this sort of ECO stuff harder to do so it may just not really be showable at the moment (without adding some sort of optimization barrier primitive).

If anyone can come up with a good example, I'm happy to add it back.

mikeurbach (approved):
I took a look, and it makes sense to me.

PR #3458, opened 2023-08-04
Switch to firtool-resolver

jackkoenig:
This uses https://github.com/jackkoenig/firtool-resolver which I will transfer to be an actual chipsalliance repo once this is merged.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

#### Release Notes

Whereas in the past, the user was expected to provide the firtool binary
on the PATH, Chisel will now manage the firtool binary automatically.
Chisel users the following 3-step process for determining which binary
to use:

1. If environment variable FIRTOOL_PATH is set, it will check for
   firtool in that directory
2. It will check the classpath for firtool provided by
   chipsalliance/llvm-firtool
3. It will download a default version of chipsalliance/llvm-firtool

Both (2) and (3) will store the firtool binary in a sensible location
depending on the user's operating system. See dirs-dev/directories-jvm
for more information. The user can override this location by setting the
FIRTOOL_CACHE environment variable.

- [ ] Did you review?

### DIFF
--- a/.github/workflows/build-scala-cli-template.yml
+++ b/.github/workflows/build-scala-cli-template.yml
@@ -5,7 +5,6 @@ on:
     inputs:
       circt:
         description: 'The version of CIRCT to use'
-        default: 'default'
         type: string
 
 jobs:
@@ -21,7 +20,7 @@ jobs:
           fetch-depth: 0
       - name: Install CIRCT
         id: install-circt
-        if: ${{ inputs.circt != 'default' }}
+        if: ${{ inputs.circt }}
         uses: circt/install-circt@v1.1.1
         with:
           version: ${{ inputs.circt }}
--- a/.github/workflows/build-scala-cli-template.yml
+++ b/.github/workflows/build-scala-cli-template.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       circt:
         description: 'The version of CIRCT to use'
-        default: 'version-file'
+        default: 'default'
         type: string
 
 jobs:
@@ -20,10 +20,18 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install CIRCT
-        uses: circt/install-circt@v1.0.0
+        id: install-circt
+        if: ${{ inputs.circt != 'default' }}
+        uses: circt/install-circt@v1.1.1
         with:
           version: ${{ inputs.circt }}
           github-token: ${{ github.token }}
+      # TODO have install-circt do this
+      - name: Set FIRTOOL_PATH
+        if: steps.install-circt.outcome == 'success'
+        run: |
+          dir=$(dirname $(which firtool))
+          echo "FIRTOOL_PATH=$dir" >> "$GITHUB_ENV"
       - name: Cache Scala-CLI
         uses: coursier/cache-action@v6
       - name: Setup Scala-CLI
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ on:
         required: true
         type: string
       circt:
-        description: 'The CIRCT version to use (must be a valid release tag or "nightly")'
+        description: 'The CIRCT version to use (must be a valid release tag, "default", "nightly", or "version-file")'
         required: true
         type: string
       ref:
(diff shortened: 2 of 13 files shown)


sequencer:
I wonder how about merge firtool-resolver into the chisel repo?

jackkoenig:
> I wonder how about merging firtool-resolver into the chisel repo?

We could. I sort of like keeping it separate because llvm-firtool (the module for re-publishing firtool binaries as Java artifacts) lives in the same repo as firtool-resolver and thus can use firtool-resolver to test itself.

sequencer:
my point is, if living in chisel repo, it might be convenient to bump with the ci-nightly branch? It is good for those following the master branch.

.github/workflows/build-scala-cli-template.yml
```diff
@@ -20,10 +20,18 @@ jobs:
...
           fetch-depth: 0
       - name: Install CIRCT
-        uses: circt/install-circt@v1.0.0
+        id: install-circt
+        if: ${{ inputs.circt != 'default' }}
+        uses: circt/install-circt@v1.1.1
         with:
           version: ${{ inputs.circt }}
           github-token: ${{ github.token }}
+      # TODO have install-circt do this
```
seldridge, line 29: I agree that this seems reasonable to do.

That would avoid this type of chained conditional steps. Such chained conditional steps usually seems like an anti-pattern. No real problem with this as-is, though.

.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ on:
         required: true
         type: string
       circt:
-        description: 'The CIRCT version to use (must be a valid release tag or "nightly")'
+        description: 'The CIRCT version to use (must be a valid release tag, "default", "nightly", or "version-file")'
```
seldridge, line 27: Default may be a bit ambiguous. This is probably fine as-is. Consider either making this empty/not a required option or something more descriptive like "use-firtool-resolver".

.github/workflows/test.yml
```diff
@@ -56,15 +56,24 @@ jobs:
...
+        run: |
+          dir=$(dirname $(which firtool))
+          echo "FIRTOOL_PATH=$dir" >> "$GITHUB_ENV"
       - name: Print firtool version
+        if: steps.install-circt.outcome == 'success'
         run: |
           echo "The CIRCT version used is:" >> $GITHUB_STEP_SUMMARY
           echo \`\`\` >> $GITHUB_STEP_SUMMARY
-          firtool -version >> $GITHUB_STEP_SUMMARY
+          $FIRTOOL_PATH/firtool -version >> $GITHUB_STEP_SUMMARY
```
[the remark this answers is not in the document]
seldridge: Also, consider inlining this step with the previous step as they have the same predicate.

seldridge (approved):
LGTM

It's great to see this finally happening.

jackkoenig:
I've marked this for backporting to 3.6. It's technical an API change since firtool on the PATH will now be ignored, but I think it's a worthwhile one for usability.

.github/workflows/test.yml
jackkoenig: Yeah good point. I think I'll go with not-required (thus empty) because that makes more sense that you're not installing a specific input, you're just leaving it up to chisel.

.github/workflows/test.yml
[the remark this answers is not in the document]
jackkoenig: My thinking here was that the above step setting FIRTOOL_PATH will eventually be folded into `circt/install-circt`, so I factored this in the way such that once that happens, we can just bump the `circt/install-circt` version and delete the step above and everything will continue to work. Basically writing this code such that the expected next change will be extremely simple, does that make sense?

jackkoenig:
> my point is, if living in chisel repo, it might be convenient to bump with the ci-nightly branch? It is good for those following the master branch.

I see your point but I don't think that's going to be an issue. firtool-resolver (and llvm-firtool) are intended to be as independent of the actual APIs they wrap as possible. We may need to update them as firtool evolves packaging-wise (adding more artifacts, etc.) but never with the actual day-to-day APIs. I don't see a problem with keeping it backwards (and forwards) compatible enough to avoid needing to change it with nightly.

PR #3460, opened 2023-08-04
[ci] Add CIRCT nightly CI

seldridge:
Add nightly testing of Chisel "main" branch with the nighlty published
"firtool" binary.  This testing will also be run on PRs targeting the
staging branch, "ci/ci-circt-nightly".

---

This runs nightly testing using the published `firtool` binary made available through this action on CIRCT: https://github.com/llvm/circt/actions/workflows/uploadFirrtlReleaseArtifacts.yml. The Chisel action (that this PR introduces) is setup to run one hour after the CIRCT one starts.

The nightly testing relies on a staging branch, `ci/ci-circt-nightly`, to deal with changes that are necessary for the nightly to pass. Nightly testing will try to update this branch (via rebase) onto the `main` branch before it builds.

Patches can be included in the `ci/ci-circt-nightly` branch by either manually updating it or by making PRs against it. If PRs are made against it, then this action will also run, but without updating the `ci/ci-circt-nightly` branch first.

For an example of how this all composes, consider the following live example. Currently, `firtool` nightly includes changes which fix a bug in the Trace API. Current Chisel `main` includes which check that the incorrect behavior occurs, i.e., the tests are wrong. Consequently, Chisel `main` w/ the `firtool` nightly will fail. This can be seen if manually triggering the nightly action here: https://github.com/chipsalliance/chisel/actions/runs/5768453794 However, the PR that fixes this will pass: https://github.com/chipsalliance/chisel/actions/runs/5768391177?pr=3451

Summarizing the complete workflow:

1. Nightly is broken as discovered by this action running.
2. A user makes a PR to update the `ci/ci-circt-nightly` branch.
3. This action is used to check that it fixed nightly.
4. The PR is merged.
5. Nightly now passes as checked by this PR.

### DIFF
--- a/.github/workflows/ci-circt-nightly.yml
+++ b/.github/workflows/ci-circt-nightly.yml
@@ -16,12 +16,104 @@ on:
   # Run on PRs that target the
   pull_request:
     branches:
-      - ${{ branch-name }}
+      - ci/ci-circt-nightly
 
 jobs:
-  hello:
-    name: Hello World
+  update-branch:
+    name: Integrate staging branch with default branch
     runs-on: ubuntu-20.04
     steps:
-      - name: Hello
-        run: echo "World"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: main
+      - run: |
+          # Leave the staging branch unchanged if this is a PR.  This is not a
+          # GitHub Actions-level "if" because we want this job to return success
+          # (so that subsequent jobs run).
+          if [[ ${{ github.event_name }} == 'pull_request' ]]; then
+            exit 0
+          fi
+
+          # Rebase the staging branch on the default branch.  Create the staging
+          # branch if it doesn't exist.
+          git checkout -b ${{ env.branch-name }}
+          if [[ `git fetch origin ${{ env.branch-name }}` ]]; then
+            git reset --hard origin ${{ env.branch-name }}
+            git rebase origin/main
+          fi
+
+          git push --force origin ${{ env.branch-name }}
+
+  # If this is running on a PR, then use the branch of the PR.  Otherwise, use
+  # the staging branch.
+  determine-branch:
+    name: Determine what branch to run tests on
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Determine Branch
+        id: determine-branch
+        run: |
+          if [[ ${{ github.event_name }} != 'pull_request' ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-circt-nightly.yml
```diff
@@ -16,12 +16,102 @@ on:
...
-        run: echo "World"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: main
+      - run: |
+          # Leave the staging branch unchanged if this is a PR.
+          if [[ ${{ github.event_name }} == 'pull_request' ]]; then
+            exit 0
+          fi
```
jackkoenig, line 34: At first I was confused why you put this `if` here rather than a Github Actions-level `if`. I know see that it's clearly do to https://github.com/orgs/community/discussions/45058, could you please add a comment above `run` that explains that?

.github/workflows/ci-circt-nightly.yml
```diff
@@ -16,12 +16,102 @@ on:
...
+    name: ci
+    needs: [update-branch, determine-branch]
+    strategy:
+      matrix:
+        system: ["ubuntu-20.04"]
+        jvm: [8]
+        scala: ["2.13.11"]
+        espresso: ["2.4"]
+        circt: ["nightly"]
+        ref: ${{ fromJSON(needs.determine-branch.outputs.branches) }}
```
jackkoenig, line 73: For my own understanding, are you using JSON here just so that you're able to include ref in the build matrix so that it shows up in the Github UI?

jackkoenig (approved):
LGTM but see request for clarifying comment.

.github/workflows/ci-circt-nightly.yml
seldridge: That would require this construction, though it's not the immediate problem this is solving.

I need to configure the build matrix. More accurately, I need to tell `test.yml` which branch to use. Either use whatever the current branch is (for the PR flow) or switch to the staging branch for the nightly flow. The previous job (`determine-branch`) figures out what the branch should be. Funneling this through JSON is the only way that I know how to do this.

Relatedly, the `env` context is not available in the `matrix` (I think). So, you wind up with a similar construction if you want to only define the branch name in one place.

.github/workflows/ci-circt-nightly.yml
seldridge: Yes. I want this to be a successful no-op, not a skip. I'll add a comment.

PR #3463, opened 2023-08-05
[ci] Change CIRCT version to exist in one place  

seldridge:
Change how the CIRCT version is specified for non-nightly flows to a single environment variable in the ci.yml script.  This makes things easier for either: (1) a human to update this or (2) a machine to update this in a CD flow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,16 +10,31 @@ on:
       - main
       - '*.x'
 
+env:
+  circt: "firtool-1.48.0"
+
 jobs:
+  set-circt-versions:
+    name: Set the CIRCT version from the env
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set CIRCT Version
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -10,16 +10,31 @@ on:
...
     name: ci
+    needs: [set-circt-versions]
     strategy:
       matrix:
         system: ["ubuntu-20.04"]
         jvm: [8]
         scala: ["2.13.11"]
         espresso: ["2.4"]
-        circt: ["firtool-1.48.0"]
+        circt: ${{ fromJSON(needs.set-circt-versions.outputs.circt-versions) }}
```
seldridge, line 37: All of this is necessary because `env` is not accessible when inside the `matrix` context. The `fromJSON` approach lets the CIRCT version still be specified in one place---in the `env`---and used in both the matrix construction and in the publish job (which does have access to the `env` context).

PR #3466, opened 2023-08-07
Implement CIRCT converter

SpriteOvO:
This PR:

- [x] Currently we just copy & paste source files to the repo, didn't make them work together, we will do it later.
- [x] llvm/circt#5790
- [x] llvm/circt#5832
- [ ] Make CI work and perform simple tests.
- [ ] ...

---
Follow-up PRs:

- [ ] Implement Verilog export.
- [ ] Figure out a way to make existing test cases also perform once for `PanamaCIRCTConverter`.
- [ ] ...

---

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?


core/src/main/scala/chisel3/internal/chisel-circt-binder/src/PanamaCIRCT.scala
```diff
@@ -0,0 +1,394 @@
...
+  def mlirModuleGetBody(module: MlirModule) = MlirBlock(CAPI.mlirModuleGetBody(arena, module.get))
+
+  def mlirModuleGetOperation(module: MlirModule) = MlirOperation(CAPI.mlirModuleGetOperation(arena, module.get))
+
+  def mlirOperationStateGet(name: String, loc: MlirLocation) = MlirOperationState(
+    CAPI.mlirOperationStateGet(arena, newString(name).get, loc.get)
+  )
+
+  def mlirOperationStateAddAttributes(state: MlirOperationState, attrs: Seq[MlirNamedAttribute]) = {
+    if (attrs.nonEmpty) {
```
sequencer, line 80: learn to use `attrs.foreach`

core/src/main/scala/chisel3/internal/chisel-circt-binder/src/PanamaCIRCT.scala
```diff
@@ -0,0 +1,394 @@
...
+
+final case class FIRRTLBundleField(name: String, isFlip: Boolean, tpe: MlirType)
+
+//
+// MLIR & CIRCT Enums
+//
+
+sealed abstract class FIRRTLNameKind(val value: Int) extends ForeignType[Int] {
+  private[circt] def get = value
+  private[circt] val sizeof = 4 // FIXME: jextract doesn't export type for C enum
```
sequencer, line 359: How to fix this?

core/src/main/scala/chisel3/internal/chisel-circt-binder/tests/src/Smoke.scala
sequencer, line 1: Although I love utest for the simplicity, but I think when upstream these codes, we should use the scalatest.

sequencer (commented):
Thanks @SpriteOvO! I think this is the first (experimental) step to link `libCIRCT.so` to Chisel.

Since it's experimental, we won't try to release it until it is stable and JVM 21 is out. We are only make it exist inside Chisel code base for let it be there, and for advance users/devs to try.

In the future, there will be a lot of new features depending on this, not limiting the
- Chisel IDE interaction
- StackTrace level `SourceInfo`
- `HasId` mapping to Verilog lines
- dynamic `TraceAPI`

core/src/main/scala/chisel3/internal/chisel-circt-binder/src/CIRCTConverter.scala
```diff
@@ -0,0 +1,39 @@
...
+package chisel3.internal
+
+import geny.Writable
+import chisel3.{Data, VerificationStatement => VerifStmt}
+import chisel3.assert.Assert
+import chisel3.assume.Assume
+import chisel3.cover.Cover
+import chisel3.internal.firrtl._
+
+abstract class CIRCTConverter {
```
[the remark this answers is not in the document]
SpriteOvO: Hmm, `abstract class` also used elsewhere in Chisel, it has a different semantics than `trait` - a class can inherit from multiple traits but only one abstract class.

Maybe `abstract class` would be more appropriate here?

core/src/main/scala/chisel3/internal/chisel-circt-binder/src/PanamaCIRCT.scala
SpriteOvO: Here we convert `attrs` to C array `(MemorySegment, Int)`, `.foreach` is not helpful.

core/src/main/scala/chisel3/internal/chisel-circt-binder/src/PanamaCIRCT.scala
SpriteOvO: The problem is `jextract` doesn't extract C enum to a Java class, so that we can't get the size of C enums.

We can simply assume the size of enums are 4, that should not be break easily from CIRCT unless someone explicitly changes the type of enumerations.

sequencer:
For those nix files it should be split to another PR

sequencer:
TODO after this PR:
- Add options based on [firtool options](https://github.com/llvm/circt/blob/08778859fc22dc1b5625b310ee18e66cca86e0e1/tools/firtool/firtool.cpp#L68-L224)
- Add passes based on [firtool passes](https://github.com/llvm/circt/blob/08778859fc22dc1b5625b310ee18e66cca86e0e1/tools/firtool/firtool.cpp#L606-L623)

binder/src/main/scala/PanamaCIRCT.scala
```diff
@@ -0,0 +1,394 @@
...
+  private val arena = Arena.openConfined()
+
+  // Create MLIR context and register dialects we need
+  private val mlirCtx = {
+    val mlirCtx = CAPI.mlirContextCreate(arena)
+
+    // Register dialects
+    Seq(
+      CAPI.mlirGetDialectHandle__firrtl__(arena),
+      CAPI.mlirGetDialectHandle__chirrtl__(arena)
```
sequencer, line 24: By default these two dialect must be registered.
But if emitting Verilog and use full feature of CIRCT need more things to add.
See dialect, firtool loaded [here](https://github.com/llvm/circt/blob/08778859fc22dc1b5625b310ee18e66cca86e0e1/tools/firtool/firtool.cpp#L573-L576)
It can be live in the future work

binder/src/main/scala/PanamaCIRCT.scala
```diff
@@ -0,0 +1,394 @@
...
+import org.llvm.circt
+import org.llvm.circt.CAPI
+
+// Wrapper for CIRCT APIs with Panama framework
+class PanamaCIRCT {
+  // Open an arena for memory management of MLIR API calling in this context instance
+  private val arena = Arena.openConfined()
+
+  // Create MLIR context and register dialects we need
+  private val mlirCtx = {
```
sequencer, line 18: Think about it:
for each ctx, how to record current Dialect, not firrtl, chirrtl only, but `chirrtl::CHIRRTLDialect, firrtl::FIRRTLDialect, hw::HWDialect, comb::CombDialect, seq::SeqDialect, om::OMDialect, sv::SVDialect, verif::VerifDialect, ltl::LTLDialect`,
how to record current transform(currently unimplemented)?
Is possible to `copy` an ctx(may not be possible), so we can pass `firrtl` to `verilog` and `arc` dialect at same time.

sequencer (commented):
Please remove the nix files for one feature one PR.

sequencer (approved):
Approve but not release!

PR #3468, opened 2023-08-08
[cd] Add CD of CIRCT Releases

seldridge:
Add a CD (continuous delivery) GitHub action that will automatically create PRs that bump CIRCT to the latest version.

There are a couple of design decisions which should be justified:

This uses `pandoc` to generate the the text of the PR title, PR body, and git commit message. I originally tried to do this with simple `bash` scripting. However, I was failing to come up with a good way of doing complicated conditional logic and iteration. E.g., I want to show an itemized list of links to `firtool` release notes. Doing this in pure `bash` made things incredibly difficult to understand what the actual template is supposed to be. By pulling in `pandoc`, the template can be kept separate and is (mostly) readable. This also opens the door for arbitrarily complex messages. Additionally, using the official `pandoc` GitHub action is surprisingly fast. (GitHub Actions show it running in "zero seconds".)

The JSON payload is an odd choice. This originally came about because I needed a way to get information _into_ `pandoc`. Pandoc will accept either a YAML or JSON configuration file (or you can specify in inline if we had a file we were passing to `pandoc`). Somewhat surprisingly, it was easier and more straightforward to use `jq` to add information to this than it was to do the equivalent with pure `bash`. The pure `bash` approach resulted in a blowup of lots of additional outputs or files that needed to be passed between steps or jobs. After switching to this approach, it proved easy to use the JSON payload to drive what commits are cherry-picked. This is also a bit odd, but the JSON gives you a single thing that you can inspect if you have some confusion about what is going on.

It is a huge pain to get a multi-line string into a GitHub Actions step or job. This is what the boilerplate in the "Setup Environment and Metadata" step is doing. This is required to use newlines properly in the "Create PR" step.

#### Example

An example of a CD run (from a `workflow_dispatch` manual trigger) is here:
  - https://github.com/chipsalliance/chisel/actions/runs/5815199883

An example of a PR created by this workflow is here: 
  - https://github.com/chipsalliance/chisel/pull/3480


.github/workflows/cd-circt.yml
```diff
@@ -1,11 +1,316 @@
...
+    # a bunch of JSON metadata about what is about to be done.  This JSON
+    # metadata is used to drive the cherry-pick action of (1), the commit
+    # message of (2), and the PR title and body of (3).
+    #
+    # This job re-clones both Chisel and CIRCT, in their entirety.  This is done
+    # because it is simpler to gate the JSON (and commit and PR messages) behind
+    # _this job_ where we know that a PR is going to be created.  We could save
+    # a CIRCT clone by doing this in the "Determine Versions" job at the cost of
+    # complicating this Workflow by needing to pass information between the
+    # jobs.
```
jackkoenig, line 89: FWIW, you always need to re-clone between different jobs, otherwise you'd have to be uploading and downloading the artifacts anyway. I'm not sure if this comment is really needed since it's basically inherent in how Github Actions work, see how every single job in `test.yml` has to clone Chisel.

.github/workflows/cd-circt.yml
```diff
@@ -1,11 +1,316 @@
...
+        id: setup-env
+        run: |
+          echo pr-title=$(cat metadata/pr-title.md) >> $GITHUB_OUTPUT
+
+          EOF=$(cat /dev/urandom | base64 -w 128 | head -n8 | tail -n1)
+          {
+            echo "commit-message<<$EOF"
+            cat metadata/commit-message.md
+            echo $EOF
+          } >> $GITHUB_OUTPUT
```
jackkoenig, line 297: Another option could be to create the commit ourselves and use `git commit -F`. But yeah this is reasonable. I wish the action had a `commit-message-path` option 🙃

jackkoenig (approved):
LGTM, awesome work @seldridge! A couple of comments to consider but merge whenever you're ready.

.github/workflows/cd-circt.yml
seldridge: It would be easy enough to fix this on the upstream repo. I'll leave this for now.

PR #3482, opened 2023-08-10
Add support for Property literals, starting with existing int types.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

This includes the API, Chisel IR, Converter, FIRRTL IR, and Serializer support for Property literals.

This is currently stacked on #3481 and can be merged after that one.

#### Desired Merge Strategy

#### Release Notes

Property literals are defined in FIRRTL spec version 3.1.0: https://github.com/chipsalliance/firrtl-spec/releases/tag/v3.1.0.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -93,10 +93,9 @@ private[chisel3] object Converter {
     // TODO Simplify
     case lit: ILit =>
       throw new InternalErrorException(s"Unexpected ILit: $lit")
-    case lit: PropertyLit =>
-      lit match {
-        case IntegerPropertyLit(lit) => fir.IntegerPropertyLiteral(lit)
-      }
+    case PropertyLit(lit: Int) => fir.IntegerPropertyLiteral(lit)
+    case PropertyLit(lit: Long) => fir.IntegerPropertyLiteral(lit)
+    case PropertyLit(lit: BigInt) => fir.IntegerPropertyLiteral(lit)
     case e @ ProbeExpr(probe) =>
       fir.ProbeExpr(convert(probe, ctx, info))
     case e @ RWProbeExpr(probe) =>
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -93,9 +93,10 @@ private[chisel3] object Converter {
     // TODO Simplify
     case lit: ILit =>
       throw new InternalErrorException(s"Unexpected ILit: $lit")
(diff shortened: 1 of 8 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -93,6 +93,9 @@ private[chisel3] object Converter {
     // TODO Simplify
     case lit: ILit =>
       throw new InternalErrorException(s"Unexpected ILit: $lit")
+    case PropertyLit(lit: Int) => fir.IntegerPropertyLiteral(lit)
+    case PropertyLit(lit: Long) => fir.IntegerPropertyLiteral(lit)
+    case PropertyLit(lit: BigInt) => fir.IntegerPropertyLiteral(lit)
```
mikeurbach, line 98: It might be better to do something more typeclassy here like we did for Property types.... I am taking a look at how to do something like that here.

mwachs5 (approved):

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
mikeurbach: I tried this out and pushed it up in https://github.com/chipsalliance/chisel/pull/3482/commits/03d2034d1afbf33f91995acedf9a3d7860cfebc3, where now the PropertyType typeclass is responsible for handing out literals, like how it hands out types. Curious if this seems better. What do you think @seldridge @jackkoenig @mwachs5? I'm happy to go either route, and I see Megan approved the initial version.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -21,6 +21,10 @@ private[chisel3] trait PropertyType[T] {
   /** Get the IR PropertyType for this PropertyType.
     */
   def getPropertyType: ir.PropertyType
+
+  /** Get the IR PropertyLit for this PropertyType.
+    */
+  def getPropertyLit(lit: T): ir.PropertyLit
 }
```
jackkoenig, line 28: This being part of `PropertyType` makes sense for the currently existing property types, but what is it going to look like for objects? Are we committing to supporting literals for arbitrary user-defined types?

This question is not blocking, just something we need to keep in mind as we continue to flesh out this API.

jackkoenig (approved):
LGTM for Integers at least, we will see how this extends to user-defined types.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Hmm that's a good point, maybe we don't want to commit to that. I could revert back to the initial version, or roll with this for now.

mikeurbach: I'm going to flip back to the original version for now and plan to merge this.

PR #3483, opened 2023-08-10
Support BaseType in BoringUtils and associated helpers.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- API modification

This generalizes a few helper methods that previously accepted Data to now accept BaseType. The main user-facing API change is to accept BaseType in `BoringUtils.bore`. This allows boring Property typed ports.

This is currently stacked on https://github.com/chipsalliance/chisel/pull/3481, which defines PropAssign.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -6,6 +6,7 @@ import scala.util.Try
 import scala.language.experimental.macros
 import scala.annotation.nowarn
 import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
+import chisel3.properties.Property
 import chisel3.internal._
 import chisel3.experimental.hierarchy.{InstanceClone, ModuleClone}
 import chisel3.internal.Builder._
@@ -176,13 +177,17 @@ abstract class RawModule extends BaseModule {
   }
   private[chisel3] val stagedSecretCommands = collection.mutable.ArrayBuffer[Command]()
 
-  private[chisel3] def secretConnection(left: Data, right: Data)(implicit si: SourceInfo): Unit = {
-    val rhs = (left.probeInfo.nonEmpty, right.probeInfo.nonEmpty) match {
-      case (true, true)                                 => ProbeDefine(si, left.lref, Node(right))
-      case (true, false) if left.probeInfo.get.writable => ProbeDefine(si, left.lref, RWProbeExpr(Node(right)))
-      case (true, false)                                => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
-      case (false, true)                                => Connect(si, left.lref, ProbeRead(Node(right)))
-      case (false, false)                               => Connect(si, left.lref, Node(right))
+  private[chisel3] def secretConnection(left: BaseType, right: BaseType)(implicit si: SourceInfo): Unit = {
+    val rhs = (left, right) match {
+      case (lprop: Property[_], rprop: Property[_]) => PropAssign(si, lprop.lref, Node(rprop))
+      case (ldata: Data, rdata: Data) =>
+        (ldata.probeInfo.nonEmpty, rdata.probeInfo.nonEmpty) match {
+          case (true, true)                                  => ProbeDefine(si, ldata.lref, Node(rdata))
+          case (true, false) if ldata.probeInfo.get.writable => ProbeDefine(si, ldata.lref, RWProbeExpr(Node(rdata)))
+          case (true, false)                                 => ProbeDefine(si, ldata.lref, ProbeExpr(Node(rdata)))
+          case (false, true)                                 => Connect(si, ldata.lref, ProbeRead(Node(rdata)))
+          case (false, false)                                => Connect(si, ldata.lref, Node(rdata))
+        }
     }
     val secretCommands = if (_closed) {
       _component.get.asInstanceOf[DefModule].secretCommands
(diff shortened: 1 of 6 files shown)


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -255,7 +256,7 @@ object BoringUtils {
             val bore = if (up) module.createSecretIO(purePortType) else module.createSecretIO(Flipped(purePortTypeBase))
             module.addSecretIO(bore)
             conLoc.asInstanceOf[RawModule].secretConnection(bore, rhs)
-            bore
+            bore.asInstanceOf[A]
```
mikeurbach, line 259: I'm not quite sure why I needed asInstanceOf here, and below. Without it, Scala gives me this error:

```
[error] /Users/mikeu/chisel/src/main/scala/chisel3/util/experimental/BoringUtils.scala:259:13: type mismatch;
[error]  found   : bore.type (with underlying type chisel3.BaseType)
[error]  required: A
[error]             bore
[error]             ^
```

I was under the impression that type `A` _is_ `BaseType`. Clearly I am out of my depth here... is this the right thing to do?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -176,13 +177,17 @@ abstract class RawModule extends BaseModule {
...
+  private[chisel3] def secretConnection(left: BaseType, right: BaseType)(implicit si: SourceInfo): Unit = {
+    val rhs = (left, right) match {
+      case (lprop: Property[_], rprop: Property[_]) => PropAssign(si, lprop.lref, Node(rprop))
+      case (ldata: Data, rdata: Data) =>
+        (ldata.probeInfo.nonEmpty, rdata.probeInfo.nonEmpty) match {
+          case (true, true)                                  => ProbeDefine(si, ldata.lref, Node(rdata))
+          case (true, false) if ldata.probeInfo.get.writable => ProbeDefine(si, ldata.lref, RWProbeExpr(Node(rdata)))
+          case (true, false)                                 => ProbeDefine(si, ldata.lref, ProbeExpr(Node(rdata)))
+          case (false, true)                                 => Connect(si, ldata.lref, ProbeRead(Node(rdata)))
+          case (false, false)                                => Connect(si, ldata.lref, Node(rdata))
```
mwachs5, line 189: should there be a non-exhaustive match case here?

src/main/scala/chisel3/util/experimental/BoringUtils.scala
mwachs5: what is the type of `bore`? It is the common ancestor of whatever `module.createSecretIO(purePortType)` and `module.createSecretIO(Flipped(purePortTypeBase)` is known to be... does createSecretIO properly return the types you want there?

mwachs5: the compiler is complaining that while A is indeed a BaseType, it wants `bore` to be the specific subclass type `A`, not just a `BaseType`. something about that `if ... createSecretIO else ...` is making it lose the understanding that it is actually an `A`

mwachs5: i think the problem is here, all it knows is that sometimes this is a Data and sometimes its an A so it concludes it can only be a BaseType:

mwachs5: I think the problem is in this bit:
```
def purePortType = (createProbe, purePortTypeBase) match {
      case (Some(pi), d: Data) if pi.writable => RWProbe(d)
      case (Some(pi), d: Data) => Probe(d)
      case (_, _) => purePortTypeBase
    }
```

some of those cases are only Data, and some are your A type, the common ancestor of which is just BaseType. Can you make the A <: Data constraint be there instead of just Data?

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: Do you mean like `case (_, _)` for the outer `match`? I didn't change the inner `match`, which I think is already exhaustive. If i'm understanding, yeah that makes sense, I can at least throw a meaningful exception in that case.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
mikeurbach: Thanks for the help! I will try it out.

mikeurbach: I tried poking around at this, and I'm still not sure the right way out of it. I'm not able to do `RWProbe(purePortTypeBase)` because you can only probe Data, which is why this change was introduced in the first place. I'm not sure how to express the constraint here.

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: I added an exception base case, but I think it might actually be impossible to hit--the Scala compiler should ensure you're either boring Data to Data, or Property to Property.

mikeurbach: I guess it would come up if we had a new kind of BaseType

src/main/scala/chisel3/util/experimental/BoringUtils.scala
mwachs5: can you do:
```
      case (Some(pi), d: A <: Data) if pi.writable => RWProbe(d)
```

mikeurbach: Oh I see what you mean conceptually. I don't think this is legal scala syntax but i will see if i can express this.

mikeurbach:
No longer applicable.

PR #3486, opened 2023-08-13
add HasExtModuleDefine.define for creating ProbeType in BlackBox

sequencer:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes
Add HasExtModuleDefine.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -112,4 +112,22 @@ package object probe extends SourceInfoDoc {
     pushCommand(ProbeRelease(sourceInfo, clock.ref, cond.ref, probe.ref))
   }
 
+  /** The mixin to define the probe signal for BlackBox. */
+  trait HasExtModuleDefine extends chisel3.util.HasExtModuleInline { this: chisel3.experimental.ExtModule =>
+
+    /** Create a ProbeType for external sources.
+      * @param tpe is the Chisel type of signal.
+      * @param path is the internal path of the signal see [[https://github.com/chipsalliance/firrtl-spec/blob/main/abi.md]]
+      */
+    def define[T <: chisel3.Element](tpe: T, path: Seq[String]): T = {
+      setInline(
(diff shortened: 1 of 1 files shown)


core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -112,4 +112,22 @@ package object probe extends SourceInfoDoc {
...
+  /** The mixin to define the probe signal for BlackBox. */
+  trait HasExtModuleDefine extends chisel3.util.HasExtModuleInline { this: chisel3.experimental.ExtModule =>
+
+    /** Create a ProbeType for external sources.
+      * @param tpe is the Chisel type of signal.
+      * @param path is the internal path of the signal see [[https://github.com/chipsalliance/firrtl-spec/blob/main/abi.md]]
+      */
+    def define[T <: chisel3.Element](tpe: T, path: Seq[String]): T = {
+      setInline(
+        s"firrtl_abi_definition_${path.mkString("_")}.sv",
```
dtzSiFive, line 124: Aw, this is what `ref` statements were intended for.  They're not really used so no worries, the verilog ABI files is what we have!

The filename (and portion after "ref_" in the define) have specific rules, should these be automatically populated based on the circuit / module (def?) names?

> Ports of ref type shall be lowered to a Verilog macro of the form `` `define
ref_<circuit name>_<module name>_<portname> <internal path from module>`` in a
file with name `ref_<circuit name>_<module name>.sv`.

These need to be output ports, and maybe if we're teaching Chisel about ABI either user is entirely responsible or we would do the necessary handling for aggregates/names?

Anyway, fly-by, just some thoughts!

sequencer: Thanks! I'll be too busy next week, I'm drafting and using this API in https://github.com/sequencer/riscv-cosim. After I think this API is good enough I'll be back to upstream to Chisel.

PR #3489, opened 2023-08-14
Add support for Class and Object on top of Properties.

mikeurbach:
### Contributor Checklist

Sending a draft now, thought it isn't quite ready.

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -8,6 +8,7 @@ import scala.annotation.nowarn
 import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
 import chisel3.internal._
 import chisel3.experimental.hierarchy.{InstanceClone, ModuleClone}
+import chisel3.properties.DynamicObject
 import chisel3.internal.Builder._
 import chisel3.internal.firrtl._
 import _root_.firrtl.annotations.{IsModule, ModuleTarget}
@@ -139,6 +140,12 @@ abstract class RawModule extends BaseModule {
         case id: assume.Assume    => id.forceName(default = "assume", _namespace)
         case id: cover.Cover      => id.forceName(default = "cover", _namespace)
         case id: printf.Printf => id.forceName(default = "printf", _namespace)
+        case id: DynamicObject => {
+          // Force name of the DynamicObject, and set its Property[ClassType] type's ref to the DynamicObject.
+          // The type's ref can't be set upon instantiation, because the DynamicObject hasn't been named yet.
+          id.forceName(default = "_object", _namespace)
+          id.getReference.setRef(id.getRef)
+        }
         case id: BaseType =>
           if (id.isSynthesizable) {
             id.topBinding match {
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -7,6 +7,7 @@ import scala.collection.mutable.ArrayBuffer
 import chisel3._
 import chisel3.experimental._
 import chisel3.experimental.hierarchy.core.{Clone, ImportDefinitionAnnotation, Instance}
+import chisel3.properties.Class
 import chisel3.internal.firrtl._
 import chisel3.internal.naming._
 import _root_.firrtl.annotations.{CircuitName, ComponentName, IsMember, ModuleName, Named, ReferenceTarget}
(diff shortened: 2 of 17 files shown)


core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -29,16 +38,24 @@ private[chisel3] trait PropertyType[T] {
   * be in the implicit scope and available for users.
   */
 private[chisel3] object PropertyType {
+  @implicitAmbiguous("unable to infer Property type. Please specify it explicitly in square brackets on the LHS.")
   implicit val intPropertyTypeInstance = new PropertyType[Int] {
-    override def getPropertyType: ir.PropertyType = ir.IntegerPropertyType
+    override def getPropertyType(_value: Option[Int]): ir.PropertyType = ir.IntegerPropertyType
   }
 
+  @implicitAmbiguous("unable to infer Property type. Please specify it explicitly in square brackets on the LHS.")
```
jackkoenig, line 46: The need for these annotations I think is indicative of my point below about not liking the fact that the only user API here is untyped. These are only necessary because the `.selectDynamic` on `Object` returns `Property[T]` but there isn't really a `T`.

jackkoenig (commented):
This is a pretty large discussion topic but I am not a huge fan of the current untyped approach using dynamic. The entire purpose of Definition/Instance was to give this sort of API while maintaining type safety. I really think we should just bite the bullet and make D/I be the intended way of defining and instantiating Classes.

Now I understand that we have some use cases that require an unsafe API for just instantiating classes by name when you certainly are not going to know anything about the actual typed interface. But rather than making the normal API embrace this untyped route, I think we can just implement that particular API directly without having its untyped-ness impact the more typical user API

```scala
val inst: DynamicObject = Class.unsafeInstanceFromName("MyClassName"))
```

Depending on how hard it is to implement D/I for classes, we could punt on implementing it fully and just make `Definition[Class]` work, then use this unsafe API as the only API to instantiate an object for the short term.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -139,6 +140,12 @@ abstract class RawModule extends BaseModule {
         case id: assume.Assume    => id.forceName(default = "assume", _namespace)
         case id: cover.Cover      => id.forceName(default = "cover", _namespace)
         case id: printf.Printf => id.forceName(default = "printf", _namespace)
+        case id: DynamicObject => {
+          // Force name of the DynamicObject, and set its Property[ClassType] type's ref to the DynamicObject.
+          // The type's ref can't be set within instantiate, because the DynamicObject hasn't been named yet.
```
jackkoenig, line 145: I think this comment needs updating

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -682,6 +683,21 @@ private[chisel3] object Builder extends LazyLogging {
...
+    currentModule match {
+      case Some(module: RawModule) =>
+        aspectModule(module) match {
+          case Some(aspect: RawModule) => aspect
+          case other => module
+        }
+      case Some(cls: Class) => cls
+      case _ =>
+        throwException(
+          "Error: Not in a RawModule or Class. Likely cause: Missed Module() or Class() wrap, bare chisel API call, or attempting to construct hardware inside a BlackBox."
```
jackkoenig, line 696:
```suggestion
          "Error: Not in a RawModule or Class. Likely cause: Missed Module() or Definition() wrap, bare chisel API call, or attempting to construct hardware inside a BlackBox."
```

plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala
```diff
@@ -89,7 +89,8 @@ class ChiselComponent(val global: Global, arguments: ChiselPluginArguments)
       shouldMatchGen(
         tq"chisel3.BaseType",
         tq"chisel3.MemBase[_]",
-        tq"chisel3.VerificationStatement"
+        tq"chisel3.VerificationStatement",
+        tq"chisel3.properties.DynamicObject"
```
jackkoenig, line 93: A++, I was getting ready to point out that this is needed but you already figured it out 🙂

core/src/main/scala/chisel3/properties/Object.scala
```diff
@@ -0,0 +1,36 @@
...
+import scala.collection.immutable.HashMap
+
+/** Represents an instance of a Class.
+  *
+  * This cannot be instantiated directly, instead see Class.unsafeGetDynamicObject.
+  *
+  * The DynamicObject is generally unsafe, in that its getField method does not check the name, type, or direction of
+  * the accessed field. It may be used with care, and a more typesafe version will be added.
+  */
+case class DynamicObject private[chisel3] (className: String) extends HasId with NamedComponent {
```
jackkoenig, line 17:
```suggestion
class DynamicObject private[chisel3] (className: String) extends HasId with NamedComponent {
```
case class implies certain things that `HadId` does not follow.

jackkoenig (approved):

PR #3496, opened 2023-08-18
Support Optionality of implicit clock and reset

jackkoenig:
This is in preparation for the new LTL operators defaulting to using the current in scope clock (needs `Option[Clock]`) and making it easy to disable clock for a region of assertions.

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

* withClock, withReset, and withClockAndReset now have forms that take Option[Clock] and Option[Reset]
* Module.clockOption and Module.resetOption return Option[Clock] and Option[Reset]

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [ ] Did you mark the proper milestone (Bug fix: `3.5.x`, `3.6.x`, or `5.x` depending on impact, API modification or big change: `6.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3497, opened 2023-08-18
Add Disable, a new API for disabling simulation constructs

jackkoenig:
~Does not actually depend on https://github.com/chipsalliance/chisel/pull/3496, but I developed them together and they conflict so will rebase once that is merged.~ Done.

Disable will be used by a following PR which will make the new LTL properties use `Disable` instead of `Bool`.

I was really struggling with how to make the user API for this reasonable, and huge h/t to @azidar for the idea to make the disable setting API (`withDisable`) a _function_ of the current reset. I originally was thinking it would be its own value but this could lead to some really strange behavior like simulation-only ports getting added to modules (for feeding into assertions) and changing the reset not affecting the disable. I think this is a much better design point than I originally had in mind--the API is restricted but we can expand it as necessary and if we need to fundamentally change it, it's much easier with the current restrictions.

### Contributor Checklist

#### Type of Improvement

- Feature

#### Desired Merge Strategy

- Squash

#### Release Notes

Disable is a new type that represents the concept of disabling a simulation construct. It is handled similarly to the implicit clock and reset except Disable is notionally a function of the current implicit reset. Its default value is "not has been reset", a two-state simulation-only construct that is 0 when simulation starts and only becomes 1 after reset has been asserted, and then de-asserted. Because Disable has simulation-only semantics, it is not a `Data` and can only be used by simulation-only APIs.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Disable.scala
+++ b/core/src/main/scala/chisel3/Disable.scala
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.internal._
+import chisel3.experimental.{IntrinsicModule, OpaqueType, SourceInfo}
+import chisel3.internal.sourceinfo.SourceInfoTransform
+
+import scala.language.experimental.macros
+import scala.collection.immutable.ListMap
+
+/** API for handling disabling of simulation constructs
+  *
+  * Disables may be non-synthesizable so they can only be used for disabling simulation constructs
+  *
+  * The default disable is the "hasBeenReset" of the currently in scope reset.
+  * It can be set by the user via the [[withDisable]] API
+  *
+  * Users can access the current `Disable` with [[Module.disable]]
+  */
+// We could just an OpaqueType, but since OpaqueTypes have some API holes, this non-Data type
+class Disable private[chisel3] (private[chisel3] val value: Bool) {
+
+  /** Logical not
+    *
+    * @return invert the logical value of this `Disable`
+    * @group Bitwise
+    */
+  final def unary_! : Disable = macro SourceInfoTransform.noArg
+
+  /** @group SourceInfoTransformMacro */
+  def do_unary_!(implicit sourceInfo: SourceInfo): Disable = new Disable(!this.value)
+}
+
+object Disable {
+
+  sealed trait Type
+
+  /** Never disable
+    *
+    * The simulation construct will always be executing.
+    */
+  case object Never extends Type
+
+  /** Disable before reset has been seen
+    *
+    * Disable until reset has been asserted and then deasserted.
+    * Also requires
+    */
+  case object BeforeReset extends Type
+
+  private[chisel3] def withDisable[T](option: Type)(block: => T): T = {
+    // Save parentScope
+    val parentDisable = Builder.currentDisable
+
+    Builder.currentDisable = option
+
+    val res = block // execute block
+
+    // Return to old scope
+    Builder.currentDisable = parentDisable
+    res
+  }
+}
+
+/** Creates a new [[Disable]] scope */
+object withDisable {
+
+  /** Creates a new Disable scope
+    *
+    * @param disable an Optional new implicit Disable, None means no disable
+    * @param block the block of code to run with new implicit Disable
+    * @return the result of the block
+    */
+  def apply[T](disable: Disable.Type)(block: => T): T = Disable.withDisable(disable)(block)
+}
+
+/**
+  */
+// Note because this uses abstract reset, it cannot be instantiable until we have the ability to
+// create both sync and sync reset instances from the same Definition
+private[chisel3] class HasBeenResetIntrinsic(implicit sourceInfo: SourceInfo)
+    extends IntrinsicModule(f"circt_has_been_reset") {
+  // Compiler plugin does not run on core so we have to suggest names
+  val clock = IO(Input(Clock())).suggestName("clock")
+  val reset = IO(Input(Reset())).suggestName("reset")
+  val out = IO(Output(Bool())).suggestName("out")
+}
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -15,6 +15,7 @@ import chisel3.properties.Class
 import chisel3.reflect.DataMirror
 import _root_.firrtl.annotations.{IsModule, ModuleName, ModuleTarget}
 import _root_.firrtl.AnnotationSeq
+import chisel3.internal.plugin.autoNameRecursively
 
 object Module extends SourceInfoDoc {
 
@@ -41,6 +42,8 @@ object Module extends SourceInfoDoc {
     val parentWhenStack = Builder.whenStack
 
     // Save then clear clock and reset to prevent leaking scope, must be set again in the Module
+    // Note that Disable is a function of whatever the current reset is, so it does not need a port
+    //   and thus does not change when we cross module boundaries
     val (saveClock, saveReset) = (Builder.currentClock, Builder.currentReset)
     val savePrefix = Builder.getPrefix
     Builder.clearPrefix()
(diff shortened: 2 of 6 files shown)


jackkoenig:
There's a question I'm currently leaving unanswered as to whether we should make Disable apply to the existing `assert`, `assume` and `cover` APIs (as opposed to just the new ltl property APIs). Applying it blanketly with not hasbeenreset as the default would be dangerous as it would change semantics of user assertions, but there are other tricks we could play if we wanted to. Will bring up in the dev meeting.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -107,6 +110,32 @@ object Module extends SourceInfoDoc {
...
+  private def hasBeenReset(implicit sourceInfo: SourceInfo): Option[Disable] = {
+    // TODO memoize this
+    (Builder.currentClock, Builder.currentReset) match {
+      case (Some(clock), Some(reset)) =>
+        val hasBeenReset = Module(new HasBeenResetIntrinsic)
+        hasBeenReset.clock := clock
+        hasBeenReset.reset := reset
+        Some(new Disable(hasBeenReset.out))
+      case _ => None
+    }
```
seldridge, line 136: Either update this with the memoization or land it before this gets in to production.

core/src/main/scala/chisel3/Disable.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /** @group SourceInfoTransformMacro */
+  def do_unary_!(implicit sourceInfo: SourceInfo): Disable = new Disable(!this.value)
+}
+
+object Disable {
+
+  sealed trait Option
+  case object None extends Option
+  case object NotHasBeenReset extends Option
```
seldridge, line 39: The shadowing of `Option` isn't awesome. I'll usually use `sealed trait Type` for this pattern.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -107,6 +110,32 @@ object Module extends SourceInfoDoc {
...
+    *
+    * Note that [[Disable]] is a function of the implicit clock and reset
+    * so having no implicit clock or reset may imply no `Disable`.
+    */
+  def disable(implicit sourceInfo: SourceInfo): Option[Disable] = {
+    Builder.currentDisable match {
+      case Disable.None            => None
+      case Disable.NotHasBeenReset => hasBeenReset.map(x => autoNameRecursively("disable")(!x))
+    }
+  }
```
seldridge, line 123: The fact that this returns an `Option[Disable]` is a bit of a non-standard API for Chisel. I'm more used to APIs which throw exceptions. My question is, in what circumstances would the user not require that `disable: Some[Disable]`?

Analogously, it's similar to having `Reg(UInt(8.W))` return `Option[UInt]`. If a user is trying to instantiate a register, they likely expect to get a register and it's faster to error if they can't instantiate a register.

WDYT?

core/src/main/scala/chisel3/Disable.scala
```diff
@@ -0,0 +1,77 @@
...
+  *
+  * Disables may be non-synthesizable so they can only be used for disabling simulation constructs
+  *
+  * The default disable is the "hasBeenReset" of the currently in scope reset.
+  * It can be set by the user via the [[withDisable]] API
+  *
+  * Users can access the current `Disable` with [[Module.disable]]
+  */
+// We could just an OpaqueType, but since OpaqueTypes have some API holes, this non-Data type
+class Disable private[chisel3] (private[chisel3] val value: Bool) {
```
seldridge, line 22: How does this work for `AsyncReset`? (I see that it does in the tests!)

seldridge (commented):
This is looking good. I like the `withDisable` style API.

One question about the use of `Option` here. Some nits/questions.

core/src/main/scala/chisel3/Module.scala
jackkoenig: I added `Module.clockOption` and `Module.resetOption` in https://github.com/chipsalliance/chisel/pull/3496 because the LTL APIs accept `Option[Clock]` and `Option[Bool]` (soon to be `Option[Disable]`), so we at least need the optional version for use there. I think it's reasonable to make `Module.disable` be a version that returns `Disable` and asserts if not and we could add `Module.disableOption` for use by the LTL Property APIs.

core/src/main/scala/chisel3/Disable.scala
jackkoenig: I added `.asDisable` on `Reset` so you can cast `Bool`, `AsyncReset`, and `Reset` to Disable via `.asDisable`.

core/src/main/scala/chisel3/Disable.scala
jackkoenig: Yeah I was not sure how I felt about the names, I'll use `Type` instead of `Option`. Do you have an idea for better name than `None` since that shadowing isn't great either? I thought about `No`, could also do `NoDisable` but the normal use of this API will be `Disable.NoDisable` so that might be a bit much (but is also maybe okay).

core/src/main/scala/chisel3/Module.scala
jackkoenig: I'll do it in a quick follow on PR, working on it.

core/src/main/scala/chisel3/Disable.scala
```diff
@@ -0,0 +1,77 @@
...
+  */
+// Note because this uses abstract reset, it cannot be instantiable until we have the ability to
+// create both sync and sync reset instances from the same Definition
+private[chisel3] class HasBeenResetIntrinsic(implicit sourceInfo: SourceInfo)
+    extends IntrinsicModule(f"circt_has_been_reset") {
+  // Compiler plugin does not run on core so we have to suggest names
+  val clock = IO(Input(Clock())).suggestName("clock")
+  val reset = IO(Input(Reset())).suggestName("reset")
+  val out = IO(Output(Bool())).suggestName("out")
+}
```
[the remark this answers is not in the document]
jackkoenig: It can't be there because it needs to be used by `Module.disable` as the default disable. I agree it's a little weird to have it in core but I don't see any other way 🤷‍♀️.

core/src/main/scala/chisel3/Disable.scala
seldridge: You could use `Type` as a sealed trait and as an object. You then get `Type` as a package-like separator to avoid confusing around `Disable.NoDisable`. Just an idea:

```scala
sealed trait Type
object Type {
  case object NoDisable extends Type
  case object NotHasBeenReset extends Type
}
```

`NoDisable` seems okay. Also, `None` isn't quite as bad if wrapped in the `Type` object.

core/src/main/scala/chisel3/Module.scala
seldridge: For consistency, we can just make this `disableOption` and either add or wait until we need a `disable` that errors. Regardless, `disableOption` sounds good to me.

seldridge (approved):

seldridge:
Other thought... calling this `disable` results in negation like `notHasBeenReset`. Would this be better expressed as `enable` which would allow the type to be `hasBeenReset` and `always`? This may be more semantically easy to grok.

jackkoenig:
> Other thought... calling this disable results in negation like notHasBeenReset. Would this be better expressed as enable which would allow the type to be hasBeenReset and always? This may be more semantically easy to grok.

Maybe, the term @fabianschuiki used for the LTL operators is Disable which derives from SVA `disable iff (cond)`. I'm not against changing our terminology but we'll need to be consistent across the board. I guess we _could_ just use the term "enable" in Chisel while still using the disable intrinsic. What do you think @seldridge and @fabianschuiki?

core/src/main/scala/chisel3/Disable.scala
```diff
@@ -0,0 +1,77 @@
...
+import chisel3.internal._
+import chisel3.experimental.{IntrinsicModule, OpaqueType, SourceInfo}
+import chisel3.internal.sourceinfo.SourceInfoTransform
+
+import scala.language.experimental.macros
+import scala.collection.immutable.ListMap
+
+/** API for handling disabling of simulation constructs
+  *
+  * Disables may be non-synthesizable so they can only be used for disabling simulation constructs
```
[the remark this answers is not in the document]
jackkoenig: They are fair questions and use cases. At the moment, you can't do anything with it except pass it as an argument (implicitly or explicitly) to the new `chisel3.ltl.*Property` functions, eg. `AssertProperty`.

It's not so much that users necessarily should not be able to use the stuff more flexibly, but more that we don't currently have any other way of stopping them from using it to influence synthesizable hardware which would be bad (you may get a warning/error in synthesis but I'm not sure, but we'd like to protect users from accidentally simulating with it as well). Another possibility would be enforcing this property as a firtool check, but I think it would turn into a global analysis which we'd probably like to avoid if possible.

FWIW, power user library writers like yourself can instantiate the intrinsic directly and then you have the value exposed as a Bool, but I'm trying very hard to keep that aspect of it out of the standard user API, at least for now.

fabianschuiki:
Yeah the property API just reused the `disable iff` terminology that people would be familiar with from SVA land. Not sure if that was a good choice.

I like `BeforeReset` has an alternative to `NotHasBeenReset` :smiley: 

Regarding `Disable`: this feels like a monad that tags/poisons values to live in the simulation or verification realm. Maybe we'll eventually want more of these? Something like `Sim[Bool]` for simulation-only constructs, or `Verif[Bool]` for verification-only constructs. For example, I could see us shoving a lot of verification-only metadata into AXI or TileLink interfaces, to make checking invariants or formally proving things easier. With AXI, you could carry the entire AXI transaction along the bus, and then just make sure that the current beat reflects the appropriate fragment of the entire transaction. That might allow us to more aggressively formally verify things in the future.

core/src/main/scala/chisel3/Disable.scala
```diff
@@ -0,0 +1,77 @@
...
+  *
+  * The default disable is the "hasBeenReset" of the currently in scope reset.
+  * It can be set by the user via the [[withDisable]] API
+  *
+  * Users can access the current `Disable` with [[Module.disable]]
+  */
+// We could just an OpaqueType, but since OpaqueTypes have some API holes, this non-Data type
```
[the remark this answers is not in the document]
jackkoenig: It works correctly.

core/src/main/scala/chisel3/Module.scala
jackkoenig: The memoization of the instance is one thing we need, we also need to use D/I on the Intrinsic definitions. There is a bug in ScalaDoc that is making this difficult to do but will handle it in https://github.com/chipsalliance/chisel/pull/3499.

core/src/main/scala/chisel3/Module.scala
jackkoenig: Renamed `disable` to `disableOption` and added a form `disable` that asserts out if no implicit disable is defined.

core/src/main/scala/chisel3/Disable.scala
jackkoenig: Using `Type`, `Never`, and `BeforeReset` (great suggestion on that one!)

PR #3500, opened 2023-08-22
add DedupGroupAnnotation

albertchen-sifive:
add annotation case class for [DedupGroupAnnotation](https://github.com/llvm/circt/blob/main/docs/Dialects/FIRRTL/FIRRTLAnnotations.md#dedupgroupannotation) (introduced in firtool 1.51.0)

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/firrtl/src/main/scala/firrtl/transforms/Dedup.scala
+++ b/firrtl/src/main/scala/firrtl/transforms/Dedup.scala
@@ -9,3 +9,8 @@ import firrtl.annotations._
 case class NoDedupAnnotation(target: ModuleTarget) extends SingleTargetAnnotation[ModuleTarget] {
   def duplicate(n: ModuleTarget): NoDedupAnnotation = NoDedupAnnotation(n)
 }
+
+/** Assign the targeted module to a dedup group. Only modules in the same group may be deduplicated. */
+case class DedupGroupAnnotation(target: ModuleTarget, group: String) extends SingleTargetAnnotation[ModuleTarget] {
+  def duplicate(n: ModuleTarget): DedupGroupAnnotation = DedupGroupAnnotation(n, group)
+}
--- a/src/test/scala/chiselTests/DedupSpec.scala
+++ b/src/test/scala/chiselTests/DedupSpec.scala
@@ -84,15 +84,29 @@ class DedupSpec extends ChiselFlatSpec {
     assert(countModules(compile { new SharedConstantValDedupTop }) === 2)
   }
 
+  def dedupGroup(m: Module, group: String): Unit = annotate(new ChiselAnnotation {
+    def toFirrtl = DedupGroupAnnotation(m.toTarget, group)
+  })
+
   it should "not dedup modules that are in different dedup groups" in {
-    def dedupGroup(m: Module, group: String): Unit = annotate(new ChiselAnnotation {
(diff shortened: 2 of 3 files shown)


firrtl/src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -9,3 +9,8 @@ import firrtl.annotations._
 case class NoDedupAnnotation(target: ModuleTarget) extends SingleTargetAnnotation[ModuleTarget] {
   def duplicate(n: ModuleTarget): NoDedupAnnotation = NoDedupAnnotation(n)
 }
+
+/** Assign the targeted module to a dedup group. Only modules in the same group may be deduplicated. */
+case class DedupGroupAnnotation(target: ModuleTarget, group: String) extends SingleTargetAnnotation[ModuleTarget] {
+  def duplicate(n: ModuleTarget): DedupGroupAnnotation = DedupGroupAnnotation(n, group)
+}
```
jackkoenig, line 16: Is this how we want to expose this or should we try changing Chisel to just emit this annotation for all Modules based on the `desiredName` of the class?

src/test/scala/chiselTests/DedupSpec.scala
```diff
@@ -81,4 +83,16 @@ class DedupSpec extends ChiselFlatSpec {
   it should "dedup modules that share a literal" in {
     assert(countModules(compile { new SharedConstantValDedupTop }) === 2)
   }
+
+  it should "not dedup modules that are in different dedup groups" in {
```
mwachs5, line 87: is it supposed to be an error if you try to put a module into different dedup groups? should we show what happens if you:
      dedupGroup(top.inst0, "inst0")
      dedupGroup(top.inst0, "anothergroup")
      dedupGroup(top.inst1, "inst1")
      dedupGroup(top.inst1, "anothergroup")

albertchen-sifive: yes it should be an error, I added a test case

firrtl/src/main/scala/firrtl/transforms/Dedup.scala
jackkoenig: I think typically we want a higher-level API than manually finding modules and group them via an aspect. I will grant that the typical use will probably want more information than just desiredName, but we should probably be integrating this information in semantically useful ways, eg. a first class prefixing API which is the stand-in for hierarchical location, and automatically generating desiredName from the parameters as a way of integrating those, etc. In any case, if we want this to be an API the user can call directly, it should be a function that takes a module, not the annotation directly since we are moving away from annotations in general.

mwachs5: agree that the function is what we should expose not the annotation.

I'm fine with making this a user API for now vs assuming its the desiredName that would be the grouping mechanism

jackkoenig: Alright, I think this is worth merging and we can come up with nicer default behavior in the future. It's useful enough as is.

jackkoenig (approved):

PR #3504, opened 2023-08-23
Implement auto typeName generation for Records

jared-barocsi:
Adds the `HasAutoTypename` trait, which mixes into `Record` and generates a `def typeName` statement constructed from the `Record`'s constructor parameters.

Example:
```
class MyBundle(foo: Int, bar: Int) extends Bundle with HasAutoTypename {
  val x = UInt(foo.W)
  val x = UInt(foo.W)
  // This line is automatically generated by the compiler plugin through HasAutoTypename
  // override def typeName: String = s"${getClass.getSimpleName}_$foo_$bar"
}

val bundle = Wire(new MyBundle(8, 4))
bundle.typeName // MyBundle_8_4
```

### Contributor Checklist

#### Type of Improvement

- New API

#### Desired Merge Strategy

Squash and merge

#### Release Notes
Add experimental `HasAutoTypename` trait
- Allows users to automatically generate a `typeName` for `Bundle` objects through the compiler plugin.

- [ ] Did you review?


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -947,6 +947,14 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record extends Aggregate {
 
+  /** The list of parameter accessors used in the constructor of this [[chisel3.Record]].
+    * @note This is automatically overrided via the compiler plugin for user-defined bundles that mix-in [[chisel3.experimental.HasAutoTypename]],
+    *       and is meant for internal Chisel use only. Can not be manually overrided by users, or else an error will be thrown.
```
mwachs5, line 952:
```suggestion
    * @note This is automatically overridden via the compiler plugin for user-defined bundles that mix-in [[chisel3.experimental.HasAutoTypename]],
    *       and is meant for internal Chisel use only. Can not be manually overridden by users, or else an error will be thrown.
```

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -189,6 +194,49 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
       elementsImpl
     }
 
+    def generateAutoTypename(bundle: ClassDef, thiz: global.This): Option[Tree] = {
+      // 1: Get immediate defined args of the Bundle
+      // 2: Determine an appropriate naming strategy based on the collection of elements:
+      //    - 1 numeric property only: Generate a name like MyBundle2
+      //    - Multiple numeric properties: Generate a name with word-like descriptors like MyBundle_${width}width
+      //    - Mixed elements: Numeric properties first, type names of Aggregate-likes after, like MyBundle_${width}property_${childBundle.typeName}
```
mwachs5, line 202:
```suggestion
      //    - Multiple numeric properties: Generate a name with word-like descriptors like MyBundle_${width}width
      //    - Mixed elements: Numeric properties first, type names of Aggregate-likes after, like MyBundle_${width}property_${childBundle.typeName}
```

Is it really putting the word `width` and `property` in there? Isn't the example just showing it with the number there?

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -189,6 +194,49 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
       elementsImpl
     }
 
+    def generateAutoTypename(bundle: ClassDef, thiz: global.This): Option[Tree] = {
+      // 1: Get immediate defined args of the Bundle
+      // 2: Determine an appropriate naming strategy based on the collection of elements:
+      //    - 1 numeric property only: Generate a name like MyBundle2
```
mwachs5, line 200:
```suggestion
      //    - 1 numeric property only: Generate a name like MyBundle2
```

don't the test examples show this is still like `MyBundle_2`?

jared-barocsi: Sorry, the comments are outdated, I need to update them

core/src/main/scala/chisel3/experimental/HasAutoTypename.scala
```diff
@@ -0,0 +1,53 @@
...
+  * (new MyBundle(3, 4)(SInt(3.W))).typeName // "MyBundle_3_4_SInt3"
+  * (new MyBundle(1, 32)(Bool())).typeName   // "MyBundle_1_32_Bool"
+  * }}}
+  */
+trait HasAutoTypename {
+  this: Record =>
+
+  /** Auto generate a type name for this Bundle using the bundle arguments supplied by the compiler plugin.
+    */
+  override def typeName: String = autoTypeName(getClass.getSimpleName, _typeNameConParams)
```
jackkoenig, line 40: We should probably think about what the behavior should be for anonymous classes, eg. what if someone writes:
```scala
class MyModule(n: Int) extends Module {
  val io = IO(new Bundle with HasAutoTypename {
    val foo = Input(UInt(n.W))
  })
}
```
I think you will get a name that is something like `_packageMyModulee0e14f5_8` which is a pretty terrible name indeed.

I think there's a compelling argument that you should not be able to mix HasAutoTypename into anonymous classes since it's going to be a disaster, and arguably not any inner class (inner classes have their outer object as the first argument, that's where `packageMyModuleee0e145f` comes from, that's the `.toString` of the instance of `MyModule` above).

Also `.getSimpleName` throws an exception in some circumstances on Java8 which is still supported by Chisel, see https://github.com/scala/bug/issues/2034 and this is why we have more complicated logic for naming Modules: https://github.com/chipsalliance/chisel/blob/8b60f1947d257ad8cdf3e5dcf72e459be868e18a/core/src/main/scala/chisel3/Module.scala#L507

I'm not going to sandbag this PR of this issue, but more careful handling either needs to be in this PR or come shortly after.

docs/src/cookbooks/naming.md
```diff
@@ -151,6 +151,42 @@ The suggested pattern for `typeName`, and subsequently `desiredName`, is to fold
...
+Auto-generated `typeName`s take the form of `{Bundle Name}_{Parameter Value 1}_{Parameter Value 2}_{...}`, and so our `MyBundle` can be equivalently expressed with:
+```scala mdoc:silent
+import chisel3.experimental.HasAutoTypename
+class MyBundle[T <: Data](gen: T, intParam: Int) extends Bundle with HasAutoTypename {
+  // ...
+  // Note: No `override def typeName` statement here
+}
+
+new MyBundle(UInt(8.W), 3).typeName // == "MyBundle_UInt8_3"
+```
```
jackkoenig, line 189:
````suggestion
```scala mdoc
import chisel3.experimental.HasAutoTypename
class MyBundle[T <: Data](gen: T, intParam: Int) extends Bundle with HasAutoTypename {
  // ...
  // Note: No `override def typeName` statement here
}

new MyBundle(UInt(8.W), 3).typeName
```
````
What happens if you just use `mdoc` here? Part of the point of mdoc is that it should evaluate the code and show the value automatically.

File: plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala [hunk 6]
```diff
@@ -189,6 +194,43 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
+      val typeNameConArgs: List[(Tree)] =
+        constructor.vparamss.flatMap(_.map { vp =>
+          val p = paramLookup(vp.name.toString)
+          // Make this.<ref>
+          val select = gen.mkAttributedSelect(thiz.asInstanceOf[Tree], p)
+          // Clone any Data parameters to avoid field aliasing, need full clone to include direction
+          val cloned = if (isData(vp.symbol)) cloneTypeFull(select.asInstanceOf[Tree]) else select
+          // Need to splat varargs
+          if (isVarArgs(vp.symbol)) q"$cloned: _*" else cloned
+        })
```
jackkoenig, line 219: I think all of this is basically copy-pasted from `generateAutoCloneType`, please figure out how to refactor the code such that the code is not duplicated and so that it only runs once.

jackkoenig (commented):
Generally looks good but please remove the code duplication in the plugin. Per anonymous Bundle issues, we can deal with that in a follow on PR.

core/src/main/scala/chisel3/experimental/HasAutoTypename.scala
jackkoenig: We chatted offline but I think the best path forward is to just find a way to ban inner classes from using `HasAutoTypename` in the compiler plugin, then we don't need to worry about the `.getSimpleName` issue either because that only affects inner classes.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -951,6 +951,14 @@ trait VecLike[T <: Data] extends IndexedSeq[T] with HasId with SourceInfoDoc {
   */
 abstract class Record extends Aggregate {
 
+  /** The list of parameter accessors used in the constructor of this [[chisel3.Record]].
+    * @note This is automatically overridden via the compiler plugin for user-defined bundles that mix-in [[chisel3.experimental.HasAutoTypename]],
+    *       and is meant for internal Chisel use only. Can not be manually overridden by users, or else an error will be thrown.
+    * @note This lives in Record rather than the [[chisel3.experimental.HasAutoTypename]] trait, due to compiler implementation details
+    *       preventing us from overriding a definition within a trait via the compiler plugin
+    */
```
mwachs5, line 959:
```suggestion
  /** The list of parameter accessors used in the constructor of this [[chisel3.Record]].
    *
    * @note This is automatically overridden via the compiler plugin for user-defined bundles that mix-in [[chisel3.experimental.HasAutoTypename]],
    *       and is meant for internal Chisel use only. Can not be manually overridden by users, or else an error will be thrown.
    * @note This lives in Record rather than the [[chisel3.experimental.HasAutoTypename]] trait, due to compiler implementation details
    *       preventing us from overriding a definition within a trait via the compiler plugin
    */
```

mega-nit to add a whitespace after first scaladoc line

src/test/scala/chiselTests/experimental/AutoTypenameSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+  val foo = UInt(width.W)
+}
+
+class DataBundle(data: Data) extends Bundle with HasAutoTypename {
+  val foo = data
+}
+
+class BundleBundle(gen: HasAutoTypename) extends Bundle with HasAutoTypename {
+  val foo = gen
+}
```
jackkoenig, line 29: Please add a test for a Bundle with multiple parameter lists, including implicit eg.
```
class MultipleParameterLists(x: Int)(y: List[Int])(implicit z: String)`

jackkoenig (approved):
Approved but please add the requested test and address Megan's comments about the ScalaDoc

PR #3505, opened 2023-08-24
Add Map properties

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
This PR changes the `Property.apply` from `def apply[T: PropertyType](): Property[T]` to `def apply[T]()(implicit tpe: PropertyType[T]): Property[tpe.Type]`. This lets implicit resolution resolve the result type, instead of having to add different apply methods for all the combinations of `T` and `Property[T]` alternatives - `Seq[Property[T]]`/`Seq[T]`, `Map[String, Property[T]]`/`Map[String, T]`, `(Property[A], Property[B])`/`(A, B)` etc.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -142,7 +142,7 @@ private[chisel3] object PropertyType {
   * describe a set of non-hardware types, so they have no width, cannot be used
   * in aggregate Data types, and cannot be connected to Data types.
   */
-sealed abstract class Property[T] extends BaseType { self =>
+abstract class Property[T] extends BaseType { self =>
   protected type TT
   protected val tpe: PropertyType[TT]
   protected def value: Option[TT]
@@ -250,27 +250,25 @@ sealed abstract class Property[T] extends BaseType { self =>
   */
 object Property {
 
+  private[chisel3] def makeWithValueOpt[T](valueOpt: Option[T])(implicit _tpe: PropertyType[T]): Property[_tpe.Type] = {
+    new Property[_tpe.Type] {
+      type TT = T
+      val tpe = _tpe
+      val value = valueOpt
+    }
+  }
+
   /** Create a new Property based on the type T.
     */
   def apply[T]()(implicit tpe: PropertyType[T]): Property[tpe.Type] = {
-    val _tpe = tpe
-    new Property[tpe.Type] {
-      type TT = T
-      val tpe = _tpe
-      val value = None
-    }
+    makeWithValueOpt(None)(tpe)
   }
 
   /** Create a new Property literal of type T.
     */
   def apply[T](lit: T)(implicit tpe: PropertyType[T]): Property[tpe.Type] = {
     val literal = ir.PropertyLit[tpe.Type, tpe.Underlying](tpe, tpe.convertUnderlying(lit))
-    val _tpe = tpe
-    val result = new Property[tpe.Type] {
-      type TT = T
-      val tpe = _tpe
-      val value = None
-    }
+    val result = makeWithValueOpt(None)(tpe)
     literal.bindLitArg(result)
   }
 }
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -20,6 +20,7 @@ import chisel3.internal.{firrtl => ir}
 import chisel3.experimental.{prefix, requireIsHardware, SourceInfo}
 import scala.reflect.runtime.universe.{typeOf, TypeTag}
 import scala.annotation.{implicitAmbiguous, implicitNotFound}
+import scala.collection.immutable.SeqMap
 
 /** PropertyType defines a typeclass for valid Property types.
   *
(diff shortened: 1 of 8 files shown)


jackkoenig:
I'm going to mark this as a `Feature` not-with-standing the binary incompatible change to `Property.apply` since these Property APIs have never been released (so from previous release 6.0.0-M2 this is just a new Feature).

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -40,38 +67,72 @@ private[chisel3] trait PropertyType[T] {
...
+      override def getPropertyType(value: Option[F[A]]): fir.PropertyType =
+        fir.MapPropertyType(implicitly[PropertyType[A]].getPropertyType(None))
+      type Underlying = Map[String, tpe.Underlying]
+      override def convert(value: Underlying, ctx: ir.Component, info: SourceInfo): fir.Expression =
+        fir.MapPropertyValue(
+          tpe.getPropertyType(None),
+          value.map {
+            case (key, value) =>
+              key -> tpe.convert(value, ctx, info)
+          }.toSeq.sortBy(_._1)
```
jackkoenig, line 131: Do we actually want to be sorting here vs. just disallowing unordered maps? I think people generally want things to match the order they defined them rather than being sorted.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -40,38 +67,72 @@ private[chisel3] trait PropertyType[T] {
...
+    override def convertUnderlying(value: F[A]) =
+      value.map(tpe.convertUnderlying(_))
   }
+
+  implicit def mapPropertyTypeInstance[A, F[A] <: Map[String, A]](implicit tpe: PropertyType[A]) =
+    new PropertyType[F[A]] {
+      type Type = F[tpe.Type]
+      override def getPropertyType(value: Option[F[A]]): fir.PropertyType =
+        fir.MapPropertyType(implicitly[PropertyType[A]].getPropertyType(None))
+      type Underlying = Map[String, tpe.Underlying]
```
jackkoenig, line 124:
```suggestion
      type Underlying = VectorMap[String, tpe.Underlying]
```
We definitely should use an ordered map here.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -40,38 +67,72 @@ private[chisel3] trait PropertyType[T] {
...
 
-  implicit val boolPropertyTypeInstance = new PropertyType[Boolean] {
-    override def getPropertyType(value: Option[Boolean]): ir.PropertyType = ir.BooleanPropertyType
+  implicit def propertyTypeInstance[T](implicit pte: PropertyType[T]) = new PropertyType[Property[T]] {
+    type Type = T
+    override def getPropertyType(value: Option[Property[T]]): fir.PropertyType = pte.getPropertyType(None)
+    override def convert(value:         Underlying, ctx: ir.Component, info: SourceInfo): fir.Expression =
+      ir.Converter.convert(value, ctx, info)
+    type Underlying = ir.Arg
+    override def convertUnderlying(value: Property[T]) = value.ref
```
jackkoenig, line 104: I get why this is here, but I think we should think about if this is the only way to accomplish the goal of making `Property.apply` work for the likes of both `Property[Seq[BigInt]]` and `Seq[Property[BigInt]]`. It using overloading so bad since we only have the 2 collection types? I think we will just need 3 in that case.

What worries me about this is it makes it legal to write things like `Property[Seq[Property[Property[Property[BigInt]]]]]` and I'm not sure what that means.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -40,38 +67,72 @@ private[chisel3] trait PropertyType[T] {
...
+      fir.SequencePropertyType(implicitly[PropertyType[A]].getPropertyType(None))
+
+    type Underlying = Seq[tpe.Underlying]
+    override def convert(value: Underlying, ctx: ir.Component, info: SourceInfo): fir.Expression =
+      fir.SequencePropertyValue(tpe.getPropertyType(None), value.map(tpe.convert(_, ctx, info)))
+    override def convertUnderlying(value: F[A]) =
+      value.map(tpe.convertUnderlying(_))
   }
+
+  implicit def mapPropertyTypeInstance[A, F[A] <: Map[String, A]](implicit tpe: PropertyType[A]) =
```
jackkoenig, line 119:
```suggestion
  implicit def mapPropertyTypeInstance[A, F[A] <: SeqMap[String, A]](implicit tpe: PropertyType[A]) =
```
Due to determinism issues, let's just not support `Map` and do `SeqMap` instead (using `VectorMap` as our concrete implementation)

core/src/main/scala/chisel3/properties/Property.scala
jackkoenig: I don't think we should sandbag on this because overall this PR is a amazing and a huge step forward.
Please leave a comment explaining why this is here with maybe a TODO to see if there's either a way to remove the need for `PropertyType[Property[T]]` or at least disallow weird `Property[Property[T]]` or better yet disallow mutually recursive types like `Property[Seq[Property[Seq[Property[T]]]]]` (ideally disallow more than 1 `Property[_]` in the type if at all possible).

We can try to "fix" this later, let's definitely get this PR in.

jackkoenig (approved):
Please change from `<: Map` to `<: SeqMap` before merging, also take a look at removing the need (or otherwise mitigating) `PropertyType[Property[T]]`, if it's too hard, go ahead and merge and we'll revisit that later.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -51,22 +51,63 @@ private[chisel3] trait PropertyType[T] {
   def convertUnderlying(value: T): Underlying
 }
 
+private[chisel3] trait RecursivePropertyType[T] extends PropertyType[T]
```
jackkoenig, line 55: A comment describing what this is doing would be good.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -51,22 +51,63 @@ private[chisel3] trait PropertyType[T] {
...
+    value.view.mapValues(tpe.convertUnderlying(_)).toMap
+}
+
+private[chisel3] trait LowPriorityPropertyTypeInstances {
+  implicit def sequencePropertyTypeInstance[A, F[A] <: Seq[A]](implicit tpe: RecursivePropertyType[A]) =
+    new SeqPropertyType[A, F, tpe.type](tpe) with RecursivePropertyType[F[A]]
+
+  implicit def mapPropertyTypeInstance[A, F[A] <: Map[String, A]](implicit tpe: RecursivePropertyType[A]) =
+    new MapPropertyType[A, F, tpe.type](tpe) with RecursivePropertyType[F[A]]
+}
```
jackkoenig: A comment describing why this is necessary would be helpful

PR #3507, opened 2023-08-25
[main] Enable MiMa for v6.0.0-M3

chiselbot:
Enable MiMa for v6.0.0-M3


jackkoenig:
We don't care about binary compatibility until RC1.

PR #3511, opened 2023-08-28
add Path property type

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
Adds new public `PathType` class that represents a path to a hardware instance or member. Adds `PathPropertyType`/`PathPropertyLiteral` and `PropertyType` instances for module/data/mem.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/PathType.scala
+++ b/core/src/main/scala/chisel3/properties/PathType.scala
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.properties
+
+import chisel3.{Data, MemBase}
+import chisel3.experimental.BaseModule
+import firrtl.annotations.{InstanceTarget, IsMember, ModuleTarget, ReferenceTarget}
+import firrtl.ir.{PathPropertyLiteral}
+
+/** Represent a Path type for referencing a hardware instance or member in a Property[PathType]
+  */
+sealed abstract class PathType {
+  private[chisel3] def toTarget(): IsMember
+
+  private[chisel3] def convert(): PathPropertyLiteral = {
+    val target = toTarget()
+    val targetType = target match {
+      case _: ModuleTarget    => "OMModuleTarget"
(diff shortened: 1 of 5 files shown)


core/src/main/scala/chisel3/properties/PathType.scala
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  /** Construct a PathType that refers to a Data
+    */
+  def apply(data: Data): PathType = {
+    new PathType {
+      def toTarget(): IsMember = data.toAbsoluteTarget
+    }
+  }
```
jackkoenig, line 42: I'm going to approve this as is, but we really validate at this point if the `Data` can have a target. I'm mainly thinking about things like `PathType(3.U)` which will result an an exception thrown later when `.toAbsoluteTarget` is actually called.

core/src/main/scala/chisel3/properties/PathType.scala
```diff
@@ -0,0 +1,51 @@
...
+package chisel3.properties
+
+import chisel3.{Data, MemBase}
+import chisel3.experimental.BaseModule
+import firrtl.annotations.{InstanceTarget, IsMember, ModuleTarget, ReferenceTarget}
+import firrtl.ir.{PathPropertyLiteral}
+
+/** Represent a Path type for referencing a hardware instance or member in a Property[PathType]
+  */
+sealed abstract class PathType {
```
jackkoenig, line 12: Should this be `PathType` or perhaps just `Path`? It's `Path` in firrtl.

jackkoenig (approved):
Implementation LGTM, just wondering if we should be calling this `Path` instead of `PathType`.

core/src/main/scala/chisel3/properties/PathType.scala
albertchen-sifive: renamed to `Path`

sequencer:
Interesting. Maybe this is the real replacement to `TraceAPI`

PR #3513, opened 2023-08-29
add tuple properties

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
This PR adds `TuplePropertyValue`/`TuplePropertyType` as well as generated `PropertyType` classes and instances for tuples of arity 1-22 using the `project/Boilerplate.scala` sbt module. Since I expect there to be some churn in the property API and because it is very tedious and error-prone to write these by hand I chose to add a sbt  task to generate these classes and instances (based on [shapeless's version](https://github.com/milessabin/shapeless/blob/80d26bc132139eb887fe1e8d39abbdb8dcb20117/project/Boilerplate.scala)). We may want to consider checking these generated files in for observability.

#### Desired Merge Strategy

- [ ] Did you review?


project/Boilerplate.sc
albertchen-sifive, line 1: symlinked this file, because I couldn't get mill to import a `.scala` file

sequencer:
Is possible to use scala plugin to resolve this?

project/Boilerplate.scala
```diff
@@ -0,0 +1,192 @@
...
+
+import scala.collection.mutable.ArrayBuffer
+
+/** This is a simplified version of https://github.com/milessabin/shapeless/blob/80d26bc132139eb887fe1e8d39abbdb8dcb20117/project/Boilerplate.scala
+  *
+  * Generate a range of boilerplate classes, those offering alternatives with 0-22 params
+  * and would be tedious to craft by hand
+  */
+object Boilerplate {
+
```
jackkoenig, line 11: I'm not super keen on adding a code generation step unless it's something we're going to need to update a lot. I have a feeling this is something we'll only need to update sparingly so I'm inclined to instead say we just write this is a scala-cli script, run it once, and commit the generated file. We can include the source code for the scala-cli script in this PR so that if we need to update the file we can. But my current opinion is that we should not include this code gen step unless we find ourselves needing to update this a lot.

jackkoenig (approved):
LGTM but I would much prefer not having the code gen as part of the build, I think we should just commit the generated file.

build.sbt
```diff
@@ -314,6 +314,15 @@ lazy val core = (project in file("core"))
   .dependsOn(macros)
   .dependsOn(firrtl)
 
+lazy val generateBoilerplate = taskKey[Unit]("generate boilerplate code for the core project")
```
albertchen-sifive, line 317: @jackkoenig is it alright if I make this a standalone sbt task that you run manually? since developers probably already have sbt installed, and that way we can write files directly into the scala directory without requiring cli args

jackkoenig:
You know, I didn't realize the generated file would end up being almost 5000 lines of Scala lol, I think I'll switch it back to your original code gen and merge this.

PR #3514, opened 2023-08-30
Property now extends Data, remove BaseType

jackkoenig:
Currently a draft PR because there is more to do:
- [x] Merge some of the code in `Property` back into `Data`
- [x] Rip out `BaseType`
- [x] Make Bundles and Vecs containing Properties work
- [x] Figure out what to do with connections containing Properties

### Contributor Checklist

#### Type of Improvement

- API modification

A new entry in the `Data` subhierarchy is a pretty substantial change that will affect users.

#### Desired Merge Strategy

- Squash

#### Release Notes

Properties can be fields of Bundles and used in wires.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -4,6 +4,7 @@ import chisel3.internal.requireIsChiselType // Fix ambiguous import
 import chisel3.internal.{throwException, Builder}
 import chisel3.experimental.SourceInfo
 import chisel3.properties.Class
+import chisel3.properties.Property
 
 object IO {
 
@@ -32,8 +33,8 @@ object IO {
     module match {
       case _: Class => {
         data match {
-          case _: Data => throwException(s"Class ports must be Property type, but found $data")
-          case _ => ()
+          case _: Property[_] => ()
+          case _ => throwException(s"Class ports must be Property type, but found $data")
         }
       }
       case _ => ()
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -18,11 +18,11 @@ object IO {
     *
     * The granted iodef must be a chisel type and not be bound to hardware.
     *
-    * Also registers a BaseType as a port, also performing bindings. Cannot be called once ports are
+    * Also registers a Data as a port, also performing bindings. Cannot be called once ports are
     * requested (so that all calls to ports will return the same information).
     * Internal API.
     */
-  def apply[T <: BaseType](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
+  def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
     val module = Module.currentModule.get // Impossible to fail
     require(!module.isClosed, "Can't add more ports after module close")
     val prevId = Builder.idGen.value
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -34,7 +34,7 @@ object IO {
       case _: Class => {
         data match {
           case _: Property[_] => ()
-          case _ => throwException(s"Class ports must be Property type, but found $data")
+          case _ => Builder.error(s"Class ports must be Property type, but found ${data._localErrorContext}.")
         }
       }
       case _ => ()
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -3,8 +3,7 @@ package chisel3
 import chisel3.internal.requireIsChiselType // Fix ambiguous import
 import chisel3.internal.{throwException, Builder}
 import chisel3.experimental.SourceInfo
-import chisel3.properties.Class
-import chisel3.properties.Property
+import chisel3.properties.{Class, Property}
 
 object IO {
 
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -256,7 +256,7 @@ private[chisel3] object BiConnect {
     sink:        Data,
     source:      Data,
     sourceInfo:  SourceInfo,
-    context_mod: RawModule
+    context_mod: BaseModule
   ): Boolean = {
 
     // check that the aggregates have the same types
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -5,12 +5,12 @@ package chisel3.internal
 import chisel3._
 import chisel3.experimental.dataview.{isView, reify, reifyToAggregate}
 import chisel3.experimental.{attach, Analog, BaseModule, SourceInfo}
+import chisel3.properties.Property
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl.{Connect, Converter, DefInvalid}
 
 import scala.language.experimental.macros
 import _root_.firrtl.passes.CheckTypes
-import chisel3.properties.Property
 
 /**
   * BiConnect.connect executes a bidirectional connection element-wise.
--- a/core/src/main/scala/chisel3/internal/BiConnect.scala
+++ b/core/src/main/scala/chisel3/internal/BiConnect.scala
@@ -115,7 +115,7 @@ private[chisel3] object BiConnect {
       }
       case (left_p: Property[_], right_p: Property[_]) =>
         // We can consider supporting this, but it's a lot of code and we should be moving away from <> anyway
-        Builder.error(s"<> is not supported by Properties, use :<>= instead")(sourceInfo)
+        Builder.error(s"${left_p._localErrorContext} does not support <>, use :<>= instead")(sourceInfo)
       // Handle Vec case
       case (left_v: Vec[Data @unchecked], right_v: Vec[Data @unchecked]) => {
         if (left_v.length != right_v.length) {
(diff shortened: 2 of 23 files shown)


src/main/scala/chisel3/stage/phases/Elaborate.scala
```diff
@@ -44,7 +44,7 @@ class Elaborate extends Phase {
          */
         case scala.util.control.NonFatal(a) =>
           if (!chiselOptions.printFullStackTrace) {
-            a.trimStackTraceToUserCode()
+            // a.trimStackTraceToUserCode()
```
jackkoenig, line 47: TODO back this out before merging, useful for debugging 😅

core/src/main/scala/chisel3/IO.scala
```diff
@@ -32,8 +33,8 @@ object IO {
     module match {
       case _: Class => {
         data match {
-          case _: Data => throwException(s"Class ports must be Property type, but found $data")
-          case _ => ()
+          case _: Property[_] => ()
+          case _ => throwException(s"Class ports must be Property type, but found $data")
```
azidar, line 37: Since we are here, can we throw an non-blocking error?

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -6,9 +6,9 @@ import chisel3._
 import chisel3.experimental.{Analog, BaseModule, SourceInfo}
 import chisel3.internal.containsProbe
 import chisel3.internal.Builder.pushCommand
-import chisel3.internal.firrtl.{Connect, Converter, DefInvalid}
+import chisel3.internal.firrtl.{Connect, Converter, DefInvalid, PropAssign}
 import chisel3.experimental.dataview.{isView, reify, reifyToAggregate}
-import chisel3.properties.Property
+import chisel3.properties.{Class, Property}
```
azidar, line 11: I don't know if this is the place to nit, but `Class` is a perhaps undesirable name, even if it is descriptive, just for its meaning as Reflection of a java Class.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -99,7 +99,7 @@ private[chisel3] object MonoConnect {
     sourceInfo:  SourceInfo,
     sink:        Data,
     source:      Data,
-    context_mod: RawModule
+    context_mod: BaseModule
```
azidar, line 102: What's the purpose for this change? I'm fine with it, just curious.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -175,7 +175,7 @@ case class SLit(n: BigInt, w: Width) extends LitArg(n, w) {
   * These are not LitArgs, because not all property literals are integers.
   */
 private[chisel3] case class PropertyLit[T, U](
-  propertyType: PropertyTypeclass[_] { type Underlying = U; type Type = T },
+  propertyType: PropertyTypeclass[T] { type Underlying = U },
```
azidar, line 178: What is this change for? I'm confused by it.

src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -213,13 +213,14 @@ class PropertySpec extends ChiselFlatSpec with MatchesAndOmits {
   }
 
   it should "fail to compile when connecting Property types of different types" in {
```
azidar, line 215: Fix test name, in that it should pass?

core/src/main/scala/chisel3/IO.scala
jackkoenig: Non-blocking meaning recoverable so that we can aggregate more errors? Sure.

core/src/main/scala/chisel3/internal/MonoConnect.scala
jackkoenig: It's worth thinking about but it's a separate discussion from this PR. Lets discuss changing it offline.

core/src/main/scala/chisel3/internal/MonoConnect.scala
jackkoenig: Class extends BaseModule and connections between Properties are legal within Classes.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -307,24 +263,19 @@ abstract class Property[T] extends BaseType { self =>
...
-      val value = valueOpt
-    }
-  }
+  private[chisel3] def makeWithValueOpt[T](valueOpt: Option[T])(implicit _tpe: PropertyType[T]): Property[T] =
+    new Property[T](_tpe, valueOpt)
 
   /** Create a new Property based on the type T.
     */
-  def apply[T]()(implicit tpe: PropertyType[T]): Property[tpe.Type] = {
+  def apply[T]()(implicit tpe: PropertyType[T]): Property[T] = {
```
albertchen-sifive, line 271: The path dependent type was necessary before so that e.g. `IO(Output(Property[Int])) := Property(Property[Int]())` would work since `:=` only worked for `Property[T]` with the same type parameter. Now that `:=` works to connect any `Property[_]` regardless of type parametrization, it is not necessary for the return types to be dependent.

But the return type is still important for users that want to layer extra type safety on top of this API e.g. 
```scala
def connectProp(prop: Property[Seq[Int]]) =
  io := prop

// this would not type check anymore
connectProp(Property(Seq(Property[Int](123), IO(Input(Property[Int]())))
```

So I think we still want to keep the return type as `Property[tpe.Type]`, or just get rid of the type parameter entirely `Property[T]` -> `Property` since the core operations on `Property` aren't parameterized on `T` anyway.

jackkoenig: Oh I understand now. Yeah even though `:=` works for any `Property` (because it works for any `Data`), the connectable operators are still typesafe, so I can test for this behavior by using `:#=` instead of `:=` connecting a `Property[Seq[Property[A]]]` to a `Property[Seq[A]]`.

Thanks for the explanation! I will restore the path dependent type.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: I originally removed the Path dependent type of Property because it didn't seem necessary and this was part of that. However, [Albert pointed out](https://github.com/chipsalliance/chisel/pull/3514/files#r1311152427) why it is, in fact, needed so I'm restoring it and backing out this change.

src/test/scala/chiselTests/properties/PropertySpec.scala
jackkoenig: It actually fails when using connectable operators so I'm just turning the `:=` into a `:#=` and restoring the failing test.

The reason for the change is that before `Property` was `Data`, it had it's own `:=` which was type safe. Now that it's a `Data`, `:=` is not type safe, but connectable operators are so we have to use them to illustrate that behavior.

jackkoenig:
Note that I carefully reconstructed the order of fields in `Data` from before `BaseType` was added (and took some of the fields out in https://github.com/chipsalliance/chisel/pull/3429), you can diff `Data.scala` with the prior commit (https://github.com/chipsalliance/chisel/commit/878733f86cb4582f7309de35384af64ed6bb16b7) to see.

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -10,6 +10,7 @@ import chisel3.internal.firrtl.{Connect, Converter, DefInvalid}
 
 import scala.language.experimental.macros
 import _root_.firrtl.passes.CheckTypes
+import chisel3.properties.Property
```
seldridge, line 13: Mega-nit: group with other `chisel3` imports.

src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -338,11 +337,11 @@ class PropertySpec extends ChiselFlatSpec with MatchesAndOmits {
...
       "output b : List<Map<List<Integer>>>",
       "output c : List<Map<List<Integer>>>",
-      "output d : List<Map<List<Integer>>>",
+      // "output d : List<Map<List<Integer>>>",
       """propassign a, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))""",
       """propassign b, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))""",
-      """propassign c, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))""",
-      """propassign d, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))"""
+      """propassign c, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))"""
+      // """propassign d, List<Map<List<Integer>>>(Map<List<Integer>>("foo" -> List<Integer>(Integer(123))))"""
```
seldridge, line 344: Nit: remove comments or do something different here.

seldridge (approved):
LGTM

Do I read this correctly in that this is working for `Wire`, too?

core/src/main/scala/chisel3/internal/BiConnect.scala
jackkoenig: That's what I get for using VScode 😅

src/test/scala/chiselTests/properties/PropertySpec.scala
jackkoenig: Lol I meant to uncomment these but forgot, will fix

jackkoenig:
> Do I read this correctly in that this is working for Wire, too?

You do indeed, I'll mention that in the release notes.

azidar (approved):

PR #3520, opened 2023-09-03
Implement Verilog output stream `PanamaCIRCTConverter`

SpriteOvO:
Based on llvm/circt#6094 and llvm/circt#6036.

~We are trying to reimplement CIRCT `firtool` in Chisel, with it, we can select and manipulate passes to export to targets with the new `CIRCTConverter` in chipsalliance/chisel#3466.~

This PR implements Verilog output stream PanamaCIRCTConverter

This PR is at a very early stage, opening the PR for discussion.

---

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?


sequencer:
We notice width truncation was broken when we try to elaborate the T1 testbench, thanks @seldridge, pointed me the [link](https://github.com/llvm/circt/blob/369537290aa77c16e87e5f3c8696723424907440/lib/Dialect/FIRRTL/FIRRTLUtils.cpp#L120-L137) saying parser is adding additional ops for padding.
Since we are directly issuing MLIR ops in this flow, there won't be any fixing from parser. So it will be slightly difference between firtool and firtool-lib. In fact firrtool-lib will be stricter(which I prefer)

Since it is stricter, I will personally force T1 and rocket-chip to avoid width padding and trunction in connect, and fix them.

But this is an issue we should try to fix in the future, here are some possible solutions:
- remove this logic in parser and add a special post-parse pass, so that we can share this logic between firtool and firtool-lib.
- add these ops in the chisel builder or converter to work around
- make the truncation and padding as warning, and forbid them in the future.(user should explicit padding and truncating)

Since we didn't release panama converter, and won't release it to normal user in a near future, when that day comes, we may already fix this issue. So it won't block the merging of this PR.

For any project which tries to use this feature. please fix the this issue in your user code.

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -552,10 +566,61 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
+    var src = util.referTo(connect.exp, loc)
+
+    val destWidth = util.getWidthOrSentinel(dest.tpe)
+    val srcWidth = util.getWidthOrSentinel(src.tpe)
+
+    if (destWidth < srcWidth) {
+      val isSignedDest = dest.tpe match {
+        case _: fir.SIntType => true
+        case _               => false
+      }
```
sequencer, line 579:
```suggestion
      val isSignedDest = dest.tpe.isInstatnce[fir.SIntType]
```

sequencer (approved):
Change in this PR doesn't affect the review flow, and possibly not be included in Chisel 6, so I mark this as `internal`.

After this PR, the chisel-circt-binding project is able to emit Verilog by calling CIRCT C-API, thanks hard work by @SpriteOvO 

The next milestone should possibly be pass plugin, e.g:
- inspect `Property` via Scala.
- use `om-linker` to link modules.

This feature is in a very early stage! please don't use in production flow for now!

PR #3522, opened 2023-09-07
Add `AnyRef` type and rework `PropertyType.getPropertyType`

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification

this adds `AnyClassType`, which cannot be instantiated and can be used to create properties of type `AnyRef` e.g.
```scala
val io: PropertyType[ClassType] = IO(Input(Property[AnyClassType]()))
// emits "input io : AnyRef"
```

This also changes the signature of `PropertyType.getPropertyType` to take no arguments instead of `Option[T]`. Previously creating a `Property[ClassType]` that wasn't backed by a `Some(ClassType)` value would result in a None.get error during serialization. This change makes it so that it is always possible to serialize the type, and also makes it so that it is no longer possible to do `Property[ClassType]()`. Now, to create a `Property[ClassType]` you must do one of the following
```scala
// AnyRef property
val io: Property[ClassType] = IO(Input(Property[AnyClassType]()) // input io: AnyRef

// property with the type of some instance of ClassType
val cls = ClassType("foobar")
val io: Property[ClassType] = IO(Input(Property[cls.Type]())) // input io: Inst<foobar>

// property with the same type as some instance of Property[ClassType]
val ref = Class.unsafeGetReferenceType("MyClass")
val io: Property[ClassType] = IO(Input(Property[ref.ClassType]())) // input io: Inst<MyClass>
val io: Property[ClassType] = IO(Input(ref.cloneType)) // input io: Inst<MyClass>
```

This also means that you cannot create properties of arbitrary collections of `Property[ClassType]` anymore, you must cast `Property[ClassType]` properties to a specific `ClassType` using the `as` extension method
```scala
val io: Property[Seq[Property[ClassType]]] = ???

// this doesn't work anymore
io := Seq(Class.unsafeGetReferenceType("foo"), Class.unsafeGetReferenceType("bar"))

// this works
val cls = ClassType("foobar")
io := Seq(Class.unsafeGetReferenceType("foo").as(cls), Class.unsafeGetReferenceType("bar").as(cls))

// this also works
val cls = Class.unsafeGetReferenceType("foobar")
io := Seq(Class.unsafeGetReferenceType("foo").as(cls), Class.unsafeGetReferenceType("bar").as(cls))

// this also works
io := Seq(Class.unsafeGetReferenceType("foo").asAnyClassType, Class.unsafeGetReferenceType("bar").asAnyClassType)
```

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -219,7 +219,7 @@ private[chisel3] object PropertyType extends TuplePropertyTypeInstances with Low
   */
 sealed trait Property[T] extends Data { self =>
   sealed trait ClassType
-  object ClassType {
+  private object ClassType {
     implicit def classTypeProvider(implicit evidence: T =:= chisel3.properties.ClassType): ClassTypeProvider[ClassType] = ClassTypeProvider(getPropertyType)
     implicit def propertyType(implicit evidence: T =:= chisel3.properties.ClassType) = new ClassTypePropertyType[Property[ClassType] with self.ClassType](getPropertyType) {
       override def convert(value: Underlying, ctx: ir.Component, info: SourceInfo): fir.Expression =
@@ -314,21 +314,17 @@ sealed trait Property[T] extends Data { self =>
 
 }
 
-trait ClassTypeProvider[A] {
+private[chisel3] sealed trait ClassTypeProvider[A] {
   val classType: fir.PropertyType
 }
 
-object ClassTypeProvider {
+private[chisel3] object ClassTypeProvider {
   def apply[A](className: String) = new ClassTypeProvider[A] {
     val classType = fir.ClassPropertyType(className)
   }
   def apply[A](_classType: fir.PropertyType) = new ClassTypeProvider[A] {
     val classType = _classType
   }
-}
-
-trait Typeable[T, Lit] {
-  def convertUnderlying(l: Lit): T
 }
 
 /** Companion object for Property.
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -33,7 +33,7 @@ sealed trait PropertyType[T] {
 
   /** Get the IR PropertyType for this PropertyType.
     */
-  def getPropertyType(value: Option[T]): fir.PropertyType
+  def getPropertyType(): fir.PropertyType
 
   /** Get convert from the underlying representation to firrtl expression
     */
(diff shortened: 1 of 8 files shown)


src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -585,4 +587,74 @@ class PropertySpec extends ChiselFlatSpec with MatchesAndOmits {
...
+  it should "not be able to get the ClassType of a Property[T] if T != ClassType" in {
+    assertTypeError("""
+      val intProp = Property[Int](123)
+      val foobar = Property[intProp.ClassType]()
+    """)
+  }
+
+  it should "emit correct types for all the ways of creating class references and properties" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new RawModule {
+      val cls = ClassType("MyClass")
```
jackkoenig, line 600: I dont think we want this to work, `Class.unsafeGetDynamicObject` is supposed to be the way to reflectively talk about classes. Now that returns a DynamicObject so I get you need a way to talk about Classes themselves, although perhaps the type should be a method on the Object rather than a `ClassType`? Eventually we'll make D/I work for Classes and then that will be the way you're typically supposed to create Classes.

At the very least can it have an annoying method name (rather than "apply") to indicate it's unsafe?

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -298,29 +314,58 @@ sealed trait Property[T] extends Data { self =>
 
 }
 
+trait ClassTypeProvider[A] {
```
jackkoenig, line 317:
```suggestion
sealed trait ClassTypeProvider[A] {
```
This should be sealed right?

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -298,29 +314,58 @@ sealed trait Property[T] extends Data { self =>
...
+object ClassTypeProvider {
+  def apply[A](className: String) = new ClassTypeProvider[A] {
+    val classType = fir.ClassPropertyType(className)
+  }
+  def apply[A](_classType: fir.PropertyType) = new ClassTypeProvider[A] {
+    val classType = _classType
+  }
+}
+
+trait Typeable[T, Lit] {
```
jackkoenig, line 330:
```suggestion
sealed trait Typeable[T, Lit] {
```
This should be sealed right?

jackkoenig (approved):
Approving but I'm a little concerned about `ClassType`, the fact that we can create classes and objects by-name is a wart that we want to discourage use of.

src/test/scala/chiselTests/properties/PropertySpec.scala
albertchen-sifive: make `ClassType.apply` private and added `ClassType.unsafeGetClassTypeByName`

core/src/main/scala/chisel3/properties/Property.scala
albertchen-sifive: make this `sealed` and `private[chisel3]`

core/src/main/scala/chisel3/properties/Property.scala
albertchen-sifive: deleted this, ended up not needing it

mikeurbach (approved):
This works for me, it's a big step in the right direction.

albertchen-sifive:
getting rid of the api-modification label, since these APIs have never been released before

jackkoenig:
@albertchen-sifive Please either rebase to cleanup the commit history before merging with a merge commit, or do squash-and-merge and cleanup the commit message, this was merged with broken commits with messages like `wip`.

PR #3533, opened 2023-09-15
Fix 'Malformed class name' errors in typeName and related implementations

jared-barocsi:
Replaces instances of `getClass.getSimpleName` in `typeName`, which is problematic in Java 8, with a common utility function (which just bundles the default `Module.desiredName` logic). Also disallows mixing `HasAutoTypename` into an anonymous Record, which is prone to generating `Malformed class name` errors.

Compiler plugin change was validated by attempting to compile anonymous `Bundle with HasAutoTypename` classes, with positive erroring.
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

Squash

#### Release Notes
- Introduce a `simpleClassName` utility object that emulates a `getClass.getSimpleName` call without throwing `Malformed class name` exceptions when Java 8 is used. `typeName` and all related implementations use this function now instead of `getClass.getSimpleName`.
- `HasAutoTypename` can no longer be mixed into an anonymous `Record`; the compiler plugin now reports this as a compilation error.

- [ ] Did you review?

### DIFF
--- a/plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
+++ b/plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
@@ -179,9 +179,14 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
 
     def generateAutoTypename(record: ClassDef, thiz: global.This, conArgsOpt: Option[List[Tree]]): Option[Tree] = {
       conArgsOpt.flatMap { conArgs =>
-        val recordIsInnerClass = !(record.symbol.isPackageClass || record.symbol.outerClass.isStaticOwner)
-        if (recordIsInnerClass) {
-          val illegalRecordTypeStr = if (record.name.toString.contains("anon")) "anonymous" else "inner"
+        // Check if the class is an inner class within an outer concrete class (not a static object or a package)
+        val recordIsInnerClassInOuterClass =
+          !(record.symbol.outerClass.isStaticOwner || record.symbol.outerClass.isPackageClass)
+        val recordIsAnon = record.symbol.isAnonOrRefinementClass
+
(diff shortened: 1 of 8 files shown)


mwachs5 (commented):
are there any tests we should add? For inner class and or anonymous record?

jared-barocsi:
> are there any tests we should add? For inner class and or anonymous record?

I'm not able to add tests since it's a compiler step, the spec would fail compilation if I attempted to mix `HasAutoTypename` into an anonymous record.

plugin/src/main/scala/chisel3/internal/plugin/BundleComponent.scala
```diff
@@ -177,21 +177,34 @@ private[plugin] class BundleComponent(val global: Global, arguments: ChiselPlugi
...
-            Flag.OVERRIDE | Flag.PROTECTED
+    def generateAutoTypename(record: ClassDef, thiz: global.This, conArgsOpt: Option[List[Tree]]): Option[Tree] = {
+      conArgsOpt.flatMap { conArgs =>
+        val recordIsAnon = record.symbol.isAnonOrRefinementClass
+
+        // Anonymous `Records` in any scope are forbidden
+        if (record.symbol.isAnonOrRefinementClass) {
+          global.reporter.error(
+            record.pos,
+            "Users can not mix 'HasAutoTypename' into an anonymous Record. Pull the Record definition into its own outer class."
```
jackkoenig, line 188:
```suggestion
            "Users cannot mix 'HasAutoTypename' into an anonymous Record. Create a named class."
```
Previous error was conflating anonymity with inner-outer, we just need them to give the Record a name.

jackkoenig (approved):

PR #3535, opened 2023-09-18
Add PolymorphicIOModule, PolymorphicIOBlackBox

seldridge:
Add a `private[chisel3]` ability to block further IO creation in a
`BaseModule`.  This is intended for advanced, in-tree APIs that need to
disallow the creation of further IO by a user.  This API has mutable state
to track if more IO is allowed to be created, a method to get the
value, and a method to set the module to disallow further IO creation.
This one-way setting (h/t @jackkoenig) tries to keep this as locked down
as possible.

Add a new module hierarchy that promotes the type of the IO of that module
to a type parameter.  This module hierarchy includes:

  1. PolymorphicIOBaseModule
  2. PolymorphicIORawModule
  3. PolymorphicIOBlackBox

Both (2) and (3) inherit from (1).  These modules are all polymorphic in
their type of IO.  It follows that no more IO are allowed to be created
other than what is in the type.

#### Release Notes
Add `PolymorphicIORawModule` and `PolymorphicIOExtModule` that have a type parameter describing what their IO is.  This is intended for advanced APIs which want to have guarantees about the IO will be.  These share a common trait, `PolymorphicIOBaseModule` which can be used to allow for substitution of one for the other.

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -392,6 +392,20 @@ package experimental {
     /** Internal check if a Module's constructor has finished executing */
     private[chisel3] def isClosed = _closed
 
+    /** Mutable state that indicates if IO is allowed to be created for this module.
+      * This can be used for advanced Chisel library APIs that want to limit
+      * what IO is allowed to be created for a module.
+      */
+    private var _isIOCreationAllowed = true
+
+    /** If true, then this module is allowed to have user-created IO. */
+    private[chisel3] def isIOCreationAllowed = _isIOCreationAllowed
+
+    /** Disallow any more IO creation for this module. */
+    private[chisel3] def disallowIOCreation(): Unit = {
+      _isIOCreationAllowed = false
+    }
+
     private[chisel3] var toDefinitionCalled:  Option[SourceInfo] = None
     private[chisel3] var modulePortsAskedFor: Option[SourceInfo] = None
 
--- a/src/main/scala/chisel3/PolymorphicIOModule.scala
+++ b/src/main/scala/chisel3/PolymorphicIOModule.scala
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, ExtModule, FlatIO, Param}
+
+/** A module or external module whose IO is generated from a specific generator.
+  * This module may have no additional IO created other than what is specified
+  * by its `ioGenerator` abstract member.
+  */
+sealed trait FixedIOBaseModule[A <: Data] extends BaseModule {
+
+  /** A generator of IO */
+  protected def ioGenerator: A
+
+  final val io = FlatIO(ioGenerator)
+  disallowIOCreation()
+
+}
+
+/** A Chisel module whose IO is determined by an IO generator.  This module
+  * cannot have additional IO created by modules that extend it.
+  *
+  * @param ioGenerator
+  */
+class FixedIORawModule[A <: Data](final val ioGenerator: A) extends RawModule with FixedIOBaseModule[A]
+
+/** A Chisel blackbox whose IO is determined by an IO generator.  This module
+  * cannot have additional IO created by modules that extend it.
+  *
+  * @param ioGenerator
+  * @param params
+  */
+class FixedIOExtModule[A <: Data](final val ioGenerator: A, params: Map[String, Param] = Map.empty[String, Param])
+    extends ExtModule(params)
+    with FixedIOBaseModule[A]
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -23,6 +23,7 @@ object IO {
     */
   def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
     val module = Module.currentModule.get // Impossible to fail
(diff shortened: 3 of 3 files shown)


core/src/main/scala/chisel3/Module.scala
```diff
@@ -392,6 +392,20 @@ package experimental {
...
     private[chisel3] def isClosed = _closed
 
+    /** Mutable state that indicates if IO is allowed to be created for this module.
+      * This can be used for advanced Chisel library APIs that want to limit
+      * what IO is allowed to be created for a module.
+      */
+    private var _allowIOCreation = true
+
+    /** If true, then this module is allowed to have user-created IO. */
+    private[chisel3] def allowIOCreation = _allowIOCreation
```
azidar, line 402: Rename to "isIOCreationAllowed" ? The "allowIOCreation" vs "disallowIOCreation()" is confusing because one is a setter, while the other is a getter, but their names imply inverses of each other (e.g. isEmpty vs nonEmpty)

src/main/scala/chisel3/PolymorphicIOModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, ExtModule, FlatIO, Param}
+
+/** A module or external module whose IO is generated from a specific generator.
+  * This module may have no additional IO created other than what is specified
+  * by its `ioGenerator` abstract member.
+  */
+sealed trait PolymorphicIOBaseModule[A <: Data] extends BaseModule {
```
azidar, line 11: Was going to suggest an abstract class here, but I take it back, I think a trait is better.

azidar (approved):
I'm not in love with the `PolymorphicIO` name. I feel like most Chisel modules could make the claim that they are polymorphic over their IOs. Isn't it more like a "FixedIO" module?

Anyways, no major issues, I approve.

core/src/main/scala/chisel3/Module.scala
seldridge: Makes sense. Good suggestion. I updated to this in the rebase.

seldridge:
"Fixed" is also a good idea. I think this will much better match what a user thinks about here. "The IO is fixed" vs. "The IO is polymorphic" (???). 👍  Changes to this in the rebase.

src/main/scala/chisel3/PolymorphicIOModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, ExtModule, FlatIO, Param}
+
+/** A module or external module whose IO is generated from a specific generator.
+  * This module may have no additional IO created other than what is specified
+  * by its `ioGenerator` abstract member.
+  */
+sealed trait FixedIOBaseModule[A <: Data] extends BaseModule {
```
mwachs5, line 11: 👍 on the FixedIO name

src/main/scala/chisel3/PolymorphicIOModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /** A generator of IO */
+  protected def ioGenerator: A
+
+  final val io = FlatIO(ioGenerator)
```
mwachs5, line 16: couldn't the order of trait-mixing-in allow a user to get around this?

```
trait MySneakyIO {
  val sneakyIO = IO(UInt(32.W))
}

class MyThing extends MySneakyIO with FixedIORawModule 
```

seldridge: Fortunately, no because `FixedIORawModule`/`FixedIOExtModule` are classes and not traits. Those have to be the first thing you mix-in, i.e., they need to follow `extends`.

A user will see the following compile-time error if they try the above:

```
[error] /repos/github.com/chipsalliance/chisel/src/test/scala/chiselTests/FixedIOModuleSpec.scala:55:43: class FixedIORawModule needs to be a trait to be mixed in
[error]     class MyThing extends MySneakyIO with FixedIORawModule[Bool](Bool())
[error]                                           ^
```

If they change the order to be legal Scala, they'll get a runtime error because the trait constructor body runs after IO creation is disallowed:

```
[info]   java.lang.IllegalArgumentException: requirement failed: This module cannot have user-created IO
[info]   at ... ()
[info]   at chiselTests.FixedIOModuleSpec$MySneakyIO$1.$anonfun$sneakyIO$2(FixedIOModuleSpec.scala:52)
[info]   at chisel3.experimental.prefix$.apply(prefix.scala:50)
[info]   at chiselTests.FixedIOModuleSpec$MySneakyIO$1.$anonfun$sneakyIO$1(FixedIOModuleSpec.scala:52)
[info]   at chisel3.internal.plugin.package$.autoNameRecursively(package.scala:33)
[info]   at chiselTests.FixedIOModuleSpec$MySneakyIO$1.$init$(FixedIOModuleSpec.scala:52)
[info]   at chiselTests.FixedIOModuleSpec$MyThing$1.<init>(FixedIOModuleSpec.scala:55)
[info]   at chiselTests.FixedIOModuleSpec.$anonfun$new$9(FixedIOModuleSpec.scala:57)
[info]   at ... ()
[info]   at ... (Stack trace trimmed to user code only. Rerun with --full-stacktrace to see the full stack trace)
```

mwachs5:
nit: can you change the name of this PR because i think it shows up in the release notes?

PR #3538, opened 2023-09-19
Make FlatIO work (as IO) for non-Record Data

seldridge:
Change `FlatIO` to work for any `Data` as opposed to just for `Record`. For non-`Record`, have this behave the same as `IO`.  I.e., there is no prefix stripping.

#### Release Notes

Change `FlatIO` to work for any `Data` as opposed to just for `Record`. For non-`Record`, have this behave the same as `IO`.  I.e., there is no prefix stripping.

### DIFF
--- a/core/src/main/scala/chisel3/experimental/package.scala
+++ b/core/src/main/scala/chisel3/experimental/package.scala
@@ -81,8 +81,11 @@ package object experimental {
   type Direction = ActualDirection
   val Direction = ActualDirection
 
-  /** The same as [[IO]] except there is no prefix for the name of the val */
-  def FlatIO[T <: Record](gen: => T)(implicit sourceInfo: SourceInfo): T = noPrefix {
+  /** The same as [[IO]] except there is no prefix when given a [[Record]] or
+    * [[Bundle]].  For [[Element]] ([[UInt]], etc.) or [[Vec]] types, this is
+    * the same as [[IO]].
+    */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/experimental/package.scala
```diff
@@ -93,20 +96,27 @@ package object experimental {
...
 
-    implicit val dv: DataView[Seq[Data], T] = DataView.mapping(
-      _ => chiselTypeClone(gen).asInstanceOf[T],
-      (seq, rec) => seq.zip(rec._elements.toSeq.reverse).map { case (port, (_, field)) => port -> field }
-    )
-    ports.viewAs[T]
+    type R = T with Record
+    gen match {
+      case _:      Element => IO(gen)
+      case _:      Vec[_] => IO(gen)
```
jackkoenig, line 103: Are we sure this is the correct behavior for Vec? I know it's unlikely people usually want ports named `0`, `1`, ... but this behavior is inconsistent with the behavior you'd get from `MixedVec` (which is actually a `Record).

seldridge: Discussed offline. Summarizing here for others.

It's unclear what should happen for `Vec[_]` other than some behavior should happen. Both `MixedVec` style behavior and the PR'd behavior (there is an `io` port) are reasonable. The PR'd behavior may be less surprising. If a user wants something different then they can create a bundle (or `MixedVec`) for that.

jackkoenig (approved):
LGTM!

PR #3539, opened 2023-09-19
MixedVec subclasses immutable.IndexedSeq

mwachs5:
In the change to scala 2.13, collection.IndexedSeq is not enough to be considered a Seq

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Change MixedVec to inherit from collection.immutable.IndexedSeq, so that it can work for Seq[Data].

- [ ] Did you review?

### DIFF
--- a/integration-tests/src/test/scala/chiselTest/MixedVecIntegrationSpec.scala
+++ b/integration-tests/src/test/scala/chiselTest/MixedVecIntegrationSpec.scala
@@ -76,14 +76,14 @@ class MixedVecZeroEntryTester extends BasicTester {
 }
 
 class MixedVecUIntDynamicIndexTester extends BasicTester {
-  val wire = Wire(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
+  val wire: MixedVec[UInt] = Wire(MixedVec(Seq(UInt(8.W), UInt(16.W), UInt(4.W), UInt(7.W))))
   val n = wire.length
 
   for (i <- 0 until n) {
     wire(i) := i.U
   }
 
-  val vecWire = VecInit(wire.toSeq)
+  val vecWire = VecInit(wire : Seq[UInt])
 
   val (cycle, done) = Counter(true.B, n)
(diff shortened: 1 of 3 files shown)


jackkoenig (approved):

jackkoenig:
@mwachs5 an integration test failed. Annoying that this change technically can break some code but the fix is usually trivial, you should just be able to remove the `.toSeq` (might also need to call splat `:_*`).

mwachs5:
> you should just be able to remove the .toSeq (might also need to call splat :_*).

Interestingly this does not work. The first gives me:

```
[error] /scratch/megan/chisel/integration-tests/src/test/scala/chiselTest/MixedVecIntegrationSpec.scala:86:17: ambiguous reference to overloaded definition,
[error] both macro method apply in object VecInit of type [T <: chisel3.Data](elt0: T, elts: T*): chisel3.Vec[T]
[error] and  macro method apply in object VecInit of type [T <: chisel3.Data](elts: Seq[T]): chisel3.Vec[T]
[error] match argument types (chisel3.util.MixedVec[chisel3.UInt])
[error]   val vecWire = VecInit(wire)
```

The second gives me:

```
error] /scratch/megan/chisel/integration-tests/src/test/scala/chiselTest/MixedVecIntegrationSpec.scala:86:29: Sequence argument type annotation `: _*` cannot be used here:
[error] the corresponding parameter has type chisel3.UInt which is not a repeated parameter type
[error]   val vecWire = VecInit(wire: _*)
```

I have confirmed that `wire` is of type `MixedVec[UInt]` ... what am I missing here?

mwachs5:
Actually, what is this test even attempting to show? It doesn't seem like it should be legal to have a VecInit of a MixedVec with different element sizes. Shouldn't VecInit be checking that all elements have the same chisel type?

mwachs5:
I guess this is just a limitation for VectInit for anything that is both a Seq and Data. Vecs have the same problem (do you want a Vec of Vecs-with-one-element or a Vec of the elements of the Vecs...)

Should VecInit force you to have at least 2 elements?

PR #3541, opened 2023-09-21
Add optional group API

seldridge:
Add optional groups.

### Example

Chisel:
```scala
object A extends group.Declaration(group.Convention.Bind) {
  object B extends group.Declaration(group.Convention.Bind)
}

class Foo extends RawModule {
  val a = IO(Input(Bool()))

  group(A) {
    val w = WireInit(a)
    dontTouch(w)
    group(A.B) {
      val x = WireInit(w)
      dontTouch(x)
    }
  }

}
```

Output FIRRTL:

```
FIRRTL version 3.2.0
circuit Foo :%[[
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~Foo|Foo>w"
  },
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~Foo|Foo>x"
  }
]]
  declgroup A, bind :
    declgroup B, bind :
  module Foo :
    input a : UInt<1>

    group A:
      wire w : UInt<1>
      connect w, a
      group B:
        wire x : UInt<1>
        connect x, w
```

Output Verilog:

```verilog
// Generated by CIRCT firtool-1.56.1-32-g4da6a4d83
module Foo_A_B(
  input _w
);

  wire x = _w;
endmodule

module Foo_A(
  input _a
);

  wire w = _a;
  wire w_probe = w;
endmodule

module Foo(
  input a
);

endmodule


// ----- 8< ----- FILE "groups_Foo_A_B.sv" ----- 8< -----

// Generated by CIRCT firtool-1.56.1-32-g4da6a4d83
`include "groups_Foo_A.sv"
`ifndef groups_Foo_A_B
`define groups_Foo_A_B
bind Foo Foo_A_B foo_A_B (
  ._w (Foo.foo_A.w_probe)
);
`endif

// ----- 8< ----- FILE "groups_Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.56.1-32-g4da6a4d83
`ifndef groups_Foo_A
`define groups_Foo_A
bind Foo Foo_A foo_A (
  ._a (a)
);
`endif
```

#### Release Notes

Add optional groups. This is a feature that can be used to add optional verification functionality, e.g., asserts or debug code, that does not affect the main design. These groups are lowered to FIRRTL Optional Groups (see: https://github.com/chipsalliance/firrtl-spec/pull/108) with the "bind" convention.

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -473,6 +473,7 @@ private[chisel3] class DynamicContext(
   val components = ArrayBuffer[Component]()
   val annotations = ArrayBuffer[ChiselAnnotation]()
   val newAnnotations = ArrayBuffer[ChiselMultiAnnotation]()
+  val groups = mutable.LinkedHashSet[group.Declaration]()
   var currentModule: Option[BaseModule] = None
 
   /** Contains a mapping from a elaborated module to their aspect
@@ -492,6 +493,7 @@ private[chisel3] class DynamicContext(
   var currentClock:         Option[Clock] = None
   var currentReset:         Option[Reset] = None
   var currentDisable:       Disable.Type = Disable.BeforeReset
+  var groupStack:           List[group.Declaration] = Nil
   val errors = new ErrorLog(warningFilters, sourceRoots, throwOnFirstError)
   val namingStack = new NamingStack
 
@@ -550,6 +552,8 @@ private[chisel3] object Builder extends LazyLogging {
 
   def components:  ArrayBuffer[Component] = dynamicContext.components
   def annotations: ArrayBuffer[ChiselAnnotation] = dynamicContext.annotations
+
+  def groups: mutable.LinkedHashSet[group.Declaration] = dynamicContext.groups
 
   def contextCache: BuilderContextCache = dynamicContext.contextCache
 
@@ -760,6 +764,11 @@ private[chisel3] object Builder extends LazyLogging {
     dynamicContext.currentDisable = newDisable
   }
 
+  def groupStack: List[group.Declaration] = dynamicContext.groupStack
+  def groupStack_=(s: List[group.Declaration]): Unit = {
+    dynamicContext.groupStack = s
+  }
+
   def inDefinition: Boolean = {
     dynamicContextVar.value
       .map(_.inDefinition)
@@ -949,6 +958,31 @@ private[chisel3] object Builder extends LazyLogging {
         case _ => None
       }.toSeq
 
+      /** Stores an adjacency list representation of groups.  Connections indicating children. */
+      val groupAdjacencyList = mutable
+        .LinkedHashMap[group.Declaration, mutable.LinkedHashSet[group.Declaration]]()
+        .withDefault(_ => mutable.LinkedHashSet[group.Declaration]())
+
+      // Populate the adjacency list.
+      groups.foreach { group =>
+        groupAdjacencyList(group.parent) = groupAdjacencyList(group.parent) += group
+      }
+
+      /** For a `group.Declaration`, walk all its children and fold them into a
+        * `GroupDecl`.  This "folding" creates one `GroupDecl` for each child
+        * nested under each parent `GroupDecl`.
+        */
+      def foldGroupDecls(decl: group.Declaration): GroupDecl = {
+        val children = groupAdjacencyList(decl)
+        val convention = decl.convention match {
+          case group.Convention.Bind => GroupConvention.Bind
+          case _                     => ???
+        }
+        GroupDecl(decl.sourceInfo, decl.name, convention, children.map(foldGroupDecls).toSeq)
+      }
+
+      val rootGroups = groupAdjacencyList(group.Declaration.Root)
+
       (
         Circuit(
           components.last.name,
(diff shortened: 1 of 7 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -313,6 +320,17 @@ private[chisel3] object Converter {
...
+              val frame = GroupFrame(groupDefine, stmts)
+              stmts = new VectorBuilder[fir.Statement]
+              scope = frame :: scope
+            case GroupDefEnd(info) =>
+              val frame = scope.head.asInstanceOf[GroupFrame]
+              val groupDefine = frame.group.copy(body = fir.Block(stmts.result()))
+              stmts = frame.outer
+              stmts += groupDefine
+              scope = scope.tail
           }
```
seldridge, line 334: I think there is a bug here where this is not properly terminating the group. I'm seeing the problem where anything after the final group are not showing up in the emitted CHIRRTL.

seldridge: The bug wasn't here, but in the Serializer. Fixed in https://github.com/chipsalliance/chisel/pull/3541/commits/8dfc4ed9e8874cdd9bbd38af43db65556be132e6.

seldridge:
The documentation will fail until `firtool` has the `LowerGroups` pass on by default. This is the case for `firtool` top-of-tree, but not on a published release, yet.

lsteveol:
Hey Schuyler, thanks for working on this. Do you guys have an idea when this will make it into a Chisel v6 release (v6.0.0-M4 I assume)?

seldridge:
> Hey Schuyler, thanks for working on this. Do you guys have an idea when this will make it into a Chisel v6 release (v6.0.0-M4 I assume)?

That's a @jackkoenig question. I think we're probably ready for a true 6.0 release and not a milestone at this point.

core/src/main/scala/chisel3/Util.scala
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  /** Return a "good" class name for an object.  This is an improved version of `Class[_].getSimpleName`.
+    * @param obj any object
+    */
+  def goodName(obj: Any): String = {
+    goodName(obj.getClass().getName())
+  }
+
+}
```
jackkoenig, line 28: Just yesterday, @jared-barocsi added `chisel3.util.simpleClassName` (https://github.com/chipsalliance/chisel/blob/4186303c85e8c5afc60685485d2bc8090ffea5bc/core/src/main/scala/chisel3/util/Naming.scala#L6) defined in `core` which has this same implementation, you should use that instead 🙂

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -949,14 +958,43 @@ private[chisel3] object Builder extends LazyLogging {
...
+        */
+      def foldGroupDecls(decl: group.Declaration): GroupDecl = {
+        val children = groupAdjacencyList(decl)
+        val convention = decl.convention match {
+          case group.Convention.Bind => GroupConvention.Bind
+          case _                     => ???
+        }
+        decl match {
+          case leaf if children.isEmpty => GroupDecl(UnlocatableSourceInfo, leaf.name, convention, Seq.empty)
+          case node                     => GroupDecl(UnlocatableSourceInfo, node.name, convention, children.map(foldGroupDecls).toSeq)
```
jackkoenig, line 983: Should we not go ahead and capture the source locator in the `Declaration`?

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -454,13 +472,24 @@ private[chisel3] object Converter {
...
   }
 
+  def rec(decl: GroupDecl): fir.GroupDeclare = {
+    val convention = decl.convention match {
+      case GroupConvention.Bind => fir.GroupConvention.Bind
+    }
+    decl match {
+      case leaf if leaf.children.isEmpty => fir.GroupDeclare(convert(leaf.sourceInfo), leaf.name, convention, Seq.empty)
+      case node                          => fir.GroupDeclare(convert(node.sourceInfo), node.name, convention, node.children.map(rec))
+    }
```
jackkoenig, line 482:
```suggestion
    fir.GroupDeclare(convert(decl.sourceInfo), decl.name, convention, decl.children.map(rec))
```
Unless I'm missing something, these are identical and mapping on empty children will not recurse

jackkoenig (approved):
This is awesome, nice work Schuyler! I've included some suggestions, but overall looks really good.

core/src/main/scala/chisel3/internal/Builder.scala
seldridge: Yep. I was being lazy. I'll pipe this through and make sure it works. I'll do the same for group defines.

seldridge: Updated in 6f173786621b74c0ca52b3d2f006a34fb919755a.

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
seldridge: Good catch. The `Builder` has the same problem. Will fix that there, too.

PR #3545, opened 2023-09-22
Add DedupGroupAnnotation phase

adkian-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Add `DedupGroupAnnotation` phase that generates deduplication groups based on module `desiredNames`

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -535,7 +535,7 @@ package experimental {
     }
 
     /** Legalized name of this module. */
-    final lazy val name =
+    lazy val name =
       try {
         // PseudoModules are not "true modules" and thus should share
         // their original modules names without uniquification
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -535,7 +535,7 @@ package experimental {
     }
 
     /** Legalized name of this module. */
-    lazy val name =
+    final lazy val name =
       try {
         // PseudoModules are not "true modules" and thus should share
         // their original modules names without uniquification
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
@@ -132,6 +132,7 @@ object Instance extends SourceInfoDoc {
       )
       class EmptyExtModule extends ExtModule with PseudoModule {
         override def desiredName: String = extModName
+        override private[chisel3] def _isImportedDefinition = true
         override def generateComponent(): Option[Component] = {
           require(!_closed, s"Can't generate $desiredName module more than once")
           _closed = true
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
@@ -130,9 +130,8 @@ object Instance extends SourceInfoDoc {
           "Imported Definition information not found - possibly forgot to add ImportDefinition annotation?"
         )
       )
-      class EmptyExtModule extends ExtModule {
+      class EmptyExtModule extends ExtModule with PseudoModule {
         override def desiredName: String = extModName
-        override lazy val name = definition.proto.name
         override def generateComponent(): Option[Component] = {
           require(!_closed, s"Can't generate $desiredName module more than once")
           _closed = true
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
@@ -5,7 +5,7 @@ package chisel3.experimental.hierarchy.core
 import scala.language.experimental.macros
 import chisel3._
 import chisel3.experimental.hierarchy.{InstantiableClone, ModuleClone}
-import chisel3.internal.{throwException, Builder}
+import chisel3.internal.{throwException, Builder, PseudoModule}
 import chisel3.experimental.{BaseModule, ExtModule, SourceInfo}
 import chisel3.internal.sourceinfo.InstanceTransform
 import chisel3.internal.firrtl.{Component, DefBlackBox, DefIntrinsicModule, DefModule, Port}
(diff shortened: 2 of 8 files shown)


jackkoenig (commented):
This LGTM, but look into the failed tests. There seems to be some weird behavior where some modules do not get their `.name` evaluated during elaboration (which is surprising, I would expect it to be evaluated for all modules). I'd like to understand what's going on here.

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -116,7 +116,7 @@ object Instance extends SourceInfoDoc {
     // Check to see if the module is already defined internally or externally
     val existingMod = Builder.components.map {
       case c: DefModule if c.id == definition.proto                 => Some(c)
-      case c: DefBlackBox if c.name == definition.proto.name        => Some(c)
+      case c: DefBlackBox[_] if c.name == definition.proto.name     => Some(c)
```
mwachs5, line 119: what is going on with this change? Can we add a unit test related to this change?

src/test/scala/circtTests/stage/ChiselStageSpec.scala
```diff
@@ -867,13 +867,21 @@ class ChiselStageSpec extends AnyFunSpec with Matchers with chiselTests.Utils {
         val a = IO(Input(Bool()))
         val b = IO(Output(Bool()))
 
+        // Chisel groups modules for deduping using their
```
mwachs5, line 870: should there be a test about black boxes deduping (or not?) based on their desiredName? Is that what is going on?

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
adkian-sifive: so this is a hacky attempt to distinguish black boxes with `Instance`-type ids here: https://github.com/chipsalliance/chisel/pull/3545/files#diff-dc0f0eefedf62d62714db80d54d7ead1df351b622236533afd6ae5eb4503d710R32

this is not currently working, could use some ideas whether or not it's workable. cc @jackkoenig

adkian-sifive: specifically what's not working is that with this change, the [dedup spec](https://github.com/chipsalliance/chisel/pull/3545/files#diff-362b0d2e1143590082fa00d7896da154acc13e6579842f044d4667a160a500b2R113) is failing to not dedup regular modules in different groups

src/test/scala/circtTests/stage/ChiselStageSpec.scala
adkian-sifive: i added desiredNames to the hierarchy tests below and it's working, i think that should be enough?

adkian-sifive: the hierarchy spec serves the same purpose since the instances are black boxed

src/main/scala/chisel3/stage/phases/AddDedupGroupAnnotations.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def transform(annotations: AnnotationSeq): AnnotationSeq = {
+    val chiselOptions = view[ChiselOptions](annotations)
+    val circuit = chiselOptions.chiselCircuit.getOrElse {
+      throw new ChiselException(
+        s"Unable to locate the elaborated circuit, did ${classOf[Elaborate].getName} run correctly"
+      )
+    }
+
+    val skipAnnos = annotations.collect { case x: DedupGroupAnnotation => x.target }
```
jackkoenig, line 30:
```suggestion
    val skipAnnos = annotations.collect { case x: DedupGroupAnnotation => x.target }.toSet
```
Need a `.toSet` here or the contains below are going to be very slow.

src/main/scala/chisel3/stage/phases/AddDedupGroupAnnotations.scala
```diff
@@ -0,0 +1,41 @@
...
+
+    @nowarn("msg=class Port")
+    val annos = circuit.components.filterNot {
+      case x @ DefBlackBox(id, _, _, _, _) => id.toString contains "Instance"
```
jackkoenig, line 34: What is this accomplishing? We need to find a better way to do this--if we need to expose some information via a new API on `BaseBlackBox`, that's okay.

adkian-sifive: this catches instances of `DefBlackBox` created here: https://github.com/chipsalliance/chisel/blob/e6c07b3760aad1b7ad1c11c9feecc2549ed406df/core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala#L142

adkian-sifive: dunno seems like a fairly fast way to accomplish this? there's usually a lot fewer black boxes this will get matched on compared to the total number of components

jackkoenig: But why do we need to catch those instances?

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
```diff
@@ -132,6 +132,7 @@ object Instance extends SourceInfoDoc {
       )
       class EmptyExtModule extends ExtModule {
         override def desiredName: String = extModName
+        override lazy val name = definition.proto.name
```
jackkoenig, line 135:
```suggestion
      class EmptyExtModule extends ExtModule with PseudoModule {
        override def desiredName: String = extModName
```

See the code for `PseudoModule` in the implementation of `final lazy val name`.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -535,7 +535,7 @@ package experimental {
     }
 
     /** Legalized name of this module. */
-    final lazy val name =
+    lazy val name =
```
jackkoenig, line 538:
```suggestion
    final lazy val name =
```

core/src/main/scala/chisel3/experimental/hierarchy/core/Instance.scala
adkian-sifive: oh right yes

jackkoenig: FWIW, I didn't think of that till reviewing your changes 😂

jackkoenig (approved):
Welp, we tried to fix it "The Right Way"TM but imported definitions are just too funky.

PR #3547, opened 2023-09-24
Changed Array to Seq in MuxLookup explantation

madsrumlenordstrom:
`MuxLookup(idx, default)(Array(0.U -> a, 1.U -> b, ...))` was deprecated in 3.6 and now it uses `MuxLookup(idx, default)(Seq(0.U -> a, 1.U -> b, ...))`. However explanation in documentation still uses an Array. This PR fixes this.


jackkoenig (approved):
Thanks!

PR #3560, opened 2023-09-29
Fix and test probe of ChiselEnum.

dtzSiFive:
Don't add enum annotations to probes of enum.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Fix probe of ChiselEnum.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/ProbeSpec.scala
+++ b/src/test/scala/chiselTests/ProbeSpec.scala
@@ -552,4 +552,21 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
       when(done) { stop() }
     }) should be(true)
   }
+
+  "Enum probe" should "work" in {
+    object MyEnum extends ChiselEnum {
+      val e0, e1, e2 = Value
+    }
+    val chirrtl = ChiselStage.emitCHIRRTL(
+      new RawModule {
+        val a = IO(Output(RWProbe(MyEnum())))
+
+        val w = WireInit(MyEnum(), MyEnum.e1)
+        val w_probe = RWProbeValue(w)
+        define(a, w_probe)
+      },
+      Array("--full-stacktrace")
+    )
+    processChirrtl(chirrtl) should contain("define a = rwprobe(w)")
+  }
 }
--- a/src/test/scala/chiselTests/ProbeSpec.scala
+++ b/src/test/scala/chiselTests/ProbeSpec.scala
@@ -557,16 +557,13 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
     object MyEnum extends ChiselEnum {
       val e0, e1, e2 = Value
     }
-    val chirrtl = ChiselStage.emitCHIRRTL(
-      new RawModule {
-        val a = IO(Output(RWProbe(MyEnum())))
+    class TestMod extends RawModule {
+      val a = IO(Output(RWProbe(MyEnum())))
 
-        val w = WireInit(MyEnum(), MyEnum.e1)
-        val w_probe = RWProbeValue(w)
-        define(a, w_probe)
-      },
-      Array("--full-stacktrace")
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -552,4 +552,21 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
...
+    }
+    val chirrtl = ChiselStage.emitCHIRRTL(
+      new RawModule {
+        val a = IO(Output(RWProbe(MyEnum())))
+
+        val w = WireInit(MyEnum(), MyEnum.e1)
+        val w_probe = RWProbeValue(w)
+        define(a, w_probe)
+      },
+      Array("--full-stacktrace")
```
seldridge, line 568: Nit: the options to Chisel aren't necessary.

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -552,4 +552,21 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
...
+    )
+    processChirrtl(chirrtl) should contain("define a = rwprobe(w)")
```
seldridge, line 570: This check won't catch the exact problem. You can use `matchesAnOmits` or roll your own check that the annotation is _not_ emitted.

E.g., `chirrtl should not include("EnumAnnotation")`. I'd check to make sure that this does actually work (fails without the change!).

seldridge (commented):
Fix looks good. Test is suspect. 🧐

src/test/scala/chiselTests/ProbeSpec.scala
dtzSiFive: Hmm, unfortunately the problem is that it never makes it that far without the fix-- instead it errors:
```
[info] - should work *** FAILED ***
[info]   firrtl.annotations.AnnotationException: Illegal component name: Node(ProbeSpec_Anon.w: Wire[MyEnum$1])
[info]   at firrtl.annotations.ComponentName.<init>(Target.scala:835)
[info]   at chisel3.internal.NamedComponent.toNamed(Builder.scala:358)
[info]   at chisel3.internal.NamedComponent.toNamed$(Builder.scala:356)
[info]   at chisel3.Data.toNamed(Data.scala:286)
[info]   at chisel3.Data.toNamed(Data.scala:286)
[info]   at chisel3.experimental.EnumAnnotations$EnumComponentChiselAnnotation.toFirrtl(EnumAnnotations.scala:20)
[info]   at chisel3.experimental.EnumAnnotations$EnumComponentChiselAnnotation.toFirrtl(EnumAnnotations.scala:19)
[info]   at chisel3.internal.firrtl.Circuit.$anonfun$firrtlAnnotations$1(IR.scala:467)
[info]   at scala.collection.immutable.List.flatMap(List.scala:293)
[info]   at scala.collection.immutable.List.flatMap(List.scala:79)
```

Which is actually the problem that brought me to this.  However you're right, and interestingly, there's still an annotation present even if I drop everything but the port (!) (without this fix it errors as above, with it we still produce some enum annotation on the rwprobe port but Chisel doesn't error).

However this example as-is (with wire + define) works through to firtool (to SV) since CIRCT doesn't even try to resolve it and  just drops such annotations on the floor.   I'll update to test that probes of enums work through the pipeline as that's what matters in some sense.

I'm not sure how to stop the annotation on the rwprobe port, although it's benign so maybe it's okay? :disappointed: 

Anyway good call, thanks!  There was definitely more to this...

seldridge (approved):

dtzSiFive:
Thanks for review + feedback! Hopefully we'll move to FIRRTL enum awesomeness soon! :+1:

PR #3561, opened 2023-09-29
[nfc] Add documentation of optional groups

seldridge:
Add documentation for groups.

This is expected to fail until `firtool-1.57.0` is used in Chisel.

### DIFF
--- a/docs/src/explanations/optional-groups.md
+++ b/docs/src/explanations/optional-groups.md
@@ -141,20 +141,8 @@ class Foo extends Module {
 
 ```
 
-After compilation, this will produce four Verilog files (one for `Foo` and one
-for each optional group definition in module `Foo`):
-
-1. `Foo`
-1. `Foo_Verification`
-1. `Foo_Verification_Assert`
-1. `Foo_Verification_Debug`
-
-The ports of each module created from an optional group will be automatically
-determined based on what that group captured from outside the group.  In the
-example above, the `Verification` group definition captured port `a`.  Both the
-`Assert` and `Debug` group definitions captured `a` and `a_d0`.
-
-Additionally, three bind files (one for each optional group declaration):
+After compilation, this will produce three group include files with the
+following filenames.  One file is created for each optional group:
 
 1. `groups_Foo_Verification.sv`
 1. `groups_Foo_Verification_Assert.sv`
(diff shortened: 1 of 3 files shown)


docs/src/explanations/optional-groups.md
```diff
@@ -0,0 +1,177 @@
...
+respectively, assertions from prints.
+
+One way to write this in Scala is the following:
+
+```scala mdoc:reset:silent
+import chisel3._
+import chisel3.group.{Convention, Declaration}
+
+// All groups are declared here.  The Assert and Debug groups are nested under
+// the Verification group.
```
mwachs5, line 110: Not sure you need to change anything here, but for my own curiousity, I assume the nesting-ness of the object matching the expected nesting usage is just for convention/sanity? Or is it important to nest the objects the same as the usages?

docs/src/explanations/optional-groups.md
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+}
+
+```
+
+After compilation, this will produce four Verilog files (one for `Foo` and one
+for each optional group definition in module `Foo`):
+
+1. `Foo`
```
mwachs5, line 147: are these supposed to have the .sv extension? you mentioned these are files?

docs/src/explanations/optional-groups.md
```diff
@@ -0,0 +1,177 @@
...
+1. `Foo_Verification_Debug`
+
+The ports of each module created from an optional group will be automatically
+determined based on what that group captured from outside the group.  In the
+example above, the `Verification` group definition captured port `a`.  Both the
+`Assert` and `Debug` group definitions captured `a` and `a_d0`.
+
+Additionally, three bind files (one for each optional group declaration):
+
+1. `groups_Foo_Verification.sv`
```
mwachs5, line 159: are these the bind files? Given the warning below, maybe you should mention something like "The ones prefixed by `groups_`". And I guess I am suprised that these files don't have the name `_bindings` or something similar in the name...

website/docs/src/main/resources/microsite/data/menu.yml
```diff
@@ -172,6 +172,10 @@ options:
         url: chisel3/docs/explanations/properties.html
         menu_type: chisel3
         menu_section: properties
+      - title: Optional Groups
```
mwachs5, line 175: thanks for adding this here. Can you add it in the same place on the ToC page of the /docs/src/explanations/explanations.md?

docs/src/explanations/optional-groups.md
seldridge: It's important that the group definition nesting matches the group declaration nesting. The declaration nesting establishes what groups depend on other groups. In this example, the `Assert` group depends on the `Verification` group, the `Assume` group also depends on the `Verification` group, but the `Assert` and `Assume` group do not depend on each other.

This then needs to be enforced so that a group definition cannot access anything that it is not supposed to. For this example, `Assert` and `Assume` can access anything in `Verification`, but `Verification` cannot access anything in `Assert` or `Assume`.

By establishing the nesting and then enforcing it (after https://github.com/chipsalliance/chisel/pull/3562), this is able to use Scala's lexical scoping rules to ensure that anything accessible in a group from Scala's perspective is accessible in a group from FIRRTL's perspective.

Put differently, this is trying to explicitly disallow the following because this changes the dependencies between the groups (`Assert` now depends on `Assume` and it is capturing things it isn't supposed to):

```scala
group(Verification) {
  group(Verification.Assume) {
    val a = Wire(Bool())
    group (Verification.Assert) {
      val b = WireInit(a)
    }
  }
}
```

We could be slightly more relaxed on the Chisel API side and allow a nested group to be used without its parent, e.g., allow code like the following where there is an implicit wrapping of each of both of these (separately) with a `group(Verification)`:

```scala
group(Verification.Assert) { /* ... */ }
group(Verification.Assume) { /* ... */ }
```

This requires a simple `MergeGroups` pass to work well: https://github.com/llvm/circt/issues/6228

docs/src/explanations/optional-groups.md
seldridge: Good catch. It should say "four Verilog modules". I'll change this.

Whether or not these are actual files isn't part of the ABI. Only the names and existence of the bind files are part of the ABI.

docs/src/explanations/optional-groups.md
seldridge: Yes, these are files that contain Verilog bind statements.

The warning should be located earlier, during the discussion of the Verilog modules that are produced, not the bind files. The bind files are part of the ABI and must exist. Good thinking.

This could or could not include "_bindings" in the filename. However, that would be better discussed on the FIRRTL spec given that `groups_<circuit>_<root>{_<nested>}*.sv` is the current format (see: https://github.com/chipsalliance/firrtl-spec/blob/main/abi.md#bind-lowering-convention). The main reason to not include `_bindings` is that an SV bind is only one envisioned convention and the file, if included, enables the group. Whether or not this is implemented via SV bind is arguably an implementation. There is expected to eventually be an `ifdef` group convention that will lower groups to inline Verilog `ifdefs`. If we keep the filename as `groups_` then the filename can be the same regardless of the group convention used.

WDYT?

PR #3563, opened 2023-10-02
Fix Decodetable

sequencer:
- bug fix for decoderbundle not being able to be lhs op
- add default field for DecodeField for DecodeTable non X case

### DIFF
--- a/src/main/scala/chisel3/util/experimental/decode/DecoderBundle.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/DecoderBundle.scala
@@ -23,11 +23,15 @@ trait DecodeField[T <: DecodePattern, D <: Data] {
 
   def chiselType: D
 
+  def default: BitPat = dc
+
   final def width: Int = chiselType.getWidth
 
   def dc: BitPat = BitPat.dontCare(width)
 
   def genTable(op: T): BitPat
+
+  require(width == default.width)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sequencer:
Ask for review, cc @jackkoenig

src/main/scala/chisel3/util/experimental/decode/DecoderBundle.scala
```diff
@@ -71,15 +75,19 @@ class DecodeBundle(fields: Seq[DecodeField[_, _]]) extends Record {
...
 
   def bundle: DecodeBundle = new DecodeBundle(fields)
 
   private val _table = patterns.map { op =>
     op.bitPat -> fields.reverse.map(field => field.genTable(op)).reduce(_ ## _)
   }
-  val table: TruthTable = TruthTable(_table, BitPat.dontCare(_table.head._2.getWidth))
+
+  val table: TruthTable = TruthTable(
+    patterns.map { op => op.bitPat -> fields.reverse.map(field => field.genTable(op)).reduce(_ ## _) },
```
jackkoenig, line 88:
```suggestion
    _table,
```
Is this not still `_table` above?

src/main/scala/chisel3/util/experimental/decode/DecoderBundle.scala
```diff
@@ -71,15 +75,19 @@ class DecodeBundle(fields: Seq[DecodeField[_, _]]) extends Record {
...
+    fields.reverse.map(_.default).reduce(_ ## _)
```
jackkoenig, line 89: I don't understand this part of the change, this feels very different than the previous use of `BitPat.dontCare`.

PR #3566, opened 2023-10-03
Support probe-define in monoconnects

rwy7:
- Enable mono/ `:=`connects for probes, which will emit a ProbeDefine.
- Also enables the fancier connect forms
- A "Probe<Bundle>" will be connected at the root

Co-authored-by: Will Dietz <will.dietz@sifive.com>

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification

#### Desired Merge Strategy

#### Release Notes

- Enable mono/ `:=`connects for probes, which will emit a ProbeDefine.
- Also enables the fancier connect forms
- A "Probe<Bundle>" will be connected at the root

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -205,14 +205,20 @@ object DataMirror {
   def getIntermediateAndLeafs(d: Data): Seq[Data] = collectAllMembers(d)
 
   /** Recursively collect just the leaf components of a data component's children
-    * (i.e. anything that isn't a `Record` or a `Vec`, but an `Element`)
+    * (i.e. anything that isn't a `Record` or a `Vec`, but an `Element`).
+    * Probes of aggregates are also considered leaves.
     *
     * @param d Data component to recursively collect leaf components.
     *
     * @return All `Element` components; intermediate fields/indices are not included
     */
   def collectLeafMembers(d: Data): Seq[Data] =
-    DataMirror.collectMembers(d) { case x: Element => x }.toVector
+    DataMirror
+      .collectMembers(d) {
+        case x: Element => x
+        case x if hasProbeTypeModifier(x) => x
+      }
+      .toVector
 
   /** Recursively collect all expanded member components of a data component, including
     * intermediate aggregate nodes
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -243,12 +243,12 @@ object DataMirror {
     def iterator = {
       val myItems = collector.lift(d).map { x => (x -> path) }
       val deepChildrenItems = d match {
-        case a: Record =>
+        case a: Record if (!hasProbeTypeModifier(a)) =>
           a._elements.iterator.flatMap {
             case (fieldName, fieldData) =>
               collectMembersAndPaths(fieldData, s"$path.$fieldName")(collector)
           }
-        case a: Vec[_] =>
+        case a: Vec[_] if (!hasProbeTypeModifier(a)) =>
           a.elementsIterator.zipWithIndex.flatMap {
             case (fieldData, fieldIndex) =>
               collectMembersAndPaths(fieldData, s"$path($fieldIndex)")(collector)
@@ -270,7 +270,7 @@ object DataMirror {
     def iterator = {
       val myItems = collector.lift(d)
       val deepChildrenItems = d match {
-        case a: Aggregate => a.elementsIterator.flatMap { x => collectMembers(x)(collector) }
+        case a: Aggregate if (!hasProbeTypeModifier(a)) => a.elementsIterator.flatMap { x => collectMembers(x)(collector) }
         case other => Nil
       }
       myItems.iterator ++ deepChildrenItems
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -270,7 +270,8 @@ object DataMirror {
     def iterator = {
       val myItems = collector.lift(d)
       val deepChildrenItems = d match {
-        case a: Aggregate if (!hasProbeTypeModifier(a)) => a.elementsIterator.flatMap { x => collectMembers(x)(collector) }
+        case a: Aggregate if (!hasProbeTypeModifier(a)) =>
+          a.elementsIterator.flatMap { x => collectMembers(x)(collector) }
         case other => Nil
       }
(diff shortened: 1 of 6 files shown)


src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -225,7 +258,20 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
...
+    val chirrtl = ChiselStage.emitCHIRRTL(
+      new RawModule {
+        val io = IO(new Bundle {
+          val in = Input(Probe(Bool()))
+          val out = Output(Probe(Bool()))
+        })
+        io.out := io.in
+      }
+    )
+    processChirrtl(chirrtl) should contain("define io.out = io.in")
```
mwachs5, line 271: can you add a case for the same as the above example, but with user error of `io.in := io.out` ? presumably we get an error message, not sure if it is in Chisel -> .fir or .fir -> .v?

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -261,6 +307,20 @@ class ProbeSpec extends ChiselFlatSpec with Utils {
     )
   }
 
+  ":= connector with aggregates of probe" should "work" in {
+    var chirrtl = ChiselStage.emitCHIRRTL(
```
jackkoenig, line 311:
```suggestion
    val chirrtl = ChiselStage.emitCHIRRTL(
```

core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -243,12 +243,12 @@ object DataMirror {
...
+        case a: Record if (!hasProbeTypeModifier(a)) =>
           a._elements.iterator.flatMap {
             case (fieldName, fieldData) =>
               collectMembersAndPaths(fieldData, s"$path.$fieldName")(collector)
           }
-        case a: Vec[_] =>
+        case a: Vec[_] if (!hasProbeTypeModifier(a)) =>
           a.elementsIterator.zipWithIndex.flatMap {
             case (fieldData, fieldIndex) =>
               collectMembersAndPaths(fieldData, s"$path($fieldIndex)")(collector)
```
jackkoenig, line 254: Why are these here? It seems to me that DataMirror should return probes as well and perhaps the onus is on the user providing a `collector` function that excludes probes if that is the behavior they want?

dtzSiFive: This stops it from recursing /through/ probes, not excluding them, FWIW.

The `collector` you're describing is used in, e.g., `excludeProbes`, but the idea here is that `Probe`s are opaque in the sense that's being iterated over here -- they're basically `Element` in that sense, with an underlying probed type that itself may have children but the probe itself does not.

As-is no `collector` definition (AFAIK?) can use this (pre-this-PR) to provide the useful behavior of treating probes as the leaves of aggregates that they are.

All that said, what this method is """supposed""" to do independent of our immediate needs I can't speak for (such as it's expected to recurse in same way other things do), so if it's not directly needed sure we can drop it, if that seems better!

mwachs5 (commented):
maybe we should add some words to this to explain how probe-ness has to match too: https://github.com/chipsalliance/chisel/blob/main/docs/src/explanations/connectable.md#terminology

In the "structural type check" bullet are we now saying that "structural type check" includes 'probe-ness', or "alignment type check" includes 'probe-ness', or should "probe-ness type check" be its own bullet

core/src/main/scala/chisel3/reflect/DataMirror.scala
jackkoenig: > This stops it from recursing /through/ probes, not excluding them, FWIW.

Good point, thanks for pointing that out

> The collector you're describing is used in, e.g., excludeProbes, but the idea here is that Probes are opaque in the sense that's being iterated over here -- they're basically Element in that sense, with an underlying probed type that itself may have children but the probe itself does not.
>
> As-is no collector definition (AFAIK?) can use this (pre-this-PR) to provide the useful behavior of treating probes as the leaves of aggregates that they are.

This is a fair point, but I think it highlights a much larger issue (and one that we do currently run into with `OpaqueTypes` as well) that we don't really have a way to talk about "leaves" other than `Element`. I think the change you're getting at here makes a lot of sense, but I think it necessitates coming up with the new API for "leaf".

> All that said, what this method is """supposed""" to do independent of our immediate needs I can't speak for (such as it's expected to recurse in same way other things do), so if it's not directly needed sure we can drop it, if that seems better!

You've convinced me that this is the right thing to do for this PR. We probably need to prioritize finding a better way to discuss leaves though.

I want to note that `collectLeafMembers` above is probably broken as it is expressed as:
```
  def collectLeafMembers(d: Data): Seq[Data] =
    DataMirror.collectMembers(d) { case x: Element => x }.toVector
```
It will ignore Aggregate probes even though those are leaves. It's also extremely likely that anyone calling that method is immediately matching on `Element` though....

dtzSiFive: :100: , agreed and interesting/good to know about `OpaqueType`.  Hopefully soon we can get into business of working on these things and making everything awesome :rocket: .

Good catch re:`collectLeafMembers`, thank you for looking / thinking about that!! I think you're right.

docs/src/explanations/connectable.md
```diff
@@ -43,7 +43,7 @@ section: "chisel3"
    * E.g. `IO(Decoupled(Bool)).ready` is a member of the parent `IO` component
  * "relative alignment" - whether two members of the same component or Chisel type are aligned/flipped, relative to one another
    * see section [below](#alignment-flipped-vs-aligned) for a detailed definition
- * "structural type check" - Chisel type `A` is structurally equivalent to Chisel type `B` if `A` and `B` have matching bundle field names and types (`Record` vs `Vector` vs `Element`), vector sizes, `Element` types (UInt/SInt/Bool/Clock etc))
+ * "structural type check" - Chisel type `A` is structurally equivalent to Chisel type `B` if `A` and `B` have matching bundle field names and types (`Record` vs `Vector` vs `Element`), probe modifiers (probe vs nonprobe), vector sizes, `Element` types (UInt/SInt/Bool/Clock)
```
rwy7, line 46: OK, I've snuck in a bit that says, A and B should have matching probe modifiers.

dtzSiFive (approved):
LGTM, thanks!

dtzSiFive:
(6.0 milestone sound right?)

jackkoenig:
> (6.0 milestone sound right?)

Yep, Probes are new in Chisel 6

PR #3575, opened 2023-10-13
Added disallowIOCreation as a public API

azidar:
### Description

This PR introduces two APIs (for discussion) for disabling IO creation.

The first disables IO creation for the rest of the execution of the body of a module. This is in effect a way to separate IO creation from body creation, without forcing a user to declare a bundle/record of the IO's ahead of time.

An example of how this is useful is below. A bunch of trait mix-ins to RawModule define IOs. Then, you can write user code expecting modules of type `HasBlahIO`, and call associated functions to do stuff. In a way, it enables the scala type of the RawModule contain useful information for connecting to the IOs it declares.

```scala
trait HasBlahIO {
  val a, b, c  = IO(Input(Bool()))
  val d, e, f  = IO(Input(Bool()))
  def tieOff(): Unit = {
    // reference IOs declared in trait, callable by concrete class
  }
}
object HasBlahIO {
  def wireUp(params)(m: HasBlahIO): Unit = {
    m.a := ???
     ...
  }
}
trait HasOtherIO {
  val out      = IO(Output(UInt(8.W)))
}


abstract class ExampleInterface extends Module with HasBlahIO with HasOtherIO {
  disableIOCreation()
}

class Example extends ExampleInterface {
  val myReg = RegInit(0.U(8.W))
  out := myReg
  tieOff()
}

// Can also use the type of the module outside an instance
class Parent(child: () => Module) extends RawModule {
  val i = Module(child())
  // Handle HasBlahIO
  i match {
    case h: HasBlahIO =>   HasBlahIO.wireUp(params)(h)
    case _ => 
  }
}
```

Basically, if you want multi-IO-modules, but also lock down IOs, FixedIORawModule doesn't work well.

The second creates a scope such that any Chisel construction within the scope cannot add an IO. It is a mechanism for the caller to ensure that the callee code does not affect its interface. In general, its a good way to enforce the expected behavior of calling reusable generator code.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Enable users to lock-down the IO-creation of any module by calling `disallowIOCreation()`. This is useful for building chisel libraries which desire this behavior, but don't want to force a user to declare the entire IO in one bundle.

- [ ] Did you review?


sequencer:
this is kinda a location dependent api(while chisel doesn't provide a useful feature to let user permute the execution sequence.) e.g. if this API exist in a trait, user can change the sequence of mixin sequence to workaround, can you describe the use case more detail? 
one use case I can think about is the forbid user in diplomacy add `InModuleBody` IO.
another question is chisel is using IO to declare `Property`, `Probe`, is it also forbid here?
another question is if user bore from other modules, is it also forbid here?

mwachs5:
In module body io creation blocking can be done with existing fixed io and refactoring diplomacy to force people to compute their iOS first... I don't think this is the right way to achieve this for diplomacy. Agree that this could get brittle with trait mixing order... but note scala can also execute code before the existing fixed io class (maybe a deprecated feature?)

seldridge:
> but note scala can also execute code before the existing fixed io class (maybe a deprecated feature?)

I thought I made this impossible to pull off, however, I figured it it is possible. `trait FixedIOBaseModule` is sealed and its only two children, `abstract class FixedIORawModule` and `abstract class FixedIOExtModule` are `abstract class`es. A user can then only mix-in `FixedIORawModule` or `FixedIOExtModule` _first_. Put differently, as long as there is single inheritance of `FixedIOBaseModule` then everything is fine. A user can get around this by creating a trait that extends one of the children and using that to achieve multiple inheritance. 🥲 (Maybe there is a way to further lock this down?)

core/src/main/scala/chisel3/MultiClock.scala
```diff
@@ -95,3 +95,15 @@ object withReset {
   }
 
 }
+
+object withoutIO {
+
+  /** Creates a new scope that IO creation is illegal
```
seldridge, line 101: Nit: 
```suggestion
  /** Creates a new scope in which IO creation cause a runtime error
```

core/src/main/scala/chisel3/Module.scala
```diff
@@ -394,17 +394,34 @@ package experimental {
...
     /** If true, then this module is allowed to have user-created IO. */
-    private[chisel3] def isIOCreationAllowed = _isIOCreationAllowed
+    private[chisel3] def isIOCreationAllowed = _whereIOCreationIsDisallowed.isEmpty
 
     /** Disallow any more IO creation for this module. */
-    private[chisel3] def disallowIOCreation(): Unit = {
-      _isIOCreationAllowed = false
+    def disallowIOCreation()(implicit si: SourceInfo): Unit = {
+      _whereIOCreationIsDisallowed = si +: _whereIOCreationIsDisallowed
+    }
```
seldridge, line 410: This public API is duplicated by `endIOCreation`. Can only one of them be public to avoid user confusion?

seldridge: Thanks for adding `SourceInfo` to this!

core/src/main/scala/chisel3/Module.scala
```diff
@@ -394,17 +394,34 @@ package experimental {
...
+    private def allowIOCreation(): Unit = {
+      if (_whereIOCreationIsDisallowed.nonEmpty) {
+        _whereIOCreationIsDisallowed = _whereIOCreationIsDisallowed.tail
+      }
+    }
+
+    def endIOCreation()(implicit si: SourceInfo): Unit = disallowIOCreation()
```
seldridge, line 417: Nit: Scaladoc.

core/src/main/scala/chisel3/IO.scala
```diff
@@ -23,7 +23,12 @@ object IO {
     */
   def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
     val module = Module.currentModule.get // Impossible to fail
-    require(module.isIOCreationAllowed, "This module cannot have user-created IO")
+    if (!module.isIOCreationAllowed)
+      Builder.error(
+        s"This module cannot have IOs instantiated after disallowing IOs: ${module._whereIOCreationIsDisallowed
+          .map(_.makeMessage { s: String => s })
+          .mkString(",")}"
+      )
```
seldridge, line 31: I didn't realize that creating the line-caret error messages were so easy. This looks like:

```
[info] - should be able to lock down their ios in a scope *** FAILED ***
[info]   chisel3.package$ChiselException: src/test/scala/chiselTests/FixedIOModuleSpec.scala 75:25: This module cannot have IOs instantiated after disallowing IOs: @[src/test/scala/chiselTests/FixedIOModuleSpec.scala 73:17]
[info]           val other = IO(Input(UInt(1.W)))
[info]                         ^
```

src/test/scala/chiselTests/FixedIOModuleSpec.scala
```diff
@@ -20,10 +20,10 @@ class FixedIOModuleSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
     class Foo extends FixedIORawModule[Bool](Bool()) {
       val a = IO(Bool())
     }
-    val exception = intercept[IllegalArgumentException] {
-      ChiselStage.emitCHIRRTL(new Foo)
+    val exception = intercept[ChiselException] {
+      ChiselStage.emitCHIRRTL(new Foo, Array("--throw-on-first-error"))
```
seldridge, line 24: Nit: unnecessary:
```suggestion
      ChiselStage.emitCHIRRTL(new Foo)
```

seldridge (approved):
LGTM

This helps pave the way to use of `FixedIORawModule`, or other evolutions of this idea, that can be used to build stronger interfaces between modules.

I had originally thought that it would be best to start this API out as `@deprecate` given the fact that it may be somewhat dangerous to rely on. However, we can claw it back if it becomes problematic. Adam has some good points about how if you are already in a `MultiIOModule`-style pattern that it can only help. I think this makes sense. 👍

src/test/scala/chiselTests/FixedIOModuleSpec.scala
azidar: Unfortunately, it isn't unnecessary because I think chisel catches the error, then throws one big error aggregating the messages :(

PR #3581, opened 2023-10-16
Add new target API to get a target relative to a specific BaseModule.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

The first commit contains some internal refactoring of how ids get named in a RawModule, which was required for the new API. The ids generated normally are named, then any `atModuleBodyEnd` blocks are evaluated and the RawModule is closed, and finally any ids generated by `atModuleBodyEnd` blocks are named. This allows the new `toRelativeTarget` to be called within an `atModuleBodyEnd` block for the root RawModule.

The second commit actually adds the new APIs.

#### Desired Merge Strategy

#### Release Notes

The new toRelativeTarget API augments the existing toAbsoluteTarget API, to provide a mechanism to get hierarchical paths to a NamedComponent relative to a known root.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -601,6 +601,32 @@ package experimental {
       }
     }
 
+    /** Returns a FIRRTL ModuleTarget that references this object, relative to an optional root.
+      *
+      * If `root` is defined, the target is a hierarchical path starting from `root`.
+      *
+      * If `root` is not defined, the target is a hierarchical path equivalent to `toAbsoluteTarget`.
+      *
+      * @note If `root` is defined, and has not finished elaboration, this must be called within `atModuleBodyEnd`.
+      * @note The BaseModule must be a descendant of `root`, if it is defined.
+      * @note This doesn't have special handling for Views.
+      */
+    final def toRelativeTarget(root: Option[BaseModule]): IsModule = {
+      (root, _parent) match {
+        // If root was defined, and we are it, return this.
+        case (Some(_), _) if root.get == this => getTarget
+        // If root was defined, and we are not there yet, recurse up.
+        case (Some(definedRoot), Some(parent)) => parent.toRelativeTarget(root).instOf(this.instanceName, name)
+        // If root was defined, and there is no parent, the root was not an ancestor.
+        case (Some(definedRoot), None) =>
+          throwException(s"Requested .toRelativeTarget relative to ${definedRoot.name}, but it is not an ancestor")
+        // If root was not defined, and there is a parent, recurse up.
+        case (None, Some(parent)) => parent.toRelativeTarget(root).instOf(this.instanceName, name)
+        // If root was not defined, and there is no parent, return this.
+        case (None, None) => getTarget
+      }
+    }
+
     /**
       * Internal API. Returns a list of this module's generated top-level ports as a map of a String
       * (FIRRTL name) to the IO object. Only valid after the module is closed.
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -423,7 +423,11 @@ package experimental {
 
     // Fresh Namespace because in Firrtl, Modules namespaces are disjoint with the global namespace
     private[chisel3] val _namespace = Namespace.empty
-    private val _ids = ArrayBuffer[HasId]()
+
+    // Expose _ids in Chisel. The ids should almost always be accessed through getIds, but there is a use-case to access
+    // the ids directly in generateComponent.
+    private[chisel3] val _ids = ArrayBuffer[HasId]()
+
     private[chisel3] def addId(d: HasId): Unit = {
       if (Builder.aspectModule(this).isDefined) {
         aspectModule(this).get.addId(d)
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -617,11 +617,15 @@ package experimental {
         case (Some(_), _) if root.get == this => getTarget
         // If root was defined, and we are not there yet, recurse up.
         case (Some(definedRoot), Some(parent)) => parent.toRelativeTarget(root).instOf(this.instanceName, name)
+        // If root was defined, and we are at a ViewParent, use its absolute target.
+        case (Some(definedRoot), None) if this == ViewParent => ViewParent.absoluteTarget
         // If root was defined, and there is no parent, the root was not an ancestor.
         case (Some(definedRoot), None) =>
           throwException(s"Requested .toRelativeTarget relative to ${definedRoot.name}, but it is not an ancestor")
         // If root was not defined, and there is a parent, recurse up.
         case (None, Some(parent)) => parent.toRelativeTarget(root).instOf(this.instanceName, name)
+        // If root was not defined, and we are at a ViewParent, use its absolute target.
+        case (None, None) if this == ViewParent => ViewParent.absoluteTarget
         // If root was not defined, and there is no parent, return this.
         case (None, None) => getTarget
       }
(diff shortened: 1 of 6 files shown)


src/test/scala/chisel3/internal/NameCollisionSpec.scala
```diff
@@ -64,15 +64,14 @@ class NameCollisionSpec extends ChiselFlatSpec with Utils {
...
+    ((the[ChiselException] thrownBy extractCause[ChiselException] {
       ChiselStage.emitCHIRRTL(
         new Module {
           // Write to an output port that isn't assigned to a val, and so doesn't get prefixed
           IO(Output(UInt(8.W))) := 123.U
         },
         Array("--throw-on-first-error")
       )
-    }).getMessage should include("try making it a public field of the Module")
+    }).getMessage should include).regex("Assign .+ to a val, or explicitly call suggestName to seed a unique name")
```
mikeurbach, line 75: This ended up being the only change I could see related to the two-phase HasId naming in RawModule. Now that we are doing a pass of naming before `checkPorts`, the error message from check ports doesn't come up in this case. We do still get a nice error that basically says the same thing, so I think this is fine.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -128,42 +173,9 @@ abstract class RawModule extends BaseModule {
...
-              case WireBinding(_, _) =>
-                id.forceName(default = "_WIRE", _namespace)
-              // probes have their refs set eagerly
-              case _ => // don't name literals
-            }
-          } // else, don't name unbound types
-      }
+    // Take a second pass through any ids generated during atModuleBodyEnd blocks to finalize names for them.
+    for (id <- _ids.drop(numInitialIds)) {
+      nameId(id)
```
jackkoenig, line 178:
```suggestion
    for (id <- _ids.view.drop(numInitialIds)) {
      nameId(id)
```
This probably won't matter much in practice, but that `.drop` does a full copy of the `ArrayBuffer`, whereas first asking for a `.view` then makes the `drop` lazy such that it won't materialize a new datastructure and instead will just let us iterate.

I experimented with a few alternatives likes so:
```scala
//> using scala "2.13.11"

object logEvent {
  private val totals = new mutable.HashMap[String, Double]()
  def apply[R](name: String)(block: => R): R = {
    val t0 = System.nanoTime()
    val result = block
    val t1 = System.nanoTime()
    val timeMillis = (t1 - t0) / 1000000.0

    val total = timeMillis + totals.getOrElse(name, 0.0)
    totals(name) = total
    println(f"$name took $timeMillis%.1f ms (total $total%.1f ms)")
    result
  }
}

object Main extends App {
  val array = logEvent("Make ArrayBuffer") {
    val xs = mutable.ArrayBuffer.empty[String]
    for (i <- 0 until 100000000) {
      xs += s"value_$i"
    }
    xs
  }

  for (_ <- 0 until 10) {

    logEvent(".drop") {
      var i = 0L
      for (x <- array.drop(array.size - 200000)) {
        i += x.size
      }
      println(i)
    }

    logEvent(".view.drop") {
      var i = 0L
      for (x <- array.view.drop(array.size - 200000)) {
        i += x.size
      }
      println(i)
    }

    logEvent(".iterator.drop") {
      var i = 0L
      for (x <- array.iterator.drop(array.size - 200000)) {
        i += x.size
      }
      println(i)
    }

    logEvent("using indexing") {
      var i = 0L
      var idx = array.size - 200000
      while (idx < array.size) {
        i += array(idx).size
        idx += 1
      }
      println(i)
    }
  }
}
```

src/test/scala/chiselTests/ToTargetSpec.scala
```diff
@@ -60,4 +94,26 @@ class ToTargetSpec extends ChiselFlatSpec with Utils {
...
+
+    chirrtl should include("~RelativeOuterModule|RelativeOuterModule/inner:RelativeInnerModule>wire")
+  }
+
+  it should "default to the root module in the requested hierarchy" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new RelativeDefaultModule)
+
+    chirrtl should include("~RelativeDefaultModule|RelativeDefaultModule/inner:RelativeInnerModule>wire")
+  }
+
```
jackkoenig, line 111: Both of these tests have the top module as the one you're relative to (even though one uses `Some` vs. `None` as the argument). It would be good if one of the relative tests were relative to a module that is not the top.

jackkoenig (approved):
Generally looks good but please see the suggestions before merging.

src/test/scala/chiselTests/ToTargetSpec.scala
mikeurbach: 🤦 yeah, thanks.

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: ah, thanks for the experimentation and pointers! btw, is there any canonical way to know this without reading the source or doing experiments? i didn't see any mention of these concerns here: https://www.scala-lang.org/api/2.13.3/scala/collection/mutable/ArrayBuffer.html#drop(n:Int):C

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -387,6 +387,26 @@ private[chisel3] trait NamedComponent extends HasId {
...
+    *
+    * If `root` is defined, the target is a hierarchical path starting from `root`.
+    *
+    * If `root` is not defined, the target is a hierarchical path equivalent to `toAbsoluteTarget`.
+    *
+    * @note If `root` is defined, and has not finished elaboration, this must be called within `atModuleBodyEnd`.
+    * @note The NamedComponent must be a descendant of `root`, if it is defined.
+    * @note This doesn't have special handling for Views.
+    */
+  final def toRelativeTarget(root: Option[BaseModule]): ReferenceTarget = {
```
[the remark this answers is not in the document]
mikeurbach: Sounds good... I was banging my head trying to get this working and cut out the view logic. I'll add a test and handle views similarly to toAbsoluteTarget.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -597,6 +601,36 @@ package experimental {
...
+        // If root was defined, and there is no parent, the root was not an ancestor.
+        case (Some(definedRoot), None) =>
+          throwException(s"Requested .toRelativeTarget relative to ${definedRoot.name}, but it is not an ancestor")
+        // If root was not defined, and there is a parent, recurse up.
+        case (None, Some(parent)) => parent.toRelativeTarget(root).instOf(this.instanceName, name)
+        // If root was not defined, and we are at a ViewParent, use its absolute target.
+        case (None, None) if this == ViewParent => ViewParent.absoluteTarget
+        // If root was not defined, and there is no parent, return this.
+        case (None, None) => getTarget
+      }
```
jackkoenig, line 631: I feel like we could simplify this logic a bit, several of the cases seem basically duplicated, especially lines 619 and 626 as well as 621 and 628.
```scala
if (root.contains(this)) getTarget
else if (this == ViewParent) ViewParent.absoluteTarget
else (root, this._parent) match {
  case (_, Some(parent)) =>  parent.toRelativeTarget(root).instOf(this.instanceName, name)
  case (Some(definedRoot), None) =>
    throwException(s"Requested .toRelativeTarget relative to ${definedRoot.name}, but it is not an ancestor")
  case (None, None) => getTarget
}
```
Maybe? I just sketched that up really fast so it may not be correct

jackkoenig: I'll leave it to you to decide how to express this in a way that is the most clear, approved for whatever you prefer.

jackkoenig (approved):

core/src/main/scala/chisel3/Module.scala
mikeurbach: Your suggestion works great as-is, just applied my old comments to the new structure. Thanks for taking the time to think hard enough about this to make that suggestion.

PR #3593, opened 2023-10-20
Blast probes to leaf elements of aggregates

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

Probes on aggregates will now get emitted as aggregates of probes, e.g. instead of `output outRWBundleProbe : RWProbe<{ x : UInt<1>, y : UInt<1>}>`, we now emit `output outRWBundleProbe : { x : RWProbe<UInt<1>>, y : RWProbe<UInt<1>>}`.

Related probe and `BoringUtils.tap` methods have been updated to support this.

- [ ] Did you review?


core/src/main/scala/chisel3/Mem.scala
```diff
@@ -54,8 +55,8 @@ sealed abstract class MemBase[T <: Data](val t: T, val length: BigInt, sourceInf
     with SourceInfoDoc {
 
   if (t.isConst) Builder.error("Mem type cannot be const.")(sourceInfo)
-
-  requireNoProbeTypeModifier(t, "Cannot make a Mem of a Chisel type with a probe modifier.")(sourceInfo)
+  if (containsProbeTypeModifier(t))
+    Builder.error("Cannot make a Mem of a Chisel type with a probe modifier.")(sourceInfo)
```
jackkoenig, line 59:
```suggestion
  if (containsProbeTypeModifier(t)) {
    Builder.error("Cannot make a Mem of a Chisel type with a probe modifier.")(sourceInfo)
  }
```
Style-nit but GOTO FAIL has scared me into requiring curly braces for if. Maybe we could enforce with scalafmt/scalafix.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -183,20 +184,31 @@ abstract class RawModule extends BaseModule {
   }
   private[chisel3] val stagedSecretCommands = collection.mutable.ArrayBuffer[Command]()
 
+  private def secretProbeDefines(left: Data, right: Data)(implicit si: SourceInfo): Seq[Command] = {
+    collectLeafMembers(left).zip(collectLeafMembers(right)).map {
```
jackkoenig, line 188: What if `left` and `right` are not the same type? Is that checked before `secretConnection` is called? You may want to consider using one of the collect functions that iterates over two types at the same tame, eg. `collectMembersOverAll` and then you can error on mismatched fields. All that being said, it really feels like this logic should be common with regular connections. I know you're just tweaking what was already there, but why is this logic necessary vs. reusing something in normal connections?

core/src/main/scala/chisel3/probe/ProbeValue.scala
```diff
@@ -2,24 +2,46 @@
...
   protected def apply[T <: Data](source: T, writable: Boolean)(implicit sourceInfo: SourceInfo): T = {
     requireIsHardware(source)
     // construct probe to return with cloned info
-    val clone = if (writable) RWProbe(source.cloneType) else Probe(source.cloneType)
+    val clone = if (writable) RWProbe(source.cloneTypeFull) else Probe(source.cloneTypeFull)
     clone.bind(OpBinding(Builder.forcedUserModule, Builder.currentWhen))
-    if (writable) {
-      clone.setRef(RWProbeExpr(source.ref))
+    // create reference for clone
+    val cloneRef = if (source.isInstanceOf[Aggregate]) {
```
jackkoenig, line 22: You should match on the type rather than using `.isInstanceOf`

core/src/main/scala/chisel3/probe/ProbeValue.scala
```diff
@@ -2,24 +2,46 @@
...
+      // intermediate probe to hook up to aggregate elements
+      val intermediate = Wire(Output(if (writable) RWProbe(source.cloneTypeFull) else Probe(source.cloneTypeFull)))
+      collectLeafMembers(intermediate).zip(collectLeafMembers(source)).foreach {
```
jackkoenig, line 25: similar to my comments above but obviously you don't need to check correctness. That being said, why can't this code use the same logic as in `secretProbeDefines` (which still feels to me like it should just be the normal connections logic. It's the same principle even if this code won't be able to exercise all code paths.

Also, it would then just handle the leaf case with the same logic I think.

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -47,25 +54,31 @@ package object probe extends SourceInfoDoc {
...
+      val sourceElems = collectLeafMembers(source)
+      val intermediateElems = collectLeafMembers(intermediate)
+      sourceElems.zip(intermediateElems).foreach { case (e, t) => t :#= do_read(e) }
+      intermediate.suggestName("probe_read")
+      intermediate.ref
+    } else {
+      requireHasProbeTypeModifier(source)
+      ProbeRead(source.ref)
+    }
+    clone.setRef(cloneRef)
```
jackkoenig, line 73: Similar comment to above that you should match rather than using `isInstanceOf`

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -89,27 +102,45 @@ package object probe extends SourceInfoDoc {
   }
 
   /** Override existing driver of a writable probe on initialization. */
-  def forceInitial(probe: Data, value: Data)(implicit sourceInfo: SourceInfo): Unit = {
-    requireHasWritableProbeTypeModifier(probe, "Cannot forceInitial a non-writable Probe.")
-    pushCommand(ProbeForceInitial(sourceInfo, probe.ref, padDataToProbeWidth(value, probe).ref))
+  def forceInitial[T <: Data](probe: T, value: T)(implicit sourceInfo: SourceInfo): Unit = {
+    collectLeafMembers(probe).zip(collectLeafMembers(value)).foreach {
```
jackkoenig, line 106: Same comment about iterating on these together (we also need to check their types which you can do while iterating).

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -89,27 +102,45 @@ package object probe extends SourceInfoDoc {
...
+      pushCommand(ProbeReleaseInitial(sourceInfo, p.ref))
+    }
   }
 
   /** Override existing driver of a writable probe. */
-  def force(clock: Clock, cond: Bool, probe: Data, value: Data)(implicit sourceInfo: SourceInfo): Unit = {
-    requireHasWritableProbeTypeModifier(probe, "Cannot force a non-writable Probe.")
-    pushCommand(ProbeForce(sourceInfo, clock.ref, cond.ref, probe.ref, padDataToProbeWidth(value, probe).ref))
+  def force[T <: Data](clock: Clock, cond: Bool, probe: T, value: T)(implicit sourceInfo: SourceInfo): Unit = {
+    collectLeafMembers(probe).zip(collectLeafMembers(value)).foreach {
```
jackkoenig, line 127: Same comment about iterating together and checking types.

core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -57,6 +57,20 @@ object DataMirror {
...
+    */
+  def containsProbeTypeModifier(x: Data): Boolean = {
+    x match {
+      case a: Aggregate =>
+        a.getElements.foldLeft(false)((accumulator, elem) => accumulator || containsProbeTypeModifier(elem))
+      case d => {
+        hasProbeTypeModifier(d)
+      }
+    }
+  }
```
jackkoenig, line 72: Can this be made tail recursive? Also it would be better to early out rather than using a fold (use `.exists` or `.forall`).

jackkoenig (commented):
Related to my comments, do we have any tests for what the errors look like when you probe, force, release, etc. aggregates of mismatched types?

core/src/main/scala/chisel3/RawModule.scala
debs-sifive: > should be common with regular connections.

Do you have pointers to the code that overlaps? I was rolling my own b/c I wasn't able to find logic that does something similar.

I'm also not sure how to use `collectMembersOverAll`... Do you know where I should be adding in the `HasMatchingZipOfChildren` trait?

cc @azidar in case you have any ideas on how to best integrate this chunk of code

mwachs5: I believe there is a `DataMirror.collectMembersOverMatches` here that we should be using rather than the raw zip of the members... @azidar could advise?

core/src/main/scala/chisel3/probe/ProbeValue.scala
mwachs5: yes again i think the `DataMirror.collectMembersOverMatches` is the suggested API here

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -8,19 +8,23 @@ import chisel3.internal.Builder.pushCommand
...
   private[chisel3] def setProbeModifier[T <: Data](data: T, probeInfo: Option[ProbeInfo]): Unit = {
     probeInfo.foreach { _ =>
-      data.probeInfo = probeInfo
-      data match {
-        case a: Aggregate =>
-          a.elementsIterator.foreach { e => setProbeModifier(e, probeInfo) }
-        case _ => // do nothing
+      collectAllMembers(data).foreach { e =>
+        e match {
+          case v: Vec[_]    => setProbeModifier(v.sample_element, probeInfo)
```
mwachs5, line 24: i wish you could just grab the leaves... for the leaves of the vector does it do this twice?

core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -30,15 +34,18 @@ package object probe extends SourceInfoDoc {
...
       Builder.error("Cannot define a probe on a non-equivalent type.")
     }
-    requireHasProbeTypeModifier(sink, "Expected sink to be a probe.")
-    requireHasProbeTypeModifier(probeExpr, "Expected source to be a probe expression.")
-    if (sink.probeInfo.get.writable) {
-      requireHasWritableProbeTypeModifier(
-        probeExpr,
-        "Cannot use a non-writable probe expression to define a writable probe."
-      )
+    collectLeafMembers(sink).zip(collectLeafMembers(probeExpr)).foreach {
```
mwachs5, line 37: collectMembersOverMatches again

core/src/main/scala/chisel3/probe/package.scala
debs-sifive: getting the leaves of vecs doesn't grab it's `sampleElement`, which is what the converter uses to get the element type. I'll add in a comment explaining what's going on here, but I don't know how else to work around this :(

debs-sifive: This also shouldn't repeat on Vec leaves, since those will all hit the last case.

The first Vec case sets probe modifiers on the sample elements of all internal Vecs.

core/src/main/scala/chisel3/RawModule.scala
azidar: Yeah, I think it should look something like:

```
collectMembersOverMatches(left, right) {
  case (l, r) => (hasProbeTypeModifier(l), hasProbeTypeModifier(r)) match {
          case (true, true)                              => ProbeDefine(si, l.lref, Node(r))
          case (true, false) if l.probeInfo.get.writable => ProbeDefine(si, l.lref, RWProbeExpr(Node(r)))
          case (true, false)                             => ProbeDefine(si, l.lref, ProbeExpr(Node(r)))
          case (false, true)                             => Connect(si, l.lref, ProbeRead(Node(r)))
          case (false, false)                            => Connect(si, l.lref, Node(r))
    }
}
```

I believe `HasMatchingZipOfChildren` should already be defined?

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -382,27 +382,25 @@ private[chisel3] object Converter {
     case t: Vec[_] =>
       val childClearDir = clearDir ||
         t.specifiedDirection == SpecifiedDirection.Input || t.specifiedDirection == SpecifiedDirection.Output
-      // if Vector is a probe, don't emit Probe<...> on its elements
-      fir.VectorType(extractType(t.sample_element, childClearDir, info, checkProbe, true, typeAliases), t.length)
+      fir.VectorType(extractType(t.sample_element, childClearDir, info, true, true, typeAliases), t.length)
```
azidar, line 385: Can you remove checkProbe entirely?

core/src/main/scala/chisel3/RawModule.scala
debs-sifive: Can you take a look at the new pattern using `collectMembersOverMatches`? E.g.:
```scala
    collectMembersOverMatches(probe.asInstanceOf[Data], value.asInstanceOf[Data]) {
      case (p: Aggregate, v: Aggregate) => // do nothing; only operate on leaves
      case (p, v) => // actual use case
```

PR #3596, opened 2023-10-21
a Pure LLVM Lit test framework

sequencer:
I wanna use #3596 to subsume #3410, see [this](https://github.com/chipsalliance/chisel/blob/ec94af02c13d9222baac943e8a3b437303a00d47/filecheck/tests/EmptyModule.sc) simple test:
```scala
//> RUN: scala-cli $THIS | firtool -format=fir
import chisel3._
import circt.stage.ChiselStage
class EmptyModule extends Module
println(ChiselStage.emitCHIRRTL(new EmptyModule))
```
In the build system, it will be processed into:
```scala
//>using scala "2.13.12"
[...]

PR #3599, opened 2023-10-24
Add llvm-lit into nix devShell

Avimitin:
Tested with llvm lit built-in example tests directory.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/nix/llvm-lit.nix
+++ b/nix/llvm-lit.nix
@@ -1,0 +1,50 @@
+{ fetchFromGitHub,
+  python3,
+  stdenv, cmake, ninja, makeWrapper }:
+let
+  src = fetchFromGitHub {
+    owner = "llvm";
+    repo = "llvm-project";
+    rev = "llvmorg-17.0.3";
+    hash = "sha256-fXkYSYwhMeY3nLuuSlO2b/TpTxa+61zxmsfFg7a2Gbo=";
+  };
+
+  llvm-lit = python3.pkgs.buildPythonPackage {
+    inherit src;
+
+    pname = "llvm-lit";
+    version = "17.0.3";
+    format = "setuptools";
+    sourceRoot = "source/llvm/utils/lit";
+    propagatedBuildInputs = [
+      python3.pkgs.setuptools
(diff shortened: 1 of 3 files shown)


sequencer:
This is required by #3596 for my development environment. We don't have CI for it ;p

dtzSiFive:
Maybe use nixpkgs-packaged lit? https://github.com/NixOS/nixpkgs/blob/75a52265bda7fd25e06e3a67dee3f0354e73243c/pkgs/development/python-modules/lit/default.nix .

`pythonPackages.lit` should work-- until channel gets that it'll be 15.0.6.

Anyway at very least can fetch this from official Pypi source instead of pulling out of the LLVM source tree which is a bit nicer :+1: .

nix/llvm-lit.nix
```diff
@@ -0,0 +1,50 @@
...
+
+  llvm-lit = python3.pkgs.buildPythonPackage {
+    inherit src;
+
+    pname = "llvm-lit";
+    version = "17.0.3";
+    format = "setuptools";
+    sourceRoot = "source/llvm/utils/lit";
+    propagatedBuildInputs = [
+      python3.pkgs.setuptools
```
dtzSiFive, line 20: I forget the rules for this for python (diff than normal packages :sweat: ) but offhand why propagate setuptools, for use of this as a build-dep of some other python thing?

sequencer:
Yeah, I just found that, CC @Avimitin for contribution points :)

Avimitin:
> Anyway at very least can fetch this from official Pypi source instead of pulling out of the LLVM source tree which is a bit nicer 👍 .

But finally we still need to build `FileCheck` and `not`, and maybe other binary dependencies. This derivation can provide a lit wrapper to handle this dependencies.

nix/llvm-lit.nix
Avimitin: Yes, this is my mistake. I should packaged it as an application not a library.

PR #3600, opened 2023-10-24
Bugfix: foo and foo_ names cause incorrect prefix calculation

azidar:
### Problem

When naming `foo`, then `foo_`, the calculation for the prefix of `foo_` is incorrect. It should not have a prefix, but instead thinks the entire `foo_` is the prefix. When the prefix is stripped to find the index, you get a `java.lang.NumberFormatException: For input string: ""`, as it tries to convert the empty string to an `Int`.

This bugfix checks the case where a prefix is calculated, but would return an empty string if stripped.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Naming `foo` and then `foo_` will no longer trigger an error.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -57,7 +57,7 @@ private[chisel3] class Namespace(keywords: Set[String], separator: Char = '_') {
       i -= 1
     }
     // Will get i == 0 for all digits or _\d+ with empty prefix, those have no prefix so returning 0 is correct
-    if (i == n.size) 0 // no digits
+    if (i == n.size || i == (n.size - 1)) 0 // no digits
     else if (n(i) != separator) 0 // no _
     else i
   }
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -57,7 +57,7 @@ private[chisel3] class Namespace(keywords: Set[String], separator: Char = '_') {
       i -= 1
     }
     // Will get i == 0 for all digits or _\d+ with empty prefix, those have no prefix so returning 0 is correct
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -57,7 +57,7 @@ private[chisel3] class Namespace(keywords: Set[String], separator: Char = '_') {
       i -= 1
     }
     // Will get i == 0 for all digits or _\d+ with empty prefix, those have no prefix so returning 0 is correct
-    if (i == n.size) 0 // no digits
+    if (i == n.size || i == (n.size - 1)) 0 // no digits
```
jackkoenig, line 60:
```suggestion
    if (i >= (n.size - 1)) 0 // no digits
```

src/test/scala/chisel3/internal/NamespaceSpec.scala
```diff
@@ -78,4 +78,10 @@ class NamespaceSpec extends AnyFlatSpec {
     name("x") should be("x_2")
     name("x_0") should be("x_0_1")
   }
+
+  they should "support resolving collisions between <name> and <name>_" in {
+    val namespace = Namespace.empty
+    namespace.name("mouth") should be("mouth")
+    namespace.name("mouth_") should be("mouth_")
```
jackkoenig, line 85: What happens if we ask for `mouth` and `mouth_` again? Might be good to include in the test.

jackkoenig (approved):
LGTM although left a suggested improvement

src/test/scala/chisel3/internal/NamespaceSpec.scala
azidar: Added it, creates `mouth_1` and `mouth__1`

PR #3601, opened 2023-10-24
Updated Path property target type serialization.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

This includes two changes to how Path property target types are serialized:
* For ModuleTargets, use the OMInstanceTarget target type, like for InstanceTargets
* Add a new API to request "member" target types

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Path.scala
+++ b/core/src/main/scala/chisel3/properties/Path.scala
@@ -10,12 +10,12 @@ import firrtl.ir.{PathPropertyLiteral}
 /** Represent a Path type for referencing a hardware instance or member in a Property[Path]
   */
 sealed abstract class Path {
-  private[chisel3] def toTarget():     IsMember
-  private[chisel3] def isMemberTarget: Boolean
+  private[chisel3] def toTarget():   IsMember
+  private[chisel3] def isMemberPath: Boolean
 
   private[chisel3] def convert(): PathPropertyLiteral = {
     val target = toTarget()
-    val targetType = if (isMemberTarget) {
+    val targetType = if (isMemberPath) {
       target match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/properties/Path.scala
```diff
@@ -27,33 +36,41 @@ object Path {
...
+      def isMemberTarget: Boolean = _isMemberTarget
     }
   }
 
   /** Construct a Path that refers to a Data
     */
-  def apply(data: Data): Path = {
+  def apply(data: Data): Path = apply(data, false)
+  def apply(data: Data, _isMemberTarget: Boolean): Path = {
     new Path {
```
jackkoenig, line 51:
```suggestion
  def apply(data: Data, isMemberTarget: Boolean): Path = {
    val _isMemberTarget = isMemberTarget // avoid name shadowing below
    new Path {
```
Since this is a public API, it should not have an `_`.

jackkoenig:
Meta point, should this be exposed as `isMemberPath`? Is there a more descriptive name we could give to the behavior? @albertchen-sifive @mikeurbach

core/src/main/scala/chisel3/properties/Path.scala
mikeurbach: 👍 thanks for pointing it out, I forgot this was a public API.

mikeurbach:
> Meta point, should this be exposed as isMemberPath? Is there a more descriptive name we could give to the behavior?

I think `isMemberPath` is better. Maybe for full granularity it could be something like `isMemberTargetType`, but I don't know how useful "target type" is in the name.

albertchen-sifive (approved):

PR #3602, opened 2023-10-25
dummy commit

seldridge:
Test which CI and firtool this uses.


seldridge:
Thanks for cleaning up this experiment!
p

PR #3605, opened 2023-10-25
remove clock and cond from probe force/release API

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Remove `clock` and `cond` from probe `force` and `release` methods.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -101,14 +101,18 @@ package object probe extends SourceInfoDoc {
   }
 
   /** Override existing driver of a writable probe. */
-  def force(clock: Clock, cond: Bool, probe: Data, value: Data)(implicit sourceInfo: SourceInfo): Unit = {
+  def force(probe: Data, value: Data)(implicit sourceInfo: SourceInfo): Unit = {
     requireHasWritableProbeTypeModifier(probe, "Cannot force a non-writable Probe.")
+    val clock = Builder.forcedClock
+    val cond = Module.disableOption.map(!_.value).getOrElse(true.B)
     pushCommand(ProbeForce(sourceInfo, clock.ref, cond.ref, probe.ref, padDataToProbeWidth(value, probe).ref))
   }
 
   /** Release driver on a probe. */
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/probe/package.scala
```diff
@@ -101,14 +101,18 @@ package object probe extends SourceInfoDoc {
   }
 
   /** Override existing driver of a writable probe. */
```
mwachs5, line 103: while we are add it can we scaladoc the probe and value fields? SHould we mention in the scaladoc that the when condition is used...?

mwachs5 (commented):
looks good to me! just always lobbying for more scaladoc :)

core/src/main/scala/chisel3/probe/package.scala
debs-sifive: IMO, I don't think we should mention the `when` stuff in the ScalaDoc b/c that's more of a firtool implementation detail. But will def add the other fields.

jackkoenig: I wouldn't call it an implementation detail. The semantic is that these ops only occur when the when condition is true.

For example, the ScalaDoc for printf says:

> Prints a message every cycle. If defined within the scope of a [when](https://s01.oss.sonatype.org/service/local/repositories/snapshots/archive/org/chipsalliance/chisel_2.13/6.0.0-M3+55-6dbbc73c-SNAPSHOT/chisel_2.13-6.0.0-M3+55-6dbbc73c-SNAPSHOT-javadoc.jar/!/chisel3/when$.html) block, the message will only be printed on cycles that the when condition is true.
>
> Does not fire when in reset (defined as the encapsulating Module's reset). If your definition of reset is not the encapsulating Module's reset, you will need to gate this externally.

I think we need something similar in these APIs, except instead of not firing when in reset, it doesn't fire "until reset has been asserted and then deasserted" through the Disable API (link to Disable).

debs-sifive: Gotcha, that makes sense.

jackkoenig (approved):

PR #3611, opened 2023-10-30
set sample_element's probeInfo for Vecs

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

For `Vec`s, also set the `probeInfo` of their `sample_element`, which is used to determine its element type.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/probe/package.scala
+++ b/core/src/main/scala/chisel3/probe/package.scala
@@ -11,16 +11,19 @@ import chisel3.experimental.SourceInfo
 import chisel3.reflect.DataMirror.{checkTypeEquivalence, hasProbeTypeModifier}
 
 import scala.language.experimental.macros
+import chisel3.reflect.DataMirror
 
 package object probe extends SourceInfoDoc {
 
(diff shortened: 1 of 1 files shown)


azidar (approved):
While I approve, i'd like to see a test case that covers this error.

dtzSiFive:
Can you add a test to check this works as intended and continues to? Thanks!

debs-sifive:
Working on it! Accidentally merged too early, so it will be coming in a separate PR.

debs-sifive:
OK actually after talking with Adam more, the `sample_element` issue no longer affects the FIRRTL that gets emitted — we were only seeing it before with the probe aggregate-blasting approach because that required probe modifiers to get nested inside of Vecs, which no longer happens/is legal to get emitted.

So it's not something we can detect anymore in the emitted firrtl... that said, I think we should keep this fix in place because it makes sense for the `sample_element` to have the same probe information to the actual vector elements in case it gets used down the line. Which I know is not going to be super satisfying b/c of this whole "source of truth" on the vector element type getting mutated in this random place :(

PR #3615, opened 2023-11-02
Add type ascription to implicits from instantiable

jackkoenig:
This is needed by users of Chisel using Scala 2.13.12

### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

This fixes Scala compiler warnings in Scala 2.13.12 requiring type ascriptions on public implicit definitions.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [ ] Did you mark the proper milestone (Bug fix: `3.5.x`, `3.6.x`, or `5.x` depending on impact, API modification or big change: `6.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3616, opened 2023-11-02
Add type ascriptions to public implicit definitions

jackkoenig:
Built on top of https://github.com/chipsalliance/chisel/pull/3615 so should be merged after it. This shouldn't be backported which is why it is a separate PR.

This makes extensive use of the ["Aux Pattern"](https://stackoverflow.com/questions/43900674/understanding-the-aux-pattern-in-scala-type-system), which is just a way of using aliases to make it easier to write type ascriptions for types with abstract type members.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?


core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
```diff
@@ -294,7 +300,7 @@ object Lookupable {
     def instanceLookup[A](that:   A => B, instance:   Instance[A]):   C = that(instance.proto)
   }
 
-  implicit def lookupInstance[B <: BaseModule](implicit sourceInfo: SourceInfo) =
+  implicit def lookupInstance[B <: BaseModule](implicit sourceInfo: SourceInfo): Simple[Instance[B]] =
```
jackkoenig, line 303: I'm wondering if we should include this `Simple` alias or just use `Aux` here making the type ascription `: Aux[Instance[B], Instance[B]]`. It's not _that_ bad and reduces the number of type aliases which may be more clear. Any thoughts?

albertchen-sifive: I'm fine with the extra alias. It could be useful for users who define their own `Lookupable`  instances.

albertchen-sifive (approved):

PR #3619, opened 2023-11-03
Add SparseVec

seldridge:
Add a new utility, SparseVec.  This exposes a normal Vec-like dynamic access API.  However, the dynamic access is implemented one of three ways:

  1. Using a binary decoder
  2. Using a one-hot decoder
  3. Using when statements

This is added to better support a design pattern ("vec--decoder pattern") where a large, sparse vector is used to describe a decoder.  This pattern, if unchecked and used in a parametric generator, can cause the vector to grow extremely large and cause backend tools to complain.

Specifically, Cadence tooling is known to set limits on the addressable range of a "memory" and will error if these are exceeded.  This vec--decoder pattern can easily exceed this if designers are not careful.

#### Release Notes

Add `SparseVec` API.

### DIFF
--- a/integration-tests/src/test/scala/chiselTest/util/SparseVecSpec.scala
+++ b/integration-tests/src/test/scala/chiselTest/util/SparseVecSpec.scala
@@ -1,0 +1,370 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiselTests.util
+
+import chisel3._
+import chisel3.testers.BasicTester
+import chisel3.util.{log2Up, Counter, SparseVec}
+import chisel3.util.SparseVec.{DefaultValueBehavior, Lookup, OutOfBoundsBehavior}
+import chiselTests.{ChiselFlatSpec, Utils}
+import _root_.circt.stage.ChiselStage
+import java.util.ResourceBundle
+
+/** Tester that checks that a a [[SparseVec]] behaves _exactly_ like a dynamic
+  * index into a [[DontCare]]-initialized dense [[Vec]].  This checks
+  * out-of-bounds behavior by indexing to the maximum addressable size of both
+  * representations.  E.g., if the size is 3, this will check indices `[0, 1, 2,
+  * 3]`.
+  *
+  * @param size the size of the vecs
+  * @param tpe the type of the vecs
+  * @param mapping a mapping of index to value
+  */
+class SparseVecDynamicIndexEquivalenceTest(
+  size:    Int,
+  tpe:     UInt,
+  mapping: Seq[(Int, UInt)],
+  debug:   Boolean = false)
+    extends BasicTester {
+
+  // The number of indices that needs to be checked.  This is larger than `size`
+  // if `size` is not a power of 2.  This is done to check out-of-bounds
+  // behavior.
+  private val paddedSize = BigInt(log2Up(size)).pow(2).toInt
+
+  // This is the reference vector that will be dynamically indexed into.
+  // Initialize all elements to DontCare.  Then set specific ones using the
+  // provided expected mapping.
+  private val denseVec = {
+    val w = Wire(Vec(size, tpe))
+    w.foreach(_ := DontCare)
+    mapping.foreach {
+      case (index, data) =>
+        w(index) := data
+    }
+    w
+  }
+
+  // Create a wire SparseVec and initialize it to the values in the mapping.
+  private val sparseVec = Wire(
+    new SparseVec(size, tpe, mapping.map(_._1), DefaultValueBehavior.DynamicIndexEquivalent, OutOfBoundsBehavior.First)
+  )
+  sparseVec.elements.values.zip(mapping.map(_._2)).foreach { case (a, b) => a :<>= b }
+
+  // Access the dense vector and the sparse vector, using all of the access
+  // types, and make sure that the results are exactly the same.
+  private val (index, wrap) = Counter(0 until paddedSize)
+  private val failed = RegInit(Bool(), false.B)
+  private val reference = denseVec(index)
+  private val sparseVecResults = Seq(Lookup.Binary, Lookup.OneHot, Lookup.IfElse).map(sparseVec(index, _))
+  if (debug) {
+    when(RegNext(reset.asBool)) {
+      printf("index, dense, binary, onehot, ifelse\n")
+    }
+    printf("%x: %x, %x, %x, %x", index, reference, sparseVecResults(0), sparseVecResults(1), sparseVecResults(2))
+  }
+  when(sparseVecResults.map(_ =/= reference).reduce(_ || _)) {
+    failed := true.B
+    if (debug)
+      printf(" <-- error")
+    else
+      assert(false.B)
+  }
+  if (debug)
+    printf("\n")
+
+  when(wrap) {
+    stop()
+  }
+}
+
+/** This test checks that a [[SparseVec]] returns expected values.  A
+  * [[SparseVec]] of size, type, and configuration parameters is created and
+  * initialized with a mapping.  It is then checked against an expected sequence
+  * of index--value pairs.  Using an expected sequence that checks indices not
+  * in the mapping, either default or out-of-bounds behaviors can be checked.
+  *
+  * @param size the size of the SparseVec
+  * @param tpe the element type of the SparseVec
+  * @param defaultValueBehavior the default value behavior
+  * @param outOfBoundsBehavior the out-of-bounds behavior
+  * @param mapping the index--value mapping that is used to initialize the vec
+  * @param expected the expected values that are read out of the vec at each index
+  */
+class SparseVecTest(
+  size:                 Int,
+  tpe:                  UInt,
+  defaultValueBehavior: DefaultValueBehavior.Type,
+  outOfBoundsBehavior:  OutOfBoundsBehavior.Type,
+  mapping:              Seq[(Int, UInt)],
+  expected:             Seq[(Int, Data)],
+  debug:                Boolean = false)
+    extends BasicTester {
+  // Create a wire SparseVec and initialize it to the values in the mapping.
+  private val sparseVec = Wire(new SparseVec(size, tpe, mapping.map(_._1), defaultValueBehavior, outOfBoundsBehavior))
+  sparseVec.elements.values.zip(mapping.map(_._2)).foreach { case (a, b) => a :<>= b }
+
+  class TestBundle extends Bundle {
+    val index = UInt()
+    val value = UInt()
+  }
+
+  val tests = Wire(
+    Vec(
+      expected.size,
+      new TestBundle
+    )
+  )
+  expected.zipWithIndex.foreach {
+    case ((index, value), testNumber) =>
+      tests(testNumber).index := index.U
+      tests(testNumber).value := value
+  }
+
+  // Access the dense vector and the sparse vector, using all of the access
+  // types, and make sure that the results are exactly the same.
+  private val (index, wrap) = Counter(0 until tests.size)
+  private val failed = RegInit(Bool(), false.B)
+  private val reference = tests(index)
+  private val sparseVecResults = Seq(Lookup.Binary, Lookup.OneHot, Lookup.IfElse).map(sparseVec(reference.index, _))
+  if (debug) {
+    when(RegNext(reset.asBool)) {
+      printf("index, dense, binary, onehot, ifelse\n")
+    }
+    printf(
+      "%x: %x, %x, %x, %x",
+      reference.index,
+      reference.value,
+      sparseVecResults(0),
+      sparseVecResults(1),
+      sparseVecResults(2)
+    )
+  }
+  when(sparseVecResults.map(_ =/= reference.value).reduce(_ || _)) {
+    failed := true.B
+    if (debug)
+      printf(" <-- error")
+    else
+      assert(false.B)
+  }
+  if (debug)
+    printf("\n")
+
+  when(wrap) {
+    when(RegNext(true.B)) {
+      stop()
+    }
+  }
+
+}
+
+class SparseVecSpec extends ChiselFlatSpec with Utils {
+  "SparseVec equivalence to Dynamic Index" should "work for a complete user-specified mapping" in {
+    assertTesterPasses(
+      new SparseVecDynamicIndexEquivalenceTest(
+        4,
+        UInt(3.W),
+        Seq(
+          0 -> 1.U,
+          1 -> 2.U,
+          2 -> 3.U,
+          3 -> 4.U
+        )
+      )
+    )
+  }
+
+  it should "work for a mapping that includes default values" in {
+    assertTesterPasses(
+      new SparseVecDynamicIndexEquivalenceTest(
+        4,
+        UInt(3.W),
+        Seq(
+          0 -> 1.U,
+          1 -> 2.U,
+          3 -> 4.U
+        )
+      )
+    )
+  }
+
+  it should "work for a mapping that includes out-of-bounds accesses" in {
+    assertTesterPasses(
+      new SparseVecDynamicIndexEquivalenceTest(
+        3,
+        UInt(3.W),
+        Seq(
+          0 -> 1.U,
+          1 -> 2.U,
+          2 -> 3.U
+        )
+      )
+    )
+  }
+
+  it should "work for a mapping that includes out-of-bounds accesses and no zeroth element" in {
+    assertTesterPasses(
+      new SparseVecDynamicIndexEquivalenceTest(
+        3,
+        UInt(3.W),
+        Seq(
+          1 -> 2.U,
+          2 -> 3.U
+        )
+      )
+    )
+  }
+
+  "SparseVec" should "work for a complete user-specified mapping" in {
+    val mapping = Seq(
+      0 -> 1.U,
+      1 -> 2.U,
+      2 -> 3.U,
+      3 -> 4.U
+    )
+    assertTesterPasses(
+      new SparseVecTest(
+        4,
+        UInt(3.W),
+        DefaultValueBehavior.Indeterminate,
+        OutOfBoundsBehavior.Indeterminate,
+        mapping,
+        expected = mapping
+      )
+    )
+  }
+
+  // This test is only checking that the indeterminate values didn't screw
+  // anything up.  We can't actually check for an indeterminate value as it
+  // could be anything.
+  it should "work for a mapping that includes default values with indeterminate behavior" in {
+    val mapping = Seq(
+      0 -> 1.U,
+      1 -> 2.U,
+      3 -> 4.U
+    )
+    assertTesterPasses(
+      new SparseVecTest(
+        4,
+        UInt(3.W),
+        DefaultValueBehavior.Indeterminate,
+        OutOfBoundsBehavior.Indeterminate,
+        mapping,
+        expected = mapping
+      )
+    )
+  }
+
+  it should "work for a mapping that includes default values" in {
+    val mapping = Seq(
+      0 -> 1.U,
+      1 -> 2.U,
+      3 -> 4.U
+    )
+    assertTesterPasses(
+      new SparseVecTest(
+        4,
+        UInt(3.W),
+        DefaultValueBehavior.UserSpecified(7.U),
+        OutOfBoundsBehavior.Indeterminate,
+        mapping,
+        expected = mapping :+ (2 -> 7.U)
+      )
+    )
+  }
+
+  // As above, there's nothing to test here other than the values put in we get
+  // out.
+  it should "work for a mapping that includes indeterminate out-of-bounds behvaior" in {
+    val mapping = Seq(
+      0 -> 1.U,
+      1 -> 2.U,
+      2 -> 3.U
+    )
+    assertTesterPasses(
+      new SparseVecTest(
+        3,
+        UInt(3.W),
+        DefaultValueBehavior.Indeterminate,
+        OutOfBoundsBehavior.Indeterminate,
+        mapping,
+        expected = mapping
+      )
+    )
+  }
+
+  it should "work for a mapping that includes \"first\" out-of-bounds behavior" in {
+    val mapping = Seq(
+      0 -> 1.U,
+      1 -> 2.U,
+      2 -> 3.U
+    )
+    assertTesterPasses(
+      new SparseVecTest(
+        3,
+        UInt(3.W),
+        DefaultValueBehavior.Indeterminate,
+        OutOfBoundsBehavior.First,
+        mapping,
+        expected = mapping :+ (3 -> mapping(0)._2)
+      )
+    )
+  }
+
+  it should "work for an empty mapping" in {
+    val mapping = Seq.empty[(Int, UInt)]
+    assertTesterPasses(
+      new SparseVecTest(
+        2,
+        UInt(3.W),
+        DefaultValueBehavior.UserSpecified(7.U),
+        OutOfBoundsBehavior.First,
+        mapping,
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,114 @@
...
+
+package chisel3.util
+
+import chisel3._
+import chisel3.experimental.SourceInfo
+import chisel3.util.{log2Up, BitPat, Mux1H}
+import chisel3.util.experimental.decode.{decoder, TruthTable}
+import scala.collection.mutable.{LinkedHashMap, LinkedHashSet}
+
+sealed trait SparseVec[A] {
```
sequencer, line 11: I think it can be a Bundle with custom fields for better Verilog

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,114 @@
...
+
+    val valuesAddrWidth = log2Up(values.size)
+    values(0) := DontCare
+    val ttable = TruthTable(
+      valueMap.zipWithIndex.map {
+        case ((k, v), i) =>
+          values(i + 1) := v
+          BitPat(k.U(addrWidth.W)) -> BitPat((i + 1).U(valuesAddrWidth.W))
+      },
+      BitPat.N(addrWidth)
```
sequencer, line 39: Default to N? What if input address is not in the Sparse Vector? Should we assert it out?

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,114 @@
...
+sealed trait SparseVec[A] {
+
+  def apply(addr: UInt)(implicit _sourceInfo: SourceInfo): A
+
+}
+
+object SparseVec {
+
+  def binary[A <: Data](
+    size:    Int,
```
sequencer, line 20: Is possible for size being inferred from mapping.size?

sequencer (commented):
Some nitpicking ;p

src/main/scala/chisel3/util/SparseVec.scala
seldridge: @jackkoenig suggested making this a `Record` and having it support writes. While I think this is possible, I'm concerned that it's subtly something different. Specifically, the `SparseVec` currently sets a default value for reading if you read something that isn't in the `Vec`. However, I'm not sure how that should work for writing... are the writes dropped or are writes to values not called out visible?

Regardless, I'll play around with making the internal storage a `Record`.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: The default value should probably be configurable. This `SparseVec` is intended to replace this pattern for something like `sourceStall`: https://github.com/chipsalliance/rocket-chip/blob/f11fb1aea0033cf87aebcc9e5d3961810d1dcf57/src/main/scala/tilelink/ToAXI4.scala#L106 We've run into issues where this `Vec` gets comically large for certain parameterizations.

The test in this PR is modeled off of this example. Roughly:

1. You create a `Vec` `Wire`.
2. You set everything in the `Vec` to zero.
3. You set a few values in the `Vec`. This can be either constants (see: `sourceTable`) or expressions (see: `sourceStall`).
4. You dynamically index out of the `Vec`.

This is basically describing a decoder which _should be optimized by synthesis tools_. However, it looks really bad and can trip limits on dynamic indexes for "memories" for some tools. Specifically, Cadence Genus will warn if you try to dynamically index something larger that 16Ki.

There are some subtleties here where the dynamic index has both a default value and an out-of-bounds value. I'll put a comment later about this.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Maybe. However, you may want to have to describe a `SparseVec` which is larger than what you have specified in the mapping. E.g., you only set element `1`, but you want to have a `SparseVec` that you can index with a `UInt(2.W)`. Maybe the address could be a `gen` parameter.

seldridge:
Note that this has some interesting interactions when trying to use this to replace a dynamic index into a `Vec`. The Scala-based FIRRTL Compiler (SFC) has always treated an out-of-bounds index into a `Vec` as returning the zero'th element. The CIRCT developers have tried to change this but we have discovered that designers have intentionally or unintentionally written designs which rely on this behavior. 😢  The current version of this PR differs in behavior from a dynamic index into a `Vec`. Commit https://github.com/chipsalliance/chisel/pull/3619/commits/7b5d6fc675b881617a3d8471689629e8b2fd8639 includes a test of this behavior that shows a difference in the results between a dynamic index and a `SparseVec`.

Normally, you would leave this behavior up to a FIRRTL compiler to sort out (as the FIRRTL spec states that an out-of-bounds access can return a different indeterminate value for each out-of-bounds element). However, with `SparseVec`, Chisel needs to eagerly choose what the out-of-bounds behavior is. Eventually, this behavior will likely diverge between `SparseVec` and dynamic out-of-bounds index. I will update this PR to use the current behavior of dynamic index out-of-bounds behavior (possibly under an option).

src/main/scala/chisel3/util/SparseVec.scala
seldridge: The right thing to do here is probably return `BitPath.dontCare(addrWidth)` though this should be configurable.

src/test/scala/chiselTests/util/SparseVecSpec.scala
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: Apache-2.0
```
jackkoenig, line 1: I've been putting tests that involve Verilog simulation in the integration tests, I don't feel strongly that they need to be there but maybe worth moving these over.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,228 @@
...
+  ): A = {
+    val readOnlyModule = Module(new ReadOnlyModule(chiselTypeOf(gen)))
+    readOnlyModule.in :<>= gen
+    readOnlyModule.out
+  }
+
+}
+
+/** A sparse vector.  This should be created using the [[SparseVec$]] factory. */
+class SparseVec[A <: Data] private (size: Int, gen: A, mapping: Seq[(Int, A)]) {
```
jackkoenig, line 35: I'm a little reluctant to give up the class name `SparseVec` to a non-Data (since it would be a breaking change to try to make it a Record after the fact). I'm not saying we need to do the Data API right this moment, but maybe we adopt a slightly different API which is just factory methods that take the `mapping`, `addr`, and `Type` and just go ahead and do the dynamic lookup there?

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,228 @@
...
+
+import chisel3._
+import chisel3.experimental.SourceInfo
+import chisel3.util.{log2Up, BitPat, Mux1H, PriorityMux}
+import chisel3.util.experimental.decode.{decoder, TruthTable}
+import chisel3.util.experimental.InlineInstance
+import scala.collection.SeqMap
+import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, LinkedHashSet}
+
+private[this] object Detail {
```
jackkoenig, line 13: What does `private[this]` do when it's not inside of a class? Is it private to package `util` or private to this file?

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,228 @@
...
+
+  /** A [[SparseVec$]] that is accessed using a [[chisel3.util.experimental.decode.decoder]]. */
+  sealed trait Decoder extends Type {
+    def encoding(index: Int, width: Int): UInt
+
+    def lookup[A <: Data](
+      index:  UInt,
+      values: VecLike[A]
+    )(
+      implicit _sourceInfo: SourceInfo
```
jackkoenig, line 180:
```suggestion
      implicit sourceInfo: SourceInfo
```
I wouldn't use an `_` in the name here, we don't in the rest of the codebase and the name of the argument is part of the public of the function.

src/test/scala/chiselTests/util/SparseVecSpec.scala
seldridge: Good idea. Moved over in https://github.com/chipsalliance/chisel/pull/3619/commits/fbd64e040becffed8cfaf404537b3df51c02fb1c.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Moved this to extend `Record`.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: I thought this was file private, but it's apparently object private. This makes it accessible inside `chisel3.util` which was not the intention. I'll move this around.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Changed this over.

integration-tests/src/test/scala/chiselTest/util/SparseVecSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+        UInt(3.W),
+        Seq(
+          0 -> 1.U,
+          1 -> 2.U,
+          2 -> 3.U,
+          3 -> 4.U
+        )
+      ),
+      // TODO: This warning suppression doesn't work for some reason.
+      annotations = Seq(WarningConfigurationAnnotation("id=4:s"))
```
jackkoenig, line 66: You should be able to fix the warning by making the indices the correct widths

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+                  BitPat(i.U(addrWidth.W)) -> BitPat(d.encoding(baseEncodedAddress, encodedSize))
+              }
+          }
+        }
+
+        // Generate the truth table.  Use a default value of zero.
+        val ttable = TruthTable(
+          legalBitPats,
+          BitPat.N(encodedSize)
+        )
```
jackkoenig, line 124: Should we make the default value a parameter an argument? It would be nice to be able to ask for dont care.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+
+        // Generate the lookup.
+        d.lookup(decoder(addr, ttable), VecInit((defaultValue ++ elements.values).toSeq)).asInstanceOf[A]
+
+      // Generate a lookup using a sequence of when statements.
+      case SparseVec.IfElse =>
+        val result = Wire(gen)
+
+        // The default value _must_ be specified to bypass initialization checking.
+        result := 0.U
```
jackkoenig, line 134: The user may want to opt-in to DontCare here.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+          // out-of-bounds accesses.
+          zeroValue match {
+            case None => Seq.empty
+            case Some(_) =>
+              (size until BigInt(addrWidth).pow(2).toInt).map {
+                case i =>
+                  BitPat(i.U(addrWidth.W)) -> BitPat(d.encoding(baseEncodedAddress, encodedSize))
+              }
+          }
+        }
```
jackkoenig, line 118: Similarly the user may want to opt-in to dont care on these.

We also may want to check QoR of doing this this way vs. using a mux in the same way we do on IfElse below--but we can do that evaluation later.

jackkoenig (commented):
Looks awesome, some nits, a maybe bug, and a request

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+  *
+  * A [[SparseVec]] will take up storage equal to the size of the provided
+  * mapping argument with one additional slot for the default value, if one is
+  * needed.
+  *
+  * @param size the apparent size of the vector
+  * @param gen the element type of the vector
+  * @param mapping the indices of the vector which are valid
+  */
+class SparseVec[A <: Data](size: Int, gen: => A, mapping: Seq[Int]) extends Record {
```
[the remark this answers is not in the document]
seldridge: It _can_ be calculated as the maximum of the mapping. @sequencer had the same suggestion. However, that prevents you from describing a dense vec-like thing that has the exact behavior you want. It could be a default like:

```scala
class SparseVec[A <: Data](mapping: Seq[Int])(size: Int = mapping.size, gen: => A) extends Record
```

This implementation is very much trying to just make this equivalent to the dense vec case, so it's likely not properly generalized. Do we want to generalize it right now?

WDYT?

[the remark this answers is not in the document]
jackkoenig: I think it's reasonable to just have the argument by default, we could have auxiliary constructors or factory methods to let the user derive the size from the mapping, eg.
```scala
  def this(gen: => A, mapping: Seq[Int]) = this(mapping.max, gen mapping)
```
Or similarly on an apply method on `object SparseVec`.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+    * @param sourceinfo implicit source locator information
+    * @return a read-only value from the specified address
+    * @throws ChiselException if the returned value is written to
+    */
+  def apply(addr: UInt, lookupType: SparseVec.Type = SparseVec.Binary)(implicit sourceinfo: SourceInfo): A = {
+    val result: A = lookupType match {
+      // Generate a lookup using a decoder.
+      case d: SparseVec.Decoder =>
+        // The number of bits required to represent all addresses of the vector.
+        val addrWidth = log2Up(size)
```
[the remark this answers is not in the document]
seldridge: This is the number of bits necessary to represent the size of the vec, not the number of bits necessary for the internal representation.

The lookup is roughly: `UInt(addrWidth.W) -> chiselTypeOf(encoding(_, encodedSize)) -> Mux`.

The `encodedSize` (number of unique entries) is passed to the encoding to let it determine the size. This looks to be correct for the one-hot, but looks overprovisioned for the binary. But, yes, the internal encoding may get much larger which is what the `encoding` method is trying to handle.

I _think_ I got that right. Does that logic make sense?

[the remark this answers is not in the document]
seldridge: I fixed the sizing of the binary case here: e4e076076 This also fixed the warnings which I incorrectly thought was due to valid use of out-of-bounds indexing. My bad.

integration-tests/src/test/scala/chiselTest/util/SparseVecSpec.scala
seldridge: Fixed in e4e076076.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+  // Determine if a default value needs to exist.  This should only exist if
+  // there are fewer map items than the size of the Vec.  (As this is a
+  // SparseVec, this is expected to be a Some.)
+  private val defaultValue: Option[A] = mapping match {
+    case full if mapping.size == size => None
+    case _                            => Some(WireInit(gen, DontCare))
+  }
+
+  /** Read a value from a [[SparseVec]] using one of several possible lookup
+    * typs. The returned value is read-only.
```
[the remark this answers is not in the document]
seldridge: Will fix in a latter commit.

src/main/scala/chisel3/util/SparseVec.scala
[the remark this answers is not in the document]
seldridge: Added alternative constructor in latest version.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: I think that's reasonable as an option. This is behavior that is different from the `DenseVec` case. E.g., making this change here will cause the tests to fail because the default value is now the first element (or however a FIRRTL compiler decides to compile this).

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Switched to `Record` in latest version.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: There's an alternative constructor that now supports this.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Adding a bunch of stuff to allow this.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: User can opt-in to indeterminate values if they want to.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Note that this is the default value in encoded space. It's not the default value. I'm adding some comments to make this more obvious. Also, there was a slight mistake here where this should be the zeroth value in encoding space. (It should have been passed into the encoding.)

src/main/scala/chisel3/util/SparseVec.scala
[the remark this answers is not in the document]
seldridge: Empty mappings work and there is a test for them now!

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,230 @@
...
+      if (!duplicates.add(index))
+        nonUniqueMapping = true
+      index.toString -> DataMirror.internal.chiselTypeClone(gen)
+  })
+
+  // Throw a runtime exception if there is a non-unique mapping.
+  // TODO: Improve this error message.
+  if (nonUniqueMapping) {
+    throw new ChiselException("non-unique mapping in SparseVec")
+  }
```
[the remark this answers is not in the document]
seldridge: Thanks for writing this out! I'm switching to this in the latest version.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,383 @@
...
+        index:  UInt,
+        values: VecLike[A]
+      )(
+        implicit sourceinfo: SourceInfo
+      ) = values(index)
+    }
+
+    /** A [[SparseVec$]] accessor that uses a one-hot-encoded lookup. */
+    case object OneHot extends Decoder {
+      override final def encoding(index: Int, width: Int) = (1 << index).U((width).W)
```
jackkoenig, line 325:
```suggestion
      override final def encoding(index: Int, width: Int) = (BigInt(1) << index).U((width).W)
```

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,383 @@
...
+  * in the `indices`
+  * @param outOfBoundsValue the out-of-bounds behavior when accessing an index
+  * larger than the largest value in `indices`
+  */
+class SparseVec[A <: Data](
+  size:             Int,
+  gen:              => A,
+  indices:          Seq[Int],
+  defaultValue:     DefaultValueBehavior.Type = DefaultValueBehavior.DynamicIndexEquivalent,
+  outOfBoundsValue: OutOfBoundsBehavior.Type = OutOfBoundsBehavior.First)
```
jackkoenig, line 75: Do we want the default behavior to be the backwards-compatible [with sparse vec implemented using a Vec with DontCare holes] or do we want this new thing to take advantage of undefined behavior by default while still letting the user opt-in to the backwards compatible behavior?

I'm asking because I think we really could go either way here.

src/main/scala/chisel3/util/SparseVec.scala
```diff
@@ -0,0 +1,383 @@
...
+        // If the elements have a value at index zero and if the user indicated
+        // they want to return this value on out-of-bounds behavior, then return
+        // it.  Otherwise, add no logic for this case.
+        (outOfBoundsValue, zeroValue) match {
+          case (SparseVec.OutOfBoundsBehavior.Indeterminate, _) | (_, None) =>
+          case (SparseVec.OutOfBoundsBehavior.First, Some(data)) =>
+            when(addr >= size.U) {
+              result := data
+            }
+        }
```
jackkoenig, line 256: I don't think this logic is correct when you have `SparseVec.OutOfBoundsBehavior.First` combined with no element at index `0`. In order to be equivalent to a dense Vec, we need to give a "fake" zero-th value which effectively is the default `value`. Basically, I think this can be fixed as follows:

```suggestion
        outOfBoundsValue match {
          case SparseVec.OutOfBoundsBehavior.Indeterminate =>
          case SparseVec.OutOfBoundsBehavior.First, Some(data)) =>
            val oob = zeroValue.getOrElse(defaultValue.getValue(lookupType)) // assign result of this expression to a val up above
            when(addr >= size.U) {
              result := oob
            }
        }
```

This could use a test. This case I'm describing here came up in the real design when I deployed this same basic idea (but in a much less reusable way, this PR is awesome).

src/main/scala/chisel3/util/SparseVec.scala
seldridge: Probably new behavior.

seldridge: I'll change it to that.

src/main/scala/chisel3/util/SparseVec.scala
seldridge: I added a test case. However, this seems to pass. I think this logic is correct for the reason that the `result` is already set, outside of any when, to `defaultValue.getValue(lookupType)`. Hence, the PR should only need to set the value to the `zeroValue` and just let it fallback to the default.

Does that make sense?

integration-tests/src/test/scala/chiselTest/util/SparseVecSpec.scala
```diff
@@ -0,0 +1,356 @@
...
+        UInt(3.W),
+        Seq(
+          1 -> 2.U,
+          2 -> 3.U
+        ),
+        true
+      )
+    )
+  }
+
```
jackkoenig, line 203: Doesn't the size need to be at least 4 for this to have out-of-bounds access?

seldridge: The _access_ needs to be padded up to a power of 2, yes. However, the size of the vec should be smaller. The test checks all values up to the next power of 2.

This test is doing is constructing a `SparseVec` and a dense `Vec` of size 3. The dense `Vec` has all values initialized to `DontCare` (elements `[0, 1, 2]`). The user-provided indices in both the `Vec` and `SparseVec` are then set (`[1, 2]`). The test then checks that all indices, including out-of-bounds indices match. I.e., all of `[0, 1, 2, 3]` match.

There's a `debug` flag to the test that shows the index/output behavior for this test. Check if this looks right to you, specifically, `0 -> 0` and `3 -> 0`:

```
index, dense, binary, onehot, ifelse
0: 0, 0, 0, 0
1: 2, 2, 2, 2
2: 3, 3, 3, 3
3: 0, 0, 0, 0
```

seldridge: I added some additional comments to the tests to indicate what is going on here.

jackkoenig (approved):
Awesome work dude, especially on the tests. This PR should serve as an example on how to test things.

Issue #2643, opened 2022-07-25
Unifying Chisel._ and chisel3._ connection semantics


mwachs5:
**Type of issue**: Feature Request

**Is your feature request related to a problem? Please describe.**

## Background: `Chisel._` vs `chisel3._` directionality

In `Chisel._`, `IO` has an implicit output direction, and `Flipped` is used for two things:

 1) Define relative direction of a field with respect to its parent bundle
 2) To change the implicit output to an implicit input of IO

```scala
val io = IO(new Bundle { // implicit output
  val x = Flipped(UInt()) // x is flipped relative to bundle, so it is an input
  val y = UInt() // y is aligned relative to bundle, so it is an output
})
```

You can declare an input port using the (2) kind of Flipped:

```scala
val io = IO(Flipped(new Bundle { // implicit input
  val x = Flipped(UInt()) // x is flipped relative to bundle, so it is an output
  val y = UInt() // y is aligned relative to bundle, so it is an input
}))
```

In `chisel3`, `IO` does not have an implicit direction, but `Input` and `Output` are specified on types, which coerce all subfields of the type to the same direction. `Flipped` inverts the absolute direction so `Input -> Output` and `Output -> Input`.

```scala
val io = IO(new Bundle { // no implicit direction
  val x = Input(UInt()) // x is an input
  val y = Output(UInt()) // y is an output
}))
```

You can switch directions with `Flipped`:

```scala
val io = IO(Flipped(new Bundle { // Swap all absolute directions
  val x = Input(UInt()) // x is flipped so it is an output
  val y = Output(UInt()) // y is flipped so it is an input
})))
```

**Describe the solution you'd like**

## Unifying directionality with new primitives

Both mechanisms of describing field directions can be unified with the following (conceptual) primitives:

- `Flipped`: a field's relative direction is reversed with respect to its parent bundle
- `Aligned`: a field's relative direction is the same with respect to its parent bundle (is implicit)
- `Outgoing`: an IO whose implicit direction is output
- `Incoming`: an IO whose implicit direction is input
- `stripFlipsOf`: a type-generator that aligns all subfields, recursively
- `reverseFlipsOf`: a type-generator that flips all subfields, recursively

Expressing Chisel._ semantics:

- `IO(new Bundle))` becomes `Outgoing(new Bundle)`
[...]

PR #3622, opened 2023-11-06
Add Incoming/Outgoing as alternatives to IO

azidar:
### Motivation

First step in implementing what was discussed in this issue:

https://github.com/chipsalliance/chisel/issues/2643

In short, `Incoming` and `Outgoing` provide a way to create input/output ports that does not "strip flips" like `Input/Output` do. The alternative way to do this was to create `IO(Flipped(..))` and `IO(..)`, which requires the implicit knowledge that `IO` creates an output port.

I'm in general curious for feedback on this API, so I'm marking as a draft for now. Thanks!

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -80,7 +80,8 @@ object Incoming {
     * The granted iodef must be a non-flipped chisel type and not be bound to hardware.
     */
   def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
-    if (SpecifiedDirection.isFlipped(iodef.specifiedDirection)) Builder.error("Incoming(..) cannot accept a chisel typed which is flipped")
+    if (SpecifiedDirection.isFlipped(iodef.specifiedDirection))
+      Builder.error("Incoming(..) cannot accept a chisel typed which is flipped")
     IO(Flipped(iodef))
   }
 }
@@ -106,7 +107,8 @@ object Outgoing {
     * The granted iodef must be a non-flipped chisel type and not be bound to hardware.
     */
   def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
-    if (SpecifiedDirection.isFlipped(iodef.specifiedDirection)) Builder.error("Outgoing(..) cannot accept a chisel typed which is flipped")
+    if (SpecifiedDirection.isFlipped(iodef.specifiedDirection))
+      Builder.error("Outgoing(..) cannot accept a chisel typed which is flipped")
     IO(iodef)
   }
 }
--- a/core/src/main/scala/chisel3/IO.scala
+++ b/core/src/main/scala/chisel3/IO.scala
@@ -9,11 +9,7 @@ object IO {
 
   /** Constructs a port for the current Module
     *
-    * This must wrap the datatype used to set the io field of any Module.
-    * i.e. All concrete modules must have defined io in this form:
-    * [lazy] val io[: io type] = IO(...[: io type])
-    *
-    * Items in [] are optional.
+    * This must wrap the datatype used to create an io port of any Module.
     *
     * The granted iodef must be a chisel type and not be bound to hardware.
     *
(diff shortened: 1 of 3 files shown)


seldridge (commented):
High level: this makes sense to me and is both better than having an implicit direction and works towards removing the confusion around things often used for port creation (`Input`/`Output`) also being useful for interacting with flip.

Bikeshed-level: I'd opt for the terser: `In` and `Out` vs. `Incoming` and `Outgoing`. This has the coincidence that `in`/`out` were the compromise names between favoring terseness and explicitness that were adopted for dialects in CIRCT.

A migration strategy on how to handle existing methods would be good. `IO` could be made deprecated in preparation for being made `private[chisel3]`?

core/src/main/scala/chisel3/IO.scala
```diff
@@ -62,3 +58,57 @@ object IO {
...
+    *  * `i.ready` is an output port
+    *
+    * Note that `Incoming(foo)` is equivalent to `IO(Flipped(foo))`
+    *
+    * This must wrap the datatype used to create an io port of any Module.
+    *
+    * The granted iodef must be a non-flipped chisel type and not be bound to hardware.
+    */
+  def apply[T <: Data](iodef: => T)(implicit sourceInfo: SourceInfo): T = {
+    if (SpecifiedDirection.isFlipped(iodef.specifiedDirection))
```
sequencer, line 83: I think we should require iodef is not `coerced` yet.

sequencer (commented):
Basically I'm thinking try to use chisel3.connectable, so maybe this API should leverage `Alignment` API for connection?

mwachs5 (commented):
in or before we land this, could we add an `Aligned` that basically nothing so that people can be writing:

val foo = Flipped(MyThing())
val bar = Aligned(MyThing())

seldridge:
What would an explicit `Aligned` do? Isn't that a no-op?

sequencer:
> What would an explicit `Aligned` do? Isn't that a no-op?
How about provide a semantic:
```
Aligned(Flipped(MyThing()))
Aligned(Aligned(MyThing()))
```
Both equal to 
```
Aligned(MyThing())
```
`Aligned` will ignore the inner `Flipped`.

mwachs5:
No @sequencer that is not what I mean. I am agreeing for a `noop` operator that does nothing, simply for the alignment (no pun intended) of the words.

darthscsi:
Is it the case that the following are indistinguishable from a connection/flow perspective?  (modulo directly inspecting the type)
val a = Incoming(Bundle(flip a : UInt))
val b = Outgoing(Bundle(a : UInt))

schoeberl:
As @seldridge mentioned, Incoming and Outgoing are unusual names. VHDL uses simply in and out. Verilog uses input and output. Both should be fine.

mwachs5:
> Is it the case that the following are indistinguishable from a connection/flow perspective? (modulo directly inspecting the type)
val a = Incoming(Bundle(flip a : UInt))
val b = Outgoing(Bundle(a : UInt))

Written in actual user Chisel not firrtly-psuedocode:

```
val a = Incoming(new Bundle{ val a = Flipped(UInt())})
val b = Outgoing(new Bundle{ val a = UInt()})

val a_wire = Wire(new Bundle{ val a = Flipped(UInt())})
val b_wire = Wire(new Bundle{ val a = UInt()})

a :<>= a_wire
b :<>= b_wire

```

Not the same: https://scastie.scala-lang.org/exwHPt8vSEWpyd7N2SsKmQ

```
...

connect a_wire.a, a.a @[src/main/scala/main.scala 13:3]
connect b.a, b_wire.a @[src/main/scala/main.scala 14:3]
```

We can tell the users to write something different, but the above is the common thing people are doing intuitively.

What if the wires have no flips to match:

```
val a = Incoming(new Bundle{ val a = Flipped(UInt())})
val b = Outgoing(new Bundle{ val a = UInt()})

val a_wire = Wire(new Bundle{ val a = UInt()})
val b_wire = Wire(new Bundle{ val a = UInt()})

a :<>= a_wire
b :<>= b_wire

```

They get an error: (https://scastie.scala-lang.org/xnj0D741RsqWhoAvSk4hZA)
```
 src/main/scala/main.scala 13:3: inversely oriented fields Example.a.a: IO[UInt] and Example.a_wire.a: Wire[UInt]
[error] a :<>= a_wire
```

mwachs5:
> This kind of feature is technically good. But for this kind of change, what are we going to ask people to do with this new feature? When the designer "needs to add a port", are we advocating the use of Outgoing and Incoming over Input and Output?


`IO(...)` and `IO(Flipped(...))` would be replaced with `Outgoing(...)` and `Incoming(...)`

The enduser doesn't write `Input` or `Output` today to make a port. they have to write `IO(...)` and through some complicated relations to whatever is in the `...` they get an `input` or `output`.

PR #3623, opened 2023-11-08
Support instanceName on HasId without a component by using localName.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

We have recently reworked when HasIds are named within a module so they can be named before atModuleBodyEnd executes, and then get referenced within atModuleBodyEnd. However, the component is still not set when they are referenced, so in this case use the local name.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -317,9 +317,10 @@ private[chisel3] trait HasId extends chisel3.InstanceId {
         case (None, _: MemBase[_]) => _ref.get.localName
         case (None, _) if _ref.isDefined => {
           // Support instance names for HasIds that don't have a _parent set yet, but do have a _ref set.
-          // This allows most HasIds to be named in atModuleBodyEnd, for example.
-          // However, this is only valid if the _ref's fullName doesn't depend on the component name.
-          // Notably, this is not valid for ports of instances, and other constructs that need the _parent's name.
+          // This allows HasIds to be named in atModuleBodyEnd, for example.
+          // In this case, we directly use the localName. This is valid, because the only time names are
+          // context-dependent is on ports. If a port doesn't have a _parent set yet, the port must be within
+          // the currently elaborating module, and should be named by its localName.
           _ref.get.localName
         }
         case (None, _) =>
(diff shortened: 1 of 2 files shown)


azidar (approved):

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -315,6 +315,13 @@ private[chisel3] trait HasId extends chisel3.InstanceId {
         case (Some(c), _) => refName(c)
         case (None, d: Data) if d.topBindingOpt == Some(CrossModuleBinding) => _ref.get.localName
         case (None, _: MemBase[_]) => _ref.get.localName
+        case (None, _) if _ref.isDefined => {
+          // Support instance names for HasIds that don't have a _parent set yet, but do have a _ref set.
+          // This allows most HasIds to be named in atModuleBodyEnd, for example.
+          // However, this is only valid if the _ref's fullName doesn't depend on the component name.
+          // Notably, this is not valid for ports of instances, and other constructs that need the _parent's name.
```
jackkoenig, line 322: This comment is a little bit concerning, what happens if you use `.toRelativeTarget` on a port of a submodule?

mikeurbach: I think I may have made this comment sound too scary, without thinking deeply enough about it. I just pushed up a commit with a submodule, is this what you had in mind? [68a31cb](https://github.com/chipsalliance/chisel/pull/3623/commits/68a31cb4a8efd6d6027344ca9da7762d03b03d40)

I think we actually are fine here. If we're trying to name a port of a submodule, the _parent component of the port will always be set, so we wouldn't come through this code path anyway. The only time we should come here is if we're naming a port of the currently-being-elaborated module, in which case the localName is safe to use (I also added a test demonstrating that).

Does this make sense? Maybe I just need to amend the comment to indicate that in fact any time we hit this code path, it's safe to use a localName.

mikeurbach: FYI I've pushed another commit that updates the comment based on what I just said here.

jackkoenig: Yep makes sense, thank you for the test and clarification!

jackkoenig (approved):

mikeurbach:
Thanks for the reviews!

PR #3624, opened 2023-11-09
Add source locators to all subclasses of BaseModule

jackkoenig:
This is implemented using the compiler plugin so that:
1. The source locator points to the first character of the class name.
2. Users don't accept implicit sourceInfos as arguments to their module classes (this would override all source locators in the body of the class).

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

The source locators point to the name of the class extends Module, RawModule, BlackBox, ExtModule, or Class. They are included in the output FIRRTL.

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -371,6 +371,10 @@ lazy val unipublish =
     .aggregate(plugin) // Also publish the plugin when publishing this project
     .settings(name := (chisel / name).value)
     .enablePlugins(ScalaUnidocPlugin)
+    .settings(
+      // Plugin isn't part of Chisel's public API, exclude from ScalaDoc
+      ScalaUnidoc / unidoc / unidocProjectFilter := inAnyProject -- inProjects(plugin)
+    )
     .settings(commonSettings: _*)
     .settings(publishSettings: _*)
     .settings(usePluginSettings: _*)
--- a/src/test/scala/chiselTests/SourceLocatorSpec.scala
+++ b/src/test/scala/chiselTests/SourceLocatorSpec.scala
@@ -4,10 +4,42 @@ package chiselTests
 
 import circt.stage.ChiselStage.emitCHIRRTL
 import chisel3._
-import chisel3.experimental.SourceLine
+import chisel3.experimental.{BaseModule, ExtModule, SourceLine}
+import chisel3.experimental.hierarchy.Definition
 import firrtl.ir.FileInfo
 
+object SourceLocatorSpec {
+  val thisFile = "src/test/scala/chiselTests/SourceLocatorSpec.scala"
+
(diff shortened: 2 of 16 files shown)


src/test/scala/chiselTests/SourceLocatorSpec.scala
```diff
@@ -41,4 +70,36 @@ class SourceLocatorSpec extends ChiselFunSpec with Utils {
...
+    }
+    it("(2.f): user-defined Classes should have a source locator") {
+      val chirrtl = emitCHIRRTL(new ClassTop)
+      chirrtl should include(s"class ClassChild : @[$thisFile 24:9]")
+    }
+    it("(2.g): Inner and anonymous modules should have a source locators") {
+      val chirrtl = emitCHIRRTL(new Outer)
+      chirrtl should include(s"module Inner : @[$thisFile 29:11]")
+      chirrtl should include(s"module AnonymousModule : @[$thisFile 31:25]")
+    }
```
mwachs5, line 103: awesome! 

are there any interactions with D/I or Instantiate that need to be tested?

jackkoenig: Definitions should get a source locator in the same way as these examples, I'll add a test!

build.sbt
```diff
@@ -371,6 +371,10 @@ lazy val unipublish =
     .aggregate(plugin) // Also publish the plugin when publishing this project
     .settings(name := (chisel / name).value)
     .enablePlugins(ScalaUnidocPlugin)
+    .settings(
+      // Plugin isn't part of Chisel's public API, exclude from ScalaDoc
+      ScalaUnidoc / unidoc / unidocProjectFilter := inAnyProject -- inProjects(plugin)
+    )
```
jackkoenig, line 377: I had to add this because the soft link of `SourceInfoFileResolver` was erroring during ScalaDoc generation saying the same class was included twice. I didn't realize Unidoc was even _trying_ to include the plugin since it's not part of this `unipublish` project (it's published separately). It never should've been included.

src/test/scala/chiselTests/SourceLocatorSpec.scala
jackkoenig: Added new test.

mikeurbach (commented):
Nice! Macro stuff is always fun. I'm no expert, but this makes sense to me.

mwachs5 (approved):

fabianschuiki:
This is awesome! Thanks a lot @jackkoenig for adding this :smiley:!

PR #3628, opened 2023-11-10
Support Instance for Definitions of Classes.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

I split this PR up into two changes: internal changes to support Instance for Definitions of Classes, and then a new StaticObject helper and safe APIs.

#### Desired Merge Strategy

#### Release Notes

This allows safe references to Objects, as well as type-safe field accessors to connect inputs or outputs via `@instantiable` and `@pulic`.

- [ ] Did you review?


core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -165,11 +173,53 @@ object Class {
...
+    }
+
+    obj
+  }
+
+  def getReferenceType(cls: Definition[Class]): Property[ClassType] = {
+    // Get the BaseModule for the Class this is a definition of.
+    val baseModule = cls.getInnerDataContext match {
+      case Some(b) => b
+      case _       => throwException("Internal Error! Class instance did not have an associated BaseModule.")
```
jackkoenig, line 193:
```suggestion
    val baseModule = cls.getInnerDataContext.getOrElse(
      throwException("Internal Error! Class instance did not have an associated BaseModule.")
    )
```
Nitty suggestion but you could use `.getOrElse` here.

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -165,11 +173,53 @@ object Class {
...
+    }
+
+    // Get a Property[ClassType] type from the Class name.
+    val classType = ClassType.unsafeGetClassTypeByName(baseModule.name)
+    Property[classType.Type]()
+  }
+
+  def getObjectInstance(cls: Instance[Class]): StaticObject = {
```
jackkoenig, line 201: This needs ScalaDoc if it's going to be a new public API

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -165,11 +173,53 @@ object Class {
...
+    // Get the BaseModule for the Class this is an instance of.
+    val baseModule = cls.getInnerDataContext match {
```
jackkoenig, line 203: Same comment about `.getOrElse`

src/test/scala/chiselTests/properties/ClassSpec.scala
```diff
@@ -132,4 +132,52 @@ class ClassSpec extends ChiselFlatSpec with MatchesAndOmits {
...
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(
+      new RawModule {
+        val cls = Definition(new Test)
+
+        val obj1 = Instance(cls)
+        val obj2 = Instance(cls)
+
+        obj2.in := obj1.out
```
jackkoenig, line 173: Should you check for this in the `matchesAndOmits`?

src/test/scala/chiselTests/properties/ObjectSpec.scala
```diff
@@ -94,4 +94,95 @@ class ObjectSpec extends ChiselFlatSpec with MatchesAndOmits {
...
+
+    matchesAndOmits(chirrtl)(
+      "module Parent",
+      "output out1 : Inst<Test>",
+      "output out2 : Inst<Test_1>",
+      "object obj1 of Test",
+      "object obj2 of Test_1",
+      "propassign out1, obj1",
+      "propassign out2, obj2"
+    )()
```
jackkoenig, line 186: What are `inst1` and `inst2`? We don't really have a similar concept for normal D/I. Should we instead just make connections support `Instance[A <: Class]`? Or perhaps we should instead make the method to get the object instance be a method on `Instance[A <: Class]` rather than on `object Class`?

jackkoenig (approved):
Generally LGTM but I have some nits and suggestions.

src/test/scala/chiselTests/properties/ObjectSpec.scala
mikeurbach: Hmm yes this is a good point. This certainly feels a little weird, and it would be great if we could directly connect the `Instance[A <: Class]`.

I implemented this a while ago and just sent the PR... trying to page in what led me to this. Looking at what I am doing in getObjectInstance, I guess we need a way to go from a `Instance[A <: Class]` to a `Property[ClassType]`, and add the appropriate binding for the property. So basically, I think whatever getObjectInstance+getReference are doing, we would do internally within `connect`. I can try some experiments.

mikeurbach: I also like your other idea as a backup, we could define a `getReference` method on `Instance[A <: Class]` to do this bookkeeping. I'll try to see if I can make it work to directly connect `Instance[A <: Class]` first.

jackkoenig: Note that for either case you probably need to be using extension methods (aka `implicit class`), eg. see `Connectable`.

mikeurbach: Makes sense, I'll poke around with it.

mikeurbach: Yep, the first idea works. I can add an extension method to `Property[ClassType]` that adds a `:=` method that accepts an Instance of some class. I need to spend a little more time playing with this to see if we can completely get rid of this StaticObject thing or if it's needed internally for bookkeeping. I'm also hoping we could statically enforce that the Instance is of the right class... at the least we can do a runtime check.

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -165,11 +173,53 @@ object Class {
...
+        cls.addCommand(DefObject(sourceInfo, obj, obj.className.name))
+        classProp.bind(ClassBinding(cls), SpecifiedDirection.Unspecified)
+      }
+      case _ => throwException("Internal Error! Property connection can only occur within RawModule or Class.")
+    }
+
+    obj
+  }
+
+  def getReferenceType(cls: Definition[Class]): Property[ClassType] = {
```
sequencer, line 189: Please add documentation, I saw it is used in the test code.

src/test/scala/chiselTests/properties/ObjectSpec.scala
mikeurbach: I don't see a way to get rid of StaticObject, since we need something to hold a reference to the instance's underlying module clone until after its been named, so we can set the ref for the Property[ClassType] we make from the instance to the module clone's ref. I also don't know the best way to enforce the types match statically, since we need the module clone's `name`, which I don't know how to bake that into a Scala type.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -348,6 +349,47 @@ object Property {
...
+      )
+
+      // Get the sink Property[ClassType] expected class name.
+      val propClassName = prop.getPropertyType match {
+        case fir.ClassPropertyType(name) => name
+        case _                           => throwException("Internal Error! Property[ClassType] underlying type was not ClassPropertyType")
+      }
+
+      // Assert the sink expects the same class as the source represents.
+      if (propClassName != baseModule.name) {
```
mikeurbach, line 366: I don't know if there is a way to move this check to Scala's type system, but that would be great. Any thoughts @albertchen-sifive or @jackkoenig ?

jackkoenig: I mentioned this on the on the other comment thread but this may be the best place to discuss it. It is possible to life the String literal value to the Scala type system, but it might be very cumbersome in typical use, that being said, we can try it out. Something like:
```scala
class Class[T <: String](value: T)

val c: Class["jack"] = new Class("jack")
val d: Class["jack"] = new Class("not jack")  // this is a compile error
```

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -348,6 +349,47 @@ object Property {
     // This cast should be safe, because there are no members of prop.ClassType to access
     def as(prop: Property[ClassType]): Property[ClassType] with prop.ClassType =
       prop.asInstanceOf[Property[ClassType] with prop.ClassType]
+
+    def :=[T <: Class](source: Instance[T])(implicit sourceInfo: SourceInfo) = {
```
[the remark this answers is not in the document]
mikeurbach: Ah yep, I forgot to write documentation here. I will add some. This is just some syntactic sugar and safety around something you could already do with the DynamicObject stuff, but yes this stuff is something we've never had before: the ability to pass a reference to an instance of some declaration around in ports.

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: We need the string in the `Property[ClassType]`, so I will see how it looks to try and lift that information there.

jackkoenig: Singleton types (which is what this is) are supported since Scala 2.13, but I will presume they are a lot more cumbersome to deal with in Scala 2 since some things users may want to do with them would required metaprogramming which is a lot more elegant in Scala 3.

mikeurbach: I poked at this a bit, and it seems possible, but I'm also happy to just start with a runtime check since that is straightforward.

core/src/main/scala/chisel3/properties/Property.scala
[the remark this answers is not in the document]
mikeurbach: I added scaladoc for this method. I am planning to write out an actual user guide for how Class and Object are supposed to work in https://www.chisel-lang.org/chisel3/docs/explanations/properties.html, but I've been holding off until some of the APIs have stabilized. What do you think about updating the explanation page in a future commit?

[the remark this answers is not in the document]
sequencer: Yeah! fell free to add the full user documentation in the following PR, I think I understand the API design now, but have some nitpick on it.

sequencer (commented):
Add some nitpick on the API design.

src/test/scala/chiselTests/properties/ObjectSpec.scala
sequencer: That's interesting! I originally thought it only can be done in Scala 3.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -348,6 +349,55 @@ object Property {
...
+
+    /** Connect a reference to an Instance[Class] to a Property[ClassType] prop.
+      *
+      * This method allows Instances of Classes to be safely connected to Property[ClassType] ports, so the references
+      * can be passed through the hierarchy.
+      *
+      * @param source An Instance[T <: Class], where T is the same class as prop's ClassType.
+      * @param sourceInfo The source location where the connection is made, for error messages.
+      */
+    def :=[T <: Class](source: Instance[T])(implicit sourceInfo: SourceInfo) = {
```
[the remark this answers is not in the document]
mikeurbach: When you put it like that, it does feel awkward. I guess what we want here is an extension method on `Instance` like you're suggesting, which converts `Instance[Class]` to the `Property[ClassType]` for that class, and then the normal `:=` for Properties works.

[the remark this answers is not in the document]
sequencer: > When you put it like that, it does feel awkward. I guess what we want here is an extension method on Instance like you're suggesting, which converts Instance[Class] to the Property[ClassType] for that class, and then the normal := for Properties works.

Yes, that what I mean, but since we don't do the release for `Property` recently, I think we can safely get this in, and do the API change in the following PRs..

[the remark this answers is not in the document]
mikeurbach: Yeah, but since this is a new API I'd also prefer to add it in the shape we really want moving forward. I'll take a look and if it's not too hard update it in this PR.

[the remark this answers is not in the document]
sequencer: Sure! let's move fast! :)

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -136,6 +144,23 @@ object AnyClassType {
 
 object Class {
 
+  /** Helper to create a Property[ClassType] type for a Definition[Class].
+    *
+    * This is useful when a Property[ClassType] type is needed for references to instances of the Class.
+    *
+    * This method is safe, and should be used over unsafeGetReferenceType when possible.
+    */
+  def getReferenceType(cls: Definition[Class]): Property[ClassType] = {
```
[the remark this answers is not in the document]
sequencer: other choice: `getClassTypeProperty`, `getClassType`
(personally prefer the last one), change it as your wish.

mikeurbach:
Ok, I reworked this slightly using extension methods:
* `Definition[T <: Class]` now has a `getPropertyType` method to get its `Property[ClassType]`
* `Instance[T <: Class]` now has a `getPropertyTypeReference` method to get a reference to the instance as a `Property[ClassType]`

I think this is better symmetry. I thought about the names of the methods, and this is where I ended, but I'm still open for discussion about that.

I also wanted to maintain the Chisel-time checks that connections between `Property[ClassType]` that I had in the previous version, so I also added a check in MonoConnect, keeping with my previous comment about sticking with a runtime check for now. We could do deeper surgery on `Property[ClassType]` to enable this in the Scala type system, but that's a bigger change.

While I was adding the runtime check in MonoConnect, I realized that properties currently don't do any checks! There was no call to `checkConnect.apply` to ensure the properties were in the right modules and directions. I tried enabling that, and it seems there are some unspecified directions leaking in. I'll open a ticket to address that separately, and at least keep the `Property[ClassType]` runtime check in this PR.

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -411,6 +412,7 @@ private[chisel3] object MonoConnect {
     source:     Property[_],
     context:    BaseModule
   ): Unit = {
+    checkConnect.checkPropertyConnection(sourceInfo, sink, source)
```
mikeurbach, line 415: The fix for `checkConnect.apply` on Property is here: https://github.com/chipsalliance/chisel/pull/3635. Once that is in, I will rebase this and make sure here we just call `checkConnect`, and the Property version of `checkConnect` calls `checkPropertyConnection`.

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -165,16 +173,69 @@ object Class {
...
+  }
+
+  implicit class ClassInstanceOps[T <: Class](instance: Instance[T]) {
+
+    /** Get a reference to an Instance[Class] as a Property[ClassType] property type.
+      *
+      * This method allows Instances of Classes to be safely connected to Property[ClassType] ports, so the references
+      * can be passed through the hierarchy.
+      */
+    def getPropertyTypeReference: Property[ClassType] = {
```
jackkoenig, line 217: `getPropertyTypeReference` feels just a little too similar to `getPropertyType`. What about just `getPropertyReference` or `getPropertyValue`?

jackkoenig (approved):
Still LGTM although maybe some method name bikeshedding.

PR #3659, opened 2023-11-29
[circt] Use new public-modules option

seldridge:
Change a removed firtool option so that tests continue to work.

This fixes the `ci` action failing on https://github.com/chipsalliance/chisel/actions/runs/7031637610, but not the `mill` (binder) failure.


sequencer:
Cherry picked to #3663

PR #3662, opened 2023-11-30
[binder] Update firtool-lib C-API to allow the creation of multiple options

### LINKED ISSUE #3658: CIRCT C-API change breaks Chisel
This issue is reported in https://github.com/chipsalliance/chisel/actions/runs/7017825652

Should be fixed after llvm/circt#6438 merge.

cc @SpriteOvO @seldridge

SpriteOvO:
Fixes #3658, context llvm/circt#6291.

### DIFF
--- a/.github/workflows/ci-circt-nightly.yml
+++ b/.github/workflows/ci-circt-nightly.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Update Staging Branch
-        uses: circt/update-staging-branch@v1.1.0
+        uses: circt/update-staging-branch@v1.2.0
         with:
           user: chiselbot
           email: chiselbot@users.noreply.github.com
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-circt-nightly.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         id: determine-branch
         shell: bash
         run: |
-          branch=$(git rev-parse --abbrev-ref HEAD)
```
SpriteOvO, line 37: This is necessary because other jobs will use this value as git `ref` to checkout, and will fail if it's a branch name from a forked repo. Replacing with a commit ID to avoid the failure.

.github/workflows/ci-circt-nightly.yml
```diff
@@ -34,7 +34,7 @@ jobs:
...
+          branch=$(git rev-parse HEAD)
```
SpriteOvO, line 37: This is necessary because other jobs will use this value as git `ref` to checkout, and will fail if it's a branch name from a forked repo. Replacing with a commit ID to avoid the failure.

sequencer:
Cherry picked to #3663

PR #3675, opened 2023-12-05
dontTouch should default to leaf marking rather than aggregate marking to match current behavior

darthscsi:
Add an argument to `dontTouch` which allows placing don't touch annotations on the leaves of aggregates.  Default this option.  This results in the same behavior as what the current `dontTouch` effectively does today.  Don't touch of an aggregate is on it's plain meaning different from don't touch on the leaves of an aggregate.  Making the annotation a bit more precise allows greater freedom in code generation and allows catching contradictory requests by the programmer.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation

#### Desired Merge Strategy

#### Release Notes

Don't touch is becoming stricter.  Specifically type lowering and don't touch of aggregates are in conflict. `dontTouch` will maintain the same behavior by default by applying `dontTouch` to every leaf when the argument is an Aggregate.  The new argument `markAgg` can be set to `true` to have `dontTouch` mark the Aggregate instead.

- [ ] Did you review?


sequencer:
How about rename to `dontTouchElements`?

> Specifically type lowering and don't touch of aggregates are in conflict.

I think we need to discuss what's the purpose of `dontTouch` on a Bundle/Vec type, for legacy reason, Chisel has dontTouch for these reasons:
- debug signal(should be replaced by probe)
- PnR anchor(should be replaced by OM with Path)
- Stable ABI of module(should be implemented by firrtl public module)

darthscsi:
> I think we need to discuss what's the purpose of `dontTouch` on a Bundle/Vec type, for legacy reason, Chisel has dontTouch for these reasons:

Yes, part of the problem with don't touch is (in practice to most designers) it's semantics are "magic sprinkles I add when I don't like the output".  The fact that we type lower don't touched things is a clear violation of the plain meaning of the words.  Which is all to say, it is wildly unspecified.  I'm actually trying to make it possible to preserve the current behavior without introducing contradictions.

> * debug signal(should be replaced by probe)

don't touch of an aggregate has nothing to do with this.

> * PnR anchor(should be replaced by OM with Path)

don't touch of an aggregate has nothing to do with this.

> * Stable ABI of module(should be implemented by firrtl public module)

don't touch of an aggregate has nothing to do with this.


Don't touch has these and other reasons to exist.  But none of those reasons mean `dontTouch(agg) == dontTouch(leaves of agg)`.  We transform aggregates all the time.  In fact it is required to compile to verilog.  So then, given that we must touch things, what does it mean when we have to touch something which is marked don't touch?  This change preserves the current behavior while not causing designs to run into that problem.

jackkoenig (approved):

PR #3682, opened 2023-12-12
Change Trace API to only trace leaves

seldridge:
Change the behavior of the Trace API to only trace leaves.  This aligns with recent changes to CIRCT which error if annotations are applied to aggregates which are then mandatorily split due to FIRRTL ABIs.  This should have no practical effect on use of the Trace API only that downstream consumers will need to work at the level of leaf types and not on aggregates.

### DIFF
--- a/core/src/main/scala/chisel3/experimental/Trace.scala
+++ b/core/src/main/scala/chisel3/experimental/Trace.scala
@@ -31,9 +31,6 @@ object Trace {
   def traceName(x: Data): Unit = {
     x match {
       case aggregate: Aggregate =>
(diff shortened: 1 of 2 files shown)


sequencer (approved):
LGTM. But eventually we should migrate TraceAPI to OMPath

sequencer:
CI failed due to 

`chiselTests.AnalogIntegrationSpec`

seldridge:
Thanks for the review!

AnalaogIntegrationSpec should be fixed with tonight's CIRCT nightly. I'll merge this so that nightly should hopefully go green when the new nightly gets picked up.

PR #3683, opened 2023-12-12
Add ModuleChoice and Groups

nandor:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Added support for the post-generation configuration of designs. Instance choices allow multiple, option-dependent targets to be specified for a single option, picking an implementation in the downstream flow.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/choice/package.scala
+++ b/core/src/main/scala/chisel3/choice/package.scala
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.util.simpleClassName
+
+/** This package contains Chisel language definitions for describing configuration options and their accepted values.
+  */
+package object choice {
+
+  /** An option group declaration. Specifies a container grouping values for some design configuration parameter.
+    *
+    * @example
+    * {{{
+    * import chisel3.option.{Group, Case}
+    * object Platform extends Group {
+    *   object FPGA extends Case
+    *   object ASIC extends Case
+    * }
+    * }}}
+    */
+  abstract class Group(implicit _sourceInfo: SourceInfo) {
+    self: Singleton =>
+
+    private[chisel3] def sourceInfo: SourceInfo = _sourceInfo
+
+    private[chisel3] def name: String = simpleClassName(this.getClass())
+
+    final implicit def group: Group = this
+  }
+
+  /** An option case declaration.
+    */
+  abstract class Case(implicit val group: Group, _sourceInfo: SourceInfo) {
+    self: Singleton =>
+
+    private[chisel3] def sourceInfo: SourceInfo = _sourceInfo
+
+    private[chisel3] def name: String = simpleClassName(this.getClass())
+
+    /** A helper method to allow ModuleChoice to use the `->` syntax to specify case-module mappings.
+      *
+      * It captures a lazy reference to the module and produces a generator to avoid instantiating it.
+      *
+      * @param module Module to map to the current case.
+      */
+    def ->[T](module: => T): (Case, () => T) = (this, () => module)
+  }
+}
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -503,6 +503,7 @@ private[chisel3] class DynamicContext(
   val annotations = ArrayBuffer[ChiselAnnotation]()
   val newAnnotations = ArrayBuffer[ChiselMultiAnnotation]()
   val layers = mutable.LinkedHashSet[layer.Layer]()
+  val options = mutable.LinkedHashSet[choice.Case]()
   var currentModule: Option[BaseModule] = None
 
   /** Contains a mapping from a elaborated module to their aspect
@@ -582,7 +583,8 @@ private[chisel3] object Builder extends LazyLogging {
   def components:  ArrayBuffer[Component] = dynamicContext.components
   def annotations: ArrayBuffer[ChiselAnnotation] = dynamicContext.annotations
 
-  def layers: mutable.LinkedHashSet[layer.Layer] = dynamicContext.layers
+  def layers:  mutable.LinkedHashSet[layer.Layer] = dynamicContext.layers
+  def options: mutable.LinkedHashSet[choice.Case] = dynamicContext.options
 
   def contextCache: BuilderContextCache = dynamicContext.contextCache
 
@@ -1010,6 +1012,15 @@ private[chisel3] object Builder extends LazyLogging {
         Layer(l.sourceInfo, l.name, convention, children.map(foldLayers).toSeq)
       }
 
+      val optionDefs = groupByIntoSeq(options)(opt => opt.group).map {
+        case (optGroup, cases) =>
+          DefOption(
+            optGroup.sourceInfo,
+            optGroup.name,
+            cases.map(optCase => DefOptionCase(optCase.sourceInfo, optCase.name))
+          )
+      }
+
       (
         Circuit(
           components.last.name,
@@ -1018,7 +1029,8 @@ private[chisel3] object Builder extends LazyLogging {
           makeViewRenameMap,
           newAnnotations.toSeq,
           typeAliases,
-          layerAdjacencyList(layer.Layer.Root).map(foldLayers).toSeq
+          layerAdjacencyList(layer.Layer.Root).map(foldLayers).toSeq,
+          optionDefs
         ),
         mod
       )
--- a/docs/src/explanations/instchoice.md
+++ b/docs/src/explanations/instchoice.md
@@ -1,0 +1,47 @@
+# Instance Choices
+
+`Instance Choice`s are instances of modules whose targets are configurable post-elaboration.
+They allow the target of an instance to be chosen from a pre-defined set after elaboration by
+enabling an option through the ABI or through specialization in the compiler.
+
+Instance choices rely on option groups to specify the available targets attached to each option:
+
+```scala mdoc:silent
+import chisel3.choice.{Case, Group}
+
+object Platform extends Group {
+  object FPGA extends Case
+  object ASIC extends Case
+}
+```
+
+The `Platform` option groups enumerates the list of platforms for which the design
+can be specialised, such as `ASIC` or `FPGA`. Specialization is not mandatory: if an
+option is left unspecified, a default variant is chosen.
+
+The modules referenced by an instance choice must all specify the same IO interface by
+deriving from `FixedIOBaseModule`. The `ModuleChoice` operator takes the default option
+and a list of case-module mappings and returns a binding to the IO of the modules.
+
+```scala mdoc:silent
+import chisel3._
+import chisel3.choice.ModuleChoice
+
+class TargetIO extends Bundle {
+  val in = Flipped(UInt(8.W))
+  val out = UInt(8.W)
+}
+
+class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO)
+
+class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO)
+
+class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO)
+
+class SomeModule extends RawModule {
+  val inst = ModuleChoice(new VerifTarget)(Seq(
+    Platform.FPGA -> new FPGATarget,
+    Platform.ASIC -> new ASICTarget
+  ))
+}
+```
--- a/firrtl/src/main/scala/firrtl/ir/Serializer.scala
+++ b/firrtl/src/main/scala/firrtl/ir/Serializer.scala
@@ -292,6 +292,14 @@ object Serializer {
       s(init); s(info)
     case DefInstance(info, name, module, _) =>
       b ++= "inst "; b ++= legalize(name); b ++= " of "; b ++= legalize(module); s(info)
+    case DefInstanceChoice(info, name, default, option, choices) =>
+      b ++= "instchoice "; b ++= legalize(name); b ++= " of "; b ++= legalize(default);
+      b ++= ", "; b ++= legalize(option); b ++= " : "; s(info)
+      choices.foreach {
+        case (choice, module) =>
+          newLineAndIndent(1)
+          b ++= legalize(choice); b ++= " => "; b ++= legalize(module)
+      }
     case DefObject(info, name, cls) =>
       b ++= "object "; b ++= legalize(name); b ++= " of "; b ++= legalize(cls); s(info)
     case DefMemory(
@@ -485,6 +493,21 @@ object Serializer {
       s(circuit.info)
       Iterator(b.toString)
     }
+    val options = if (circuit.options.nonEmpty) {
+      implicit val b = new StringBuilder
+      circuit.options.foreach { optGroup =>
+        newLineAndIndent(1)
+        b ++= s"option ${optGroup.name} :"
+        s(optGroup.info)
+        optGroup.cases.foreach { optCase =>
+          newLineAndIndent(2)
+          b ++= optCase.name
+          s(optCase.info)
+        }
+        newLineNoIndent()
+      }
+      Iterator(b.toString)
+    } else Iterator.empty
     val typeAliases = if (circuit.typeAliases.nonEmpty) {
       implicit val b = new StringBuilder
       circuit.typeAliases.foreach(ta => { b ++= s"${NewLine}"; doIndent(1); s(ta) })
@@ -502,6 +525,7 @@ object Serializer {
       Iterator(b.toString)
     } else Iterator.empty
     prelude ++
+      options ++
       typeAliases ++
       layers ++
       circuit.modules.iterator.zipWithIndex.flatMap {
--- a/src/main/scala/chisel3/choice/ModuleChoice.scala
+++ b/src/main/scala/chisel3/choice/ModuleChoice.scala
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.choice
+
+import scala.language.experimental.macros
+import scala.collection.immutable.ListMap
+
+import chisel3.{Data, FixedIOBaseModule, Module, SourceInfoDoc}
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal.{groupByIntoSeq, Builder, WireBinding}
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl.DefInstanceChoice
+import chisel3.internal.sourceinfo.InstChoiceTransform
+
+object ModuleChoice extends SourceInfoDoc {
+
+  /** A wrapper method for Module instantiation based on option choices.
+    * (necessary to help Chisel track internal state).
+    * @see [[chisel3.choice.Group]] on how to declare the options for the alternatives.
+    *
+    * Example:
+    * ```
+    * val module = ModuleChoice(new DefaultModule)(Seq(
+    *   Platform.FPGA -> new FPGATarget,
+    *   Platform.ASIC -> new ASICTarget
+    * ))
+    * ```
+    *
+    * @param default the Module to instantiate if no module is specified
+    * @param choices the mapping from cases to module generators
+    *
+    * @return the input module `m` with Chisel metadata properly set
+    *
+    * @throws java.lang.IllegalArgumentException if the cases do not belong to the same option.
+    */
+  def apply[T <: Data](
+    default: => FixedIOBaseModule[T]
+  )(choices: => Seq[(Case, () => FixedIOBaseModule[T])]
+  ): T =
+    macro InstChoiceTransform.apply[T]
+
+  /** @group SourceInfoTransformMacro */
+  def do_apply[T <: Data](
+    default: => FixedIOBaseModule[T],
+    choices: Seq[(Case, () => FixedIOBaseModule[T])]
+  )(
+    implicit sourceInfo: SourceInfo
+  ): T = {
+    val instDefaultModule = Module.evaluate(default)
+
+    val choiceModules = choices.map {
+      case (choice, module) =>
+        val instModule = Module.evaluate(module())
+        if (!instModule.io.typeEquivalent(instDefaultModule.io)) {
+          Builder.error("Error: choice module IO bundles are not type equivalent")
+        }
+        Builder.options += choice
+        (choice, instModule)
+    }
+
+    groupByIntoSeq(choiceModules.map(_._1))(opt => opt).foreach {
+      case (_, group) =>
+        if (group.size != 1) {
+          throw new IllegalArgumentException(s"Error: duplicate case '${group.head.name}'")
+        }
+    }
+
+    val groupedChoices = choices.map(_._1.group).distinct
+    if (groupedChoices.size == 0) {
+      throw new IllegalArgumentException("Error: at least one alternative must be specified")
+    }
+    if (groupedChoices.size != 1) {
+      val groupNames = groupedChoices.map(_.name).mkString(", ")
+      throw new IllegalArgumentException(s"Error: cannot mix choices from different groups: ${groupNames}")
+    }
+    val group = groupedChoices.head.name
+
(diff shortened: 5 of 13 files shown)


firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -292,6 +292,14 @@ object Serializer {
...
     case DefInstance(info, name, module, _) =>
       b ++= "inst "; b ++= legalize(name); b ++= " of "; b ++= legalize(module); s(info)
+    case DefInstanceChoice(info, name, default, option, choices) =>
+      b ++= "inst "; b ++= legalize(name); b ++= " of "; b ++= legalize(default);
+      b ++= " "; b ++= legalize(option); s(info)
+      choices.foreach {
+        case (choice, module) =>
+          newLineAndIndent(1)
+          b ++= legalize(choice); b ++= " => "; b ++= legalize(module)
+      }
```
seldridge, line 302: Is this the right textual format? `s/inst/inst_choice/`, nittily comma after the default, missing colon after option name.

firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -485,6 +493,21 @@ object Serializer {
...
+    val options = if (circuit.options.nonEmpty) {
+      implicit val b = new StringBuilder
+      circuit.options.foreach { optGroup =>
+        newLineAndIndent(1)
+        b ++= s"option ${optGroup.name} :"
+        s(optGroup.info)
+        optGroup.cases.foreach { optCase =>
+          newLineAndIndent(2)
+          b ++= s"case ${optCase.name} :"
+          s(optCase.info)
```
seldridge, line 505: No semicolon after case names, right?

src/main/scala/chisel3/ModuleChoice.scala
```diff
@@ -0,0 +1,77 @@
...
+    * @param default the Module to instantiate if no module is specified
+    * @param choices the mapping from cases to module generators
+    *
+    * @return the input module `m` with Chisel metadata properly set
+    *
+    * @throws IllegalArgumentException if the cases do not belong to the same option.
+    */
+  def apply[T <: Data](
+    default: => FixedIOBaseModule[T]
+  )(choices: => ListMap[option.Case, () => FixedIOBaseModule[T]]
```
seldridge, line 30: You're using `ListMap` in order to guarantee an ordering here. The alternative is to use a `Seq` and put the checking of uniqueness when this method is called (as opposed to putting it in the type). There are already _other_ run time checks, so this seems fine to delay similarly. I'd suggest changing this. @jackkoenig: do you have any thoughts here?

`ListMap` is just weird as it's very rarely what you want do to its terrible performance characteristics (it is not an ordered map, it is literally just a linked list that will iterate over the list to find the key). Scala 2 has no immutable ordered list, only mutable ordered lists.

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+      out := inst.out
+    }
+
+    intercept[IllegalArgumentException] { ChiselStage.emitCHIRRTL(new MixedOptions) }.getMessage() should include(
+      "module choices must belong to the same group of options"
+    )
+
+  }
+
+  "Module choice checking" should "require that at least one alternative is present" in {
```
seldridge, line 77: Nit: you can substitute `it` for "Module choice checking" here.

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.util.simpleClassName
+
+/**
+  * This object contains Chisel language definitions for describing configuration options and their accepted values.
+  */
```
seldridge, line 10: Nit throughout: 
```suggestion
/** This object contains Chisel language definitions for describing configuration options and their accepted values.
  */
```

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,44 @@
...
+    private[chisel3] def name: String = simpleClassName(this.getClass())
+
+    /**
+      * A helper method to allow ModuleChoice to use the `->` syntax to specify case-module mappings.
+      *
+      * It captures a lazy reference to the module and produces a generator to avoid instantiating it.
+      *
+      * @param module Module to map to the current case.
+      */
+    def ->[T](module: => T): (Case, () => T) = (this, () => module)
```
seldridge, line 42: Re: the `ListMap` discussion at the end and guaranteeing uniqueness of arguments. There may be something clever that could be done to link the type of the `Option` into the case as a type parameter which could then be checked.

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
```
seldridge, line 44: @jackkoenig: Can you review this implementation?

firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -485,6 +493,21 @@ object Serializer {
...
+        }
+        b ++= s"${NewLine}"
```
seldridge, line 507: `newLineNoIndent()` or some other method here?

src/main/scala/chisel3/ModuleChoice.scala
```diff
@@ -0,0 +1,77 @@
...
+import scala.language.experimental.macros
+import scala.collection.immutable.ListMap
+
+import chisel3.FixedIOBaseModule
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal._
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl._
+import chisel3.internal.sourceinfo.InstChoiceTransform
+import chisel3.internal.{Builder}
```
seldridge, line 14: Nit: please avoid wildcard imports. I don't think there are that many things imported here?

Also, the following is sufficient, no curly braces:

```
import chisel3.internal.Builder
```

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+
+class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO) {}
+
+class ModuleChoiceSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
+  it should "emit options and cases" in {
+    class ModuleWithChoice extends Module {
+      val out = IO(UInt(8.W))
+
+      val inst =
+        ModuleChoice(new VerifTarget)(Seq(Platform.FPGA -> new FPGATarget, Platform.ASIC -> new ASICTarget))
```
jackkoenig, line 36: Is this something that should have to be specified at every call site, or is this more of a property of the type of the default choice? I'm envisioning some sort of typeclass API, eg.

```scala
// ...
object ModuleChoice {
  def apply[T <: Data : ModuleChoices](default: => FixedIOBaseModule[T]): T = ???
}
// ...

class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO) {}

class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO) {}

class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO) {}
object VerifTarget {
  implicit val choices = Choices[TargetIO](Platform.FPGA -> new FPGATarget, Platform.ASIC -> new ASICTarget))
}
// ...
val inst = ModuleChoice(new VerifTarget)
```
Then of course if there are different choices for different call sites the user can provide the argument explicitly using normal typeclass pattern mechanisms.

nandor: This feature should be used at call sites. Users should feel free to add wrappers to group some choices, but that's not something that the Chisel feature should bake in.

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,41 @@
...
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.util.simpleClassName
+
+/** This object contains Chisel language definitions for describing configuration options and their accepted values.
+  */
+object option {
+
+  /** An option group declaration.
+    */
+  abstract class Option(implicit _sourceInfo: SourceInfo) {
```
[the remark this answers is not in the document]
nandor: How about `TargetOption`?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1010,6 +1012,18 @@ private[chisel3] object Builder extends LazyLogging {
         Layer(l.sourceInfo, l.name, convention, children.map(foldLayers).toSeq)
       }
 
+      val optionDefs = options.toSeq
+        .groupBy(opt => opt.group)
```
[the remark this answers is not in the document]
nandor: Thanks for clarifying this!

core/src/main/scala/chisel3/Option.scala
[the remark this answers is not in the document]
jackkoenig: That sounds good to me

src/test/scala/chiselTests/ModuleChoiceSpec.scala
jackkoenig: But why? You're already opting in at the call site by using `ModuleChoice` instead of `Module`. The question is how often for a given type do you expect the user to use different choices or is it probably basically always going to be the same? Yes the user can write a function, but making the default behavior more elegant in the common case is still generally a good idea.

nandor: I was a bit confused by the code you sketched up since it hides the variants from the default.

core/src/main/scala/chisel3/Option.scala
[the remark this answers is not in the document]
nandor: I went for `OptionGroup` since we talked about it in the spec already. Not sure if I want to bring in the notion of a "Target" in here specifically.

mwachs5 (commented):
high level comment, can we add a `.md` in `docs/src/explanations` for this new feature?

src/main/scala/chisel3/ModuleChoice.scala
```diff
@@ -0,0 +1,92 @@
...
+
+object ModuleChoice extends SourceInfoDoc {
+
+  /** A wrapper method for Module instantiation based on option choices.
+    * (necessary to help Chisel track internal state). @see [[OptionGroup]]
+    * on how to declare the options for the alternatives.
+    *
+    * Example:
+    * ```
+    * val module = ModuleChoice(new DefaultModule)(Seq(
```
mwachs5, line 23: what is the scala type of `module`? Is this thing figuring out the scala-common type between them?

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+
+      inst.in := 42.U(8.W)
+      out := inst.out
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(new ModuleWithChoice, Array("--full-stacktrace"))
+
+    info("CHIRRTL emission looks coorect")
```
mwachs5, line 45:
```suggestion
    info("CHIRRTL emission looks correct")
```

src/main/scala/chisel3/ModuleChoice.scala
nandor: It's going to be the type parameter of the `FixedIOModule`s used.

nandor:
Done, added `instchoice.md`.

src/test/scala/chiselTests/ModuleChoiceSpec.scala
jackkoenig: > I was a bit confused by the code you sketched up since it hides the variants from the default

What I wrote is just how you express typeclasses in Scala 2, by virtue of putting the variants in the companion object for the default, it's very much associated with the default.

I agree it's fine to go with the more basic API for the time being, although my intuition is still that the API people end up wanting is something closer to the typeclass-y API and it will be slightly unfortunate that we have to give the higher-level abstraction a different name. We can cross that bridge later though.

core/src/main/scala/chisel3/Option.scala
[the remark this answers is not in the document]
jackkoenig: That sounds good, thanks!

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,50 @@
...
+    * ```
+    */
+  abstract class OptionGroup(implicit _sourceInfo: SourceInfo) {
+    self: Singleton =>
+
+    private[chisel3] def sourceInfo: SourceInfo = _sourceInfo
+
+    private[chisel3] def name: String = simpleClassName(this.getClass())
+
+    implicit def group: OptionGroup = this
```
jackkoenig, line 30: Should this be final or is it something users are allowed to override?

core/src/main/scala/chisel3/Option.scala
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.util.simpleClassName
+
+/** This object contains Chisel language definitions for describing configuration options and their accepted values.
+  */
+object option {
```
jackkoenig, line 10: I think this PR looks good, my only hangup is on the name of this object. It just feels very generic for this currently specific thing--and it isn't really related to what `Option` means in Scala more generally.

We will eventually want to add an Option type to Chisel (maybe called `Opt` to keep the pattern with `Vec` being Chisel's `Vector`), but that would probably live in just package `chisel3`.

Does anyone have any ideas? @nandor @azidar @seldridge @mwachs5?

nandor: `option_group`?

seldridge: Maybe call it what it is: `chisel3.parameterization`?

jackkoenig:
Note that this is failing some tests:
```
[error] 	chiselTests.SourceLocatorSpec
[error] 	chiselTests.properties.ClassSpec
[error] 	chiselTests.properties.ObjectSpec
[error] 	chiselTests.DedupSpec
```

mwachs5 (commented):
super nit -- are we supposed to be bumping some firrtl version somewhere when we add new features to emit?

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+import chisel3.parameterization.{Case, OptionGroup}
+import chiselTests.{ChiselFlatSpec, MatchesAndOmits, Utils}
+import _root_.circt.stage.ChiselStage
+
+object Platform extends OptionGroup {
+  object FPGA extends Case
+  object ASIC extends Case
+}
+
+class TargetIO extends Bundle {
```
mwachs5, line 15: what happens if TargetIO takes a parameter and the different targets differ in their (new TargetIO(6) vs new TargetIO(8))? Mostly just curious where this error would be detected and what the error message looks like

nandor: `firtool` definitely checks for this. Should we also verify the structure of the IO bundles in Chisel?

nandor: Added a check to ensure that the bundles are type equivalent.

src/main/scala/chisel3/ModuleChoice.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.collection.immutable.ListMap
+
+import chisel3.FixedIOBaseModule
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal.{groupByIntoSeq, Builder, WireBinding}
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl.DefInstanceChoice
+import chisel3.internal.sourceinfo.InstChoiceTransform
+
+object ModuleChoice extends SourceInfoDoc {
```
jackkoenig, line 15: I think one other thing was to put this inside of `object choice` and just call it `Module`, then then general use will be `import chisel3.choice` and `choice.Module(...)`. Is this what y'all were thinking @azidar and @seldridge?

nandor: I don't want to overload the `Module` name, I think we should have a different one, even if it's in its own package.

src/main/scala/chisel3/ModuleChoice.scala
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.choice
```
jackkoenig, line 3: Stylistically, we need to add a directory for this package, so this file should be: `src/main/scala/chisel3/choice/ModuleChoice.scala`

core/src/main/scala/chisel3/Choice.scala
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.util.simpleClassName
+
+/** This package contains Chisel language definitions for describing configuration options and their accepted values.
+  */
+package object choice {
```
jackkoenig, line 10: Stylistically, this file should be named: `core/src/main/scala/chisel3/choice/package.scala`

core/src/main/scala/chisel3/choice/package.scala
```diff
@@ -0,0 +1,50 @@
...
+  /** An option group declaration. Specifies a container grouping values for some design configuration parameter.
+    *
+    * Example:
+    * ```
+    * import chisel3.option.{Group, Case}
+    * object Platform extends Group {
+    *   object FPGA extends Case
+    *   object ASIC extends Case
+    * }
+    * ```
```
jackkoenig, line 21:
```suggestion
    * @example
    * {{{
    * import chisel3.option.{Group, Case}
    * object Platform extends Group {
    *   object FPGA extends Case
    *   object ASIC extends Case
    * }
    * }}}
```
This is the way examples are done in Scala 2 ScalaDoc (I think Scala 3 might use markdown, alas).

core/src/main/scala/chisel3/internal/Builder.scala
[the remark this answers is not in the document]
jackkoenig: This still needs to actually use `groupByIntoSeq`

docs/src/explanations/instchoice.md
```diff
@@ -0,0 +1,53 @@
+---
+layout: docs
+title:  "Instance Choices"
+section: "chisel3"
+---
```
jackkoenig, line 5:
```suggestion
```
We need to clean up the other files, but this header stuff was used by Jekyll, it is no longer used by the new Docusaurus website.

docs/src/explanations/instchoice.md
```diff
@@ -0,0 +1,53 @@
...
+class TargetIO extends Bundle {
+  val in = Flipped(UInt(8.W))
+  val out = UInt(8.W)
+}
+
+class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO) {}
+
+class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO) {}
+
+class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO) {}
```
jackkoenig, line 45:
```suggestion
class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO)

class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO)

class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO)
```
Style nit. Also you probably don't need to specify the type parameter since it will be inferred, but if you prefer it it is also fine.

docs/src/explanations/instchoice.md
```diff
@@ -0,0 +1,53 @@
...
+
+class SomeModule extends RawModule {
+  val inst = ModuleChoice(new VerifTarget)(Seq(
+      Platform.FPGA -> new FPGATarget,
+      Platform.ASIC -> new ASICTarget
+  ))
+}
```
jackkoenig, line 52:
```suggestion
class SomeModule extends RawModule {
  val inst = ModuleChoice(new VerifTarget)(Seq(
    Platform.FPGA -> new FPGATarget,
    Platform.ASIC -> new ASICTarget
  ))
}
```
Tabbing is inconsistent.

src/main/scala/chisel3/choice/ModuleChoice.scala
```diff
@@ -0,0 +1,95 @@
...
+          throw new IllegalArgumentException(s"Error: duplicate case '${group.head.name}'")
+        }
+    }
+
+    val groupedChoices = choices.map(_._1.group).toSeq.distinct
+    if (groupedChoices.size == 0) {
+      throw new IllegalArgumentException("Error: at least one alternative must be specified")
+    }
+    if (groupedChoices.size != 1) {
+      throw new IllegalArgumentException("Error: module choices must belong to the same group of options")
```
jackkoenig, line 73: It would be good to include information about what groups are used in the error message.

src/main/scala/chisel3/choice/ModuleChoice.scala
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    groupByIntoSeq(choiceModules.map(_._1))(opt => opt).foreach {
+      case (_, group) =>
+        if (group.size != 1) {
+          throw new IllegalArgumentException(s"Error: duplicate case '${group.head.name}'")
+        }
+    }
+
+    val groupedChoices = choices.map(_._1.group).toSeq.distinct
```
jackkoenig, line 68:
```suggestion
    val groupedChoices = choices.map(_._1.group).distinct
```
`choices` is already a `Seq` so `.toSeq` is redundant

src/main/scala/chisel3/choice/ModuleChoice.scala
```diff
@@ -0,0 +1,95 @@
...
+    val binding = instDefaultModule.io.cloneTypeFull
+    binding.bind(WireBinding(Builder.forcedUserModule, Builder.currentWhen))
+
+    pushCommand(
+      DefInstanceChoice(
+        sourceInfo,
+        binding,
+        instDefaultModule,
+        group,
+        choiceModules.toSeq.map {
```
jackkoenig, line 86:
```suggestion
        choiceModules.map {
```
`choiceModules` is already a `Seq` so `.toSeq` is redundant

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+class TargetIO(width: Int) extends Bundle {
+  val in = Flipped(UInt(width.W))
+  val out = UInt(width.W)
+}
+
+class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO(8)) {}
+
+class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO(8)) {}
+
+class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO(8)) {}
```
jackkoenig, line 24:
```suggestion
class FPGATarget extends FixedIOExtModule[TargetIO](new TargetIO(8))

class ASICTarget extends FixedIOExtModule[TargetIO](new TargetIO(8))

class VerifTarget extends FixedIORawModule[TargetIO](new TargetIO(8))
```
Style nit, also same comment about not needing to specify the type parameter (but of course you can).

jackkoenig (approved):
1 problem (need `groupByIntoSeq`) and a few nits. LGTM otherwise.

src/main/scala/chisel3/choice/ModuleChoice.scala
nandor: If no case is specified, unfortunately we can't even infer the group.

jackkoenig: But if you have more than 1 you can mention what groups there were.

PR #3684, opened 2023-12-12
[main] Enable MiMa for v6.0.0-RC1

chiselbot:
Enable MiMa for v6.0.0-RC1


jackkoenig:
We're gonna have at least another RC so will hold off on enabling binary compatibility checking.

PR #3685, opened 2023-12-13
Define typeEquivalent specially for Property.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

When we changed Property to extend Element, we needed to update some internal logic that checked "are these Elements the same type". That logic needs to work a little differently for Property, and this internal API is one other place that needed to be updated.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Data.scala
+++ b/core/src/main/scala/chisel3/Data.scala
@@ -12,6 +12,7 @@ import chisel3.internal.Builder.pushCommand
 import chisel3.internal._
 import chisel3.internal.sourceinfo._
 import chisel3.internal.firrtl._
+import chisel3.properties.Property
 import chisel3.reflect.DataMirror
 import chisel3.util.simpleClassName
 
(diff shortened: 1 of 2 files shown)


mikeurbach:
You can see I needed to add something like this while making Property an Element here: https://github.com/chipsalliance/chisel/blob/d7c203a55d9341d9129559b6ff9e3cf8c853d3cf/core/src/main/scala/chisel3/experimental/dataview/package.scala#L149-L150

I missed `typeEquivalent` in `Data` originally, since I hadn't written an interesting enough test before. Since it's an internal method, I'm using FlatIO in the new test to exercise the `typeEquivalent` and the fix.

While I'm at it, I grepped for any other places where we do `getClass ==` in the codebase, since that doesn't work as expected with Property. I only found one other usage: https://github.com/chipsalliance/chisel/blob/d7c203a55d9341d9129559b6ff9e3cf8c853d3cf/core/src/main/scala/chisel3/Data.scala#L187

I'm not really familiar with that API, but I don't think it's meant to be used on Properties, so I haven't touched it.

jackkoenig (approved):

PR #3690, opened 2023-12-15
[binder] Truncate for aggregate connection, fix inner_sym

SpriteOvO:
Some fixes for binder:

- Truncate for aggregate connection.

- Set `inner_sym` only when needed.

### DIFF
--- a/binder/src/main/scala/PanamaCIRCTConverter.scala
+++ b/binder/src/main/scala/PanamaCIRCTConverter.scala
@@ -581,6 +581,124 @@ class PanamaCIRCTConverter extends CIRCTConverter {
         .build()
       firCtx.newItem(id, op.results(0))
     }
+
+    def emitConnect(dest: Reference.Value, srcVal: Reference.Value, loc: MlirLocation): Unit = {
+      val indexType = circt.mlirIntegerTypeGet(32)
+      var src = srcVal
+
+      // TODO: Strict connect
+
+      (dest.tpe, src.tpe) match {
+        case (fir.BundleType(fields), fir.BundleType(srcFields)) => {
+          assert(srcFields.size == fields.size)
+
+          def subField(index: Int, value: Reference.Value): Reference.Value = {
+            val opName = if (circt.firrtlTypeIsAOpenBundle(circt.mlirValueGetType(value.value))) {
+              "firrtl.opensubfield"
+            } else {
+              "firrtl.subfield"
+            }
+            val fieldTpe = value.tpe.asInstanceOf[fir.BundleType].fields(index).tpe
+            val op = util
+              .OpBuilder(opName, firCtx.currentBlock, loc)
+              .withNamedAttr("fieldIndex", circt.mlirIntegerAttrGet(indexType, index))
+              .withOperand(value.value)
+              .withResult(util.convert(fieldTpe))
+              .build()
+            Reference.Value(op.results(0), fieldTpe)
+          }
+
+          for (index <- 0 until fields.size) {
+            var destField = subField(index, dest)
+            var srcField = subField(index, src)
+            if (fields(index).flip == fir.Flip) {
+              emitConnect(srcField, destField, loc)
+            } else {
+              emitConnect(destField, srcField, loc)
+            }
+          }
+          return
+        }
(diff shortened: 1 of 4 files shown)


binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -574,13 +581,136 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
-        .withNamedAttr("inner_sym", circt.hwInnerSymAttrGet(name))
         .withOperand(input)
         .withResult(util.convert(resultType))
         // .withResult( /* ref */ )
         .build()
+      firCtx.ops += ((id._id, op.op))
       firCtx.newItem(id, op.results(0))
     }
+
+    def emitConnect(dest: Reference.Value, srcVal: Reference.Value, loc: MlirLocation): Unit = {
```
sequencer, line 592: This function is too mutable, before real releasing, we need to refactor.

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -574,13 +581,136 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
+      val indexType = circt.mlirIntegerTypeGet(32)
+      var src = srcVal
+
+      // TODO: Strict connect
+
+      if (dest.tpe.isInstanceOf[fir.BundleType]) {
```
sequencer, line 598: Use pattern match.

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -574,13 +581,136 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
+          if (destBundle.fields(index).flip == fir.Flip) {
+            emitConnect(srcField, destField, loc)
+          } else {
+            emitConnect(destField, srcField, loc)
+          }
+        }
+        return
+      }
+
+      if (dest.tpe.isInstanceOf[fir.VectorType]) {
```
sequencer, line 633: use pattern match

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -574,13 +581,136 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
+              .withNamedAttrs(Seq(("amount", circt.mlirIntegerAttrGet(circt.mlirIntegerTypeGet(32), destWidth))))
+              .withOperands(Seq(src.value))
+              .withResult(util.convert(dest.tpe))
+              .build()
+              .results(0),
+            dest.tpe
+          )
+        }
+      } else {
+        // TODO: const cast
```
sequencer, line 705: match and throw unimplemented error

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -574,13 +581,136 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
         .build()
+      firCtx.ops += ((id._id, op.op))
       firCtx.newItem(id, op.results(0))
     }
+
+    def emitConnect(dest: Reference.Value, srcVal: Reference.Value, loc: MlirLocation): Unit = {
+      val indexType = circt.mlirIntegerTypeGet(32)
+      var src = srcVal
+
+      // TODO: Strict connect
```
sequencer, line 596: match and throw unimplemented error

sequencer (changes requested):
Nitpick on the C++ style of writing Scala

binder/src/main/scala/PanamaCIRCTConverter.scala
SpriteOvO: The determination of strict connection may need adding something to FIRRTL C-API on the CIRCT side, I prefer just leave a comment here at the moment and fix it later.

binder/src/main/scala/PanamaCIRCTConverter.scala
SpriteOvO: The same reason as strict connection.

sequencer (approved):

PR #3691, opened 2023-12-18
dontTouch: Skip probe and properties.

dtzSiFive:
### Contributor Checklist

  - (None, title seems sufficient?)

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/dontTouch.scala
+++ b/core/src/main/scala/chisel3/dontTouch.scala
@@ -3,6 +3,8 @@
 package chisel3
 
 import chisel3.experimental.{annotate, requireIsHardware, ChiselAnnotation}
+import chisel3.properties.Property
+import chisel3.reflect.DataMirror
 import firrtl.transforms.DontTouchAnnotation
 
 /** Marks that a signal's leaves are an optimization barrier to Chisel and the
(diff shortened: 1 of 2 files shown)


jackkoenig:
I see you marked this API modification and BugFix, I went with BugFix for the label (which is used by release notes) because it seems better for `dontTouch` of something with Probes or Properties in it to "work" (for some definition of work) rather than crash. The API modification is something that used to crash now does something.

Happy to change it if @dtzSiFive or others disagree though.

dtzSiFive:
Okay, I agree with all that.  I can drop the "API modification" part, but no objection to your labeling or reasoning. :+1: .

core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -34,6 +36,8 @@ object dontTouch {
   def apply[T <: Data](data: T): T = {
     requireIsHardware(data, "Data marked dontTouch")
     data match {
+      case d if DataMirror.hasProbeTypeModifier(d) => ()
+      case _:   Property[_] => ()
```
seldridge, line 40: Explicitly returning `Unit`, i.e. `()`, is probably best here. Most places in the code base will not make this explicit and just put no text after the `=>`. I like the style here of `()`, though.

seldridge (approved):

mikeurbach (commented):
This makes sense to me, thanks for include Property along with Probe. If the others are on board, I am on board with this.

core/src/main/scala/chisel3/dontTouch.scala
mikeurbach: 👍 i agree that i like explicitly returning Unit

dtzSiFive: (Oh, neat! TIL, thanks!)

dtzSiFive:
Thanks, all!

PR #3696, opened 2023-12-19
Bump firtool-resolver to 1.1.0

jackkoenig:
This is sort of an BugFix to an unreleased API (added after 6.0.0-RC1), see the Release Notes for how this changes the behavior of `FIRTOOL_PATH` in a desirable way.

### Contributor Checklist

#### Type of Improvement

- Bugfix
- Dependency update

#### Desired Merge Strategy

- Squash

#### Release Notes

The logic for 
  * If FIRTOOL_PATH is set, the --version regex failing no longer will return fail, instead it will set version to `<unknown>`.
  * If FIRTOOL_PATH is set and something fails (firtool not found or cannot execute the binary), then the error will be returned rather than firtool-resolver falling back to fetching the default version.
  * Errors from running firtool will be logged and returned rather than unconditionally turned into "Firtool binary not on FIRTOOL_PATH".

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -199,7 +199,7 @@ lazy val chiselSettings = Seq(
     "org.scalatest" %% "scalatest" % "3.2.16" % "test",
     "org.scalatestplus" %% "scalacheck-1-16" % "3.2.14.0" % "test",
     "com.lihaoyi" %% "upickle" % "3.1.0",
-    "org.chipsalliance" %% "firtool-resolver" % "1.0.0"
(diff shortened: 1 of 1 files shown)


dtzSiFive (approved):
I have not performed the reviewer checklist, apologies.

However, I did test that this works for me using `FIRTOOL_PATH` pointing to a firtool with a non-release-tag firtool version string, as well as when not setting `FIRTOOL_PATH` and the downloaded executable does not work (producing nice diagnostic).

LGTM, thanks!

PR #3707, opened 2023-12-31
[binder] Add function for exporting split verilog

SpriteOvO:


binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -737,6 +737,21 @@ class PanamaCIRCTConverter extends CIRCTConverter {
     }
   }
 
+  def exportSplitVerilog(directory: String) = {
```
sequencer, line 740: Use `os.Path`

sequencer (changes requested):

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -737,6 +737,21 @@ class PanamaCIRCTConverter extends CIRCTConverter {
     }
   }
 
+  def exportSplitVerilog(directory: os.Path): Unit = {
+    def assertResult(result: MlirLogicalResult): Unit = {
+      assert(circt.mlirLogicalResultIsSuccess(result))
+    }
+
+    val pm = circt.mlirPassManagerCreate()
+    val options = circt.circtFirtoolOptionsCreateDefault()
```
sequencer, line 746: Use default options for now. In the future PRs, we should add api for config.

sequencer (approved):

PR #3714, opened 2024-01-06
Provide ImplicitClock and ImplicitReset

jackkoenig:
This also deleted a long-deprecated protected method that didn't do anything.

- [x] TODO add an example to the mdoc

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

These traits implement the functionality formerly only implemented in Module such that they can now be used by RawModules. They also define new protected virtual methods `implicitClock` and `implicitReset` that can be overridden within `Module` to change what values are used as the implicit clock and implicit reset respectively.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -237,8 +237,8 @@ abstract class Module extends RawModule with ImplicitClock with ImplicitReset {
   final val reset: Reset = IO(Input(mkReset))(this._sourceInfo).suggestName("reset")
   // TODO add a way to memoize hasBeenReset iff it is used
 
-  override def implicitClock: Clock = clock
-  override def implicitReset: Reset = reset
+  override protected def implicitClock: Clock = clock
+  override protected def implicitReset: Reset = reset
 
   private[chisel3] def mkReset: Reset = {
     // Top module and compatibility mode use Bool for reset
@@ -278,7 +278,7 @@ abstract class Module extends RawModule with ImplicitClock with ImplicitReset {
   *   // Define a Clock value, it need not be called "implicitClock"
   *   val clk = IO(Input(Clock()))
   *   // Implement the virtual method to tell Chisel about this Clock value
-  *   override def implicitClock = clk
+  *   override protected def implicitClock = clk
   *   // Now we have a Clock to use in this RawModule
   *   val reg = Reg(UInt(8.W))
   * }
@@ -287,7 +287,7 @@ abstract class Module extends RawModule with ImplicitClock with ImplicitReset {
 trait ImplicitClock { self: RawModule =>
 
   /** Method that should point to the user-defined Clock */
-  def implicitClock: Clock
+  protected def implicitClock: Clock
 
   Builder.currentClock = Some(Delayed(implicitClock))
 }
@@ -304,7 +304,7 @@ trait ImplicitClock { self: RawModule =>
   *   // Define a Reset value, it need not be called "implicitReset"
   *   val rst = IO(Input(AsyncReset()))
   *   // Implement the virtual method to tell Chisel about this Reset value
-  *   override def implicitReset = clk
+  *   override protected def implicitReset = clk
   *   // Now we have a Reset to use in this RawModule
   *   // Registers also require a clock
   *   val clock = IO(Input(Clock()))
@@ -315,7 +315,7 @@ trait ImplicitClock { self: RawModule =>
 trait ImplicitReset { self: RawModule =>
 
   /** Method that should point to the user-defined Reset */
-  def implicitReset: Reset
+  protected def implicitReset: Reset
 
   Builder.currentReset = Some(Delayed(implicitReset))
 }
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -295,6 +295,7 @@ abstract class Module extends RawModule with ImplicitClock with ImplicitReset {
   *   // Define a Clock value, it need not be called "implicitClock"
   *   val clk = IO(Input(Clock()))
   *   // Implement the virtual method to tell Chisel about this Clock value
+  *   // Note that though this is a def, the actual Clock is assigned to a val (clk above)
   *   override protected def implicitClock = clk
   *   // Now we have a Clock to use in this RawModule
   *   val reg = Reg(UInt(8.W))
@@ -312,7 +313,7 @@ trait ImplicitClock { self: RawModule =>
 /** Provides an implicit Reset for use _within_ the [[RawModule]]
   *
   * Be careful to define the Reset value before trying to use it.
-  * Due to Scala initialization order, the actual val defining the Reset must occur before any
+  * Due to Scala initialization order, the actual val defining the Reset object must occur before any
   * uses of the implicit Reset.
   *
   * @example
@@ -321,6 +322,7 @@ trait ImplicitClock { self: RawModule =>
   *   // Define a Reset value, it need not be called "implicitReset"
   *   val rst = IO(Input(AsyncReset()))
   *   // Implement the virtual method to tell Chisel about this Reset value
+  *   // Note that though this is a def, the actual Reset is assigned to a val (rst above)
   *   override protected def implicitReset = clk
   *   // Now we have a Reset to use in this RawModule
   *   // Registers also require a clock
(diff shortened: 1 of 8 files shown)


core/src/main/scala/chisel3/Module.scala
```diff
@@ -271,21 +254,76 @@ abstract class Module extends RawModule {
...
+  * @example {{{
+  * class MyModule extends Module with OverrideReset {
+  *   val gate = IO(Input(Bool()))
+  *   // It it useful to name this signal
+  *   val myReset = reset.asBool || gate
+  *   internalReset := myReset
+  * }
+  * }}}
+  */
+trait OverrideReset { self: Module =>
```
mwachs5, line 288: must this be a Module? can it work for a RawModule as well?

mwachs5 (commented):
Would the following work (not sure why i would want to override the clock and reset when there is already a .clock and .reset, that seems like an anti-pattern?)

class MyRawModule extends RawModule with OverrideClock {

  val myClock = IO(Input(Clock))
  val myReset = IO(Input(Reset))
  internalClock := myClock
  withClockAndReset(Module.clock, myReset) {
    ???
  }
}

jackkoenig:
I agree with making this apply to RawModule, that requires a bit more generalization but I think it's a good improvement to the codebase and API. It will make it possible for Module to not be "special" in that a user could create their own thing just like Module except with a different name for clock and reset, or including clock and reset but also with an automatic clock gate on the clock.

core/src/main/scala/chisel3/Module.scala
jackkoenig: Now it can!

core/src/main/scala/chisel3/Module.scala
```diff
@@ -283,6 +283,60 @@ abstract class Module extends RawModule {
...
+  *   // Define a Clock value, it need not be called "implicitClock"
+  *   val clk = IO(Input(Clock()))
+  *   // Implement the virtual method to tell Chisel about this Clock value
+  *   override protected def implicitClock = clk
+  *   // Now we have a Clock to use in this RawModule
+  *   val reg = Reg(UInt(8.W))
+  * }
+  * }}}
+  */
+trait ImplicitClock { self: RawModule =>
```
seldridge, line 304: Does this need to be `RawModule` or can it be `BaseModule`?

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -11,6 +11,7 @@ import chisel3.internal.Builder.Prefix
 import scala.util.Try
 import scala.annotation.implicitNotFound
 import scala.collection.mutable
+import chisel3.ChiselException
```
seldridge, line 14: Is this used?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -791,13 +794,35 @@ private[chisel3] object Builder extends LazyLogging {
...
+    result
+  }
+
+  /** Safely get the current Clock for use */
+  def currentClock: Option[Clock] =
+    dynamicContext.currentClock.map(d => getDelayed("clock", d))
+
+  /** Get the underlying box around current Clock, only used for saving the value */
+  def currentClockDelayed: Option[Delayed[Clock]] = dynamicContext.currentClock
+  def currentClock_=(newClock: Option[Delayed[Clock]]): Unit = {
```
seldridge, line 815: Do these defs need to return the delayed, or should they return the underlying value?

seldridge (approved):
Go. For it.

core/src/main/scala/chisel3/Module.scala
jackkoenig: `RawModule` is the parent of all modules that are allowed to create hardware--it's not clear to me what being able to mix this into a `BlackBox` would accomplish. Note that this API only handles the implicit clock _within the module_, if we were to add a similar API for the outside ("implicitly hook up this port to the implicit clock when instantiated"), then I agree it should work for BlackBoxes as well.

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -143,6 +144,14 @@ package object internal {
     }
   }
 
+  // This is similar to cats.Eval.later but we don't depend on Cats
+  private[chisel3] class Delayed[A](a: => A) {
```
mwachs5, line 148: i know this is private except within `chisel3` but it's also public for developers of this codebase (?) could we add some scaladoc for it

src/test/scala/chiselTests/ClockSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  property("Should be able to add an implicit clock to a RawModule") {
+    val verilog = ChiselStage.emitSystemVerilog(new RawModule with ImplicitClock {
+      val foo = IO(Input(Bool()))
+      val in = IO(Input(UInt(8.W)))
+      val out = IO(Output(UInt(8.W)))
+      override protected val implicitClock = (!foo).asClock
+
+      val r = Reg(UInt(8.W))
```
mwachs5, line 60: can we do a test that withClock(Module.clock) {...} still gives me the expected `implicitClock`

mwachs5: I meant that comment to apply for a `Module` as well

core/src/main/scala/chisel3/Module.scala
```diff
@@ -283,6 +283,60 @@ abstract class Module extends RawModule {
...
+  /** Method that should point to the user-defined Clock */
+  protected def implicitClock: Clock
+
+  Builder.currentClock = Some(Delayed(implicitClock))
+}
+
+/** Provides an implicit Reset for use _within_ the [[RawModule]]
+  *
+  * Be careful to define the Reset value before trying to use it.
+  * Due to Scala initialization order, the actual val defining the Reset must occur before any
```
mwachs5, line 315: nit: You say "the actual val" but then below you use a `def`. I can certainly imagine someone doing `override protected def implicitClock = InstantiateAClockGate(...)` and being very sad by all the clock gates getting put down. Should we use a `val` in the example body?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -791,13 +794,35 @@ private[chisel3] object Builder extends LazyLogging {
...
 
-  def currentClock: Option[Clock] = dynamicContext.currentClock
-  def currentClock_=(newClock: Option[Clock]): Unit = {
+  // Helper for reasonable errors when clock or reset value not yet initialized
+  private def getDelayed[A](field: String, dc: Delayed[A]): A = {
+    val result = dc.value
+    if (result == null) {
+      // TODO add SourceInfo and change to Builder.exception
+      throwException(
+        s"The implicit $field is null which means its definition probably has not yet been initialized."
```
mwachs5, line 803:
```suggestion
        s"The implicit $field is null which means the code that sets its definition has not yet executed."
```

maybe? This is an error in execution order that I can certainly imagine people running into.

core/src/main/scala/chisel3/internal/package.scala
jackkoenig: Will do

src/test/scala/chiselTests/ClockSpec.scala
jackkoenig: Are you basically asking that `Module.clock` returns `implicitClock`? I should even be able to just assert that

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: Yeah I think that phrasing will be better for most users (even if the "key phrase" to Google for is "Scala initialization order")

core/src/main/scala/chisel3/Module.scala
jackkoenig: The point I'm trying to illustrate is that the actual reset object needs to be a value, a `val`, but `implicitReset` can be a def. I could simplify the example a little bit by glossing over this subtly, but I was trying to show this point. What do you think I should do here? I could make this example very simple and put the more subtle example in a cookbook example.

jackkoenig: I could also just add more text explaining this fact.

jackkoenig: I've decided to keep the def but add a clarifying comment

PR #3715, opened 2024-01-07
[binder] Handle secret ports for instance and fix extmodule emission

SpriteOvO:


sequencer:
Please split it into two commits

sequencer:
LGTM
We are going to add more test after T1 is successfully elaborated.

sequencer (approved):

binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -781,10 +781,12 @@ class PanamaCIRCTConverter extends CIRCTConverter {
 
     val builder = util
       .OpBuilder("firrtl.extmodule", firCtx.circuitBlock, circt.unkLoc)
+      .withRegionNoBlock()
```
SpriteOvO, line 784: This line fixed the assertion at https://github.com/llvm/circt/blob/29ba8077af794dac78b90a8707da376489c398fe/lib/Dialect/FIRRTL/Transforms/LowerOpenAggs.cpp#L318

PR #3716, opened 2024-01-08
[binder] Handle secret commands and probe ops without `Node`

SpriteOvO:


binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -1518,7 +1523,8 @@ private[chisel3] object PanamaCIRCTConverter {
   }
   def visitDefModule(defModule: DefModule)(implicit cvt: CIRCTConverter): Unit = {
     cvt.visitDefModule(defModule)
-    defModule.commands.zip(defModule.commands.map(Some(_)).drop(1) :+ None).foreach {
+    val commands = defModule.commands ++ defModule.secretCommands
+    commands.zip(commands.map(Some(_)).drop(1) :+ None).foreach {
```
sequencer, line 1527: why drop 1

sequencer: Please add comments for understanding.

sequencer (approved):

PR #3728, opened 2024-01-12
make SRAM targets accessible from SRAMInterface

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Makes `SRAM` targets accessible from `SRAMInterface`s instantiated via the `SRAM` object, which sets the `underlying` field in `SRAMInterface`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import firrtl.annotations.{IsMember, Named, ReferenceTarget}
-import chisel3.internal.ExceptionHelpers
+import chisel3.internal.{ExceptionHelpers, NamedComponent}
 import chisel3.experimental.BaseModule
 
 import java.util.{MissingFormatArgumentException, UnknownFormatConversionException}
@@ -416,7 +416,7 @@ package object chisel3 {
 
   /** Contains universal methods for target accesses.
     */
-  trait HasTarget {
+  sealed trait HasTarget {
     def toTarget:         ReferenceTarget
     def toAbsoluteTarget: ReferenceTarget
     def toRelativeTarget(root: Option[BaseModule]): ReferenceTarget
@@ -427,7 +427,7 @@ package object chisel3 {
     /** This wrapping hides the actual object, ensuring users only have access
       * to the target methods (instead of the type of the underlying object).
       */
-    def wrap(t: HasTarget): HasTarget = new HasTarget {
+    private[chisel3] def apply(t: NamedComponent): HasTarget = new HasTarget {
       def toTarget = t.toTarget
       def toAbsoluteTarget = t.toAbsoluteTarget
       def toRelativeTarget(root: Option[BaseModule]) = t.toRelativeTarget(root)
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import firrtl.annotations.{IsMember, Named}
+import firrtl.annotations.{IsMember, Named, ReferenceTarget}
 import chisel3.internal.ExceptionHelpers
+import chisel3.experimental.BaseModule
 
 import java.util.{MissingFormatArgumentException, UnknownFormatConversionException}
 import scala.collection.mutable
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -79,14 +79,18 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
...
   */
 class SRAMInterface[T <: Data](
   memSize:           BigInt,
   tpe:               T,
   numReadPorts:      Int,
   numWritePorts:     Int,
   numReadwritePorts: Int,
-  masked:            Boolean = false)
+  masked:            Boolean = false,
+  val underlying:    Option[HasTarget] = None)
```
debs-sifive, line 93: @jackkoenig i am concerned that the `SRAMInterface.underlying` can be set to any random `HasTarget`… is that ok? Should i make a private setter or something? Or is that overkill

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    val chirrtl = chirrtlCircuit.serialize
+    chirrtl should include("module Top :")
+    chirrtl should include("smem sram_mem : UInt<8> [32]")
+    chirrtl should include(
+      "wire sram : { readPorts : { flip address : UInt<6>, flip enable : UInt<1>, data : UInt<8>}[0], writePorts : { flip address : UInt<6>, flip enable : UInt<1>, flip data : UInt<8>}[0], readwritePorts : { flip address : UInt<6>, flip enable : UInt<1>, flip isWrite : UInt<1>, readData : UInt<8>, flip writeData : UInt<8>}[1]}"
+    )
+
+    val dummyAnno = annos.collectFirst { case DummyAnno(t) => (t.toString) }
+    require(dummyAnno.nonEmpty)
+    dummyAnno.get should include("~Top|Top>sram_mem")
```
jackkoenig, line 43:
```suggestion
    dummyAnno should be(Some("~Top|Top>sram_mem"))
```

jackkoenig: We use `include` for the chirrtl matches mainly because there's a lot of other stuff in the chirrtl, but it's best to be precise when we can be.

src/main/scala/chisel3/util/SRAM.scala
jackkoenig: It's a fair point and making it a private setter would be more conservative. It would then be possible to make the underlying not propagate when the type is cloned which might be useful. There's still possibly some utility in having it be public and settable by users since someone could have their own implementation of SRAM that returns SRAMInterface and have the underlying point to their own `HasTarget` (maybe a module?).

In any case, I think going with the more conservative option is probably best since we can always open up the API later, so yeah go ahead and make it a private setter, something like:
```scala
  private[chisel3] var _underlying: Option[HasTarget] = None
  /** Useful ScalaDoc */
  def underlying: Option[HasTarget] = _underlying
```

core/src/main/scala/chisel3/package.scala
```diff
@@ -412,4 +413,25 @@ package object chisel3 {
...
+    /** This wrapping hides the actual object, ensuring users only have access
+      * to the target methods (instead of the type of the underlying object).
+      */
+    def wrap(t: HasTarget): HasTarget = new HasTarget {
+      def toTarget = t.toTarget
+      def toAbsoluteTarget = t.toAbsoluteTarget
+      def toRelativeTarget(root: Option[BaseModule]) = t.toRelativeTarget(root)
+    }
+
+  }
```
jackkoenig, line 436:
```suggestion
  sealed trait HasTarget {
    def toTarget:         ReferenceTarget
    def toAbsoluteTarget: ReferenceTarget
    def toRelativeTarget(root: Option[BaseModule]): ReferenceTarget
  }

  object HasTarget {

    /** This wrapping hides the actual object, ensuring users only have access
      * to the target methods (instead of the type of the underlying object).
      */
    private[chisel3] def apply(t: NamedComponent): HasTarget = new HasTarget {
      def toTarget = t.toTarget
      def toAbsoluteTarget = t.toAbsoluteTarget
      def toRelativeTarget(root: Option[BaseModule]) = t.toRelativeTarget(root)
    }
  }
```
I know I talked you into the original implementation, but being a little more conservative here we can make sure users don't inherit this type and instead just make it a box type that only Chisel internals can create. We can always open up the API a bit later.

jackkoenig (commented):
LGTM but I think we should be more closed with the API for now.

jackkoenig:
FWIW, I would call this a `Feature` as it's new information we're providing via the `SRAMInterface` API and I don't think it's breaking.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -79,14 +79,18 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
...
-  numWritePorts:     Int,
-  numReadwritePorts: Int,
-  masked:            Boolean = false)
+  memSize:                          BigInt,
+  tpe:                              T,
+  numReadPorts:                     Int,
+  numWritePorts:                    Int,
+  numReadwritePorts:                Int,
+  masked:                           Boolean = false,
+  private[chisel3] val _underlying: Option[HasTarget] = None)
```
jackkoenig, line 93: This isn't doing quite what you want it to, the constructor is public and thus users still have the ability to pass the argument. This really needs to be a `private[chisel3] var _underlying` down by `def` underlying (and then mutably set after running the constructor down below)

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -104,6 +105,14 @@ class SRAMInterface[T <: Data](
   val writePorts: Vec[MemoryWritePort[T]] = Vec(numWritePorts, new MemoryWritePort(tpe, addrWidth, masked))
   val readwritePorts: Vec[MemoryReadWritePort[T]] =
     Vec(numReadwritePorts, new MemoryReadWritePort(tpe, addrWidth, masked))
+
+  private[chisel3] var _underlying: Option[HasTarget] = None
+
+  /** Set underlying target information. This is intended to store information
+    * about the SRAM that instantiates this interface and should be set to None
+    * by default if this interface was instantiated directly.
+    */
```
jackkoenig, line 114: This comment should be a little more user-facing, something like "Target information for annotating the underlying SRAM if it is known"

jackkoenig (approved):

PR #3737, opened 2024-01-14
[main] Enable MiMa for v6.0.0-RC2

chiselbot:
Enable MiMa for v6.0.0-RC2


jackkoenig:
Not needed, we have https://github.com/chipsalliance/chisel/pull/3743

Issue #3735, opened 2024-01-13
Inconsistent behavior about zero-width operands


SpriteOvO:
**Type of issue**: Bug Report

```scala
class ZeroWidthTest extends RawModule {
  val a = IO(Input(UInt(32.W)))
  val b = IO(Input(SInt(32.W)))
  val c = IO(Output(UInt(32.W)))
  c := Cat(a >> 32, b)
}
```

**What is the current behavior?**

`c := Cat(a >> 32, b)` is represented as `UInt(32.W) := Cat(UInt(0.W), UInt(32.W))` in Chisel IR

**What is the expected behavior?**

In CIRCT MLIR, it represented as `UInt(33.W) := Cat(UInt(1.W), UInt(32.W))`, the first operand of the `Cat` is 1 width instead of zero-width.

```mlir
module {
  firrtl.circuit "ZeroWidthTest" {
    firrtl.module @ZeroWidthTest(in %a: !firrtl.uint<32>, in %b: !firrtl.sint<32>, out %c: !firrtl.uint<32>) attributes {convention = #firrtl<convention scalarized>} {
      %0 = firrtl.shr %a, 32 : (!firrtl.uint<32>) -> !firrtl.uint<1>
      %_c_T = firrtl.node interesting_name %0 : !firrtl.uint<1>
      %1 = firrtl.asUInt %b : (!firrtl.sint<32>) -> !firrtl.uint<32>
      %c_lo = firrtl.node interesting_name %1 : !firrtl.uint<32>
      %2 = firrtl.cat %_c_T, %c_lo : (!firrtl.uint<1>, !firrtl.uint<32>) -> !firrtl.uint<33>
      %_c_T_1 = firrtl.node interesting_name %2 : !firrtl.uint<33>
      %3 = firrtl.tail %_c_T_1, 1 : (!firrtl.uint<33>) -> !firrtl.uint<32>
      firrtl.strictconnect %c, %3 : !firrtl.uint<32>
    }
  }
}
```

**What is the use case for changing the behavior?**

It caused problems in Binder, when we visit the `connect` in the example we get IR `connect UInt(32.W), UInt(32.W)`, so there is no way to add a `tail` op like CIRCT does, as the width of the 2 operands are the same.

sequencer:
Basically, this problem was coming from the firtool parser did too much things, binder should not depend on the width inference result(but firtool can).

SpriteOvO:
According to the FIRRTL spec:

```
[...]

PR #3742, opened 2024-01-18
[binder] Fix emission of `firrtl.invalidvalue`

SpriteOvO:
Based on the merged PR in upstream CIRCT: llvm/circt#6577

Unsolved issue: #3735

CC @sequencer

### DIFF
--- a/binder/src/main/scala/PanamaCIRCTConverter.scala
+++ b/binder/src/main/scala/PanamaCIRCTConverter.scala
@@ -721,6 +721,114 @@ class PanamaCIRCTConverter extends CIRCTConverter {
         .withOperand( /* src */ src.value)
         .build()
     }
+
+    case class RecursiveTypeProperties(isPassive: Boolean, containsAnalog: Boolean)
+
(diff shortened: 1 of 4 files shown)


binder/src/main/scala/PanamaCIRCTConverter.scala
```diff
@@ -721,6 +725,114 @@ class PanamaCIRCTConverter extends CIRCTConverter {
...
+        }
+      ) match {
+        case 0 => Flow.None
+        case 1 => Flow.Source
+        case 2 => Flow.Sink
+        case 3 => Flow.Duplex
+      }
+    }
+
+    def emitInvalidate(value: Reference.Value, loc: MlirLocation): Unit = emitInvalidate(value, loc, foldFlow(value))
```
sequencer, line 782: I'm getting scared about binder require `flow`, which means: Rather than plain emit IR, Chisel need to query the states from MFC. I personally prefer we should improve the semantic of FIRRTL.

PR #3757, opened 2024-01-22
Update Scala CLI Template

jackkoenig:
### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

* Use "dep" instead of deprecated "lib"
* Move ChiselStage import above `import chisel3._`
  * This makes users less likely to run into issues adding `import chisel3.util._`
* Bump versions in Github workflows

- [ ] Did you review?

### DIFF
--- a/.github/workflows/build-scala-cli-template/chisel-template.scala
+++ b/.github/workflows/build-scala-cli-template/chisel-template.scala
@@ -1,11 +1,12 @@
 //> using repository "sonatype-s01:snapshots"
 //> using scala "2.13.12"
-//> using lib "org.chipsalliance::chisel::@VERSION@"
+//> using dep "org.chipsalliance::chisel::@VERSION@"
 //> using plugin "org.chipsalliance:::chisel-plugin::@VERSION@"
 //> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Xfatal-warnings", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"
 
 import chisel3._
-import circt.stage.ChiselStage
+// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._
+import _root_.circt.stage.ChiselStage
 
 class Foo extends Module {
   val a, b, c = IO(Input(Bool()))
--- a/.github/workflows/build-scala-cli-template.yml
+++ b/.github/workflows/build-scala-cli-template.yml
@@ -34,9 +34,9 @@ jobs:
       - name: Cache Scala-CLI
         uses: coursier/cache-action@v6
       - name: Setup Scala-CLI
-        uses: VirtusLab/scala-cli-setup@v0.2.1
+        uses: VirtusLab/scala-cli-setup@v1
         with:
-          jvm: adoptium:1.17
+          jvm: adoptium:17
           apps: sbt
       - name: Generate Scala CLI Template
         shell: bash


.github/workflows/build-scala-cli-template/chisel-template.scala
```diff
@@ -1,11 +1,11 @@
 //> using repository "sonatype-s01:snapshots"
 //> using scala "2.13.12"
-//> using lib "org.chipsalliance::chisel::@VERSION@"
+//> using dep "org.chipsalliance::chisel::@VERSION@"
 //> using plugin "org.chipsalliance:::chisel-plugin::@VERSION@"
 //> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Xfatal-warnings", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"
 
-import chisel3._
 import circt.stage.ChiselStage
+import chisel3._
```
jackkoenig, line 8: People are most likely to add `import chisel3.util._` right below `import chisel3._`, this makes it less likely that someone ends up with imports like:
```scala
import chisel3._
import chisel3.util._
```
Which leads the compiler to complain that `stage` is not a member of package `chisel3.util.circt`

seldridge: Consider using `_root_` here to keep things alphabetic.

seldridge (approved):

.github/workflows/build-scala-cli-template/chisel-template.scala
jackkoenig: so instead of
```scala
import chisel3._
```
Are you suggesting?
```scala
import chisel3._
import _root._circt.stage.ChiselStage
```
I don't hate it. I don't care that much about alphabetical order but I do want to try to help users avoid the common `object stage is not a member of package chisel3.util.circt` confusion.

What do you prefer?

mwachs5: sorry what is the underlying issue? Is the rule that if you do a wildcard import chisel always assumes that the prefix for anything after that?? Or is the issue that we have both a `_root` circt package and a `chisel3.circt` package? (or and a chisel3.util.circt` package?

jackkoenig: > if you do a wildcard import chisel always assumes that the prefix for anything after that?

Not Chisel, the Scala compiler, the problem being if you have

```scala
import chisel3.util._
```
Then because `chisel3.util.circt` exists, it then looks for `chisel3.util.circt.stage` which does not.

FWIW I am not in this spot trying to discuss ways to resolve that more fundamental issue. This is a problem in existing versions so I at least want to tweak our examples to guide users away from hitting it.

mwachs5: i think either is fine, but can we add a comment above wherever we put the ChiselStage import to this effect

azidar: I think using `_root_` is probably best, but I also think adding a comment would be helpful.

jackkoenig: `_root_` with comment coming right up!

jackkoenig:
@Mergifyio backport 6.x

jackkoenig:
Somehow this didn't get backported to 6.x, it may have occurred just after 6.0.0 was released but before the backporting flow was set up properly.

PR #3765, opened 2024-01-24
Add DataMirror.getLayerColor

seldridge:
Add a utility to reflect on the optional color of a layer.  This will be used by libraries that need to know if they need to put generated logic in a layerblock or not.

#### Release Notes

Add `DataMirror.getLayerColor` API to return the optional layer color of a `Data`.

### DIFF
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -61,9 +61,8 @@ object DataMirror {
     * @param x the `Data` to examine
     * @return a `Some[Layer]` if the data has a layer color, `None` otherwise
     */
-  def getLayerColor(x: Data): Option[layer.Layer] = x.probeInfo match {
-    case None                           => None
-    case Some(Data.ProbeInfo(_, color)) => color
+  def getLayerColor(x: Data): Option[layer.Layer] = x.probeInfo.collect {
+    case Data.ProbeInfo(_, Some(color)) => color
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -57,6 +57,15 @@ object DataMirror {
...
   def hasProbeTypeModifier(x: Data): Boolean = x.probeInfo.nonEmpty
 
+  /** Return the optional layer color of a `Data`.
+    * @param x the `Data` to examine
+    * @return a `Some[Layer]` if the data has a layer color, `None` otherwise
+    */
+  def getLayerColor(x: Data): Option[layer.Layer] = x.probeInfo match {
+    case None                           => None
+    case Some(Data.ProbeInfo(_, color)) => color
+  }
```
jackkoenig, line 67:
```suggestion
  def getLayerColor(x: Data): Option[layer.Layer] =
    x.probeInfo.collect { case Data.ProbeInfo(_, color) => color }
```
`case None => None` always suggests a `.collect` to me. But feel free to ignore 🙂

jackkoenig (approved):

core/src/main/scala/chisel3/reflect/DataMirror.scala
seldridge: This is a bit cleaner. Nice suggestion. Changed in https://github.com/chipsalliance/chisel/pull/3765/commits/2f1284f52e836357f845225483cbcc6a9267083a.

PR #3771, opened 2024-01-27
Convert cloneType to an extension method

jackkoenig:
This removes much of the use of this.type in Chisel which is necessary to upgrade to Scala 3.

There is still `this.type` used for other purposes:
* suggestName
* autoSeed
* LitArg
* ChiselEnum

I figure we can fix these in multiple IRs, this is the biggest one that is the most likely to cause breakages. While this is almost entirely source compatible, you may notice a few places I had to add `this.` in front of a `cloneType`. It will break any code where the user calls `cloneType` from within a `Bundle` or `Record` but leaves of the implicit `this.`. I'm not sure how big of an issue this is, we could potentially warn about it with the compiler plugin on 6.x.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -368,9 +368,7 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
     */
   def apply(idx: Int): T = self(idx)
 
-  override def cloneType: this.type = {
-    new Vec(gen.cloneTypeFull, length).asInstanceOf[this.type]
-  }
+  override protected def _cloneType: Data = new Vec(gen.cloneTypeFull, length)
 
   override def getElements: Seq[Data] = self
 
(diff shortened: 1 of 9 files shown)


jackkoenig:
I also may be entirely wrong about `this.type` in Scala 3. I tried adding `-Xsource:3` and didn't notice any errors from `this.type` so I tried using it in Scala3 and it seems to work...? https://scastie.scala-lang.org/dxX1lfOYQHSmYTTjHgcQEg

mwachs5:
That example compiles but isn't actually exercising anything but the .tostring? I changed the tostring and the message changed. Not sure what you are trying to show in the example?

jackkoenig:
As far as I can tell, this isn't actually needed for Scala 3 so I am going to close this. We may still want to make this change eventually, but it's a separate discussion from preparing the codebase for Scala 3 support.

PR #3772, opened 2024-01-28
Update installation.md

mwachs5:
Not everyone has `.` on their `$PATH`

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

#### Release Notes

Update Quickstart instructions for those without `.` on their `$PATH`

- [ ] Did you review?

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -13,7 +13,7 @@ The easiest way to install Chisel is to [install Scala CLI](https://scala-cli.vi
 
 ```bash
 curl -O -L https://github.com/chipsalliance/chisel/releases/latest/download/chisel-template.scala
-scala-cli chisel-template.scala
+./scala-cli chisel-template.scala
 ```
 
 The Chisel template is a simple, single-file example of Chisel that emits Verilog to the screen.


schoeberl:
Why `./`? 'scala-cli' could be anywhere in your path?

seldridge:
I agree with Martin here... this assumes that `scala-cli` is in the same directory where you have the template. I would assume that most users would follow the instructions to install `scala-cli` and put it on their `PATH`. This is what the instructions for installing `scala-cli` have the user do. After that, this command would work as-is. Making the change in this PR would _force_ users to put `scala-cli` into `.` which seems odd.

mwachs5:
For some reason I thought that these instructions were responsible for installing scala-cli

mwachs5:
The "for some reason" is that a reading of the line above this instruction sounds like this instruction is going to install scala-cli for you. Shouldn't we break the line up to say something like, "first install scala-cli, instructions <here>". Newline, "then get this project and run it with Scala-cli".

I apparently installed scala-cli a while back and forgot about it

jackkoenig:
> The "for some reason" is that a reading of the line above this instruction sounds like this instruction is going to install scala-cli for you. Shouldn't we break the line up to say something like, "first install scala-cli, instructions ". Newline, "then get this project and run it with Scala-cli".

Great point. Writing good instructions is hard 😅. I'll open a PR.

PR #3783, opened 2024-01-30
Support isProperty query in DataMirror.

mikeurbach:
### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference).

#### Release Notes

This adds an API to DataMirror to query if a Data is a Property.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [ ] Did you mark the proper milestone (Bug fix: `3.6.x`, `5.x`, or `6.x` depending on impact, API modification or big change: `7.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #3784, opened 2024-01-30
Support Property in BoringUtils.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This adds support for BoringUtils.bore to bore and connect Property ports.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -7,7 +7,7 @@ import scala.language.experimental.macros
 import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
 import chisel3.internal._
 import chisel3.experimental.hierarchy.{InstanceClone, ModuleClone}
-import chisel3.properties.{DynamicObject, StaticObject}
+import chisel3.properties.{DynamicObject, Property, StaticObject}
 import chisel3.internal.Builder._
 import chisel3.internal.firrtl.ir._
 import _root_.firrtl.annotations.{IsModule, ModuleTarget}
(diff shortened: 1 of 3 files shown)


mikeurbach:
Note that this PR is currently stacked on top of https://github.com/chipsalliance/chisel/pull/3783.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -205,7 +205,11 @@ abstract class RawModule extends BaseModule {
       case (true, false) if left.probeInfo.get.writable => ProbeDefine(si, left.lref, RWProbeExpr(Node(right)))
       case (true, false)                                => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
       case (false, true)                                => Connect(si, left.lref, ProbeRead(Node(right)))
-      case (false, false)                               => Connect(si, left.lref, Node(right))
+      case (false, false) =>
+        (left, right) match {
+          case (_: Property[_], _: Property[_]) => PropAssign(si, left.lref, Node(right))
+          case (_, _) => Connect(si, left.lref, Node(right))
```
sequencer, line 211: What if `left: Property[_]` and `right: _`? this should be forbidden explicitly and throw it at elaboration time.

sequencer (commented):
small nitpick

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: True. Maybe I can delegate to MonoConnect which checks these things. Or just add the runtime check here.

seldridge, line 211: What should happen for boring from property to data and when/what error will show up? I assume this will blow up later. Is the error message reasonable?

seldridge (approved):
Looks good. Question about errors on property/data connections.

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: Yep Jiuyang had a similar question. I will add an error test and make sure the error is reasonable.

mikeurbach: Ok, so this logic is only ever used to connect ports/wires created by BoringUtils itself. If the user bores from a Property to a Data, the Property ports will be bored and secret connections from Property to Property made here, but then when the bored output is connected to a non-Property data, you'll get the usual error:

```
Connection between sink (Foo.a: IO[Bool]) and source (chisel3.properties.Property$$anon$9@371acd0e) failed @: Sink (Bool) and Source (chisel3.properties.Property$$anon$9@66435227) have different types.
```

mikeurbach: See my comment on the other thread--this would only happen if there was a bug in the implementation of BoringUtils, so I think it is reasonable to just deal with the two cases we care about.

PR #3792, opened 2024-01-30
Add a new DynamicObject.apply method to create Class instances.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This is in addition to the existing support in Definition.apply. Sometimes it is not possible to use Definition.apply, for example, if you plan to bore ports through the Class being constructed. The new DynamicObject.apply method supports this, and creates a DynamicObject from the newly elaborated Class.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -133,7 +133,7 @@ abstract class RawModule extends BaseModule {
       // This also updates the source Class ref to the DynamicObject ref now that it's named.
       id.forceName(default = "_object", _namespace)
       id.getReference.setRef(id.getRef)
-      id.getSourceClass.foreach(_.setRef(id.getRef, true))
+      id.setSourceClassRef()
     }
     case id: StaticObject => {
       // Set the StaticObject's ref and Property[ClassType] type's ref to the BaseModule for the Class.
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -130,8 +130,10 @@ abstract class RawModule extends BaseModule {
     case id: DynamicObject => {
       // Force name of the DynamicObject, and set its Property[ClassType] type's ref to the DynamicObject.
       // The type's ref can't be set upon instantiation, because the DynamicObject hasn't been named yet.
+      // This also updates the source Class ref to the DynamicObject ref now that it's named.
       id.forceName(default = "_object", _namespace)
       id.getReference.setRef(id.getRef)
+      id.getSourceClass.foreach(_.setRef(id.getRef, true))
     }
     case id: StaticObject => {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/chisel3/properties/Object.scala
```diff
@@ -21,6 +25,15 @@ class DynamicObject private[chisel3] (val className: ClassType) extends HasId wi
...
   _parent.foreach(_.addId(this))
 
+  // Keep state for a reference to the Class from which the DynamicObject was created.
+  // This is used to update the Class ref to the DynamicObject ref, for Classes created via DynamicObject.apply.
+  private var _class: Option[Class] = None
+  protected[chisel3] def setSourceClass(cls: Class): Unit = {
+    require(!_class.isDefined, "Cannot set DynamicObject class multiple times")
+    _class = Some(cls)
+  }
+  protected[chisel3] def getSourceClass: Option[Class] = _class
```
jackkoenig, line 35:
```suggestion
  private var _class: Class = null
  protected[chisel3] def setSourceClass(cls: Class): Unit = {
    require(_class != null, "Cannot set DynamicObject class multiple times")
    _class = cls
  }
  protected[chisel3] def getSourceClass: Option[Class] = Option(_class)
```
It's a micro-optimization but one that saves a fair amount of memory if we end up having a lot of `DyanmicObjects` for whatever reason. We've done this same optimization for similar fields in `Data`.

core/src/main/scala/chisel3/properties/Object.scala
```diff
@@ -44,6 +57,36 @@ class DynamicObject private[chisel3] (val className: ClassType) extends HasId wi
   }
 }
 
+object DynamicObject {
```
jackkoenig, line 60: Now that it's possible for users to create these instances, we should include ScalaDoc with an example on `class DynamicObject`

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -130,8 +130,10 @@ abstract class RawModule extends BaseModule {
     case id: DynamicObject => {
       // Force name of the DynamicObject, and set its Property[ClassType] type's ref to the DynamicObject.
       // The type's ref can't be set upon instantiation, because the DynamicObject hasn't been named yet.
+      // This also updates the source Class ref to the DynamicObject ref now that it's named.
       id.forceName(default = "_object", _namespace)
       id.getReference.setRef(id.getRef)
+      id.getSourceClass.foreach(_.setRef(id.getRef, true))
```
jackkoenig, line 136: Since this is called twice and contains some real weirdness, should maybe this functionality be a method on `DynamicObject` that you call? eg.
```scala
id.setSourceClassRef()
```
then `setSourceClassRef` is something like
```scala
def setSourceClassRef(): Unit = this.getSourceClass.foreach(_.setRef(this.getRef, true))
```

jackkoenig (approved):
LGTM but see suggestions

PR #3799, opened 2024-01-30
Allow modules to globally enable layers

seldridge:
Add a function that allows the layer to be "enabled" within the current module scope.  This has the effect of emitting the module with the "enablelayer <layer>" FIRRTL format.  This is not currently in the FIRRTL spec---it is being prototyped inside CIRCT and is expected to land in the spec soon.

This is done to allow for the use case of a Test Harness needing to enable one or more layers (more than one layer is the union of several layers) inside a DUT in order to drive it properly.  The specific use case is to get access to an expensive core trace inside a microprocessor that is hidden behind a layer.

This was originally written using an API where the layer provided a trait that enabled the layer.  This runs afoul of Scala's restriction on inheriting the same trait multiple times.

FYI: @rwy7

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -49,6 +49,7 @@ object layer {
       case Layer.Root => name
       case _          => s"${parent.fullName}.$name"
     }
+
   }
 
   object Layer {
@@ -93,6 +94,12 @@ object layer {
     thunk
     Builder.pushCommand(LayerBlockEnd(sourceInfo))
     Builder.layerStack = Builder.layerStack.tail
+  }
+
+  /** Call this function from within a `Module` body to enable this layer globally for that module. */
+  final def enable(layer: Layer): Unit = layer match {
+    case Layer.Root =>
+    case _          => Builder.enabledLayers = layer :: Builder.enabledLayers
   }
 
 }
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -49,7 +49,6 @@ object layer {
       case Layer.Root => name
       case _          => s"${parent.fullName}.$name"
     }
-
   }
 
   object Layer {
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -99,7 +99,7 @@ object layer {
   /** Call this function from within a `Module` body to enable this layer globally for that module. */
   final def enable(layer: Layer): Unit = layer match {
     case Layer.Root =>
-    case _          => Builder.enabledLayers = layer :: Builder.enabledLayers
+    case _          => Builder.enabledLayers += layer
   }
 
 }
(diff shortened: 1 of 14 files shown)


core/src/main/scala/chisel3/Layer.scala
```diff
@@ -49,6 +49,7 @@ object layer {
       case Layer.Root => name
       case _          => s"${parent.fullName}.$name"
     }
+
```
jackkoenig, line 52: Nit I would remove this whitespace

src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -66,6 +68,21 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
     )()
   }
 
+  they should "be enabled with a trait" in {
+
+    class Foo extends RawModule {
+      layer.enable(A.B)
+      layer.enable(C)
+      // This should be a no-op.
+      layer.enable(layer.Layer.root)
```
jackkoenig, line 77: What happens if you enable a layer multiple times?

jackkoenig (approved):
Looks good to me, just a question and a nit

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -529,6 +529,7 @@ private[chisel3] class DynamicContext(
   var currentClock:   Option[Delayed[Clock]] = None
   var currentReset:   Option[Delayed[Reset]] = None
   var currentDisable: Disable.Type = Disable.BeforeReset
+  var enabledLayers:  List[layer.Layer] = Nil
```
mwachs5, line 532: should this be a `Set`?

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -456,10 +456,11 @@ private[chisel3] object Converter {
   }
 
   def convert(component: Component, typeAliases: Seq[String]): fir.DefModule = component match {
-    case ctx @ DefModule(id, name, ports, cmds) =>
+    case ctx @ DefModule(id, name, layers, ports, cmds) =>
       fir.Module(
         convert(id._getSourceLocator),
         name,
+        layers.map(_.fullName).reverse.distinct,
```
mwachs5, line 463: ah you put the .distinct here instead of the `Set` at first

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: We should definitely maintain order so if we were to use a Set, it should be an ordered one.

seldridge: A `LinkedHashSet[layer.Layer]` might be better and save the later `.distinct`. The only other requirement is it seems reasonable to preserve the order in which layers are enabled.

src/test/scala/chiselTests/LayerSpec.scala
seldridge: As Megan noticed, there's are made `.distinct` before going to FIRRTL. Currently, if a layer is enabled multiple times in CIRCT it has no additional effect (it's not disallowed). This should be ironed out on the FIRRTL spec side.

core/src/main/scala/chisel3/internal/Builder.scala
seldridge: Switched to `LinkedHashSet` in https://github.com/chipsalliance/chisel/pull/3799/commits/f19965417095daee7af695d5c9e0dd79d83959c6.

core/src/main/scala/chisel3/Layer.scala
seldridge: Fixed in https://github.com/chipsalliance/chisel/pull/3799/commits/54d826170888f31d1e3116ac674d34d2a3a9005b.

PR #3802, opened 2024-01-31
Add format strings to assert and assume

uenoku:
### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

This supports format strings for assert and assume statements as proposed in https://github.com/chipsalliance/firrtl-spec/pull/166. This change might break existing code if message contains `%`. 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -219,7 +219,11 @@ private[chisel3] object Converter {
           convert(probe, ctx, sourceInfo)
         )
       )
-    case e @ Verification(_, op, info, clk, pred, msg) =>
+    case e @ Verification(_, op, info, clk, pred, pable) =>
+      val (fmt, args) = unpack(pable, ctx)
+      if (op == Formal.Cover && args.nonEmpty) {
+        throwException("cover message cannot be used as a format string")
+      }
       val firOp = op match {
         case Formal.Assert => fir.Formal.Assert
         case Formal.Assume => fir.Formal.Assume
@@ -232,7 +236,8 @@ private[chisel3] object Converter {
           convert(clk, ctx, info),
           convert(pred, ctx, info),
           firrtl.Utils.one,
-          fir.StringLit(msg),
+          fir.StringLit(fmt),
+          args.map(a => convert(a, ctx, info)),
           e.name
         )
       )
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -221,9 +221,6 @@ private[chisel3] object Converter {
       )
     case e @ Verification(_, op, info, clk, pred, pable) =>
       val (fmt, args) = unpack(pable, ctx)
-      if (op == Formal.Cover && args.nonEmpty) {
-        throwException("cover message cannot be used as a format string")
-      }
       val firOp = op match {
         case Formal.Assert => fir.Formal.Assert
         case Formal.Assume => fir.Formal.Assume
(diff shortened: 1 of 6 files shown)


sequencer:
The Panama part lgtm. Although we don’t have enough test for now.

dtzSiFive (approved):
LGTM, but may be good to get another approval :+1: .

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -219,7 +219,11 @@ private[chisel3] object Converter {
           convert(probe, ctx, sourceInfo)
         )
       )
-    case e @ Verification(_, op, info, clk, pred, msg) =>
+    case e @ Verification(_, op, info, clk, pred, pable) =>
+      val (fmt, args) = unpack(pable, ctx)
+      if (op == Formal.Cover && args.nonEmpty) {
+        throwException("cover message cannot be used as a format string")
+      }
```
seldridge, line 226: This can be done as a `require` inside the constructor of the `Verification` op. (Arguably that motivates splitting this into separate IR structures for assert, assume, and cover, but that's not necessary.)

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -436,6 +436,7 @@ object Formal extends Enumeration {
   pred: Expression,
   en:   Expression,
   msg:  StringLit,
+  args: Seq[Expression],
```
seldridge, line 439: If we're doing this the data class way, then this needs an `@since(...)`. @jackkoenig: how do you want to handle this?

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -448,14 +449,17 @@ object Formal extends Enumeration {
...
+    Verification(op, info, clk, pred, en, msg, args, name)
   }
 }
 object Verification {
-  def unapply(s: Verification): Some[(Formal.Value, Info, Expression, Expression, Expression, StringLit)] = {
-    Some((s.op, s.info, s.clk, s.pred, s.en, s.msg))
+  def unapply(
+    s: Verification
+  ): Some[(Formal.Value, Info, Expression, Expression, Expression, StringLit, Seq[Expression])] = {
+    Some((s.op, s.info, s.clk, s.pred, s.en, s.msg, s.args))
```
seldridge, line 462: @jackkoenig: is this necessary to update for data class?

seldridge (approved):
This looks good. I'm requesting @jackkoenig to take a look for some of the data class pieces.

firrtl/src/main/scala/firrtl/ir/IR.scala
jackkoenig: If we keep the data class (implying we're trying to maintain source and binary compatibility), we should not update this (but should instead deprecate it). As is, this is making a breaking change, but I support just breaking it, see my other comment.

firrtl/src/main/scala/firrtl/ir/IR.scala
jackkoenig: We've deprecated everything in package firrtl._ and we should probably make it all private as part of Chisel 7 so I'm fine with just making the breaking change. WIth that, I'm fine with this as is, or we can change it back to a case class which would let us delete the unapply (and companion object) rather than updating it.

WDYT?

jackkoenig (approved):
I approve but please just change `@data class Verification` to a `case class`. To do this, replace `@data` with `case`, delete `@since`, delete `copy`, and delete the companion object and `unapply` method.

firrtl/src/main/scala/firrtl/ir/IR.scala
seldridge: I greatly prefer making all this private!

uenoku:
@jackkoenig @seldridge Could you merge this? Thanks!

PR #3804, opened 2024-01-31
Report source locator in when scoping error messages

jackkoenig:
Not really a bugfix but that seemed closest

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Data.scala
+++ b/core/src/main/scala/chisel3/Data.scala
@@ -582,7 +582,7 @@ abstract class Data extends HasId with NamedComponent with SourceInfoDoc {
     rec(leftType, rightType)
   }
 
-  private[chisel3] def isVisible: Boolean = isVisibleFromModule && isVisibleFromWhen
+  private[chisel3] def isVisible: Boolean = isVisibleFromModule && visibleFromWhen.isEmpty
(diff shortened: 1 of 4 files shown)


seldridge (approved):
Code looks good and this is definitely an improvement.

This hints at missing infrastructure for pretty printing multiple source locators. E.g., it would be useful if after doing one line-carat error pointing at one bad operation, you could then show the when that it escaped from. Just showing a bare source locator is better. 

Something like `attachNote` (https://mlir.llvm.org/docs/Diagnostics/#attaching-notes) is what is ideal.

jackkoenig:
> This hints at missing infrastructure for pretty printing multiple source locators. E.g., it would be useful if after doing one line-carat error pointing at one bad operation, you could then show the when that it escaped from. Just showing a bare source locator is better.

Yeah definitely. I thought about trying to tackle that when I was making that changed but just went with the smaller touch. These errors could also be changed to be non-fatal (report error rather than throwing exception).

jackkoenig:
It also would be super nice to have the source locator for where the current data was bound (eg. where the `Wire` was created). That's not currently very easy to figure out, but would definitely help improve this sort of error reporting quite a bit.

PR #3810, opened 2024-02-02
Add Property expressions, starting with addition.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This allows Properties to be used to build up expressions in terms of input Properties and literals.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Binding.scala
+++ b/core/src/main/scala/chisel3/internal/Binding.scala
@@ -3,9 +3,11 @@
 package chisel3.internal
 
 import chisel3._
-import chisel3.experimental.BaseModule
-import chisel3.internal.firrtl.ir.{LitArg, PropertyLit}
+import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal.firrtl.ir.{Arg, LitArg, PropertyLit}
 import chisel3.properties.Class
+
+import _root_.firrtl.ir.{PropPrimOp, PropertyType}
 
 import scala.collection.immutable.VectorMap
 
@@ -176,3 +178,16 @@ private[chisel3] case class BundleLitBinding(litMap: Map[Data, LitArg]) extends 
 private[chisel3] case class VecLitBinding(litMap: VectorMap[Data, LitArg]) extends LitBinding
 // Literal binding attached to a Property.
 private[chisel3] case object PropertyValueBinding extends UnconstrainedBinding with ReadOnlyBinding
+
+/** Binding for Property expressions.
+  *
+  * This binding is a little different, because Property expressions don't emit Nodes, so we sore all the information we
+  * need to create an in-memory Arg that captures the operation and operands.
+  *
+  * @param sourceInfo Source location where the expression was created
+  * @param enclosure BaseModule where the expression was created
+  * @param op Primitive operation for this expression
+  * @param args Arguments to this expression
+  */
+private[chisel3] case class PropExprBinding(sourceInfo: SourceInfo, enclosure: BaseModule, op: PropPrimOp, args: Arg*)
+    extends ConstrainedBinding
--- a/core/src/main/scala/chisel3/internal/Binding.scala
+++ b/core/src/main/scala/chisel3/internal/Binding.scala
@@ -3,11 +3,9 @@
 package chisel3.internal
 
 import chisel3._
-import chisel3.experimental.{BaseModule, SourceInfo}
-import chisel3.internal.firrtl.ir.{Arg, LitArg, PropertyLit}
+import chisel3.experimental.BaseModule
+import chisel3.internal.firrtl.ir.{LitArg, PropertyLit}
 import chisel3.properties.Class
-
-import _root_.firrtl.ir.{PropPrimOp, PropertyType}
 
 import scala.collection.immutable.VectorMap
 
@@ -178,16 +176,3 @@ private[chisel3] case class BundleLitBinding(litMap: Map[Data, LitArg]) extends 
 private[chisel3] case class VecLitBinding(litMap: VectorMap[Data, LitArg]) extends LitBinding
 // Literal binding attached to a Property.
 private[chisel3] case object PropertyValueBinding extends UnconstrainedBinding with ReadOnlyBinding
-
-/** Binding for Property expressions.
-  *
-  * This binding is a little different, because Property expressions don't emit Nodes, so we sore all the information we
-  * need to create an in-memory Arg that captures the operation and operands.
-  *
-  * @param sourceInfo Source location where the expression was created
-  * @param enclosure BaseModule where the expression was created
-  * @param op Primitive operation for this expression
-  * @param args Arguments to this expression
-  */
-private[chisel3] case class PropExprBinding(sourceInfo: SourceInfo, enclosure: BaseModule, op: PropPrimOp, args: Arg*)
-    extends ConstrainedBinding
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -186,6 +186,15 @@ private[chisel3] object ir {
     }
   }
 
+  /** Property expressions.
+    *
+    * Property expressions are conceptually similar to Nodes, but only exist as a tree of Args in-memory.
+    */
+  case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: Arg*)
+      extends Arg {
+    def name: String = throwException("Internal Error! PropExpr has no name")
+  }
+
   case class Ref(name: String) extends Arg
 
   /** Arg for ports of Modules
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -192,6 +192,9 @@ private[chisel3] object ir {
     */
   case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: List[Arg])
       extends Arg {
+    // PropExpr is different from other Args, because this is only used as an internal data structure, and we never name
+    // the Arg or use the name in textual FIRRTL. This is always expected to be the exp of a PropAssign, and it would be
+    // an internal error to request the name.
     def name: String = throwException("Internal Error! PropExpr has no name")
   }
 
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -190,7 +190,7 @@ private[chisel3] object ir {
     *
     * Property expressions are conceptually similar to Nodes, but only exist as a tree of Args in-memory.
     */
-  case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: Arg*)
+  case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: List[Arg])
       extends Arg {
     def name: String = throwException("Internal Error! PropExpr has no name")
   }
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -324,6 +324,8 @@ sealed trait Property[T] extends Element { self =>
     }
   }
 
+  final def +(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
+    ev.add(this, that)
 }
 
 private[chisel3] sealed trait ClassTypeProvider[A] {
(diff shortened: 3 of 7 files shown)


mikeurbach:
Opening a draft to share an early cut of this. The first commit shows the one time effort to support Property expressions, and new kinds of Property expressions can be added pretty easily similar to the second commit.

sequencer:
Is there any firrtl spec for it? Looks like an essential feature for property propagation

mwachs5 (commented):
Exciting! Maybe add a sentence to the Property explanations doc?

src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -617,4 +617,44 @@ class PropertySpec extends ChiselFlatSpec with MatchesAndOmits {
...
+      d := t + a
+      e := w + (a + b)
+    })
+
+    matchesAndOmits(chirrtl)(
+      "propassign w, integer_add(a, b)",
+      "propassign c, integer_add(a, b)",
+      "propassign d, integer_add(integer_add(a, b), a)",
+      "propassign e, integer_add(w, integer_add(a, b))"
+    )()
```
seldridge, line 645: This is interesting as it demonstrates an emission strategy for more compact FIRRTL. I.e., we could emit this way instead of with many, many nodes and the FIRRTL would be conceptually much closer to the underlying Chisel expressions.

mikeurbach: Yeah, I think this is the main thing I'm interested in getting feedback from in this draft PR. It's quite different from how expressions like this normally work, but I like this emission because a) we have always aimed for static single assignment of properties, b) it is completely trivial to parse in the FIRRTL compiler, and standard CSE makes short work of any duplicated subexpressions.

To expand a bit on how this is implemented... here's roughly the way it works for other Data:

Create new Data subtype object in memory
-> give it an OpBinding 
-> push a DefPrim command in the Builder
-> when the Data is used as an rvalue, make a Node

Compared to the new approach I tried out here:

Create a new Property object in memory
-> give it a PropExprBinding that holds the primitive operation and the operands
-> *dont* push any commands in the Builder
-> when the Data is used as an rvalue, make a PropExpr

The new PropExpr Arg is handled differently in the Converter and Serializer. Rather than the Converter seeing a DefPrim, converting it, and defining a new Node, the Converter essentially passes through the same information that was present in the PropExprBinding and PropExpr. The Serializer then recursively serializes the expression tree in whatever rvalue context it is used (only the RHS of propassign or as a sub-expression, presently).

I like the result, but since it diverges from how we normally use Bindings and Args, I am curious if this is a valid implementation or if I should look into other techniques.

mwachs5: we have no plan for any sort of last-connect semantics for properties, right? is all the complexity in the existing system described really there to enable that?

mikeurbach: > we have no plan for any sort of last-connect semantics for properties, right?

Yep, see 5. and 6. here: https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#property-assignments

> is all the complexity in the existing system described really there to enable that?

The use of Node for pretty much any rvalue pre-dates me diving into Chisel, so I'm not sure the historical reasons this has been done. I think making Nodes everywhere certainly makes some things simpler. I don't know if I'd say the existing approach exists to enable last connect though, since that's more about what you can do with lvalues than how you represent rvalues internally.

jackkoenig: > This is interesting as it demonstrates an emission strategy for more compact FIRRTL. I.e., we could emit this way instead of with many, many nodes and the FIRRTL would be conceptually much closer to the underlying Chisel expressions.

> Yeah, I think this is the main thing I'm interested in getting feedback from in this draft PR. It's quite different from how expressions like this normally work, but I like this emission because a) we have always aimed for static single assignment of properties, b) it is completely trivial to parse in the FIRRTL compiler, and standard CSE makes short work of any duplicated subexpressions.

I'm all for more compact FIRRTL that is closer to the original Chisel, but there are some important considerations:
1. What about when the user _wants_ to name the subexpression? Synthetic names are terrible so if the user assigns some subexpression to a val, we want that name which implies a `node` or something very much like a node.
2. We also need to think about how to track different source locators. `e := w + (a + b)` has 3 source locators, at least 3 columns, but if you had function calls they could different in line and source file as well. How do we record that information? It's possible with alternative source locator encoding, eg. SourceMaps, but it's generally important information.

mikeurbach: Thanks Jack, interesting points.

1. I know that this has been the Chisel -> Verilog contract, but I'm wondering if we don't need to uphold that contract any more. For one, using `val` names to control what is visible in the Verilog output has been a major source of frustration, both for users and FIRRTL compiler developers. For properties, we already have a different contract--the names of your classes and their output ports are what is guaranteed to be visible in the output (by the implementation, not the spec!). Maybe that is actually a decent contract, and we can treat intermediate property expressions more like how software compilers treat them--they have a name in the source, and you can look up the associated value while debugging if necessary (which gets into the second point).
2. I need to look into how Chisel is doing source locations, but there is probably something we can do here. All the source locations are there in memory, so if there is a format to emit compound source locations into, we can use it.

In general--I've been interested in this approach because the property stuff is new, and I'm trying to see if we can use an implementation that is closer to what we actually want for both Chisel and the FIRRTL compiler. The compiler is already operating on SSA form internally, so I think the value of nodes in the textual FIRRTL is pretty low. In order for this to work for end users, we need to figure out how to map the source locations in-memory in Chisel to the appropriate SSA values after parsing. Otherwise, I think this approach doesn't lose much for properties, and more closely matches how the compiler wants to see them.

On the flip side, we can easily go the more traditional Chisel route of spilling all intermediate expressions into nodes/wires if we decide that is preferable. I just wanted to try out something new and share it.

mikeurbach: > we can easily go the more traditional Chisel route of spilling all intermediate expressions into nodes/wires if we decide that is preferable

Discussed this synchronously today, and seems the consensus is this is preferable, so I will amend the first commit of this PR.

This did kick off an interesting discussion that we should continue, which might lead to a better FIRRTL spec around (sub-)expressions.

mikeurbach: One more nail in the coffin... the current approach doesn't support BoringUtils.bore from a property expression, which is something this should definitely support. I'll add a test for that as I update the branch.

mikeurbach:
Ok, I have updated the implementation slightly. The APIs for creating Property expressions now immediately create a temp wire and propassign the expression into it, returning the wire for use in other Property expressions.

I think this is the output we are looking for, but I'm curious for feedback on the internal implementation. I still kept the new PropExprBinding and PropExpr as an internal data structures that we can use instead of creating Nodes internally. Does this make sense given our goals? I've looked at how DefPrim works and is converted, and I'm not sure if we want to treat property expressions like a "command", or not.

mikeurbach:
Separately, I'll write some docs for this new feature.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -186,6 +186,15 @@ private[chisel3] object ir {
     }
   }
 
+  /** Property expressions.
+    *
+    * Property expressions are conceptually similar to Nodes, but only exist as a tree of Args in-memory.
+    */
+  case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: Arg*)
```
jackkoenig, line 193:
```suggestion
  case class PropExpr(sourceInfo: SourceInfo, tpe: firrtlir.PropertyType, op: firrtlir.PropPrimOp, args: List[Arg])
```
Don't use varargs in classes, just pick the container type. I'm guessing this is only ever 1-2 so List is a good choice.

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -176,3 +178,16 @@ private[chisel3] case class BundleLitBinding(litMap: Map[Data, LitArg]) extends
...
+  * This binding is a little different, because Property expressions don't emit Nodes, so we sore all the information we
+  * need to create an in-memory Arg that captures the operation and operands.
+  *
+  * @param sourceInfo Source location where the expression was created
+  * @param enclosure BaseModule where the expression was created
+  * @param op Primitive operation for this expression
+  * @param args Arguments to this expression
+  */
+private[chisel3] case class PropExprBinding(sourceInfo: SourceInfo, enclosure: BaseModule, op: PropPrimOp, args: Arg*)
+    extends ConstrainedBinding
```
jackkoenig, line 193: So this is tricky and highlight the weirdness of us both having `chisel3.internal.ir._` and the binding. But basically binding traditionally was only used to store enough information to check if the lexical scoping of something--basically to check if it is legal to access it in a given scope. The normal IR (ie. the `_ref` of a `Data`) is where information for emitting FIRRTL was stored.

There was obviously a bit of feature creep in the Literal bindings and in the DataView bindings which include things that are arguably more appropriate for the ref than for the binding.

This is a mess, and we need to clean it up, but IMO both the PropPrimOp and the Args do not belong here and are in fact redundant with the information you're putting in the IR node `PropExpr`.

It's true that SourceInfo is also not usually stored here, but I think it is reasonable to do so since it's useful in error reporting.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -315,11 +316,16 @@ sealed trait Property[T] extends Element { self =>
     requireIsHardware(this)
     requireVisible()
     topBindingOpt match {
+      case Some(PropExprBinding(sourceInfo, _, op, lhs, rhs)) =>
+        // For Property expressions, we create a special kind of Arg that builds in-memory trees without creating nodes.
+        ir.PropExpr(sourceInfo, tpe.getPropertyType(), op, lhs, rhs)
```
jackkoenig, line 321: I don't like getting this from the Binding rather than just setting the ref. The current system is a mess but this would just add yet-another-way information is stored and later derived.

jackkoenig (changes requested):
I don't like putting information that traditionally goes in `_ref` inside of the Binding, let's meet to chat about how we should do this.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
mikeurbach: Sounds good. I was copying this from DefPrim, but I can see why its better to not use varargs.

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -335,6 +338,66 @@ private[chisel3] object ClassTypeProvider {
   }
 }
 
+/** Typeclass for Property arithmetic.
+  */
+@implicitNotFound("arithmetic operations are not supported on Property type ${T}")
+trait PropertyArithmeticOps[T] {
```
jackkoenig, line 344:
```suggestion
sealed trait PropertyArithmeticOps[T] {
```
This should be sealed right?

mikeurbach: Oh, yes, I don't see why not.

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -257,6 +257,29 @@ case class PathPropertyLiteral(value: String) extends Expression with UseSeriali
 
 case class SequencePropertyValue(tpe: Type, values: Seq[Expression]) extends Expression with UseSerializer
 
+/** Property primitive operations.
+  */
+sealed trait PropPrimOp
+object PropPrimOp {
+  case object AddOp extends PropPrimOp {
+    override def toString(): String = "integer_add"
```
jackkoenig, line 265:
```suggestion
    override def toString: String = "integer_add"
```
Nit but `()` imply mutation which is obviously not the case here.

You could also consider making `PropPrimOp` an `abstract class` that takes the String representation as an argument and overriding `toString` there so that future Ops are a little more concise to add, feel free to not doing this though.

src/test/scala/chiselTests/properties/PropertySpec.scala
```diff
@@ -617,4 +618,124 @@ class PropertySpec extends ChiselFlatSpec with MatchesAndOmits {
...
+      e := w + (a + b)
+    })
+
+    matchesAndOmits(chirrtl)(
+      "wire t : Integer",
+      "propassign t, integer_add(a, b)",
+      "wire w : Integer",
+      "propassign w, t",
+      "propassign c, t",
+      "wire _d_WIRE : Integer",
```
jackkoenig, line 647: Do you want to give a less jarring name than `_WIRE`? You could do `.autoSeed` on the wires you're creating and give them something descriptive, eg. `_propOp` or something (do start it with an `_` though so that it's not treated as an important name downstream). eg. see what I did here: https://github.com/chipsalliance/chisel/blob/ff3766b6741fc519d6e7dd86127c9ee3f1b43809/core/src/main/scala/chisel3/Bits.scala#L692.

Note that it isn't picking up your val name `_wire` because the Chisel compiler plugin is not run on the core of Chisel.

jackkoenig (approved):
LGTM, please see comment about sealing `PropertyArithmeticOps` and other suggestions.

PR #3816, opened 2024-02-09
Update BoringUtils test for FIRRTL 4.0

seldridge:
Fix a BoringUtils test that was checking for the name of a define produce by CIRCT.  This was changed slightly in FIRRTL 4.0.  Align the test with the new Verilog macro name.


dtzSiFive (approved):
LGTM, thanks!

PR #3817, opened 2024-02-09
Revert "Shift right produce at least 1 bit width result (#3752)"

seldridge:
Revert changes to the width of static shift right.  Originally, Chisel would return a 0-bit value.  However, this violated the FIRRTL spec which mandated that this return a 1-bit value.  FIRRTL compilers would then clean this up, however, if a user looks at the Chisel-determined width they would get the wrong answer.

Commit 1634320ce299775307aafa4ffbffe8e18022e81f made Chisel align with FIRRTL.  After additional discussion, it was decided that the best course of action was to make FIRRTL _align with Chisel_ for UInt, but not for SInt.  Specifically, the following behaviors are what we want to move towards:

  1. The smallest width of a UInt shifted right is 0-bit
  2. The smallest width of an SInt shifted right is 1-bit

This will then be handled with changes to the FIRRTL spec and by FIRRTL compilers.  In the mean time, do not introduce incorrect behavior in Chisel and preserve things the way they were.

This reverts commit 1634320ce299775307aafa4ffbffe8e18022e81f.

### DIFF
--- a/core/src/main/scala/chisel3/Width.scala
+++ b/core/src/main/scala/chisel3/Width.scala
@@ -13,7 +13,7 @@ sealed abstract class Width {
   def max(that:              Width): Width = this.op(that, _ max _)
   def +(that:                Width): Width = this.op(that, _ + _)
   def +(that:                Int):   Width = this.op(this, (a, b) => a + that)
-  def shiftRight(that:       Int): Width = this.op(this, (a, b) => 1.max(a - that))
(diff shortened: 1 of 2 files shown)


seldridge:
CC: @sequencer @SpriteOvO

SpriteOvO (approved):
LGTM, nice to see it going in a more appropriate direction.

jackkoenig (approved):
Thanks! I am working on the firtool changes for the new width behavior so we should be able to shift over to it soon.

PR #3822, opened 2024-02-12
[website] Make dark theme styling better

jackkoenig:
* Restore default for links, "hyperlink blue" is very harsh in dark mode
* Change color of hero banner, primary color (Cal gold) is very harsh

We should probably just pick better primary colors, Cal gold is a bit much, but this is at least an improvement.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -37,6 +37,13 @@
   --docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.3);
 }
 
+/* Make hero banner less ugly for dark theme */
+ [data-theme='dark'] .hero--primary {
+  /** Defaults to primary color which is horrible for the hero background */
+  --ifm-hero-background-color: #2b307d; /* --ifm-color-primary-lightest from light theme */
+  --ifm-hero-text-color: var(--ifm-font-color-base);
+}
+
 .header-chipsalliance-link::before {
   content: '';
   background: url("/static/img/chips-alliance.svg");
@@ -45,12 +52,21 @@
   display: flex;
 }
 
-/* Make links more obvious */
+/* Make links more obvious [for light theme] */
 a {
   color: blue;
 }
 a:hover {
   color: blue;
+}
+
+/* Use default link color for dark theme */
+[data-theme='dark'] a {
+  color: var(--ifm-color-primary);
+}
+
+[data-theme='dark'] a:hover {
+  color: var(--ifm-color-primary);
 }
 
 /* Styling for the result of TOCInline


jackkoenig:
## Before

### Landing Page

Screenshot 2024-02-12 at 10 41 12 AM


### Page with Links

Screenshot 2024-02-12 at 10 41 28 AM



## After

### Landing Page

Screenshot 2024-02-12 at 10 40 13 AM

### Page with Links

Screenshot 2024-02-12 at 10 40 48 AM

seldridge (commented):
I can't comment on the CSS changes.

The link change overall looks good!

website/src/css/custom.css
```diff
@@ -37,6 +37,13 @@
   --docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.3);
 }
 
+/* Make hero banner less ugly for dark theme */
+ [data-theme='dark'] .hero--primary {
+  /** Defaults to primary color which is horrible for the hero background */
+  --ifm-hero-background-color: #2b307d; /* --ifm-color-primary-lightest from light theme */
+  --ifm-hero-text-color: var(--ifm-font-color-base);
```
richardxia, line 44: I think you should add these to the `[data-theme='dark']` rule above (lines 29–38), since it looks the variable name itself has "hero" in its name already, so there's no need to specifically select the `.hero--primary` element. This would allow the variables to be accessible for all elements under element that has the `theme="dark"` attribute set.

To help explain this a bit more, the formal name for CSS variables is [custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties), and they actually behave exactly normal properties (e.g. `font-size`, etc.) except that they have no builtin behavior. They are "cascaded" ("inherited") by all elements below them, so if you treat them as "variables", then it's as if the variable is scoped to all descendant elements. Also, in case if you're not familiar, `[data-theme='dark']` is the syntax for selecting an element based on its attributes, and the `data-` prefix is specifically for custom attributes, so it implies that there is an actual element in the DOM somewhere that has `data-theme="dark"` set somewhere, likely somewhere very high in the hierarchy. The values of the variables can change dynamically, if the `data-theme` attribute's value is changed dynamically, causing the other selectors to be applied and cascade down the tree.

The tl;dr for the previous paragraph is that you should basically treat the first two blocks in this file as essentially defining "global" variables that are conditionally applied based on the value of the `data-theme` attribute, and you should probably declare all other variables in those blocks unless you really want to override the variable for something more specific. I've been saying "variable" but they are really "constants" for all intents and purposes, since CSS rules cannot change the value of a variable as seen by a higher scope.

For consistency, you might also consider putting the light mode version in the `:root` rule above it, rather than relying on the default.

website/src/css/custom.css
```diff
@@ -45,14 +52,23 @@
...
 }
 
+/* Use default link color for dark theme */
+[data-theme='dark'] a {
+  color: var(--ifm-color-primary);
+}
+
+[data-theme='dark'] a:hover {
+  color: var(--ifm-color-primary);
+}
```
richardxia, line 70: Possibly out of scope, but there was a design fad for probably the last decade or so where people were removing the underlines from links for aesthetic reasons. I think we are probably starting to move back away from that because it impairs accessibility, and even for people who are fully sighted, it is sometimes difficult to identify links if the colors are too similar to the normal color. You might consider adding the underlines back for both light and dark themes (`text-decoration: underline`).

To make things look a bit prettier even with underlines, you could consider using some relatively new CSS properties for controlling the [offset](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset) and [thickness](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-thickness) of the underlines. (Note that older browsers, particularly Safari, [might not support these](https://caniuse.com/text-decoration), but it's still safe to put in the CSS properties because they degrade gracefully for browsers that don't support them.)

jackkoenig: I agree with this idea but want to do it as a separate change.

I have tried it out in the past, and it's less about the ugliness (although I do appreciate the suggestions for slight prettification, will definitely keep in mind). The main problem is the straightforward way to do it adds underlines _everywhere_, even to headers, sidebars, footers. I would like to apply the change but be a little more intentional with it. Thus, it just is more work than I'm intending for this particular PR.

website/src/css/custom.css
jackkoenig: I actually tried that originally, and as far as I can tell, it doesn't work, maybe we should chat offline about why because I have no idea 😂

PR #3824, opened 2024-02-15
Change the width of static shift right

jackkoenig:
This requires a recent change to CIRCT (https://github.com/llvm/circt/pull/6698).

Once firtool is released with the linked change, bumped in this repository, and this PR is merged, I think we should crack 7.0.0-M1 as it will serve as a useful milestone for dealing with this semantic change.

I also added a CLI option to emulate the old behavior: `--legacy-shift-right-width`. I'm not wedded to the name so if anyone has a better name for it, please suggest it.

Instructions on using this option will be in the release notes but we should probably also add a page to the website about migrating from Chisel 6 to Chisel 7 that calls this out.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Rebase

#### Release Notes

* A UInt shifted right by a static amount >= its width will now result
  in a 0-bit UInt
* An SInt shifted right by a static amount >= its width will now result
  in a 1-bit SInt (the sign bit)

This is a change for SInts which Chisel would treat the output as a 0-bit SInt. However, FIRRTL implemented different behavior where both UInts and SInts would result in 1-bit values (which shifted right by an amount >= the width of the input).

Users can emulate the old behavior by providing CLI option `--legacy-shift-right-width`. Users are encouraged to generate Verilog with and without this option and diff it to ensure the width change does not affect the correctness of their design. Note that this option is purely for code migration and should not be used long term--it will eventually be removed.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Bits.scala
+++ b/core/src/main/scala/chisel3/Bits.scala
@@ -677,7 +677,7 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
   override def do_<<(that: UInt)(implicit sourceInfo: SourceInfo): UInt =
     binop(sourceInfo, UInt(this.width.dynamicShiftLeft(that.width)), DynamicShiftLeftOp, that)
   override def do_>>(that: Int)(implicit sourceInfo: SourceInfo): UInt =
-    binop(sourceInfo, UInt(this.width.shiftRight(that)), ShiftRightOp, validateShiftAmount(that))
+    binop(sourceInfo, UInt(this.width.unsignedShiftRight(that)), ShiftRightOp, validateShiftAmount(that))
   override def do_>>(that: BigInt)(implicit sourceInfo: SourceInfo): UInt =
     this >> castToInt(that, "Shift amount")
   override def do_>>(that: UInt)(implicit sourceInfo: SourceInfo): UInt =
@@ -990,7 +990,7 @@ sealed class SInt private[chisel3] (width: Width) extends Bits(width) with Num[S
   override def do_<<(that: UInt)(implicit sourceInfo: SourceInfo): SInt =
     binop(sourceInfo, SInt(this.width.dynamicShiftLeft(that.width)), DynamicShiftLeftOp, that)
   override def do_>>(that: Int)(implicit sourceInfo: SourceInfo): SInt =
-    binop(sourceInfo, SInt(this.width.shiftRight(that)), ShiftRightOp, validateShiftAmount(that))
+    binop(sourceInfo, SInt(this.width.signedShiftRight(that)), ShiftRightOp, validateShiftAmount(that))
   override def do_>>(that: BigInt)(implicit sourceInfo: SourceInfo): SInt =
     this >> castToInt(that, "Shift amount")
   override def do_>>(that: UInt)(implicit sourceInfo: SourceInfo): SInt =
--- a/core/src/main/scala/chisel3/Bits.scala
+++ b/core/src/main/scala/chisel3/Bits.scala
@@ -17,6 +17,7 @@ import chisel3.internal.firrtl.ir.PrimOp._
 import _root_.firrtl.{ir => firrtlir}
 import chisel3.internal.{castToInt, Builder, Warning, WarningID}
 import chisel3.util.simpleClassName
+import scala.annotation.nowarn
 
 /** Exists to unify common interfaces of [[Bits]] and [[Reset]].
   *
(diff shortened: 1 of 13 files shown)


src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -124,38 +126,50 @@ trait ChiselRunners extends Assertions {
     assert(!runTester(t, additionalVResources, annotations))
   }
 
-  def assertKnownWidth(expected: Int)(gen: => Data): Unit = {
+  def assertKnownWidth(expected: Int, args: Iterable[String] = Nil)(gen: => Data)(implicit pos: Position): Unit = {
```
azidar, line 129: Huh, so this is for scalatest to properly pass the source locator to the assert? Interesting.

core/src/main/scala/chisel3/Bits.scala
```diff
@@ -676,8 +677,26 @@ sealed class UInt private[chisel3] (width: Width) extends Bits(width) with Num[U
     this << castToInt(that, "Shift amount")
   override def do_<<(that: UInt)(implicit sourceInfo: SourceInfo): UInt =
     binop(sourceInfo, UInt(this.width.dynamicShiftLeft(that.width)), DynamicShiftLeftOp, that)
-  override def do_>>(that: Int)(implicit sourceInfo: SourceInfo): UInt =
-    binop(sourceInfo, UInt(this.width.shiftRight(that)), ShiftRightOp, validateShiftAmount(that))
+
+  // Implement legacy [buggy] UInt shr behavior for both Chisel and FIRRTL
+  @nowarn("msg=method shiftRight in class Width is deprecated")
```
azidar, line 682: What does nowarn mean again?

azidar (approved):
LGTM!

core/src/main/scala/chisel3/Bits.scala
jackkoenig: I deprecated `Width.shiftRight` and it would trigger a warning here (and warnings are errors), so this `@nowarn` suppresses it. Obviously we are allowed to use our own deprecated APIs, but the API is public so I couldn't/shouldn't just delete it.

src/test/scala/chiselTests/SIntOps.scala
```diff
@@ -198,4 +198,110 @@ class SIntOpsSpec extends ChiselPropSpec with Utils {
...
+    assertInferredWidth(1, args) {
+      val in = IO(Input(SInt(0.W)))
+      val w = WireInit(SInt(), in)
+      w >> 8
+    }
+    assertInferredWidth(1, args) {
+      val in = IO(Input(SInt(0.W)))
+      val w = WireInit(SInt(), in)
+      w >> 0
+    }
```
seldridge, line 295: There's a lot of duplication between this and the previous test. Consider if the patch could combine them in some way.

There is related commonality between these and the UInt tests.

src/main/scala/chisel3/stage/ChiselOptions.scala
```diff
@@ -7,20 +7,22 @@ import chisel3.internal.WarningFilter
...
-  val chiselCircuit:       Option[Circuit] = None,
-  val sourceRoots:         Vector[File] = Vector.empty,
-  val warningFilters:      Vector[WarningFilter] = Vector.empty) {
+  val printFullStackTrace:   Boolean = false,
+  val throwOnFirstError:     Boolean = false,
+  val outputFile:            Option[String] = None,
+  val chiselCircuit:         Option[Circuit] = None,
+  val sourceRoots:           Vector[File] = Vector.empty,
+  val warningFilters:        Vector[WarningFilter] = Vector.empty,
+  val legacyShiftRightWidth: Boolean = false) {
```
seldridge, line 16: Nit: the naming of this may be better as `useLegacyShiftRightWidth`, i.e., something which is more obviously true/false.

seldridge (commented):
LGTM

Nits and suggestions for compacting some code and tests.

core/src/main/scala/chisel3/Width.scala
```diff
@@ -9,12 +9,18 @@ object Width {
...
+  def +(that:   Width): Width = this.op(that, _ + _)
+  def +(that:   Int):   Width = this.op(this, (a, b) => a + that)
+  @deprecated(
+    "The width of shift-right now differs by type, use unsignedShiftRight and signedShiftRight",
+    "Chisel 7.0.0"
+  )
+  def shiftRight(that:         Int): Width = this.op(this, (a, b) => 0.max(a - that))
+  def unsignedShiftRight(that: Int): Width = this.op(this, (a, b) => 0.max(a - that))
+  def signedShiftRight(that:   Int): Width = this.op(this, (a, b) => 1.max(a - that))
+  def dynamicShiftLeft(that:   Width): Width =
```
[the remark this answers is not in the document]
jackkoenig: I think the sealing is more than sufficient. I probably don't actually want them to be final because I have the ulterior motive of possibly introducing a "WidthChanged" private subclass of `KnownWidth` that could capture information about semantic changes in widths and warn when they are observable. I don't intend to implement it any time soon but there may be future changes were we would want to.

PR #3825, opened 2024-02-16
Add ifElseFatal intrinsics; use it for chisel3.assert emission

uenoku:
* Replaced printf-encoded assertion with ifElseFatal intrinsics[0]. Currently the intrinsic is a private object of core since we don't want to expose it other than chisel3.assert. 
* Removed outdated comments regarding verification statements. 
* Added a new case class `PrintableParam` for lazily constructing a string parameter for format string. 

[0] https://github.com/llvm/circt/blob/main/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md#circtchisel_ifelsefatal

### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

Support `circt_chisel_ifelsefatal` intrinsics and use it for chisel3.assert emission. 

- [x] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -27,7 +27,8 @@ package experimental {
   case class IntParam(value: BigInt) extends Param
   case class DoubleParam(value: Double) extends Param
   case class StringParam(value: String) extends Param
-  case class PrintableParam(value: chisel3.Printable, id: BaseModule) extends Param
+  /** Creates a parameter from the Printable's resulting format String */
+  case class PrintableParam(value: chisel3.Printable, context: BaseModule) extends Param
 
   /** Unquoted String */
   case class RawParam(value: String) extends Param
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -27,6 +27,7 @@ package experimental {
   case class IntParam(value: BigInt) extends Param
   case class DoubleParam(value: Double) extends Param
   case class StringParam(value: String) extends Param
+  case class PrintableParam(value: chisel3.Printable, id: BaseModule) extends Param
 
   /** Unquoted String */
   case class RawParam(value: String) extends Param
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -27,6 +27,7 @@ package experimental {
   case class IntParam(value: BigInt) extends Param
   case class DoubleParam(value: Double) extends Param
   case class StringParam(value: String) extends Param
+
   /** Creates a parameter from the Printable's resulting format String */
   case class PrintableParam(value: chisel3.Printable, context: BaseModule) extends Param
 
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -168,7 +168,7 @@ case class PString(str: String) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) =
     (str.replaceAll("%", "%%"), List.empty)
 
-  final def unpack_args(): Seq[Bits] = Seq.empty
+  final def unpackArgs: Seq[Bits] = List.empty
 }
 
 /** Superclass for Firrtl format specifiers for Bits */
@@ -178,7 +178,7 @@ sealed abstract class FirrtlFormat(private[chisel3] val specifier: Char) extends
     (s"%$specifier", List(bits.ref.fullName(ctx)))
   }
 
-  def unpackArgs(): Seq[Bits] = List(bits)
+  def unpackArgs: Seq[Bits] = List(bits)
 }
 object FirrtlFormat {
   final val legalSpecifiers = List('d', 'x', 'b', 'c')
@@ -219,17 +219,17 @@ case class Character(bits: Bits) extends FirrtlFormat('c')
 /** Put innermost name (eg. field of bundle) */
 case class Name(data: Data) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) = (data.ref.name, List.empty)
-  final def unpack_args(): Seq[Bits] = Seq.empty
+  final def unpackArgs: Seq[Bits] = List.empty
 }
 
 /** Put full name within parent namespace (eg. bundleName.field) */
 case class FullName(data: Data) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) = (data.ref.fullName(ctx), List.empty)
-  final def unpack_args(): Seq[Bits] = Seq.empty
+  final def unpackArgs: Seq[Bits] = List.empty
 }
 
 /** Represents escaped percents */
 case object Percent extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) = ("%%", List.empty)
-  final def unpack_args(): Seq[Bits] = Seq.empty
+  final def unpackArgs: Seq[Bits] = List.empty
 }
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -51,7 +51,7 @@ sealed abstract class Printable {
     */
   def unpack(ctx: Component): (String, Iterable[String])
 
-  /** Unpack into a List of captured Bits arguments
+  /** Unpack into a Seq of captured Bits arguments
     */
   def unpack_args(): Seq[Bits]
 
@@ -168,7 +168,7 @@ case class PString(str: String) extends Printable {
   final def unpack(ctx: Component): (String, Iterable[String]) =
     (str.replaceAll("%", "%%"), List.empty)
 
-  final def unpack_args(): Seq[Bits] = List.empty
+  final def unpack_args(): Seq[Bits] = Seq.empty
 }
 
 /** Superclass for Firrtl format specifiers for Bits */
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -53,7 +53,7 @@ sealed abstract class Printable {
 
   /** Unpack into a Seq of captured Bits arguments
     */
-  def unpack_args(): Seq[Bits]
+  def unpackArgs: Seq[Bits]
 
   /** Allow for appending Printables like Strings */
   final def +(that: Printable): Printables = Printables(List(this, that))
@@ -160,7 +160,7 @@ case class Printables(pables: Iterable[Printable]) extends Printable {
     (fmts.mkString, args.flatten)
   }
 
-  final def unpack_args(): Seq[Bits] = pables.map(_.unpack_args()).flatten.toSeq
+  final def unpackArgs: Seq[Bits] = pables.view.flatMap(_.unpackArgs).toList
 }
 
 /** Wrapper for printing Scala Strings */
@@ -178,7 +178,7 @@ sealed abstract class FirrtlFormat(private[chisel3] val specifier: Char) extends
     (s"%$specifier", List(bits.ref.fullName(ctx)))
   }
 
-  def unpack_args(): Seq[Bits] = Seq(bits)
+  def unpackArgs(): Seq[Bits] = List(bits)
 }
 object FirrtlFormat {
   final val legalSpecifiers = List('d', 'x', 'b', 'c')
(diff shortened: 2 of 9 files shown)


src/test/scala/chiselTests/naming/NamePluginSpec.scala
```diff
@@ -75,14 +75,12 @@ class NamePluginSpec extends ChiselFlatSpec with Utils {
       val foo, bar = IO(Input(UInt(8.W)))
 
       {
-        val x1 = chisel3.assert(1.U === 1.U)
```
uenoku, line 78: Deleted because a label doesn't make sense for chisel3.assert

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -151,12 +177,8 @@ object assert extends VerifPrintMacrosDoc {
   )(
     implicit sourceInfo: SourceInfo
   ): Assert = {
-    val id = new Assert()
-    when(!Module.reset.asBool) {
```
mwachs5, line 155: Should we remove the "this will be removed" comment above...?

core/src/main/scala/chisel3/experimental/VerificationIntrinsics.scala
```diff
@@ -0,0 +1,33 @@
...
+import chisel3.internal.firrtl.ir._
+import chisel3.experimental.{BaseModule, SourceInfo}
+
+import scala.annotation.nowarn
+import scala.reflect.macros.blackbox
+import chisel3.experimental.IntrinsicModule
+
+/** Create a module for `ifelsefatal` style assertion.
+  */
+private[chisel3] class Chisel_IfElseFatal[T <: Data](
```
mwachs5, line 17: Not sure about this underscore naming pattern

src/test/scala/chiselTests/naming/NamePluginSpec.scala
uenoku: The problem is that chisel3.assert never use plain FIRRTL assert. IfElseFatal style assertion doesn't have a label and we have ignored labels. I think it's fair to add a new API for emitting a plain FIRRTL assert and use that API here though.

core/src/main/scala/chisel3/experimental/VerificationIntrinsics.scala
uenoku: Thanks, I'll change it to `IfElseFatalIntrinsic`.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -139,6 +135,36 @@ object assert extends VerifPrintMacrosDoc {
...
+    def extract_args(pable: Printable): Seq[Bits] = pable match {
+      case Printables(pables) =>
+        val args = pables.map(p => extract_args(p))
+        args.flatten.toSeq
+      case PString(_) | Name(_) | FullName(_) | Percent => List.empty
+      case format: FirrtlFormat => List(format.bits)
+    }
+    val id = Builder.forcedUserModule // It should be safe since we push commands anyway.
+    val data = extract_args(format)
+    val inst = Module(new Chisel_IfElseFatal()(sourceInfo, id, format, data))
```
fabianschuiki, line 156: You might be able to use Definition/Instance for this, to limit the number of identical `intmodule`s that gets created in the FIRRTL output. LTL does a similar thing:
- https://github.com/chipsalliance/chisel/blob/main/src/main/scala/chisel3/ltl/LTL.scala#L112-L116

jackkoenig: We should probably do this, but since the format String is a parameter (rather than an argument), there isn't going to be a lot of benefit. This change is going to be pretty expensive in IR size. @uenoku can you provide some numbers for `.fir` size difference from this change for large designs?

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -15,15 +15,17 @@ import firrtl.options.{
 import firrtl.options.internal.WriteableCircuitAnnotation
 import firrtl.options.Viewer.view
 import chisel3.{deprecatedMFCMessage, ChiselException, Module}
-import chisel3.RawModule
+import chisel3.{RawModule}
```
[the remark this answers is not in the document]
jackkoenig: We only run scalafmt, not scalafix and scalafix is what handles imports. Relatedly, we should run scalafix 😂

core/src/main/scala/chisel3/VerificationStatement.scala
dtzSiFive: Oh, so that's where/how that's done!

With format strings there might not be a huge amount of re-use, especially as long as the strings are likely to contain the source information directly.  Might be good to do anyway, or as a follow-up?

I was briefly concerned about D/I'ing with arguments that need width or reset inference, the idea being if they all instantiate the same the inference behavior might do unexpected things, especially regarding reset.

Not sure what the spec says about these things, but in practice they become operations during parsing where this won't be a problem, just a thought.

jackkoenig: > I was briefly concerned about D/I'ing with arguments that need width or reset inference, the idea being if they all instantiate the same the inference behavior might do unexpected things, especially regarding reset.

That's a good point, not a problem here but I have had that problem elsewhere. I wonder if we could/should add a way to ask FIRRTL to monomorphize multiply-instantiated things rather than keeping a single instance (I have proposed this as the behavior for future explicit width parameters, but it might be useful to make it specifiable even in that case).

uenoku: I tried using DI but I got an error that `Instanciate` https://github.com/chipsalliance/chisel/blob/ef7de527452df171872349aec0c0493605bea526/core/src/main/scala/chisel3/experimental/hierarchy/Instantiate.scala#L48 requires macros. Is it possible to avoid this kind of error? The FIR text size increase is within 1% in large cores so I think it's fine to use the current approach.

jackkoenig: You would have to call the [underlying API](https://github.com/chipsalliance/chisel/blob/9448d32685b102f23bd6d97923fc45294635e1d0/core/src/main/scala/chisel3/experimental/hierarchy/Instantiate.scala#L113) of `Instantiate` directly since this is in `core` (the macro is just sugar around calling `Instantiate._impl`). In any case, since this is within 1% and I'm skeptical that D/I will save much here, let's just leave that as future work.

core/src/main/scala/chisel3/experimental/VerificationIntrinsics.scala
```diff
@@ -0,0 +1,25 @@
...
+private[chisel3] class IfElseFatalIntrinsic[T <: Data](
+  implicit sourceInfo: SourceInfo,
+  id:                  BaseModule,
+  format:              Printable,
+  data:                Seq[T])
+    extends IntrinsicModule(
+      "circt_chisel_ifelsefatal",
+      Map("format" -> chisel3.experimental.PrintableParam(format, id))
+    ) {
+  val clock = IO(Input(Clock())).suggestName("clock")
```
jackkoenig, line 18:
```suggestion
  // Because this code is in core, the Chisel compiler plugin does not run on it so we must .suggestName
  val clock = IO(Input(Clock())).suggestName("clock")
```

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -45,6 +45,10 @@ private[chisel3] object Converter {
     reportInternalError(s"Trying to convert a cloned IO of $mod inside of $mod itself$loc!")
   }
 
+  // Create a map from base module to the component.
+  def getIdToComponent(components: Seq[Component]): VectorMap[BaseModule, Component] =
+    VectorMap[BaseModule, Component](components.map(c => c.id -> c): _*)
```
jackkoenig, line 50:
```suggestion
    VectorMap(components.map(c => c.id -> c): _*)
```
This is a tiny nit, but is the duplication of the type parameters necessary?

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -427,11 +431,16 @@ private[chisel3] object Converter {
...
   }
 
-  def convert(name: String, param: Param): fir.Param = param match {
+  def convert(name: String, param: Param, idToComponent: VectorMap[BaseModule, Component]): fir.Param = param match {
     case IntParam(value)    => fir.IntParam(name, value)
     case DoubleParam(value) => fir.DoubleParam(name, value)
     case StringParam(value) => fir.StringParam(name, fir.StringLit(value))
-    case RawParam(value)    => fir.RawStringParam(name, value)
+    case PrintableParam(value, id) => {
+      val ctx = idToComponent.get(id).get
```
jackkoenig, line 439: Style nit that's irrelevant because we should get rid of the map, but `myMap.get(key).get` is the same thing as just `myMap(key)`, but see below:

`_component` is a package private field of `BaseModule`: https://github.com/chipsalliance/chisel/blob/9448d32685b102f23bd6d97923fc45294635e1d0/core/src/main/scala/chisel3/Module.scala#L829

You shouldn't need this map, just use `id._component.get`.  It's an `Option` because it is set at the end of Module construction, but that means it will be set before Conversion so it's safe to `.get`.

src/test/scala/chiselTests/naming/NamePluginSpec.scala
jackkoenig: You're saying it never had a label, but if it didn't, how did this test pass previously?

uenoku: chisel3.assert emitted an assert op along with print op. The assert op is actually a dummy op deleted by a parser and this test passed because the dummy assert had a label.

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -51,6 +51,10 @@ sealed abstract class Printable {
     */
   def unpack(ctx: Component): (String, Iterable[String])
 
+  /** Unpack into a Seq of captured Bits arguments
+    */
+  def unpack_args(): Seq[Bits]
```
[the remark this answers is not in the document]
jackkoenig: `Seq` is not `List`, it is the abstract type for ordered iterables so it's the exact right type to use here. It gives us the freedom to use `List` for small collections or `Vector` for large ones.

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -155,12 +159,16 @@ case class Printables(pables: Iterable[Printable]) extends Printable {
     val (fmts, args) = pables.map(_.unpack(ctx)).unzip
     (fmts.mkString, args.flatten)
   }
+
+  final def unpack_args(): Seq[Bits] = pables.map(_.unpack_args()).flatten.toSeq
```
jackkoenig, line 163:
```suggestion
  final def unpackArgs: Seq[Bits] = pables.view.flatMap(_.unpackArgs).toList
```
`.view` will ensure we don't create intermediate collections, `.flatMap` == `.map` + `.flatten`, and we should pick concrete types. These are typically small (and never massive) so `List` is best.

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -169,6 +177,8 @@ sealed abstract class FirrtlFormat(private[chisel3] val specifier: Char) extends
   def unpack(ctx: Component): (String, Iterable[String]) = {
     (s"%$specifier", List(bits.ref.fullName(ctx)))
   }
+
+  def unpack_args(): Seq[Bits] = Seq(bits)
```
jackkoenig, line 181:
```suggestion
  def unpackArgs(): Seq[Bits] = List(bits)
```
`Seq.apply` does create a `List`, but we should be concrete (note that this codebase very frequently uses `Seq(...)` to construct things but we should stop doing that).

core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -27,6 +27,7 @@ package experimental {
   case class IntParam(value: BigInt) extends Param
   case class DoubleParam(value: Double) extends Param
   case class StringParam(value: String) extends Param
+  case class PrintableParam(value: chisel3.Printable, id: BaseModule) extends Param
```
jackkoenig, line 30:
```suggestion
  /** Creates a parameter from the Printable's resulting format String */
  case class PrintableParam(value: chisel3.Printable, context: BaseModule) extends Param
```
Just some minor nits to make what this does a little more clear.

jackkoenig (approved):
A few last nits but looks good

uenoku:
@jackkoenig Thank you for reviews! Could you merge this?

dtzSiFive (approved):
LGTM!

PR #3826, opened 2024-02-16
Make `SRAMInterface` parameters publicly available

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature

#### Desired Merge Strategy

#### Release Notes

`memSize`, `dataType`, `numReadPorts`, `numWritePorts`, `numReadwritePorts`, `masked` parameters are now visible for `SRAMInterface`.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -82,12 +82,12 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
   * @param masked Whether the memory is write masked
   */
 class SRAMInterface[T <: Data](
-  memSize:           BigInt,
-  tpe:               T,
-  numReadPorts:      Int,
-  numWritePorts:     Int,
-  numReadwritePorts: Int,
-  masked:            Boolean = false)
+  val memSize:           BigInt,
+  val tpe:               T,
+  val numReadPorts:      Int,
+  val numWritePorts:     Int,
+  val numReadwritePorts: Int,
+  val masked:            Boolean = false)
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -82,12 +82,12 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
...
   */
 class SRAMInterface[T <: Data](
-  memSize:           BigInt,
-  tpe:               T,
-  numReadPorts:      Int,
-  numWritePorts:     Int,
-  numReadwritePorts: Int,
-  masked:            Boolean = false)
+  val memSize:           BigInt,
+  val tpe:               T,
```
jackkoenig, line 86: This actually can't be public in this way because making an argument to a class a `val` makes it a public field of the class in a way that is indistinguishable with public vals in the body. Since this is of type `Data`, it will become a field of the Bundle.

Instead, you can keep `tpe` not a `val` but add a public def below that just returns it, something like `def dataType: T = tpe`

debs-sifive: hmm I'm actually thinking of putting this into a case class so the whole interface doesn't have to get passed around all the time (per what Megan said). So that should avoid the `Data` issue

debs-sifive: ok nvm, switching back to a getter because of what Andrew said
> using case classes has the long-term issue of dead fields

jackkoenig (approved):
Note that "API Modification" means breaking something, just making stuff public is a new "Feature"

PR #3829, opened 2024-02-16
[svsim] Better error message when verilator not on PATH

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/verilator/Backend.scala
+++ b/svsim/src/main/scala/verilator/Backend.scala
@@ -3,7 +3,8 @@
 package svsim.verilator
 
 import svsim._
-import java.io.{BufferedReader, InputStreamReader}
+import scala.sys.process._
+import scala.collection.mutable
 
 object Backend {
   object CompilationSettings {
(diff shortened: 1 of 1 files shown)


svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -22,10 +23,12 @@ object Backend {
...
-    val process = Runtime.getRuntime().exec(Array("which", "verilator"))
-    val outputReader = new BufferedReader(new InputStreamReader(process.getInputStream()))
-    val executablePath = outputReader.lines().findFirst().get()
-    process.waitFor()
+    val output = mutable.ArrayBuffer.empty[String]
+    val exitCode = List("which", "verilator").!(ProcessLogger(output += _))
+    if (exitCode != 0) {
+      throw new Exception(s"verilator not found on the PATH!\n${output.mkString("\n")}")
+    }
+    val executablePath = output.head.trim
```
seldridge, line 31: This definitely works on macOS and linux. Does this work on Windows?

It may be better to instead do `verilator --version` which may be more platform independent.

seldridge (approved):
Question about avoiding `which`. May be able to be ignored.

I can't think of any easy way to test this code path other than with the something like applying the suggested change of using `verilator --lint-only` and then aliasing `verilator` to `false` in the test.

svsim/src/main/scala/verilator/Backend.scala
jackkoenig: Yeah I doubt `which` works on Windows except maybe WSL/WSL2. `verilator --version` won't give the path to it, which is what this code is trying to do (not sure why, it's how it's always been). I think I'll merge this fix as is (although will try to add a test), but point taken about trying to make it platform independent. That being said, I'm not sure how well Verilator works on Windows anyway since it also uses a Makefile that is very much written in a *nix style... Checking all of this on Windows is probably just a big task across the board.

jackkoenig: Rereading your suggestion (from review) about a test, I don't think there's any reasonable way to test the behavior of "which" here, at least not via ScalaTest. I'm not sure how valuable it is that this code is getting the full path to Verilator, but it does have some nice properties like ensuring the generated `Makefile` for replaying points to the exact Verilator rather than requiring the user to have the same environment.

PR #3835, opened 2024-02-17
Fix bug where Definitions are incorrectly public

### LINKED ISSUE #3834: Public is Incorrectly Applied to Definitions
A definition inside a public module (or the top module?) incorrectly gets marked as public. Consider the following:

```scala
//> using scala "2.13.12"
//> using lib "org.chipsalliance::chisel::6.0.0+74-0a437d8f-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::6.0.0+74-0a437d8f-SNAPSHOT"

import chisel3._
import chisel3.experimental.hierarchy.{
  Definition,
  Instance,
  instantiable
}

@instantiable
class Bar extends RawModule

class Foo extends Module {
  private val barDef = Definition(new Bar)
  private val bar = Instance(barDef)
}

object Main extends App {
  println(ChiselStage.emitCHIRRTL(new Foo))
}
```

This produces the following FIRRTL with `scala-cli Foo.scala`:

```
FIRRTL version 4.0.0
circuit Foo :
  public module Bar :

    skip

  public module Foo :
    input clock : Clock
    input reset : UInt<1>

    inst bar of Bar
```

Module `Bar` should not be `public`.

seldridge:
Drop an argument from the Builder.build method which is only used to
change behavior when building Definitions.  Instead, rely on the
`inDefinition` variable made available by the DynamicContext.

Fix a bug where a Definition's root module would be incorrectly marked
public.  Definitions currently create a separate Circuit for each
Definition.  The logic to control marking the main module public relied on
marking the last module in the circuit public.  However, this incorrectly
assumed that there was one Circuit when, in actuality, there are
multiple.  (This commit ignores questions of whether or not it should work
this way.)

Instead, use the DyanmicContext `inDefinition` variable to execute this
code path to set the main module public when the Builder is not working
with a Definition.

Fixes #3834.

### DIFF
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
@@ -112,7 +112,7 @@ object Definition extends SourceInfoDoc {
       )
     }
     dynamicContext.inDefinition = true
-    val (ir, module) = Builder.build(Module(proto), dynamicContext, false)
+    val (ir, module) = Builder.build(Module(proto), dynamicContext)
     Builder.components ++= ir.components
     Builder.annotations ++= ir.annotations: @nowarn // this will go away when firrtl is merged
     module._circuit = Builder.currentModule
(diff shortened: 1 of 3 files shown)


seldridge:
This PR is structured as two logical commits. The first commit does some light refactoring of the existing code. The second actually applies the change.

jackkoenig (approved):
LGTM!

PR #3844, opened 2024-02-22
Add Property expression for integer multiplication.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This adds an API for integer Property multiplication.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -321,9 +321,13 @@ sealed trait Property[T] extends Element { self =>
     }
   }
 
+  /** Perform addition as defined by FIRRTL spec section Integer Add Operation.
+    */
   final def +(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
     ev.add(this, that)
 
+  /** Perform multiplication as defined by FIRRTL spec section Integer Multiply Operation.
+    */
   final def *(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
     ev.mul(this, that)
 }
--- a/core/src/main/scala/chisel3/properties/Property.scala
+++ b/core/src/main/scala/chisel3/properties/Property.scala
@@ -323,6 +323,9 @@ sealed trait Property[T] extends Element { self =>
 
   final def +(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
     ev.add(this, that)
+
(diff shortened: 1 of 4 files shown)


core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -323,6 +323,9 @@ sealed trait Property[T] extends Element { self =>
 
   final def +(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
```
mwachs5, line 324:
```suggestion
  /** Perform addition as defined by FIRRTL spec section 25.1.1. (as of 4.0.0) */
  final def +(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
```

or maybe there is a more future-proof way of referring to that

core/src/main/scala/chisel3/properties/Property.scala
```diff
@@ -323,6 +323,9 @@ sealed trait Property[T] extends Element { self =>
...
     ev.add(this, that)
+
+  final def *(that: Property[T])(implicit ev: PropertyArithmeticOps[Property[T]], sourceInfo: SourceInfo): Property[T] =
+    ev.mul(this, that)
```
mwachs5, line 328: can you copy the great info from `properties.md` as scaladoc here?

docs/src/explanations/properties.md
```diff
@@ -143,6 +143,7 @@ class IntegerArithmeticExample extends RawModule {
 The following table lists the possible arithmetic operators that are supported
 on integral `Property` typed values.
 
-| Operation | Description                                                |
-| --------- | -----------                                                |
-| `+`       | Perform addition as defined by FIRRTL spec section 25.1.1. |
+| Operation | Description                                                      |
+| --------- | -----------                                                      |
+| `+`       | Perform addition as defined by FIRRTL spec section 25.1.1.       |
```
mwachs5, line 148: nit: it seems likely that these section numbers would move around; are we planning to keep updating this? Maybe we can refer to the section by name instead which seems likely to be more stable?

core/src/main/scala/chisel3/properties/Property.scala
mikeurbach: Ah, yes I'll add scaladoc to the public methods, thanks for pointing it out.

docs/src/explanations/properties.md
mikeurbach: hmm yeah, they probably will drift. i think referencing by name sounds good.

mwachs5 (approved):

PR #3857, opened 2024-02-23
Relax legality of defines, align with FIRRTL spec

seldridge:
Change the checking logic for the legality of a probe define destination to align with FIRRTL specification language introduced recently [1]. Specifically, this enforces the following sentence:

> An operation that "writes" to a probe must be in a region whose layer color is enabled when the probe's layer color is enabled.

Previously, Chisel was more restrictive and would only allow writing to a probe whose color was the same as the layer block you are currently in. However, as has been worked out in CIRCT and elsewhere, it is fine to drive a probe from any region that is enabled when the probe is also enabled.

Practically, this gets some Chisel code passing that was legal by a FIRRTL compiler, but Chisel was refusing to generate it.

CC: @rwy7

[1]: https://github.com/chipsalliance/firrtl-spec/commit/673aa08cc31883aa90eca5ceaa6bf0017f3165fa

#### Release Notes

Align Chisel's checks of define destinations with the FIRRTL specification. Previously, Chisel was too strict.

### DIFF
--- a/core/src/main/scala/chisel3/internal/package.scala
+++ b/core/src/main/scala/chisel3/internal/package.scala
@@ -143,7 +143,7 @@ package object internal {
         color
       case _ => return
     }
-    val enabledLayers = Builder.enabledLayers ++ Builder.layerStack.headOption
+    val enabledLayers = Builder.enabledLayers.view ++ Builder.layerStack.headOption
     if (enabledLayers.exists(_.canWriteTo(destLayer))) {
       return
     }
--- a/core/src/main/scala/chisel3/internal/package.scala
+++ b/core/src/main/scala/chisel3/internal/package.scala
@@ -143,7 +143,7 @@ package object internal {
         color
       case _ => return
     }
-    val enabledLayers = Builder.enabledLayers ++= Builder.layerStack.headOption
+    val enabledLayers = Builder.enabledLayers ++ Builder.layerStack.headOption
     if (enabledLayers.exists(_.canWriteTo(destLayer))) {
       return
     }
--- a/core/src/main/scala/chisel3/internal/package.scala
+++ b/core/src/main/scala/chisel3/internal/package.scala
@@ -143,8 +143,11 @@ package object internal {
         color
       case _ => return
     }
-    if (!Builder.layerStack.exists(_ == destLayer))
-      Builder.error(errorMessage)
+    val enabledLayers = Builder.enabledLayers ++= Builder.layerStack.headOption
+    if (enabledLayers.exists(_.canWriteTo(destLayer))) {
+      return
+    }
+    Builder.error(errorMessage)
   }
 
   // TODO this exists in cats.Traverse, should we just use that?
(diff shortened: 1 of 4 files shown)


src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -48,23 +48,29 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
       layer.block(A) {
-        define(a, ProbeValue(in))
+        define(x, ProbeValue(c))
+        define(y, ProbeValue(d))
         layer.block(A.B) {
-          define(b, ProbeValue(in))
+          define(y, ProbeValue(e))
         }
       }
     }
```
seldridge, line 66: This will produce illegal FIRRTL as the probes are defined more than once. However, it makes the test terser to write it this way.

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -143,8 +143,11 @@ package object internal {
         color
       case _ => return
     }
-    if (!Builder.layerStack.exists(_ == destLayer))
-      Builder.error(errorMessage)
+    val enabledLayers = Builder.enabledLayers ++= Builder.layerStack.headOption
```
jackkoenig, line 146: I don't fully grok this mutation, shouldn't the enabling of layers happen separately from the checking of them?

src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -83,43 +89,48 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
 
-  "Layers error checking" should "require that a nested layer definition matches its declaration nesting" in {
+  they should "allow to define additional layer-colored probes using enables" in {
 
     class Foo extends RawModule {
-      layer.block(A.B) {
-        val a = Wire(Bool())
+      // Without this enable, this circuit is illegal because `C` is NOT enabled
+      // when `A` is enabled.  Cf. tests checking errors of this later.
+      layer.enable(A)
```
jackkoenig, line 97: Is enable the global enable for this Module?

jackkoenig (approved):
I don't fully grok some of the logic but the tests look good

core/src/main/scala/chisel3/internal/package.scala
seldridge: This is actually wrong. It should just be `++`. No mutation. Will fix...

src/test/scala/chiselTests/LayerSpec.scala
seldridge: Yes, it's indicating that this module has layer `A` enabled. This will emit:

```
public module Foo enablelayer A :
  ; ...
```

If there is more than one enable, then those will get appended, e.g., `layer.enable(A); layer.enable(B)` will be:

```
public module Foo enablelayer A enablelayer B :
  ; ...
```

core/src/main/scala/chisel3/internal/package.scala
seldridge: Fixed in https://github.com/chipsalliance/chisel/pull/3857/commits/023776502bd2207a37222eeca60b598bb1cffc22.

PR #3865, opened 2024-02-23
[main] Enable MiMa for v7.0.0-M1

chiselbot:
Enable MiMa for v7.0.0-M1


jackkoenig:
We don't want binary compatibility checking yet.

PR #3866, opened 2024-02-23
Create method for marking all ports of a `Module` with `DontTouchAnnotation`

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Create `dontTouch.modulePorts` method for marking all ports of a `Module` with `DontTouchAnnotation`.

- [ ] Did you review?


seldridge:
Is this necessary now that `Public` exists?

core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -46,4 +47,18 @@ object dontTouch {
...
+  /** Mark a module so that its ports will not get optimized away.
+    *
+    * @param module whose ports shouldl not get optimized away
+    */
+  def modulePorts(module: BaseModule): BaseModule = {
+
+    annotate(new ChiselMultiAnnotation {
+      def toFirrtl: Seq[Annotation] = (DataMirror.fullModulePorts(module).map {
+        case (_, data) => DontTouchAnnotation(data.toNamed)
+      })
```
jackkoenig, line 59: `fullModulePorts` is returns every Aggregate and Element so this will result in dontTouching things multiple times. Can we instead maybe just use `.modulePorts` and call `dontTouch.apply` above (which does the recursion and only marks `Elements`)?

jackkoenig (commented):
Generally LGTM, I don't love the name `dontTouch.modulePorts` but I also don't have a better idea so 🤷‍♀️

@seldridge @azidar do you have any opinions on the name? Also this API makes it such that it will dontTouch any ports added via BoringUtils even after the function is called (because the call to `DataMirror.fullModulePorts` is delayed

jackkoenig:
> Is this necessary now that Public exists?

I agree with moving away from dontTouch generally but Public does impose additional constraints w.r.t. inferred resets and port widths that dontTouch doesn't.

core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -46,4 +47,18 @@ object dontTouch {
     data
   }
 
+  /** Mark a module so that its ports will not get optimized away.
+    *
+    * @param module whose ports shouldl not get optimized away
```
mwachs5, line 52:
```suggestion
    * @param module whose ports should not get optimized away
```

core/src/main/scala/chisel3/dontTouch.scala
```diff
@@ -46,4 +47,18 @@ object dontTouch {
     data
   }
 
+  /** Mark a module so that its ports will not get optimized away.
```
mwachs5, line 50: does this already do the right thing w.r.t. property ports and probe ports?

seldridge:
> I agree with moving away from dontTouch generally but Public does impose additional constraints w.r.t. inferred resets and port widths that dontTouch doesn't.

Is not enforcing those constraints a requirement?

I'm trying to conserve concepts here and if this API is being used to create testable units, then I think those also probably want public and what it entails.

I won't hold this up as don't touch is a thing and having a utility to don't touch all the ports is generally useful. However, this is an intermediate API that I think we generally want to move off of.

jackkoenig:
I think there are 3 main use cases of this API:

1. Give me this module "as is" in the Verilog -- as you noted this is just `Public`
2. Give me the ports of this module such that I can access them with Verilog XMRs -- this can be handled with Probes and RBProbes
3. I don't understand what is happening and want to debug, leave my ports alone so I can see them in the Verilog while I debug! -- I don't think we really have a great way to do this. Public sort of, but now the additional constraints are a problem when all you're trying to do is debug. Probes, again sort of, but you have to thread them to the top (BoringUtils helps but still extra boilerplate), and then of course read Probes intentionally do not block optimizations and RWProbes also are not really talking about keeping the ports themselves as much as keeping a forceable reference around.

(1) and (2) are what you should do in code that actually hits main, but (3) is still important IMO.

seldridge:
I think that makes sense. Is there anyway that we can prevent users using (3) in situations where they should be using (1) or (2)? Annoyingly, all I could think of is not providing this API. 🙃

core/src/main/scala/chisel3/dontTouch.scala
jackkoenig: As we discussed in dev, we're fine to add this API but we should direct users to `Public` and `Probe` here in the API docs depending on the use case.

src/test/scala/chiselTests/DontTouchSpec.scala
```diff
@@ -123,4 +123,66 @@ class DontTouchSpec extends ChiselFlatSpec with Utils {
     // Ensure can compile the result.
     compile(new HasProbesAndProperties())
   }
+
+  "dontTouch.modulePorts" should "mark all ports of a module" in {
+    class ModuleWithPorts extends Module {
```
mwachs5, line 128: can we add a test for a Module with property and/or Probe ports? Im unclear if those are expected to be annotated, or there is supposed to be an error, or...?

PR #3868, opened 2024-02-24
[ci] Update `jextract` download URL

SpriteOvO:
Some unexpected CI failures occurred in PR #3842, logs https://github.com/chipsalliance/chisel/actions/runs/8029630555

- Updates the `jextract` download URL, the previous one is no longer accessible.

PR #3869, opened 2024-02-26
Fix tests to avoid inferred width ports on public modules.

dtzSiFive:
Public modules can't have inferred widths or abstract reset, fix tests to emit spec-compliant FIRRTL 4.0.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Main module are "public" (#3813, soon mandatory) which means they cannot have inferred widths or abstract resets.  Code relying on old behavior will presently encounter an error in `firtool`.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/ChiselSpec.scala
+++ b/src/test/scala/chiselTests/ChiselSpec.scala
@@ -154,22 +154,23 @@ trait ChiselRunners extends Assertions {
       val testPoint = gen
       assert(!testPoint.isWidthKnown, s"Asserting that width should be inferred yet width is known to Chisel!")
       // Sanity check that firrtl doesn't change the width
-      val out = IO(Output(chiselTypeOf(testPoint)))
+      val widthcheck = Wire(Output(chiselTypeOf(testPoint)))
+      dontTouch(widthcheck)
       val zero = 0.U(0.W).asTypeOf(chiselTypeOf(testPoint))
       if (DataMirror.isWire(testPoint)) {
         testPoint := zero
       }
-      out := zero
-      out := testPoint
+      widthcheck := zero
+      widthcheck := testPoint
     }
     val verilog =
       ChiselStage.emitSystemVerilog(new TestModule, args.toArray :+ "--dump-fir", Array("-disable-all-randomization"))
     expected match {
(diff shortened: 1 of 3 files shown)


dtzSiFive:
CI failure re:jextract is due to the usual problem of the URL expiring, and is fixed here: https://github.com/chipsalliance/chisel/pull/3868 .

src/test/scala/chiselTests/ChiselSpec.scala
```diff
@@ -154,22 +154,23 @@ trait ChiselRunners extends Assertions {
       val testPoint = gen
       assert(!testPoint.isWidthKnown, s"Asserting that width should be inferred yet width is known to Chisel!")
       // Sanity check that firrtl doesn't change the width
-      val out = IO(Output(chiselTypeOf(testPoint)))
+      val widthcheck = Wire(Output(chiselTypeOf(testPoint)))
```
seldridge, line 157: Does this need `Output` here to strip flips?

seldridge (approved):
LGTM

One questions about if this needs output coercion or not.

src/test/scala/chiselTests/ChiselSpec.scala
dtzSiFive: That makes sense, and matches how it worked before.  Will `:=` hook everything up from `testPoint` if the dest is coerced / Output?
(Basically wondering if this presently supports things with flips or not)

Anyway good catch ty I didn't mean to change this, will fix :+1: .

dtzSiFive: Whoops, I didn't change this, you're asking if it needs this ("still")?

Inclined to leave it as-is since it was present before but I'm not sure the story here re:expectations/assumptions/behavior.

Tests pass without it but that doesn't mean it's right to drop but just FYI.

jackkoenig: These tests are assuming passive types so I don't think it matters either way, I think it's fine to keep or remove.

dtzSiFive: Dropped here and from sibling checker.

seldridge (approved):

Issue #3836, opened 2024-02-19
Verilator requires a C++14 or newer compiler


PeterWang-dev:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**
1. Install Verilator 5.021
2. Use chisel-template as a test project
3. Run `mill __.test`

**What is the current behavior?**
Verilator compile error while testing, complaining Verilator need a newer compiler above c++ 14. 

**What is the expected behavior?**
Test should work properly.

**Please tell us about your environment:**

- version: `6.0`
- OS: `Linux ArchLinux 6.7.4-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Mon, 05 Feb 2024 22:07:37 +0000 x86_64 GNU/Linux`
- Verilator version: `Verilator 5.021 devel rev v5.020-105-ga187a16e5` (integrated with `oss-cad-suite` )

**Other Information**

Error logs of test:

``` txt
LANG=C mill __.test

[83/83] TestProject.test.test 
GCDSpec:
- Gcd should calculate proper greatest common denominator *** FAILED ***
  java.lang.Exception: make: Entering directory '/tmp/chisel3.simulator.EphemeralSimulator/357569@PeterWang-ArchLinux/workdir-default'
ls . | grep -v Makefile | grep -v execution-script.txt | xargs rm -rf
\
/opt/oss-cad-suite/bin/verilator \
        '--cc' \
        '--exe' \
        '--build' \
        '-o' \
        '../simulation' \
        '--top-module' \
        'svsimTestbench' \
        '--Mdir' \
        'verilated-sources' \
        '-CFLAGS' \
        '-std=c++11 -I/tmp/chisel3.simulator.EphemeralSimulator/357569@PeterWang-ArchLinux/workdir-default -DSVSIM_ENABLE_VERILATOR_SUPPORT' \
        '../primary-sources/DecoupledGcd.sv' 'testbench.sv' '../generated-sources/c-dpi-bridge.cpp' '../generated-sources/simulation-driver.cpp'
make[1]: Entering directory '/tmp/chisel3.simulator.EphemeralSimulator/357569@PeterWang-ArchLinux/workdir-default/verilated-sources'
[...]

PR #3876, opened 2024-02-27
Use c++14 to pass the verilator check

Emin017:
When I update to verilator v5.022, the svsim test will fail (someone has already raised an [issue](https://github.com/chipsalliance/chisel/issues/3836)). I found out this was because verilator added a [check](https://github.com/verilator/verilator/commit/e3e5f30ba1b02a202aff32e10c95e824e855a6df) for the c++14 standard. 
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Change ChiselSim to requires a C++14 compiler.

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/verilator/Backend.scala
+++ b/svsim/src/main/scala/verilator/Backend.scala
@@ -120,7 +120,7 @@ final class Backend(
                 case OptimizationStyle.OptimizeForCompilationSpeed => Seq("-O1")
               },
 
-              Seq("-std=c++11"),
+              Seq("-std=c++14"),
                 
               additionalHeaderPaths.map { path => s"-I${path}" },
               


seldridge (approved):
Generally seems fine to me. There shouldn't be an issue with requiring a compiler that supports C++14.

We haven't bumped the version of Tabby CAD OSS Suite in a while, so this likely hasn't cropped up. (We're using an older version of Verilator in CI.) We should probably do that, too, put that on latest, or add nightly/CD that tries to keep Tabby CAD up to date.

jackkoenig:
We should merge this, but a longer term fix is we should probably make the arguments version-specific so that we can support some older Linux distros and Verilator versions out-of-the-box.

jackkoenig:
I also wouldn't call this "No release notes", we do now require a compiler with C++14 support which should be most, but lots of users in the hardware industry are on surprisingly old Linux distros.

seldridge:
@jackkoenig: Can we backport this and/or get this onto a non-snapshot release? I think this may be necessary for newer Verilator. I'm thinking this (https://github.com/chipsalliance/chisel-template/issues/127) is related.

jackkoenig:
@seldridge sure, no reason not to release 6.2.0 (or 6.1.1?) with this.

PR #3880, opened 2024-02-27
Add explanation for Class and Object, with examples.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/docs/src/explanations/properties.md
+++ b/docs/src/explanations/properties.md
@@ -169,7 +169,7 @@ created in their body--from the outside, the inputs must be supplied and only
 the outputs may be accessed.
 
 The graphs represented by `Class` declarations and `Object` instantiations
-coexist within the hardware graph instance graph. `Object` instances can exist
+coexist within the hardware instance graph. `Object` instances can exist
 within hardware modules, providing domain-specific information, but hardware
 instances cannot exist within `Class` declarations.
 
--- a/docs/src/explanations/properties.md
+++ b/docs/src/explanations/properties.md
@@ -148,3 +148,249 @@ on integral `Property` typed values.
 | `+`       | Perform addition as defined by FIRRTL spec section Integer Add Operation            |
 | `*`       | Perform multiplication as defined by FIRRTL spec section Integer Multiply Operation |
 | `>>`      | Perform shift right as defined by FIRRTL spec section Integer Shift Right Operation |
+
+### Classes and Objects
+
+Classes and Objects are to `Property` types what modules and instances are to
+hardware types. That is, they provide a means to declare hierarchies through
+which `Property` typed values flow. `Class` declares a hierarchical container
+with input and output `Property` ports, and a body that contains `Property`
+connections and `Object`s. `Object`s instantiate a `Class`, which requires any
+input `Property` ports to be assigned, and allows any output `Property` ports to
+be read.
+
+This allows domain-specific data models to be built using the basic primitives
+of an object-oriented programming language, and embedded directly in the
+instance graph Chisel is constructing. Intuitively, inputs to a `Class` are like
+constructor arguments, which must be supplied to create an `Object`. Similarly,
+outputs from a `Class` are like fields, which may be accessed from an `Object`.
+This separation allows `Class` declarations to abstract over any `Object`s
+created in their body--from the outside, the inputs must be supplied and only
+the outputs may be accessed.
+
+The graphs represented by `Class` declarations and `Object` instantiations
+coexist within the hardware graph instance graph. `Object` instances can exist
+within hardware modules, providing domain-specific information, but hardware
+instances cannot exist within `Class` declarations.
+
+`Object`s can be referenced, and references to `Object`s are a special kind of
+`Property[ClassType]` type. This allows the data model captured by `Class`
+declarations and `Object` instances to form arbitrary graphs.
+
+To understand how `Object` graphs are represented, and can ultimately be
+queried, consider how the hardware instance graph is elaborated. To build the
+`Object` graph, we first pick an entrypoint module to start elaboration. The
+elaboration process works according to the Verilog spec's definition of
+elaboration--instances of modules and `Object`s are instantiated in-memory,
+with connections to their inputs and outputs. Inputs are supplied, and outputs
+may be read. After elaboration completes, the `Object` graph is exposed in terms
+of the output ports, which may contain any `Property` types, including
+references to `Object`s.
+
+To illustrate how these pieces come together, consider the following examples:
+
+```scala mdoc:silent
+import chisel3.properties.Class
+import chisel3.experimental.hierarchy.{instantiable, public, Definition, Instance}
+
+// An abstract description of a CSR, represented as a Class.
+@instantiable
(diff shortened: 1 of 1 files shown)


docs/src/explanations/properties.md
```diff
@@ -148,3 +148,249 @@ on integral `Property` typed values.
...
+
+// An abstract description of a CSR, represented as a Class.
+@instantiable
+class CSRDescription extends Class {
+  // An output Property indicating the CSR name.
+  val identifier = IO(Output(Property[String]()))
+  // An output Property describing the CSR.
+  val description = IO(Output(Property[String]()))
+  // An output Property indicating the CSR width.
+  val width = IO(Output(Property[Int]()))
```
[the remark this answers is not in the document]
mikeurbach: Yeah, maybe they should. Nothing in the example accesses them at Chisel time, but they'd need to be `@public` if so.

docs/src/explanations/properties.md
```diff
@@ -148,3 +148,249 @@ on integral `Property` typed values.
...
 | `*`       | Perform multiplication as defined by FIRRTL spec section Integer Multiply Operation |
 | `>>`      | Perform shift right as defined by FIRRTL spec section Integer Shift Right Operation |
+
+### Classes and Objects
+
+Classes and Objects are to `Property` types what modules and instances are to
+hardware types. That is, they provide a means to declare hierarchies through
+which `Property` typed values flow. `Class` declares a hierarchical container
+with input and output `Property` ports, and a body that contains `Property`
+connections and `Object`s. `Object`s instantiate a `Class`, which requires any
```
mwachs5, line 158:
```suggestion
connections and `Object`s. `Object`s are the instantiation of a `Class`, which requires any
```

maybe?

docs/src/explanations/properties.md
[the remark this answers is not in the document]
mwachs5: is it important to show the @instantiable and @public in this simple example...? It seems to be mixing concepts. The important thing here is that it is an IO(Output(Property...)), right?

 Maybe the @instantiable/@public could be a follow-on example that says this stuff works with D/I?

docs/src/explanations/properties.md
```diff
@@ -148,3 +148,249 @@ on integral `Property` typed values.
...
+are output `Property` type ports on the `Class`.
+
+To capture concrete values at each `Object` instantiation, we have corresponding
+input `Property` type ports, which are connected directly to the outputs. This
+is how we would represent something like a Scala `case class` using `Class`.
+
+```scala mdoc:silent
+// A hardware module representing a CSR and its description.
+class CSRModule(
+  csrDescDef:     Definition[CSRDescription],
```
sequencer, line 233: Wondering why passing `Definition[CSRDescription]` as a parameter `csrDescDef:     Definition[CSRDescription]`?

docs/src/explanations/properties.md
```diff
@@ -148,3 +148,249 @@ on integral `Property` typed values.
...
+  // Instantiate a couple CSR modules.
+  val mcycle = Module(new CSRModule(csrDescDef, 64, "mcycle", "Machine cycle counter."))
+  val minstret = Module(new CSRModule(csrDescDef, 64, "minstret", "Machine instructions-retired counter."))
+
+  // Assign references to the CSR description objects to the property port.
+  descriptions := Property(Seq(mcycle.description.as(csrDescType), minstret.description.as(csrDescType)))
+}
+```
+
+The `Top` module represents the entrypoint. It creates the `Definition` of the
```
sequencer, line 290: The `Definition` usage makes sense to me, but w/o the D/I API, does it also works?

docs/src/explanations/properties.md
```diff
@@ -148,3 +148,249 @@ on integral `Property` typed values.
...
+    connect mcycle.reset, reset
+    inst minstret of minstret
+    connect minstret.clock, clock
+    connect minstret.reset, reset
+    propassign descriptions, List<Inst<CSRDescription>>(mcycle.description, minstret.description)
+```
+
+To understand the `Object` graph that is constructed, we will consider an
+entrypoint to elaboration, and then show a hypothetical JSON representation of
+the `Object` graph. The details of how we go from IR to an `Object` graph are
```
sequencer, line 356: Currently, we don't have the API to get the hypothetical JSON representation. we can demo it in the panamaom framework.

sequencer (commented):
Thanks for the clear documentation. LGTM, I'll propose a end-to-end demonstration recently.

docs/src/explanations/properties.md
mikeurbach: There are a couple annoying reasons. The short answer is that this is the way I'd write it today, so I'm trying to show the happy(est) path on the example.

Slightly longer answer... For one, we want to get the type of a reference to a CSRDescription in the Top module for the `definitions` IO, so it is useful to make the Definition there. Unfortunately, with the way D/I works today, if we also made new Definitions or used Instantiate within the CSRModule class, it will make duplicate definitions. I've spent some time trying to figure this out, but it in general it seems like Definitions are not shared across multiple calls to `Module`. So, the best way to get the FIRRTL we want, with one CSRDescription, is to only call Definition once, and pass the Definition to where it's needed. If you don't mind creating duplicate Definitions, it's not a problem, but I really wanted to show FIRRTL with one Definition of CSRDescription.

docs/src/explanations/properties.md
mikeurbach: Yep. I haven't show an example using the "unsafe" APIs, but it's possible to create classes and objects without D/I. I wanted the first example to show the "safe" APIs with D/I, but I can add a follow up showing the unsafe version.

docs/src/explanations/properties.md
mikeurbach: Yeah that would be great. I will update the wording here slightly, since panamaom isn't an "external tool". Maybe we can add an example in this section of the doc showing a concrete use of panamaom with this IR. It could traverse the descriptions and print something out, so we could show the code snippet and the expected output, or something like that.

docs/src/explanations/properties.md
mikeurbach: I'll add some description about why we recommend Definition/Instance, and just mention DynamicObject, which has scaladoc for the unsafe APIs. We can add an example using the unsafe APIs separately if it's helpful.

docs/src/explanations/properties.md
sequencer: I see, the reason I ask this is I wanna Module is able to serialized and de-serilaized, when add a `Definition` in `Module`, it's really hard to pickle it. I think we can still use the traditional way to instantiate a `Object` w/o passing `CSRDescription` as parameter, I'll tweak it.

docs/src/explanations/properties.md
sequencer: Nice!

sequencer (approved):
Thanks for @mikeurbach's work! Ask @SpriteOvO to copy&paste the examples to scala-cli test, so we can use the binding framework testing it.

PR #3886, opened 2024-02-29
Print the optional Printable passed to stop ops

nandor:
### Contributor Checklist

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

#### Release Notes

The message passed to `stop` is no longer ignored. The construct was extended to accept Printable.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -401,20 +401,34 @@ object cover extends VerifPrintMacrosDoc {
 
 object stop {
 
-  /** Terminate execution, indicating success.
+  /** Terminate execution, indicating success and printing a message.
     *
-    * @param message a string describing why the simulation was stopped
+    * @param message a message describing why simulation was stopped
     */
-  def apply(message: String = "")(implicit sourceInfo: SourceInfo): Stop = {
-    val stp = new Stop()
-    when(!Module.reset.asBool) {
-      pushCommand(Stop(stp, sourceInfo, Builder.forcedClock.ref, 0))
-    }
-    stp
-  }
+  def apply(message: String)(implicit sourceInfo: SourceInfo): Stop = buildStopCommand(Some(PString(message)))
+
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/Stop.scala
```diff
@@ -4,16 +4,25 @@ package chiselTests
...
 class StopTester() extends BasicTester {
-  stop()
+  chisel3.stop()
+}
+
+class StopWithMessageTester() extends BasicTester {
+  val cycle = RegInit(0.U(4.W))
+  cycle := cycle + 1.U
+  when (cycle === 4.U) {
+    chisel3.stop(p"cycle: $cycle")
```
jackkoenig, line 17:
```suggestion
    chisel3.stop(cf"cycle: $cycle")
```
I thought we deprecated `p"..."`, but in any case we should use `cf"..."` in examples

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -401,20 +401,30 @@ object cover extends VerifPrintMacrosDoc {
...
-    }
-    stp
-  }
+  def apply(message: Printable)(implicit sourceInfo: SourceInfo): Stop = buildStopCommand(Some(message))
+
+  /** Terminate execution, indicating success.
+    */
+  def apply()(implicit sourceInfo: SourceInfo): Stop = buildStopCommand(None)
+
+  import VerificationStatement._
```
jackkoenig, line 414: Can we make the import here more precise?

jackkoenig (approved):
LGTM, a couple of minor comments, and don't forget to run `sbt fmt` to run formatting.

jackkoenig (commented):
If you don't mind adding back `def apply(message: String)(implicit sourceInfo: SourceInfo): Stop`, then we can backport this to 6.x and it's also probably better to overload rather than rely on the implicit conversion from String to Printable anyway.

jackkoenig (approved):

PR #3890, opened 2024-03-01
Get ChiselSim working with CIRCT 1.66+

seldridge:
Fix an incorrect assumption in ChiselSim that a 'filelist.f' will exist that includes all modules, _including blackboxes_, in the design.  Change this to try to also copy files from the black box filelist.  This makes ChiselSim work with CIRCT 1.66 and earlier (which put everything in 'filelist.f') and CIRCT 1.66+ (which does not include blackboxes in 'filelist.f').

Note that this is intended as a temporary solution until FIRRTL has a rock solid ABI that is implemented by CIRCT where all modules can be discovered under an arbitrary public module.

See CIRCT changes here: https://github.com/llvm/circt/pull/6729

#### Release Notes

Change ChiselSim to also move blackboxes as indicated by the default black box filelist.  This fixes issues with ChiselSim and CIRCT 1.67.0 introduced in (https://github.com/llvm/circt/pull/6729). This is a backwards compatible change that has no effect on earlier versions of CIRCT.

### DIFF
--- a/src/main/scala/chisel3/simulator/package.scala
+++ b/src/main/scala/chisel3/simulator/package.scala
@@ -107,24 +107,37 @@ package object simulator {
         )
       )
 
-      // Move the relevant files over to primary-sources
-      val filelist =
-        new java.io.BufferedReader(new java.io.FileReader(s"${workspace.supportArtifactsPath}/filelist.f"))
-      try {
-        filelist.lines().forEach { immutableFilename =>
-          var filename = immutableFilename
-          /// Some files are provided as absolute paths
-          if (filename.startsWith(workspace.supportArtifactsPath)) {
-            filename = filename.substring(workspace.supportArtifactsPath.length + 1)
+      // Move the files indicated by a filelist.
+      def moveFiles(filename: String) = {
+        val filelist = new java.io.BufferedReader(new java.io.FileReader(filename))
+        try {
+          filelist.lines().forEach { immutableFilename =>
+            var filename = immutableFilename
+            /// Some files are provided as absolute paths
+            if (filename.startsWith(workspace.supportArtifactsPath)) {
+              filename = filename.substring(workspace.supportArtifactsPath.length + 1)
+            }
+            java.nio.file.Files.move(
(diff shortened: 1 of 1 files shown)


seldridge:
@jackkoenig I'm going to merge this and ask for post-merge review (as this is holding up some other things).

jackkoenig:
@seldridge does this work with older version of firtool? Can/should we backport this to 5.x?

jackkoenig:
RTFM, yes:

> This is a backwards compatible change that has no effect on earlier versions of CIRCT.

src/main/scala/chisel3/simulator/package.scala
```diff
@@ -107,25 +107,38 @@ package object simulator {
...
-      } finally {
-        filelist.close()
       }
 
+      // Move a file in a filelist which may not exist.
+      def maybeMoveFiles(filename: String) = try {
+        moveFiles(filename)
+      } catch {
+        case _ @(_: java.nio.file.NoSuchFileException | _: java.io.FileNotFoundException) =>
+      }
```
jackkoenig, line 135: It would be better to write this as "check if exists, then move if so" but whatever, very Pythonic if you 😉

seldridge:
I tested it with 1.66.0 and assumed that was enough to setup the inductive proof... (Empirically this works on 1.66 as this passed CI and works on the 1.67 candidate as it passed nightly after I manually ran it.)

This may, however, be working because of the aggressive exception hiding in `maybeMoveFiles` now that I think about it. `maybeMoveFiles` is catching `java.nio.file.NoSuchFileException` and `java.io.FileNotFoundException`. What I think is going on is that in 1.67, this will only ever hit one of these if when it fails to find the black box resource filelist. However, in 1.66, the blackbox files may be duplicated in both the black box resource filelist _and the main filelist_. This means that this is working because it is catching both the case of the black box filelist not existing and the case of the files _already_ having been moved.

tl;dr: empirically this should be safe to backport. The exception catching logic may be a little jank.

jackkoenig:
If we need to tweak it on 5.x or 6.x that's fine, thanks for getting the fix in! We can tweak the exception logic later, nbd.

PR #3893, opened 2024-03-01
[chiselsim] Better fix for CIRCT 1.66- and 1.66+

seldridge:
Change the way ChiselSim moves files to track and not move files which may be duplicated in both the default black box filelist and in the main CIRCT-produced filelist.  Previously, this was working because a java.nio exception was being ignored.  However, this may have been hiding bugs where the filelist was actually pointing at a file that did not exist.

### DIFF
--- a/src/main/scala/chisel3/simulator/package.scala
+++ b/src/main/scala/chisel3/simulator/package.scala
@@ -2,6 +2,8 @@ package chisel3
 
 import svsim._
 import chisel3.reflect.DataMirror
+import scala.collection.mutable
+import java.nio.file.{Files, Path, Paths}
 
 package object simulator {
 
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/simulator/package.scala
```diff
@@ -117,21 +120,26 @@ package object simulator {
...
       }
 
-      // Move a file in a filelist which may not exist.
+      // Move a file in a filelist which may not exist.  Do nothing if the
+      // filelist does not exist.
       def maybeMoveFiles(filename: String) = try {
         moveFiles(filename)
       } catch {
-        case _ @(_: java.nio.file.NoSuchFileException | _: java.io.FileNotFoundException) =>
+        case _: java.io.FileNotFoundException =>
```
jackkoenig, line 142: Might be a good opportunity for this to check file existence rather than catching exceptions.

src/main/scala/chisel3/simulator/package.scala
```diff
@@ -117,21 +120,26 @@ package object simulator {
...
-              java.nio.file.Paths.get(s"${workspace.supportArtifactsPath}/$filename"),
-              java.nio.file.Paths.get(s"${workspace.primarySourcesPath}/$filename")
-            )
+            val src = java.nio.file.Paths.get(s"${workspace.supportArtifactsPath}/$filename").normalize()
+            if (!movedFiles.contains(src)) {
+              java.nio.file.Files.move(
+                src,
+                java.nio.file.Paths.get(s"${workspace.primarySourcesPath}/$filename")
+              )
+              movedFiles.addOne(src)
```
jackkoenig, line 129: Nit but typical style would just be `movedFiles += src`

jackkoenig (approved):
A couple of suggestions, feel free to merge when you're ready.

src/main/scala/chisel3/simulator/package.scala
seldridge: Fixed in the force push!

src/main/scala/chisel3/simulator/package.scala
seldridge: Switched to that. This was also starting to push things in the direction of this should be using `java.nio.file.Path` instead of strings. I then rewrote the `moveFiles` method to work this way while, I think, preserving the original logic.

jackkoenig (approved):
LGTM!

PR #3895, opened 2024-03-01
Fix ChiselStage and Builder handling of logging

jackkoenig:
I originally set out just to fix the minor issue of Chisel not doing `Logger.withScope`, that ended up being more involved than I expected.

Long-term, we should probably ditch our Logger code and switch to a more robust solution, probably [scribe](https://github.com/outr/scribe) or scribe + a thin wrapper around it. But this fix is needed for anything using firtool-resolver, otherwise it's really painful for the user to debug what is going wrong with that.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

* Adds support for `--log-level` to `circt.stage.ChiselStage`
* Previously, `object circt.stage.ChiselStage` was ignoring the Logger.
* Also, Chisel was not creating its own logger scope which could lead to clobbering of the Console when running invoking Chisel in the same process multiple times.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/stage/phases/Elaborate.scala
+++ b/src/main/scala/chisel3/stage/phases/Elaborate.scala
@@ -13,14 +13,18 @@ import chisel3.stage.{
   ThrowOnFirstErrorAnnotation
 }
 import firrtl.AnnotationSeq
-import firrtl.options.Phase
+import firrtl.options.{Dependency, Phase}
 import firrtl.options.Viewer.view
+import logger.LoggerOptions
 
 /** Elaborate all [[chisel3.stage.ChiselGeneratorAnnotation]]s into [[chisel3.stage.ChiselCircuitAnnotation]]s.
   */
 class Elaborate extends Phase {
 
-  override def prerequisites = Seq.empty
+  override def prerequisites: Seq[Dependency[Phase]] = Seq(
+    Dependency[chisel3.stage.phases.Checks],
+    Dependency(_root_.logger.phases.Checks)
+  )
   override def optionalPrerequisites = Seq.empty
   override def optionalPrerequisiteOf = Seq.empty
   override def invalidates(a: Phase) = false
@@ -28,6 +32,7 @@ class Elaborate extends Phase {
   def transform(annotations: AnnotationSeq): AnnotationSeq = annotations.flatMap {
     case ChiselGeneratorAnnotation(gen) =>
       val chiselOptions = view[ChiselOptions](annotations)
+      val loggerOptions = view[LoggerOptions](annotations)
       try {
         val context =
           new DynamicContext(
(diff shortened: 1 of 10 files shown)


src/main/scala/circt/stage/Shell.scala
```diff
@@ -0,0 +1,60 @@
...
+  /** Include options for logging
+    *
+    * Defaults to true, override to false when mixing in to [[firrtl.options.Shell]] for use in a [[firrtl.options.Phase]]
+    */
+  def includeLoggerOptions: Boolean = true
+
+  if (includeLoggerOptions) {
+    parser.note("Logger options")
+    Seq(LogLevelAnnotation, ClassLogLevelAnnotation, LogClassNamesAnnotation).foreach(_.addOptions(parser))
+  }
```
seldridge, line 32: This may be better as a `CLI` trait in the logger package that is mixed into the `Shell` as opposed to a method that can be overridden. If going with the method approach, this would be better as `protected`.

seldridge (approved):
Looks good. Consider use of a trait and not a member that is overridden. The tests look great.

src/main/scala/circt/stage/Shell.scala
jackkoenig: It definitely should be protected, will fix. The reason it's a method is because this trait needs to get mixed in to the Shell used by _class_ ChiselStage, but that has to use `firrtl.options.Shell` (required as the type by `Stage`) which already includes the Logger options. So this way makes it _reasonably_ neat and re-usable by class and object ChiselStage.

seldridge:
I think this is causing problems with internal designs as I'm now seeing the unrecognized `LogLevelAnnotation` show up. Generally, this shouldn't be passed to CIRCT unless it's supposed to understand it.

src/main/scala/chisel3/stage/phases/Elaborate.scala
```diff
@@ -13,21 +13,26 @@ import chisel3.stage.{
...
+import logger.LoggerOptions
 
 /** Elaborate all [[chisel3.stage.ChiselGeneratorAnnotation]]s into [[chisel3.stage.ChiselCircuitAnnotation]]s.
   */
 class Elaborate extends Phase {
 
-  override def prerequisites = Seq.empty
+  override def prerequisites: Seq[Dependency[Phase]] = Seq(
+    Dependency[chisel3.stage.phases.Checks],
+    Dependency(_root_.logger.phases.Checks)
```
seldridge, line 26: This brings in a transitive dependency on `_root_logger.phases.AddDefaults` which will cause a default `LogLevelAnnotation` to get pulled in.

I think that `Check` should probably _not_ have `AddDefaults` as a prerequisite and instead it should be an optional one. I'll see about changing this.

seldridge: This is probably not the issue. The real issue is that the logger annotations shouldn't be emitted. I'm testing https://github.com/chipsalliance/chisel/pull/3901 which marks them unserializable and prevents them from showing up in the output CHIRRTL.

PR #3901, opened 2024-03-02
Make logger annotations unserializable

seldridge:
Change logger annotations to mix-in the Unserializable trait so that they will not emitted by a stage.  These annotations are not intended to be seen by CIRCT and these should be stripped from the output FIRRTL text.

#### Release Notes

Change all logger annotations to be unserializable. This avoids a bug where these could be seen by CIRCT (which rightly does not understand them) and cause it to reject the FIRRTL it sees.

### DIFF
--- a/firrtl/src/main/scala/logger/LoggerAnnotations.scala
+++ b/firrtl/src/main/scala/logger/LoggerAnnotations.scala
@@ -3,7 +3,7 @@
 package logger
 
 import firrtl.annotations.{Annotation, NoTargetAnnotation}
-import firrtl.options.{HasShellOptions, ShellOption}
+import firrtl.options.{HasShellOptions, ShellOption, Unserializable}
 
 /** An annotation associated with a Logger command line option */
 sealed trait LoggerOption { this: Annotation => }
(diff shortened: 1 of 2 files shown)


sequencer (approved):

sequencer:
We might need to maintain a list to filter out what circt can parse?

sequencer:
E.g. https://github.com/chipsalliance/chisel/commit/ce0ba40b3d660c1d1b893ac3da6928397dfd607a

seldridge:
> We might need to maintain a list to filter out what circt can parse?

Something like that. The FIRRTL spec is probably the right place for this as annotations are, until removed by converting to other first-class constructs, effectively part of the FIRRTL language.

seldridge:
Will merge and ask for post-commit review from @jackkoenig again on this. 😅 This is getting split Chisel/CIRCT flows back to green. What might be good here, too, is to have a test that is doing a split Chisel/CIRCT flow to make sure that continues to work.

sequencer:
> Something like that. The FIRRTL spec is probably the right place for this as annotations are, until removed by converting to other first-class constructs, effectively part of the FIRRTL language.

Agree. And we should make a plan to purge annotation while making `Property` first class.

PR #3902, opened 2024-03-02
[binder] Specify the new attribute `layers` for `firrtl.instance`

SpriteOvO:
Follow up llvm/circt#6663.


sequencer (approved):
LGTM, thanks for your hard work!

PR #3905, opened 2024-03-05
Support serializing large (> 2 GiB) annotation files

jackkoenig:
This probably also improves performance and reduces memory since creating a large String is expensive.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
+++ b/firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
@@ -15,6 +15,7 @@ import org.json4s.native.Serialization
 import org.json4s.native.Serialization.{read, write, writePretty}
 
 import scala.collection.mutable
+import java.io.{StringWriter, Writer}
 
 trait HasSerializationHints {
   // For serialization of complicated constructor arguments, let the annotation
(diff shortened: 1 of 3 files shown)


firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -167,13 +168,21 @@ object JsonProtocol extends LazyLogging {
...
-    val tags = getTags(annos)
+  def serializeTry(annos: Seq[Annotation]): Try[String] = serializeTry(annos, new StringWriter).map(_.toString)
+
+  /** Serialize annotations to a [[java.io.Writer]]
+    *
+    * @param annos Annotations to serialize
+    * @param out Writer to which the serialized annotations will be written
+    * @return
+    */
+  def serializeTry[W <: Writer](annos: Iterable[Annotation], out: W): Try[W] = {
```
seldridge, line 179: An alternative is to make this `private`. There is then nothing to actually test here (other than seeing if 2GiB of annotations work... 😓). I'm really questioning if this is necessary to make public.

seldridge (approved):
LGTM

firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: I assumed I'd have to call it from Chisel but ended up not needing to. That being said, the current test is in package `firrtlTests` so would have to mess with that. We have already deprecated any import of `package firrtl` (via the compiler plugin) so I'm a bit 🤷‍♀️ about it.

PR #3908, opened 2024-03-06
Allow BoringUtils.bore to work on probes

seldridge:
Extends BorginUtils.bore (and related) to work where the source is a probe.  Currently, if a user tries to do this, they will get an error about being unable to probe a probe because the old code would unconditionally wrap whatever the source was in `Probe`/`RWProbe`.  Change this to not wrap if the source is already a probe.

#### Release Notes

Allow boring from a source which is already a probe.

### DIFF
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -224,9 +224,14 @@ object BoringUtils {
     else if (DataMirror.hasOuterFlip(source)) Flipped(chiselTypeOf(source))
     else chiselTypeOf(source)
     def purePortType = createProbe match {
-      case Some(pi) if pi.writable => RWProbe(purePortTypeBase)
-      case Some(pi)                => Probe(purePortTypeBase)
-      case None                    => purePortTypeBase
+      case Some(pi) =>
+        // If the source is already a probe, don't double wrap it in a probe.
+        purePortTypeBase.probeInfo match {
+          case Some(_)             => purePortTypeBase
+          case None if pi.writable => RWProbe(purePortTypeBase)
+          case None                => Probe(purePortTypeBase)
+        }
+      case None => purePortTypeBase
     }
     def isPort(d: Data): Boolean = d.topBindingOpt match {
       case Some(PortBinding(_)) => true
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -224,9 +224,14 @@ object BoringUtils {
     else if (DataMirror.hasOuterFlip(source)) Flipped(chiselTypeOf(source))
     else chiselTypeOf(source)
     def purePortType = createProbe match {
-      case Some(pi) if pi.writable => RWProbe(purePortTypeBase)
-      case Some(pi)                => Probe(purePortTypeBase)
-      case None                    => purePortTypeBase
+      case Some(pi) =>
+        // If the source is already a probe, don't double wrap it in a probe.
+        purePortTypeBase.probeInfo match {
```
dtzSiFive, line 229: Things get strange if the bore/tap target is mixed probe/non-probe, but that's already true.

(or if has incompatible probe type vs requested kind)

Not sure what the errors look like if this is wrong.

dtzSiFive: Not something to be addressed here unless it seems easy/reasonable to address while visiting (?), just a note.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -224,9 +224,14 @@ object BoringUtils {
...
+          case Some(_)             => purePortTypeBase
+          case None if pi.writable => RWProbe(purePortTypeBase)
+          case None                => Probe(purePortTypeBase)
+        }
+      case None => purePortTypeBase
```
dtzSiFive, line 234: Some of the below logic maybe can/should be tweaked, not sure and maybe better to defer that to a later simplification.
For example, if the target is a rwprobe port (new L251, old L246), and this is used by rwTap, it'll add another RWProbe for that.  Possibly error but haven't traced all the paths here...

dtzSiFive (approved):
Looks like a useful extension, although I'm a little uncertain about this method's behavior in all the various cases.

Don't see this breaking things that work today AFAIK, so LGTM.

PR #3921, opened 2024-03-08
toBigInt should round identically for Double and BigDecimal

mvanbeirendonck:
This very minor PR fixes a bug in `Num.toBigInt(x: BigDecimal, binaryPoint: Int)`. 

The current function gives the impression to round the input BigDecimal:

```scala
val result = (x * multiplier).rounded.toBigInt
```

However:

```scala
BigDecimal(1.4).rounded.toBigInt // : BigInt = 1
BigDecimal(1.5).rounded.toBigInt // : BigInt = 1
BigDecimal(1.6).rounded.toBigInt // : BigInt = 1
```

I'm assuming this is not the intended behaviour. 

The underlying issue is with the `rounded` function which uses the BigDecimal's `MathContext`. I changed this to use `setScale(0)` instead.

I added a testcase for rounding positive numbers. I would've added a testcase for negative ones, but as far as I know there is no `RoundingMode` that replicates the rounding of Doubles there (round half to positive infinity).

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Num.scala
+++ b/core/src/main/scala/chisel3/Num.scala
@@ -200,13 +200,22 @@ trait NumObject {
     result
   }
 
-  /**
-    * How to create a bigint from a big decimal with a specific binaryPoint (int)
-    * @param x           a BigDecimal value
-    * @param binaryPoint a binaryPoint that you would like to use
+  /** Create a bigint from a big decimal with a specific binaryPoint (Int)
+    *
+    * @param x           the BigDecimal value
+    * @param binaryPoint the binaryPoint to use
     * @return
     */
-  def toBigInt(x: BigDecimal, binaryPoint: Int, roundingMode: RoundingMode = HALF_UP): BigInt = {
+  def toBigInt(x: BigDecimal, binaryPoint: Int): BigInt = toBigInt(x, binaryPoint, HALF_UP)
+
+  /** Create a bigint from a big decimal with a specific binaryPoint (Int)
+    *
+    * @param x           the BigDecimal value
+    * @param binaryPoint the binaryPoint to use
+    * @param roundingMode the RoundingMode to use
+    * @return
+    */
+  def toBigInt(x: BigDecimal, binaryPoint: Int, roundingMode: RoundingMode): BigInt = {
     val multiplier = math.pow(2, binaryPoint)
     val result = (x * multiplier).setScale(0, roundingMode).toBigInt
     result
--- a/core/src/main/scala/chisel3/Num.scala
+++ b/core/src/main/scala/chisel3/Num.scala
@@ -5,6 +5,7 @@ package chisel3
 import chisel3.internal.sourceinfo.SourceInfoTransform
 
 import scala.language.experimental.macros
+import scala.math.BigDecimal.RoundingMode._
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/LiteralExtractorSpec.scala
```diff
@@ -59,6 +59,20 @@ class LiteralExtractorSpec extends ChiselFlatSpec {
     bigIntFromDouble should be(bigInt53)
   }
 
+  "doubles and big decimals" should "be rounded identically" in {
+
+    for (double <- Seq(1.0, 1.1, 1.5, 1.6, 2.5)) {
```
seldridge, line 64: This has both `1.5` and `2.5` which means that this is checking to make sure it's not using `HALF_EVEN`. 👍

core/src/main/scala/chisel3/Num.scala
```diff
@@ -5,6 +5,7 @@ package chisel3
 import chisel3.internal.sourceinfo.SourceInfoTransform
 
 import scala.language.experimental.macros
+import scala.math.BigDecimal.RoundingMode._
```
seldridge, line 8: Nit: I'd avoid the wildcard import here and just do:

```suggestion
import scala.math.BigDecimal.RoundingMode.RoundingMode
```

seldridge (approved):
Minor comments, but LGTM otherwise. Nice test.

core/src/main/scala/chisel3/Num.scala
```diff
@@ -205,9 +206,9 @@ trait NumObject {
     * @param binaryPoint a binaryPoint that you would like to use
     * @return
     */
-  def toBigInt(x: BigDecimal, binaryPoint: Int): BigInt = {
+  def toBigInt(x: BigDecimal, binaryPoint: Int, roundingMode: RoundingMode = HALF_UP): BigInt = {
```
jackkoenig, line 209: Since this is a bugfix, I'd like to backport it. Stable branches check binary compatibility and obnoxiously in Scala (2 at least, not sure about 3) it is not binary compatible to add an argument with a default parameter--instead you have to just overload the method, ie.:
```scala
def toBigInt(x: BigDecimal): BigInt = toBigInt(x, HALF_UP)
def toBigInt(x: BigDecimal, binaryPoint: Int, roundingMode: RoundingMode): BigInt = ...
```
I will push this change to your branch, I just wanted to note it before we merge.

jackkoenig: I just realized the version without the default argument already exists, I'm surprised this isn't causing a problem...

jackkoenig:
@mvanbeirendonck would you like a release with this fix in the near term? What version of Chisel are you using?

mvanbeirendonck:
> @mvanbeirendonck would you like a release with this fix in the near term? What version of Chisel are you using?

Not very urgent! Just wanted to report this one. I am still on Chisel 3.5.6.

jackkoenig:
@mvanbeirendonck fix released in Chisel v6.3.0 and v5.2.0, hopefully you can get there from Chisel 3.5 😅. This will also be in v3.6.1 when I eventually get to it.

PR #3940, opened 2024-03-16
Some fixes related to OM

SpriteOvO:
- Add method for reading mlirbc in `PanamaCIRCTConverter`

- Construct `MlirStringRef` manually for `Array[Byte]`

  `Array[Byte]` may not end with `'\0'`, so constructing a `MlirStringRef` with `Array[Byte]` via `mlirStringRefCreateFromCString` can sometimes cause a crash in CIRCT.

- Fix infinite recursion for displaying `PanamaCIRCTOMEvaluatorValuePath`

### DIFF
--- a/panamaconverter/src/PanamaCIRCTConverter.scala
+++ b/panamaconverter/src/PanamaCIRCTConverter.scala
@@ -1665,6 +1665,12 @@ object PanamaCIRCTConverter {
     cvt.mlirRootModule = circt.mlirModuleCreateParse(mlir)
     cvt
   }
+  def newWithMlirBc(mlirbc: Array[Byte]): PanamaCIRCTConverter = {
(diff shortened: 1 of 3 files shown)


sequencer:
I still don't think Converter should be the Top module, but since OM is not released to users yet, the bug fix LGTM.
The next step is decoupling Converter and OM, 
Converter should only depend on PanamaCIRCT, OM should only depend on PanamaCIRCT, don't couple these two packages together.

sequencer (approved):

SpriteOvO:
> I still don't think Converter should be the Top module ... The next step is decoupling Converter and OM, Converter should only depend on PanamaCIRCT, OM should only depend on PanamaCIRCT, don't couple these two packages together.

Agree, I'm going to refactor them next.

PR #3943, opened 2024-03-18
Fix incorrect width for negative `SInt` literals

### LINKED ISSUE #3942: `PanamaCIRCTConverter` emits incorrect firrtl for negative `SInt` literals.
**Type of issue**: Bug Report

When developing integration tests for it (#3936), I noticed the following case will fail:

```scala
class WireAndReg extends Module {
  val r = IO(Input(Bool()))
  val o = IO(Output(UInt(2.W)))

  val o_next = RegInit(false.B)
  val flip = Reg(Bool())
  val magic = Wire(SInt(8.W))

  o := o_next && magic(7)
  o_next := flip
  flip := flip ^ r
  magic := -42.S
}

println(lit.utility.panamaconverter.firrtlString(new WireAndReg))
```

For Wire `magic`, PanamaConverter will emit `connect magic, pad(SInt<7>(22), 8)`. For some reason, the sign bit is truncated, and in Verilog generation the whole circuit will be optimized out.

For reference, `circt.stage.ChiselStage.emitCHIRRTL` emits `connect magic, asSInt(UInt<7>(0h56))`.

`nix develop` shell on HEAD as of 31b222e4445b1d6c58321e1d4d1752b345e02326.

Give correct negative `SInt`s.

SpriteOvO:
Fixes #3942.

The reason of this bug is that the result of Scala's `BigInt.bitLength` excludes the sign bit.

Reference from Chisel

https://github.com/chipsalliance/chisel/blob/7b9c07c4e5fe4a8174b4f788cf798b3870039e7e/core/src/main/scala/chisel3/internal/firrtl/Converter.scala#L88-L92

I haven't figured out why Chisel always generates `asSInt(UInt())` for `SLit`. However, since `pad` is sign-extends this behavior is documented. I think the fixed version `connect magic, pad(SInt<7>(-42), 8)` and `connect magic, asSInt(UInt<7>(0h56))` should be equivalent?

Feel free to correct me if I'm wrong.

```
## Padding Operations

| Name | Arguments | Parameters | Arg Types | Result Type | Result Width |
|------|-----------|------------|-----------|-------------|--------------|
| pad  | \(e\)     | \(n\)      | (UInt)    | UInt        | max(w~e~,n)  |
|      |           |            | (SInt)    | SInt        | max(w~e~,n)  |

If e's bit width is smaller than n, then the pad operation zero-extends or sign-extends e up to the given width n.
Otherwise, the result is simply e.
n must be non-negative.
```

```diff
--- a/chisel.fir
+++ b/panama.fir
@@ -6,13 +6,13 @@ circuit WireAndReg :
     input r : UInt<1>
     output o : UInt<2>
 
-    regreset o_next : UInt<1>, clock, reset, UInt<1>(0h0)
+    regreset o_next : UInt<1>, clock, reset, UInt<1>(0)
     reg flip : UInt<1>, clock
     wire magic : SInt<8>
     node _o_T = bits(magic, 7, 7)
     node _o_T_1 = and(o_next, _o_T)
-    connect o, _o_T_1
+    connect o, pad(_o_T_1, 2)
     connect o_next, flip
     node _flip_T = xor(flip, r)
     connect flip, _flip_T
-    connect magic, asSInt(UInt<7>(0h56))
+    connect magic, pad(SInt<7>(-42), 8)
```

---

CC @OceanS2000 Could you review this PR? Thanks!

### DIFF
--- a/panamaconverter/src/PanamaCIRCTConverter.scala
+++ b/panamaconverter/src/PanamaCIRCTConverter.scala
@@ -197,7 +197,7 @@ class PanamaCIRCTConverter(val circt: PanamaCIRCT, fos: Option[FirtoolOptions], 
     def convert(name: String, parameter: Param): MlirAttribute = {
       val (tpe, value) = parameter match {
         case IntParam(value) =>
-          val tpe = circt.mlirIntegerTypeGet(max(bitLength(value), 32))
+          val tpe = circt.mlirIntegerTypeGet(max(value.bitLength, 32))
           (tpe, circt.mlirIntegerAttrGet(tpe, value.toLong))
         case DoubleParam(value) =>
           val tpe = circt.mlirF64TypeGet()
@@ -239,8 +239,6 @@ class PanamaCIRCTConverter(val circt: PanamaCIRCT, fos: Option[FirtoolOptions], 
         .withNamedAttr("portSyms", circt.mlirArrayAttrGet(ports.symAttrs))
         .withNamedAttr("portLocations", circt.mlirArrayAttrGet(ports.locAttrs))
     }
-
-    def bitLength(n: BigInt): Int = max(n.bitLength, 1)
 
     def widthShl(lhs: fir.Width, rhs: fir.Width): fir.Width = (lhs, rhs) match {
       case (l: fir.IntWidth, r: fir.IntWidth) => fir.IntWidth(l.width << r.width.toInt)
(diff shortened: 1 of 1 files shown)


panamaconverter/src/PanamaCIRCTConverter.scala
```diff
@@ -486,7 +484,7 @@ class PanamaCIRCTConverter(val circt: PanamaCIRCT, fos: Option[FirtoolOptions],
       def referToNewConstant(n: BigInt, w: Width, isSigned: Boolean): Reference.Value = {
         val (firWidth, valWidth) = w match {
           case _: UnknownWidth =>
-            val bitLen = util.bitLength(n)
+            val bitLen = if (!isSigned) max(n.bitLength, 1) else n.bitLength + 1
```
sequencer, line 487: plz add the documentation to this for explaining the reason.

sequencer (commented):
just need a few comment as reminder why we are doing this.(like the explanation on your comment.

panamaconverter/src/PanamaCIRCTConverter.scala
SpriteOvO: Done.

sequencer (commented):
Request review from @OceanS2000 to see if this PR fix #3942.

OceanS2000 (approved):
It seems to me that #3942 is fixed with current changes.

sequencer:
@SpriteOvO please rebase

SpriteOvO:
@sequencer Rebased.

PR #3955, opened 2024-03-28
Make SRAMs directly emit FIRRTL memories

debs-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- Backend code generation

#### Desired Merge Strategy

#### Release Notes

`SRAM`s updated to directly emit FIRRTL memories (as opposed to creating `SyncReadMem`s.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -160,7 +160,9 @@ abstract class RawModule extends BaseModule {
           // probes have their refs set eagerly
           case _ => // don't name literals
         }
-      } // else, don't name unbound types
+      }
+    case m: SramTarget =>
+      id.forceName(default = "_MEM", _namespace)
   }
 
   private[chisel3] override def generateComponent(): Option[Component] = {
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -162,7 +162,7 @@ abstract class RawModule extends BaseModule {
         }
       }
     case m: SramTarget =>
-      id.forceName(default = "_MEM", _namespace)
+      id.forceName(default = "MEM", _namespace)
   }
 
   private[chisel3] override def generateComponent(): Option[Component] = {
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -219,7 +219,7 @@ private[chisel3] object ir {
       // NOTE: mod eq ctx.id only occurs in Target and Named-related APIs
       if (mod eq ctx.id) localName else name
   }
-  case class Slot(imm: Node, name: String) extends Arg {
+  case class Slot(imm: Arg, name: String) extends Arg {
     override def contextualName(ctx: Component): String = {
       val immName = imm.contextualName(ctx)
       if (immName.isEmpty) name else s"$immName.$name"
@@ -288,6 +288,16 @@ private[chisel3] object ir {
     readUnderWrite: fir.ReadUnderWrite.Value)
       extends Definition
 
+  case class FirrtlMemory(
+    sourceInfo:         SourceInfo,
+    id:                 HasId,
+    t:                  Data,
+    size:               BigInt,
+    readPortNames:      Seq[String],
+    writePortNames:     Seq[String],
+    readwritePortNames: Seq[String])
+      extends Definition
+
   case class DefMemPort[T <: Data](
     sourceInfo: SourceInfo,
     id:         T,
@@ -310,7 +320,7 @@ private[chisel3] object ir {
   case class WhenEnd(sourceInfo: SourceInfo, firrtlDepth: Int, hasAlt: Boolean = false) extends Command
   case class AltBegin(sourceInfo: SourceInfo) extends Command
   case class OtherwiseEnd(sourceInfo: SourceInfo, firrtlDepth: Int) extends Command
-  case class Connect(sourceInfo: SourceInfo, loc: Node, exp: Arg) extends Command
+  case class Connect(sourceInfo: SourceInfo, loc: Arg, exp: Arg) extends Command
   case class PropAssign(sourceInfo: SourceInfo, loc: Node, exp: Arg) extends Command
   case class Attach(sourceInfo: SourceInfo, locs: Seq[Node]) extends Command
   case class ConnectInit(sourceInfo: SourceInfo, loc: Node, exp: Arg) extends Command
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -524,22 +524,25 @@ object SRAM {
     )
 
     // connect firrtl memory ports to user-facing interface
-    for ((memReadPort, firrtlReadPort) <- _out.readPorts.zip(firrtlReadPorts)) {
+    for (((memReadPort, firrtlReadPort), readClock) <- _out.readPorts.zip(firrtlReadPorts).zip(readPortClocks)) {
       firrtlReadPort.addr := memReadPort.address
-      firrtlReadPort.clk := Module.clock
+      firrtlReadPort.clk := readClock
       memReadPort.data := firrtlReadPort.data.asInstanceOf[Data]
       firrtlReadPort.en := memReadPort.enable
     }
-    for ((memWritePort, firrtlWritePort) <- _out.writePorts.zip(firrtlWritePorts)) {
+    for (((memWritePort, firrtlWritePort), writeClock) <- _out.writePorts.zip(firrtlWritePorts).zip(writePortClocks)) {
       firrtlWritePort.addr := memWritePort.address
-      firrtlWritePort.clk := Module.clock
+      firrtlWritePort.clk := writeClock
       firrtlWritePort.data.asInstanceOf[Data] := memWritePort.data
       firrtlWritePort.en := memWritePort.enable
       assignMask(memWritePort.data, memWritePort.mask, firrtlWritePort.getRef, "mask")
     }
-    for ((memReadwritePort, firrtlReadwritePort) <- _out.readwritePorts.zip(firrtlReadwritePorts)) {
+    for (
+      ((memReadwritePort, firrtlReadwritePort), readwriteClock) <-
+        _out.readwritePorts.zip(firrtlReadwritePorts).zip(readwritePortClocks)
+    ) {
       firrtlReadwritePort.addr := memReadwritePort.address
-      firrtlReadwritePort.clk := Module.clock
+      firrtlReadwritePort.clk := readwriteClock
       firrtlReadwritePort.en := memReadwritePort.enable
       memReadwritePort.readData := firrtlReadwritePort.rdata.asInstanceOf[Data]
       firrtlReadwritePort.wdata.asInstanceOf[Data] := memReadwritePort.writeData
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -3,6 +3,7 @@ package chisel3.util
 import chisel3._
 
 import chisel3.internal.{Builder, FirrtlMemTypeBinding, NamedComponent, SramPortBinding}
+import chisel3.internal.plugin.autoNameRecursively
 import chisel3.experimental.SourceInfo
 import chisel3.internal.sourceinfo.{MemTransform, SourceInfoTransform}
 import chisel3.internal.firrtl.ir.{Arg, FirrtlMemory, ILit, Index, Node, Ref, Slot}
@@ -473,7 +474,7 @@ object SRAM {
     }
 
     // underlying target
-    val mem = new SramTarget()
+    val mem = autoNameRecursively("sram")(new SramTarget)
 
     // user-facing interface into the SRAM
     val sramIntfType = new SRAMInterface(size, tpe, numReadPorts, numWritePorts, numReadwritePorts, isVecMem)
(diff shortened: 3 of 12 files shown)


src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -530,3 +618,40 @@ object SRAM {
     s"$rdPorts$wrPorts$rwPorts"
   }
 }
+
+/** Contains fields that map to a FIRRTL memory read port.
+  *
+  * @param readPort used to parameterize this class
+  */
+private[chisel3] final class FirrtlMemoryReader[T <: Data](readPort: MemoryReadPort[T]) extends Bundle {
```
debs-sifive, line 626: @jackkoenig should these bundles live here discreetly? The file is getting kind of long, so I was thinking about separating these out into a different file.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -288,6 +288,16 @@ private[chisel3] object ir {
     readUnderWrite: fir.ReadUnderWrite.Value)
       extends Definition
 
+  case class FirrtlMemory(
```
debs-sifive, line 291: Thoughts on naming this `DefMem` and renaming `DefMem` to `CDefMem` for name alignment? Or not that important?

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -7,7 +7,9 @@ import chisel3.util.SRAM
 import chisel3.experimental.{annotate, ChiselAnnotation}
 import chiselTests.ChiselFlatSpec
 import _root_.circt.stage.ChiselStage.emitCHIRRTL
+import _root_.circt.stage.ChiselStage.emitSystemVerilog
```
jackkoenig, line 10: You can combine these two imports

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -7,7 +7,9 @@ import chisel3.util.SRAM
...
 import firrtl.annotations.{Annotation, ReferenceTarget, SingleTargetAnnotation}
+import chisel3.util.MemoryReadWritePort
```
jackkoenig, line 12: You can merge this with the `import chisel3.util.SRAM` above

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -33,13 +35,25 @@ class SRAMSpec extends ChiselFlatSpec {
...
+    chirrtl should include("read-latency => 1")
+    chirrtl should include("write-latency => 1")
+    chirrtl should include("readwriter => RW0")
+    chirrtl should include("read-under-write => undefined")
+    chirrtl should include("connect _sram_MEM.RW0.addr, sram.readwritePorts[0].address")
+    chirrtl should include("connect _sram_MEM.RW0.clk, clock")
+    chirrtl should include("connect _sram_MEM.RW0.en, sram.readwritePorts[0].enable")
+    chirrtl should include("connect sram.readwritePorts[0].readData, _sram_MEM.RW0.rdata")
+    chirrtl should include("connect _sram_MEM.RW0.wdata, sram.readwritePorts[0].writeData")
+    chirrtl should include("connect _sram_MEM.RW0.wmode, _sram_MEM.RW0.wmode")
```
seldridge, line 53: This make me so happy...

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -33,13 +35,25 @@ class SRAMSpec extends ChiselFlatSpec {
     val (chirrtlCircuit, annos) = getFirrtlAndAnnos(new Top)
     val chirrtl = chirrtlCircuit.serialize
     chirrtl should include("module Top :")
-    chirrtl should include("smem sram_mem : UInt<8> [32]")
     chirrtl should include(
       "wire sram : { readPorts : { flip address : UInt<5>, flip enable : UInt<1>, data : UInt<8>}[0], writePorts : { flip address : UInt<5>, flip enable : UInt<1>, flip data : UInt<8>}[0], readwritePorts : { flip address : UInt<5>, flip enable : UInt<1>, flip isWrite : UInt<1>, readData : UInt<8>, flip writeData : UInt<8>}[1]}"
     )
+    chirrtl should include("mem _sram_MEM")
```
jackkoenig, line 41: The default name can't be `_sram_MEM`. The reason this changed is that `val mem` in `SRAM.scala` is of type `SyncReadMem` which gets handled by the compiler plugin. You can do the naming yourself by wrapping the RHS of `val sram = ` with `autoNameRecursively` [[1]](https://github.com/chipsalliance/chisel/blob/9177535ff1be47ffd99034bf0154c0f1ec637419/core/src/main/scala/chisel3/internal/plugin/package.scala#L30) (make sure to leave a comment explaining why you're doing this).

src/main/scala/chisel3/util/SRAM.scala
jackkoenig: I think they're fine here, but leave a comment about why they're private and mention that the write and readwrite bundles are missing `mask` (and explain why).

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -470,54 +472,140 @@ object SRAM {
...
+        firrtlReadPort.en := memReadPort.enable
     }
+    _out.writePorts.zip(firrtlWritePorts).foreach {
+      case (memWritePort, firrtlWritePort) =>
+        firrtlWritePort.addr := memWritePort.address
+        firrtlWritePort.clk := Module.clock
+        firrtlWritePort.data.asInstanceOf[Data] := memWritePort.data
+        firrtlWritePort.en := memWritePort.enable
+        assignMask(memWritePort.data, memWritePort.mask, firrtlWritePort.getRef, "mask")
+    }
```
jackkoenig, line 546:
```suggestion
    for((memWritePort, firrtlWritePort) <- _out.writePorts.zip(firrtlWritePorts)) {
      firrtlWritePort.addr := memWritePort.address
      firrtlWritePort.clk := Module.clock
      firrtlWritePort.data.asInstanceOf[Data] := memWritePort.data
      firrtlWritePort.en := memWritePort.enable
      assignMask(memWritePort.data, memWritePort.mask, firrtlWritePort.getRef, "mask")
    }
```
This is a not a strongly held position, but for simple foreaches like this, I prefer `for` style since it reduces a level of indention

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -470,54 +472,140 @@ object SRAM {
...
+        val name = s"W$idx"
+        p.setRef(Slot(Node(mem), name))
+        name
+    }
+    val firrtlReadwritePortNames: Seq[String] = firrtlReadwritePorts.zipWithIndex.map {
+      case (p, idx) =>
+        val name = s"RW$idx"
+        p.setRef(Slot(Node(mem), name))
+        name
+    }
```
jackkoenig, line 507:
```suggestion
    def nameAndSetRef(ports: Seq[Data], namePrefix: String): Seq[Name] = {
      ports.zipWithIndex.map { case (p, idx) =>
        val name = namePrefix + idx
        p.setRef(Slot(Node(mem), name))
        name        
      }
    }
    val firrtlReadPortNames: Seq[String] = nameAndSetRef(firrtlReadPorts, "R")
    val firrtlWritePortNames: Seq[String] = nameAndSetRef(firrtlWritePorts, "W")
    val firrtlReadwritePortNames: Seq[String] = nameAndSetRef(firrtlReadwritePorts, "RW")
```
Reduce, reuse, recycle

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -70,4 +84,119 @@ class SRAMSpec extends ChiselFlatSpec {
...
+            val x = UInt(3.W)
+            val y = Vec(4, Bool())
+          }
+        ),
+        numReadPorts = 1,
+        numWritePorts = 1,
+        numReadwritePorts = 0
+      )
+      val maskedVecRecordMemIo = IO(maskedVecRecordMem.cloneType)
+      maskedVecRecordMemIo <> maskedVecRecordMem
```
jackkoenig, line 142: Good job writing good tests 🙂

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: I don't think it's important, fine as is IMO

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -160,7 +160,9 @@ abstract class RawModule extends BaseModule {
           // probes have their refs set eagerly
           case _ => // don't name literals
         }
-      } // else, don't name unbound types
+      }
+    case m: SramTarget =>
+      id.forceName(default = "_MEM", _namespace)
```
jackkoenig, line 165:
```suggestion
      id.forceName(default = "MEM", _namespace)
```
This is not related to my comment below because if you do that, we should never actually hit this default, but we should just default to `MEM` (like `MemBase` above)

jackkoenig (approved):
LGTM, mostly style nits, please fix the names of the memories before merging.

src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -70,4 +84,119 @@ class SRAMSpec extends ChiselFlatSpec {
...
+        tpe = new Bundle {
+          val x = UInt(3.W)
+          val y = Vec(4, Bool())
+        },
+        numReadPorts = 0,
+        numWritePorts = 0,
+        numReadwritePorts = 1
+      )
+      val unmaskedRecordMemIo = IO(unmaskedRecordMem.cloneType)
+      unmaskedRecordMemIo <> unmaskedRecordMem
```
mwachs5, line 189: shoudl these tests use `:<>=` instead of `<>`?

src/main/scala/chisel3/util/SRAM.scala
debs-sifive: oh cool, that does look nice! I have never seen or used this pattern before :x

jackkoenig (approved):

seldridge (approved):

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -219,7 +219,7 @@ private[chisel3] object ir {
       // NOTE: mod eq ctx.id only occurs in Target and Named-related APIs
       if (mod eq ctx.id) localName else name
   }
-  case class Slot(imm: Node, name: String) extends Arg {
+  case class Slot(imm: Arg, name: String) extends Arg {
```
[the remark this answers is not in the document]
jackkoenig: Good eye, so recall (or learn very quickly) that `Arg` is the Chisel IR Expression. `Node` is the `Arg` that wraps a `HasId`. In SRAM, we manually build some nested Args to emit the connections to the mask fields of the ports because representing those as actual Chisel types is really hard. We needed to build some Slots that didn't have real Chisel Data backing them up (thus no `HasIds`). We figured it was pretty weird that `Slot` _required_ a `HasId` (whereas other `Args` like `Index` don't), so this change makes more sense anyway and let us do what we needed to do to emit the connections for the mask.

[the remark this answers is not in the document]
seldridge: Follow up question: https://github.com/chipsalliance/chisel/pull/3955/files#diff-f01d2f59c3998b1bb4f17da0c509aa21205a6682b1cd42e274dca0fa92ebf33dR477 is preventing this alternative id-less `Slot` from showing up when doing a select query. This seems to make sense with your explanation.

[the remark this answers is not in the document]
jackkoenig: Yeah that is the only place where this change had any effect. @debs-sifive maybe we should write a test on what happens if you do a Select query for connections on a Module with an SRAM. These connections to the real ports of the SRAM are iffy and that Select may let users get at them...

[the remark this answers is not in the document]
debs-sifive: I played around with this a bit, and we can indeed see some of the memory ports, but it's all output stuff:
```
List(PredicatedConnect(List(),MemWrapper.foo.readPorts[0].data: Wire[UInt<8>],MemWrapper.foo_sram.R0.data: [UInt<8>],false), PredicatedConnect(List(),MemWrapper.foo.readwritePorts[0].readData: Wire[UInt<8>],MemWrapper.foo_sram.RW0.rdata: [UInt<8>],false))
```

So it looks like we can't drive anything on the actual memory with Select. If we want Select functionality with SRAM, I say we do it in a separate PR :)

sequencer:
Really appreciate. And we finally have this feature! I wanna kill all `Mem` usage for a long time. After this feature leading to chisel, maybe we can finally deprecate `Mem` API. 
Some questions but I don’t think it will block this PR:

- does it support `suggestName`?(some time sram isn’t sram, but RF). Originally it doesn’t because it warp `cmem`. 

- maybe in the future. It can be directly lowered to SRAM Intrinsic? Then we can simply add more interface to sram. For example extend MBIST interface.

- is possible to emit sram with Instance Select API for avoiding different lower pass for memory behavior models?

debs-sifive:
@sequencer I haven't really worked with the Instance Select API, so not sure about that Q.

> does it support suggestName?(some time sram isn’t sram, but RF). Originally it doesn’t because it warp cmem.

It currently supports `suggestName` of the `SRAMInterface`. I believe we can support `suggestName` of the memory itself, since the memory target is a `HasID`. But it's not implemented in this iteration.

PR #3960, opened 2024-04-01
Add a new BoringUtils.drive API for boring to drive a sink.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

This API allows users to bore to a sink they plan to drive, which complements the existing API to bore from a source to read.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -211,10 +211,11 @@ abstract class RawModule extends BaseModule {
       case (true, false) if left.probeInfo.get.writable => ProbeDefine(si, left.lref, RWProbeExpr(Node(right)))
       case (true, false)                                => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
       case (false, true)                                => Connect(si, left.lref, ProbeRead(Node(right)))
-      case (false, false) =>
+      case (false, false)                               =>
+        // For non-probe, directly create Nodes for lhs, skipping visibility check to support BoringUtils.drive.
         (left, right) match {
-          case (_: Property[_], _: Property[_]) => PropAssign(si, left.lref, Node(right))
-          case (_, _) => Connect(si, left.lref, Node(right))
+          case (_: Property[_], _: Property[_]) => PropAssign(si, Node(left), Node(right))
+          case (_, _) => Connect(si, Node(left), Node(right))
         }
     }
     val secretCommands = if (_closed) {
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -300,8 +300,8 @@ object BoringUtils {
       Builder.error(s"Cannot bore from $source to ${thisModule.name}, as they do not share a least common ancestor")
     }
     val (upPath, downPath) = lcaResult.get
-    val lcaSource = drill(source, upPath.dropRight(1), upPath.dropRight(1), true && !isDrive)
-    val sink = drill(lcaSource, downPath.reverse.tail, downPath.reverse, false || isDrive)
+    val lcaSource = drill(source, upPath.dropRight(1), upPath.dropRight(1), up = !isDrive)
+    val sink = drill(lcaSource, downPath.reverse.tail, downPath.reverse, up = isDrive)
 
     if (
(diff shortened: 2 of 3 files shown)


src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -376,4 +376,65 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      val a = Wire(Bool())
+    }
+
+    class Bar extends RawModule {
+      val foo = Module(new Foo)
+
+      BoringUtils.drive(foo.a) := 1.B
+    }
+
+    val chirrtl = circt.stage.ChiselStage.emitCHIRRTL(new Bar, Array("--full-stacktrace"))
```
seldridge, line 410: Nit: `--full-stacktrace` is unnecessary. Ditto below.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -290,8 +300,8 @@ object BoringUtils {
       Builder.error(s"Cannot bore from $source to ${thisModule.name}, as they do not share a least common ancestor")
     }
     val (upPath, downPath) = lcaResult.get
-    val lcaSource = drill(source, upPath.dropRight(1), upPath.dropRight(1), true)
-    val sink = drill(lcaSource, downPath.reverse.tail, downPath.reverse, false)
+    val lcaSource = drill(source, upPath.dropRight(1), upPath.dropRight(1), true && !isDrive)
+    val sink = drill(lcaSource, downPath.reverse.tail, downPath.reverse, false || isDrive)
```
seldridge, line 304: Interesting choice of representing this that way. I assume this is done because it is perceived as more understandable than doing the boolean optimization?

```suggestion
    val lcaSource = drill(source, upPath.dropRight(1), upPath.dropRight(1), up=!isDrive)
    val sink = drill(lcaSource, downPath.reverse.tail, downPath.reverse, up=isDrive)
```

seldridge (approved):
LGTM

src/main/scala/chisel3/util/experimental/BoringUtils.scala
mikeurbach: Interesting is putting it politely... I added without thinking about the redundancy.

src/test/scala/chiselTests/BoringUtilsSpec.scala
mikeurbach: Oops yes leftover from debugging and then copied. Thanks.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -211,10 +211,11 @@ abstract class RawModule extends BaseModule {
...
       case (true, false)                                => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
       case (false, true)                                => Connect(si, left.lref, ProbeRead(Node(right)))
-      case (false, false) =>
+      case (false, false)                               =>
+        // For non-probe, directly create Nodes for lhs, skipping visibility check to support BoringUtils.drive.
         (left, right) match {
-          case (_: Property[_], _: Property[_]) => PropAssign(si, left.lref, Node(right))
-          case (_, _) => Connect(si, left.lref, Node(right))
+          case (_: Property[_], _: Property[_]) => PropAssign(si, Node(left), Node(right))
+          case (_, _) => Connect(si, Node(left), Node(right))
```
dtzSiFive, line 218: I don't think node-of-non-passive (e.g., property or probe) are valid and regardless they don't currently work in CIRCT.

I'm not sure what to suggest here -- probably a wire? -- and if this path isn't already tested can you add one for it?

mikeurbach: There is a difference between Chisel internals Node and `node` appearing in the IR. The `lref` method on Data produces a Chisel internals Node already, this is just skipping over some visibility checking that `lref` does, which is expected to fail in the context BoringUtils.drive uses this method. Both of the happy path tests I added go through this code path, and but no `node`s are emitted into the FIRRTL.

dtzSiFive: Awesome! Thanks, appreciate the explanation and glad to hear it!

PR #3974, opened 2024-04-06
[binder] Support constructing integers larger than 64 bits

### LINKED ISSUE #3963: [Binder] Generate wrong IR in some cases of SInt
https://github.com/ucb-bar/berkeley-hardfloat/blob/b3c8a38c286101973b3bc071f7918392343faba7/hardfloat/src/main/scala/primitives.scala#L76

```scala
val shift = (BigInt(-1)<<numInVals.toInt).S>>in
```

Right outputs

```firrtl
node roundMask_shift = dshr(asSInt(UInt<65>(0h10000000000000000)), roundMask_lsbs_2)
```

Binder generates

```firrtl
node roundMask_shift = dshr(SInt<65>(0), roundMask_lsbs_2)
```

SpriteOvO:
Fixes #3963, upstream PR: llvm/circt#6893

This PR fixes the first problem in the issue mentioned above, for the second one, it seems to be just the style of output from CIRCT that doesn't affect the final result and there's nothing we can do about it from Chisel side.

### DIFF
--- a/circtpanamabinding/includeFunctions.txt
+++ b/circtpanamabinding/includeFunctions.txt
@@ -104,6 +104,7 @@ firrtlAttrGetNameKind
 firrtlAttrGetRUW
 firrtlAttrGetMemoryInit
 firrtlAttrGetMemDir
+firrtlAttrGetIntegerFromString
 
 # Utility API
 firrtlValueFoldFlow
--- a/panamaconverter/src/PanamaCIRCTConverter.scala
+++ b/panamaconverter/src/PanamaCIRCTConverter.scala
@@ -326,10 +326,10 @@ class PanamaCIRCTConverter(val circt: PanamaCIRCT, fos: Option[FirtoolOptions], 
       def buildBefore(ref: Op): Op = buildImpl(circt.mlirBlockInsertOwnedOperationBefore(parent, ref.op, _))
     }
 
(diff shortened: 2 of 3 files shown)


circtpanamabinding/includeFunctions.txt
```diff
@@ -104,6 +104,7 @@ firrtlAttrGetNameKind
 firrtlAttrGetRUW
 firrtlAttrGetMemoryInit
 firrtlAttrGetMemDir
+firrtlAttrGetIntegerFromString
```
sequencer, line 107: Please remove the original C-API call and usages to avoid dead code.

sequencer (changes requested):

circtpanamabinding/includeFunctions.txt
SpriteOvO: `mlirIntegerAttrGet` is still used in elsewhere so it can't be removed.

sequencer: Is possible to align them to `firrtlAttrGetIntegerFromString`

SpriteOvO: For small integers (less than 64 bits), they don't need to be parsed from a string with the hack, aligning them will downgrade the performance.

sequencer (approved):
LGTM merge it when ci is green

sequencer:
please rebase

jackkoenig:
Thanks!

PR #3984, opened 2024-04-10
Add layers for colored probe types or enablelayer

### LINKED ISSUE #3983: Layers are not emitted unless a block exists
The use of a layer in a layer colored probe or with an enable layer isn't enough to have Chisel internals cause the layer to be emitted. Consider the following example:

``` scala
//> using scala "2.13.11"
//> using lib "org.chipsalliance::chisel::7.0.0-M1+55-ea2d1036-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::7.0.0-M1+55-ea2d1036-SNAPSHOT"

import chisel3._
import chisel3.probe.Probe
import chisel3.layer.{Layer, Convention, enable}

object LayerA extends Layer(Convention.Bind)

class Foo extends RawModule {
  val a = IO(Output(Probe(UInt(1.W), LayerA)))
  enable(LayerA)
}

object Main extends App {
  println(ChiselStage.emitCHIRRTL(new Foo))
}
```

This produces the following illegal FIRRTL:

``` firrtl
FIRRTL version 4.0.0
circuit Foo :
  public module Foo enablelayer LayerA :
    output a : Probe<UInt<1>, LayerA>

    skip
```

h/t @nandor for finding this issue.

seldridge:
Fix a bug where a layer definition would only be added to Builder state if it was used in a layer block.  Also add these layers to the Builder state if they are used by a layer-colored probe type or by a module's enable layer.

Fixes #3983.

#### Release Notes

Fix a bug where illegal FIRRTL (that was missing a layer) unless a layer block using that layer was provided.

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -106,7 +106,9 @@ object layer {
   /** Call this function from within a `Module` body to enable this layer globally for that module. */
   final def enable(layer: Layer): Unit = layer match {
     case Layer.Root =>
-    case _          => Builder.enabledLayers += layer
+    case _          =>
+      addLayer(layer)
+      Builder.enabledLayers += layer
   }
 
 }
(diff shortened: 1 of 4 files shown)


jackkoenig (approved):
LGTM!

PR #3990, opened 2024-04-11
use versionfile for mill

sequencer:
Reference: https://mill-build.com/mill/contrib/versionfile.html

This provide a unified version management target for future releasing.

After switch to mill, I'd like to store all dependency versions to `version` folder for reference, CI bump and review.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?


jackkoenig:
With the current SBT flow, every push to main publishes a SNAPSHOT based on the current commit. That wouldn't work with a version file.

Similarly, the current flow for tagged releases uses the git tag as the version. We could use a version file but that would require making an extra commit for every release that isn't needed in the current flow.

sequencer:
> With the current SBT flow, every push to main publishes a SNAPSHOT based on the current commit. That wouldn't work with a version file.
Similarly, the current flow for tagged releases uses the git tag as the version. We could use a version file but that would require making an extra commit for every release that isn't needed in the current flow.

Yes, we should dig out a way to maintain version in mill, this is just an example, we can override functions in the `VersionFileModule` trait, or even get rid of it. This is just for discussion.

PR #3993, opened 2024-04-12
Optimize truth table merge

jackkoenig:
I'm collecting numbers, but it should be pretty obvious that not essentially doing a deep copy of every TruthTable both before and after running espresso minimization in the common case where the default value is uniform (i.e. all ones, all zeros, or, most commonly, all dont care) will speed things up.

To implement this, I also added some useful methods to BitPat that you can see the tests for.

Doing split and merge on the `TruthTable` from `TruthTableSpec`:
```scala
  val table = TruthTable(
    Map(
      // BitPat("b000") -> BitPat("b0"),
      BitPat("b001") -> BitPat("b?"),
      BitPat("b010") -> BitPat("b?"),
      // BitPat("b011") -> BitPat("b0"),
      BitPat("b100") -> BitPat("b1"),
      BitPat("b101") -> BitPat("b1"),
      // BitPat("b110") -> BitPat("b0"),
      BitPat("b111") -> BitPat("b1")
    ),
    BitPat("b0")
  )
```
Before this change:
```
Benchmark               Mode  Cnt  Score   Error   Units
ChiselBenchmark.Merge  thrpt   10  0.366 ± 0.003  ops/us
ChiselBenchmark.Split  thrpt   10  0.786 ± 0.006  ops/us
```
After this change:
```
Benchmark               Mode  Cnt    Score   Error   Units
ChiselBenchmark.Merge  thrpt   10  453.041 ± 1.422  ops/us
ChiselBenchmark.Split  thrpt   10   83.122 ± 0.184  ops/us
```

Also from a large real design, I measured it was spending 2 **minutes** in the little filter function in split: https://github.com/chipsalliance/chisel/blob/d584dd7aee4e8b738440386739a66b40c0d53ee6/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala#L145 (this should use a Set instead of a Seq)
With this change, it now spends 0.3ms in that function which just shows how much unnecessary splitting we were doing 🙃.

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Rebase

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/BitPat.scala
+++ b/src/main/scala/chisel3/util/BitPat.scala
@@ -437,6 +437,18 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, val width: Int)
     }
   }
 
+  /** Are any bits of this BitPat `?` */
+  private[chisel3] def hasDontCares: Boolean = width > 0 && mask != ((BigInt(1) << width) - 1)
+
+  /** Are all bits of this BitPat `0` */
+  private[chisel3] def allZeros: Boolean = value == 0 && !hasDontCares
+
+  /** Are all bits of this BitPat `1` */
+  private[chisel3] def allOnes: Boolean = !hasDontCares && value == mask
+
+  /** Are all bits of this BitPat `?` */
+  private[chisel3] def allDontCares: Boolean = mask == 0
+
   override def isEmpty: Boolean = false
 
   /** Generate raw string of a `BitPat`. */
--- a/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
+++ b/src/main/scala/chisel3/util/experimental/decode/TruthTable.scala
@@ -162,7 +162,13 @@ object TruthTable {
     def index(bitPat: BitPat, bpType: Char): Seq[Int] =
       bitPat.rawString.zipWithIndex.filter(_._1 == bpType).map(_._2)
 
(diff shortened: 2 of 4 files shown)


src/main/scala/chisel3/util/BitPat.scala
```diff
@@ -437,6 +437,18 @@ sealed class BitPat(val value: BigInt, val mask: BigInt, val width: Int)
     }
   }
 
+  /** Are any bits of this BitPat `?` */
+  private[chisel3] def hasDontCares: Boolean = width > 0 && mask != ((BigInt(1) << width) - 1)
```
sequencer, line 441: width of bp cannot be 0. Maybe assert out?

jackkoenig: I'm future proofing this such that if we do allow it, these should be correct for it (see stubs in tests).

sequencer:
> With this change, it now spends 0.3ms in that function which just shows how much unnecessary splitting we were doing 🙃.

the reason of splitting is because cannot support the existence of both `0`, `1` in the result of table. To support this, we have to split it to multiple runs of espresso.

sequencer (approved):
LGTM. Let’s make common case fast!
Since split is only designed for long decode output with 0 and 1

PR #3998, opened 2024-04-15
In BoringUtils.drive, don't bore into the final module for inputs.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

This follows up on https://github.com/chipsalliance/chisel/pull/3960 to address the behavior at the final module for the sink being driven. We already have special handling for boring out from the original source, and we need the inverse here. When we reach the final sink, if it is an input port, we should use it, rather than boring into the module and connecting to it.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/experimental/BoringUtils.scala
+++ b/src/main/scala/chisel3/util/experimental/BoringUtils.scala
@@ -243,6 +243,9 @@ object BoringUtils {
       case Some(PortBinding(_)) => true
       case _                    => false
     }
+    def isDriveDone(d: Data): Boolean = {
+      isDrive && isPort(d) && DataMirror.directionOf(d) == ActualDirection.Input
+    }
     def boringError(module: BaseModule): Unit = {
       (module.fullyClosedErrorMessages ++ Seq(
         (si, s"Can only bore into modules that are not fully closed: ${module.name} was fully closed")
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -254,8 +257,9 @@ object BoringUtils {
       path.zip(connectionLocation).foldLeft(source) {
         case (rhs, (module, conLoc)) if (module.isFullyClosed) => boringError(module); DontCare.asInstanceOf[A]
         case (rhs, (module, _))
-            if (up && module == path(0) && isPort(rhs) && (!createProbe.nonEmpty || !createProbe.get.writable)) => {
-          // When drilling from the original source, if it's already a port just return it.
+            if ((up || isDriveDone(rhs)) && module == path(0) && isPort(rhs) &&
```
mikeurbach, line 260: This diff isn't great due to line wrapping, but the change was `up` -> `(up || isDriveDone(rhs))`, which has the logic to know if we need to stop at the last boundary or go inside it. We stop on an input, but we go inside on an output. The tests show what I mean. Previously, we'd always go inside, which creates illegal FIRRTL for driving and input.

dtzSiFive: Aside to I suppose mostly myself: We could probably change the RWProbe bores from inside behavior now, actually.

dtzSiFive (approved):
LGTM!

src/main/scala/chisel3/util/experimental/BoringUtils.scala
mikeurbach: This logic could use a cleanup in general... I've only made it worse. So maybe settling on the final RWProbe behavior could be done as part of that.

PR #3912, opened 2024-03-07
Add a label "chisel3_builtin" to ifElseFatal intrinsic

uenoku:
This is a follow-up to #3825.

MFC has attached a laebl "chisel3_builtin" to every ifElseFatal assertion. Usually a label on ifElseFatal is just ignored but the label is emitted when compiled with `-emit-chisel-asserts-as-sva`. So labels on assertions converted from ifElseFatal disappear due to #3825. This PR restores labels to preserve the old behavior.

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you add appropriate documentation in `docs/src`?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference).

#### Release Notes

Add a label "chisel3_builtin" to ifElseFatal intrinsic.
### Reviewer Checklist (only modified by reviewer)
- [x] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [x] Did you mark the proper milestone (Bug fix: `3.6.x`, `5.x`, or `6.x` depending on impact, API modification or big change: `7.0`)?
- [x] Did you review?
- [x] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #4001, opened 2024-04-17
Add ifElseFatal intrinsics; use it for chisel3.assert emission

dtzSiFive:
Restore https://github.com/chipsalliance/chisel/pull/3912 and https://github.com/chipsalliance/chisel/pull/3825.
Use intrinsic expression instead of intrinsic module.

This reverts commit https://github.com/chipsalliance/chisel/commit/347d82b8c1c41f8cfd26d07004491e6bc0c549fb.

### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

Support `circt_chisel_ifelsefatal` intrinsics and use it for chisel3.assert emission.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/naming/PrefixSpec.scala
+++ b/src/test/scala/chiselTests/naming/PrefixSpec.scala
@@ -460,17 +460,15 @@ class PrefixSpec extends ChiselPropSpec with Utils {
 
       {
         val x5 = {
-          val x1 = chisel3.assert(1.U === 1.U)
           val x2 = cover(foo =/= bar)
           val x3 = chisel3.assume(foo =/= 123.U)
           val x4 = printf("foo = %d\n", foo)
-          x1
+          x2
         }
       }
     }
     val chirrtl = ChiselStage.emitCHIRRTL(new Test)
-    (chirrtl should include).regex("assert.*: x5")
-    (chirrtl should include).regex("cover.*: x5_x2")
+    (chirrtl should include).regex("cover.*: x5")
     (chirrtl should include).regex("assume.*: x5_x3")
     (chirrtl should include).regex("printf.*: x5_x4")
   }
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -28,6 +28,9 @@ package experimental {
   case class DoubleParam(value: Double) extends Param
   case class StringParam(value: String) extends Param
 
+  /** Creates a parameter from the Printable's resulting format String */
+  case class PrintableParam(value: chisel3.Printable, context: BaseModule) extends Param
+
   /** Unquoted String */
   case class RawParam(value: String) extends Param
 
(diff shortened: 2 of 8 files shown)


src/test/scala/chiselTests/naming/PrefixSpec.scala
```diff
@@ -460,17 +460,15 @@ class PrefixSpec extends ChiselPropSpec with Utils {
...
-          x1
+          x2
         }
       }
     }
     val chirrtl = ChiselStage.emitCHIRRTL(new Test)
-    (chirrtl should include).regex("assert.*: x5")
-    (chirrtl should include).regex("cover.*: x5_x2")
+    (chirrtl should include).regex("cover.*: x5")
     (chirrtl should include).regex("assume.*: x5_x3")
```
jackkoenig, line 472: I know this PR is just a re-application of 2 previously reverted PRs so I'm not going to block on this issue (since it was in a PR I previously approved), but this is not ideal. We should be able to maintain the old label behavior, but it will probably require a new Printable that wraps the Assert() object that we return from `chisel3.assert`.

jackkoenig (approved):
LGTM although we should follow up by restoring the labeling behavior.

Merge once you're confident that the old performance issue is fixed (I'm pretty sure it is but it's important to check!)

src/test/scala/chiselTests/naming/PrefixSpec.scala
dtzSiFive: Ah, so instead of "chisel3_builtin" being the label for all of these, they get the "valname"/name included that would previously be mixed in by putting it in the "optional name" (trailing ": asdf" this test checks for)?

Yes that seems rather useful!

So Printable would help get a friendly name available (that's not available when IntrinsicExpr apply is called) so we could append it to the label StringParam while converting to the FIRRTL IR object, thereabouts? Makes sense to me!

dtzSiFive:
> Merge once you're confident that the old performance issue is fixed (I'm pretty sure it is but it's important to check!)

Yes, it's MUCH better now! Smaller FIRRTL too (compared to printf encoding).

Going to sit on this for at least a bit though to test things more thoroughly re:heavy intrinsic (expression) usage, let's ensure this sticks best way can this time 😉 .

src/test/scala/chiselTests/naming/PrefixSpec.scala
jackkoenig: > So Printable would help get a friendly name available (that's not available when IntrinsicExpr apply is called) so we could append it to the label StringParam while converting to the FIRRTL IR object, thereabouts? Makes sense to me!

Exactly!

> FYI previous behavior was just dropping optional names (assert_ label in the test) emitted for assert op. chisel3_builtin was used regardless of optional names.

Perhaps for these asserts in firtool, but the intent was always to try to replace printf-encoded with real asserts with reasonable label behavior. This has been used by ChiselTest users, we just never fully implemented it in firtool, nor for printf-encoded asserts, but we should.

PR #4013, opened 2024-04-18
Add Tywaves project to community projects page

rameloni:
### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/website/src/pages/community.md
+++ b/website/src/pages/community.md
@@ -52,7 +52,11 @@ If you want to add your project to the list, let us know on the [Chisel users ma
 | [RISCVAssembler](https://github.com/carlosedp/riscvassembler)                                                                                | A RISC-V assembler library for Scala/Chisel projects            | [`@carlosedp`](https://github.com/carlosedp)                                     | [Site](https://index.scala-lang.org/carlosedp/riscvassembler), [Demo Site](https://carlosedp.github.io/rvasmweb/)) |
 | [SoC-Now](https://github.com/merledu/SoC-Now-Generator)                                                                                | A plug and play supported RISC-V System-on-Chip Generator, 5-stage RISC-V Microprocessor            | [`@merledu`](https://github.com/merledu)                                     | [Site](https://github.com/merledu/SoC-Now-Generator), [Poster](https://woset-workshop.github.io/WOSET2022.html#article-6), [Video](https://woset-workshop.github.io/Videos/2022/6-Shahzaib.mp4) |
 
+### Tooling
 
+| Project                                                    | Destription                           | Author                                     | Links                                                                                                                                                    |
+| ---------------------------------------------------------- | ------------------------------------- | ------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [Tywaves](https://github.com/rameloni/tywaves-chisel-demo) (demo) | Type-based waveform viewer for Chisel | [`@rameloni`](https://github.com/rameloni) | [Chisel backend](https://github.com/rameloni/tywaves-chisel-demo), [Surfer-Tywaves fork](https://gitlab.com/rameloni/surfer-tywaves-demo/-/tree/tywaves) |
 
 ### FIRRTL
 


jackkoenig (approved):
Great! Thanks!

sequencer:
Nice to see it! I'm also using the surfer for debugging! would it be possible to upstream to surfer?

rameloni:
Yes, the plan is that. However, the demo is at the moment a "proof of concept" rather than a fully working tool. I am currently working on a more stable version which will integrate better into Chisel and CIRCT compiler (firtool) and will also provide a better visualization in surfer. 
I added a short description here about the main drawbacks of the demo and why it needs another approach: https://github.com/rameloni/tywaves-chisel-demo?tab=readme-ov-file#drawbacks

sequencer:
Yes, this is a good reason, `hgldd` is mainly for Verdi, while chisel only exposes source info to `firtool` for debugging, which makes your tool depending the `ChiselIR` which isn't stable enough.

Since chisel has an unstable type, what if only use hgldd+sourceinfo+surfer to provide an IDE-like experiencing, like what hgldd+sourceinfo+verdi does. It may avoid serializing Chisel type and remapping. IMHO, remapping chisel seems to be an extremely hard since there are too much things to think about, e.g. `elements` in `Record`...  

Would it be possible to use this tool to provide a language-server like experience? You can think about it like this: all chisel type is still dynamic and store in the JVM context. source info can tell the wave-viewer where line to goes to, user can debug vcd or fst w/ an IDE(e.g. JetBrains) reading mlir and vcd from plugin, so that you don't need to expose the Chisel type to surfer, but can still debug it via sourceinfo: e.g: click a signal in surfer and show the sourceinfo or click a sourceinfo from IDE and show it in surfer.

The vcd rewrite phase is certainly a burden to align the symbol from verilator to chisel types. However we can release this constraint to reconstruct the hierarchy to firrtl type. which is much easier to do!

For the rewrite pass in surfer, how about using a symbol type mapper to map different symbols? which reduces development burden to map different simulator tools in different version symbol tables. and I believe that much simpler for upstream!

rameloni:
> Since chisel has an unstable type, what if only use hgldd+sourceinfo+surfer to provide an IDE-like experiencing, like what hgldd+sourceinfo+verdi does. It may avoid serializing Chisel type and remapping. IMHO, remapping chisel seems to be an extremely hard since there are too much things to think about, e.g. `elements` in `Record`...

Hgldd is however not stable, since it is made mainly for verdi, it may change basing on what the would need. 

> Would it be possible to use this tool to provide a language-server like experience? You can think about it like this: all chisel type is still dynamic and store in the JVM context. source info can tell the wave-viewer where line to goes to, user can debug vcd or fst w/ an IDE(e.g. JetBrains) reading mlir and vcd from plugin, so that you don't need to expose the Chisel type to surfer, but can still debug it via sourceinfo: e.g: click a signal in surfer and show the sourceinfo or click a sourceinfo from IDE and show it in surfer.

This would be certainly possible, especially jumping to source from surfer would be a great feature to add to surfer. 
However, with Type-based waveform viewer I would want to highlight the main nature of the tool, that is showing a typed representation in the viewer. This project was born from another (Tydi-Chisel and Tydi-lang) in which users can specify custom data types. Tydi is an open-source specification to map complex, dynamically sized data structures onto hardware streams and allows to describe those types and streams with high abstractions thanks to tydi-lang. While Tydi-Chisel is an implementation of the specification in Chisel. Briefly, in the tydi-flow users specify circuit components in tydi-lang, this is transpiled to boilerplate Chisel and then they can add behaviour of core blocks without thinking about stream details and use Chisel constructs and testing. Indeed, it initially aimed to be a viewer for Tydi. But, since Tydi is now implemented in Chisel, I had to start from Chisel and then move to tydi.
In this domain, not only the type itself of a variable would matter but also the representation of its values. For example, a user in tydi-lang can specify a stream of strings in the following way:
```
Char_t = Bit(8);
Char_stream = Stream(Char_t, t=3.0, d=1, c=1);
```
Which internally will be seen as a bus of `UInt<8>`. However, types in Tydi matter and users may want to see the stream of chars as a string in the viewer itself. Also, since Tydi-Chisel abstracts the internal details of its implementation, users may not want to see a bunch of signals of tydi streams, but some higher-level representation.

Because of this, the viewer needs to know exact information about Chisel, starting from types of the objects, I will extend it with custom representations and so on (maybe that users can also specify).

> The vcd rewrite phase is certainly a burden to align the symbol from verilator to chisel types. However we can release this constraint to reconstruct the hierarchy to firrtl type. which is much easier to do!

Do you mean during compilation from firrtl to verilog?

> For the rewrite pass in surfer, how about using a symbol type mapper to map different symbols? which reduces development burden to map different simulator tools in different version symbol tables. and I believe that much simpler for upstream!

I'm sorry but I don't get this last point: `a symbol type mapper to map different symbols`. Anyway, the vcd rewrite in surfer is currently a workaround to make it working for the demo as it is now. Surfer, internally, loads every signal in the vcd first and then allows the user to select each of these signals. Once the signal is selected, its value is passed to a transformer trait which is responsible for rendering it. This makes rendering aggregates difficult since only the flattened fields could be selected. Selecting signals differently would've required to me to much time that I didn't have at the moment I was implementing the demo.

rameloni:
@sequencer 
Probably an Ide-like experience might be useful for the old chisel versions since I'm currently working on the development branch, the viewer won't be compatible with older chisel versions

PR #4014, opened 2024-04-19
Support D/I in toRelativeTarget with a specified root.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

If we are in a Definition, it would be best to not throw an exception. The only sane thing to do in this case seems to be to treat the Definition as the root.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/ToTargetSpec.scala
+++ b/src/test/scala/chiselTests/ToTargetSpec.scala
@@ -3,6 +3,7 @@
 package chiselTests
 
 import chisel3._
+import chisel3.experimental.hierarchy.{instantiable, public, Instantiate}
 import chisel3.properties.{Path, Property}
 import circt.stage.ChiselStage
 
@@ -79,6 +80,46 @@ class RelativeSiblingsModule extends RawModule {
 
   atModuleBodyEnd {
     val reference = middle1.inner.wire.toRelativeTarget(Some(middle2))
+  }
+}
+
+@instantiable
+class RelativeDefinitionGrandChildModule(relativeRoot: RawModule) extends RawModule {
+  @public val grandChildWire = Wire(Bool())
+}
+
+@instantiable
+class RelativeDefinitionChildModule(relativeRoot: RawModule) extends RawModule {
+  val inProgressChild = IO(Output(Property[Path]()))
+  val inProgressGrandChild = IO(Output(Property[Path]()))
+
+  @public val childWire = Wire(Bool())
+
+  @public val grandChild = Module(new RelativeDefinitionGrandChildModule(relativeRoot))
+
+  atModuleBodyEnd {
+    // Check relative targets created from an in-progress Definition.
+    val childReference = childWire.toRelativeTarget(Some(relativeRoot))
+    inProgressChild := Property(Path(childReference))
+
+    val grandChildReference = grandChild.grandChildWire.toRelativeTarget(Some(relativeRoot))
+    inProgressGrandChild := Property(Path(grandChildReference))
+  }
+}
+
+class RelativeDefinitionParentModule extends RawModule {
+  val completedChild = IO(Output(Property[Path]()))
+  val completedGrandChild = IO(Output(Property[Path]()))
+
+  val child = Instantiate(new RelativeDefinitionChildModule(this))
+
+  atModuleBodyEnd {
+    // Check relative targets created from a completed Definition.
+    val childReference = child.childWire.toRelativeTarget(Some(this))
+    completedChild := Property(Path(childReference))
+
+    val grandChildReference = child.grandChild.grandChildWire.toRelativeTarget(Some(this))
+    completedGrandChild := Property(Path(grandChildReference))
   }
 }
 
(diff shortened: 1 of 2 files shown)


jackkoenig:
What happens if both the target and the root are in the same Definition but the root is not an ancestor of the target?

mikeurbach:
Let me add a unit test for this scenario. I would hope that would fail with the previous error, but I have a suspicion this isn't what I want.

Is there a good way to tell from instance variables or global state that a given RawModule is being elaborated to become a Defintion (besides the _parent instance variable being None, which is helpful but not sufficient)? I used toDefinitionCalled for the check on a RawModule that was finished, but I'm wondering about the in-progress case? If it isn't easy to check, maybe we can start with only supporting this for completed Definitions using toDefinitionCalled.

src/test/scala/chiselTests/ToTargetSpec.scala
```diff
@@ -82,6 +83,46 @@ class RelativeSiblingsModule extends RawModule {
...
+    val grandChildReference = grandChild.grandChildWire.toRelativeTarget(Some(relativeRoot))
+    inProgressGrandChild := Property(Path(grandChildReference))
+  }
+}
+
+class RelativeDefinitionParentModule extends RawModule {
+  val completedChild = IO(Output(Property[Path]()))
+  val completedGrandChild = IO(Output(Property[Path]()))
+
+  val child = Instantiate(new RelativeDefinitionChildModule(this))
```
sequencer, line 114: When `Instantiate` two `RelativeDefinitionChildModule` do we need to elaborate twice?

Rather than passing `this` to `RelativeDefinitionChildModule`, what would happen if passing other modules to it?

I wonder, rather than passing it in parameter, how about passing the target from `Input(Property[Path]())`? This might keep the module parameter stable?

sequencer (commented):
When `Instantiate` two `RelativeDefinitionChildModule` do we need to elaborate twice?

Rather than passing `this` to `RelativeDefinitionChildModule`, what would happen if passing other modules to it?

I wonder, rather than passing it in parameter, how about passing the target from `Input(Property[Path]())`? This might keep the module parameter stable?

mikeurbach:
> I wonder, rather than passing it in parameter, how about passing the target from Input(Property[Path]())? This might keep the module parameter stable?

Yes, this is not really how I'd write this code, I just wanted an easy way to trigger the exception so I could test the change that using toRelativeTarget under a Definition should stop if it reaches the Definition boundary, rather than throwing an exception.

mikeurbach:
> What happens if both the target and the root are in the same Definition but the root is not an ancestor of the target?

I tried to write something like this:

```
@instantiable
class RelativeDefinitionWrongChildModule extends RawModule {
  @public val target = Module(new RawModule {})
  @public val root = Module(new RawModule {})
}

class RelativeDefinitionWrongParentModule extends RawModule {
  val referenceOut = IO(Output(Property[Path]()))

  val child = Instantiate(new RelativeDefinitionWrongChildModule)

  atModuleBodyEnd {
    val reference = child.target.toRelativeTarget(Some(child.root))
    referenceOut := Property(Path(reference))
  }
}
```

But it doesn't seem possible to write this toRelativeTarget call, since the instances within RelativeDefinitionWrongChildModule are hierarchy.Instance, not BaseModule. Instead, I can write something like this at least:

```
@instantiable
class Target extends RawModule {
  @public val wire = Wire(Bool())
}

@instantiable
class RelativeDefinitionWrongChildModule extends RawModule {
  @public val target = Module(new Target)
}

class RelativeDefinitionWrongParentModule extends RawModule {
  val referenceOut = IO(Output(Property[Path]()))

  val child = Instantiate(new RelativeDefinitionWrongChildModule)
  val root = Module(new RawModule {})

  atModuleBodyEnd {
    val reference = child.target.wire.toRelativeTarget(Some(root))
    referenceOut := Property(Path(reference))
  }
}
```

This does fail as expected. I think you're more interested in something else though.

For example, I can at least write this:

```
@instantiable
class RelativeDefinitionWrongChildModule extends RawModule {
  val referenceOut = IO(Output(Property[Path]()))

  val a = Module(new RawModule {})
  val b = Module(new RawModule {})

  atModuleBodyEnd {
    val reference = a.toRelativeTarget(Some(b))
    referenceOut := Property(Path(reference))
  }
}

class RelativeDefinitionWrongParentModule extends RawModule {
  val child = Instantiate(new RelativeDefinitionWrongChildModule)
}
```

This does succeed erroneously, and makes the target relative to `RelativeDefinitionWrongChildModule`.

So I think the crux here is the call to Builder.inDefinition where I'm trying to check if the current module is being elaborated to become a Definition. Builder.inDefinition is not what I want, but I don't think anything like this exists. At that point in time, the RawModule is just a RawModule, not an InstanceClone or ModuleClone or anything like that.

mikeurbach:
I can update this PR to dial back the check for Builder.inDefinition, and let it continue to throw errors in the case we use toRelativeTarget for something inside an in-progress Definition elaboration.

I think we can at least be certain if an elaborated Definition is done with toDefinitionCalled. But even in that case, we can't tell the difference between a Definition that was indeed contained by the requested root or not, since a Definition doesn't have any parents, so the question isn't really well formed. Maybe this just isn't a good idea, and we shouldn't mix toRelativeTarget with Definition in these situations.

PR #4018, opened 2024-04-19
Fix Nested Instantiate

sequencer:
When Instantiate a instantiable module, if it has other instantiable modules inside it, inner module will miss cache.

### Contributor Checklist

#### Type of Improvement
- Bugfix

#### Desired Merge Strategy
- Squash

#### Release Notes
Fix Nested Instantiate

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -455,7 +455,9 @@ private[chisel3] class DynamicContext(
   val defaultNamespace:      Option[Namespace],
   // Definitions from other scopes in the same elaboration, use allDefinitions below
   val outerScopeDefinitions: List[Iterable[Definition[_]]],
-  val loggerOptions:         LoggerOptions) {
+  val loggerOptions:         LoggerOptions,
+  val definitions:           ArrayBuffer[Definition[_]],
+  val contextCache:          BuilderContextCache) {
   val importedDefinitionAnnos = annotationSeq.collect { case a: ImportDefinitionAnnotation[_] => a }
 
   // Map from proto module name to ext-module name
@@ -506,7 +508,6 @@ private[chisel3] class DynamicContext(
   }
 
   val components = ArrayBuffer[Component]()
-  val definitions = ArrayBuffer[Definition[_]]()
   val annotations = ArrayBuffer[ChiselAnnotation]()
   val newAnnotations = ArrayBuffer[ChiselMultiAnnotation]()
   val layers = mutable.LinkedHashSet[layer.Layer]()
@@ -520,8 +521,6 @@ private[chisel3] class DynamicContext(
 
   // Views that do not correspond to a single ReferenceTarget and thus require renaming
   val unnamedViews: ArrayBuffer[Data] = ArrayBuffer.empty
-
-  val contextCache: BuilderContextCache = BuilderContextCache.empty
 
   // Set by object Module.apply before calling class Module constructor
   // Used to distinguish between no Module() wrapping, multiple wrappings, and rewrapping
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -454,10 +454,9 @@ private[chisel3] class DynamicContext(
   val sourceRoots:           Seq[File],
   val defaultNamespace:      Option[Namespace],
   // Definitions from other scopes in the same elaboration, use allDefinitions below
-  val outerScopeDefinitions: List[Iterable[Definition[_]]],
-  val loggerOptions:         LoggerOptions,
-  val definitions:           ArrayBuffer[Definition[_]],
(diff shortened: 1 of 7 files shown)


src/test/scala/chiselTests/experimental/hierarchy/InstantiateSpec.scala
```diff
@@ -475,4 +477,24 @@ class InstantiateSpec extends ChiselFunSpec with Utils {
...
+  @instantiable
+  class Foo(i: Int) extends Module {
+    val om = Instantiate(new FooOM)
+  }
+  it("Instantiate should work with Class") {
+    class MyTop extends Top {
+      val inst0 = Instantiate(new Foo(1))
+      val inst1 = Instantiate(new Foo(2))
+    }
+    assert(convert(new MyTop).modules.map(_.name).toSet == Seq("FooOM", "Foo", "Foo_1", "Top").toSet)
```
sequencer, line 498: Reproduce via:
```
mill chiselut[2.13.12].testOnly chiselTests.experimental.hierarchy.InstantiateSpec
```
Here is log:
```
- Instantiate should work with Class *** FAILED ***
  HashSet("FooOM", "Top", "Foo_1", "Foo", "FooOM_1") did not equal Set("FooOM", "Foo", "Foo_1", "Top") (InstantiateSpec.scala:498)
  Analysis:
  HashSet(missingInRight: [FooOM_1])
1 targets failed
chiselut[2.13.12].testOnly 1 tests failed:
  chiselTests.experimental.hierarchy.InstantiateSpec Instantiate should work with Class
```

src/test/scala/chiselTests/experimental/hierarchy/InstantiateSpec.scala
```diff
@@ -475,4 +491,12 @@ class InstantiateSpec extends ChiselFunSpec with Utils {
       )
     )
   }
+
+  it("Nested Instantiate should work") {
```
sequencer, line 495: Reproduce via:
```
mill chiselut[2.13.12].testOnly chiselTests.experimental.hierarchy.InstantiateSpec
```
Here is log:
```
List("Bar", "Bar_1", "Bar_2", "Bar_3", "Baz", "Baz_1", "Baz_2", "Baz_3", "Foo", "Foo_1", "Top") did not equal List("Bar", "Bar_1", "Baz", "Foo", "Foo_1", "Top")
ScalaTestFailureLocation: chiselTests.experimental.hierarchy.InstantiateSpec at (InstantiateSpec.scala:500)
Expected :List("Bar", "Bar_1", "Baz", "Foo", "Foo_1", "Top")
Actual   :List("Bar", "Bar_1", "Bar_2", "Bar_3", "Baz", "Baz_1", "Baz_2", "Baz_3", "Foo", "Foo_1", "Top")
```

sequencer (commented):
This bug is because in each definition elaboration, previous elaborated definition and cache doesn't exist, so `Definition`/`Instantiate` inside other `Definition` never worked before.
This doesn't affect `Module` since it can be resolved by firtool dedup, but for `Class`, multiple `Instantiate` lead to different type: e.g. `FooOM`, `FooOM_1`, `FooOM_2`.

mikeurbach:
Thanks for the PR. I had been playing around with Instantiate and noticed this behavior, but I never got around to sending a PR or filing an issue. I took a quick look and I think this makes sense, I had something similar when I was trying to fix this on my end once.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -455,7 +455,9 @@ private[chisel3] class DynamicContext(
   val defaultNamespace:      Option[Namespace],
   // Definitions from other scopes in the same elaboration, use allDefinitions below
   val outerScopeDefinitions: List[Iterable[Definition[_]]],
-  val loggerOptions:         LoggerOptions) {
+  val loggerOptions:         LoggerOptions,
+  val definitions:           ArrayBuffer[Definition[_]],
```
jackkoenig, line 459: See `outerScopeDefinitions` above (added in https://github.com/chipsalliance/chisel/pull/3895). I added that to fix passing Definitions down to other Definitions, but it does essentially remove nested definitions. I think just switching to use the `ArrayBuffer` is probably right (I'm unsure why I was thinking we need to remove definitions added by the nested Definitions...).

Regardless, please remove `outerScopeDefinitions` and make sure everything works with just definitions because this change makes `outerScopeDefinitions` very weird (everything in the list is going to be the same ArrayBuffer).

sequencer: Hahahahahaha, I thought about this, but dear not to touch it...
Will do it tonight.

sequencer:
@jackkoenig Another question: should `Definition` also takes like `Instantiate`? e.g. when found a `Definition` is defined. don't elaborate it twice.

SpriteOvO (approved):

PR #4021, opened 2024-04-22
Add Scala Compiler Plugin for SerializableModuleGenerator

sequencer:
In this PR, I'm going to add a new Phase of ChiselPlugin to create `SomeSerializableModuleGenerator` object for `SomeSerializableModule` which extends from a `SerializableModule[SomeSerializableModuleParameter]` which contains an main function to elaborate a mlirbc file, which will be used for tool automation to generate SomeSerializableModuleGenerator with a json string.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -21,6 +21,7 @@ object v {
   val jmhVersion = "1.37"
   val osLib = ivy"com.lihaoyi::os-lib:0.9.1"
   val upickle = ivy"com.lihaoyi::upickle:3.1.0"
+  val mainargs = ivy"com.lihaoyi::mainargs:0.7.0"
   val firtoolResolver = ivy"org.chipsalliance::firtool-resolver:2.0.0"
   val scalatest = ivy"org.scalatest::scalatest:3.2.14"
   val scalacheck = ivy"org.scalatestplus::scalacheck-1-15:3.2.11.0"
@@ -95,6 +96,8 @@ trait Core extends common.CoreModule with ChiselPublishModule with CrossSbtModul
   def osLibModuleIvy = v.osLib
 
   def upickleModuleIvy = v.upickle
+
+  def mainargsModuleIvy = v.mainargs
 
   def firtoolResolverModuleIvy = v.firtoolResolver
 
--- a/common.sc
+++ b/common.sc
@@ -45,6 +45,8 @@ trait CoreModule extends ScalaModule with HasMacroAnnotations {
 
   def upickleModuleIvy: Dep
 
+  def mainargsModuleIvy: Dep
+
   def firtoolResolverModuleIvy: Dep
 
   override def moduleDeps = super.moduleDeps ++ Seq(macrosModule, firrtlModule)
@@ -52,6 +54,7 @@ trait CoreModule extends ScalaModule with HasMacroAnnotations {
   override def ivyDeps = super.ivyDeps() ++ Agg(
     osLibModuleIvy,
     upickleModuleIvy,
+    mainargsModuleIvy,
     firtoolResolverModuleIvy
   )
 }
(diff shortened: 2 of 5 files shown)


plugin/src/main/scala/chisel3/internal/plugin/SerializableModule.scala
```diff
@@ -0,0 +1,52 @@
...
+        unit.body = new MyTypingTransformer(unit).transform(unit.body)
+      }
+    }
+  }
+
+  private class MyTypingTransformer(unit: CompilationUnit) extends TypingTransformer(unit) with ChiselInnerUtils {
+    override def transform(tree: Tree): Tree = tree match {
+      case packageDef @ PackageDef(pid, stats) =>
+        val newStats = stats.flatMap {
+          case classDef: ClassDef if isASerializableModule(classDef.symbol) && !classDef.mods.hasFlag(Flag.ABSTRACT) =>
```
jackkoenig, line 39: Rather than transforming at the package level, I suspect you can just do this at the `ClassDef` level and I think it would be best to insert the generated object into the companion object for the Class. We do that sort of thing in the `@instantiable` macro. Macros aren't exactly the same as plugins but there's a lot of similarity.

plugin/src/main/scala/chisel3/internal/plugin/SerializableModule.scala
```diff
@@ -0,0 +1,52 @@
...
+            val parameterTypeName = classDef.tparams.filter(t => isASerializableModuleParameter(t.symbol)).head.name.toTypeName
+            val moduleTypeName = classDef.name.toTypeName
+            val moduleTermName = classDef.name.toTermName
+            val generatorObject = q"""object ${TermName(moduleTermName.toString + "Generator")} extends App { _root_.chisel3.experimental.SerializableModuleGenerator(classOf[${moduleTypeName}], upickle.default.read[${parameterTypeName}](args.head)(implictly[upickle.default.Read[${parameterTypeName}]])) } """
```
jackkoenig, line 43: Please break this line up. It's also really important to check that this generated object doesn't collide with anything existing--find a way to check the names of symbols in the companion object when generating it--what we do if there's a collision is debatable, we could have some unification algorithm, we could just not generate it, or we could error.

sequencer: Sure! I'm planing migrate to a trait implementation. I have some release ddl these two days, I'll go back to it this weekend.

plugin/src/main/scala/chisel3/internal/plugin/SerializableModule.scala
sequencer: > I think it would be best to insert the generated object into the companion object for the Class.

Here is of my original considerations:
- If adding a companion object, we may need to forbid user adding their own companion, this is possible, but I'm not sure if is necessary. or we need to implement a compile time logic to detect if there is already an companion, then adding logic to it, if not, creating a companion.
- This is the reason I choose to add a new object that user should never implement, e.g. `FooModuleGenerator`.

jackkoenig: > If adding a companion object, we may need to forbid user adding their own companion, this is possible, but I'm not sure if is necessary. or we need to implement a compile time logic to detect if there is already an companion, then adding logic to it, if not, creating a companion.

No need to forbid, you just create the object only if it doesn't exist, otherwise you just insert into the existing one. This is pretty standard practice (e.g. case classes generate methods in the companion object, creating it if necessary), it's a normal Scala-y thing to do so I would expect there to be an API.

PR #4023, opened 2024-04-22
Add Attribute annotation back to chisel in-tree firrtl

lordspacehog:
### Contributor Checklist

- [X] Did you add Scaladoc to every public function/method?
- [X] Did you add at least one test demonstrating the PR?
- [X] Did you add appropriate documentation in `docs/src`?

#### Type of Improvement
Bugfix

#### Desired Merge Strategy
Squash: The PR will be squashed and merged (choose this if you have no preference).

#### Release Notes
Add AttributeAnnotation back to in-tree firrtl

- [ ] Did you review?

### DIFF
--- a/firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
+++ b/firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+package firrtl
+
+import firrtl.annotations.{Named, SingleTargetAnnotation}
+
+/** Firrtl implementation for verilog attributes
+  * @param target       target component to tag with attribute
+  * @param description  Attribute string to add to target
+  */
+case class AttributeAnnotation(target: Named, description: String) extends SingleTargetAnnotation[Named] {
+  def duplicate(n: Named): AttributeAnnotation = this.copy(target = n, description = description)
+}
--- a/firrtl/src/test/scala/firrtlTests/annotationTests/AttributeAnnotationSpec.scala
+++ b/firrtl/src/test/scala/firrtlTests/annotationTests/AttributeAnnotationSpec.scala
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package firrtlTests.annotationTests
+
+import firrtl.AttributeAnnotation
+import firrtl.annotations._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class AttributeAnnotationSpec extends AnyFreeSpec with Matchers {
+  "AttributeAnnotation should be correctly parsed from a string" in {
+    val attribAnno = new AttributeAnnotation(
+      ComponentName("attrib", ModuleName("ModuleAttrib", CircuitName("CircuitAttrib"))),
+      "X_INTERFACE_INFO = \"some:interface:type:1.0 SIGNAL\""
+    )
+
+    val annoString = JsonProtocol.serializeTry(Seq(attribAnno)).get
+    val loadedAnnos = JsonProtocol.deserializeTry(annoString).get
+    attribAnno should equal(loadedAnnos.head)
+  }
+}
--- a/src/main/scala/chisel3/util/AttributeAnnotation.scala
+++ b/src/main/scala/chisel3/util/AttributeAnnotation.scala
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-
 package chisel3.utils
 
 import firrtl.AttributeAnnotation
@@ -8,8 +7,34 @@ import firrtl.annotations.Named
 import chisel3.Data
 import chisel3.experimental.{annotate, requireIsAnnotatable, ChiselAnnotation}
 
-object annotateAttribute {
-  def apply[T <: Data](target: T, annoString: String): Unit = {
+/** Helper Object for applying Attribute Annotations */
+object addAttribute {
+
+  /** Add attribute annotation to a chisel target.
+    *
+    *  == Example ==
+    *  {{{
+    *  import chisel3._
+    *  import chisel3.util.addAttribute
+    *  class AttributeExample extends Module {
+    *    val io = IO(new Bundle{
+    *      val input = Input(UInt(8))
+    *      val output = Output(UInt(8))
+    *
+    *      val reg = RegNext(io.input)
+    *
+    *      addAttribute(reg, "synthesis translate_off")
+    *
+    *      io.output := reg
+    *    }
+    *  }
+    *
+    *  }}}
+    *
+    * @param target Chisel target. Must be Reg, Wire, or RawModule type.
+    * @param annoString attribute string to add to target.
+    */
+  def apply(target: Data, annoString: String): Unit = {
     requireIsAnnotatable(target, "target must be annotatable")
     annotate(new ChiselAnnotation {
       def toFirrtl = AttributeAnnotation(target.toNamed, annoString)
--- a/src/main/scala/chisel3/util/AttributeAnnotation.scala
+++ b/src/main/scala/chisel3/util/AttributeAnnotation.scala
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.utils
+
+import firrtl.AttributeAnnotation
+import firrtl.annotations.Named
+
+import chisel3.Data
+import chisel3.experimental.{annotate, requireIsAnnotatable, ChiselAnnotation}
+
+object annotateAttribute {
+  def apply[T <: Data](target: T, annoString: String): Unit = {
+    requireIsAnnotatable(target, "target must be annotatable")
+    annotate(new ChiselAnnotation {
+      def toFirrtl = AttributeAnnotation(target.toNamed, annoString)
+    })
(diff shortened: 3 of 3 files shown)


firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import firrtl.annotations.{Named, SingleTargetAnnotation}
+
+/** Firrtl implementation for verilog attributes
+  * @param target       target component to tag with attribute
+  * @param description  Attribute string to add to target
+  */
+case class AttributeAnnotation(target: Named, description: String) extends SingleTargetAnnotation[Named] { 
+  def duplicate(n: Named): AttributeAnnotation = this.copy(target = n, description = description)
+}
```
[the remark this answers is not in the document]
lordspacehog: Open to suggestions on this. It seems like it should live in chisel3 or circt and not the firrtl tree?

Also unclear if i should try to have the chisel implementation detect invalid annotation targets and error at compile time, any idea how to constrain this to only the supported types for now? Module, reg, wire etc seem pretty easy, however im not sure what "node" maps to, is that "FirrtlNode" in scala?

firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
```diff
@@ -0,0 +1,11 @@
+package firrtl
```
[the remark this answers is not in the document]
lordspacehog: updated, and should have fixed the formatting issue as well.

firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import firrtl.annotations.{Named, SingleTargetAnnotation}
+
+/** Firrtl implementation for verilog attributes
+  * @param target       target component to tag with attribute
+  * @param description  Attribute string to add to target
+  */
+case class AttributeAnnotation(target: Named, description: String) extends SingleTargetAnnotation[Named] {
+  def duplicate(n: Named): AttributeAnnotation = this.copy(target = n, description = description)
```
sequencer, line 11: Q to @seldridge, do we still need to implement `duplicate`? this might be an old API from SFC world?

seldridge: You unfortunately, do as this is an abstract member. However, this method will never be used. We should remove this abstract member.

firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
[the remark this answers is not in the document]
seldridge: Probably in the `chisel` package is better.

I wouldn't constrain it here. You could try to constrain it in the Chisel utility via `DataMirror` methods. However, I'm not aware of other similar utilities trying to do this.

firrtl/src/main/scala/firrtl/AttributeAnnotation.scala
jackkoenig: There is still 1 case that makes this necessary. Because views (from DataView) may be a single data that targets multiple unrelated Data, we do a special renaming at the end of Chisel elaboration which uses this duplicate method. Cleaning this up requires changes to the annotation API--easiest if we ban user-defined annotations or otherwise we have to change the API for how the user creates annotations from Chisel (forcing them to contemplate views).

sequencer: > easiest if we ban user-defined annotations or otherwise we have to change the API for how the user creates annotations from Chisel (forcing them to contemplate views).

I think, in the future plan we should remove the user-defined annotations, and maybe use `binding` for views?

firrtl/src/test/scala/firrtlTests/annotationTests/AttributeAnnotationSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+  "AttributeAnnotation should be correctly parsed from a string" in {
+    val attribAnno = new AttributeAnnotation(
+      ComponentName("attrib", ModuleName("ModuleAttrib", CircuitName("CircuitAttrib"))),
+      "X_INTERFACE_INFO = \"some:interface:type:1.0 SIGNAL\""
+    )
+
+    val annoString = JsonProtocol.serializeTry(Seq(attribAnno)).get
+    val loadedAnnos = JsonProtocol.deserializeTry(annoString).get
+    attribAnno should equal(loadedAnnos.head)
+  }
```
jackkoenig, line 20: What is the current status of support in firtool? Does it support attribute annotations on some things but not all? It would be good to have a test through to Verilog here for something that firtool supports

src/main/scala/chisel3/util/AttributeAnnotation.scala
```diff
@@ -0,0 +1,18 @@
...
+import chisel3.experimental.{annotate, requireIsAnnotatable, ChiselAnnotation}
+
+object annotateAttribute {
+  def apply[T <: Data](target: T, annoString: String): Unit = {
+    requireIsAnnotatable(target, "target must be annotatable")
+    annotate(new ChiselAnnotation {
+      def toFirrtl = AttributeAnnotation(target.toNamed, annoString)
+    })
+  }
+}
```
jackkoenig, line 18: This needs ScalaDoc with a simple example. I also think we should have a name that doesn't refer to annotations as that's an implementation detail that will hopefully change. Not sure what we should call it `addAttribute`? Any inspiration?

firrtl/src/test/scala/firrtlTests/annotationTests/AttributeAnnotationSpec.scala
seldridge: It works on declarations (wires, registers), but not on ports.

kivikakk:
Hiya! I might be getting ahead of myself or this PR, but I'm wondering how this might be adapted for modules\*. From what I could tell, HEAD firtool does support them, going off [LowerAnnotations.cpp](https://github.com/llvm/circt/blob/139c97bd90ec3b0ffa4fbfc98d7bb5195f34d50f/lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp#L330-L333) and this [test case (?)](https://github.com/llvm/circt/blob/139c97bd90ec3b0ffa4fbfc98d7bb5195f34d50f/test/firtool/sv-attr.fir#L3)), but I'll admit I'm very late to this party and I'm just seeing what I can see.

I tried importing this code into my own project, swapped out the `T <: Data` bound for `T <: BaseModule`, commented out the `requireIsAnnotatable` check (since that also needs `T <: Data`), and it does indeed work end-to-end. We can't overload `apply` with two different generic bounds, so it might take a separate `addAttributeModule` object or something.

\* I actually want it for blackboxes, but that looks like it'll take a trip to CIRCT too so I figure I'll go one step at a time.

jackkoenig:
If it works end-to-end, I see no reason why we shouldn't just enable it for Modules as well. If it doesn't work for BlackBoxes yet, maybe we should overload with `RawModule` (until BlackBoxes are also supported). Note that we can't overload if we use generics, but there's no real reason to use a generic here (since the function returns `Unit`), we can just define for `Data` and `RawModule`.

src/main/scala/chisel3/util/AttributeAnnotation.scala
```diff
@@ -0,0 +1,18 @@
...
+package chisel3.utils
+
+import firrtl.AttributeAnnotation
+import firrtl.annotations.Named
+
+import chisel3.Data
+import chisel3.experimental.{annotate, requireIsAnnotatable, ChiselAnnotation}
+
+object annotateAttribute {
+  def apply[T <: Data](target: T, annoString: String): Unit = {
```
jackkoenig, line 12:
```suggestion
  def apply(target: Data, annoString: String): Unit = {
```
I don't think we get any benefit from the generic here and it would prevent us from overloading this method for other types (e.g. `RawModule`). You only really need a generic if it somehow affects other arguments (e.g. due to context bounds) or the return type.

lordspacehog: Makes sense, I'll update this later today and test locally. I ended up getting sidetracked with a bunch of business side things and haven't had a chance to do much technical work over the last few weeks.

src/main/scala/chisel3/util/AttributeAnnotation.scala
lordspacehog: I'll make sure to capture the list of what CIRCT can process in the docs as well.

src/main/scala/chisel3/util/AttributeAnnotation.scala
jackkoenig: Thanks! Was not trying to prod you or anything, was just leaving a comment while I was thinking about it.

lordspacehog: all good! was going to come back to this soon anyway! looks like im not the only one hitting this too, which def gives me energy to get it moving again, haha.

lordspacehog: currently CIRCT only supports wire, node, reg, and module. should i just go ahead and add overloaded versions of this for those 4 types?

jackkoenig (approved):

PR #4028, opened 2024-04-25
Subdivide OM primitive values into Scala types

SpriteOvO:
Subdivide OM primitive values into Scala types

### DIFF
--- a/panamaom/src/PanamaCIRCTOM.scala
+++ b/panamaom/src/PanamaCIRCTOM.scala
@@ -62,7 +62,7 @@ object PanamaCIRCTOMEvaluatorValue {
     } else if (circt.omEvaluatorValueIsAReference(value)) {
       newValue(circt, circt.omEvaluatorValueGetReferenceValue(value))
     } else if (circt.omEvaluatorValueIsAPrimitive(value)) {
-      new PanamaCIRCTOMEvaluatorValuePrimitive(circt, value)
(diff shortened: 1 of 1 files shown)


sequencer:
please add test

lit/tests/Property/Good.sc
```diff
@@ -76,15 +76,24 @@ args.head match {
...
     converter.foreachHwModule(name => println(s"module{$name}"))
+
+    val aInt = obj.field("a")
+      .asInstanceOf[PanamaCIRCTOMEvaluatorValueList]
+      .getElement(0)
+      .asInstanceOf[PanamaCIRCTOMEvaluatorValueList]
+      .getElement(0)
+      .asInstanceOf[PanamaCIRCTOMEvaluatorValuePrimitiveInteger]
+      .integer
+    assert(123 == aInt)
```
sequencer, line 96: It's a dilemma to get these type since Scala doesn't know the real type.
In the following PR, we may need to think about how to safely get these Scala type.
think about provide these API to end users in the future:
https://github.com/com-lihaoyi/upickle/blob/3845e9ef57851075756b3d32efe3898e2dc13391/ujson/src/ujson/Value.scala#L17-L89

sequencer (approved):
It's still an unsafe design API, but look better.
In the following PRs, please decouple the chisel omreader out from chisel as a standalong project to make it be able to be standalone release.
In a near future(before formally release), we should figure out a clean user-level API like `value.strOpt: Option[String]`.

PR #4032, opened 2024-04-25
Improve source locators for intrinsic helpers.

dtzSiFive:
Better to point to where the intrinsic was used,
not where the wrapper is defined.

### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

#### Release Notes

Use source information where intrinsic wrapper is used, instead of location of the wrapper's implementation.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/circt/PlusArgsValue.scala
+++ b/src/main/scala/chisel3/util/circt/PlusArgsValue.scala
@@ -5,6 +5,7 @@ package chisel3.util.circt
 import scala.language.reflectiveCalls
 
 import chisel3._
+import chisel3.experimental.SourceInfo
 import chisel3.internal.Builder
 
 /** Create an intrinsic which generates a verilog \$value\$plusargs.  This returns a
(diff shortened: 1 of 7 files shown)


src/main/scala/chisel3/util/circt/PlusArgsValue.scala
```diff
@@ -21,7 +22,7 @@ object PlusArgsValue {
     * b.value
     * }}}
     */
-  def apply[T <: Data](gen: T, str: String) = {
+  def apply[T <: Data](gen: T, str: String)(implicit sourceInfo: SourceInfo) = {
```
jackkoenig, line 25: I hate to have to point out (because the fix is a little annoying), but this is a source incompatible change. If a user were doing `PlusArgsValue(UInt(8.W), "whatever")(7)` to take just the 7th bit of the value, the code will no longer compile. This is the whole `do_apply` pattern. Naturally this is fixed in Scala 3, at the moment we have to do the macro.

dtzSiFive: I'll take a look at that, thanks, that totally makes sense.

PR #4036, opened 2024-05-01
Add private Module API and internal DataMirror API for moduleIOs.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

The DataMirror API allows users who know what they're doing to access a module's ports before it is closed.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -577,16 +577,7 @@ package experimental {
     /** Get IOs that are currently bound to this module.
       */
     private[chisel3] def getIOs: Seq[Data] = {
-      _ids.flatMap { id =>
-        id match {
-          case (data: Data) if data.isSynthesizable =>
-            data.topBinding match {
-              case PortBinding(_) => Some(data)
-              case _              => None
-            }
-          case _ => None
-        }
-      }.toSeq
+      _ports.map(_._1).toSeq
     }
 
     // These methods allow checking some properties of ports before the module is closed,
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -574,6 +574,21 @@ package experimental {
       _ports.toSeq
     }
 
+    /** Get IOs that are currently bound to this module.
+      */
+    private[chisel3] def getIOs: Seq[Data] = {
+      _ids.flatMap { id =>
+        id match {
+          case (data: Data) if data.isSynthesizable =>
+            data.topBinding match {
+              case PortBinding(_) => Some(data)
+              case _              => None
+            }
+          case _ => None
+        }
+      }.toSeq
+    }
+
     // These methods allow checking some properties of ports before the module is closed,
     // mainly for compatibility purposes.
     protected def portsContains(elem: Data): Boolean = {
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -213,15 +213,16 @@ object DataMirror {
       target.cloneTypeFull
     }
 
-    /** Returns the IOs of a module.
+    /** Returns the current ports of an in-progress module.
       *
       * This method does not necessarily return the final ports of the target module. It consults Chisel's internal data
-      * structures to extract the module's IOs. For this reason, it is generally safer to prefer [[modulePorts]], but
-      * this method may be used for certain use cases that want the current list of ports before the module is closed.
+      * structures to extract the module's IOs. For this reason, it is generally not safe, and users should prefer
+      * [[DataMirror.modulePorts]], but this method may be used for certain use cases that want the current list of
+      * ports before the module is closed.
       *
       * @param target BaseModule to get IOs from
       */
-    def moduleIOs(target: BaseModule): Seq[Data] = target.getIOs
+    def currentModulePorts(target: BaseModule): Seq[Data] = target.getIOs
   }
 
   // Old definition of collectLeafMembers
--- a/core/src/main/scala/chisel3/reflect/DataMirror.scala
+++ b/core/src/main/scala/chisel3/reflect/DataMirror.scala
@@ -212,6 +212,16 @@ object DataMirror {
     def chiselTypeClone[T <: Data](target: T): T = {
       target.cloneTypeFull
     }
+
+    /** Returns the IOs of a module.
+      *
+      * This method does not necessarily return the final ports of the target module. It consults Chisel's internal data
+      * structures to extract the module's IOs. For this reason, it is generally safer to prefer [[modulePorts]], but
+      * this method may be used for certain use cases that want the current list of ports before the module is closed.
+      *
+      * @param target BaseModule to get IOs from
+      */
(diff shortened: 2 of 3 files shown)


core/src/main/scala/chisel3/reflect/DataMirror.scala
```diff
@@ -212,6 +212,16 @@ object DataMirror {
...
+
+    /** Returns the IOs of a module.
+      *
+      * This method does not necessarily return the final ports of the target module. It consults Chisel's internal data
+      * structures to extract the module's IOs. For this reason, it is generally safer to prefer [[modulePorts]], but
+      * this method may be used for certain use cases that want the current list of ports before the module is closed.
+      *
+      * @param target BaseModule to get IOs from
+      */
+    def moduleIOs(target: BaseModule): Seq[Data] = target.getIOs
```
sequencer, line 224: Maybe constraint target to be `FixedIOModule`? I'm afraid in some complex design pattern, this function will execute to early.

mwachs5, line 224: can we call this something like `currentModuleIOs` or `currentModulePorts`? Are we trying to convey a difference between the Ports and IOs returned between the two APIs? Do we not wnat to return the String name yet because that is not finalized...?

src/test/scala/chiselTests/reflect/DataMirrorSpec.scala
```diff
@@ -245,4 +245,21 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
+      val child = Module(new RawModule {})
+
+      val ports = DataMirror.internal.moduleIOs(this)
+    }
+
+    ChiselStage.emitCHIRRTL(new RawModule {
+      val foo = Module(new Foo)
+      foo.ports.size should be(2)
+      foo.ports(0).toNamed.name should be("in")
+      foo.ports(1).toNamed.name should be("out")
```
mwachs5, line 262: can we show a test what happens if you then add more ports (e.g by boring), or right after the `val ports = ...` call?

core/src/main/scala/chisel3/reflect/DataMirror.scala
mikeurbach: For the use case I had in mind, I don't want to add that constraint. Indeed, this is not generally safe, which I tried to explain in the scaladoc and by putting this into `DataMirror.internal` object.

src/test/scala/chiselTests/reflect/DataMirrorSpec.scala
```diff
@@ -245,4 +245,21 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
 
+  "moduleIOs" should "return an in-progress module's IOs" in {
+    class Foo extends RawModule {
+      val in = IO(Input(Bool()))
+      val out = IO(Output(Bool()))
+      val wire = Wire(Bool())
+      val child = Module(new RawModule {})
+
+      val ports = DataMirror.internal.moduleIOs(this)
+    }
```
mwachs5, line 256:
```suggestion

      val ports = DataMirror.internal.moduleIOs(this)
      
      val whatHappensHere = IO(Input(Bool())
      val andWhatHappensHere = DataMirror.internal.moduleIOs(this)
    }
```

core/src/main/scala/chisel3/reflect/DataMirror.scala
mwachs5: we need this for more than FixedIOModule.

mikeurbach: Yeah I wasn't sure what to call this, I think currentModulePorts is a good name, so we have `DataMirror.modulePorts` and `DataMirror.internal.currentModulePorts`.

I didn't include the string names because I didn't need them, but I can add that for symmetry.

mikeurbach: Oh, actually, getting the names is a little tricky. It is possible to get the names in atModuleBodyEnd, but only for IOs that were added before the atModuleBodyEnd. I could add the names but users will see exceptions if they call this too early. Since I didn't need them, I am inclined to no add them, and that way this API should always succeed.

sequencer: I see, this sounds to be an advance unsafe API, wish normal user won't touch it;p

sequencer: Maybe we can try to add a hook in builder, if user called this function and new IO is created, log an warning to user saying the IO they got is not enough.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -574,6 +574,21 @@ package experimental {
...
+      _ids.flatMap { id =>
+        id match {
+          case (data: Data) if data.isSynthesizable =>
+            data.topBinding match {
+              case PortBinding(_) => Some(data)
+              case _              => None
+            }
+          case _ => None
+        }
+      }.toSeq
```
jackkoenig, line 589:
```suggestion
      _ports.toSeq
```
You can just use `this._ports`!

src/test/scala/chiselTests/reflect/DataMirrorSpec.scala
```diff
@@ -245,4 +245,31 @@ class DataMirrorSpec extends ChiselFlatSpec {
...
+
+      val ports1 = DataMirror.internal.currentModulePorts(this)
+    }
+
+    ChiselStage.emitCHIRRTL(new RawModule {
+      val foo = Module(new Foo)
+
+      foo.ports0.size should be(2)
+      foo.ports0(0).toNamed.name should be("in")
+      foo.ports0(1).toNamed.name should be("out")
```
jackkoenig, line 267:
```suggestion
      foo.ports0(0).toNamed.name should be("in")
      foo.ports0(1).toNamed.name should be("out")
```
Could this just be `foo.ports0 should be(Seq(foo.in, foo.out))`?

jackkoenig (approved):
LGTM, consider tweaking the test but merge when you're ready.

core/src/main/scala/chisel3/reflect/DataMirror.scala
mikeurbach: I will leave the names off for now.

mikeurbach: I think it's fine to start with the new API "hidden" in `DataMirror.internal`, with a comment explaining the dangers. If users start using this and continue to be surprised by the documented behavior, I guess we could add the extra state to give warnings.

src/test/scala/chiselTests/reflect/DataMirrorSpec.scala
```diff
@@ -245,4 +245,24 @@ class DataMirrorSpec extends ChiselFlatSpec {
     DataMirror.getLayerColor(foo.c) should be(Some(A))
   }
 
+  "moduleIOs" should "return an in-progress module's IOs" in {
```
mwachs5, line 248: super nit you changed the name of this function and maybe we should have changed this string

mikeurbach: thanks, i wish we could push NFC changes to main.

PR #4053, opened 2024-05-03
[LTL] Port LTL intrinsics to use IntrinsicExpr instead of IntrinsicModule

dobios:
**[WIP]**  
This PR is replaces all Intrinsic module defined LTL intrinsics with `IntrinsicExpr` defined ones.  

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
+++ b/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
@@ -5,13 +5,14 @@ package chisel3.util.circt
 import chisel3._
 import chisel3.experimental.{IntParam, IntrinsicModule, Param, StringParam}
 import chisel3.experimental.hierarchy.{instantiable, public}
+import chisel3.experimental.SourceInfo
 
-private object Utils {
+/*private object Utils {
   private[chisel3] def withoutNone(params: Map[String, Option[Param]]): Map[String, Param] =
     params.collect { case (name, Some(param)) => (name, param) }
-}
+}*/
 
-@instantiable
+/*@instantiable
 private[chisel3] class BaseIntrinsic(
   intrinsicName: String,
   maybeParams:   Map[String, Option[Param]] = Map.empty[String, Option[Param]])
@@ -22,9 +23,26 @@ private[chisel3] class BaseIntrinsic(
     case x              => x.toString
   })
   override def desiredName = paramValues.fold(this.getClass.getSimpleName) { (a, b) => a + "_" + b }
+}*/
+
+/** Base instrinsic for circt related intrinsics
+  * @param intrinsicName name of the intrinsic
+  * @param ret return type of the expression
+  * @param params parameter name/value pairs, if any.  Parameter names must be unique.
+  * @param sourceInfo where the data is from
+  * @return intrinsic expression that returns the specified return type
+  */
+object BaseIntrinsic {
+  def apply[T <: Data](
+    intrinsicName: String,
+    ret: => T,
+    maybeParams:   Seq[(String, Param)] = Seq()
+  )(implicit sourceInfo: SourceInfo) = {
+    IntrinsicExpr(f"circt_$intrinsicName", ret, maybeParams:_*)()
+  }
 }
 
-/** Base class for all unary intrinsics with `in` and `out` ports. */
+/*/** Base class for all unary intrinsics with `in` and `out` ports. */
 @instantiable
 private[chisel3] class UnaryLTLIntrinsic(
   intrinsicName: String,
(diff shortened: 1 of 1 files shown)


dtzSiFive:
Awesome, thanks!!

> API modification

I /think/ (?) this is more "Backend Code Generation" since it changes the emitted FIRRTL but not the user-facing API or what that accomplishes?

dtzSiFive (approved):
LGTM!!

We could thread sourceInfo on all the `LTL.scala` methods so that the source locators point to user-code, but that might be a breaking change (see #4032).  It might not since I don't think you can use parens on the returned type (?).

I wonder why the nodes all have `_T_*` names -- usually they get named fine even when wrapped, like here: https://github.com/chipsalliance/chisel/blob/69ec7c2b54a911e3798ad692590fa280d32f42e3/src/test/scala/chiselTests/util/circt/PlusArgsValueSpec.scala#L34 .

Maybe because the return value is a Sequence, not a Data?

dtzSiFive:
> I wonder why the nodes all have `_T_*` names -- usually they get named fine even when wrapped, like here:

This little patch gives them names after the intrinsic they're calling, but doesn't use `val` name from user.

```patch
From e4b2f958d89f7e9815107ab9c704c38ce0e3c114 Mon Sep 17 00:00:00 2001
From: Will Dietz <will.dietz@sifive.com>
Date: Tue, 7 May 2024 09:04:11 -0500
Subject: [PATCH] LTLIntrinsics: suggestName?

Also use BinaryLTLIntrinsic for delay/clock too.
---
 .../chisel3/util/circt/LTLIntrinsics.scala    | 22 +++++++++----------
 1 file changed, 11 insertions(+), 11 deletions(-)

diff --git a/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala b/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
index a04ccbf73..898283b6b 100644
--- a/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
+++ b/src/main/scala/chisel3/util/circt/LTLIntrinsics.scala
@@ -30,27 +30,27 @@ private[chisel3] object BaseIntrinsic {

 /** Base intrinsic for all unary intrinsics with `in` and `out` ports. */
 private[chisel3] object UnaryLTLIntrinsic {
-  def apply(
+  def apply[T <: Data](
     intrinsicName: String,
     params:        Seq[(String, Param)] = Seq()
-  )(_in:           Bool
+  )(_in:           T
   )(
   ): Bool =
-    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(_in)
+    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(_in).suggestName(intrinsicName)
 }

 /** Base instrinsic for all binary intrinsics with `lhs`, `rhs`, and `out` ports. */
 private[chisel3] object BinaryLTLIntrinsic {
-  def apply(
+  def apply[T <: Data, U <: Data](
     intrinsicName: String,
     params:        Seq[(String, Param)] = Seq()
-  )(lhs:           Bool,
-    rhs:           Bool
+  )(lhs:           T,
+    rhs:           U
   )(
   ): Bool =
-    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs)
+    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs).suggestName(intrinsicName)
 }

 /** A wrapper intrinsic for the CIRCT `ltl.and` operation. */
@@ -104,13 +104,13 @@ private[chisel3] object LTLEventuallyIntrinsic {
 /** A wrapper intrinsic for the CIRCT `ltl.clock` operation. */
 private[chisel3] object LTLClockIntrinsic {
   def apply(in: Bool, clock: Clock)(implicit sourceInfo: SourceInfo) =
-    BaseIntrinsic("ltl_clock", Bool())(in, clock)
+    BinaryLTLIntrinsic("clock")(in, clock)
 }

 /** A wrapper intrinsic for the CIRCT `ltl.disable` operation. */
 private[chisel3] object LTLDisableIntrinsic {
   def apply(in: Bool, cond: Bool)(implicit sourceInfo: SourceInfo) =
-    BaseIntrinsic("ltl_disable", Bool())(in, cond)
+    BinaryLTLIntrinsic("disable")(in, cond)
 }

 /** Base class for assert, assume, and cover intrinsics. */
@@ -119,8 +119,8 @@ private[chisel3] object VerifAssertLikeIntrinsic {
     val name = f"verif_$intrinsicName"

     label match {
-      case None    => BaseIntrinsic(name, Bool())(prop)
-      case Some(l) => BaseIntrinsic(name, Bool(), Seq("label" -> StringParam(l)))(prop)
+      case None    => BaseIntrinsic(name, Bool())(prop).suggestName(intrinsicName)
+      case Some(l) => BaseIntrinsic(name, Bool(), Seq("label" -> StringParam(l)))(prop).suggestName(intrinsicName)
     }
   }
 }
--
2.43.2
```

For example:
```firrtl
    node assert = intrinsic(circt_verif_assert : UInt<1>, a) @[src/main/scala/chisel3/ltl/LTL.scala 372:68]
```

Which seems a bit nicer (?), but means these always have "interesting" names even when not named by a user.
Leaving here in case it's something we'd like.

cc @jackkoenig any insights/suggestions re:naming story here? Thanks!

dobios:
Thanks for the review @dtzSiFive ! I updated the naming like you suggested.

jackkoenig:
Better naming would be to make Properties hook into the compiler plugin, https://github.com/chipsalliance/chisel/blob/69ec7c2b54a911e3798ad692590fa280d32f42e3/plugin/src/main/scala/chisel3/internal/plugin/ChiselComponent.scala#L90, but we can do that in a follow on PR.

src/main/scala/chisel3/ltl/LTL.scala
```diff
@@ -412,7 +364,7 @@ sealed abstract class AssertPropertyLike {
     apply(Sequence.BoolSequence(cond), Some(clock), Some(disable), Some(label))
   }
 
-  def createIntrinsic(label: Option[String]): Instance[VerifAssertLikeIntrinsic]
+  def createIntrinsic(label: Option[String]): (Bool) => Unit
```
jackkoenig, line 367: We should make these protected or package private and backport a deprecation (since we've made a source incompatible change anyway).

PR #4056, opened 2024-05-06
OM evaluator instantiate method returns Option

SpriteOvO:
If the instantiation is failed, method returns `None` rather than assertion failure.

CC @sequencer


sequencer (approved):

sequencer:
I suspect the `Option` return can be used for most of evaluating functions.

PR #4067, opened 2024-05-14
Add .toRelativeTargetToHierarchy

mwachs5:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Add .toRelativeTargetToHierarchy for getting .toRelativeTarget functionality when the root is a Definition or Instance.

- [x] Did you mark the proper milestone (Bug fix: `3.6.x`, `5.x`, or `6.x` depending on impact, API modification or big change: `7.0`)?
- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -762,6 +762,9 @@ package experimental {
         s"No common ancestor between\n  ${this.toAbsoluteTarget} and\n  ${root.get.toAbsoluteTarget}"
       )
 
+      // Algorithm starts from the top of both absolute paths
+      // and walks down them checking for equality,
+      // and terminates once the root is just a ModuleTarget
       def recurse(thisRelative: IsModule, rootRelative: IsModule): IsModule = {
         (thisRelative, rootRelative) match {
           case (t: IsModule, r: ModuleTarget) => {
@@ -776,14 +779,13 @@ package experimental {
         }
       }
 
-      // If root was defined, and we are it, return this.
       if (root.isEmpty) (this.toAbsoluteTarget)
-      // Not yet handling DataView
+      else if (this == ViewParent) ViewParent.absoluteTarget
       else {
-        var thisRelative = this.toAbsoluteTarget
-        var rootRelative = root.get.toAbsoluteTarget
-        if (thisRelative.circuit != rootRelative.circuit) fail()
-        recurse(thisRelative, rootRelative)
+        val thisAbsolute = this.toAbsoluteTarget
+        val rootAbsolute = root.get.toAbsoluteTarget
+        if (thisAbsolute.circuit != thisAbsolute.circuit) fail()
+        recurse(thisAbsolute, rootAbsolute)
       }
     }
 
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -17,6 +17,7 @@ import _root_.firrtl.annotations.{IsModule, ModuleName, ModuleTarget}
 import _root_.firrtl.AnnotationSeq
 import chisel3.internal.plugin.autoNameRecursively
 import chisel3.util.simpleClassName
+import chisel3.experimental.hierarchy.Hierarchy
 
 object Module extends SourceInfoDoc {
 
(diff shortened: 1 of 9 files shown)


core/src/main/scala/chisel3/Module.scala
```diff
@@ -738,12 +739,47 @@ package experimental {
...
+      *
+      * If `root` is not defined, the target is a hierarchical path equivalent to `toAbsoluteTarget`.
+      *
+      * @note If `root` is defined, and has not finished elaboration, this must be called within `atModuleBodyEnd`.
+      * @note The BaseModule must be a descendant of `root`, if it is defined.
+      * @note This doesn't have special handling for Views.
+      */
+    final def toRelativeTargetToHierarchy(root: Option[Hierarchy[BaseModule]]): IsModule = {
+      // If root was defined, and we are it, return this.
+      if (root.map(_.proto).contains(this)) getTarget
```
mwachs5, line 762: @azidar this is the check that should work that is failing. Is there a correct incantation here...?

src/test/scala/chiselTests/ToTargetSpec.scala
```diff
@@ -183,4 +214,17 @@ class ToTargetSpec extends ChiselFlatSpec with Utils {
 
     (e.getMessage should include).regex("Requested .toRelativeTarget relative to .+, but it is not an ancestor")
   }
+
+  it should "work to get relative targets to an instance of an Instance" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new RelativeSiblingsInstancesParent)
+
+    chirrtl should include("~RelativeSiblingsInstancesParent|Foo>inner")
```
mwachs5, line 221: not the right check yet, but not even getting to the point of chirrtl emission yet

core/src/main/scala/chisel3/Module.scala
mwachs5: > theoretically, we should be able to resolve this given the actual target (e.g. the IsModule) datastructure, rather than the hierarchy datastructure.  convert both the root and the this to absoluteTargets, and then inspect those two things to figure out how to represent the relative path. 

Maybe need some sort of helper on IsModule.relativeTo(another: IsModule)

mwachs5: todo: see Target.scala in firrtl, use moduleOpt and striphierarchy to cobble something together here

azidar (approved):
LGTM!

core/src/main/scala/chisel3/Module.scala
```diff
@@ -738,12 +739,56 @@ package experimental {
...
 
+    /** Returns a FIRRTL ModuleTarget that references this object, relative to an optional root.
+      *
+      * If `root` is defined, the target is a hierarchical path starting from `root`.
+      *
+      * If `root` is not defined, the target is a hierarchical path equivalent to `toAbsoluteTarget`.
+      *
+      * @note If `root` is defined, and has not finished elaboration, this must be called within `atModuleBodyEnd`.
+      * @note The BaseModule must be a descendant of `root`, if it is defined.
+      * @note This doesn't have special handling for Views.
```
jackkoenig, line 758: What does this comment mean? Since this is only involving Modules, views shouldn't be relevant (but I might be missing something).

jackkoenig: Oh I see the logic about `ViewParent`, but I think that should never come up since that's only relevant as the parent of views (which are `Data`).

jackkoenig: I see how that this is [obviously] also implemented for Data, just the same Scaladoc, makes sense!

core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
```diff
@@ -78,6 +78,18 @@ object Definition extends SourceInfoDoc {
       * @return absoluteTarget of this instance
       */
     def toAbsoluteTarget: IsModule = d.proto.toAbsoluteTarget
+
+    /** If this is an instance of a Module, returns the toAbsoluteTarget of this instance
+      * @return absoluteTarget of this instance
+      */
+    def toRelativeTarget(root: Option[BaseModule]): IsModule = d.proto.toRelativeTarget(root)
+
+    /** If this is an instance of a Module, returns the toAbsoluteTarget of this instance
```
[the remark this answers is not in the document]
mikeurbach: I think these comments have been here since before I started contributing to Chisel, but maybe we can at least update the new ones?

mikeurbach (approved):
I think this makes sense, definitely seems better than the hacky thing I was trying as evidenced by the various positive/negative test scenarios.

core/src/main/scala/chisel3/experimental/hierarchy/core/Definition.scala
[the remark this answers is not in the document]
mwachs5: i will clean these up more comprehenstively in a follow-up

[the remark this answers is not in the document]
mwachs5: oh wait I added these comments 😂 sorry for the noise.

PR #4070, opened 2024-05-17
Add support for Vec literals of empty Vecs

jackkoenig:
Also tweak Vec.Lit code to use cloneAsSuperType rather than implementing its own version of the same concept.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -489,7 +489,6 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
             s"are less than zero or greater or equal to than Vec length"
         )
       }
-      cloneSupertype(elementInitializers.map(_._2), s"Vec.Lit(...)")
 
       // look for literals of this vec that are wider than the vec's type
       val badLits = elementInitializers.flatMap {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -489,7 +489,6 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
             s"are less than zero or greater or equal to than Vec length"
         )
       }
-      cloneSupertype(elementInitializers.map(_._2), s"Vec.Lit(...)")
```
jackkoenig, line 492: This was being used as a check, but it is what was blocking size-0 from working in the `Vec(0, ...).Lit` use case. Rather, that check is only needed when using the `Vec.Lit(...)` API (since no element type is given) so I basically moved this check there but also went ahead and used it to figure out the resulting `Vec` type.

src/test/scala/chiselTests/BundleLiteralSpec.scala
```diff
@@ -360,4 +360,11 @@ class BundleLiteralSpec extends ChiselFlatSpec with Utils {
...
     (chirrtl should include).regex(wire)
   }
+
+  "Empty bundle literals" should "be supported" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new RawModule {
+      val lit = (new Bundle {}).Lit()
+      lit.litOption should equal(Some(0))
+    })
+  }
 }
```
jackkoenig, line 370: I also added this test, but it already worked correctly.

mwachs5 (approved):

PR #4078, opened 2024-05-22
Remove line contents from assertion messages

jackkoenig:
I'm waffling on whether to call this an API Modification or a Backend code generation. It sort of toes that line. I decided to keep the `Assertion failed:` prefix (at least for now) even when there is a user-provided message just because I know that that String is often used for determining when an assertion fires. Removing that is something we should probably do (when the user provides a String), but I'll leave that as a separate change that should be marked as an API modification.

This change feels very... late lol. This feels like the sort of thing in Chisel before 3.0 yet it has lingered for a very long time. Capturing the user's line has a lot of negative consequences, like including comments. It's also very ugly.

### Contributor Checklist

#### Type of Improvement

- Backend code generation

#### Desired Merge Strategy

- Squash

#### Release Notes

Also remove line and column when a message is provided.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -433,11 +433,11 @@ abstract class VerificationStatement extends NamedComponent {
 /** Helper functions for common functionality required by stop, assert, assume or cover */
 private object VerificationStatement {
 
-  type SourceLineInfo = (String, Int, String)
+  type SourceLineInfo = (String, Int)
 
   def getLine(c: blackbox.Context): SourceLineInfo = {
     val p = c.enclosingPosition
-    (p.source.file.name, p.line, p.lineContent.trim): @nowarn // suppress, there's no clear replacement
+    (p.source.file.name, p.line): @nowarn // suppress, there's no clear replacement
   }
 
   def formatFailureMessage(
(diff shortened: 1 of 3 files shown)


seldridge (commented):
Dropping `Assertion failed` would be nice. It is unfortunately part of the heuristic for matching Chisel asserts: https://github.com/llvm/circt/blob/da2ca8cc9f8a305a0bcc3d3ac61175402138babb/lib/Dialect/FIRRTL/Import/FIRParserAsserts.cpp#L65 😢

seldridge (approved):

dtzSiFive:
> Dropping `Assertion failed` would be nice. It is unfortunately part of the heuristic for matching Chisel asserts: https://github.com/llvm/circt/blob/da2ca8cc9f8a305a0bcc3d3ac61175402138babb/lib/Dialect/FIRRTL/Import/FIRParserAsserts.cpp#L65 😢

FWIW as of https://github.com/llvm/circt/pull/7030 this is only still around for the sole purpose of warning people using the printf-encoded format (error, actually).  So Chisel today gets nothing from including this message, since it doesn't use printf's.

jackkoenig:
Nice, thanks for that info @dtzSiFive! It's also used in Chisel's own tests for detecting test failures: https://github.com/chipsalliance/chisel/blob/7e0cd105339417e32baa3dfc0b0269725b4f2be2/src/test/scala/chiselTests/ChiselSpec.scala#L107, but that obviously should be cleaned up. I think we somehow need to figure out how to have SVSim detect an assertion failure (easy in Verilator, not sure how to do it in VCS, not sure if there's a DPI way to do this which would be preferred).

PR #4083, opened 2024-05-23
Fix boring tap of non-passive source from parent.

### LINKED ISSUE #3557: Boring from parent module with probe doesn't get a necessary :#= assignment
**Type of issue**: Bug Report

See the test in #3556. Even if fixing the issue reported there, the test will not pass, because as shown in the picture below, we implement the single black arrow not the dotted two-part arrow:
Screenshot 2023-09-26 at 4 50 21 PM

Note the bottom arrow in the picture: when we are goign straight from a bidirectional thing, and then turning it into an Input, we dont have any part of the code that does the stripping of direction and then using a `:#=` operator to drive the stripped-direction input to the bidirectional "real thing".

When there is a layer of hierarchy in the middle:
Screenshot 2023-09-27 at 11 05 13 AM

we have to first create the probe, which handles the stripping and the define works without the `#:=`.

We need to not emit a `connect` from a bidirectional thing to a monodirectional thing.

This can be done by either emitting the individual fields (as in the `:#=` would handle) or by making a probe define and then read in the same layer of hierarchy.

The former is a little tricky to implement because we can't directly use the `:#=` operator in the boring APIs, we can only add commands to the secretCommands, and there is no direct translation from `:#=` to the commands we would want to add to secretCommands.

It might be simpler to just always make the probe, define, and read even if we are in the same layer of heirarchy.

**Other Information**

Need to get code compiling. There is a possible workaround -- users can make the Wire(Output(...)) themself in the source of the tap.

dtzSiFive:
Expected result is fully aligned, which is what happens when reading from a probe.

When the source is already at the LCA (from parent) there's no probe and the secret connections only support a few commands.

For lack of a way to do, e.g., `:#=` here,
use `read(probe(x))` to get the fully aligned result that's expected and bored down to the child.

This isn't ideal but fixes this using only the sorts of expressions and commands that we've already committed to supporting.

Fixes #3557.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Fix tapping mix-alignment signal from parent.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/BoringUtilsTapSpec.scala
+++ b/src/test/scala/chiselTests/BoringUtilsTapSpec.scala
@@ -470,13 +470,7 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
 
     class Foo extends RawModule {
       val a = WireInit(DecoupledIO(Bool()), DontCare)
-      val dummyA = Wire(Output(chiselTypeOf(a)))
-      // FIXME we shouldn't need this intermediate wire
-      // https://github.com/chipsalliance/chisel/issues/3557
-      dummyA :#= a
-      dontTouch(a)
-
-      val bar = Module(new Bar(dummyA))
+      val bar = Module(new Bar(a))
     }
 
     val chirrtl = circt.stage.ChiselStage.emitCHIRRTL(new Foo, Array("--full-stacktrace"))
@@ -486,9 +480,7 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
       "input bore : { ready : UInt<1>, valid : UInt<1>, bits : UInt<1>}",
       "module Foo :",
       "wire a : { flip ready : UInt<1>, valid : UInt<1>, bits : UInt<1>}",
-      // FIXME shouldn't need intermediate wire
-      "wire dummyA : { ready : UInt<1>, valid : UInt<1>, bits : UInt<1>}",
-      "connect bar.bore, dummyA"
+      "connect bar.bore, read(probe(a))"
     )()
 
     // Check that firtool also passes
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -470,13 +470,7 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
     class Foo extends RawModule {
       val a = WireInit(DecoupledIO(Bool()), DontCare)
-      val dummyA = Wire(Output(chiselTypeOf(a)))
-      // FIXME we shouldn't need this intermediate wire
-      // https://github.com/chipsalliance/chisel/issues/3557
-      dummyA :#= a
-      dontTouch(a)
-
-      val bar = Module(new Bar(dummyA))
+      val bar = Module(new Bar(a))
```
jackkoenig, line 473: I'm a little confused by this test `Bar` is a child of `Foo` yet we're taping and reading a value from `Foo` in `Bar`?

dtzSiFive: Yep, that's what is happening!

"Tap" kinda has come to mean "run probes up, non-probes down" and this is just that second half.  And the issue here is specifically when this is ONLY the second half (never a probe) we expect the "fully aligned" but nothing makes that happen so emit a busted connect.  The input port created is fully aligned, but since we didn't (and shouldn't matter if we did) have to probe something we need a way to connect the two.  And the secret commands are limited, here I'm just taking all of that as fixed and working within what we've allowed there.

This is kinda the intuitive backing for the approach taken here, not that it matters really but .. now we read+probe (at least if that does anything) even if didn't have to run probes upwards or probe at all, which unsurprisingly is effective for providing uniform behavior.
(not that I don't wish/think we could do something better but also it's maybe less random that it appears, lol)

jackkoenig: Got it, thanks for the explanation!

jackkoenig (approved):

PR #4158, opened 2024-06-10
Add DPI intrinsics and API 

uenoku:
This PR adds DPI intrinsic support and API to define DPI functions declaratively. The test passes locally but requires a new version of firtool. 

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/circt/DPI.scala
+++ b/src/main/scala/chisel3/util/circt/DPI.scala
@@ -115,27 +115,27 @@ object RawClockedVoidFunctionCall {
 
 // A common trait for DPI functions.
 trait DPIFunctionImport {
-  val functionName: String
-  val inputNames: Option[Seq[String]] = None
+  def functionName: String
+  def inputNames: Option[Seq[String]] = None
 }
 
 // Base trait for a non-void function that returns `T`.
 trait DPINonVoidFunctionImport[T <: Data] extends DPIFunctionImport {
-  val ret:     T
-  val clocked: Boolean
-  val outputName: Option[String] = None
-  def callWithEnable(enable: Bool, data: Data*): T =
+  def ret:     T
+  def clocked: Boolean
+  def outputName: Option[String] = None
+  final def callWithEnable(enable: Bool, data: Data*): T =
     if (clocked) {
       RawClockedNonVoidFunctionCall(functionName, ret, inputNames, outputName)(Module.clock, enable, data: _*)
     } else {
       RawUnclockedNonVoidFunctionCall(functionName, ret, inputNames, outputName)(enable, data: _*)
     }
-  def call(data: Data*): T = callWithEnable(true.B, data: _*)
+  final def call(data: Data*): T = callWithEnable(true.B, data: _*)
 }
 
 // Base trait for a clocked void function.
 trait DPIClockedVoidFunctionImport extends DPIFunctionImport {
-  def callWithEnable(enable: Bool, data: Data*): Unit =
+  final def callWithEnable(enable: Bool, data: Data*): Unit =
     RawClockedVoidFunctionCall(functionName, inputNames)(Module.clock, enable, data: _*)
-  def call(data: Data*): Unit = callWithEnable(true.B, data: _*)
+  final def call(data: Data*): Unit = callWithEnable(true.B, data: _*)
 }
--- a/src/main/scala/chisel3/util/circt/DPI.scala
+++ b/src/main/scala/chisel3/util/circt/DPI.scala
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.util.circt.dpi
+
+import chisel3._
+object RawClockedNonVoidFunctionCall {
+
+  /** Creates an intrinsic that calls non-void DPI function at its clock posedge.
+    *
+    * @example {{{
+    * val a = RawClockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), clock, enable, b, c)
+    * }}}
+    */
+  def apply[T <: Data](functionName: String, ret: => T)(clock: Clock, enable: Bool, data: Data*): Data = {
+    IntrinsicExpr("circt_dpi_call", ret, "functionName" -> functionName, "isClocked" -> 1)(
+      (Seq(clock, enable) ++ data): _*
+    )
+  }
+}
+
+object RawUnlockedNonVoidFunctionCall {
+
+  /** Creates an intrinsic that calls non-void DPI function when input values are changed.
+    *
+    * @example {{{
+    * val a = RawUnlockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), enable, b, c)
+    * }}}
+    */
+  def apply[T <: Data](functionName: String, ret: => T)(enable: Bool, data: Data*): Data = {
+    IntrinsicExpr("circt_dpi_call", ret, "functionName" -> functionName, "isClocked" -> 0)(
+      (Seq(enable) ++ data): _*
+    )
+  }
+}
+
+object RawClockedVoidFunctionCall {
+
+  /** Creates an intrinsic that calls void DPI function at its clock posedge.
+    *
+    * @example {{{
+    * RawClockedVoidFunctionCall("dpi_func_foo", UInt(1.W), clock, enable, b, c)
+    * }}}
+    */
+  def apply(functionName: String)(clock: Clock, enable: Bool, data: Data*) = {
+    Intrinsic("circt_dpi_call", "functionName" -> functionName, "isClocked" -> 1)(
+      (Seq(clock, enable) ++ data): _*
+    )
+  }
+}
--- a/src/main/scala/chisel3/util/circt/DPI.scala
+++ b/src/main/scala/chisel3/util/circt/DPI.scala
@@ -3,31 +3,89 @@
 package chisel3.util.circt.dpi
 
 import chisel3._
+
+import chisel3.experimental.{IntParam, IntrinsicModule, Param, StringParam}
+
+private object GetDPIParams {
+  def apply(
+    functionName: String,
+    isClocked:    Boolean,
+    inputNames:   Option[Seq[String]],
+    outputName:   Option[String] = None
+  ): Seq[(String, Param)] = {
+    val inputNamesParam =
+      inputNames.map(_.mkString(";")).map(x => Seq("inputNames" -> StringParam(x))).getOrElse(Seq())
+    val outputNameParam = outputName.map(x => Seq("outputName" -> StringParam(x))).getOrElse(Seq())
+    Seq[(String, Param)](
+      "functionName" -> functionName,
+      "isClocked" -> (if (isClocked) 1 else 0)
+    ) ++ inputNamesParam ++ outputNameParam
+  }
+}
+
 object RawClockedNonVoidFunctionCall {
 
   /** Creates an intrinsic that calls non-void DPI function at its clock posedge.
+    * The result values behave like registers and the DPI function is used as a state
+    * transfer function of them.
     *
+    * `enable` operand is used to conditionally call the DPI since DPI call could be quite
+    * more expensive than native constructs.
+    *
+    * When an `enable` is false, it means the state transfer function is not called. Hence
+    * their values will not be modified in that clock.
+    *
+    * Please refer https://github.com/llvm/circt/blob/main/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md#dpi-intrinsic-abi for DPI function ABI.
     * @example {{{
     * val a = RawClockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), clock, enable, b, c)
     * }}}
     */
-  def apply[T <: Data](functionName: String, ret: => T)(clock: Clock, enable: Bool, data: Data*): Data = {
-    IntrinsicExpr("circt_dpi_call", ret, "functionName" -> functionName, "isClocked" -> 1)(
+  def apply[T <: Data](
+    functionName: String,
+    ret:          => T,
+    inputNames:   Option[Seq[String]] = None,
+    outputName:   Option[String] = None
+  )(clock:        Clock,
+    enable:       Bool,
+    data:         Data*
+  ): T = {
+    IntrinsicExpr(
+      "circt_dpi_call",
+      ret,
+      GetDPIParams(functionName, true, inputNames, outputName): _*
+    )(
       (Seq(clock, enable) ++ data): _*
     )
   }
 }
 
-object RawUnlockedNonVoidFunctionCall {
+object RawUnclockedNonVoidFunctionCall {
 
-  /** Creates an intrinsic that calls non-void DPI function when input values are changed.
+  /** Creates an intrinsic that calls non-void DPI function for its input value changes.
+    * The DPI call is considered as a combinational logic.
     *
+    * `enable` operand is used to conditionally call the DPI since DPI call could be quite
+    * more expensive than native constructs.
+    * When `enable` is false, results of unclocked calls are undefined and evaluated into X.
+    *
+    * Please refer https://github.com/llvm/circt/blob/main/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md#dpi-intrinsic-abi for DPI function ABI.
     * @example {{{
-    * val a = RawUnlockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), enable, b, c)
+    * val a = RawUnclockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), enable, b, c)
     * }}}
     */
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  val dpiImpl = s"""
+                   |#include <stdint.h>
+                   |#include <iostream>
+                   |
+                   |extern "C" void hello()
+                   |{
+                   |    std::cout << "hello from c++\\n";
+                   |}
+                   |
+                   |extern "C" void add(int lhs, int rhs, int* result)
```
uenoku, line 25: comment for link.

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,49 @@
...
+import chisel3._
+object RawClockedNonVoidFunctionCall {
+
+  /** Creates an intrinsic that calls non-void DPI function at its clock posedge.
+    *
+    * @example {{{
+    * val a = RawClockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), clock, enable, b, c)
+    * }}}
+    */
+  def apply[T <: Data](functionName: String, ret: => T)(clock: Clock, enable: Bool, data: Data*): Data = {
```
sequencer, line 14: what if the return type being a aggregate?

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  // Void function
+  RawClockedVoidFunctionCall("hello")(clock, true.B)
+
+  // Stateless function with result
+  val result = RawClockedNonVoidFunctionCall("add", UInt(32.W))(clock, true.B, io.a, io.b)
+
+  // A function that allocates a state and returns a pointer.
+  val counter = RawClockedNonVoidFunctionCall("create_counter", UInt(64.W))(clock, reset.asBool)
```
sequencer, line 61: Originally in SystemVerilog, if put multiple `RawClockedNonVoidFunctionCall` in an always block, they will be executed in order. This is be used to schedule the order of different DPI calls with dependencies, e.g. peek `valid` signal and poke the `ready` signal for handshake. It seems that we need to merge two DPIs together to achieve this?

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,49 @@
...
+
+object RawUnlockedNonVoidFunctionCall {
+
+  /** Creates an intrinsic that calls non-void DPI function when input values are changed.
+    *
+    * @example {{{
+    * val a = RawUnlockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), enable, b, c)
+    * }}}
+    */
+  def apply[T <: Data](functionName: String, ret: => T)(enable: Bool, data: Data*): Data = {
```
sequencer, line 29: I have been thinking: how to pass a string to DPI in Chisel, as well as Plusarg, maybe we also need a `StringLit` type in Chisel.

sequencer (commented):
just small thoughts.

src/main/scala/chisel3/util/circt/DPI.scala
uenoku: You can return passive aggregate. I'll add a similar document but here is ABI for DPI: https://github.com/llvm/circt/pull/7163

src/main/scala/chisel3/util/circt/DPI.scala
uenoku: Yes. We'd like to pass and create a string, open array or chandle as well (chandle can be representable with UInt<64> though). My plan is to introduce "intrinsic/foreign type" concept so they can be created by intrinsics.

src/test/scala/chiselTests/DPISpec.scala
uenoku: > It seems that we need to merge two DPIs together to achieve this?

Exactly, that's recommended way to model DPI dependency. It's extremely hard and breaking change to represent dependency within always clock with the current FIRRTL representation.

src/main/scala/chisel3/util/circt/DPI.scala
sequencer: But we may need a Chisel String data type.

uenoku: Yes, I think creating Chisel/FIRRTL String type is a right way but it will require more work. I guess we can use `Vec<UInt<8>, n>` as a way to represent string :)

sequencer: mind blowing. But yes, we can do it

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,49 @@
...
+    * }}}
+    */
+  def apply[T <: Data](functionName: String, ret: => T)(clock: Clock, enable: Bool, data: Data*): Data = {
+    IntrinsicExpr("circt_dpi_call", ret, "functionName" -> functionName, "isClocked" -> 1)(
+      (Seq(clock, enable) ++ data): _*
+    )
+  }
+}
+
+object RawUnlockedNonVoidFunctionCall {
```
sequencer, line 21:
```suggestion
object RawUnclockedNonVoidFunctionCall {
```

nit.

sequencer:
Maybe `RawUnclockedVoidFunctionCall` is missing?

sequencer:
Two questions:
- How to add a DPI function which is called at `initial`?(maybe need some preset flag to `true.B` as guard of this DPI, then configured it to false after initial)
- How to export DPI functions? This is not designed for driver, but can be used for logging.

uenoku:
> Maybe RawUnclockedVoidFunctionCall is missing?

It's intentional since it's illegal to call void-function on non-procedural region. 

> How to add a DPI function which is called at initial?(maybe need some preset flag to true.B as guard of this DPI, then configured it to false after initial)

Good point. This is not currently supported. The reason why we cannot simply add sort of `initial` flag to DPI call op is that it will create dependency between two values in the same timing, e.g:

```
wire w;
node b = dpi_call_initial<func>(w);
node a = dpi_call_initial<func>(b);
w <= a

--> 

initial begin
   a = func(b);
   b = func(a);
end

or 

initial begin
   b = func(a);
   a = func(b);
end
```

I think we can restrict operands of dpi_call_initial op to be constant types but there are few other representation so I'm still iterating on the IR design. 


> How to export DPI functions? This is not designed for driver, but can be used for logging.

This is out-of-scope now since currently FIRRTL has nothing to export :) I think we first need to introduce FIRRTL functions as first class citizen. CIRCT SV dialect recently got function support but for FIRRTL side it's still in design phase .

sequencer:
Another nit pick: when passing Reg to DPI, it will trigger the comb logic cycle error.

src/main/scala/chisel3/util/circt/DPI.scala
sequencer, line 14:
```suggestion
  def apply[T <: Data](functionName: String, ret: => T)(clock: Clock, enable: Bool, data: Data*): T = {
```

src/main/scala/chisel3/util/circt/DPI.scala
sequencer, line 29:
```suggestion
  def apply[T <: Data](functionName: String, ret: => T)(enable: Bool, data: Data*): T = {
```

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,49 @@
...
+
+object RawClockedVoidFunctionCall {
+
+  /** Creates an intrinsic that calls void DPI function at its clock posedge.
+    *
+    * @example {{{
+    * RawClockedVoidFunctionCall("dpi_func_foo", UInt(1.W), clock, enable, b, c)
+    * }}}
+    */
+  def apply(functionName: String)(clock: Clock, enable: Bool, data: Data*) = {
```
sequencer, line 44:
```suggestion
  def apply(functionName: String)(clock: Clock, enable: Bool, data: Data*): Unit = {
```

sequencer (commented):
Some nit on the return tpe

uenoku:
Sorry for delay. It's ready for review now.

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,181 @@
...
+    val b = Input(UInt(32.W))
+    val add_clocked_result = Output(UInt(32.W))
+    val add_unclocked_result = Output(UInt(32.W))
+  })
+
+  EmitDPIImplementation()
+
+  Hello()
+  val result_clocked = AddClocked(io.a, io.b)
+  val result_unclocked = AddUnclocked(io.a, io.b)
```
uenoku, line 100: comment for link

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,181 @@
...
+
+object Hello extends DPIClockedVoidFunctionImport {
+  val functionName = "hello"
+  final def apply() = super.call()
+}
+
+object AddClocked extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "add"
+  val ret = UInt(32.W)
+  val clocked = true
```
fabianschuiki, line 73: Should these also be `override`?

fabianschuiki (approved):
Very exciting to see this work!

src/test/scala/chiselTests/DPISpec.scala
uenoku: They are abstract class members so override is optional but I'll add for the consistency, thanks!

jackkoenig: Stylistically I think we prefer to add `override` to call out that it's implementing a virtual method (but this is a relatively recent stylistic opinion so the codebase is pretty inconsistent). Maybe we can do this with a lint or rewrite rule...

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,181 @@
...
+}
+
+object Hello extends DPIClockedVoidFunctionImport {
+  val functionName = "hello"
+  final def apply() = super.call()
+}
+
+object AddClocked extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "add"
+  val ret = UInt(32.W)
```
jackkoenig, line 72: I'm of two minds of this API here. It's common in Chisel for user APIs to be inheritance based, but I don't think it's generally the best option. I'm not sure if I have a better idea though and I don't want to sandbag so maybe we just go with it and change it later if we need to. Any thoughts @seldridge?

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,141 @@
...
+  val ret:     T
+  val clocked: Boolean
+  val outputName: Option[String] = None
+  def callWithEnable(enable: Bool, data: Data*): T =
+    if (clocked) {
+      RawClockedNonVoidFunctionCall(functionName, ret, inputNames, outputName)(Module.clock, enable, data: _*)
+    } else {
+      RawUnclockedNonVoidFunctionCall(functionName, ret, inputNames, outputName)(enable, data: _*)
+    }
+  def call(data: Data*): T = callWithEnable(true.B, data: _*)
```
jackkoenig, line 133: Should these be `final` or do you want users overriding them?

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,141 @@
...
+trait DPIFunctionImport {
+  val functionName: String
+  val inputNames: Option[Seq[String]] = None
+}
+
+// Base trait for a non-void function that returns `T`.
+trait DPINonVoidFunctionImport[T <: Data] extends DPIFunctionImport {
+  val ret:     T
+  val clocked: Boolean
+  val outputName: Option[String] = None
```
jackkoenig, line 126:
```suggestion
trait DPIFunctionImport {
  def functionName: String
  def inputNames: Option[Seq[String]] = None
}

// Base trait for a non-void function that returns `T`.
trait DPINonVoidFunctionImport[T <: Data] extends DPIFunctionImport {
  def ret:     T
  def clocked: Boolean
  def outputName: Option[String] = None
```

Virtual methods should be defs.

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,141 @@
...
+    *
+    * Please refer https://github.com/llvm/circt/blob/main/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md#dpi-intrinsic-abi for DPI function ABI.
+    * @example {{{
+    * val a = RawUnclockedNonVoidFunctionCall("dpi_func_foo", UInt(1.W), enable, b, c)
+    * }}}
+    */
+  def apply[T <: Data](
+    functionName: String,
+    ret:          => T,
+    inputNames:   Option[Seq[String]] = None,
```
jackkoenig, line 79: Is there a reason for this to be `Option[Seq[String]]` vs. just `Seq[String]`? Is there an important difference between `None` and `Some(Seq.empty)`?

src/main/scala/chisel3/util/circt/DPI.scala
```diff
@@ -0,0 +1,141 @@
...
+    outputName:   Option[String] = None
+  )(enable:       Bool,
```
sequencer, line 81:
```suggestion
  )(enable:       Bool = When.cond,
```
Just a small suggestion to capture the `when` condition.

jackkoenig: I think this is handled by firtool but @uenoku can correct me if I'm wrong.

sequencer: If handled in `firtool`, what's the behavior of a DPI call inside the `when` block, but also providing the `enable` signal?

uenoku: When condition will be added to enable as conjunction

src/main/scala/chisel3/util/circt/DPI.scala
uenoku: No I don't. These should be final, thanks!

src/main/scala/chisel3/util/circt/DPI.scala
uenoku: I want to distinguish nullary function and user defined names not being provided. If we require users to provide input and output names, it's ok for me to just use Seq[String].

src/test/scala/chiselTests/DPISpec.scala
seldridge: There's some precedence for this kind of API for things like layers where you want to "declare" something and then refer to it multiple times. Using the object here achieves that (though I mentioned to @uenoku to self-type this with `this: Singleton =>` to mandate an object). Also, this could be slightly cleaner as an `abstract class` in order to avoid the copious overrides.

Let's try this and see where it bottoms out.

jackkoenig: Sounds good to me, thanks @seldridge!

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -0,0 +1,181 @@
...
+          dpi.io.a.poke(24.U)
+          dpi.io.b.poke(12.U)
+          dpi.io.add_clocked_result.peek()
+          dpi.io.add_clocked_result.expect(60)
+          dpi.io.add_unclocked_result.peek()
+          dpi.io.add_unclocked_result.expect(36)
+        }
+        .result
+
+      val outputFile = io.Source.fromFile("test_run_dir/simulator/DPIIntrinsic/workdir-verilator/simulation-log.txt")
```
jackkoenig, line 134: Ideally we would be a little DRY-er on these paths, deriving them from values defined above, but I'm not gonna block this PR on that.

jackkoenig (approved):
This is fantastic, thank you @uenoku!

dtzSiFive:
Great work! 🚀

uenoku:
Thank you for review!

sequencer:
For those who come into this PR, and interested in the usage of Chisel DPI, here is a concrete opensource example of using `DPIIntrinsic` + `verilator`:
- all signals are driven by DPI, see:
https://github.com/chipsalliance/t1/blob/master/ipemu/src/TestBench.scala
https://github.com/chipsalliance/t1/blob/master/ipemu/src/TestBench.scala

- compile verilog via verialtor and it will elaborate header files, see:
https://github.com/chipsalliance/t1/blob/master/ipemu/src/TestBench.scala

- implement them in the native languages(C++, C, Rust), (T1 use rust):
https://github.com/chipsalliance/t1/blob/master/ipemu/src/TestBench.scala

then link the dpi w/ verilator, then you can drive the Verilog with your native language then.

VCS should work as well, but due to the NDA limitation those cannot be opensourced.

Issue #4159, opened 2024-06-10
Inconsistent width for ChiselEnums cast to UInt


jackkoenig:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**

Consider the following Chisel:

```scala
//> using scala "2.13.12"
//> using dep "org.chipsalliance::chisel:6.4.0"
//> using plugin "org.chipsalliance:::chisel-plugin:6.4.0"
//> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Xfatal-warnings", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"

import chisel3._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._
import _root_.circt.stage.ChiselStage

object MyEnum extends ChiselEnum {
  val e0, e1, e2 = Value

  val e10 = Value(10.U)
}

class Foo extends Module {
  val out = IO(Output(UInt(8.W)))

  val x = MyEnum.e1.asUInt
  val y = WireInit(x)
  val z = WireInit(chiselTypeOf(x), x)
  dontTouch(y)
  dontTouch(z)

  out := x
}

object Main extends App {
  println(ChiselStage.emitCHIRRTL(new Foo))
  println(
    ChiselStage.emitSystemVerilog(
      gen = new Foo,
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

**What is the current behavior?**

Running the above gives:
```firrtl
FIRRTL version 3.3.0
circuit Foo :%[[
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~Foo|Foo>y"
  },
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~Foo|Foo>z"
  }
]]
  module Foo : @[Users/koenig/work/t/chiselenum/chisel-example.scala 16:7]
    input clock : Clock @[Users/koenig/work/t/chiselenum/chisel-example.scala 16:7]
    input reset : UInt<1> @[Users/koenig/work/t/chiselenum/chisel-example.scala 16:7]
    output out : UInt<8> @[Users/koenig/work/t/chiselenum/chisel-example.scala 17:15]

    node x = asUInt(UInt<1>(0h1)) @[Users/koenig/work/t/chiselenum/chisel-example.scala 19:21]
[...]

PR #4161, opened 2024-06-11
Add a test for issue 4159

jackkoenig:
Enshrine the behavior of #4159 - this test will be useful for backporting and then showing the change when we fix it.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/ChiselEnum.scala
+++ b/src/test/scala/chiselTests/ChiselEnum.scala
@@ -401,8 +401,8 @@ class ChiselEnumSpec extends ChiselFlatSpec with Utils {
     assertTesterPasses(new CastToUIntTester)
   }
 
-  // This is a bug but fixing it may break user code
-  // See https://github.com/chipsalliance/chisel/issues/4159
+  // This is a bug, but fixing it may break user code.
+  // See: https://github.com/chipsalliance/chisel/issues/4159
   it should "preserve legacy width behavior" in {
     val verilog = ChiselStage.emitSystemVerilog(new RawModule {
       val out1, out2, out3 = IO(Output(UInt(8.W)))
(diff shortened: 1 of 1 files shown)


src/test/scala/chiselTests/ChiselEnum.scala
```diff
@@ -401,6 +401,30 @@ class ChiselEnumSpec extends ChiselFlatSpec with Utils {
     assertTesterPasses(new CastToUIntTester)
   }
 
+  // This is a bug but fixing it may break user code
+  // See https://github.com/chipsalliance/chisel/issues/4159
```
seldridge, line 405: Mega nits, also throughout. Full sentence comments should be a period:

```suggestion
  // This is a bug, but fixing it may break user code.
  // See: https://github.com/chipsalliance/chisel/issues/4159
```

src/test/scala/chiselTests/ChiselEnum.scala
```diff
@@ -401,6 +401,30 @@ class ChiselEnumSpec extends ChiselFlatSpec with Utils {
...
+      x.getWidth should be(7)
+      x.getWidth should be(EnumExample.getWidth)
+      y.widthOption should be(None)
+      z.getWidth should be(7)
+    })
+    // The bug is that all of these should be the same as out3, or the widths above are wrong
+    verilog should include("assign out1 = 8'h3;")
+    verilog should include("assign out2 = 8'h3;")
+    verilog should include("assign out3 = 8'h81;")
+  }
```
seldridge, line 426: Scalatest doesn't have XFAIL, right? This would be better as that if it's supported. Alternatively, this could be a `should not include` which is the correct thing.

seldridge (approved):

src/test/scala/chiselTests/ChiselEnum.scala
jackkoenig: I don't think ScalaTest has an XFAIL per se, more of `should not include` yeah. Are you suggesting I also add `should not` for the "correct" output when we fix this (ie.  `verilog should not include("assign out1 = 8'h81;")` and `verilog should not include("assign out2 = 8'h81;")`)? Or are you suggesting something else?

PR #4168, opened 2024-06-12
Preserve literals across .asTypeOf

jackkoenig:
Built on top of https://github.com/chipsalliance/chisel/pull/4160 so should merge this one after that one (might need rebase but git seems smarter about cherry-picks these days).

This is very valuable for ChiselSim-style testing because it means that you can much more easily construct literals in tests.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

#### Release Notes

Casting a literal (of any type) to another type with .asTypeOf will result in a literal of the new type. For non-literals, the FIRRTL representation will now be a little bit more efficient.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ChiselEnum.scala
+++ b/core/src/main/scala/chisel3/ChiselEnum.scala
@@ -45,13 +45,8 @@ abstract class EnumType(private[chisel3] val factory: ChiselEnum, selfAnnotating
     pushOp(DefPrim(sourceInfo, Bool(), op, this.ref, other.ref))
   }
 
-  private[chisel3] override def connectFromBits(
-    that: Bits
-  )(
-    implicit sourceInfo: SourceInfo
-  ): Unit = {
-    this := factory.apply(that.asUInt)
-  }
+  override private[chisel3] def _fromUInt(that: UInt)(implicit sourceInfo: SourceInfo): Data =
+    factory.apply(that.asUInt)
 
   final def ===(that: EnumType): Bool = macro SourceInfoTransform.thatArg
   final def =/=(that: EnumType): Bool = macro SourceInfoTransform.thatArg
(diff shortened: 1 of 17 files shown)


jackkoenig:
I was hoping to backport this because I think it's potentially a very useful feature for ChiselSim and 6.x is closer to 3.6.x which makes for easier porting, but alas it seems that it is a really common pattern to use `0.U.asTypeOf(new MyBundle)` to initialize a Wire of type `MyBundle` that then as a few fields connected to.

This change is still good and worth making, but it needs to be on a major version since it will break those uses (they just need to add a `WireInit`).

jackkoenig:
Converted to draft temporarily. I want to deprecate assigning to the wire returned by `.asTypeOf` (and backport to 6.x) before merging this. I am almost done with the PR doing that deprecation (which involves adding my long desired `.readOnly` feature!), should open it tomorrow.

jackkoenig:
This is ready to go

src/test/scala/chiselTests/stage/WarningConfigurationSpec.scala
```diff
@@ -380,10 +380,10 @@ class WarningConfigurationSpec extends AnyFunSpec with Matchers with chiselTests
...
     }
 
-    it("should number AsTypeOfReadOnly as 8") {
-      val args = Array("--warn-conf", "id=8:e,any:s", "--throw-on-first-error")
+    it("should now error on AsTypeOfReadOnly") {
+      val args = Array("--throw-on-first-error")
       val e = the[Exception] thrownBy ChiselStage.emitCHIRRTL(new AsTypeOfReadOnly, args)
-      e.getMessage should include("[W008] Return values of asTypeOf will soon be read-only")
+      e.getMessage should include("Return values of asTypeOf are now read-only")
     }
```
jackkoenig, line 387: So this is the first time we'll be elevating a warning to an error. I haven't really worked through it all, we could add a corresponding `E008` for the old `W008`, but no other errors are numbered yet (and of course, unlike warnings, you will not be allowed to configure errors). Any thoughts?

mwachs5: it seems like if we want to add error numbers we should do that more holistically? separately from this PR?

core/src/main/scala/chisel3/ChiselEnum.scala
```diff
@@ -73,6 +68,27 @@ abstract class EnumType(private[chisel3] val factory: ChiselEnum, selfAnnotating
   def do_>=(that: EnumType)(implicit sourceInfo: SourceInfo): Bool =
     compop(sourceInfo, GreaterEqOp, that)
 
+  // This is a hack to preserve the _workaround_ for https://github.com/chipsalliance/chisel/issues/4159
```
mwachs5, line 71:
```suggestion
  // This preserves the _workaround_ for https://github.com/chipsalliance/chisel/issues/4159
```

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -67,6 +69,40 @@ package object internal {
     if (headOk) res else s"_$res"
   }
 
+  // Hack to work around https://github.com/chipsalliance/chisel/issues/4162
```
mwachs5, line 72:
```suggestion
  // Work around https://github.com/chipsalliance/chisel/issues/4162
```

mwachs5 (approved):

PR #4184, opened 2024-06-17
Rework When IR to be Region-based

seldridge:
Change the internal representation of `When` in Chisel to be region based as opposed to token-based.  Concretely, change `WhenContext` to not push token commands indicating when the region starts/stops, has an alternative, etc.  Instead, use a single object that has an "if" and an "else" region where commands can be pushed.

CC: @sequencer, @SpriteOvO: I didn't fixup the bindings here.

### DIFF
--- a/core/src/main/scala/chisel3/When.scala
+++ b/core/src/main/scala/chisel3/When.scala
@@ -68,7 +68,7 @@ private object Scope {
     override def toString: String = "if"
   }
 
-  /** Indicatest that the `WhenContext` is in the "else" clause. */
+  /** Indicates that the `WhenContext` is in the "else" clause. */
   object Else extends Type {
     override def toString: String = "else"
   }
--- a/core/src/main/scala/chisel3/When.scala
+++ b/core/src/main/scala/chisel3/When.scala
@@ -93,12 +93,6 @@ final class WhenContext private[chisel3] (
   // For capturing conditions from prior whens or elsewhens
   altConds: List[() => Bool]) {
 
-  /** Commands that occur under the "if" leg of this when statement. */
-  val ifRegion = new VectorBuilder[Command]()
-
-  /** Commands that occurr under the "else" leg of this when statement. */
-  val elseRegion = new VectorBuilder[Command]()
-
   /** Indicate if the `WhenContext` is "closed" (`None`) or if this is writing to
     * the "if" or "else" region.
     */
@@ -131,7 +125,7 @@ final class WhenContext private[chisel3] (
   )(
     implicit sourceInfo: SourceInfo
   ): WhenContext = {
-    Builder.forcedUserModule.withRegion(elseRegion) {
+    Builder.forcedUserModule.withRegion(whenCommand.elseRegion) {
       new WhenContext(sourceInfo, () => elseCond, block, firrtlDepth + 1, cond :: altConds)
     }
   }
@@ -146,25 +140,24 @@ final class WhenContext private[chisel3] (
   def otherwise(block: => Any)(implicit sourceInfo: SourceInfo): Unit = {
     Builder.pushWhen(this)
     scope = Some(Scope.Else)
-    Builder.forcedUserModule.withRegion(elseRegion) {
+    Builder.forcedUserModule.withRegion(whenCommand.elseRegion) {
       block
     }
     scope = None
     Builder.popWhen()
   }
 
+  /** Return true if this `WhenContext` is currently constructing operations. */
   def active: Boolean = scope.isDefined
 
-  /*
-   *
-   */
-  // if (firrtlDepth > 0) { pushCommand(AltBegin(sourceInfo)) }
-  // cond.foreach(c => pushCommand(WhenBegin(sourceInfo, c().ref)))
-  pushCommand(When(sourceInfo, cond().ref, ifRegion, elseRegion))
+  // Create the `When` operation and run the `block` thunk inside the
+  // `ifRegion`.  Any commands that this thunk creates will be put inside this
+  // block.
+  private val whenCommand = pushCommand(When(sourceInfo, cond().ref))
   Builder.pushWhen(this)
   scope = Some(Scope.If)
   try {
-    Builder.forcedUserModule.withRegion(ifRegion) {
+    Builder.forcedUserModule.withRegion(whenCommand.ifRegion) {
       block
     }
   } catch {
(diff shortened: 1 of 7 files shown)


SpriteOvO:
:ok_hand: Got it! I will fix the circt converter after this PR lands.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -326,10 +326,14 @@ private[chisel3] object ir {
...
-  case class WhenBegin(sourceInfo: SourceInfo, pred: Arg) extends Command
-  case class WhenEnd(sourceInfo: SourceInfo, firrtlDepth: Int, hasAlt: Boolean = false) extends Command
-  case class AltBegin(sourceInfo: SourceInfo) extends Command
-  case class OtherwiseEnd(sourceInfo: SourceInfo, firrtlDepth: Int) extends Command
+
+  case class When(
+    sourceInfo: SourceInfo,
+    pred:       Arg,
+    ifRegion:   VectorBuilder[Command] = new VectorBuilder[Command],
+    elseRegion: VectorBuilder[Command] = new VectorBuilder[Command])
```
jackkoenig, line 334: It might be worth making the `elseRegion` default to `null` and doing the `Option` boxing to make the normal API safer. Quick maths shows that an empty `VectorBuilder` costs 72-96 bytes (depending on heap size and thus if compressed pointers are in use).

seldridge: Good idea. I changed this, roughly in this direction, with https://github.com/chipsalliance/chisel/pull/4184/commits/6e53f83e60ad58998c78c84d34dd8bf6dbb52659. This kind of thing really means `case class` isn't appropriate. One nice thing about this approach is that the `unapply` can then do the conversion to `Seq[Command]` and avoid needing to do calls to `.result()` all over the place.

seldridge:
@SpriteOvO: If you have cycles, would you be able to push to this branch and fix the `PanamaCIRCTConverter.scala` failures? I _think_ that this should be relatively easy and would remove a lot of the janky context tracking that you had to add.

sequencer:
@SpriteOvO please fix the `???`.

SpriteOvO:
Yea, I'm working on it

core/src/main/scala/chisel3/When.scala
```diff
@@ -60,6 +60,20 @@ object when {
...
 
+private object Scope {
+  sealed trait Type
+
+  /** Indicates that the `WhenContext` is in the "if" clause. */
+  object If extends Type {
+    override def toString: String = "if"
+  }
+
+  /** Indicatest that the `WhenContext` is in the "else" clause. */
```
youngar, line 71: "Indicatest"

jackkoenig: I love that, "indicate by testing"

SpriteOvO:
I have fixed the CIRCT converter on this branch, and I love this new Region-based When IR, it's super intuitive. :laughing:

core/src/main/scala/chisel3/When.scala
seldridge: Fixed in https://github.com/chipsalliance/chisel/pull/4184/commits/4b1c8f1d811cf8c5897c33b9510e6bb063e499e9.

seldridge:
Awesome @SpriteOvO! Thanks for the cooperative PR work. 💯

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -326,10 +326,26 @@ private[chisel3] object ir {
...
+    }
+  }
+
+  object When {
+    def unapply(when: When): Option[(SourceInfo, Arg, Seq[Command], Seq[Command])] = {
+      Some(
+        (when.sourceInfo, when.pred, when.ifRegion.result(), when._elseRegion.map(_.result()).getOrElse(Seq.empty))
+      )
+    }
+  }
```
jackkoenig, line 347:
```suggestion
    private var _elseRegion: VectorBuilder[Command] = null
    def elseRegion: VectorBuilder[Command] = {
      if (_elseRegion == null) {
        _elseRegion = new VectorBuilder[Command]
      }
      _elseRegion
    }

  object When {
    def unapply(when: When): Option[(SourceInfo, Arg, Seq[Command], Seq[Command])] = {
      Some(
        (when.sourceInfo, when.pred, when.ifRegion.result(), Option(when._elseRegion).map(_.result()).getOrElse(Seq.empty))
      )
    }
  }
```
`null` is awful in public APIs, but it saves 16-bytes per instance of `When` that has an has an `else`. Since it's only relevent in about 6 lines of code, it's worth it. Note that `Option(_)` boxing the `null` "reifies" it into the Option world safely, and because that's a function call, it can (and will in practice) be inlined and optimized away (unlike a field of type Option which cannot be optimized, or at least isn't by the Scala compiler today).

jackkoenig: Also consider `.map(_.result()).getOrElse(Seq.empty))` => `.fold(Seq.empty)(_.result())`. I prefer that but I respect others may not 🙂

src/main/scala/chisel3/aop/Select.scala
```diff
@@ -512,27 +526,25 @@ object Select {
...
+      commands:       Seq[Command],
+      preds:          Seq[Predicate],
+      processCommand: (Command, Seq[Predicate]) => Unit
+    ): Unit = commands.foreach {
+      case FWhen(_, cond, ifRegion, elseRegion) =>
+        val pred = cond match {
+          case Node(pred: Bool) => When(pred)
+          case l: LitArg if l.num == BigInt(1) => When(true.B)
+          case l: LitArg if l.num == BigInt(0) => When(false.B)
+          case _ => ???
```
jackkoenig, line 540:
```suggestion
          case _ =>
            sys.error(s"Something went horribly wrong! I was expecting $pred to be a lit or a bool!")
```
Any reason not to maintain the old error behavior?

jackkoenig (approved):
LGTM, minor suggestions, merge at your leisure.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
seldridge: Good suggestion. Fixed in: https://github.com/chipsalliance/chisel/pull/4184/commits/9279b9ccef7ca754c91a7df19bbc6584940a28ab

src/main/scala/chisel3/aop/Select.scala
seldridge: Not really. Fixed in c1f8b2302.

PR #4186, opened 2024-06-18
Distinguish identity views from single-target views

### LINKED ISSUE #4185: FIRRTL bulk connections emitted incorrectly for views of same type but non-identity mapping
**Type of issue**: Bug Report

Consider the following Chisel:
```scala
//> using scala "2.13.12"
//> using dep "org.chipsalliance::chisel:6.4.0"
//> using plugin "org.chipsalliance:::chisel-plugin:6.4.0"

import chisel3._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._

object Util {
  type ReversedVec[T <: Data] = Vec[T]
  implicit def reversedVecView[T <: Data]: DataView[Vec[T], ReversedVec[T]] =
    DataView.mapping[Vec[T], ReversedVec[T]](v => v.cloneType, { case (a, b) => a.reverse.zip(b) })
}
import Util._
  val in = IO(Input(Vec(2, UInt(8.W))))
  val turducken, legacy = IO(Output(Vec(2, UInt(8.W))))

  val view = in.viewAs[ReversedVec[UInt]]
  turducken :<>= view
  legacy <> view
}

object Main extends App {
  println( ChiselStage.emitCHIRRTL(new MyModule))
  println(
    ChiselStage.emitSystemVerilog(
      gen = new MyModule,
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

This emits the following Verilog:
```verilog
// Generated by CIRCT firtool-1.62.0
module MyModule(
  input        clock,
               reset,
  input  [7:0] in_0,
               in_1,
  output [7:0] turducken_0,
               turducken_1,
               legacy_0,
               legacy_1
);

  assign turducken_0 = in_1;
  assign turducken_1 = in_0;
  assign legacy_0 = in_0;
  assign legacy_1 = in_1;
endmodule
```

Obviously those connections to `legacy` are wrong. This is due to the fact that Chisel is emitting a bulk connect to FIRRTL:
```firrtl
FIRRTL version 3.3.0
circuit MyModule :
  module MyModule : @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 18:7]
    input clock : Clock @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 18:7]
    input reset : UInt<1> @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 18:7]
    input in : UInt<8>[2] @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 19:14]
    output turducken : UInt<8>[2] @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 20:29]
    output legacy : UInt<8>[2] @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 20:29]

    connect turducken[0], in[1] @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 23:13]
    connect turducken[1], in[0] @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 23:13]
    connect legacy, in @[Users/koenig/work/t/badbulkconnect/chisel-example.scala 24:10]
```

**Other Information**

### LINKED ISSUE #4187: Views of Probes of Aggregates cause Stack Overflow errors
**Type of issue**: Bug Report

Consider the following Chisel:
```scala
//> using scala "2.13.12"
//> using dep "org.chipsalliance::chisel:6.4.0"
//> using plugin "org.chipsalliance:::chisel-plugin:6.4.0"

import chisel3._
import chisel3.probe._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._

  val a = UInt(8.W)
  val b = UInt(8.W)
}

class Foo extends RawModule {
  val out = IO(Output(Probe(new MyBundle)))
}

class Top extends RawModule {
  val f = Module(new Foo)

  f.out.viewAs.isLit
}

object Main extends App {
  println(
    ChiselStage.emitCHIRRTL(
      gen = new Top
    )
  )
}
```

The `f.out.viewAs.isLit` will cause a stack overflow.

Views of Probes of Aggregates should work!

**Other Information**

jackkoenig:
This is a minor overhaul of DataViews internals that make the various use cases of what were previously referred to as "1-1 views" more clear. This incidentally fixes an issue with bulk connections and views.

While for users this is just a bug fix, this has larger implications for developers. This replaces `reifySingleData` with two more precise functions:
* `reifyIdentityView` - Fast, used for things like Probe legality (which by having pointer semantics, only really works with identity views).
* `reifySingleTarget` - Slow, used for legacy annotation support which was, and remains, iffy for views.

A subtle but critical aspect of this PR is that `reifyIdentityView` works in cases when the `reifySingleData` did not. It is properly hierarchical: e.g. if you have an Aggregate that is a child of a non-identity view parent yet the child itself *is* an identity view, it will return the target for that child. This is critical for use cases like FlatIO where the parent Bundle isn't an identity but every single child is.

One negative of this PR is that it makes the awkward "unnamed rename map" logic slower. I benchmarked this on a large design (that uses lots of views) and it's such a small piece of the overall runtime that it's _okay_ (~1%). Regardless that logic is ludicrously slow and is a complete waste. I am thinking about ways to get rid of it but this PR has scope creeped enough. I think we just have to change the annotation API so it's really about coming up with the best way to do it.

This PR actually fixes several bugs, I will file them separately as they illustrate all of the weird corner cases this PR cleans up.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash
#### Release Notes

Fixes #4185, Fixes #4187

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chisel3/experimental/dataview/ReifySpec.scala
+++ b/src/test/scala/chisel3/experimental/dataview/ReifySpec.scala
@@ -1,0 +1,317 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.experimental.dataview
+
+import circt.stage.ChiselStage
+import chisel3._
+import chisel3.probe.Probe
+import chisel3.properties.Property
+import chisel3.experimental.Analog
+import chisel3.experimental.dataview._
+
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers._
+import chisel3.reflect.DataMirror
+import chisel3.experimental.hierarchy.{instantiable, public, Instantiate}
+
+object ReifySpec {
+  // Views can be single target of the same type, but still not identity.
+  type ReversedVec[T <: Data] = Vec[T]
+  implicit def reversedVecView[T <: Data]: DataView[Vec[T], ReversedVec[T]] =
+    DataView.mapping[Vec[T], ReversedVec[T]](v => v.cloneType, { case (a, b) => a.reverse.zip(b) })
+
+  class AllElementsBundle extends Bundle {
+    val u = UInt(8.W)
+    val s = SInt(8.W)
+    val b = Bool()
+    val r = Reset()
+    val d = AsyncReset()
+    val c = Clock()
+    val a = Analog(8.W)
+    val p = Probe(UInt(8.W))
+    val prop = Property[String]()
+  }
+  implicit val allElementsView: DataView[Seq[Data], AllElementsBundle] =
+    DataView.mapping[Seq[Data], AllElementsBundle](
+      _ => new AllElementsBundle,
+      { case (a, b) => a.zip(b.getElements) }
+    )
+
+  class SimpleBundle extends Bundle {
+    val value = UInt(8.W)
+  }
+  class NestedBundle extends Bundle {
+    val child = new SimpleBundle
+  }
+  class TargetBundle extends Bundle {
+    val fizz = new NestedBundle
+    val buzz = new SimpleBundle
+    val vec = Vec(2, UInt(8.W))
+  }
+  class ViewChildBundle extends Bundle {
+    val a = Vec(2, UInt(8.W))
+    val b = Vec(2, UInt(8.W))
+  }
+  class ViewBundle extends Bundle {
+    val foo = UInt(8.W) // used to ensure ViewBundle isn't 1-1
+    val bar = new ViewChildBundle
+  }
+  // The key thing about this mapping is that ViewBundle.bar maps 1-1 (but not identity) to TargetBundle
+  // and that ViewBundle.bar.b identity maps to TargetBundle.vec.
+  implicit val myView: DataView[(UInt, TargetBundle), ViewBundle] =
+    DataView.mapping[(UInt, TargetBundle), ViewBundle](
+      _ => new ViewBundle,
+      {
+        case ((u, t), v) =>
+          Seq(u -> v.foo, t.fizz.child.value -> v.bar.a(0), t.buzz.value -> v.bar.a(1), t.vec -> v.bar.b)
+      }
+    )
+}
+
+import ReifySpec._
+
+class ReifySpec extends AnyFunSpec {
+
+  describe("dataview.reify") {
+
+    it("should reify single targets and identity for all non-view Elements") {
+      ChiselStage.convert(new Module {
+        val wires = (new AllElementsBundle).getElements.map(Wire(_))
+
+        // .getElements returns Data so we have to match that these are Elements.
+        wires.foreach {
+          case elt: Element =>
+            reify(elt) should be(elt)
+            reifyIdentityView(elt) should be(Some(elt))
+            reifySingleTarget(elt) should be(Some(elt))
+        }
+      })
+    }
+
+    it("should reify single targets and identity for all non-view Elements even as children of an Aggregate") {
+      ChiselStage.convert(new Module {
+        val bundle = IO(new AllElementsBundle)
+
+        // .getElements returns Data so we have to match that these are Elements.
+        bundle.getElements.foreach {
+          case elt: Element =>
+            reify(elt) should be(elt)
+            reifyIdentityView(elt) should be(Some(elt))
+            reifySingleTarget(elt) should be(Some(elt))
+        }
+      })
+    }
+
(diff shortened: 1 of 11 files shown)


src/test/scala/chisel3/experimental/dataview/ReifySpec.scala
```diff
@@ -0,0 +1,318 @@
...
+    val a = Vec(2, UInt(8.W))
+    val b = Vec(2, UInt(8.W))
+  }
+  class ViewBundle extends Bundle {
+    val foo = UInt(8.W) // used to ensure ViewBundle isn't 1-1
+    val bar = new ViewChildBundle
+  }
+  // The key thing about this mapping is that ViewBundle.bar maps 1-1 to TargetBundle
+  // but the relationship is tricky.
+  // TODO should we add an aggregate child of ViewBundle that maps 1-1 to an aggregate child of TargetBundle?
```
jackkoenig, line 61:
```suggestion
  // The key thing about this mapping is that ViewBundle.bar maps 1-1 (but not identity) to TargetBundle
  // and that ViewBundle.bar.b identity maps to TargetBundle.vec
```

src/test/scala/chisel3/experimental/dataview/ReifySpec.scala
```diff
@@ -0,0 +1,318 @@
...
+}
+
+import ReifySpec._
+
+class ReifySpec extends AnyFunSpec {
+
+  describe("dataview.reify") {
+
+    it("should reify single targets and identity for all non-view Elements") {
+      ChiselStage.emitCHIRRTL(new Module {
```
seldridge, line 79: Slightly faster to only run. Ditto throughout:

```suggestion
      ChiselStage.elaborate(new Module {
```

seldridge (approved):
This is pretty far into dark corners of Chisel that I don't have cached context to comment on well. I gave the code a once over and it seemed fine. I'm approving this to keep things going here as I'm confident that you will deal with any impacts of this.

High level comments: 

There seems like there is a lot of complexity that is stemming from `reifySingleTarget`. Do we actually need to have the ability annotate a view? Would things be clearer/cleaner if this support was instead removed as opposed to making this a failable operation?

The concerns with performance are brought up a couple of times in the PR. Are there expectations that `reifySingleTarget` has internal users which could see a performance degradation from this? Restating the first high-level comment, the performance impact goes away if you can't do this.

Granted, I tend to think of `DataView` as closer to a mechanism for a user to define how two things should be connected. (You could argue that `DataView` is actually a connectable type class.) From that perspective, I do think that the first high-level comment is made stronger as there is less of an expectation that if you have defined how two things "connect" that you can do anything with that type class implementation other than connect them.

jackkoenig:
Thanks for the review!


> There seems like there is a lot of complexity that is stemming from `reifySingleTarget`. Do we actually need to have the ability annotate a view? Would things be clearer/cleaner if this support was instead removed as opposed to making this a failable operation?

I think it would be reasonable to ban annotating [non-identity] views. Targets are similar to Probes in that they are references and I do think there's just fundamentally a clash between the semantics of pointers/references and the value-semantics of views. I think annotation authors should still be able to annotate the targets of views, but we could do that by tweaking the annotation API and exposing reification via `DataMirror` (or some other way). The annotation author can then choose to support them or not as appropriate (e.g. obviously `dontTouch` can easily work on views, it's fine with being split up).

> The concerns with performance are brought up a couple of times in the PR. Are there expectations that `reifySingleTarget` has internal users which could see a performance degradation from this? Restating the first high-level comment, the performance impact goes away if you can't do this.
> 
> Granted, I tend to think of `DataView` as closer to a mechanism for a user to define how two things should be connected. (You could argue that `DataView` is actually a connectable type class.) From that perspective, I do think that the first high-level comment is made stronger as there is less of an expectation that if you have defined how two things "connect" that you can do anything with that type class implementation other than connect them.

I measured the performance impact and it seems small in practice. Whatever we do, the unnamed views renamemap needs to go away. We can do this by tweaking the annotation API to only operate on the targets of views or banning annotating views directly but providing users the ability to do reification themselves.

I think we should go ahead and do this on `main` while leaving the current sketchy but functional thing going on 6.x.
I'm only backporting this because it's needed for `.readOnly` (next PR), and I need that on 6.x to deprecate some stuff so I can change it on main 🙂.

PR #4188, opened 2024-06-18
Add Region to Chisel IR

seldridge:
Add a Region operation to Chisel IR.  This is a command which is a container of other commands.  This is intended to provide an "insertion point" that can be added to the command buffer which can then be referenced later to add more commands.  This avoids the problem of the command buffer being append only while _not_ going fully in the direction of using a datastructure which allows arbitrary insertion (i.e., a linked list).

This is stacked on #4184.

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -282,11 +282,7 @@ private[chisel3] object Converter {
         )
       )
     case Region(info, region) =>
-      if (region.isEmpty) {
-        None
-      } else {
-        Some(fir.Block(convert(region, ctx, typeAliases)))
-      }
+      Some(fir.Block(convert(region, ctx, typeAliases)))
     case _ => None
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -281,6 +281,12 @@ private[chisel3] object Converter {
           convert(elseRegion, ctx, typeAliases)
         )
       )
+    case Region(info, region) =>
+      if (region.isEmpty) {
+        None
+      } else {
+        Some(fir.Block(convert(region, ctx, typeAliases)))
+      }
```
jackkoenig, line 289:
```suggestion
      Option.when(region.nonEmpty) {
        fir.Block(convert(region, ctx, typeAliases))
      }
```
YMMV but I like `Option.when`

jackkoenig (approved):

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
seldridge: This cleans up the code as written. There is actually a slightly different fix that is necessary here as this needs to return `None` only if it isn't handled. I.e., this should just be a `Some`. Will push a fix and land.

seldridge: Fixed in: https://github.com/chipsalliance/chisel/pull/4188/commits/3738c1fa0102240f370781dab0c595fe15e66ff0

Issue #1267, opened 2019-12-05
API to turn Data into read-only reference


jackkoenig:
**Type of issue**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: request

**What is the use case for changing the behavior?**

This is a proposal to have an API to turn any `Data` object into a read-only reference. You can do this with `UInt`s by performing some no-op (eg. `x + 0.U`), but there is no way that I can think of to do the same with an `Aggregate`.

I'm not sure what the API should, maybe `.readOnly` or `ReadOnlyWireInit(...)`, but it can easily just map to a node in FIRRTL: `node _T_1 = originalSignal`, const prop will remove the node.

PR #4190, opened 2024-06-18
Add support for marking things as readOnly

jackkoenig:
Also add internal support for marking things as "deprecated read only". This will promptly be used to mark the return value of `.asTypeOf` as "deprecated read only" in 6.x in preparation of https://github.com/chipsalliance/chisel/pull/4168.

I implemented this as part of the view binding for the following reasons:
1. While one could mutate the underlying Data, there are use cases where an API may want to return a read-only reference/view of their underlying Data while reserving the right to connect to it themselves (I intend to do this in proprietary code)
2. We could implement this as a separate Binding (i.e. not part of the existing Views) or as an Arg. Either approach could certainly work, but it would need to compose with views _anyway_. Consider some of the tests. Given some Data that is marked readOnly that is then combined with other (potentially not read-only) Data to be viewed as some Aggregate type--the read-only aspect of that target needs to be preserved. This "writeability" is thus an intrinsic property of reification anyway, so we might as well just embrace it and make it part of the API so that callers handle it as appropriate (which is what I implemented).

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Users can call `.readOnly` on any `Data` to prevent connections to the returned value. Resolves https://github.com/chipsalliance/chisel/issues/1267.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/dataview/package.scala
+++ b/core/src/main/scala/chisel3/experimental/dataview/package.scala
@@ -20,7 +20,13 @@ package object dataview {
     * Calling `viewAs` also requires an implementation of [[DataView]] for the target type
     */
   implicit class DataViewable[T](target: T) {
-    def viewAs[V <: Data](implicit dataproduct: DataProduct[T], dataView: DataView[T, V], sourceInfo: SourceInfo): V = {
+    private def _viewAsImpl[V <: Data](
+      writability: ViewWriteability
+    )(
+      implicit dataproduct: DataProduct[T],
+      dataView:             DataView[T, V],
+      sourceInfo:           SourceInfo
+    ): V = {
       // TODO put a try catch here for ExpectedHardwareException and perhaps others
       // It's likely users will accidentally use chiselTypeOf or something that may error,
       // The right thing to use is DataMirror...chiselTypeClone because of composition with DataView.andThen
@@ -28,7 +34,7 @@ package object dataview {
       val result: V = dataView.mkView(target)
       requireIsChiselType(result, "viewAs")
 
-      doBind(target, result, dataView)
+      doBind(target, result, dataView, writability)
 
       // Setting the parent marks these Data as Views
       result.setAllParents(Some(ViewParent))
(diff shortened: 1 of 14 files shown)


jackkoenig:
I'm going to peel off the bit that actually marks the return value as "deprecated read-only" into its own PR.

core/src/main/scala/chisel3/internal/BiConnect.scala
```diff
@@ -360,11 +401,11 @@ private[chisel3] object BiConnect {
...
         //    CURRENT MOD   CHILD MOD
-        case (Input, Input)    => issueConnectL2R(left, right)
-        case (Internal, Input) => issueConnectL2R(left, right)
+        case (Input, Input)    => issueConnectL2R(left, lwr, right, rwr)
+        case (Internal, Input) => issueConnectL2R(left, lwr, right, rwr)
 
-        case (Output, Output)   => issueConnectR2L(left, right)
-        case (Internal, Output) => issueConnectR2L(left, right)
+        case (Output, Output)   => issueConnectR2L(left, lwr, right, rwr)
+        case (Internal, Output) => issueConnectR2L(left, lwr, right, rwr)
```
jackkoenig, line 409: The diff here can probably be made lower touch. The `issueConnect` functions call `.lref` so we could probably rely on the error reporting there, although that means it would construct wires in the error case (to provide a legal lref for a read-only thing) whereas doing it at this level allows us to elide issuing the connection altogether.

Separate from that, this code is also pretty inelegant in that it's checking `.topBinding` for each left and right multiple times. This may be optimized by the JIT, but it's pretty inefficient if it isn't.

core/src/main/scala/chisel3/experimental/dataview/package.scala
```diff
@@ -222,15 +251,19 @@ package object dataview {
...
+        val aggWritability = Option.when(writability.isReadOnly)(
+          Map((agg: Data) -> writability)
+        )
+        agg.bind(AggregateViewBinding(fullResult, aggWritability))
     }
   }
 
   // Traces an Element that may (or may not) be a view until it no longer maps
   // Inclusive of the argument
+  // Note that this does *not* include writability so do not use this in place of reify
```
seldridge, line 263: Usual nit: full sentence comments. should be full sentences with punctuation.

seldridge (approved):
This approach sounds good to me. I don't have time to do a thorough review today, so please bug me for a more detailed post-merge review.

core/src/main/scala/chisel3/internal/Binding.scala
```diff
@@ -166,6 +236,20 @@ private[chisel3] object binding {
...
+  case class DeprecatedReadOnlyViewBinding(target: Data, getWarning: SourceInfo => Warning)
+      extends Binding
+      with ConditionalDeclarable {
+    def location: Option[BaseModule] = target.binding.flatMap(_.location)
+    def visibility: Option[WhenContext] = target.binding.flatMap {
+      case c: ConditionalDeclarable => c.visibility
+      case _ => None
+    }
+  }
+
```
jackkoenig, line 252:
```suggestion
```
This is unused.

PR #4193, opened 2024-06-20
Move registers out of when blocks

seldridge:
Change the way that Chisel IR creates registers to move these outside of when blocks.  This works towards a world where the only statements allowed under FIRRTL when blocks are connects.


seldridge:
We've decided against this in favor of enshrining this behavior in FIRRTL with https://github.com/chipsalliance/firrtl-spec/commit/b5bc360587703b370684aea8f795dfacf7665328.

PR #4198, opened 2024-06-20
Mark return value of .asTypeOf as "deprecated read-only"

jackkoenig:
Users will now get a deprecation warning if they connect to the result of .asTypeOf.

Marked for backporting to 6.x so that we can change this in 7.0.

### Contributor Checklist

#### Type of Improvement

- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

Previously, `.asTypeOf` would return a `Wire`. To get the old behavior, wrap the `.asTypeOf` call in `WireInit(...)`.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/Vec.scala
+++ b/src/test/scala/chiselTests/Vec.scala
@@ -319,9 +319,9 @@ class VecSpec extends ChiselPropSpec with Utils {
 
     })
     chirrtl should include("output io : { foo : UInt<1>, bar : UInt<1>[0]}")
-    chirrtl should include("wire zero : { foo : UInt<1>, bar : UInt<1>[0]}")
-    chirrtl should include("connect zero.foo, UInt<1>(0h0)")
-    chirrtl should include("connect io, zero")
+    chirrtl should include("wire _zero_WIRE : { foo : UInt<1>, bar : UInt<1>[0]}")
+    chirrtl should include("connect _zero_WIRE.foo, UInt<1>(0h0)")
+    chirrtl should include("connect io, _zero_WIRE")
     chirrtl should include("wire w : UInt<1>[0]")
     chirrtl should include("connect w, m.io.bar")
   }
--- a/core/src/main/scala/chisel3/Data.scala
+++ b/core/src/main/scala/chisel3/Data.scala
@@ -815,7 +815,9 @@ abstract class Data extends HasId with NamedComponent with SourceInfoDoc {
   def do_asTypeOf[T <: Data](that: T)(implicit sourceInfo: SourceInfo): T = {
     val thatCloned = Wire(that.cloneTypeFull)
     thatCloned.connectFromBits(this.asUInt)
-    thatCloned
+    thatCloned.viewAsReadOnlyDeprecated(siteInfo =>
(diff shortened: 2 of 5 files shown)


src/test/scala/chiselTests/Vec.scala
```diff
@@ -319,9 +319,9 @@ class VecSpec extends ChiselPropSpec with Utils {
 
     })
     chirrtl should include("output io : { foo : UInt<1>, bar : UInt<1>[0]}")
-    chirrtl should include("wire zero : { foo : UInt<1>, bar : UInt<1>[0]}")
-    chirrtl should include("connect zero.foo, UInt<1>(0h0)")
-    chirrtl should include("connect io, zero")
+    chirrtl should include("wire _zero_WIRE : { foo : UInt<1>, bar : UInt<1>[0]}")
+    chirrtl should include("connect _zero_WIRE.foo, UInt<1>(0h0)")
+    chirrtl should include("connect io, _zero_WIRE")
```
jackkoenig, line 324: This change to the .fir is unfortunate, I will look into this later (perhaps we need naming to contemplate views).

seldridge (approved):
LGTM

sequencer:
This PR introduced bug to T1. Related code snippet:
```
  val slotControl: Vec[InstructionControlRecord] =
    RegInit(
      VecInit(
        Seq.fill(parameter.chainingSize)(0.U.asTypeOf(new InstructionControlRecord(parameter)))
      )
    )
```
error message:
```
Caused by: chisel3.package$ChiselException: Connection between left (Lane.slotControl_?[0]: Wire[InstructionControlRecord]) and source (Lane.slotControl_?: Wire[InstructionControlRecord]) failed @.mask.bits: Locally unclear whether Left or Right (both internal)

```

jackkoenig:
Thanks for the report, looking into it!

jackkoenig:
This issue is caused by a weird interaction between how `VecInit` and directionality work. `VecInit` picks `<>` if the source is bidirectional[^1], which then due to the specialized logic to sometimes emit FIRRTL bulk connects, will emit a FIRRTL bulk connect in this case[^2]. This PR broke that because it is conservative and won't issue bulk connections if either side is marked read-only[^3]. That could change, we could still issue the bulk connect, but the real issue is actually https://github.com/chipsalliance/chisel/issues/4204. The direction of your `InstructionControlRecord` is being reported as `Bidirectional(Default)` even though it isn't bidirectional due to the bug in the linked issue. So I'm going to fix that.

[^1]: https://github.com/chipsalliance/chisel/blob/00c7a6288386b237be45fe471e67d9b9cc3192a4/core/src/main/scala/chisel3/Aggregate.scala#L646
[^2]: https://github.com/chipsalliance/chisel/blob/00c7a6288386b237be45fe471e67d9b9cc3192a4/core/src/main/scala/chisel3/internal/BiConnect.scala#L196
[^3]: https://github.com/chipsalliance/chisel/blob/00c7a6288386b237be45fe471e67d9b9cc3192a4/core/src/main/scala/chisel3/internal/BiConnect.scala#L133

PR #4200, opened 2024-06-21
Fix width of ChiselEnum values in emitted FIRRTL

jackkoenig:
This is primarily a bug fix, but it changes widths in a way that could silently break user code so I'm framing this as an API Modification that has instructions for users when bumping Chisel. In https://github.com/chipsalliance/chisel/pull/3824 we added `--use-legacy-shift-right-width` to help users migrate to the fixed width semantics for shift-right. I have renamed that option `--use-legacy-width` and added this width behavior change to the option.

`--use-legacy-width` is purely a code migration option, this behavior will only be "optionally preserved" for migrating code to Chisel 7, it will be removed in Chisel 8.

### Contributor Checklist

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

Fixes https://github.com/chipsalliance/chisel/issues/4159.

Previously, the width reported by Chisel under `.getWidth` was inconsistent with the width of the emitted FIRRTL for `ChiselEnum` values cast to `UInt`.

Temporarily preserve the old behavior under CLI option `--use-legacy-width` (formerly known as `--use-legacy-shift-right-width`). Users are encouraged to build Verilog with and without this option enabled and diff the result to verify that this change in width behavior did not silently affect the correctness of their designs.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/stage/ChiselAnnotations.scala
+++ b/src/main/scala/chisel3/stage/ChiselAnnotations.scala
@@ -362,17 +362,27 @@ object ChiselOutputFileAnnotation extends HasShellOptions {
   */
 case class DesignAnnotation[DUT <: RawModule](design: DUT) extends NoTargetAnnotation with Unserializable
 
-/** Use legacy Chisel shift-right behavior
+/** Use legacy Chisel width behavior.
   *
   * '''This should only be used for checking for unexpected semantic changes when bumping to Chisel 7.0.0'''
   *
-  * Use as CLI option `--use-legacy-shift-right-width`.
+  * Use as CLI option `--use-legacy-width`.
   *
-  * This behavior is inconsistent between Chisel and FIRRTL
-  * - Chisel will report the width of a UInt or SInt shifted right by a number >= its width as a 0-bit value
-  * - FIRRTL will implement the width for these UInts and SInts as 1-bit
+  * There are two width bugs fixed in Chisel 7.0 that could affect the semantics of user code.
+  *
+  * 1. The width of shift-right when shift amount is >= the width of the argument
+  *
+  * This behavior is inconsistent between Chisel and FIRRTL.
+  * - Chisel will report the width of a UInt or SInt shifted right by a number >= its width as a 0-bit value.
+  * - FIRRTL will implement the width for these UInts and SInts as 1-bit.
+  *
+  * 2. The width of `ChiselEnum` values
+  *
+  * This behavior is inconsistent between the width reported by Chisel and the FIRRTL emitted by Chisel.
+  * - Calling .getWidth of a ChiselEnum value will give the width needed to encode the enum.
+  * - The resulting FIRRTL will have the minimum width needed to encode the literal value for that enum value.
   */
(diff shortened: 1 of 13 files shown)


src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -362,28 +362,37 @@ object ChiselOutputFileAnnotation extends HasShellOptions {
...
   *
-  * This behavior is inconsistent between Chisel and FIRRTL
-  * - Chisel will report the width of a UInt or SInt shifted right by a number >= its width as a 0-bit value
-  * - FIRRTL will implement the width for these UInts and SInts as 1-bit
+  * There are two width bugs fixed in Chisel 7.0 that could affect the semantics of user code.
+  *
+  * 1. The width of shift-right when shift amount is >= the width of the argument
+  *
+  * This behavior is inconsistent between Chisel and FIRRTL.
+  * - Chisel will report the width of a UInt or SInt shifted right by a number >= its width as a 0-bit value.
```
mwachs5, line 376: I'm a bit confused about the tense in this doc. So what is the state of affaris after this PR is merged?

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -362,28 +362,37 @@ object ChiselOutputFileAnnotation extends HasShellOptions {
...
   *
   * '''This should only be used for checking for unexpected semantic changes when bumping to Chisel 7.0.0'''
   *
-  * Use as CLI option `--use-legacy-shift-right-width`.
+  * Use as CLI option `--use-legacy-width`.
   *
-  * This behavior is inconsistent between Chisel and FIRRTL
-  * - Chisel will report the width of a UInt or SInt shifted right by a number >= its width as a 0-bit value
-  * - FIRRTL will implement the width for these UInts and SInts as 1-bit
+  * There are two width bugs fixed in Chisel 7.0 that could affect the semantics of user code.
```
mwachs5, line 371:
```suggestion
  * There are two width bugs fixed in Chisel 7.0 that could affect the semantics of user code. Setting this argument continues to use the legacy buggy behavior
```

src/main/scala/chisel3/stage/ChiselAnnotations.scala
jackkoenig: This part of doc is the same as before, but this is a good opportunity to tighten this phrasing up (especially since this should probably be turned into a doc on the website).

The present tense here is describing what enabling the flag does.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
jackkoenig: I did something similar, see below.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
jackkoenig: I redid the ScalaDoc a little, hopefully it's more clear now.

mwachs5 (approved):

Issue #4218, opened 2024-06-25
Bulk connection emission for <> fails for "input" wires


jackkoenig:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**

Consider the following Chisel:
```scala
//> using scala "2.13.12"
//> using dep "org.chipsalliance::chisel:6.4.0"
//> using plugin "org.chipsalliance:::chisel-plugin:6.4.0"
//> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Xfatal-warnings", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"

import chisel3._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._
import _root_.circt.stage.ChiselStage

class MyBundle extends Bundle {
  val foo = Input(UInt(8.W))
}

class Foo extends Module {
  val producer = IO(Flipped(new MyBundle))
  val consumer = IO(new MyBundle)

  val w1, w2 = Wire(new MyBundle)
  w1 <> producer
  w2 <> w1
  consumer <> w2
}

object Main extends App {
  println(
    ChiselStage.emitSystemVerilog(
      gen = new Foo,
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

**What is the current behavior?**

This fails with:
```
Exception in thread "main" chisel3.package$ChiselException: Connection between left (Foo.w2: Wire[MyBundle]) and source (Foo.w1: Wire[MyBundle]) failed @.foo: Locally unclear whether Left or Right (both internal)
```

Now this seems obvious. Of course it fails, it has always failed for `<>` of wires right? Wellllll sort of. The special emission of FIRRTL bulk connect when possible actually makes this legal if `val foo` were `Flipped` instead of `Input`, and user code relies on this behavior sometimes. The reason is an overly pessimistic check here [[1]](https://github.com/chipsalliance/chisel/blob/2c5e185deffee52cd40e739da31b5f75e33103ba/core/src/main/scala/chisel3/internal/MonoConnect.scala#L299) that should only care about direction being Input if the `sink` is a port (so the fix is super easy).

I wouldn't bother even reporting this but https://github.com/chipsalliance/chisel/pull/4205 exposes some cases because it means the Bundle with `Flipped` now has `Input` direction (whereas previously it incorrectly had `Bidirectional(_)` direction.

[...]

PR #4219, opened 2024-06-25
Emit FIRRTL bulk connects even for "input" wires

jackkoenig:
Fixes https://github.com/chipsalliance/chisel/issues/4218

One can question the need for this, as it's just making `<>` not error in a few more cases. The main motivation is that https://github.com/chipsalliance/chisel/pull/4205 causes some existing uses of `<>` that used to work no longer work. It's an unnecessary breakage that is fixed by this change.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/BulkConnectSpec.scala
+++ b/src/test/scala/chiselTests/BulkConnectSpec.scala
@@ -112,4 +112,15 @@ class BulkConnectSpec extends ChiselPropSpec {
     chirrtl should include("connect out1[0], in1[1]")
     chirrtl should include("connect out1[1], in1[0]")
   }
+
+  property("Chisel should emit FIRRTL bulk connect for \"input\" wires") {
+    class MyBundle extends Bundle {
+      val foo = Input(UInt(8.W))
+    }
+    val chirrtl = ChiselStage.emitCHIRRTL(new Module {
+      val w1, w2 = Wire(new MyBundle)
+      w2 <> w1
+    })
+    chirrtl should include("connect w2, w1")
+  }
 }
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/BulkConnectSpec.scala
```diff
@@ -112,4 +112,15 @@ class BulkConnectSpec extends ChiselPropSpec {
...
+
+  property("Chisel should emit FIRRTL bulk connect for \"input\" wires") {
+    class MyBundle extends Bundle {
+      val foo = Input(UInt(8.W))
+    }
+    val chirrtl = ChiselStage.emitCHIRRTL(new Module {
+      val w1, w2 = Wire(new MyBundle)
+      w2 <> w1
+    })
+    chirrtl should include("connect w2, w1")
```
dtzSiFive, line 124: 💯

mwachs5: i never thought that `<>` was very sound for connecting two wires, so this is no worse than before 🤷‍♀️

jackkoenig: Until recently, I didn't know it even _could_ work for two wires. As I mention in https://github.com/chipsalliance/chisel/issues/4218, it does because of the special logic to sometimes emit FIRRTL bulk connects. So that work which was intended to be purely an optimization and not API visible actually extended the `<>` API in some funky ways.

This PR just makes it slightly more consistent in its funkiness 😇

mwachs5 (approved):

PR #4222, opened 2024-06-27
Name the targets of identity views

jackkoenig:
I'm calling this a bugfix because it fixes an issue exacerbated by https://github.com/chipsalliance/chisel/pull/4198, but it's also sort of a feature. The crux of the motivation for this change is visible in the change to `VecSpec`. Basically, if the user returns a view in lieu of the underlying target Data, the actual Data does not currently get named. This PR fixes it by having identity views forward their naming on to the underlying target. This PR is made much more important by https://github.com/chipsalliance/chisel/pull/4198 because this now comes up quite a bit in user code, `VecSpec` showing an example.

Future work is handling non-identity views because it is reasonable to forward naming in those cases as well, but perhaps with additional prefixing based on the "local" name of the target within the view Aggregate.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -18,7 +18,7 @@ import _root_.firrtl.AnnotationSeq
 import _root_.firrtl.renamemap.MutableRenameMap
 import _root_.firrtl.util.BackendCompilationUtilities._
 import _root_.firrtl.{ir => fir}
-import chisel3.experimental.dataview.{reify, reifySingleTarget}
+import chisel3.experimental.dataview.{reify, reifyIdentityView, reifySingleTarget}
 import chisel3.internal.Builder.Prefix
 import logger.{LazyLogging, LoggerOption}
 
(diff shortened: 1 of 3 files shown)


mikeurbach (approved):

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -903,6 +903,12 @@ private[chisel3] object Builder extends LazyLogging {
         case Clone(m: experimental.hierarchy.ModuleClone[_]) => namer(m.getPorts, prefix)
         case _ =>
       }
+    case (d: Data) =>
+      // Views are often returned in lieu of the original, so name the original (as appropriate)
+      // If a view but not identity, return the view and name it since it shows up in .toString and error messages
+      // TODO recurse on targets of non-identity views, perhaps with additional prefix from the view
```
jackkoenig, line 909:
```suggestion
      // Views are often returned in lieu of the target, so name the target (as appropriate).
      // If a view but not identity, return the view and name it since it shows up in .toString and error messages.
      // TODO recurse on targets of non-identity views, perhaps with additional prefix from the view.
```

PR #4226, opened 2024-06-27
Bundle elements ordering fix

### LINKED ISSUE #4223: asUInt packing inconsistent for Bundles with type parameters
**Type of issue**: Bug Report

The code below defines a Bundle with type parameters and four fields, all resolving to type `Bool()`.

Generating Verilog for this code:

```scala
import chisel3._

class Types {
  val D = Bool()
  val B = Bool()
}

class InBundle [T <: Types, U <: Data] (TypeC: U, TypeBD: T) extends Bundle {
  val a = Bool()
  val b = TypeBD.B
  val c = TypeC
  val d = TypeBD.D
}

class BundleTest extends Module() {
  val io = IO (new Bundle {
    val in1  =  Input(new InBundle (TypeBD = new Types, TypeC  = Bool()))
    val in2  =  Input(new InBundle (TypeC  = Bool(),    TypeBD = new Types))
    val out1 = Output(UInt(in1.getWidth.W))
    val out2 = Output(UInt(in2.getWidth.W))
  })
    io.out1 := io.in1.asUInt
    io.out2 := io.in2.asUInt
  }

object VerilogMain extends App {
  ChiselStage.emitSystemVerilogFile(new BundleTest)
}
```
results in that `a`, `b`, `c`, and `d` (all bools) are packed differently in `out1` and `out2` in the generated Verilog file:
```Verilog
assign io_out1 = {io_in1_d, io_in1_b, io_in1_c, io_in1_a}; // {d, b, c, a}
assign io_out2 = {io_in2_c, io_in2_d, io_in2_b, io_in2_a}; // {c, d, b, a}
```
 
Packing of Bundle fields depend on:

- the order of type definitions 
  - e.g. `Types.D` defined before `Types.B` in the code above
- the order in which parameters are assigned at the creation of the Bundle 
  - `TypeBD = new Types, TypeC = Bool()` gives different result than `TypeC = Bool(), TypeBD = new Types`
- the order of declaring the fields of the Bundle

Packing order should only be influenced by the order of the declaration of the Bundle fields. In the example above I expect both outputs to be assigned to `{a, b, c, d}`

- version: 6.3.0

**Other Information**

Interfacing with pure Verilog modules requires the field positions of a bundle to be controllable. Having to write an explicit mapping function from Bundle fields to bits of a bit vector shouldn't be needed.

adkian-sifive:
Fixes #4223. Straighten out the reversing of bundle elements by removing sorting. Add reversals to fix ordering and fix tests

### Contributor Checklist

#### Type of Improvement

- API Modification

#### Desired Merge Strategy

#### Release Notes

All `Bundle` elements will be ordered by the order of their declaration, including by inheritance.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -410,7 +410,7 @@ private[chisel3] object Converter {
           fir.Field(getRef(elt, info).name, fir.Flip, extractType(elt, false, info, checkProbe, true, typeAliases))
       }
       if (!t._isOpaqueType)
-        fir.BundleType(t._elements.toIndexedSeq.reverse.map { case (_, e) => eltField(e) })
+        fir.BundleType(t._elements.toIndexedSeq.map { case (_, e) => eltField(e) })
       else
         extractType(t._elements.head._2, childClearDir, info, checkProbe, true, typeAliases)
     }
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -410,7 +410,7 @@ private[chisel3] object Converter {
           fir.Field(getRef(elt, info).name, fir.Flip, extractType(elt, false, info, checkProbe, true, typeAliases))
       }
       if (!t._isOpaqueType)
-        fir.BundleType(t._elements.toIndexedSeq.map { case (_, e) => eltField(e) })
+        fir.BundleType(t._elements.toIndexedSeq.reverse.map { case (_, e) => eltField(e) })
       else
         extractType(t._elements.head._2, childClearDir, info, checkProbe, true, typeAliases)
     }
--- a/src/test/scala/chiselTests/BundleElementsSpec.scala
+++ b/src/test/scala/chiselTests/BundleElementsSpec.scala
@@ -44,13 +44,13 @@ class BpipDecoupled extends BpipOneField {
 class HasDecoupledBundleByInheritance extends Module {
   val out1 = IO(Output(new BpipDecoupled))
   assertElementsMatchExpected(out1)(
-    "bpipDecoupledDecoupled" -> _.bpipDecoupledDecoupled,
-    "bpipDecoupledVec" -> _.bpipDecoupledVec,
-    "bpipDecoupledSInt" -> _.bpipDecoupledSInt,
-    "bpipOneFieldTwo" -> _.bpipOneFieldTwo,
-    "bpipOneFieldOne" -> _.bpipOneFieldOne,
+    "bpipSuperTraitGood" -> _.bpipSuperTraitGood,
     "bpipTraitGood" -> _.bpipTraitGood,
-    "bpipSuperTraitGood" -> _.bpipSuperTraitGood
+    "bpipOneFieldOne" -> _.bpipOneFieldOne,
+    "bpipOneFieldTwo" -> _.bpipOneFieldTwo,
+    "bpipDecoupledSInt" -> _.bpipDecoupledSInt,
+    "bpipDecoupledVec" -> _.bpipDecoupledVec,
+    "bpipDecoupledDecoupled" -> _.bpipDecoupledDecoupled,
   )
 }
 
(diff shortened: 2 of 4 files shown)


src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -498,17 +545,17 @@ class BundleElementsSpec extends AnyFreeSpec with Matchers {
...
+      assertElementsMatchExpected(r)("a" -> _.a, "b" -> _.b)
     }
 
     class Module1 extends Module {
       val i = IO(Input(new Bundle1))
       val m0 = Module(new Module0)
       m0.i := i
       m0.o := DontCare
-      assertElementsMatchExpected(i)("b" -> _.b, "a" -> _.a)
+      assertElementsMatchExpected(i)( "a" -> _.a, "b" -> _.b)
```
adkian-sifive, line 558:
```suggestion
      assertElementsMatchExpected(i)("a" -> _.a, "b" -> _.b)
```

jackkoenig:
I want to clarify my statement that "The right order is the order of the fields in the definition of the Bundle". The order I'm talking about here is in the FIRRTL (and thus the Verilog). It is a well-known mild annoyance that the elements of a `Bundle` are stored in reverse order in the `elements` during elaboration. We should **not** change that, at least not lightly, as that is known and accounted for in lots of user code that iterates on elements of Bundles.

For fixing https://github.com/chipsalliance/chisel/issues/4223 I think we should stick to just fixing the issue where object creation order sometimes violates field order.

mwachs5 (commented):
make sure this cookbook is still valid whatever you do... my most bookmarked: https://www.chisel-lang.org/docs/cookbooks/cookbook#how-do-i-create-a-bundle-from-a-uint

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -207,7 +207,44 @@ class UsesBundleWithGeneratorField extends Module {
...
   out := DontCare
 
-  assertElementsMatchExpected(out)("superQux" -> _.superQux, "superFoo" -> _.superFoo)
+  assertElementsMatchExpected(out)(
+    "superFoo" -> _.superFoo,
+    "superQux" -> _.superQux
+  )
+}
+
+
```
mwachs5, line 216: seems like the cases we want to see are cross products of: 
a. vals in the constructor args
b. vals in the body
c. vals in superclass/traits bodies

We should have a clear story on the order of these and the order within each of a, b, c and then show that as expected they go in order with a few test cases.

mwachs5: before the PR, i think that `b` had expected behavior that we don't want to change in this PR except for weird interactions flagged by the bug

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -410,7 +410,7 @@ private[chisel3] object Converter {
           fir.Field(getRef(elt, info).name, fir.Flip, extractType(elt, false, info, checkProbe, true, typeAliases))
       }
       if (!t._isOpaqueType)
-        fir.BundleType(t._elements.toIndexedSeq.reverse.map { case (_, e) => eltField(e) })
+        fir.BundleType(t._elements.toIndexedSeq.map { case (_, e) => eltField(e) })
```
mwachs5, line 413: you can't make this change here because it is affecting all Records, not just Bundles. So even if you undo the reversing for Bundles, this change is now messing with records so .asUint on a record would have different behavior.

YOu could show this by taking the cookbook here https://www.chisel-lang.org/docs/cookbooks/cookbook#how-do-i-create-a-bundle-from-a-uint and rewriting it with a record and convincing yourself it doesn't work the same after this change

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -44,27 +44,27 @@ class BpipDecoupled extends BpipOneField {
 class HasDecoupledBundleByInheritance extends Module {
```
mwachs5, line 44: add more tests that mess with object creation order like have static objects, call-by-name, gens, etc

mwachs5:
it seems like part of the confusion here is that .asUInt is not actually documented in the firrtl spec in anyway how it is supposed to function.

src/test/scala/chiselTests/BundleElementsSpec.scala
```diff
@@ -469,16 +503,26 @@ class BundleElementsSpec extends AnyFreeSpec with Matchers {
...
+  "Bundle elements should be ordered in order of declaration" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new SimpleBundleElemOrder)
+    chirrtl should include("input in1 : { a : UInt<1>, b : UInt<8>}")
+    chirrtl should include("input in2 : { a : UInt<1>, b : UInt<8>}")
+
+    val chirrtl2 = ChiselStage.emitCHIRRTL(new SimpleBundleElemOrderByName)
+    chirrtl2 should include("input in0 : { a : UInt<8>, b : UInt<8>, c : UInt<8>}")
+    chirrtl2 should include("input in1 : { a : UInt<8>, b : UInt<8>, c : UInt<8>}")
+  }
+
```
jackkoenig, line 525: These are good, but it would also be nice to do some  assertElementsMatchExpected as that's the thing that changed (i.e. that you fixed) in this PR.

jackkoenig (approved):

PR #4228, opened 2024-06-28
Make isView safe outside of Builder context

jackkoenig:
Contemplating views in naming (https://github.com/chipsalliance/chisel/pull/4222) exposed a longstanding bug that I patched out recently in https://github.com/chipsalliance/chisel/pull/4147. Basically `isView` does not work outside of Builder contexts. Now it is safe.

We do have lots of tests for things like literals outside of Builder contexts, but they didn't trigger this because `ViewParent` is a `lazy val` and thus if _any_ Chisel elaboration has occurred on the same thread, then the problem doesn't manifest. Thus this "test" looks very strange, but by virtue of defining a literal assigned to a val first thing in a ScalaTest spec, it ensures no Chisel elaboration has run before when this line executes, and the naming plugin + #4222 ensures that this value is checked if it's a view.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/dataview/package.scala
+++ b/core/src/main/scala/chisel3/experimental/dataview/package.scala
@@ -273,7 +273,7 @@ package object dataview {
   }
 
   // Safe for all Data
-  private[chisel3] def isView(d: Data): Boolean = d._parent.contains(ViewParent)
+  private[chisel3] def isView(d: Data): Boolean = d._parent.exists(_ == ViewParent)
 
   /** Turn any [[Element]] that could be a View into a concrete Element
     *
--- a/src/test/scala/chiselTests/UIntOps.scala
+++ b/src/test/scala/chiselTests/UIntOps.scala
@@ -275,6 +275,11 @@ trait ShiftRightWidthBehavior { self: ChiselRunners =>
 
 class UIntOpsSpec extends ChiselPropSpec with Matchers with Utils with ShiftRightWidthBehavior {
 
+  // This is intentionally a val outside of any ScalaTest constructs to check that it is legal
+  // to create a literal outside of a Chisel context and *before* and Chisel contexts have been created
+  // in this thread.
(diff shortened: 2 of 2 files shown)


core/src/main/scala/chisel3/experimental/dataview/package.scala
```diff
@@ -273,7 +273,7 @@ package object dataview {
   }
 
   // Safe for all Data
-  private[chisel3] def isView(d: Data): Boolean = d._parent.contains(ViewParent)
+  private[chisel3] def isView(d: Data): Boolean = d._parent.exists(_ == ViewParent)
```
jackkoenig, line 276: This looks equivalent but because `contains` takes its argument by value, even if `_parent` is `None`, `ViewParent` must be initialized (it's a lazy val) which requires a Chisel context (because it's a Module)... yay side effects 🤮.

src/test/scala/chiselTests/UIntOps.scala
```diff
@@ -275,6 +275,11 @@ trait ShiftRightWidthBehavior { self: ChiselRunners =>
 
 class UIntOpsSpec extends ChiselPropSpec with Matchers with Utils with ShiftRightWidthBehavior {
 
+  // This is intentionally a val outside of any ScalaTest constructs to check that it is legal
+  // to create a literal outside of a Chisel context and *before* and Chisel contexts have been created
```
mwachs5, line 279: And-> any

mwachs5 (approved):
Not sure why the difference matters but looks good!

src/test/scala/chiselTests/UIntOps.scala
jackkoenig, line 279:
```suggestion
  // to create a literal outside of a Chisel context and *before* any Chisel contexts have been created
```

PR #4238, opened 2024-07-02
Share ChildBinding objects between siblings

jackkoenig:
It's hard to measure the benefit from this, but I used the following little snippet of Chisel to make a bundle of Bundles:
```scala
import chisel3._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._

  val a, b, c, d, e, f, g = UInt(8.W)
}

class Foo(n: Int) extends Module {
  val in = IO(Input(Vec(n, new MyBundle)))
  val out = IO(Output(Vec(n, new MyBundle)))

  out :#= in
}

object Main extends App {
  val n = args(0).toInt
  ChiselStage
    .emitCHIRRTLFile(
      gen = new Foo(n)
    )
    .toSeq
}
```
Then I built a fat assembly jar with Mill, and checked the heap size with `find_heap_bound`:

```
./firrtl/benchmark/scripts/find_heap_bound.py -vvv --start-size 4G --min-step 100M --context 2 -- -cp assembly.jar Main 200000
```

With this, I measured a ~2.5% memory reduction for this very Bundle-heavy code.

I had thought about a more aggressive change, where `ChildBinding` becomes a `case object` and we instead get the `parent` from `ref` (which does duplicate this information). However, using the above benchmark, I measured the exact same memory improvement. Since I cannot measure any benefit from the much more aggressive and hacky alternative approach, I think we should do the simpler thing which gets more-or-less the same benefit.

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

This reduces memory use by `n - 1` times 16-bytes for an Aggregate with `n` elements.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -215,8 +215,10 @@ sealed class Vec[T <: Data] private[chisel3] (gen: => T, val length: Int) extend
 
     val resolvedDirection = SpecifiedDirection.fromParent(parentDirection, specifiedDirection)
     sample_element.bind(SampleElementBinding(this), resolvedDirection)
+    // Share same object for binding of all children.
+    val childBinding = ChildBinding(this)
     for (child <- elementsIterator) { // assume that all children are the same
-      child.bind(ChildBinding(this), resolvedDirection)
+      child.bind(childBinding, resolvedDirection)
     }
 
     // Since all children are the same, we can just use the sample_element rather than all children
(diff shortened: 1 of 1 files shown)


jackkoenig:
I validated that the ChildBinding objects are **not** currently shared and that with this change they are now shared with the Eclipse MemoryAnalyzer (https://eclipse.dev/mat/). This changed the measured memory use per UInt child of the Bundles in my example above from 72 bytes shallow, 144 retained to 72 bytes shallow, 128 retained, which measures my expectation. I'm also impressed that MAT is able to figure out that the individual elements should not be dinged for retaining the object that they all share.

mwachs5 (approved):

jackkoenig:
I figured out how to benchmark a little bit better by avoiding serialization

```scala
import chisel3._
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._

  val a, b, c, d, e, f, g = UInt(8.W)
}

class Foo(n: Int) extends Module {
  val in = IO(Input(Vec(n, new MyBundle)))
  val out = IO(Output(Vec(n, new MyBundle)))

  out :#= in
}

object Main extends App {
  val n = args(0).toInt
  val phase = new chisel3.stage.phases.Elaborate
  val annos = Seq(
    chisel3.stage.ChiselGeneratorAnnotation(() => new Foo(n))
  )
  println(phase.transform(annos).size)
}
```

Using the same basic approach but small min-step and more context:
```
./firrtl/benchmark/scripts/find_heap_bound.py -vvv --start-size 4G --min-step 10M --context 5 -- -cp assembly.jar Main 200000
```

Without this change:

Xmx | Max RSS (MiB) | Wall Clock (s) | User Time (s)
-- | -- | -- | --
1G | 1344 | 7.09 | 56.45
920M | 1269 | 6.92 | 88.26
910M | 1220 | 7.61 | 90.95
900M | 1198 | 9.98 | 116.99
890M | - | - | -
870M | - | - | -

With this change:

Xmx | Max RSS (MiB) | Wall Clock (s) | User Time (s)
-- | -- | -- | --
1G | 1324 | 6.75 | 74.59
880M | 1202 | 6.86 | 58.98
870M | 1221 | 7.1 | 82.23
860M | 1164 | 9.19 | 98.01
850M | - | - | -
840M | - | - | -

Using the best Max RSS for each shows a memory reduction of 2.8%. Small, but measurable!

PR #4254, opened 2024-07-09
Optimize Record._elements to not duplicate VectorMap if possible

jackkoenig:
The code is a little janky, but it is a massive reduction in memory use.

For a typical Bundle with 7 UInt fields, this reduces the memory footprint from 1992 bytes to 1552 bytes (the duplicate VectorMap of only 7 elements is 440 bytes!), a 22% reduction. Combined with my other recent performance improvement PRs (#4251, #4252, #4253 but **excluding** #4242), this 1992 bytes is reduced to 1424 for a total reduction of 28.5% memory use for this typical Bundle with 7 UInt fields.

I suspect we can do better by getting rid of any Map here at all and instead just use an `Array[Data]` (where the String names of the fields are stored in the children Data themselves as they are already stored there). Whereas the VectorMap of 7 elements is 440 bytes, an Array of 7 elements is 48 bytes.

That being said, this is a much simpler change that imparts a huge benefit so is worth it in the meantime.

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

This reduces memory use of a typical bundle by 20%.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -1287,7 +1287,12 @@ abstract class Record extends Aggregate {
   // without having to recurse over all elements after the Record is
   // constructed. Laziness of _elements means that this check will
   // occur (only) at the first instance _elements is referenced.
-  // Also used to sanitize names and convert to more optimized VectorMap datastructure
+  // Also used to sanitize names and convert to more optimized VectorMap datastructure (if necessary)
+  // TODO We should figure out how to combine this with elements
+  // The main trick there is storing the sanitized names and making sure every Chisel API uses them
+  // Strawman proposal is to put sanitized name in the Slot refs of the children, and then replace
+  // _elements and elements with a single Array and deprecate elements as a public API
+  // This would also give an opportunity to stop storing elements in reverse order
   private[chisel3] lazy val _elements: VectorMap[String, Data] = {
     // Since elements is a map, it is impossible for two elements to have the same
     // identifier; however, Namespace sanitizes identifiers to make them legal for Firrtl/Verilog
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -1287,25 +1287,46 @@ abstract class Record extends Aggregate {
...
-          )
-        }
-        // namespace.name also sanitizes for firrtl, leave name alone for OpaqueTypes
-        val sanitizedName = namespace.map(_.name(name, leadingDigitOk = true)).getOrElse(name)
-        sanitizedName -> field
-    }.to(VectorMap) // VectorMap has O(1) lookup whereas ListMap is O(n)
+    val originalElements = elements
+    // Don't create a new map unless necessary, this is much more memory efficient in common case
+    // This is true if elements is not a VectorMap or if any names need sanitization
+    var needNewMap = !originalElements.isInstanceOf[VectorMap[_, _]]
```
mwachs5, line 1306: is there a reason why this has to be done with a `var` and also why we aren't just doing it something like:

```
def makeNewMap(newNames: Seq[String) =       newNames.view
        .zip(originalElements)
        .map { case (name, (_, data)) => name -> data }
        .to(VectorMap) // VectorMap has O(1) lookup whereas ListMap is O(n)

val newElements = originalElements match {
  case original: VectorMap[_, _] => {
    val newNames = ...
    val allNamesUnchanged = newNames == origNames
    if (allNamesUnchanged) original else makeNewMap(newNames)
  case _ => makeNewMap(newNames)
}

...? I guess you are trying to only iterate over it once ...?

jackkoenig: > I guess you are trying to only iterate over it once ...?

Correct, I'm trying to minimize the total number of traversals

adkian-sifive (approved):
LGTM, the changes make sense

PR #4275, opened 2024-07-12
Fix emitted FIRRTL for dynamic index of size 0 Vec

jackkoenig:
https://github.com/chipsalliance/chisel/pull/4268 introduced a subtle bug for when dynamically indexing Vecs of size 0. This is currently deprecated but obviously we need to emit legal FIRRTL until it becomes an error.

### Contributor Checklist

#### Type of Improvement

- Bugfix (but no release notes because this bug isn't released)

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Aggregate.scala
+++ b/core/src/main/scala/chisel3/Aggregate.scala
@@ -182,7 +182,8 @@ trait VecFactory extends SourceInfoDoc {
     implicit sourceInfo: SourceInfo
   ): UInt = {
     val w = (n - 1).bitLength
-    if (n <= 1) 0.U
+    if (n <= 1) WireInit(0.U) // Need the Wire otherwise we emit vec[0] which is illegal FIRRTL.
+    // Other cases do not need a Wire because the literal is truncated to fit.
     else if (idx.width.known && idx.width.get <= w) idx
     else if (idx.width.known) idx(w - 1, 0)
     else (idx | 0.U(w.W))(w - 1, 0)
--- a/src/test/scala/chiselTests/Vec.scala
+++ b/src/test/scala/chiselTests/Vec.scala
@@ -460,4 +460,18 @@ class VecSpec extends ChiselPropSpec with Utils {
     }))
     log should be("")
   }
+
+  property("Indexing a size 0 Vec should warn but also emit legal FIRRTL") {
+    val (log, chirrtl) = grabLog(emitCHIRRTL(new RawModule {
+      val vec = IO(Input(Vec(0, UInt(8.W))))
+      val idx = IO(Input(UInt(2.W)))
+      val out = IO(Output(UInt(8.W)))
+      out := vec(idx)
+    }))
+    log should include("Cannot extract from Vec of size 0.")
+    chirrtl should include("input vec : UInt<8>[0]")
+    chirrtl should include("wire _out_WIRE : UInt")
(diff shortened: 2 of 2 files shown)


core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -182,10 +183,15 @@ trait VecFactory extends SourceInfoDoc {
...
-    else (idx | 0.U(w.W))(w - 1, 0)
+    val name = "_vec_index_truncate"
+    // Give any truncation logic a name to make debugging easier
+    autoNameRecursively(name)(prefix(name) {
+      if (n <= 1) WireInit(0.U) // Need the Wire otherwise we emit vec[0] which is illegal FIRRTL.
+      // Other cases do not need a Wire because the literal is truncated to fit.
+      else if (idx.width.known && idx.width.get <= w) idx
+      else if (idx.width.known) idx(w - 1, 0)
+      else (idx | 0.U(w.W))(w - 1, 0)
+    })
```
seldridge, line 194: Nit: A match would be more readable, IMO.

seldridge: Also, the naming change would be better broken out. However, I'll allow it.

core/src/main/scala/chisel3/Aggregate.scala
```diff
@@ -182,10 +183,15 @@ trait VecFactory extends SourceInfoDoc {
...
   ): UInt = {
     val w = (n - 1).bitLength
-    if (n <= 1) 0.U
-    else if (idx.width.known && idx.width.get <= w) idx
-    else if (idx.width.known) idx(w - 1, 0)
-    else (idx | 0.U(w.W))(w - 1, 0)
+    val name = "_vec_index_truncate"
+    // Give any truncation logic a name to make debugging easier
+    autoNameRecursively(name)(prefix(name) {
+      if (n <= 1) WireInit(0.U) // Need the Wire otherwise we emit vec[0] which is illegal FIRRTL.
```
seldridge, line 189: Is it cleaner to use a zero-width, invalidated wire here?

seldridge (approved):
Generally looks fine.

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: > Nit: A match would be more readable, IMO.

Agreed, although I'm mostly keeping the logic untouched.

> Also, the naming change would be better broken out. However, I'll allow it.

That's fair, I think I'll back it out and then the diff will better justify my decision to not change it into a match 😇

core/src/main/scala/chisel3/Aggregate.scala
jackkoenig: Great question, I'll try it out. This wire is closer to what Chisel had been emitting, but that might be cleaner and not rely on something that probably should be illegal in FIRRTL anyway.

jackkoenig: I tried it, it seemed to work for the most part but caused some subtle failure in the QueueSpec integration tests. I'm not convinced it's a real issue, but I couldn't quickly figure out what's wrong so I'll just leave it alone for now. We want to get rid of this truncation anyway so I'm not _that_ concerned with making sure it's the most elegant.

PR #4278, opened 2024-07-15
[main] Enable MiMa for v7.0.0-M2

chiselbot:
Enable MiMa for v7.0.0-M2


jackkoenig:
No MiMa for pre-release

PR #4285, opened 2024-07-18
Add permissions to website deployment job

jackkoenig:
I think chipsalliance changed something about default repo settings and now GITHUB_TOKEN only defaults to read permissions. We need to be more explicit with permissions to ensure various automation can continue working. I noticed the website deployment started failing but it is likely that other automation jobs will need similar permissions (publishing release artifacts, release notes, etc.).

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,8 @@ jobs:
   deploy_website:
     name: Deploy Website
     runs-on: ubuntu-22.04
+    permissions:
+      contents: write
     needs: [all_tests_passed]
     # Only Deploy website on pushes to main, may change to a stable branch
     if: (github.event_name == 'push') && (github.ref_name == 'main')


bensternthal:
@jackkoenig So, I upgraded our Github organization yesterday (which would have happened automatically in a few weeks).

I did not change any permissions, but what I did change was to remove some of the folks who were admins over the entire CHIPS enterprise account. I did not expect this to cause issues for individual projects. 

To me I think this means I probably need to audit our organization and how it is setup. You should not need to be an admin over the entire org to administer a project.

Let me know if you need to be added back as as admin or if the above has solved the problem.

jackkoenig:
@bensternthal Thanks for the info!

I did some poking around and I don't think you did anything wrong or need to audit anything. I think this is just a setting that is stricter in Github Enterprise, see the screenshot below.

Screenshot 2024-07-24 at 1 17 27 PM

We used to have this set to `Read and write permissions` which, honestly, is not a great idea to have as your default setting. I think having to enable greater permissions per job that needs them is a good thing so I don't think any action is needed on your end (and I'm not sure if it's even possible to change this setting, I certainly wouldn't suggest doing so if it is!). Thank you for following up!

bensternthal:
@jackkoenig Thanks for tagging me in on this. I ran into a similar issue on the chipsalliance website github action and your response saved me quite a bit of time hunting around for the fix!

PR #4295, opened 2024-07-19
[firrtl] Emit commas for smem/cmem

seldridge:
Fix a bug where smem/cmem did not emit commas which are now mandatory in the FIRRTL spec.

h/t @dtzSiFive for identifying the problem/fix


dtzSiFive (approved):
LGTM, thanks!

PR #4298, opened 2024-07-22
Add property support to SRAMInterface-based SRAM

sequencer:
This PR adds Property support for `SRAMInterface`, instantiating the `SRAMOM` when using `SRAMInterface`. In our flow this is intend to replace the repl-sram and provide a unified Verilog for both synthesis tool and simulation tools.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes
add property to SRAMInterface

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -84,45 +84,15 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
 /** Description to the SRAM, encoded by the [[chisel3.properties]] API.
   * User can access it via CIRCT API.
   */
-@instantiable
-final class SRAMDescription extends Class {
-  val depth:           Property[BigInt] = IO(Output(Property[BigInt]()))
-  val width:           Property[Int] = IO(Output(Property[Int]()))
-  val masked:          Property[Boolean] = IO(Output(Property[Boolean]()))
-  val read:            Property[Int] = IO(Output(Property[Int]()))
-  val write:           Property[Int] = IO(Output(Property[Int]()))
-  val readwrite:       Property[Int] = IO(Output(Property[Int]()))
-  val maskGranularity: Property[Int] = IO(Output(Property[Int]()))
-  val hierarchy:       Property[Path] = IO(Output(Property[Path]()))
-
-  @public
-  val depthIn: Property[BigInt] = IO(Input(Property[BigInt]()))
-  @public
-  val widthIn: Property[Int] = IO(Input(Property[Int]()))
-  @public
-  val maskedIn: Property[Boolean] = IO(Input(Property[Boolean]()))
-  @public
-  val readIn: Property[Int] = IO(Input(Property[Int]()))
-  @public
-  val writeIn: Property[Int] = IO(Input(Property[Int]()))
-  @public
-  val readwriteIn: Property[Int] = IO(Input(Property[Int]()))
-  @public
-  val maskGranularityIn: Property[Int] = IO(Input(Property[Int]()))
-  @public
-  val hierarchyIn: Property[Path] = IO(Input(Property[Path]()))
-
-  depth := depthIn
-  width := widthIn
-  masked := maskedIn
-  read := readIn
-  write := writeIn
-  readwrite := readwriteIn
-  maskGranularity := maskGranularityIn
-  hierarchy := hierarchyIn
-}
-object SRAMDescription {
-  val definition: Definition[SRAMDescription] = Instantiate.definition(new SRAMDescription)
+final class SRAMDescription extends Bundle {
+  val depth:           Property[BigInt] = Property[BigInt]()
+  val dataWidth:       Property[Int] = Property[Int]()
+  val masked:          Property[Boolean] = Property[Boolean]()
+  val read:            Property[Int] = Property[Int]()
+  val write:           Property[Int] = Property[Int]()
+  val readwrite:       Property[Int] = Property[Int]()
+  val maskGranularity: Property[Int] = Property[Int]()
+  val hierarchy:       Property[Path] = Property[Path]()
 }
 
 /** A IO bundle of signals connecting to the ports of a memory, as requested by
@@ -171,7 +141,7 @@ class SRAMInterface[T <: Data](
   def underlying: Option[HasTarget] = _underlying
 
   /** SRAM Description */
-  val description: Property[ClassType] = SRAMDescription.definition.getPropertyType
+  val description = new SRAMDescription
 }
 
 /** A memory file with which to preload an [[SRAM]]
@@ -604,36 +574,22 @@ object SRAM {
       assignMask(firrtlReadwritePort.wmask, memReadwritePort.mask)
     }
 
-    // Hack to ScalaDoc Bug, see [[LTLIntrinsicInstanceMethodsInternalWorkaround]]
-    implicit class SRAMDescriptionInstanceMethods(underlying: Instance[SRAMDescription]) {
-      implicit val mg: internal.MacroGenerated = new chisel3.internal.MacroGenerated {}
-      def depthIn = underlying._lookup(_.depthIn)
-      def widthIn = underlying._lookup(_.widthIn)
-      def maskedIn = underlying._lookup(_.maskedIn)
-      def readIn = underlying._lookup(_.readIn)
-      def writeIn = underlying._lookup(_.writeIn)
-      def readwriteIn = underlying._lookup(_.readwriteIn)
-      def maskGranularityIn = underlying._lookup(_.maskGranularityIn)
-      def hierarchyIn = underlying._lookup(_.hierarchyIn)
-    }
-
     // Add metadata to the SRAM.
-    val descriptionInstance: Instance[SRAMDescription] = Instantiate(new SRAMDescription)
-    descriptionInstance.depthIn := Property(size)
-    descriptionInstance.widthIn := Property(tpe.getWidth)
-    descriptionInstance.maskedIn := Property(isVecMem)
-    descriptionInstance.readIn := Property(numReadPorts)
-    descriptionInstance.writeIn := Property(numWritePorts)
-    descriptionInstance.readwriteIn := Property(numReadwritePorts)
-    descriptionInstance.maskGranularityIn := Property(
+    val description = _out.description
+    description.depth := Property(size)
+    description.dataWidth := Property(tpe.getWidth)
+    description.masked := Property(isVecMem)
+    description.read := Property(numReadPorts)
+    description.write := Property(numWritePorts)
+    description.readwrite := Property(numReadwritePorts)
+    description.maskGranularity := Property(
       Option
         .when(isVecMem)(tpe match {
           case t: Vec[_] => t.sample_element.getWidth
         })
         .getOrElse(0)
     )
-    descriptionInstance.hierarchyIn := Property(Path(mem))
-    _out.description := descriptionInstance.getPropertyReference
+    description.hierarchy := Property(Path(mem))
 
     _out
   }
--- a/src/main/scala/chisel3/util/SRAM.scala
+++ b/src/main/scala/chisel3/util/SRAM.scala
@@ -1,19 +1,22 @@
 package chisel3.util
 
 import chisel3._
-
 import chisel3.internal.{Builder, NamedComponent}
 import chisel3.internal.binding.{FirrtlMemTypeBinding, SramPortBinding}
 import chisel3.internal.plugin.autoNameRecursively
 import chisel3.experimental.{OpaqueType, SourceInfo}
+import chisel3.experimental.hierarchy.{instantiable, public, Definition, Instance, Instantiate}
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -78,6 +81,45 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
   }
 }
 
+@instantiable
+/** Object Module for SRAM, used for storing metadata of an SRAM. */
+class SRAMOM extends Class {
+  val depth:           Property[Int] = IO(Output(Property[Int]()))
```
jackkoenig, line 87: The size of a memory is a BigInt, not an Int (mainly for use in simulation but still) so I think this should probably be a BigInt, do we support `Property[BigInt]` or does that matter @mikeurbach?

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -78,6 +81,45 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
   }
 }
 
+@instantiable
+/** Object Module for SRAM, used for storing metadata of an SRAM. */
+class SRAMOM extends Class {
```
jackkoenig, line 86: The term "Object Model" is a rocket-chip term, not a Chisel one, and I'm not keen to introduce it here. I think we should stick to something more generic like "Metadata" or "Properties", any thoughts @seldridge or @mikeurbach?

mikeurbach: The Property term is what we're using in Chisel 🤷 The Property docs have some nuggets that we could also use: "non-hardware information", and "domain-specific data models" show up in the introductory paragraphs.

src/main/scala/chisel3/util/SRAM.scala
mikeurbach: We do support Property[BigInt]. It's all the same (arbitrary precision signed integers) in FIRRTL, but we should probably use BigInt here if the size of the memory is BigInt.

src/main/scala/chisel3/util/SRAM.scala
mikeurbach: I have been thinking about this a bit in out of tree projects as well, and thinking about what it would look like if certain Modules were expected to have Objects of certain Classes attached.

I personally don't like the term "metadata", since in some teams/flows, this information can be as important as the RTL. But I do think it would be good to pick a name that something we can use to refer to these Objects we always attach to certain Modules. I'm not sure what a good name would be, but I'll use "Description" as an example since I've seen it downstream. So this could be the SRAMDescription, but we could also have other Models like a hypothetical ALUDescription or whatever.

mikeurbach: I will also leave this as food for thought, which was something we discussed quite a bit while working on the Object Model dialect in CIRCT. Yes, we did pick, that name, and yes, I'm agreeing with Jack that that's maybe not the name to use here.

> Your domain model is not your object model. Your domain model is not your abstract data type model. Your object model is not your abstract data type model.

In my mind, Properties and Objects are the building blocks for representing an object model. But when you say an SRAM has some "description"/"metadata"/"whatever", that is about your domain model, and the object model is what you project your domain model into.

https://www.sicpers.info/2018/03/why-inheritance-never-made-any-sense/

mikeurbach (commented):
I took a look through this, and I don't have any major concerns with the implementation.

This is opening an interesting door though...now that Chisel can support such information, what APIs merit having Objects built in within Chisel?  Should this always be added in SRAM, or maybe there is a subclass or some other class that users can pick if they want the description.

I'm also keenly aware that CIRCT produces this information as it compiles memories. We have actually recently added support for this same information to be added automatically within CIRCT and bubbled up to the compilation unit's top level for all SRAMS. So there are a couple things to think about:

* This shouldn't conflict with that, but it would be good to confirm
* This could potentially replace the CIRCT pass adding the information

The last point it a worthy goal, but the CIRCT pass has some...peculiar requirements. And due to Hyrum's Law, those have now become entrenched. To fully replace the existing CIRCT pass that's already adding basically this information, we'd need to work out all the weird uses downstream. Specifically, there is a behavior in CIRCT that the `hierarchy` field is the hierarchical path before any SRAM extraction has happened, whereas if this is implemented in Chisel, it will always point to the final hierarchical path of the SRAM. That's arguably the right thing, but I know of downstream users who depend on the previous, non-intuitive behavior.

src/main/scala/chisel3/util/SRAM.scala
sequencer: > In my mind, Properties and Objects are the building blocks for representing an object model. But when you say an SRAM has some "description"/"metadata"/"whatever", that is about your domain model, and the object model is what you project your domain model into.

This does make sense to me!

I'll change it to `SRAMDescription` in the following commit.

sequencer:
> This is opening an interesting door though...now that Chisel can support such information, what APIs merit having Objects built in within Chisel? Should this always be added in SRAM, or maybe there is a subclass or some other class that users can pick if they want the description.

In my team, I was proposing: 
- for each module, they should be serializable module and put all parameters and main class to Property, so we are planing provide each module the Serializable OM description;
- power information: register that need to be associated with Path, it will be collected and generate the upf;
- IP integration: interface and memory region definition that required by IP-XACT;

> To fully replace the existing CIRCT pass that's already adding basically this information, we'd need to work out all the weird uses downstream. Specifically, there is a behavior in CIRCT that the hierarchy field is the hierarchical path before any SRAM extraction has happened, whereas if this is implemented in Chisel, it will always point to the final hierarchical path of the SRAM. That's arguably the right thing, but I know of downstream users who depend on the previous, non-intuitive behavior.

I do agree we should replace the `CreateSiFiveMetadata` pass, and because the name, I believe SiFive is the only downstream to that pass. So maybe migration should be simple?

sequencer:
cc @debs-sifive and @jackkoenig,  I reverted #4364 in a8cdaf4: since I think exposing the `underlying` API to users is not a good idea, and may introduce some inconvenience for future API change. I added 09254b9 instead with a private API and only used by the OM feature. For any users need the SRAM path, they can always retrieve the the Path via OM API in CIRCT.
WDYT?

debs-sifive:
@sequencer thanks for looking at this — I think your approach is definitely safer and more user-friendly than using `underlying`! There's some discussion on our end about being able to see `hierarchy`, which is the field I need. I'm copy-pasting @albertchen-sifive 

> could we make SRAMDescription a bundle with just property fields instead of an Class? I don’t think we have anyway of doing a field lookup like Property[SRAMDescription].hierarchy for properties? but if SRAMDescription was a Bundle and we had val description: SRAMDescription in SRAMInterface then we could do sram.description.hierarchy

sequencer:
> could we make SRAMDescription a bundle with just property fields instead of an Class? I don’t think we have anyway of doing a field lookup like Property[SRAMDescription].hierarchy for properties? but if SRAMDescription was a Bundle and we had val description: SRAMDescription in SRAMInterface then we could do sram.description.hierarchy

If we are construction another standalone OM, e.g. `SomeCoreL1DCacheDescription: Class`, and only need a `Path`, it does makes sense. I think this flow can only add `Path` to `SomeCoreL1DCacheDescription.hierarchy` field, and append other fields, and do post-process, this reasonable. My question is what's the difference of using `hierarchy: Property[Path]` or `SRAMDescription: Class`? Since they are always used in the CIRCT side.

My concern is: a `Class` is a property for the whole SRAM metadata, and we only use the `description` after firtool compiling `mlirbc`. And I personally hope using this metadata to replace whole `repl-sram` pass and the `metadata` folder from circt output for better SRAM replacing experience.
Basically `Path` or `Class` are both OK in the flow designing, and for a standard ABI, I personally prefer using `Class` for `SRAM`, any other downstream user can always use the `SRAMDescription` to do the SRAM replacement flow. e.g. add their own pass find and match the `Class`.

jackkoenig:
> My question is what's the difference of using hierarchy: Property[Path] or SRAMDescription: Class? Since they are always used in the CIRCT side.

The problem is that there is currently no way in Chisel to access the fields of the `Property[SRAMDescription]`. I think this is an issue with our current `Property` API, not with this PR per se, but we need the specific `Property[Path]` from `SRAMDescription` in the short term. I'm going to open a PR with a test that shows the lacking API in `Property[_ <: Class]` to show what I mean and then we can tackle it.

> Basically Path or Class are both OK in the flow designing, and for a standard ABI, I personally prefer using Class for SRAM,

It's not `Class` vs. `Path`, it's `Class` vs. `Bundle` of the same properties that the `Class` had since it is possible to access fields of a `Bundle`. I do agree that `Class` is better though and maybe we can just make that work, see my coming PR showing what I mean.

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -121,7 +165,10 @@ class SRAMInterface[T <: Data](
   private[chisel3] var _underlying: Option[HasTarget] = None
 
   /** Target information for annotating the underlying SRAM if it is known. */
-  def underlying: Option[HasTarget] = _underlying
+  def underlying:                    Option[HasTarget] = _underlying
+  private val descriptionDefinition: Definition[SRAMDescription] = Instantiate.definition(new SRAMDescription)
+  private val descriptionType:       ClassType = descriptionDefinition.getClassType
+  val description:                   Property[ClassType] = Property[descriptionType.Type]()
```
jackkoenig, line 171: The `Definition` should not be a field of the `Bundle`, it should possibly be a field of the companion object (if we're saying there's only 1 definition for all `SRAMDescriptions`), or possibly an argument to the constructor. I'm thinking maybe we should go with argument to constructor because I suspect users are going to want to optionally add their own metadata to SRAMs which implies there may not be a single definition for all of them.

One thing to note is this is a very backwards incompatible change, anyone using 

We also want the description to be typed, `Property[SRAMDescription]`, otherwise this is weird and erased and not useful as a Chisel API (e.g. our use case of accessing the `hierarchy` field from it).

src/main/scala/chisel3/util/SRAM.scala
```diff
@@ -78,6 +81,47 @@ class MemoryReadWritePort[T <: Data](tpe: T, addrWidth: Int, masked: Boolean) ex
   }
 }
 
+/** Description to the SRAM, encoded by the [[chisel3.properties]] API.
+  * User can access it via CIRCT API.
+  */
+@instantiable
+private[chisel3] class SRAMDescription extends Class {
```
jackkoenig, line 88:
```suggestion
@instantiable
final class SRAMDescription extends Class {
```
This should be public and should be used in the type of the `description` property in the `SRAMInterface`.

jackkoenig:
Marked API Modification because this will break any use of `SRAMInterface` where the user instantiates it themselves rather than via the `SRAM` factory methods. They will have to hook up the new `description` field.

sequencer:
> The problem is that there is currently no way in Chisel to access the fields of the Property[SRAMDescription].

Yes, That's a problem, in our side, we just throw the entire Property to another Property rather than filter out some internal property LOL, the output Top OM might be verbosity though...

src/main/scala/chisel3/util/SRAM.scala
sequencer: > it should possibly be a field of the companion object

Sounds good, I will do it later. For user adding their own metadata, Explicitly instantiating the `SRAMDescription` and connect sounds dangerous, I think we should consider what to add and how to add. e.g. SRAM description string, Initial ROM, UPF fields, etc. 

> accessing the hierarchy field from it

yes if you need to access the `hierarchy` field, we need add new API to `Class` to extract it.

src/main/scala/chisel3/util/SRAM.scala
sequencer: Yes, if you are going to use `hierarchy: Property[Path]` in Chisel, this should be public, but if not, private it will be better.

jackkoenig:
I pushed a commit showing what I want to see, I thought we could get by with `.getField` but that doesn't seem to be defined on `Property[ClassType]` which seems like an oversight. I think if the test I pushed can pass, we're in good shape.

sequencer:
> I pushed a commit showing what I want to see, I thought we could get by with .getField but that doesn't seem to be defined on Property[ClassType] which seems like an oversight. I think if the test I pushed can pass, we're in good shape.

I see. this is another key feature I also need, <del>T1 has a fat OM due to this limitation</del>

sequencer:
> it does feel weird that classes are Module-like

I was confused by this for a long time. But this is the only way I made it successfully work😅

jackkoenig:
I changed `SRAMDescription` to a `Bundle` and it seems to work better.

jackkoenig (approved):
I'm happy with this using a `Bundle`, we can revisit if any issues come up though.

sequencer:
cc @jackkoenig, `Bundle` make OM API extensively tricky to design, thus I switch to `Class` instead, I believe we need more APIs for Class API. rather than using Bundle for storage.

PR #4300, opened 2024-07-22
Layers: Add support for inline convention.

dtzSiFive:
Add inline-layers, a kind of Layer that lowers to macro ifdefs, inline, in the resulting verilog.

- Rename the "bind layers" concept to "extract layers".
- An inline-layer, in contrast to extract-layers, do not have output directories. When determining the output directory of an extract-layer, skip over and ignore any parent 
- Deprecate the concept of a layer's convention in favour of two `Layer` configuration types: `LayerConfig.Extract` and `LayerConfig.Inline`. Update all code inspecting the convention of a layer to instead match against the configuration.
 
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Layer API has changed from convention as argument to constructing specific layer kind directly.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -19,20 +19,25 @@ object layer {
   /** Enumerations of different optional layer conventions.  A layer convention
     * says how a given layer should be lowered to Verilog.
     */
+  @deprecated("`Convention` is being removed in favor of `LayerConfig`", "Chisel 7.0.0")
   object Convention {
     sealed trait Type
 
-    /** Internal type used as the parent of all layers. */
-    private[chisel3] case object Root extends Type
-
     /** The layer should be lowered to a SystemVerilog `bind`. */
-    case object Bind extends Type
+    final case object Bind extends Type
   }
 
   sealed trait OutputDirBehavior
   final case object DefaultOutputDir extends OutputDirBehavior
   final case object NoOutputDir extends OutputDirBehavior
   final case class CustomOutputDir(path: Path) extends OutputDirBehavior
+
+  sealed trait LayerConfig
+  object LayerConfig {
+    final case class Extract(outputDirBehavior: OutputDirBehavior = DefaultOutputDir) extends LayerConfig
+    final case object Inline extends LayerConfig
+    private[chisel3] final case object Root extends LayerConfig
+  }
 
   /** A layer declaration.
     *
@@ -41,12 +46,20 @@ object layer {
     * @param _parent the parent layer, if any
     */
   abstract class Layer(
-    val convention:        Convention.Type,
-    val outputDirBehavior: OutputDirBehavior = DefaultOutputDir
+    val config: LayerConfig
   )(
     implicit _parent: Layer,
     _sourceInfo:      SourceInfo) {
     self: Singleton =>
+
+    @deprecated("`Convention` is being removed in favor of `LayerConfig`", "Chisel 7.0.0")
+    def this(
+      convention:        Convention.Type,
+      outputDirBehavior: OutputDirBehavior = DefaultOutputDir
+    )(
+      implicit _parent: Layer,
+      _sourceInfo:      SourceInfo
+    ) = this(LayerConfig.Extract(outputDirBehavior))
 
     /** This establishes a new implicit val for any nested layers. */
     protected final implicit val thiz: Layer = this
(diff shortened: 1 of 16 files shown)


src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -201,4 +201,37 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+    matchesAndOmits(ChiselStage.emitCHIRRTL(new Foo))(
+      """layer A, inline, "A" :""",
+      s"""layer B, inline, "A${sep}B" :""",
+      "define x = probe(a)",
+      "define y1 = probe(b)",
+      "define x = probe(c)",
+      "define y2 = probe(d)",
+      "define y3 = probe(e)"
+    )()
+  }
```
seldridge, line 236: This test is overkill for what it is checking. Can this PR instead add an inline layer D and check that this emits as expected?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1106,8 +1106,9 @@ private[chisel3] object Builder extends LazyLogging {
       def foldLayers(l: layer.Layer): Layer = {
         val children = layerAdjacencyList(l)
         val convention = l.convention match {
-          case layer.Convention.Bind => LayerConvention.Bind
-          case _                     => ???
+          case layer.Convention.Bind   => LayerConvention.Bind
+          case layer.Convention.Inline => LayerConvention.Inline
+          case _                       => ???
```
seldridge, line 1111: I know this is some old code I wrote. In a follow-up, removing the `???` and just letting this error with a default match error may be better.

seldridge (approved):

src/test/scala/chiselTests/LayerSpec.scala
dtzSiFive: Trimmed this down 👍 .

core/src/main/scala/chisel3/internal/Builder.scala
dtzSiFive: Replaced with covering all cases (add a new convention that isn't handled? -> compile error) and throwing an exception -- wasn't sure if this was the best error-reporting mechanism here or not?  Still want this to evaluate to `Convention`-type value on the non-error path.

(And untested since this isn't exposed outside that package, so should be unreachable...)

seldridge: This approach LGTM.

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -40,13 +26,7 @@ object layer {
...
     * @param _parent the parent layer, if any
     */
-  abstract class Layer(
-    val convention:        Convention.Type,
-    val outputDirBehavior: OutputDirBehavior = DefaultOutputDir
-  )(
-    implicit _parent: Layer,
-    _sourceInfo:      SourceInfo) {
-    self: Singleton =>
+  abstract sealed class Layer(implicit _parent: Layer, _sourceInfo: SourceInfo) {
```
seldridge, line 29: Scala nit: use of `abstract` on the three classes in this diff isn't accurate as there are no abstract members. You _can_ do this in Scala if you want to prevent a user from doing `new Layer` and force them to do `new Layer {}`. However, I don't think there is much benefit in that---particularly for the `BindLayer` and `InlineLayer` which are self-typed with `Singleton` which I think mandates that you can't construct them via `new`.

seldridge: This change is a hard API breakage. We can do this. It will just break code and should be communicated as such.

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -63,6 +43,22 @@ object layer {
...
+    final private[chisel3] def canWriteTo(that: Layer): Boolean = that match {
+      case null              => false
+      case _ if this == that => true
+      case _                 => this.canWriteTo(that.parent)
+    }
+  }
+
+  abstract class BindLayer(
+    val outputDirBehavior: OutputDirBehavior = DefaultOutputDir
+  )(
```
seldridge, line 56: Nit: consider just switching to `ExtractLayer` as opposed to `BindLayer` to make this more generic terminology-wise. We've also considered doing this for the "bind" convention in FIRRTL as well.

seldridge (commented):
This is a breaking change. However, using the type system like this is cleaner than using a `require`. We're at the 7.0 boundary and are free to break this in this way. If we are planning to do this, then it would be good to backport a change to 6.5.0 that adds the new layer stuff and deprecates the old one.

(Alternatively and circuitously, I'm considering if we want to support the ability for directories on inline layers to not break this. 🙃)

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -19,35 +19,48 @@ object layer {
...
-    case object Bind extends Type
+    final case object Bind extends Type
   }
 
   sealed trait OutputDirBehavior
   final case object DefaultOutputDir extends OutputDirBehavior
   final case object NoOutputDir extends OutputDirBehavior
   final case class CustomOutputDir(path: Path) extends OutputDirBehavior
 
+  sealed abstract class LayerConfig
```
seldridge, line 35: Nit: a `trait` is slightly cleaner here.

```suggestion
  sealed trait LayerConfig
```

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -19,35 +19,48 @@ object layer {
   /** Enumerations of different optional layer conventions.  A layer convention
     * says how a given layer should be lowered to Verilog.
     */
+  @deprecated
```
seldridge, line 22: The `@deprecated` annotation takes two optional arguments:

1. A string suggesting what the user should do instead.
2. A string indicating since what version this is deprecated.

Can the PR include something useful here that is actionable by a user?

```suggestion
  @deprecated("`Convention` is being removed in favor of `LayerConfig`", "Chisel 7.0.0")
```

Same comment for the deprecated auxiliary constructor.

seldridge (commented):
Approach looks great. 👍

dtzSiFive:
Woohoo! 🎉

PR #4303, opened 2024-07-22
Use Region for Layer Blocks

seldridge:
Change Chisel IR for `layer.block` operations to be region based instead
of token based.  This is a similar refactor that was done for when blocks
previously.

Remove complexity from the Chisel IR Converter which only existed to
facilitate converting token-based IR structures like WhenBegin/WhenEnd and
LayerBlockBegin/LayerBlockEnd.  Now that all of these structures have been
removed, there is no need for the added complexity.

Change the name of a function internal to the Chisel IR/FIRRTL IR
converter.

#### Metadata

This is structured as logical commits that may be reviewed independently.


jackkoenig (approved):
LGTM! Nice!

PR #4304, opened 2024-07-23
Create layer blocks on demand

seldridge:
Change `chisel3.layer.block` to create parent layers on demand and only error if the layer of the current layer block is not an ancestor of the desired layer.  This is a relaxation of the previous requirement that the current layer must be the parent of the desired layer.

The reason for this relaxation is to make future APIs easier to write. E.g., there is a desire to put asserts under layers automatically.  Part of such an API will need to automatically create layer blocks if the user wants to put all assertions in a layer that is not a child of the root layer.  Ergo, an API like what `block` was modified to be is necessary either for `block` or in this new assertion code.  It seemed better to make this available to all users and library authors.

Note: this is narrowly an API extension as things which used to not work will now work.

#### Release Notes

Extend `chisel3.layer.block` to work under any `block` with a layer that is an ancestor of the desired layer.

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -125,8 +125,8 @@ object layer {
     * @param layer the layer this block is associated with
     * @param thunk the Chisel code that goes into the layer block
     * @param sourceInfo a source locator
-    * @throws IllegalArgumentException if the layer of the currnet layerblock is
-    * not an ancestor of the desired layer
+    * @throws java.lang.IllegalArgumentException if the layer of the currnet
+    * layerblock is not an ancestor of the desired layer
     */
   def block[A](
     layer: Layer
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/Layer.scala
```diff
@@ -117,26 +117,48 @@ object layer {
...
+    * This function automatically creates parent layers as the layer of the
+    * current layerblock is an ancestor of the desired layer.  The ancestor may
+    * be the current layer which causes no layer block to be created.  (This is
+    * not a _proper_ ancestor requirement.)
+    *
     * @param layer the layer this block is associated with
     * @param thunk the Chisel code that goes into the layer block
     * @param sourceInfo a source locator
+    * @throws IllegalArgumentException if the layer of the currnet layerblock is
+    * not an ancestor of the desired layer
```
jackkoenig, line 129: ScalaDoc is complaining about not being able to find `IllegalArgumentException`, maybe `java.lang.IllegalArgumentException`?

jackkoenig (approved):
LGTM

core/src/main/scala/chisel3/Layer.scala
seldridge: 🧐 I'll try that, thanks!

PR #4307, opened 2024-07-24
[simulator] Enable all layers for simulation

seldridge:
Change Chisel's simulator to enable all layers that exist in the circuit.
Do this by copying over the layer bind files and including them in
compilation.

In terms of implementation, this works by copying over all files in the filelist (as before) and then copying over all ABI-specified `layers_` files. All files copied over are included in the compilation (which turns on all the layers). In order for `` `include `` directives to be properly evaluated, this adds include directories to common compilation options. These include directories map to the same option in either Verilator or VCS: `+incdir`.

#### Release Notes

Enable all layers during simulation. Previously, all layers were disabled.

### DIFF
--- a/src/main/scala/chisel3/simulator/package.scala
+++ b/src/main/scala/chisel3/simulator/package.scala
@@ -141,9 +141,11 @@ package object simulator {
         (abs, rel) match {
           case (abs, _) if movedFiles.contains(abs) =>
           case (abs, rel) =>
+            val dest = Paths.get(workspace.primarySourcesPath).resolve(rel)
+            dest.getParent.toFile.mkdirs
             Files.move(
               abs,
-              Paths.get(workspace.primarySourcesPath).resolve(rel)
+              dest
             )
             movedFiles += abs
         }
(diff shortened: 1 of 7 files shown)


seldridge:
@jackkoenig 
> but it would be cool if the user could customize which layers are included (defaulting to all)

I entirely agree! I'm intending to add this support in the next PR.

src/main/scala/chisel3/simulator/package.scala
```diff
@@ -127,30 +127,35 @@ package object simulator {
...
+        // Move the file into primarySourcesPath if it has not already been moved.
+        (abs, rel) match {
+          case (abs, _) if movedFiles.contains(abs) =>
+          case (abs, rel) =>
+            val dest = Paths.get(workspace.primarySourcesPath).resolve(rel)
+            dest.getParent.toFile.mkdirs
+            Files.move(
+              abs,
+              dest
+            )
```
jackkoenig, line 149: This is the nittiest of nits but I would make this 1 line, your changes seem to make it no longer need to be split.

seldridge: I noticed this and was relying on formatting the handle it, though it didn't fix it. I was very confused. Apparently the formatting won't touch this.

This is fixed here: eac5decc5

PR #4312, opened 2024-07-26
[simulator] Add LayerControl infra

seldridge:
Add some infrastructure for controlling which layers are enabled by
Chisel's simulator.  This is not currently exposed to the user, yet.  This
makes no changes to the existing behavior of all layers being enabled.

Add a function argument that can be used to choose (filter) which Verilog
files should be include in the build.  This is a _very_ low level API that
is being added to allow a user to enable bind convention layers when
writing a test.

Add the ability for the user of the EphemeralSimulator to control which
layers are enabled.

#### Metadata

This is broken into logical commits that can be reviewed independently.

### DIFF
--- a/src/main/scala/chisel3/simulator/LayerControl.scala
+++ b/src/main/scala/chisel3/simulator/LayerControl.scala
@@ -29,11 +29,6 @@ object LayerControl {
     override protected def shouldEnable(layerFilename: String) = true
   }
 
-  /** Disable all layers */
-  case object DisableAll extends Type {
-    override protected def shouldEnable(filename: String) = false
-  }
-
   /** Enable only the specified layers
     *
     * Nested layers should use a `.` as a delimiter.
@@ -41,10 +36,15 @@ object LayerControl {
     * @param layers a variadic list of layer names
     */
   case class Enable(layers: Layer*) extends Type {
-    private val re = {
-      val layersRe = layers.map(_.fullName.split("\\.").mkString("_")).mkString("|")
-      s"^layers_\\w+_($layersRe)\\.sv$$".r
+    private val _shouldEnable: String => Boolean = {
+      layers match {
+        case Nil => _ => false
+        case _ =>
+          val layersRe = layers.map(_.fullName.split("\\.").mkString("_")).mkString("|")
+          val re = s"^layers_\\w+_($layersRe)\\.sv$$".r
+          re.matches(_)
+      }
     }
-    override protected def shouldEnable(filename: String) = re.matches(filename)
+    override protected def shouldEnable(filename: String) = _shouldEnable(filename)
   }
 }
--- a/src/main/scala/chisel3/simulator/LayerControl.scala
+++ b/src/main/scala/chisel3/simulator/LayerControl.scala
@@ -1,0 +1,50 @@
+package chisel3.simulator
+
+import java.io.File
+import chisel3.layer.Layer
+
+/** Utilities for enabling and disabling Chisel layers */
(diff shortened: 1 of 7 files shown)


seldridge:
Thinking about this overnight... it might be better to only expose a generic `fileFilter(file: File): Boolean` in `svsim` and have the actual logic to create file filters _for layers_ on the Chisel side. I'm slightly worried about breaking the svsim/Chisel firewall which has been generally viewed as a good thing.

seldridge:
The most recent version switches changed this to make the `svsim` interface a basic file filter. This is exposed more ergonomically to Chisel users by allowing them to provide a `LayerControl` argument to the `EphemeralSimulator` so that layers can be enabled or disabled.

src/main/scala/chisel3/simulator/LayerControl.scala
```diff
@@ -0,0 +1,50 @@
...
+    override protected def shouldEnable(filename: String) = false
+  }
+
+  /** Enable only the specified layers
+    *
+    * Nested layers should use a `.` as a delimiter.
+    *
+    * @param layers a variadic list of layer names
+    */
+  case class Enable(layers: Layer*) extends Type {
```
rwy7, line 43:
```suggestion
  final case class Enable(layers: Layer*) extends Type {
```

seldridge: Made both `EnableAll` and `Enable` final in https://github.com/chipsalliance/chisel/pull/4312/commits/e15215f50e41de0b106004a1603cda7636548c1b.

seldridge:
https://github.com/chipsalliance/chisel/pull/4312/commits/e272775d9b218e7315e2bf86872cd209e3ca47c2 and https://github.com/chipsalliance/chisel/pull/4312/commits/602a938af35ff8d0380fc03f9a2c65517fa94eb2 remove `DisableAll` as this is the same thing as `Enable()`. The logic is updated for `Enable()` to make this actually reject all layers it sees. (This shouldn't use a regex, and should just return `false`.)

rwy7 (approved):

src/main/scala/chisel3/simulator/LayerControl.scala
```diff
@@ -0,0 +1,50 @@
...
+      layers match {
+        case Nil => _ => false
+        case _ =>
+          val layersRe = layers.map(_.fullName.split("\\.").mkString("_")).mkString("|")
+          val re = s"^layers_\\w+_($layersRe)\\.sv$$".r
+          re.matches(_)
+      }
+    }
+    override protected def shouldEnable(filename: String) = _shouldEnable(filename)
+  }
```
jackkoenig, line 49: This is minor and just a thought, but should there be a built in `EnableNone`? Could just be `val EnableNone = Enable()`, it's a little more obvious about intent when used in code. A case object named `EnableNone` would potentially show up better in ScalaDoc.

jackkoenig (approved):
This looks great

src/main/scala/chisel3/simulator/LayerControl.scala
seldridge: I wrote up the `val DisableAll = Enable()` at one point and didn't include it. I can add that back.

seldridge: Added `DisableAll` and tests here: https://github.com/chipsalliance/chisel/pull/4312/commits/30d3ab1fd16f1af0cace2dc60024b665aee15df9

PR #4317, opened 2024-07-29
Fix svsim DPI function template

uenoku:
This fixes two bugs in https://github.com/chipsalliance/chisel/pull/4286

* VCS seems to be unhappy with `initTestBenchScope;` so replace it with `initTestBenchScope();`. 
* `getBitWidth` exported DPI functions are used both for gettable/settable ports but #4286 only added DPI decl for getter so this PR fixes by removing a filter. 

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/Workspace.scala
+++ b/svsim/src/main/scala/Workspace.scala
@@ -245,9 +245,12 @@ final class Workspace(
       l(" svScope setScopeToTestBench();")
       for ((port, index) <- ports) {
       l(CreateFunctionForPort.createGetBitWidth(port.name))
-      }
-      for ((port, index) <- ports.filter(_._1.isGettable)) {
+        if (port.isGettable) {
       l(CreateFunctionForPort.createGetBits(port.name))
+        }
+        if (port.isSettable) {
+      l(CreateFunctionForPort.createSetBits(port.name))
+        }
       }
       l()
       l("int port_getter(int id, int *bitWidth, void (**getter)(uint8_t*)) {")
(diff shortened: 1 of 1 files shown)


mwachs5 (approved):

mwachs5:
is there a testcase we can add for this ?

svsim/src/main/scala/Workspace.scala
```diff
@@ -243,9 +243,11 @@ final class Workspace(
       l()
       l("extern \"C\" {")
       l(" svScope setScopeToTestBench();")
+      for ((port, index) <- ports) {
+      l(CreateFunctionForPort.createGetBitWidth(port.name))
+      }
       for ((port, index) <- ports.filter(_._1.isGettable)) {
       l(CreateFunctionForPort.createGetBits(port.name))
```
jackkoenig, line 250:
```suggestion
      for ((port, index) <- ports) {
      l(CreateFunctionForPort.createGetBitWidth(port.name))
        if (port.isGettable) {
      l(CreateFunctionForPort.createGetBits(port.name))
        }
```
Stylistically, I think these would be better in a single loop (it also would keep the functions for a given port next to each other).

uenoku:
> is there a testcase we can add for this ?

I tried but it looks tricky to reproduce the error in chisel repo. Internally we use svsim in a slightly different way and it looks there is no issue if we used `chisel3.simulator` directly.

svsim/src/main/scala/Workspace.scala
uenoku: Fused a setter and getter generation into a single loop.

jackkoenig (approved):

sequencer:
> I tried but it looks tricky to reproduce the error in chisel repo. Internally we use svsim in a slightly different way and it looks there is no issue if we used chisel3.simulator directly.

Our team do provide the VCS infrastructure for rocket-chip and T1. Do we need to adopt it for chisel?

uenoku:
> Our team do provide the VCS infrastructure for rocket-chip and T1. Do we need to adopt it for chisel?

No I don't think so. It was not backend specific and internally we use some wrapper for svsim that manually constructs `ModuleInfo` .

PR #4322, opened 2024-07-31
Add --remap-layers to ChiselStage

seldridge:
Add a new annotation/option to Chisel, `RemapLayer`, which is intended to
be used to allow a Chisel compilation to change which layers are used by
an upstream project's libraries in a downstream project.  The motivation
here is to fix mismatches between upstream and downstream when they have
different layers.

Add the ability for ChiselStage to globally remap a layer to another
layer.  This is intended to be used as a Chisel-level mechanism for a
downstream project to change the layers of all upstream projects.  This
may be necessary in order to produce compatible FIRRTL designs where the
upstream project and downstream project have different layers, but the
upstream project has put certain constructs (e.g., asserts) into
specific layers.

#### Release Notes

Add --remap-layers ChiselStage option for changing all layers in an upstream project.

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -129,11 +129,12 @@ object layer {
     * layerblock is not an ancestor of the desired layer
     */
   def block[A](
-    layer: Layer
-  )(thunk: => A
+    _layer: Layer
+  )(thunk:  => A
   )(
     implicit sourceInfo: SourceInfo
   ): Unit = {
+    val layer = Builder.layerMap.getOrElse(_layer, _layer)
     var layersToCreate = List.empty[Layer]
     var currentLayer = layer
     while (currentLayer != Builder.layerStack.head && currentLayer != Layer.Root) {
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -129,24 +129,24 @@ object layer {
     * layerblock is not an ancestor of the desired layer
     */
   def block[A](
-    _layer: Layer
-  )(thunk:  => A
+    layer: Layer
+  )(thunk: => A
   )(
     implicit sourceInfo: SourceInfo
   ): Unit = {
-    val layer = Builder.layerMap.getOrElse(_layer, _layer)
+    val _layer = Builder.layerMap.getOrElse(layer, layer)
     var layersToCreate = List.empty[Layer]
-    var currentLayer = layer
+    var currentLayer = _layer
     while (currentLayer != Builder.layerStack.head && currentLayer != Layer.Root) {
       layersToCreate = currentLayer :: layersToCreate
       currentLayer = currentLayer.parent
     }
     require(
       currentLayer != Layer.Root || Builder.layerStack.head == Layer.Root,
-      s"a layerblock associated with layer '${layer.fullName}' cannot be created under a layerblock of non-ancestor layer '${Builder.layerStack.head.fullName}'"
+      s"a layerblock associated with layer '${_layer.fullName}' cannot be created under a layerblock of non-ancestor layer '${Builder.layerStack.head.fullName}'"
     )
 
-    addLayer(layer)
+    addLayer(_layer)
 
     def createLayers(layers: List[Layer])(thunk: => A): A = layers match {
       case Nil => thunk
--- a/src/main/scala/chisel3/stage/ChiselAnnotations.scala
+++ b/src/main/scala/chisel3/stage/ChiselAnnotations.scala
@@ -406,12 +406,12 @@ case object UseLegacyWidthBehavior
   * This is intended to be used by a downstream Chisel project that is using an
   * upstream Chisel project which has different layers and the user would like
   * to align the upstream project with the downstream.
+  * @param oldLayer the old layer that should be remapped
+  * @param newLayer the new layer that the old layer should be replaced with
   */
 case class RemapLayer(oldLayer: Layer, newLayer: Layer) extends NoTargetAnnotation with ChiselOption with Unserializable
 
 object RemapLayer extends HasShellOptions {
-
-  private val re = "([\\w\\$.]+),([\\w\\$.]+)".r
 
   private def getLayer(name: String): Layer = try {
     Class.forName(name).getField("MODULE$").get(null).asInstanceOf[Layer]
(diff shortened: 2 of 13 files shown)


seldridge:
This also needs to remap layer colored probes.

seldridge:
This should also probably grow a feature for remapping to the root layer with the effect. This has overlap with CIRCT specialization flows, but is necessary in order to produce legal FIRRTL.

sequencer:
Should this feature goes to circt by adding some cmdline option?

sequencer:
another idea is we can simply append the package name of each layer to identify them(or provide an user defined override-able val instead.)

seldridge:
> Should this feature goes to circt by adding some cmdline option?

A variation of this will need to show up as a FIRRTL language feature. However, I don't want to have CIRCT handle corrections to illegal FIRRTL to make it legal. I want to have Chisel always be able to emit legal FIRRTL.

E.g., we need some kind of "layer alias" or "layer remapping" at an extmodule boundary:

```
circuit Foo:
  layer A:
  extmodule Bar:
    layer B as A:
  module Foo:
    inst bar of Bar    
```

In the above `Bar` was compiled to Verilog with a layer `Bar::B`. (Which means it has a bunch of optional functionality that can be enabled by including a file `layers_Bar_B.sv` in the compilation). From the FIRRTL side, we need to know how to enable `layer B` in relation to the existing layers that `Foo` has. The above would cause layers `A` and `Bar::B` to be enabled at the same time and allow for any intermixing of layer colors `A` and Bar::B`

> another idea is we can simply append the package name of each layer to identify them(or provide an user defined override-able val instead.)

I wanted a more Chisel language level way of doing this but couldn't come up with something. The problem is that there is no "global configuration" for Chisel where you can do this. Where could you put a `val` that would set this globally? It could be an `implicit` that is passed everywhere. However, that `implicit` has the possibility of being overridden and for you to get legal FIRRTL out of Chisel it can't be overridden. There's similar problems with things like `withLayer[A](thunk: => A)` which can be overridden by some upstream library.

All that said, _practically_, I don't think that this feature will be frequently used. I want most people to align on the upstream provided layers that Chisel has. Draft PR here: https://github.com/chipsalliance/chisel/pull/4323/files#diff-ef2aee2c3e90ffd56af995e76ac6573da3db4f939ea66b5c2f95756c29041504 These then give us a common place to put assertions for standard library generators or internal things. Users are then free to add more layers based on what they care about. (There is a missing feature to add a layer that is a parent of an _existing_ layer.)

seldridge:
> This also needs to remap layer colored probes.

Added in: fdfdff9aca5b3a83becab0a5bc8e7892576a9aca

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -129,11 +129,12 @@ object layer {
     * layerblock is not an ancestor of the desired layer
     */
   def block[A](
-    layer: Layer
-  )(thunk: => A
+    _layer: Layer
+  )(thunk:  => A
```
jackkoenig, line 133:
```suggestion
    layer: Layer
  )(thunk:  => A
```
I would keep this the same and use a different name for the value below. The argument is part of the public API (named arguments and ScalaDoc), and thus I think it's better to keep it as the nicer argument name.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -399,3 +400,54 @@ case object UseLegacyWidthBehavior
...
+  *
+  * This is intended to be used by a downstream Chisel project that is using an
+  * upstream Chisel project which has different layers and the user would like
+  * to align the upstream project with the downstream.
+  */
+case class RemapLayer(oldLayer: Layer, newLayer: Layer) extends NoTargetAnnotation with ChiselOption with Unserializable
+
+object RemapLayer extends HasShellOptions {
+
+  private val re = "([\\w\\$.]+),([\\w\\$.]+)".r
```
jackkoenig, line 414:
```suggestion
  private val Remap = "([\\w\\$.]+),([\\w\\$.]+)".r
```
Nit but IMO this should be UpperCamelCase and some sort of more meaningful name to make the use below more clear.

src/test/scala/circtTests/stage/ChiselStageSpec.scala
```diff
@@ -1231,6 +1245,17 @@ class ChiselStageSpec extends AnyFunSpec with Matchers with chiselTests.Utils {
       (ChiselStage.emitCHIRRTL(new RawModule {}) should not).include("LogLevelAnnotation")
       (ChiselStage.emitCHIRRTL(new RawModule {}, Array("-ll", "info")) should not).include("LogLevelAnnotation")
     }
+
+    it("should allow remapping one layer to another") {
+      val chirrtl = ChiselStage.emitCHIRRTL(
+        new ChiselStageSpec.LayerRemappingTest,
+        Array("--remap-layer", "circtTests.stage.ChiselStageSpec$A$,circtTests.stage.ChiselStageSpec$B$")
```
jackkoenig, line 1252: IMO, it would be better if we could make this work using the expected Scala package path of the object rather than the actual classname, ie. I would prefer to see this as:
```scala
Array("--remap-layer", "circtTests.stage.ChiselStageSpec.A,circtTests.stage.ChiselStageSpec.B")
```
We already require layers to be Singletons so there's no risk of ambiguity, we can insert and remove the `$` ourselves and just include them in error messages as appropriate (e.g. we could print the thing we couldn't find in case something is going wrong).

WDYT?

jackkoenig (commented):
Generally looks good, a couple of nits and a suggestion

src/test/scala/circtTests/stage/ChiselStageSpec.scala
seldridge: I agree. However, `.` is annoyingly ambiguous as any number of these could definitely be package or object separators (and could also be a literal `.`, maybe):

1. `.` 
2. `$`
3. `$u002E` (In the weird case of `` object `A.B` ``.)

I think an algorithm could work that tries all possibilities of, for the provided example:

1. `circtTests.stage.ChiselStageSpec.A` (Maybe the user means the class and made a mistake?)
1. `circtTests.stage.ChiselStageSpec.A$`
2. `circtTests.stage.ChiselStageSpec$A$`
3. `circtTests.stage$ChiselStageSpec$A$`
4. `circtTests$stage$ChiselStageSpec$A$`

However, it seemed weird to do this "guess and check" approach and instead require the user to just specify the exact object that they want. Given that this is intended to be a very advanced user API, I figured the precision was appreciated.

WDYT? Do you want/like the guess and check algorithm?

jackkoenig: Fair points, and we can also just go with this for now and add something more sugary later.

Also maybe Scala reflection gives something nicer here where the name munging is hidden under a more intentional API...

In any case, I think we should move forward with this and do that later.

jackkoenig (approved):
LGTM, but still see nits.

core/src/main/scala/chisel3/Layer.scala
seldridge: Good ABI catch! Fixed in 2778c92f0.

src/main/scala/chisel3/stage/ChiselAnnotations.scala
seldridge: I'm not used to seeing `val`s with upper camel case. However, I'll make this more descriptive and move it closer to its use site. Changed in: 167b1bebe

src/test/scala/circtTests/stage/ChiselStageSpec.scala
seldridge: Sounds good. I'll keep this as-is and we can revisit it later. I'm also very unconcerned about changing this advanced API in the future if we need to. Thanks for the discussion.

jackkoenig: Agreed on all counts! Thanks!

PR #4323, opened 2024-07-31
Add layer specification to LTL, verification ops

seldridge:
Put verification operations into layer blocks.

#### Release Notes

Put verification operations (asserts, assumes, covers, and stop) into Chisel-defined layers by default. Assertions are placed in `Verification.Assert`, assumptions are placed in `Verification.Assume`, covers are placed in `Verification.Cover`, and stop is placed in `Verification`. For users of ChiselSim/svsim, all layers are enabled by default, so no changes are necessary. For users that previously relied on these operations being enabled by default, either enable in-CIRCT specialization with `-enable-layers=Verification,Verification::Assert,Verification::Assume,Verification::Cover` or follow the FIRRTL ABI and include bind files for the layers that should be enabled.

### DIFF
--- a/core/src/main/scala/chisel3/VerificationStatement.scala
+++ b/core/src/main/scala/chisel3/VerificationStatement.scala
@@ -6,6 +6,8 @@ import chisel3.experimental.{BaseModule, SourceInfo}
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal._
 import chisel3.internal.firrtl.ir._
+import chisel3.layer.block
+import chisel3.layers
 import chisel3.util.circt.IfElseFatalIntrinsic
 import scala.annotation.nowarn
 import scala.language.experimental.macros
@@ -143,8 +145,10 @@ object assert extends VerifPrintMacrosDoc {
   )(
     implicit sourceInfo: SourceInfo
   ): Assert = {
-    val id = Builder.forcedUserModule // It should be safe since we push commands anyway.
-    IfElseFatalIntrinsic(id, format, "chisel3_builtin", clock, predicate, enable, format.unpackArgs: _*)(sourceInfo)
+    block(layers.Verification.Assert, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      val id = Builder.forcedUserModule // It should be safe since we push commands anyway.
+      IfElseFatalIntrinsic(id, format, "chisel3_builtin", clock, predicate, enable, format.unpackArgs: _*)(sourceInfo)
+    }
     new Assert()
   }
 
@@ -304,9 +308,11 @@ object assume extends VerifPrintMacrosDoc {
     implicit sourceInfo: SourceInfo
   ): Assume = {
     val id = new Assume()
-    when(!Module.reset.asBool) {
-      val formattedMsg = formatFailureMessage("Assumption", line, cond, message.map(Printable.pack(_, data: _*)))
-      Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, formattedMsg))
+    block(layers.Verification.Assume, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      when(!Module.reset.asBool) {
+        val formattedMsg = formatFailureMessage("Assumption", line, cond, message.map(Printable.pack(_, data: _*)))
+        Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, formattedMsg))
+      }
     }
     id
   }
@@ -320,10 +326,12 @@ object assume extends VerifPrintMacrosDoc {
     implicit sourceInfo: SourceInfo
   ): Assume = {
     val id = new Assume()
-    message.foreach(Printable.checkScope(_))
-    when(!Module.reset.asBool) {
-      val formattedMsg = formatFailureMessage("Assumption", line, cond, message)
-      Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, formattedMsg))
+    block(layers.Verification.Assume, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      message.foreach(Printable.checkScope(_))
+      when(!Module.reset.asBool) {
+        val formattedMsg = formatFailureMessage("Assumption", line, cond, message)
+        Builder.pushCommand(Verification(id, Formal.Assume, sourceInfo, Module.clock.ref, cond.ref, formattedMsg))
+      }
     }
     id
   }
@@ -385,8 +393,10 @@ object cover extends VerifPrintMacrosDoc {
     implicit sourceInfo: SourceInfo
   ): Cover = {
     val id = new Cover()
-    when(!Module.reset.asBool) {
-      Builder.pushCommand(Verification(id, Formal.Cover, sourceInfo, Module.clock.ref, cond.ref, ""))
+    block(layers.Verification.Cover, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      when(!Module.reset.asBool) {
+        Builder.pushCommand(Verification(id, Formal.Cover, sourceInfo, Module.clock.ref, cond.ref, ""))
+      }
     }
     id
   }
@@ -415,10 +425,12 @@ object stop {
 
   private def buildStopCommand(message: Option[Printable])(implicit sourceInfo: SourceInfo): Stop = {
     val stopId = new Stop()
-    message.foreach(Printable.checkScope(_))
-    when(!Module.reset.asBool) {
-      message.foreach(printf.printfWithoutReset(_))
-      pushCommand(Stop(stopId, sourceInfo, Builder.forcedClock.ref, 0))
+    block(layers.Verification, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      message.foreach(Printable.checkScope(_))
+      when(!Module.reset.asBool) {
+        message.foreach(printf.printfWithoutReset(_))
+        pushCommand(Stop(stopId, sourceInfo, Builder.forcedClock.ref, 0))
+      }
     }
     stopId
   }
--- a/src/main/scala/chisel3/ltl/LTL.scala
+++ b/src/main/scala/chisel3/ltl/LTL.scala
@@ -3,6 +3,7 @@
 package chisel3.ltl
 
 import chisel3._
+import chisel3.layer.{block, Layer}
 import chisel3.util.circt._
 import chisel3.experimental.hierarchy.{Instance, Instantiate}
 import chisel3.experimental.SourceInfo
@@ -391,7 +392,7 @@ object Property {
 /** The base class for the `AssertProperty`, `AssumeProperty`, and
   * `CoverProperty` verification constructs.
   */
-sealed abstract class AssertPropertyLike {
+sealed abstract class AssertPropertyLike(defaultLayer: Layer) {
 
   /** Assert, assume, or cover that a property holds.
     *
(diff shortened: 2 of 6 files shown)


src/main/scala/chisel3/ltl/LTL.scala
```diff
@@ -414,8 +415,10 @@ sealed abstract class AssertPropertyLike {
   )(
     implicit sourceInfo: SourceInfo
   ): Unit = {
-    val clocked = clock.fold(prop)(prop.clock(_))
-    createIntrinsic(label)(sourceInfo)(clocked.inner, disable.map(!_.value))
+    block(layers.Verification.Assert) {
+      val clocked = clock.fold(prop)(prop.clock(_))
+      createIntrinsic(label)(sourceInfo)(clocked.inner, disable.map(!_.value))
+    }
```
uenoku, line 421: I think this code path could create LTL assume/cover so I think they need to go to assume/cover layers.

dtzSiFive (commented):
Change LGTM!

Release notes should be updated (not just the Assert layer), and PR description.

Regarding:
> it is expected that asserts will not be put under
their corresponding layer.

What does that mean?

Anyway, please add at least one test to check this change in output (beyond the indirect btor test).

Should release notes (+label +milestone/target) make it more clear that this is a breaking change (worse than not compiling, your asserts/etc may now silently disappear if you don't know about this)?  Not sure what to suggest re:compatibility story for migration.

All users of these LTL operations will need to make sure they use flows that either specialize or pull in the right files, so if they expect/are used to just getting SV with what they asked for they'll need to adapt.   (right?)

Are there any in-tree comments or documentation appropriate so that this behavior is communicated to new folks using these?

src/main/scala/chisel3/ltl/LTL.scala
```diff
@@ -414,8 +415,10 @@ sealed abstract class AssertPropertyLike {
   )(
     implicit sourceInfo: SourceInfo
   ): Unit = {
-    val clocked = clock.fold(prop)(prop.clock(_))
-    createIntrinsic(label)(sourceInfo)(clocked.inner, disable.map(!_.value))
+    block(layer) {
+      val clocked = clock.fold(prop)(prop.clock(_))
+      createIntrinsic(label)(sourceInfo)(clocked.inner, disable.map(!_.value))
+    }
```
jackkoenig, line 421: I think it's important to handle the case (and we should show this in a test) of what happens if a user has their own layer and then creates an `AssertProperty` (or friend). As we discussed offline, I think that this `layer` should be treated as a default that only applies if the user hasn't specified a layer. WDYT?

Regardless, we need a test for that case (and for the default cases, just checking the emitted CHIRRTL is fine).

seldridge: The `layer.block` API is extended to provide this behavior here: https://github.com/chipsalliance/chisel/pull/4323/commits/2b80245b484cbeab6d5c542c2fd8d632c7769564

The LTL ops are updated to use it here: https://github.com/chipsalliance/chisel/pull/4323/commits/3a57da039b59ef7cb9f2ba0f337dc656d48a2adf These will no longer be put into the `Verification.Assert` block if the user decides to put them in _another_ block first.

seldridge:
> What does that mean?

I think the sentence is wrong. `s/not//`.

> Anyway, please add at least one test to check this change in output (beyond the indirect btor test).

Test added.

> Should release notes (+label +milestone/target) make it more clear that this is a breaking change (worse than not compiling, your asserts/etc may now silently disappear if you don't know about this)? Not sure what to suggest re:compatibility story for migration.
> 
> All users of these LTL operations will need to make sure they use flows that either specialize or pull in the right files, so if they expect/are used to just getting SV with what they asked for they'll need to adapt. (right?)

Previous commits make this _not_ a breaking change for normal users. Specifically asserts are still enabled for ChiselSim/svsim flows: https://github.com/chipsalliance/chisel/pull/4307 Additionally, any flow which is globbing for files is going to continue to work. If flows aren't globbing files, then they can glob files and continue to behave the same as before.

For other users, I expect that this is the exact right output that they want. I.e., there have been asks for `-extract-test-code` style emission from the SFC back when this was a SiFive-internal transform before it was made available through CIRCT. Additionally, for any user that is synthesizing designs, this cannot be breaking because the ops which are moved behind binds are not synthesizable.

> Are there any in-tree comments or documentation appropriate so that this behavior is communicated to new folks using these?

This should be updated! Good point. The docs for layers are relatively comprehensive (https://www.chisel-lang.org/docs/explanations/layers). The docs for assertions should be updated to indicate this.

seldridge:
I marked this as draft again as I was concerned about the ability to turn this off. 

However, with @youngar getting the layer specialization online, I think we are good here. A user who _doesn't_ want this can use `-enable-layer=Verification` `-enable-layer=Verification.Assert` `-enable-layer=Verification.Assume` `-enable-layer=Verification.Cover` to get the exact old emission behavior.

I think this also addresses the concerns from @dtzSiFive above as there is that salad of `-enable-layer` options can be used to turn this off. (There is already some use of this in tests: https://github.com/chipsalliance/chisel/pull/4323/files#diff-8174efeaf4221f2b797fc1f4b27899f2691366e4a37b7982251cb9b2276b6978R1178)

src/main/scala/chisel3/layers/Layers.scala
```diff
@@ -0,0 +1,16 @@
+package chisel3.layers
+
+import chisel3.layer.{Convention, Layer}
+
+/** The root [[chisel3.layer.Layer]] for all shared verification collateral. */
```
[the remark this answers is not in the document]
jackkoenig: Also importantly, note that when users use built-in DV constructs (like `AssertProperty`, not `chisel3.assert` yet but probably it should also be in a layer @seldridge), they will default into these layers[^1]. User code will be using these layers so they're pretty universal.

[^1]: Unless the user is creating the AssertProperty (and friends) in a layer block.

seldridge:
Does this do the right thing (what is the right thing) when one of these assertions is put under a module that has layers enabled?

sequencer:
Another related thing I want to propose is that: I'm thinking a design pattern that providing layers to be a module generator's parameter, i.e. `case class FooModuleP(assertLayer: Layer) extends SerializableModuleParameter`, and module can only use the `assertLayer` which already defined in the parameter, in this case, user can easily config the layer they want to use.

seldridge:
https://github.com/chipsalliance/chisel/pull/4323/commits/c5ed27272cd621a09bfc31b07d54fcf0c97658cd and https://github.com/chipsalliance/chisel/pull/4323/commits/78b1af0ca90caff56f57e5f08b3e5d84addd963f add escape hatches to `AssertPropertyLike.apply` to allow for users to _not_ put these statements inside layers. This is another escape hatch if users really want to opt out of this behavior without relying on specialization in CIRCT.

mwachs5 (commented):
should this PR include tests that ensure I can make my own MegansSpecialLayer_Assert layer and ensure that I am able to wrap `assert` such that it shows up in MegansSpecialLayer_Assert? Or are those tests already there and I am just missing them?

seldridge:
@mwachs5 wrote:
> should this PR include tests that ensure I can make my own MegansSpecialLayer_Assert layer and ensure that I am able to wrap assert such that it shows up in MegansSpecialLayer_Assert? Or are those tests already there and I am just missing them?

See: https://github.com/chipsalliance/chisel/pull/4323/files#diff-6ab973311ef3003e84ac5a740f734da637d081fbc5f9247da425eceb6f7fa763R253

If you put an assert under your own layer, it will be under that layer and not under the default layer.

seldridge:
The observed failures on this in the `LowerLayers` pass during integration tests will be addressed by the CIRCT 1.81.0 release.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -6,6 +6,8 @@ import chisel3.experimental.{BaseModule, SourceInfo}
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal._
 import chisel3.internal.firrtl.ir._
+import chisel3.layer.block
+import chisel3.layers
```
jackkoenig, line 10: It's sort of weird that we have both the `layer` and `layers` packages... Not really this PR but maybe we should just move the default layers to package `layer`? Idk, maybe it's just something to document. I do see utility in having the default layers in a separate package from the core `layer` APIs.

jackkoenig (approved):

core/src/main/scala/chisel3/VerificationStatement.scala
seldridge: Rare @ucbjrl "👍". 🐝 🍯 

Yeah, this can probably be collapsed down to a single package. The motivation was that `layer` is APIs for creating singular things: `object Foo extends layer.Layer` is creating _a_ layer or `layer.block(Foo)` is creating _a_ layer block. The `layers` package actually contains the plural layers. This isn't strong reasoning other than the imports vibed better in my head this way.

`layers` is going into the next major release, so this can change anytime.

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -659,6 +659,8 @@ class ProbeSpec extends ChiselFlatSpec with MatchesAndOmits with Utils {
       define(b.refs.reg, RWProbeValue(r))
     }
     runTester(new BasicTester {
+      layer.enable(layers.Verification)
+      layer.enable(layers.Verification.Assert)
```
seldridge, line 663: Without these, the movement of the assertions into bound modules will cause this test to fail. This test is _extremely_ brittle because there are both forces and asserts and these need to happen in a very specific order for everything to work. As soon as the always blocks that this code was producing are fractured, then the order of things is no longer deterministic. This may speak to `force` being a footgun of a Chisel API or something that needs to be reconsidered as it is extremely procedural. `force_initial` is probably fine, though.

I did not try this test with `-extract-test-code`. I expect that would produce the same problems.

src/test/scala/chiselTests/DataEqualitySpec.scala
```diff
@@ -22,7 +22,9 @@ class EqualityTester(lhsGen: => Data, rhsGen: => Data) extends BasicTester {
 
   assert(equalityModule.out)
 
-  stop()
+  when(RegNext(next = true.B, init = false.B)) {
+    stop()
+  }
```
seldridge, line 27: This is necessary because this test relies on the ordering of asserts and stops. With this PR, the assert and stop will be in separate always blocks which have no defined execution order. I expect that because `stop` goes into the `Verification` layer and `assert` goes into the `Verification.Assert` layer and that these are then bound in with `Verification` first and then `Verification.Assert` that a simulator will likely evaluate the `stop` before the `assert`. (There is obviously no guarantee of this, though.)

Adding an explicit delay to the `stop` of one cycle avoids this problem.

(This test was bad originally at it relied on assumptions about ordering of side effects which moving things into layers exposed.)

src/test/scala/chiselTests/ProbeSpec.scala
dtzSiFive: Apologies for the brittle test, it does indeed excessively rely on statement ordering. 😞 .

Empirically, sinking the asserts into a child instance appears to resolve the dependence on its ordering.
I wonder why bound-in child instance behaves differently?  Hopefully the spec makes this clear...

All verif-y things (including printf, assert, so on) are entirely unordered re:Chisel and FIRRTL semantics and any reliance on their ordering is problematic and will cause problems in practice today, I'm not sure this is an issue with `force` specifically.  This has always been a problem but so far we haven't had to care.

`force` definitely is weird, though, and alternatives for use cases that reach for force seem like a great idea to iterate on :+1: for when that's not explicitly what's required.  This would help with flows not going to SV!

A simple test with prints and stops will be emitted out-of-order (printf/stop/printf), even without layers, for example.

Increasing usage of procedural statements in FIRRTL will continue to expose the unreliability of their ordering until we develop mechanisms to order them predictably / as needed or find means to not rely on their ordering.

Thanks for wrangling this test!

PR #4337, opened 2024-08-09
Support BoringUtils.bore on OpaqueType wrapping a Property.

mikeurbach:
### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

#### Release Notes

This ensures boring from an OpaqueType that wraps a Property uses the correct connection operator in the IR.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -4,7 +4,7 @@ package chisel3
 
 import scala.util.Try
 import scala.language.experimental.macros
-import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
+import chisel3.experimental.{BaseModule, OpaqueType, SourceInfo, UnlocatableSourceInfo}
 import chisel3.internal._
 import chisel3.internal.binding._
 import chisel3.experimental.hierarchy.{InstanceClone, ModuleClone}
@@ -239,6 +239,10 @@ abstract class RawModule extends BaseModule {
       case (false, false)                               =>
         // For non-probe, directly create Nodes for lhs, skipping visibility check to support BoringUtils.drive.
         (left, right) match {
+          case (lhsOpaque: OpaqueType, rhsOpaque: OpaqueType)
+              if DataMirror
+                .isProperty(lhsOpaque.allElements.head) && DataMirror.isProperty(rhsOpaque.allElements.head) =>
+            PropAssign(si, Node(lhsOpaque.allElements.head), Node(rhsOpaque.allElements.head))
           case (_: Property[_], _: Property[_]) => PropAssign(si, Node(left), Node(right))
           // Use `connect lhs, read(probe(rhs))` if lhs is passive version of rhs.
           // This provides solution for this: https://github.com/chipsalliance/chisel/issues/3557
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -232,7 +232,7 @@ abstract class RawModule extends BaseModule {
         throwException(s"BoringUtils currently only support identity views, ${_right} has multiple targets.")
       )
 
-    def computeConnection(left: Data, right: Data, si: SourceInfo): Command = {
+    def computeConnection(left: Data, right: Data): Command = {
       (left.probeInfo.nonEmpty, right.probeInfo.nonEmpty) match {
         case (true, true)                                 => ProbeDefine(si, left.lref, Node(right))
         case (true, false) if left.probeInfo.get.writable => ProbeDefine(si, left.lref, RWProbeExpr(Node(right)))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -239,6 +239,10 @@ abstract class RawModule extends BaseModule {
       case (false, false)                               =>
         // For non-probe, directly create Nodes for lhs, skipping visibility check to support BoringUtils.drive.
         (left, right) match {
+          case (lhsOpaque: OpaqueType, rhsOpaque: OpaqueType)
+              if DataMirror
+                .isProperty(lhsOpaque.allElements.head) && DataMirror.isProperty(rhsOpaque.allElements.head) =>
```
mikeurbach, line 244: I'm not sure if there is a better way to query the underlying Data than `allElements.head`... if this is about the best we can do, maybe I can add an internal-only helper method to access this?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -239,6 +239,10 @@ abstract class RawModule extends BaseModule {
...
+            PropAssign(si, Node(lhsOpaque.allElements.head), Node(rhsOpaque.allElements.head))
```
jackkoenig, line 245:
```suggestion
          case (lhsOpaque: Record, rhsOpaque: Record)
              if lhsOpaque._isOpaqueType && rhsOpaque._isOpaqueType && DataMirror.isProperty=>
            secretConnection(lhsOpaque, rhsOpaque)
```
`OpaqueType` the trait isn't enough to check (and note the trait is self-typed by `Record`), instead we match on `Record` and check `_isOpaqueType`. Also this needs to be recursive in some way because you can have `OpaqueTypes` of `OpaqueTypes`. It's turtles all the way down!

core/src/main/scala/chisel3/RawModule.scala
jackkoenig: See [my comment](https://github.com/chipsalliance/chisel/pull/4337/files#r1712166891) on all 4 lines with suggestion.

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: Ah good point. What if DataMirror.isProperty handles _isOpaqueType and recursively checking? Then we could have one case here that just uses the `if DataMirror.isProperty` condition?

mikeurbach: Ugh I guess it's not that easy, this needs to recursively pick off the allElements.head till it gets to the bottom.

mikeurbach: I think [fa90b8f](https://github.com/chipsalliance/chisel/pull/4337/commits/fa90b8f5237aa9cf4cdd99700ecc4858ac7300ea) should handle it, this is now using _isOpaqueType and recursively unwrapping OpaqueTypes.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -231,24 +231,34 @@ abstract class RawModule extends BaseModule {
...
-          // Use `connect lhs, read(probe(rhs))` if lhs is passive version of rhs.
-          // This provides solution for this: https://github.com/chipsalliance/chisel/issues/3557
-          case (_, _)
-              if !DataMirror.checkAlignmentTypeEquivalence(left, right) &&
-                DataMirror.checkAlignmentTypeEquivalence(left, Output(chiselTypeOf(right))) =>
-            Connect(si, Node(left), ProbeRead(ProbeExpr(Node(right))))
-          case (_, _) => Connect(si, Node(left), Node(right))
-        }
+
+    def computeConnection(left: Data, right: Data, si: SourceInfo): Command = {
```
jackkoenig, line 235:
```suggestion
    def computeConnection(left: Data, right: Data): Command = {
```
Just delete this argument, you already capture `si` above since this is an inner method. The compilation error in CI is due to the fact that `lref` needs the implicit argument and by shadowing the name of the variable, it makes the `implicit si` above not available (lol, fixed by not needing to name `using` parameters in Scala 3)!

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -231,24 +231,34 @@ abstract class RawModule extends BaseModule {
...
+            case (_, _)
+                if !DataMirror.checkAlignmentTypeEquivalence(left, right) &&
+                  DataMirror.checkAlignmentTypeEquivalence(left, Output(chiselTypeOf(right))) =>
+              Connect(si, Node(left), ProbeRead(ProbeExpr(Node(right))))
+            case (_, _) => Connect(si, Node(left), Node(right))
+          }
+      }
     }
+
+    val rhs = computeConnection(left, right, si)
```
jackkoenig, line 261:
```suggestion
    val rhs = computeConnection(left, right)
```

jackkoenig (approved):
LGTM, see suggestions though to fix the compile issue.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -231,24 +231,34 @@ abstract class RawModule extends BaseModule {
...
+        case (true, false)                                => ProbeDefine(si, left.lref, ProbeExpr(Node(right)))
+        case (false, true)                                => Connect(si, left.lref, ProbeRead(Node(right)))
+        case (false, false)                               =>
+          // For non-probe, directly create Nodes for lhs, skipping visibility check to support BoringUtils.drive.
+          (left, right) match {
+            case (lhsOpaque: Record, rhsOpaque: Record)
+                if lhsOpaque._isOpaqueType && rhsOpaque._isOpaqueType && DataMirror.isProperty(
+                  lhsOpaque.allElements.head
+                ) && DataMirror.isProperty(rhsOpaque.allElements.head) =>
+              computeConnection(lhsOpaque.allElements.head, rhsOpaque.allElements.head, si)
```
jackkoenig, line 248:
```suggestion
              computeConnection(lhsOpaque.allElements.head, rhsOpaque.allElements.head)
```

core/src/main/scala/chisel3/RawModule.scala
mikeurbach: Oops, I didn't notice that in my local test somehow.

mikeurbach:
Oh, thanks for fixing the `si` shadowing @jackkoenig!

PR #4349, opened 2024-08-15
Make layer.addLayer API public

seldridge:
Make the `chisel3.layer.addLayer` API public.  This is useful if a user wants to guarantee that a layer will show up in CHIRRTL even if no layer block refers to this layer.

This solves a problem where downstream build flows want to expect that certain layers will always exist.  However, without this API, there is no way to do this other than to create dummy layer blocks.  This is slightly cleaner without the empty layer block trick.  Additionally, there is no real harm to having this API public.

#### Release Notes

Add `chisel3.layer.addLayer` API.  This can be used to cause a layer to be emitted in FIRRTL even if it has no layer block users.  This API is added (it was private before) so that users can make their custom layers always be there.  This is important for downstream build flows which may want to enable layers or specialize them away and having the unpredictability of whether or not a layer will exist is not ideal and complicates the build system.

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -127,10 +127,14 @@ object layer {
     implicit val root: Layer = Root
   }
 
-  /** Add a layer and all of its parents to the Builder.  This lets the Builder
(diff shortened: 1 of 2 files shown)


dtzSiFive (approved):
LGTM.

PR #4359, opened 2024-08-19
Fix Arg.name and earlyLocalName for probes

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- Bugfix

Names for probe Args will be the default `toString`. For example, if you use `DataMirror` to get the `earlyName` of a probe (derived from `Arg.name`), you will see this default `toString`.

```scala
import chisel3._
import chisel3.probe._

class ProbedClockedData[D <: Data](gen: => D) extends Bundle {
  val dataProbe = Probe(gen)
  def data      = probe.read(dataProbe)

  val clockProbe = Probe(Clock())
  def clock      = probe.read(clockProbe)

  val resetProbe = Probe(Reset())
  def reset      = probe.read(resetProbe)

  def tapAndConnect(ref: D, clock: Clock, reset: Reset) = {
    dataProbe  := BoringUtils.tap(ref)
    clockProbe := BoringUtils.tap(clock)
    resetProbe := BoringUtils.tap(reset)
  }
}

class Dut extends Module {
  val io = IO(new Bundle {
    val in  = Input(UInt(4.W))
    val out = Output(UInt(4.W))
  })
}

class TestHarness extends Module {
  val dut    = Module(new Dut)
  val probed = Wire(new ProbedClockedData(chiselTypeOf(dut.io)))
  probed.tapAndConnect(dut.io, dut.clock, dut.reset)

  println(s"probed.data name: ${DataMirror.queryNameGuess(probed.data)}")
}

object Main extends App {
  println(ChiselStage.emitCHIRRTL(gen = new TestHarness))
}
```

```
probed.data name: Node(TestHarness.probed.dataProbe: Wire[AnonymousBundle])
```

If that name is ever used for an identifier, Chisel will throw an exception due to the invalid name.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -262,7 +262,7 @@ private[chisel3] object ir {
 
   sealed trait ProbeDetails { this: Arg =>
     val probe: Arg
-    override def name: String = s"$probe"
+    override def name: String = s"${probe.name}"
   }
   case class ProbeExpr(probe: Arg) extends Arg with ProbeDetails
   case class RWProbeExpr(probe: Arg) extends Arg with ProbeDetails
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -101,6 +101,9 @@ private[chisel3] object ir {
       case Some(Index(Node(imm), arg))    => s"${earlyLocalName(imm, includeRoot)}[${arg.localName}]"
       case Some(Slot(Node(imm), name))    => s"${earlyLocalName(imm, includeRoot)}.$name"
       case Some(OpaqueSlot(Node(imm)))    => s"${earlyLocalName(imm, includeRoot)}"
+      case Some(ProbeExpr(Node(ref)))     => s"probe(${earlyLocalName(ref, includeRoot)})"
+      case Some(RWProbeExpr(Node(ref)))   => s"rwprobe(${earlyLocalName(ref, includeRoot)})"
+      case Some(ProbeRead(Node(ref)))     => s"read(${earlyLocalName(ref, includeRoot)})"
       case Some(arg) if includeRoot       => arg.name
       case None if includeRoot =>
         id match {
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -101,9 +101,9 @@ private[chisel3] object ir {
       case Some(Index(Node(imm), arg))    => s"${earlyLocalName(imm, includeRoot)}[${arg.localName}]"
       case Some(Slot(Node(imm), name))    => s"${earlyLocalName(imm, includeRoot)}.$name"
       case Some(OpaqueSlot(Node(imm)))    => s"${earlyLocalName(imm, includeRoot)}"
-      case Some(ProbeExpr(Node(ref)))     => s"probe(${earlyLocalName(ref, includeRoot)})"
-      case Some(RWProbeExpr(Node(ref)))   => s"rwprobe(${earlyLocalName(ref, includeRoot)})"
-      case Some(ProbeRead(Node(ref)))     => s"read(${earlyLocalName(ref, includeRoot)})"
+      case Some(ProbeExpr(Node(ref)))     => s"${earlyLocalName(ref, includeRoot)}"
+      case Some(RWProbeExpr(Node(ref)))   => s"${earlyLocalName(ref, includeRoot)}"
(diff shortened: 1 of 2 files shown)


mikeurbach:
I think this makes sense, but can you please add a unit test like the example in the PR description that shows the issue and confirms we don't regress in the future?

dtzSiFive:
Thanks for identifying and chasing this down!!

jackkoenig:
LGTM but agree it needs a test. Not sure of the best place for this specific test, something in ProbeSpec would be fine or it could be treated as a DataMirror check like here: https://github.com/chipsalliance/chisel/blob/da4cb2dc6ea72c8aaba64ca135ecb98dc2541440/src/test/scala/chiselTests/reflect/DataMirrorSpec.scala#L141

tmckay-sifive:
I added a simple unit test. But, there may be more to this. This test

https://github.com/tmckay-sifive/chisel/blob/02f98df3c2c47a087b47ee91b0321a2435cd1727/src/test/scala/chiselTests/ProbeSpec.scala#L370-L389

is failing with

```
"signalName/pathName should be called after circuit elaboration: ProbeSpec_Anon.w: Wire[Bool], Some(chiselTests.ProbeSpec$$anon$29@6780e0a6)" did not include substring "Cannot define a probe on a non-equivalent type." (ProbeSpec.scala:384)
```

jackkoenig:
Lol, looks like that test was expecting the previous bad output, you should just update that test!

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -383,7 +383,7 @@ class ProbeSpec extends ChiselFlatSpec with MatchesAndOmits with Utils {
     }
     exc.getMessage should include("Cannot define a probe on a non-equivalent type.")
     exc.getMessage should include(
-      "Left (ProbeSpec_Anon.p: IO[UInt<4>]) and Right (ProbeSpec_Anon.Node(ProbeSpec_Anon.w: Wire[Bool]): OpResult[Bool]) have different types"
+      "Left (ProbeSpec_Anon.p: IO[UInt<4>]) and Right (ProbeSpec_Anon.probe(w): OpResult[Bool]) have different types"
```
tmckay-sifive, line 386: This test was failing due to missing cases in `earlyLocalName` for probes. I added cases for them, and this error message seems sensible to me now.

src/test/scala/chiselTests/ProbeSpec.scala
```diff
@@ -733,4 +733,13 @@ class ProbeSpec extends ChiselFlatSpec with MatchesAndOmits with Utils {
...
     )()
   }
+
+  "Probes" should "have valid names" in {
+    class TestMod extends RawModule {
+      val a = IO(Output(Probe(UInt())))
+      val w = WireInit(UInt(32.W), 0.U)
+      a := w
+      require(reflect.DataMirror.queryNameGuess(a) == "a")
+    }
```
jackkoenig, line 743:
```suggestion
    class TestMod extends RawModule {
      val a = IO(Output(Probe(UInt())))
      val w = WireInit(UInt(32.W), 0.U)
      a := w
      require(reflect.DataMirror.queryNameGuess(a) == "a")
    }
    ChiselStage.emitCHIRRTL(new TestMod)
```
You have to actually run elaboration for this test to do anything. This is why I always make sure my unit tests fail before they pass (e.g. change the check to say it's equal to `"b"`, see it fail, then fix it).

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -101,6 +101,9 @@ private[chisel3] object ir {
       case Some(Index(Node(imm), arg))    => s"${earlyLocalName(imm, includeRoot)}[${arg.localName}]"
       case Some(Slot(Node(imm), name))    => s"${earlyLocalName(imm, includeRoot)}.$name"
       case Some(OpaqueSlot(Node(imm)))    => s"${earlyLocalName(imm, includeRoot)}"
+      case Some(ProbeExpr(Node(ref)))     => s"probe(${earlyLocalName(ref, includeRoot)})"
+      case Some(RWProbeExpr(Node(ref)))   => s"rwprobe(${earlyLocalName(ref, includeRoot)})"
+      case Some(ProbeRead(Node(ref)))     => s"read(${earlyLocalName(ref, includeRoot)})"
```
jackkoenig, line 106:
```suggestion
      case Some(ProbeExpr(Node(ref)))     => s"${earlyLocalName(ref, includeRoot)}"
      case Some(RWProbeExpr(Node(ref)))   => s"${earlyLocalName(ref, includeRoot)}"
      case Some(ProbeRead(Node(ref)))     => s"${earlyLocalName(ref, includeRoot)}"
```
Probe information is not part of the name

jackkoenig (approved):
Looks great, thanks Trevor!

PR #4361, opened 2024-08-20
[ci] Try to enable binder tests

SpriteOvO:


jackkoenig:
Why is this PR against nightly instead of main?

SpriteOvO:
The required C-API was still in nightly when this PR was created, but now we can change it to be based on `main`. :+1:

jackkoenig:
I'm unsure what Scala-CLI issue you're running into, but I've had issues with Scala-CLI hanging in Github Actions: https://github.com/VirtusLab/scala-cli/issues/3054

Also https://github.com/VirtusLab/scala-cli/issues/3013 about Scala CLI downloading its own Java (although that has to do with the scala-cli action so may not apply).

SpriteOvO:
> I'm unsure what Scala-CLI issue you're running into

scala-cli complains `java.nio.file.FileAlreadyExistsException: /home/runner/.cache/scalacli/local-repo/.1.5.0.tmp/version`.

rm this file before running mill doesn't help, there seems to be another process creating this file. It's not even fully reproducible, it just happens randomly. 🤦

PR #4368, opened 2024-08-27
Change Path to emit relative targets

jackkoenig:
Chisel will now reject attempts to create a Path within a Module that is not a direct ancestor of the target. Note that such paths are already rejected by firtool.

I'm marking this as API Modification since it is a pretty big behavioral change but I don't think the change is actually observable by users unless they are doing things with the paths in the `.fir`.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash
- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/properties/Path.scala
+++ b/core/src/main/scala/chisel3/properties/Path.scala
@@ -2,7 +2,7 @@
 
 package chisel3.properties
 
-import chisel3.{Data, HasTarget, MemBase, SramTarget}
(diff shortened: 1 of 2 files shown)


jackkoenig:
FWIW, in my opinion, Paths should always be local so `Path.apply` should do a lexical scoping check and error if you're not in the same module. This change is a bit of a compromise where XMR use cases should continue working while making paths that can be local actually be local instead of "absolute".

seldridge (approved):
LGTM

PR #4380, opened 2024-09-04
Enhance test and document for DPI and tweak trait name.

uenoku:
This PR adds a test for open array and document `docs/src/explanations/dpi.md` which describes DPI API usage.
This also changes DPIClockedVoidFunctionImport to DPIVoidFunctionImport for more consistent naming (this is breaking change but I believe there is no user of DPIClockedVoidFunctionImport so it should be ok). 

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Enhance test and document for DPI and tweak trait name

- [ ] Did you review?

### DIFF
--- a/docs/src/explanations/dpi.md
+++ b/docs/src/explanations/dpi.md
@@ -1,0 +1,143 @@
+---
+layout: docs
+title:  "Calling Native Functions from Chisel (DPI)"
+section: "chisel3"
+---
+
+# Calling Native Functions from Chisel (DPI)
+
+## DPI Basics
+
+DPI API allows you to integrate native code into your Chisel hardware designs. This enables you to leverage existing libraries or implement functionality that is difficult to express directly in Chisel.
+
+Here's a simple example that demonstrates printing a message from a C++ function:
+```c++
+extern "C" void hello()
+{
+    std::cout << "hello from c++\\n";
+}
+```
+
+To call this function from Chisel, we need to define a corresponding DPI object.
+
+```scala
+object Hello extends DPIVoidFunctionImport {
+  override val functionName = "hello"
+  override val clocked = true
+  final def apply() = super.call()
+}
+
+Hello() // Print
+```
+
+Explanation:
+
+* `Hello` inherits from `DPIVoidFunctionImport` because the C++ function doesn't return a value (void).
+* `functionName` specifies the C-linkage name of the C++ function.
+* `clocked = true` indicates that its function call is invoked at clock's posedge.
+* We recommend defining the apply method for a more Scala-like syntax.
+
+## Type ABI
+
+Unlike normal Chisel compilation flow, we use a specific ABI for types to interact with DPI.
+
+### Argument Types
+
+* Operand and result types must be passive.
+* A vector is lowered to an *unpacked* *open* array type, e.g., `a: Vec<4, UInt>` to `byte a []`.
+* A bundle is lowered to a packed struct.
+* Integer types are lowered into 2-state types.
+Small integer types (< 64 bit) must be compatible with C-types and arguments are passed by value. Users are required to use specific integer types for small integers shown in the table below. Large integers are lowered to bit and passed by reference.
+
+
+| Width | Verilog Type | Argument Passing Modes |
+| ----- | ------------ | ---------------------- |
+| 1     | bit          | value                  |
+| 8     | byte         | value                  |
+| 16    | shortint     | value                  |
+| 32    | int          | value                  |
+| 64    | longint      | value                  |
+| > 64  | bit [w-1:0]  | reference              |
+
+### Function Types
+The type of DPI object you need depends on the characteristics of your C++ function:
+
+* Return Type
+  * For functions that don't return a value (like hello), use `DPIVoidFunctionImport`.
+  * For functions with a return type (e.g., integer addition), use `DPINonVoidFunctionImport[T]`, where `T` is the return type.
+  * The output argument must be the last argument in DPI fuction.
+* Clocked vs. Unclocked:
+  * Clocked: The function call is evaluated at the associated clock's positive edge. By default, Chisel uses the current module's clock. For custom clocks, use `withClocked(clock) {..}`. If the function has a return value, it will be available in the next clock cycle.
+  * Unclocked: The function call is evaluated immediately.
+
+## Example: Adding Two Numbers
+Here's an example of a DPI function that calculates the sum of two numbers:
+
+```c++
+extern "C" void add(int lhs, int rhs, int* result)
+{
+    *result = lhs + rhs;
+}
+```
+
+```scala
+object Add extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "add"
+  override val ret = UInt(32.W)
+  override val clocked = false
+  override val inputNames = Some(Seq("lhs", "rhs"))
+  override val outputName = Some("result")
+  final def apply(lhs: UInt, rhs: UInt): UInt = super.call(lhs, rhs)
+}
+
+val result = Add(2.U(32.W), 3.U(32.W))
+```
+
+Explanation:
+
+* `Add` inherits from `DPINonVoidFunctionImport[UInt]` because it returns a 32-bit unsigned integer.
+* `ret` specifies the return type.
+* `clocked` indicates that this is a clocked function call.
+* `inputNames` and `outputName` provide optional names for the function's arguments and return value (these are just for Verilog readability).
+
+## Example: Sum of an array
+Chisel vectors are converted into SystemVerilog open arrays when used with DPI. Since memory layout can vary between simulators, it's recommended to use `svSize` and `svGetBitArrElemVecVal` to access array elements.
+
+```c++
+extern "C" void sum(const svOpenArrayHandle array, int* result) {
+  // Get a length of the open array.
+  int size = svSize(array, 1);
+  // Initialize the result value.
+  *result = 0;
+  for(size_t i = 0; i < size; ++i) {
+    svBitVecVal vec;
+    svGetBitArrElemVecVal(&vec, array, i);
+    *result += vec;
+  }
+}
+```
+
+```scala
+object Sum extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "sum"
+  override val ret = UInt(32.W)
+  override val clocked = false
+  override val inputNames = Some(Seq("array"))
+  override val outputName = Some("result")
+  final def apply(array: Vec[UInt]): UInt = super.call(array)
+}
+
+val io = IO(new Bundle {
+  val a = Input(Vec(3, UInt(32.W)))
+  val b = Input(Vec(6, UInt(32.W)))
+})
+
+val sum_a = Sum(io.a) // compute a[0] + a[1] + a[2]
+val sum_b = Sum(io.b) // compute b[0] + ... + b[5]
+```
+
+# FAQ
+
+* Can we export functions? -- No, not currently. Consider using a black box for such functionality.
+* Can we call a DPI function in initial block? -- No, not currently. Consider using a black box for initialization.
+* Can we call two clocked DPI calls and pass the result to another within the same clock? -- No, not currently. Please merge the DPI functions into a single function.
--- a/docs/src/explanations/dpi.md
+++ b/docs/src/explanations/dpi.md
@@ -8,7 +8,7 @@ section: "chisel3"
 
 ## DPI Basics
 
-DPI API allows you to integrate native code into your Chisel hardware designs. This enables you to leverage existing libraries or implement functionality that is difficult to express directly in Chisel.
+Chisel's DPI API allows you to integrate native code into your Chisel hardware designs. This enables you to leverage existing libraries or implement functionality that is difficult to express directly in Chisel.
 
 Here's a simple example that demonstrates printing a message from a C++ function:
 ```c++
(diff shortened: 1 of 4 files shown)


docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
+---
```
mwachs5, line 1: can you please add this to the ToC on the [explanations/index.md?](https://raw.githubusercontent.com/chipsalliance/chisel/main/docs/src/explanations.md)?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+To call this function from Chisel, we need to define a corresponding DPI object.
+
+```scala
+object Hello extends DPIVoidFunctionImport {
+  override val functionName = "hello"
+  override val clocked = true
+  final def apply() = super.call()
+}
+
+Hello() // Print
```
mwachs5, line 30: should this `Hello()` call be inside a Module?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+```c++
+extern "C" void hello()
+{
+    std::cout << "hello from c++\\n";
+}
+```
+
+To call this function from Chisel, we need to define a corresponding DPI object.
+
+```scala
```
mwachs5, line 23: can you use `mdoc` so that we can check the code compiles at least?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Type ABI
+
+Unlike normal Chisel compilation flow, we use a specific ABI for types to interact with DPI.
+
+### Argument Types
+
+* Operand and result types must be passive.
+* A vector is lowered to an *unpacked* *open* array type, e.g., `a: Vec<4, UInt>` to `byte a []`.
+* A bundle is lowered to a packed struct.
```
mwachs5, line 48:
```suggestion
* A `Vec` is lowered to an *unpacked* *open* array type, e.g., `a: Vec<4, UInt>` to `byte a []`.
* A `Bundle` is lowered to a packed struct.
```

for vec i think you mean a chisel3.Vec not a scala `Vector`, right?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+* Integer types are lowered into 2-state types.
```
mwachs5, line 49:
```suggestion
* `Int`, `SInt`, `Clock`, `Reset` types are lowered into 2-state types.
```

? Are clock and reset lowered?

mwachs5: I am not actually sure if you are talkign about Chisel Data subclasses or not

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+object Add extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "add"
+  override val ret = UInt(32.W)
+  override val clocked = false
+  override val inputNames = Some(Seq("lhs", "rhs"))
+  override val outputName = Some("result")
+  final def apply(lhs: UInt, rhs: UInt): UInt = super.call(lhs, rhs)
+}
+
+val result = Add(2.U(32.W), 3.U(32.W))
```
mwachs5, line 93: again would be helpful to see this with mdoc, I assume this Add function belongs inside a Module or a RawModule given that we say `clocked` is False

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Explanation:
+
+* `Add` inherits from `DPINonVoidFunctionImport[UInt]` because it returns a 32-bit unsigned integer.
+* `ret` specifies the return type.
+* `clocked` indicates that this is a clocked function call.
```
mwachs5, line 100: but this is set to `false` so isn't NOT a clocked function call?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+    *result += vec;
+  }
+}
+```
+
+```scala
+object Sum extends DPINonVoidFunctionImport[UInt] {
+  override val functionName = "sum"
+  override val ret = UInt(32.W)
+  override val clocked = false
```
mwachs5, line 124: should clocked be true?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+  *result = 0;
+  for(size_t i = 0; i < size; ++i) {
+    svBitVecVal vec;
+    svGetBitArrElemVecVal(&vec, array, i);
+    *result += vec;
+  }
+}
+```
+
+```scala
```
mwachs5, line 120: again please use mdoc to check for compile if not chisel elaboration

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+  val a = Input(Vec(3, UInt(32.W)))
+  val b = Input(Vec(6, UInt(32.W)))
+})
+
+val sum_a = Sum(io.a) // compute a[0] + a[1] + a[2]
+val sum_b = Sum(io.b) // compute b[0] + ... + b[5]
+```
+
+# FAQ
+
```
mwachs5, line 140: can you show an example where having clocked = false makes sense? Is that done in a `when` context...?

mmaloney-sf:
I'm not 100% sure of Chisel's documentation conventions, but quickly glancing at other `.md` files, I think we should stick to a "one sentence per line" convention. 

This makes diffs easier to read.

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+Hello() // Print
+```
+
+Explanation:
+
+* `Hello` inherits from `DPIVoidFunctionImport` because the C++ function doesn't return a value (void).
+* `functionName` specifies the C-linkage name of the C++ function.
+* `clocked = true` indicates that its function call is invoked at clock's posedge.
```
mmaloney-sf, line 37: "the clock's posedge"

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+* We recommend defining the apply method for a more Scala-like syntax.
```
mmaloney-sf, line 38: Rather than saying "we recommend", just point out that you did this.

"We define an `apply` method for a function-like syntax. This allows us to call the DPI function with `Hello()`."

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Type ABI
+
+Unlike normal Chisel compilation flow, we use a specific ABI for types to interact with DPI.
```
mmaloney-sf, line 42: I'm not sure what this means. Could you elaborate?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Argument Types
+
+* Operand and result types must be passive.
```
mmaloney-sf, line 46: If we have another document that explains what "passive" means, we should link to it here. This is a rather technical term, and will otherwise cause confusion.

docs/src/explanations/dpi.md
mmaloney-sf: It may also be useful to explicitly point out that what we're doing in this section is showing how Chisel types map to C types.

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+| ----- | ------------ | ---------------------- |
+| 1     | bit          | value                  |
+| 8     | byte         | value                  |
+| 16    | shortint     | value                  |
+| 32    | int          | value                  |
+| 64    | longint      | value                  |
+| > 64  | bit [w-1:0]  | reference              |
+
+### Function Types
+The type of DPI object you need depends on the characteristics of your C++ function:
```
mmaloney-sf, line 63: The type of DPI object I need for what?

mmaloney-sf: Reading ahead, I think this section would be better typed out in prose, rather than summarized as a bulleted list.

"There are several base classes which are used to define DPI functions in Chisel. Which one you use depends on two factors: the return type of the function in C++ and how the DPI is called in relation to the clock."

Then outline the information in the bullet points in two paragraphs after that.

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+  * The output argument must be the last argument in DPI fuction.
+* Clocked vs. Unclocked:
+  * Clocked: The function call is evaluated at the associated clock's positive edge. By default, Chisel uses the current module's clock. For custom clocks, use `withClocked(clock) {..}`. If the function has a return value, it will be available in the next clock cycle.
+  * Unclocked: The function call is evaluated immediately.
+
+## Example: Adding Two Numbers
+Here's an example of a DPI function that calculates the sum of two numbers:
+
+```c++
+extern "C" void add(int lhs, int rhs, int* result)
```
mmaloney-sf, line 77: It surprised me to see here that the `int* result` is an out-parameter. Is that always the case? (If so, maybe it needs to be called out earlier. Or maybe I just missed it?)

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+* `Add` inherits from `DPINonVoidFunctionImport[UInt]` because it returns a 32-bit unsigned integer.
+* `ret` specifies the return type.
+* `clocked` indicates that this is a clocked function call.
+* `inputNames` and `outputName` provide optional names for the function's arguments and return value (these are just for Verilog readability).
+
+## Example: Sum of an array
+Chisel vectors are converted into SystemVerilog open arrays when used with DPI. Since memory layout can vary between simulators, it's recommended to use `svSize` and `svGetBitArrElemVecVal` to access array elements.
+
+```c++
+extern "C" void sum(const svOpenArrayHandle array, int* result) {
```
mmaloney-sf, line 107: What is `svOpenArrayHandle`?

I guess, more generally, what libraries does Chisel depend on in the C++ DPI? We can't simply assume that a general Chisel user will have whatever library exposes `DPI.h` or whatever it is. 

It would be good to have a section at the very start that gives the C++ prerequisites.

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+  val b = Input(Vec(6, UInt(32.W)))
+})
+
+val sum_a = Sum(io.a) // compute a[0] + a[1] + a[2]
+val sum_b = Sum(io.b) // compute b[0] + ... + b[5]
+```
+
+# FAQ
+
+* Can we export functions? -- No, not currently. Consider using a black box for such functionality.
```
mmaloney-sf, line 141: What do you mean by "export" here?

docs/src/explanations/dpi.md
```diff
@@ -0,0 +1,143 @@
...
+* Can we call a DPI function in initial block? -- No, not currently. Consider using a black box for initialization.
+* Can we call two clocked DPI calls and pass the result to another within the same clock? -- No, not currently. Please merge the DPI functions into a single function.
```
mmaloney-sf, line 143: I'm not sure what this last point means.

It feels like what we actually want is a light explanation of the underlying "framework" for the DPI calls, and that the above comment is actually a restriction imposed on this framework. 

For example,  is this true?: "All DPI calls on the same clock are dispatched at the same time. Because of this, it is not possible to use the result of one DPI call as an argument to another."

src/test/scala/chiselTests/DPISpec.scala
```diff
@@ -17,6 +17,7 @@ private object EmitDPIImplementation {
     val dpiImpl = s"""
                      |#include <stdint.h>
                      |#include <iostream>
+                     |#include <svdpi.h>
```
mmaloney-sf, line 20: To reiterate an earlier point: Let's add a prerequisites section near the top. Let's make sure to name what library(ies?) are needed for Chisel DPI, and let's be sure to include the `svdpi.h` header in our examples, since otherwise, the code doesn't work.

mmaloney-sf: Let's also link to the documentation so that Chisel users can quickly cross-reference the DPI primitives, such as `svBitVecVal` or `svOpenArrayHandle`.

mmaloney-sf (changes requested):

docs/src/explanations/dpi.md
uenoku: For OpenArray example both clocked and unclocked work. So I didn't explicitly describe the example. It depends on whether user wants to evaluate the value at every input value changes or clock's posedge. 
This is an example for Add https://github.com/chipsalliance/chisel/blob/ead5fc99a836d1d547463021dae31e65b77b48c6/src/test/scala/chiselTests/DPISpec.scala#L158-L161

docs/src/explanations/dpi.md
uenoku: Yes, thank you for pointing out. Totally missed there is `mdoc` support in chisel repo as well.

docs/src/explanations/dpi.md
uenoku: Yes. The snippet was meant to be pseudo code so I'll update with mdoc.

docs/src/explanations/dpi.md
uenoku: > All DPI calls on the same clock are dispatched at the same time. Because of this, it is not possible to use the result of one DPI call as an argument to another

Yes, that's correct. This is fundamental restrictions for side-effecting operations in Chisel but DPI is a first example that has both side-effect and results.

docs/src/explanations/dpi.md
uenoku: Unclocked calls be useful when user wants to replace pure (but expensive) combinatorial logics with dpi call (actually `Add` is a good example of such use case). 
An unclocked call under `when(cond)` is lowered into `always_comb` + `if(cond)` and DPI is conditionally invoked as well. I updated `Add` example to include both of them. 

```scala
class AddTest extends Module {
  val io = IO(new Bundle {
    val a = Input(UInt(32.W))
    val b = Input(UInt(32.W))
    val c = Output(UInt(32.W))
    val d = Output(UInt(32.W))
    val en = Input(Bool())
  })

  // Call DPI only when `en` is true.
  when (io.en) {
    io.c := AddClocked(io.a, io.b)
    io.d := AddUnclocked(io.a, io.b)
  } .otherwise {
    io.c := 0.U(32.W)
    io.d := 0.U(32.W)
  }
}
```
```verilog
module AddTest(
  input         clock,
                reset,
  input  [31:0] io_a,
                io_b,
  output [31:0] io_c,
                io_d,
  input         io_en
);

  logic [31:0] _add_0;
  reg   [31:0] _GEN;
  always @(posedge clock) begin
    if (io_en) begin
      add(io_a, io_b, _add_0);
      _GEN <= _add_0;
    end
  end // always @(posedge)
  reg   [31:0] _GEN_0;
  always_comb begin
    if (io_en) begin
      add(io_a, io_b, _GEN_0);
    end
    else
      _GEN_0 = 32'bx;
  end // always_comb
  assign io_c = io_en ? _GEN : 32'h0;
  assign io_d = io_en ? _GEN_0 : 32'h0;
```

uenoku (commented):
Thank you for comments! I added code examples with mdoc. I'll mark it draft until all comments were addressed.

Issue #4353, opened 2024-08-15
Mill Blackbox Resource Loading Doesn't Work


seldridge:
When running tests with mill, tests involving reading from resources do not work. I think this is because mill will, by default, use a different project structure (https://stackoverflow.com/questions/73370982/cannot-load-resource-files-using-scala-mill).

This can be reproduced with the following command:

```
./mill 'chiselut[2.13.14].testOnly' chiselTests.simulator.SimulatorSpec -- -z "runs a design that includes an external module"
```

This produces:

```
- runs a design that includes an external module *** FAILED ***
  os.ResourceNotFoundException: jdk.internal.loader.ClassLoaders$AppClassLoader@200a570f/chisel3/simulator/Baz.sv
  at os.ResourcePath.getInputStream(ResourcePath.scala:21)
  at os.read$bytes$.apply(ReadWriteOps.scala:272)
  at os.read$.apply(ReadWriteOps.scala:229)
  at os.read$.apply(ReadWriteOps.scala:227)
  at chisel3.util.HasExtModuleResource.addResource(ExtModuleUtils.scala:26)
  at chisel3.util.HasExtModuleResource.addResource$(ExtModuleUtils.scala:21)
  at chiselTests.simulator.SimulatorSpec$Baz$1.addResource(SimulatorSpec.scala:114)
  at chiselTests.simulator.SimulatorSpec$Baz$1.<init>(SimulatorSpec.scala:116)
  at chiselTests.simulator.SimulatorSpec$Foo$1.$anonfun$new$19(SimulatorSpec.scala:127)
  at chisel3.Module$.evaluate(Module.scala:94)
```

This is the same resource path that exists when running with sbt, but mill can't load it.

PR #4384, opened 2024-09-05
[mill] Get unit testing working

jackkoenig:
Fixes https://github.com/chipsalliance/chisel/issues/4353.

Another step toward replacing SBT.

I went ahead and used the way of defining tests in Mill rather than doing it more manually. This ensures we get the SBT default `src/test/scala` and `src/test/resource` directories for free. It's also just more idiomatic as this is what the Mill examples do.

These tests can be run in parallel with `./mill -j0 __.test`. I also figured out how to create "aggregate commands" that run in parallel (if you tell mill to use parallelism with `-j`):

```scala
object utils {
  // Copied from https://github.com/com-lihaoyi/mill/blob/4caadfb62581b4c1772c14c6f8dc8963011eaec7/main/define/src/mill/define/Task.scala#L55
  // This makes it possible to run multiple tasks in parallel (when passing `-j <n>` to mill)
  // Mill's source has MIT license
  class Sequence[+T](inputs0: Seq[Task[T]]) extends Task[Seq[T]] {
    val inputs: Seq[Task[_]] = inputs0
    def evaluate(ctx: mill.api.Ctx): mill.api.Result[Seq[T]] = {
      for (i <- 0 until ctx.args.length)
        yield ctx.args(i).asInstanceOf[T]
    }
  }
  // end copied from Mill

  // Sugar API over Sequence
  def parallelize[T](tasks: Seq[Task[T]]): Task[Seq[T]] = new Sequence(tasks)
}

object test extends TaskModule {

  override def defaultCommandName() = "run"

  val tasks = Seq(
    svsim(v.scalaVersion).test.test(),
    firrtl(v.scalaVersion).test.test(),
    chisel(v.scalaVersion).test.test(),
    integrationTests(v.scalaVersion).test.test()
  ) 
  def run() = T.command {
    utils.parallelize(tasks)()
  }
}
```

I don't think we need this yet, but if we ever want command "aliases", this approach might be useful.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -122,28 +122,19 @@ trait Firrtl extends common.FirrtlModule with ChiselPublishModule with CrossSbtM
   def commonTextIvy = v.commonText
 
   def scoptIvy = v.scopt
+
+  object test extends SbtModuleTests with TestModule.ScalaTest {
+    def ivyDeps = Agg(v.scalatest, v.scalacheck)
+  }
 }
 
 object svsim extends Cross[Svsim](v.scalaCrossVersions)
 
 trait Svsim extends common.SvsimModule with ChiselPublishModule with CrossSbtModule with ScalafmtModule {
   def millSourcePath = super.millSourcePath / os.up / "svsim"
-}
 
-object firrtlut extends Cross[FirrtlUnitTest](v.scalaCrossVersions)
-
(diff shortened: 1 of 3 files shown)


jackkoenig:
From a mill dev, you can just do `T.sequence(tasks)()`, we don't need to copy the `Sequence` class into our `build.sc`.

PR #4386, opened 2024-09-06
[svsim] Handle Chisel assertions more elegantly

jackkoenig:
This is a bugfix but I'm not sure if it should be backported [yet] because adding new arguments, even with default arguments, to public methods breaks binary compatibility. I don't feel like dealing with that at the moment. If we need to backport it later, we can. We could backport just the bugfix part of it I guess.

The returned exception is now a lot more elegant than `svsim.Simulation.UnexpectedEndOfMessages`. This also fixes an issue where the Verilator crashes could create core dumps depending on the user's ulimit settings.

I have not yet tested this with VCS, so there may be follow up work (possibly in a follow up PR).

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

* svsim now overrides Verilators $fatal handling to more elegantly report the error.
* svsim can now accept a java.io.OutputStream to consume the stderr from the simulation process. This is especially useful for testing.

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/Simulation.scala
+++ b/svsim/src/main/scala/Simulation.scala
@@ -4,6 +4,7 @@ package svsim
 
 import scala.collection.mutable.Queue
 import scala.util.Try
+import scala.concurrent.{ExecutionContext, Future}
 import java.io.{BufferedReader, BufferedWriter, File, InputStreamReader, OutputStreamWriter}
 
 final class Simulation private[svsim] (
@@ -18,7 +19,8 @@ final class Simulation private[svsim] (
     conservativeCommandResolution: Boolean = false,
     verbose:                       Boolean = false,
     traceEnabled:                  Boolean = false,
-    executionScriptLimit:          Option[Int] = None
+    executionScriptLimit:          Option[Int] = None,
+    stderrStream:                  java.io.OutputStream = Console.err
   )(body:                          Simulation.Controller => T
   ): T = {
     val cwd = settings.customWorkingDirectory match {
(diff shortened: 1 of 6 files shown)


jackkoenig:
What's weird is that these failing tests are currently printing failures and maybe even crashing on **main**, but for some reason SVSim returns success. I think my changes are just exposing some other underlying issue. I had been planning to revisit how we deal with `$stop` in a follow on PR, but it might just have to be done in this PR as well.

svsim/src/main/scala/Simulation.scala
```diff
@@ -41,6 +42,22 @@ final class Simulation private[svsim] (
...
+      val err = process.getErrorStream()
+      val buffer = new Array[Byte](BufSize)
+      var read = 0
+      while ({
+        read = err.read(buffer, 0, BufSize)
+        read >= 0
+      }) {
+        stderrStream.write(buffer, 0, read)
+      }
+    }
```
dobios, line 60: kind of a drive-by comment, but is there a particular reason why you're handling this as a Java Stream rather than using Scala's nice streaming data-types and stream programming constructs? Is there non Scala equivalent to `Console.err` that you can use?

jackkoenig: I think Scala's streaming constructs are generally for streaming data, whereas for process-level I/O streams I think it's still pretty typical to just use `java.io` streams (as evidenced by the fact that scala.Console.err is a `java.io.OutputStream`). I guess I'm not aware of any advantage to wrapping this in a Scala type but if anyone has a suggestion I'm all ears 🙂.

I could use `System.err`, I'm not really sure when one should use Java's `System` vs. Scala's `Console` for these sorts of things, but we often use things like `Console.withOut` to capture stdout, so I figured I'd go with the Scala API.

dobios (approved):
For what it's worth, this looks good to me given that I couldn't find a better way to handle those streams either. Not sure why CI is broken tho

jackkoenig:
Thanks for the review @dobios! I think I sadly have to bite the bullet and also implement `$stop` for this to pass CI... will revisit eventually 😭

jackkoenig:
This has been made obsolete by @seldridge's recent work.

PR #4395, opened 2024-09-12
Select.collectDeep and related: search into layers and whens

dtzSiFive:
Use `collect` in `instances` and other methods, such that they and API using them like `collectDeep` work with regions.

Also tweak `collect` to visit the layer command itself, on principle, should this be reachable or useful in the future.

Add two tests confirming collectDeep now searches within when and layer regions.

### Contributor Checklist

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

#### Release Notes

More Select methods support searching into whens and layers.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/aop/Select.scala
+++ b/src/main/scala/chisel3/aop/Select.scala
@@ -43,7 +43,13 @@ object Select {
           else
             Seq.empty
         head ++ collect(ifRegion)(f) ++ collect(elseRegion)(f)
-      case LayerBlock(_, _, region)  => collect(region)(f)
+      case cmd @ LayerBlock(_, _, region) =>
+        val head =
+          if (f.isDefinedAt(cmd))
+            Seq(f(cmd))
+          else
+            Seq.empty
+        head ++ collect(region)(f)
       case cmd if f.isDefinedAt(cmd) => Some(f(cmd))
       case _                         => None
     }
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/aop/Select.scala
```diff
@@ -43,7 +43,13 @@ object Select {
...
             Seq.empty
         head ++ collect(ifRegion)(f) ++ collect(elseRegion)(f)
-      case LayerBlock(_, _, region)  => collect(region)(f)
+      case cmd @ LayerBlock(_, _, region) =>
+        val head =
+          if (f.isDefinedAt(cmd))
+            Seq(f(cmd))
+          else
+            Seq.empty
+        head ++ collect(region)(f)
```
jackkoenig, line 52:
```suggestion
        val head = f.lift(cmd).toSeq
        head ++ collect(region)(f)
```
You can `lift` a `PartialFunction` to return `Option[R]`, then just `.toSeq` to get a `Seq`. You could alternatively leave it as an `Option` and prepend with `head ++: collect(region)(f)` but either way is fine.

dtzSiFive: Awesome, thanks! I reached for that but wasn't entirely sure and so just copied this from above.
I like your suggestion, thanks!

jackkoenig (approved):
LGTM!

Issue #4108, opened 2024-05-29
BoringUtils + when/layer/region produces invalid FIRRTL


dtzSiFive:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**

Consider these tests: https://github.com/dtzSiFive/chisel3/commit/d8a3f1213c32a9e69fd6e15eaf4c9030ce2235fb .

Steps to reproduce are to checkout that branch / apply that commit and run the tests.

**What is the current behavior?**

The tests fail due to generating invalid FIRRTL because BoringUtils appends connect statements to the end where not all declarations are visible.

Example FIRRTL for the layer test example:
```
FIRRTL version 4.0.0
circuit Top :
  layer TestLayer, bind : @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 167:36]
  module Bar : @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 168:11]
    output out : UInt<1> @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 169:19]
    input out_bore : UInt<1> @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 170:36]

    connect out, out_bore @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 170:11]

  module Foo : @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 172:11]
    input out_bore : UInt<1> @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 170:36]

    layerblock TestLayer : @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 173:30]
      inst bar of Bar @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 174:25]

    connect bar.out_bore, out_bore @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 170:36]

  public module Top : @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 177:11]

    wire parentWire : UInt<1> @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 178:28]
    inst foo of Foo @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 179:23]
    connect foo.out_bore, parentWire @[src/test/scala/chiselTests/BoringUtilsTapSpec.scala 170:36]
```

`connect bar.out_bore, out_bore` is invalid, `bar` is only visible within the layer.

**What is the expected behavior?**

[...]

PR #4399, opened 2024-09-12
Flush staged commands before/after enter/exit regions.

dtzSiFive:
Fix BoringUtils (see tests) placement of connection commands so that they are added to the current command buffer.

Flush staged commands before changing regions to avoid needing multiple "staging" buffer lists.
Alternatively the current list could be saved and restored.

Tests from #4108 are included and finished, with this change two of them now pass (the ones encountered in practice, FWIW). The third is included but marked "ignore" as future work.

This is a bit of a change regarding "staged" commands, as now before the module is closed they are no longer "secret" since they are added to the normal command buffers.
They are still staged (deferred) which appears to be useful to ensure they appear after what's currently being constructed (specifically, not before).

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Fix BoringUtils with regions in some cases.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -97,9 +97,13 @@ abstract class RawModule extends BaseModule {
   }
 
   private[chisel3] def withRegion[A](newRegion: VectorBuilder[Command])(thunk: => A): A = {
+    _currentRegion ++= stagedSecretCommands
+    stagedSecretCommands.clear()
     var oldRegion = _currentRegion
     changeRegion(newRegion)
     val result = thunk
+    _currentRegion ++= stagedSecretCommands
+    stagedSecretCommands.clear()
     changeRegion(oldRegion)
     result
   }
@@ -220,6 +224,7 @@ abstract class RawModule extends BaseModule {
 
     // Secret connections can be staged if user bored into children modules
     component.secretCommands ++= stagedSecretCommands
+    stagedSecretCommands.clear()
     _component = Some(component)
     _component
   }
--- a/src/test/scala/chiselTests/BoringUtilsTapSpec.scala
+++ b/src/test/scala/chiselTests/BoringUtilsTapSpec.scala
@@ -89,8 +89,8 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
       "      define w_bore = bar.w_bore"
     )()
 
-    // Check is valid FIRRTL and builds to SV.
-    circt.stage.ChiselStage.emitSystemVerilog(new Top)
+    // Check is valid FIRRTL.
+    circt.stage.ChiselStage.emitFIRRTLDialect(new Top)
   }
 
   it should "work upwards from child to parent" in {
@@ -169,8 +169,8 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
       "      connect bar.out_bore, out_bore"
     )()
 
-    // Check is valid FIRRTL and builds to SV.
-    circt.stage.ChiselStage.emitSystemVerilog(new Top)
+    // Check is valid FIRRTL.
+    circt.stage.ChiselStage.emitFIRRTLDialect(new Top)
   }
 
   it should "work upwards from grandchild to grandparent into layer" in {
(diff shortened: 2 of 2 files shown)


core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -220,6 +224,7 @@ abstract class RawModule extends BaseModule {
 
     // Secret connections can be staged if user bored into children modules
     component.secretCommands ++= stagedSecretCommands
+    stagedSecretCommands.clear()
```
seldridge, line 227: Nice that this is now cleaning up after itself. 👍

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -97,9 +97,13 @@ abstract class RawModule extends BaseModule {
   }
 
   private[chisel3] def withRegion[A](newRegion: VectorBuilder[Command])(thunk: => A): A = {
+    _currentRegion ++= stagedSecretCommands
+    stagedSecretCommands.clear()
```
seldridge, line 101: An alternative to flushing here would be save/restore the old staged commands. I can't think of a good reason as to why this would be better.

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -109,6 +145,63 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
+      val foo = Module(new Foo(parentWire))
+    }
+    val chirrtl = circt.stage.ChiselStage.emitCHIRRTL(new Top)
+
+    // The connect should be at the end of the when block.
+    matchesAndOmits(chirrtl)(
+      "    when UInt<1>(0h1) :",
+      "      inst bar of Bar",
+      "      connect bar.out_bore, out_bore"
+    )()
```
seldridge, line 170: We really need something better than `matchesAndOmits` (FileCheck) as this test will, I think, pass without this change. Given that, the test might be better written as simply running `emitFIRRTLDialect` which will cause CIRCT to check that this isn't busted. Could you check if that would cause this to fail before the commit and pass after?

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -109,6 +145,63 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
+    // The connect should be at the end of the layerblock.
+    matchesAndOmits(chirrtl)(
+      "    layerblock TestLayer :",
+      "      inst bar of Bar",
+      "      connect bar.out_bore, out_bore"
+    )()
+
+    // Check is valid FIRRTL and builds to SV.
+    circt.stage.ChiselStage.emitSystemVerilog(new Top)
+  }
```
seldridge, line 203: There's no additional coverage that this test is providing. However, I think it's good to have it in here. 👍

seldridge (approved):

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
dtzSiFive: The whitespace is what ensures this fails without the change, just re-checked that this is the case.
Whoa, TIL about `emitFIRRTLDialect`!

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
dtzSiFive: It's checking that the FIRRTL isn't invalid, or that was the idea.  It previously failed but as you say w.r.t. checking the validity kinda doesn't add more directly other than it's not broken otherwise I guess.

I'm very amenable to just running `emitFIRRTLDialect`  (firtool through `ir-fir`, maybe/apparently?) instead of both of these!

core/src/main/scala/chisel3/RawModule.scala
dtzSiFive: Agreed, noted this in the original commit message actually haha because I thought this as well but (also) wasn't sure which was better to be honest.

jackkoenig (approved):
LGTM!

dtzSiFive:
Thank you very much for review!

PR #4403, opened 2024-09-16
Support cross-compiling FIRRTL source files with Scala3

adkian-sifive:
Part of the larger work of preparing for Scala3 support

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

#### Release Notes

Refactor FIRRTL sources to support cross-compilation with Scala3

- [ ] Did you review?

### DIFF
--- a/firrtl/src/main/scala-2/Macros.scala
+++ b/firrtl/src/main/scala-2/Macros.scala
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 package firrtl.macros
 
 import scala.reflect
--- a/firrtl/src/main/scala-2/Macros.scala
+++ b/firrtl/src/main/scala-2/Macros.scala
@@ -1,0 +1,9 @@
+package firrtl.macros
+
+import scala.reflect
+
+object Macros {
+  def isSingletonImpl(obj: AnyRef): Boolean = {
+    reflect.runtime.currentMirror.reflect(obj).symbol.isModuleClass
+  }
+}
--- a/firrtl/src/main/scala/firrtl/Emitter.scala
+++ b/firrtl/src/main/scala/firrtl/Emitter.scala
@@ -6,6 +6,7 @@ import firrtl.annotations.NoTargetAnnotation
 import firrtl.options.Viewer.view
 import firrtl.options.CustomFileEmission
 import firrtl.stage.FirrtlOptions
+import firrtl.stage.FirrtlOptionsView
 
 // ***** Annotations for results of emission *****
 sealed abstract class EmittedComponent {
--- a/firrtl/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/firrtl/src/main/scala/firrtl/graph/DiGraph.scala
@@ -279,11 +279,11 @@ class DiGraph[T](private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]) {
         if (frame.childCall.isEmpty) {
           if (lowlinks(v) == indices(v)) {
             val scc = new mutable.ArrayBuffer[T]
-            do {
+            while (scc.last != v) {
               val w = stack.pop()
               onstack -= w
               scc += w
-            } while (scc.last != v);
+            }
             sccs.append(scc.toSeq)
           }
           callStack.pop()
--- a/firrtl/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/firrtl/src/main/scala/firrtl/graph/DiGraph.scala
@@ -279,14 +279,12 @@ class DiGraph[T](private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]) {
         if (frame.childCall.isEmpty) {
           if (lowlinks(v) == indices(v)) {
             val scc = new mutable.ArrayBuffer[T]
-            val w = stack.pop()
-            onstack -= w
-            scc += w
-            while (scc.last != v) {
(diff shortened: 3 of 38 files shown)


firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -372,28 +371,32 @@ case class PropAssign(info: Info, loc: Expression, expr: Expression) extends Sta
...
   def copy(info: Info = info, ret: Int = ret, clk: Expression = clk, en: Expression = en): Stop = {
-    Stop(info, ret, clk, en, name)
+    new Stop(info, ret, clk, en, name)
   }
 }
 object Stop {
   def unapply(s: Stop): Some[(Info, Int, Expression, Expression)] = {
     Some((s.info, s.ret, s.clk, s.en))
   }
 }
```
jackkoenig, line 392: The custom `copy` and `unapply` are there to preserve binary compatibility (which is why we used `@data` in the first place. We should just break binary compatibility and go ahead and delete the manual copy and unapply as well. Same for `Print` below.

jackkoenig:
I'm happy with this as is, but I think we should also add `-Xsource:3` to the `scalacOptions` for `firrtl` so that we can enforce we don't regress on some of these changes. Do you want to do that here or in the next PR?

adkian-sifive:
I like the idea of adding -Xsource:3 on a subproject basis

firrtl/src/main/scala/firrtl/Emitter.scala
```diff
@@ -30,7 +31,7 @@ trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation with C
 }
 
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
-  override def getBytes = value.value.getBytes
+  override def getBytes: scala.collection.mutable.ArraySeq.ofByte = value.value.getBytes
```
jackkoenig, line 34: This matters a lot less here since we've deprecated the entire `import firrtl`, but this is going to be a common theme as we add `-Xsource:3` so I want to describe what we should do in cases like these.

We don't want the locally inferred type. In this case, we want the type from `CustomFileEmission.getBytes` which is `Iterable[Byte]`. I'm a little surprised you need a type annotation here though since my understanding is that `-Xsource-features:infer-override` would do what we want (making the type `Iterable[Byte]`). Maybe you made this change before adding that option?

Regardless, we do not want this type, same for similar things below. We also need to report types that have changed in release notes. I don't think we need to do it for this compilation unit (since `import firrtl._`) is deprecated anyway, but we will need to do so for `core` and `chisel` since it could affect users in some rare cases. I think we can do this by enabling `-Xsource:3` _without_ `-Xsource-features:infer-override`, collecting the errors for release notes, then adding `-Xsource-features:infer-override`. We will need to make sure we don't hit the max number of errors when doing that.

adkian-sifive: hm not sure i understand - all type annotation additions in commit 2dfb9d0 are quickfixable compiler rewrites

adkian-sifive: that is i did not add these manually, they were added by enabling infer-overrides

jackkoenig: My point is that the quickfix rewrite is not what we want. We want the type that Scala 3 will infer which is what the error is about. That type in this case is `Iterable[Byte]`. This is a case where we want to change Chisel's public API a bit in prep for the Scala 3 upgrade.

firrtl/src/main/scala-2/Macros.scala
```diff
@@ -0,0 +1,9 @@
+package firrtl.macros
```
jackkoenig, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0

package firrtl.macros
```

firrtl/src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -279,11 +279,14 @@ class DiGraph[T](private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]) {
...
-            do {
+            val w = stack.pop()
+            onstack -= w
+            scc += w
+            while (scc.last != v) {
               val w = stack.pop()
               onstack -= w
               scc += w
-            } while (scc.last != v);
+            }
```
jackkoenig, line 289: Isn't the suggested transformation to just make the look the condition?
```scala
while({
  val w = stack.pop()
  onstack -= w
  scc += w
  scc.last != v
}) ()
// or you can end it more explicitly with
}) { /* empty body, logic is in condition */ }
```

jackkoenig (approved):
Generally LGTM but see suggestion

firrtl/src/main/scala/firrtl/graph/DiGraph.scala
sequencer: Is DiGraph still using? How about directly remove it?

seldridge: It's used by the `DependencyManager` for computing phase orderings. It can't go just yet. However, it could be made private if that is useful.

sequencer: oh I forgot this. They should go away eventually.

PR #4409, opened 2024-09-21
feat: implement SerializableModuleElaborator

unlsycn:
This PR implements the SerializableModuleElaborator trait, which is used to create elaborators for SerializableModule. In t1, we compile all the Scala code into a jar file and specify different main classes to elaborate different modules. 
The config API is used to guarantee the type safety of the parameters, and the design API actually produces the circuit, which is used this way:
```scala
  @main
  def config(
    @arg(name = "parameter") parameter:  GCDParameterMain,
    @arg(name = "target-dir") targetDir: os.Path = os.pwd
  ) =
    os.write.over(targetDir / s"${topName}.json", configImpl(parameter.convert))

  @main
  def design(
    @arg(name = "parameter") parameter:  os.Path,
    @arg(name = "target-dir") targetDir: os.Path = os.pwd
  ) = {
    val (firrtl, annos) = designImpl[GCD, GCDParameter](os.read.stream(parameter))
    os.write.over(targetDir / s"${topName}.fir", firrtl)
    os.write.over(targetDir / s"${topName}.anno.json", annos)
  }
```

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Add SerializableModuleElaborator

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
+++ b/src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
@@ -1,0 +1,80 @@
+package chisel3.experimental.util
+
+import chisel3.RawModule
+import chisel3.experimental.{SerializableModule, SerializableModuleGenerator, SerializableModuleParameter}
+
+import scala.reflect.runtime.universe
+import scala.reflect.runtime.universe.{runtimeMirror, typeOf}
+
+/** The target design emitted by [[SerializableModuleElaborator]], which contains
+  * @param fir the firrtl circuit
+  * @param annos annotations which is serializable
+  */
+case class SerializableModuleDesign(fir: firrtl.ir.Circuit, annos: Seq[firrtl.annotations.Annotation]) {
+  def firFile = fir.serialize
+  def annosFile = firrtl.annotations.JsonProtocol.serializeRecover(annos)
+}
+
+/** Mixin this trait to produce elaborators for [[SerializableModule]]
+  */
+trait SerializableModuleElaborator {
+
+  /**
+    * Implementation of a config API to serialize the [[SerializableModuleParameter]]
+    * @example
+    * {{{
+    *  def config(parameter: MySerializableModuleParameter) =
+    *    os.write.over(os.pwd / "config.json", configImpl(parameter))
+    * }}}
+    */
+  def configImpl[P <: SerializableModuleParameter: universe.TypeTag](
+    parameter: P
+  )(
+    implicit rwP: upickle.default.Writer[P]
+  ) = upickle.default.write(parameter)
+
+  /**
+    * Implementation of a design API to elaborate [[SerializableModule]]
+    *
+    * @return [[SerializableModuleDesign]]
+    * @example
+    * {{{
+    *  def design(parameter: os.Path) = {
+    *    val design = designImpl[MySerializableModule, MySerializableModuleParameter](os.read(parameter))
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
```diff
@@ -0,0 +1,80 @@
+package chisel3.experimental.util
```
sequencer, line 1: Please add Apache license header.

src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
```diff
@@ -0,0 +1,80 @@
...
+import scala.reflect.runtime.universe.{runtimeMirror, typeOf}
+
+/** The target design emitted by [[SerializableModuleElaborator]], which contains
+  * @param fir the firrtl circuit
+  * @param annos annotations which is serializable
+  */
+case class SerializableModuleDesign(fir: firrtl.ir.Circuit, annos: Seq[firrtl.annotations.Annotation]) {
+  def firFile = fir.serialize
+  def annosFile = firrtl.annotations.JsonProtocol.serializeRecover(annos)
+}
```
sequencer, line 16: How about directly introducing the tuple for result? Adding new API here for only one usage might be over design, and I don't like to getting it deprecate in the future.

unlsycn:
> How about directly introducing the tuple for result? Adding new API here for only one usage might be over design, and I don't like to getting it deprecate in the future.

I was using it originally to provide unserialized structures to the user. Now we return a tuple of Writable, which should be used in this way:
```scala
val (firrtl, annos) = designImpl[MySerializableModule, MySerializableModuleParameter](os.read(parameter))
os.write.over(os.pwd / "GCD.fir", firrtl)
os.write.over(os.pwd / "GCD.anno.json", annos)
```

src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
unlsycn: I used it originally to provide unserialized structures to user. Now we return a tuple of geny.Readable hence we can use it in this way:
```scala
val (firrtl, annos) = designImpl[MySerializableModule, MySerializableModuleParameter](os.read.stream(parameter))
os.write.over(os.pwd / "GCD.fir", firrtl)
os.write.over(os.pwd / "GCD.anno.json", annos)
```

src/main/scala/chisel3/experimental/util/SerializableModuleElaborator.scala
```diff
@@ -0,0 +1,77 @@
...
+      ): firrtl.AnnotationSeq
+    ) { case (annos, stage) => stage.transform(annos) }
+      .flatMap {
+        case firrtl.stage.FirrtlCircuitAnnotation(circuit) =>
+          fir = circuit
+          None
+        case _: firrtl.options.Unserializable => None
+        case a => Some(a)
+      }
+    val firrtlStream: Readable = fir.serialize
```
sequencer, line 73: I still doubt there is an extra copy to Sting here, but the return type is good now and we can improve the internal performance with StringBuild in the future.

sequencer (approved):
Maybe you or someone can add more documents in the documentation on how to use the SerializableModule in the future, I think for now the chisel-nix project already give a comprehensive example to it. Please update that to use this Elaborator, and point new users to there.

PR #4415, opened 2024-09-26
Deprecate all Injecting Aspect Code

seldridge:
Deprecate injecting aspect code as of Chisel 6.6.0.  These were planned to be deprecated some time ago (see ROADMAP.md).  However, these had a long user in @sequencer.  He is now off of these and there is nothing blocking their removal.

I will follow this with a commit to remove these from `main` for Chisel 7.

For similar, verification-focused functionality can be selectively added to the circuit, use layers.

#### Release Notes

Deprecate injecting aspects as of Chisel 6.6.0 and plan for removal in Chisel 7.


jackkoenig (approved):
LGTM

PR #4419, opened 2024-09-27
Add documentation for CIRCT Converter

SpriteOvO:
Add documentation for the major components and basic concepts of CIRCT Converter.


jackkoenig:
I like this doc, but I don't think it belongs on the normal website. The website is for Chisel users who are expected to use Chisel via the published artifacts, this is more of documentation for Developers. Can you instead add it to `CONTRIBUTING.md`?

docs/src/appendix/experimental-features.md
```diff
@@ -256,3 +256,61 @@ for working examples.
...
+top.foreachField((name, value) => println(s".$name => { ${value.toString} }"))
+```
+
+### Directory `lit`
+
+We used llvm-lit and scala-cli to test CIRCT Converter.
+
+To run the test, make sure the environment variables are defined first:
+
+- `CIRCT_INSTALL_PATH`: CIRCT installation path.
```
jackkoenig, line 307: This is no longer necessary as the Mill build will download it for you based on the version specified in `etc/circt.json`: https://github.com/chipsalliance/chisel/blob/9ee6842bc700bc11b1bdeae9249008b1751f9a4b/build.sc#L67

docs/src/appendix/experimental-features.md
```diff
@@ -256,3 +256,61 @@ for working examples.
...
+- `JAVA_HOME`: The directory where the Java runtime environment.
```
jackkoenig, line 308: This should probably specify that you need JDK 21+

jackkoenig (approved):

CONTRIBUTING.md
```diff
@@ -32,3 +32,65 @@ From the Chisel root directory, run:
...
+val pm = converter.passManager()
+assert(pm.populateFinalizeIR())
+assert(pm.run())
+
+val om = converter.om()
+val evaluator = om.evaluator()
+
+val top = evaluator.instantiate("Top_Class", Seq(om.newBasePathEmpty)).get
+top.foreachField((name, value) => println(s".$name => { ${value.toString} }"))
+```
```
sequencer, line 80: Please provide an example on how to use the PanamaCIRCTOM in `docs/src/cookbooks`, see our previous [example](https://github.com/chipsalliance/chisel/blob/main/docs/src/cookbooks/serialization.md) here.

CONTRIBUTING.md
```diff
@@ -32,3 +32,65 @@ From the Chisel root directory, run:
...
+
+- `JAVA_HOME`: The directory where the Java runtime environment. The Panama framework requires Java 21 and higher.
+
+```
+mill -i lit[_].run
+```
+
+#### Debugging
+
+When interacting with CIRCT, it is easy to get a JVM crash, which is usually due to an unexpected error inside CIRCT (e.g. an assertion failure). Checking the call stack of the crash will help you to localize the bug.
```
sequencer, line 96: Please add more documentation for debugging, you can put the documentation on `docs/src/developers`, since this flow has been agnostic to all other developers. I think you need to provide a `scala-cli` as an example, you can copy from some thing from the lit folder. Please give a example how to configure the CICRT for debugging, including
1. how to recompile CIRCT to dynamic library to debug info(`-DCMAKE_BUILD_TYPE=RelWithDebInfo`);
2. how to generate the jextract header(point a link to [`common.sc`](https://github.com/chipsalliance/chisel/blob/4f392323e9160440961b9f06e383d3f2742d2f3e/common.sc#L173))
3. how to link the jvm program to the generated dynamic library(some JVM options added on the scala-cli, maybe the [`SmokeTest.sc`](https://github.com/chipsalliance/chisel/blob/main/lit/tests/SmokeTest.sc) is enough);
4. how to hook gdb to to dynamic library(this is really important, since for most of other developer doesn't know this flow, including myself);

SpriteOvO: I don't use gdb to debug it either. All I need is just `println` :/

PR #4430, opened 2024-09-30
Add svsim -Xsource:3 for Scala3 cross compilation support

adkian-sifive:
Part of the larger work of preparing for Scala3 support

### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous printlns/debugging code?
- [ ] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [ ] Did you request a desired merge strategy?
- [ ] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Feature (or new API)
- API modification

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference).

#### Release Notes

* `svsim.Simulation.Port` apply and copy methods are now private, these should never have been public.
* `svsim.CommonCompilationSettings.VerilogPreprocessorDefine` apply and copy methods now private, use other public apply methods.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [ ] Did you mark the proper milestone (Bug fix: `3.6.x`, `5.x`, or `6.x` depending on impact, API modification or big change: `7.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #4431, opened 2024-09-30
[Scala3] Add core -Xsource:3 for Scala3 cross compilation support

adkian-sifive:
Part of the larger work of preparing for Scala3 support

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Adds -Xsource:3 to core in preparation of Scala3 migration. Add stricter type ascriptions for Scala3:
* ResetType.litOption return type changed from `None` to `Option[BigInt]`
* AsyncReset.litOption return type changed from `None` to `Option[BigInt]`
* RequireAsyncReset.resetType type changed from `ResetType.Type` to `Asynchronous.type`
* RequireAsyncReset.resetType type changed from `ResetType.Type` to `Synchronous.type`
* CrossModuleBinding.location return type changed from `None.type` to `Option[BaseModule]`

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -316,7 +316,8 @@ lazy val core = (project in file("core"))
       "-Xcheckinit",
       "-Xlint:infer-any",
       "-Xsource:3",
-      "-Xsource-features:case-apply-copy-access"
+      "-Xsource-features:case-apply-copy-access",
+      "-Xsource-features:infer-override",
 //      , "-Xlint:missing-interpolator"
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -314,7 +314,9 @@ lazy val core = (project in file("core"))
       "-language:reflectiveCalls",
       "-unchecked",
       "-Xcheckinit",
-      "-Xlint:infer-any"
+      "-Xlint:infer-any",
+      "-Xsource:3",
+      "-Xsource-features:case-apply-copy-access"
 //      , "-Xlint:missing-interpolator"
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -316,8 +316,6 @@ lazy val core = (project in file("core"))
       "-Xcheckinit",
       "-Xlint:infer-any",
       "-Xsource:3",
-      "-Xsource-features:case-apply-copy-access",
-      "-Xsource-features:infer-override",
 //      , "-Xlint:missing-interpolator"
     )
   )
--- a/core/src/main/scala/chisel3/BitsImpl.scala
+++ b/core/src/main/scala/chisel3/BitsImpl.scala
@@ -531,7 +531,7 @@ private[chisel3] trait ResetTypeImpl extends Element { self: Reset =>
 
   def cloneType: this.type = Reset().asInstanceOf[this.type]
 
-  override def litOption: None.type = None
+  override def litOption = None
 
   /** Not really supported */
   def toPrintable: Printable = PString("Reset")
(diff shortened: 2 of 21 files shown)


build.sbt
```diff
@@ -310,7 +310,9 @@ lazy val core = (project in file("core"))
       "-language:reflectiveCalls",
       "-unchecked",
       "-Xcheckinit",
-      "-Xlint:infer-any"
+      "-Xlint:infer-any",
+      "-Xsource:3",
+      "-Xsource-features:case-apply-copy-access"
```
jackkoenig, line 315: Unlike firrtl and svsim, this is going to be changing APIs that may actually affect users so we need to potentially report it in the release notes. If you remove this option, what are the errors?

core/src/main/scala/chisel3/BitsImpl.scala
```diff
@@ -528,7 +528,7 @@ private[chisel3] trait ResetTypeImpl extends Element { self: Reset =>
 
   def cloneType: this.type = Reset().asInstanceOf[this.type]
 
-  override def litOption = None
+  override def litOption: None.type = None
```
jackkoenig, line 531: This is similar to the discussion here: https://github.com/chipsalliance/chisel/pull/4403#discussion_r1763647813

We actually want `-Xsource-features:infer-override` because we would rather the type be `Option[BigInt]` than `None.type`. **However**, unlike in the discussion on the previous PR, this change matters so needs to be documented in the release notes.

Looking through, there aren't very many where it matters so that's good, but for any public API, we need to document that the type has changed, e.g.

* `chisel3.ResetType.litOption` type changed from `None.type` to `Option[BigInt]`

jackkoenig: As another example, see the changes I made to the Release Notes in https://github.com/chipsalliance/chisel/pull/4430.
The APIs in svsim matter less since they are commonly used, but many of the APIs here in Chisel matter, like this one.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -276,10 +277,10 @@ abstract class RawModule extends BaseModule {
 
 /** Enforce that the Module.reset be Asynchronous (AsyncReset) */
 trait RequireAsyncReset extends Module {
-  override final def resetType = Module.ResetType.Asynchronous
+  override final def resetType: Module.ResetType.Type = Module.ResetType.Asynchronous
```
mwachs5, line 280: is this really what you want here? I would think that we *do* want to tighten the type here. Same with below

jackkoenig: It's debatable, I could go either way but I think you're probably right that the more concrete type is better here.

core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -127,6 +127,7 @@ case class ClassType private[chisel3] (name: String) { self =>
         override def convertUnderlying(value: Property[ClassType] with self.Type) = value.ref
       }
   }
+  def copy(name: String = this.name) = new ClassType(name)
```
jackkoenig, line 130: I think we want this as private

jackkoenig (approved):
Generally LGTM although a few Release Note tweaks:
* EmptyAlignment is private so can remove that one
* Lots of things have private copy and companion apply methods, those should be mentioned: `Bidirectional`, `Width` (private copy), `RecordAlias`, `Definition`, `Instance`

Also see comment in code.

core/src/main/scala/chisel3/properties/Class.scala
adkian-sifive: afaict these need to be public for scala3..

PR #4434, opened 2024-10-02
Layers: color check all not any and don't consider enabled layers.

dtzSiFive:
Drop test specifically checking this incorrect behavior.

Add tests checking don't need to be under a layer block to define a probe with more colors, and that enabled layers don't change what is allowed to be defined.

### Contributor Checklist

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

#### Release Notes

Fix checking of layer-color probes to avoid rejecting valid uses and to detect actual problems that would otherwise be caught by CIRCT.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/package.scala
+++ b/core/src/main/scala/chisel3/internal/package.scala
@@ -180,8 +180,7 @@ package object internal {
         color
       case _ => return
     }
-    val enabledLayers = Builder.enabledLayers.view ++ Builder.layerStack.headOption
-    if (enabledLayers.exists(_.canWriteTo(destLayer))) {
+    if (Builder.layerStack.headOption.forall(_.canWriteTo(destLayer))) {
       return
     }
     Builder.error(errorMessage)
--- a/src/test/scala/chiselTests/LayerSpec.scala
+++ b/src/test/scala/chiselTests/LayerSpec.scala
@@ -160,6 +160,33 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
     )()
   }
 
+  they should "allow for defines to layer-colored probes without layer blocks" in {
+
+    class Foo extends RawModule {
+      val a, b = IO(Input(Bool()))
+      val x = IO(Output(Probe(Bool(), A)))
+      val y = IO(Output(Probe(Bool(), A.B)))
+      define(x, ProbeValue(a))
+      define(y, ProbeValue(b))
+    }
+
+    ChiselStage.convert(new Foo)
+  }
+
+  they should "allow for defines to layer-colored probes regardless of enabled layers" in {
+
+    class Foo extends RawModule {
+      val a, b = IO(Input(Bool()))
+      val x = IO(Output(Probe(Bool(), A)))
+      val y = IO(Output(Probe(Bool(), A.B)))
+      layer.enable(C)
+      define(x, ProbeValue(a))
+      define(y, ProbeValue(b))
+    }
+
+    ChiselStage.convert(new Foo)
+  }
+
   they should "be enabled with a trait" in {
 
     class Foo extends RawModule {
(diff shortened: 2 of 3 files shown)


src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -160,6 +160,39 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+      val x = IO(Output(Probe(Bool(), A)))
+      val y = IO(Output(Probe(Bool(), A.B)))
+      define(x, ProbeValue(a))
+      define(y, ProbeValue(b))
+    }
+
+    matchesAndOmits(ChiselStage.emitCHIRRTL(new Foo))(
+      "define x = probe(a)",
+      "define y = probe(b)"
+    )()
```
seldridge, line 176: `ChiselStage.elaborate` is sufficient to only check that it doesn't error.

src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -176,22 +209,6 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
-    class Foo extends RawModule {
-      // Without this enable, this circuit is illegal because `C` is NOT enabled
-      // when `A` is enabled.  Cf. tests checking errors of this later.
-      layer.enable(A)
-      val a = IO(Output(Probe(Bool(), A)))
-      layer.block(C) {
-        val b = Wire(Bool())
-        define(a, ProbeValue(b))
-      }
-    }
```
seldridge, line 190: Great to remove this (and fix the checking) given that this emits with an illegal reference. This could be made to work, but we don't even have a way to describe this in the FIRRTL spec, i.e., there is no `layerblock A.C`.

(You know all this! I'm more leaving comments for bread crumbs if others come back to this later...)

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -180,8 +180,7 @@ package object internal {
         color
       case _ => return
     }
-    val enabledLayers = Builder.enabledLayers.view ++ Builder.layerStack.headOption
-    if (enabledLayers.exists(_.canWriteTo(destLayer))) {
+    if (Builder.layerStack.headOption.forall(_.canWriteTo(destLayer))) {
```
seldridge, line 183: Clever use of `forall`. 👍

seldridge (approved):
LGTM

src/test/scala/chiselTests/LayerSpec.scala
dtzSiFive: Breadcrumbs are great! Thanks! I'm unsure about this one, let's check!

I think the code in the test should not be supported as the probe target is not guaranteed to exist when `A` is enabled (`C` must be enabled, so the probe must at least have that color).

Additionally, enabling a layer should still allow sending out probes that aren't colored (not required to have color `A` [1].

This matches CIRCT's interpretation, AFAIK (FWIW).

I'm not sure what the mentioned `A.C` would be, and if it's some sort of sets-of-layers specification (?) not sure that changes anything here?

core/src/main/scala/chisel3/internal/package.scala
dtzSiFive: Revisiting, this is holdover from replacing the previous `exists` with `forall` -- but now that it's a single element (Option) it's not necessary but will leave as-is.

If a block could be colored by `A` AND `B` (where each could be activated independently), we'd want "all" but presently we can continue to just check the top of the stack.

src/test/scala/chiselTests/LayerSpec.scala
dtzSiFive: Good call, thanks! Replaced with `ChiselStage.convert`!

PR #4435, opened 2024-10-03
fix: wrap all intrinsics in layerblocks

### LINKED ISSUE #4429: [FIRRTL] LTL intrinsics are out of the verification layer block
**Type of issue**: Bug Report

There are two problems here: 
First, properties generates LTL intrinsics that are outside the Verification layer block.
Second, the above behavior will cause a firtool error `verification operation used in a non-verification context`, which only occurs in specific cases that I have not figured out, and I will provide a minial example to reproduce below.

```scala
AssumeProperty(
  inputNotValid |=> not(inputFire),
  label = Some("GCD_ASSUMPTION_INPUT_NOT_VALID")
)
```
This Scala code generates the following FIRRTL:

```firrtl
node not = intrinsic(circt_ltl_not : UInt<1>, _T)
node delay = intrinsic(circt_ltl_delay<delay = 1, length = 0> : UInt<1>, UInt<1>(0h1))
node concat = intrinsic(circt_ltl_concat : UInt<1>, _T_5, delay)
node implication = intrinsic(circt_ltl_implication : UInt<1>, concat, not)
node x4_has_been_reset = intrinsic(circt_has_been_reset : UInt<1>, clock, reset)
node x4 = eq(x4_has_been_reset, UInt<1>(0h0))
layerblock Verification :
  layerblock Assume :
    node clock_1 = intrinsic(circt_ltl_clock :
    node _T_6 = eq(x4, UInt<1>(0h0))
    intrinsic(circt_verif_assume<label = "GCD_ASSUMPTION_INPUT_NOT_VALID">, clock_1, _T_6)
```

which works fine with firtool, although I think the intrinsics should be inside the layerblock.

However, this Scala code that just adds another property will break the firtool:

```scala
AssumeProperty(
  inputNotValid |=> not(inputFire),
  label = Some("GCD_ASSUMPTION_INPUT_NOT_VALID")
)

AssertProperty(
  inputFire |=> inputNotFire.repeatAtLeast(1) ### outputFire,
  label = Some("GCD_ALWAYS_RESPONSE")
)
```

The diff tool tells me that the difference between the two firrtl is only the extra property but firtool encounters this error:

```
gcd/src/GCDFormal.scala:68:19: error: verification operation used in a non-verification context
    inputNotValid |=> not(inputFire),
                  ^
gcd/src/GCDFormal.scala:68:19: note: see current operation: %10 = "ltl.delay"(%1) {delay = 1 : i64, length = 0 : i64} : (i1) -> !ltl.sequence
gcd/src/GCDFormal.scala:72:17: note: leaking outside verification context here
  AssertProperty(
                ^
```

The full firrtl files are here:

[good.fir](http://fars.ee/Sf-_)
[bad.fir](http://fars.ee/QI5_)

unlsycn:
### Contributor Checklist

This PR is based on #4436 and fixes #4429.

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Fix LTL intrinsics are generated outside the verification layer block

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/ltl/LTL.scala
+++ b/src/main/scala/chisel3/ltl/LTL.scala
@@ -408,14 +408,15 @@ sealed abstract class AssertPropertyLike(defaultLayer: Layer) {
     *   emitted as `label: assert(...)`.
     */
   def apply(
-    prop:    Property,
+    prop:    => Property,
     clock:   Option[Clock] = Module.clockOption,
     disable: Option[Disable] = Module.disableOption,
     label:   Option[String] = None
   )(
     implicit sourceInfo: SourceInfo
   ): Unit = block(defaultLayer, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
-    val clocked = clock.fold(prop)(prop.clock(_))
+    val _prop = prop // evaluate prop expression once
+    val clocked = clock.fold(_prop)(_prop.clock(_))
     createIntrinsic(label)(sourceInfo)(clocked.inner, disable.map(!_.value))
   }
 
--- a/src/test/scala/chiselTests/LTLSpec.scala
+++ b/src/test/scala/chiselTests/LTLSpec.scala
@@ -383,4 +383,28 @@ class LTLSpec extends AnyFlatSpec with Matchers with ChiselRunners {
       }
     })
   }
+
+  class LayerBlockMod extends RawModule {
+    val a, b = IO(Input(Bool()))
+    implicit val info = SourceLine("Foo.scala", 1, 2)
+    AssertProperty(Sequence(Delay(), a))
+    AssumeProperty(a |-> b)
+  }
+
+  it should "wrap all intrinsics in layerblocks" in {
+    val chirrtl = ChiselStage.emitCHIRRTL(new LayerBlockMod)
+    val sourceLoc = "@[Foo.scala 1:2]"
+
+    val assertBlockLoc = chirrtl.indexOf(s"layerblock Assert : $sourceLoc")
+    val delayIntrinsicLoc = chirrtl.indexOf(
+      s"intrinsic(circt_ltl_delay<delay = 1, length = 0> : UInt<1>, a)"
+    )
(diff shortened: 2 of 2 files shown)


src/main/scala/chisel3/ltl/LTL.scala
```diff
@@ -408,7 +408,7 @@ sealed abstract class AssertPropertyLike(defaultLayer: Layer) {
     *   emitted as `label: assert(...)`.
     */
   def apply(
-    prop:    Property,
+    prop:    => Property,
```
sequencer, line 411: Interesting fixup. LGTM

sequencer (approved):
Wait for #4436, this fix lgtm.

src/main/scala/chisel3/ltl/LTL.scala
jackkoenig: Why is this change here? If you look below, `prop` is accessed multiple times so now this is going to execute multiple times.

jackkoenig (changes requested):

src/main/scala/chisel3/ltl/LTL.scala
unlsycn: It is intended to delay `prop` to be evaluated in the `block`. Actually, it will only execute once since the definition of `fold` is
```scala
@inline final def fold[B](ifEmpty: => B)(f: A => B): B =
  if (isEmpty) ifEmpty else f(this.get)
```

jackkoenig: Good point and very clever, perhaps too clever 🙂. This was not obvious to me (despite the fact that I should know better), so I suspect it won't be obvious to others in the future updating this code. So in the interest of defensive programming against future bugs, please create a val above `val clocked` (but obviously inside the `block` that you assign `prop` to with a comment saying "evaluate by name parameter once", and then use that value in the `fold`.

jackkoenig (approved):
Nice!

src/main/scala/chisel3/ltl/LTL.scala
unlsycn: Thank you for the suggestion.

PR #4437, opened 2024-10-03
switch SRAMDescription to Class

sequencer:
We use this api in our flow to replace the sram-repl: 
See https://github.com/chipsalliance/t1/pull/791/commits/7862959c20f7bee83047b9fb33102aac83036b33

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/util/SRAMSpec.scala
+++ b/src/test/scala/chiselTests/util/SRAMSpec.scala
@@ -3,8 +3,7 @@
 package chiselTests.util
 
 import chisel3._
-import chisel3.properties.{Path, Property}
-import chisel3.util.{MemoryReadWritePort, SRAM, SRAMInterface}
+import chisel3.util.{MemoryReadWritePort, SRAM}
 import chisel3.experimental.{annotate, ChiselAnnotation, OpaqueType}
 import chiselTests.ChiselFlatSpec
 import _root_.circt.stage.ChiselStage.{emitCHIRRTL, emitSystemVerilog}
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/util/SRAMSpec.scala
```diff
@@ -228,25 +228,6 @@ class SRAMSpec extends ChiselFlatSpec {
     val sv = emitSystemVerilog(new Top)
   }
 
-  it should "be possible to access SRAM description information" in {
```
sequencer, line 231: I don't think the needs of accessing the value of "SRAM description information" at elaboration time. actually we just choose nest inner instance's OM to outer module and access them at circt.
See our use case: https://github.com/chipsalliance/t1/pull/791/commits/ec76744ae76c3f2ceebd0a71142fae1d66ddb4f1

jackkoenig: We need the ability to access just the `Path`, that is our specific use case. The `Path` only isn't there right now because firtool currently doesn't have the ability to materialize `Path` Properties in leaf modules, but that's a bug that needs to be fixed.

jackkoenig: If there can be a way to deal with classes and select specific fields that would be great, but AFAIK that does not exist, maybe @mikeurbach will have more info.

sequencer: Maybe we can add an additional property for `Path` in SRAM? It that possible to workaround?

sequencer:
We are prototyping a new API for `SRAMDescription` to extract SRAM based on the `panamaom`, thus snap this into master, feel free to modify for better API.

PR #4443, opened 2024-10-04
Introduce Blocks, track, use for improved scope checking and boring

dtzSiFive:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -2,7 +2,7 @@
 
 package chisel3
 
-import scala.collection.immutable.ListMap
+import scala.collection.immutable.{ListMap, VectorBuilder}
 import scala.collection.mutable.{ArrayBuffer, HashMap, LinkedHashSet}
 
 import chisel3.internal._
@@ -60,6 +60,7 @@ private[chisel3] trait ObjectModuleImpl {
 
     val parent = Builder.currentModule
     val parentWhenStack = Builder.whenStack
+    val parentBlockStack = Builder.blockStack
     val parentLayerStack = Builder.layerStack
 
     // Save then clear clock and reset to prevent leaking scope, must be set again in the Module
@@ -79,10 +80,12 @@ private[chisel3] trait ObjectModuleImpl {
     //   - set currentModule
     //   - unset readyForModuleConstr
     //   - reset whenStack to be empty
+    //   - reset blockStack to body or empty if none
     //   - reset layerStack to be root :: nil
     //   - set currentClockAndReset
     val module: T = bc // bc is actually evaluated here
 
+    require(Builder.blockDepth == module.getBody.size, "block leftover")
     if (Builder.whenDepth != 0) {
       throwException("Internal Error! when() scope depth is != 0, this should have been caught!")
     }
@@ -104,6 +107,7 @@ private[chisel3] trait ObjectModuleImpl {
     // scope of the current Module.
     Builder.currentModule = parent // Back to parent!
     Builder.whenStack = parentWhenStack
+    Builder.blockStack = parentBlockStack
     Builder.layerStack = parentLayerStack
     Builder.currentClock = saveClock // Back to clock and reset scope
     Builder.currentReset = saveReset
@@ -168,10 +172,12 @@ private[chisel3] trait ObjectModuleImpl {
     val parent = Builder.currentModule
     val whenStackOpt = Option.when(Builder.hasDynamicContext)(Builder.whenStack)
     val layerStackOpt = Option.when(Builder.hasDynamicContext)(Builder.layerStack)
+    val blockStackOpt = Option.when(Builder.hasDynamicContext)(Builder.blockStack)
     val module: T = bc // bc is actually evaluated here
     if (!parent.isEmpty) { Builder.currentModule = parent }
     whenStackOpt.foreach(Builder.whenStack = _)
     layerStackOpt.foreach(Builder.layerStack = _)
+    blockStackOpt.foreach(Builder.blockStack = _)
 
     module
   }
@@ -452,6 +458,23 @@ package experimental {
 
     /** Represents an eagerly-determined unique and descriptive identifier for this module */
     final val definitionIdentifier = _definitionIdentifier
+
+    // TODO: this weird re:evaluation/initialization ordering, revisit?
+    // Modules that contain bodies should override this.
+    private[chisel3] def getBody : Option[Block] = None
+
+    // Current block at point of creation.
+    private var _blockVar:       Block = Builder.currentBlock.getOrElse(null)
+    private[chisel3] def _block: Option[Block] = {
+      Option(_blockVar)
+    }
+    private[chisel3] def _block_=(target: Option[Block]): Unit = {
+      _blockVar = target.getOrElse(null)
+    }
+
+    // Return Block containing the instance of this module.
+    protected[chisel3] def getInstantiatingBlock : Option[Block] = _block
+
     //
     // Builder Internals - this tracks which Module RTL construction belongs to.
     //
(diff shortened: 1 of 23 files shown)


core/src/main/scala/chisel3/properties/Class.scala
```diff
@@ -89,10 +89,14 @@ class Class extends BaseModule {
 
   /** Internal state and logic to maintain a buffer of commands.
     */
-  private val _commands = new ArrayBuffer[Command]()
```
dtzSiFive, line 92: By moving this to `Block`, the backing here is now a `VectorBuilder[Command]` instead of an `ArrayBuffer`.

Not sure if there's a preference here or more investigation/characterization of their usage needed, but just drawing attention in case it matters 👍 .

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -274,10 +275,37 @@ object BoringUtils {
...
+                  case Some(bb: BlockBinding) => bb.parentBlock
+                  // Special handling to reach in and get instantiating block for ports.
+                  case Some(pb: PortBinding) if pb.enclosure._parent == Some(conLoc) =>
+                    pb.enclosure.getInstantiatingBlock
+                  case Some(spb: SecretPortBinding) if spb.enclosure._parent == Some(conLoc) =>
+                    spb.enclosure.getInstantiatingBlock
+                  // Otherwise, default behavior.
+                  case _ => None
+                }
+            }
```
jackkoenig, line 302: Feels weird to use `match` for `true` and `false`, but I'll leave it to you to decide.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
```diff
@@ -274,10 +275,37 @@ object BoringUtils {
...
-            if (isDrive) {
-              conLoc.asInstanceOf[RawModule].secretConnection(rhs, bore)
-            } else {
-              conLoc.asInstanceOf[RawModule].secretConnection(bore, rhs)
+
+            // TODO: Check for wiring non-probes not in same block, reject/diagnose.
+
+            // `module` contains the new port.
+            // `conLoc` is the module with the `rhs` value, and where we want to insert.
+            require(conLoc == module || Some(conLoc) == module._parent)
```
jackkoenig, line 283: If this is truly a `require` (implying that bad input from the user can trip it), then it must have an error message for the user to tell them what they did wrong.

jackkoenig (commented):
Generally looks great and love the tests, but see comments.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -789,6 +790,27 @@ private[chisel3] object Builder extends LazyLogging {
...
+    val lastBlock = dynamicContext.blockStack.head
+    dynamicContext.blockStack = dynamicContext.blockStack.tail
+    lastBlock
+  }
+
+  def currentBlock: Option[Block] = dynamicContextVar.value match {
+    case Some(dc) => dc.blockStack.headOption
+    case _        => None
+  }
+
```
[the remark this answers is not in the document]
dtzSiFive: Yes, we definitely should drop the other stacks -- actually, this change is /nearly/ enough to drop the whenStack entirely, but I restored it since it's needed for now to implement `cond`.  Presently Blocks point to their owner commands, which is at different level/layer vs Chisel Data.  But it'd be great to be able to track this with one stack, yes!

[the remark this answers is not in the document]
jackkoenig: Incidentally, my original idea for how to implement `when.cond` was that when you call it, it drops a Bool wire connected to false just before the outermost when (or at the very top of the module but that'd be uglier), then connects it to true in the current when scope. But we couldn't do this because we didn't have the ability to insert at arbitrary points. Your work here would actually make that implementation possible.

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -452,6 +458,23 @@ package experimental {
 
     /** Represents an eagerly-determined unique and descriptive identifier for this module */
     final val definitionIdentifier = _definitionIdentifier
+
+    // TODO: this weird re:evaluation/initialization ordering, revisit?
+    // Modules that contain bodies should override this.
+    private[chisel3] def getBody: Option[Block] = None
```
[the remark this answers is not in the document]
dtzSiFive: I'll play with this, thanks!

It was a bit of a hack to get things online that I meant to revisit.

Looks like https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html covers the situation.

We want the body block on the blockStack ASAP -- to ensure anything evaluated finds the right block.

Presently similar Builder state (whenStack, layerStack) is initialized as part of the BaseModule's body/primary constructor evaluation, at which point a per-subclass override (or definition) for `_body` isn't initialized yet.

src/main/scala/chisel3/util/experimental/BoringUtils.scala
dtzSiFive: Oh wow, yes indeed.  It had more before 🙃 .  I'll address this.

core/src/main/scala/chisel3/ModuleImpl.scala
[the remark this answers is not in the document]
dtzSiFive: I think I have a solution, I'll push soon.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -339,23 +339,40 @@ private[chisel3] object ir {
...
+      _stagedCommands += c
+    }
+    def getCommands(): Seq[Command] = {
+      if (_stagedCommands != null) {
+        _commands ++= _stagedCommands.result()
+        _stagedCommands.clear()
+      }
+      _commands.result()
+    }
   }
```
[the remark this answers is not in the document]
dtzSiFive: Added a boolean to track this to be defensive, thanks!

src/main/scala/chisel3/util/experimental/BoringUtils.scala
dtzSiFive: Done! It's internal but message doesn't hurt anyway 👍 .

core/src/main/scala/chisel3/internal/MonoConnect.scala
```diff
@@ -602,6 +596,16 @@ private[chisel3] object checkConnect {
...
+    checkBlockVisibility(sink) match {
+      case Some(blockInfo) => throw SinkEscapedBlockScopeException(sink, blockInfo)
+      case None            => ()
+    }
+
+    checkBlockVisibility(source) match {
+      case Some(blockInfo) => throw SourceEscapedBlockScopeException(source, blockInfo)
+      case None            => ()
+    }
   }
```
jackkoenig, line 609: I don't have a problem with this, but just curious why you moved this below the Module scoping checks?

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -339,23 +339,47 @@ private[chisel3] object ir {
...
+    private val _commands = new VectorBuilder[Command]
+    private var _secretCommands: mutable.ArrayBuffer[Command] = null
+    private var _closed:         Boolean = false
+
+    def addCommand(c: Command): Unit = {
+      require(!_closed, "cannot add more commands after block is closed")
+      _commands += c
+    }
+    def getCommands(): Seq[Command] = {
+      _closed = true;
```
jackkoenig, line 352:
```suggestion
      _closed = true
```

jackkoenig (approved):
LGTM! As we discussed offline, we probably need to micro-optimize a little to mitigate some of the memory use increase from extra objects, but it's small and we can do that in follow on work.

core/src/main/scala/chisel3/internal/MonoConnect.scala
jackkoenig: Ahh I see in the commit messages:
```
Report error involving general access or module-level access
before more narrowly diagnosing problem with block scope.

ConnectSpec.scala is fixed with this, but seems generally reasonable.
```

PR #4449, opened 2024-10-07
[Scala3] Refactor Printf and VerificationStatement macros

adkian-sifive:
Part of the larger work of preparing for Scala3 support. 

This is likely the first of a series of changes that position macros for compatibility with Scala3. 

### Contributor Checklist

#### Type of Improvement

- Scala3

#### Desired Merge Strategy

#### Release Notes

Separate and move Printf and VerificationStatement macros in preparation of Scala3 support

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala-2/chisel3/VerificationStatementMacros.scala
+++ b/core/src/main/scala-2/chisel3/VerificationStatementMacros.scala
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.SourceInfo
+import chisel3.PrintfMacrosCompat._
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox
+import scala.reflect.macros.blackbox.Context
+import scala.annotation.nowarn
+
+object VerifStmtMacrosCompat {
+  type SourceLineInfo = (String, Int)
+
+  def getLine(c: blackbox.Context): SourceLineInfo = {
+    val p = c.enclosingPosition
+    (p.source.file.name, p.line): @nowarn // suppress, there's no clear replacement
+  }
+
+  object assert {
+
+    /** @group VerifPrintMacros */
+    def _applyMacroWithInterpolatorCheck(
+      c:          blackbox.Context
+    )(cond:       c.Tree,
+      message:    c.Tree,
+      data:       c.Tree*
+    )(sourceInfo: c.Tree
+    ): c.Tree = {
+      import c.universe._
+      _checkFormatString(c)(message)
+      val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
+      q"$apply_impl_do($cond, ${getLine(c)}, _root_.scala.Some(_root_.chisel3.Printable.pack($message, ..$data)))($sourceInfo)"
+    }
+
+    /** @group VerifPrintMacros */
+    def _applyMacroWithStringMessage(
+      c:          blackbox.Context
+    )(cond:       c.Tree,
+      message:    c.Tree,
+      data:       c.Tree
+    )(sourceInfo: c.Tree
+    ): c.Tree = {
+      import c.universe._
+      val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("_applyWithSourceLinePrintable"))
+      q"$apply_impl_do($cond, ${getLine(c)},_root_.scala.Some(_root_.chisel3.Printable.pack($message,..$data)))($sourceInfo)"
+    }
+
+    /** @group VerifPrintMacros */
+    def _applyMacroWithPrintableMessage(
+      c:          blackbox.Context
+    )(cond:       c.Tree,
+      message:    c.Tree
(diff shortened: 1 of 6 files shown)


core/src/main/scala/chisel3/Printable.scala
```diff
@@ -65,7 +65,7 @@ object Printable {
 
   /** Pack standard printf fmt, args* style into Printable
     */
-  def pack(fmt: String, data: Data*): Printable = {
+  def pack(fmt: String, data: DataImpl*): Printable = {
```
adkian-sifive, line 68: undo this maybe

core/src/main/scala/chisel3/Printf.scala
```diff
@@ -109,19 +78,7 @@ object printf {
...
-  )(fmt:        c.Tree,
-    data:       c.Tree*
-  )(sourceInfo: c.Tree
-  ): c.Tree = {
-    import c.universe._
-    _checkFormatString(c)(fmt)
-    val apply_impl_do = symbolOf[this.type].asClass.module.info.member(TermName("printfWithReset"))
-    q"$apply_impl_do(_root_.chisel3.Printable.pack($fmt, ..$data))($sourceInfo)"
-  }
+    macro PrintfMacrosCompat._applyMacroWithInterpolatorCheck
```
jackkoenig, line 81: FWIW, I'm not sure that this will work, I'm pretty sure `macro` the keyword only works in Scala 2. I think we have to do a similar "veneer" thing where the full public method exists only in `src/main/scala-2`:
```scala
def apply(fmt: String, data: Bits*)(implicit sourceInfo: SourceInfo): Printf =
  macro PrintfMacrosCompat._applyMacroWithInterpolatorCheck
```

adkian-sifive: this in fact works in scala3 since the scala3 compiler can read scala2 macro definitions. the compiler can pick between scala2 and scala3 definitions, so while transitioning we need to add scala3 macro definitions while keeping the scala2 ones in the same artifact

jackkoenig: Relevant documentation: https://docs.scala-lang.org/scala3/guides/migration/tutorial-macro-mixing.html

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -30,6 +31,9 @@ trait VerifPrintMacrosDoc
 
 object assert extends VerifPrintMacrosDoc {
```
jackkoenig, line 32: I am a little confused, does this `object assert` not conflict with the one in `scala-2/chisel3/VerificationStatementMacros.scala`?

jackkoenig: Oh I see, it's `chisel3.VerifStmtMacrosCompat.assert`

core/src/main/scala-2/chisel3/VerificationStatementMacros.scala
```diff
@@ -0,0 +1,240 @@
...
+      case Some(msg) =>
+        p"$kind failed: $msg\n"
+      case None => p"$kind failed at $lineMsg\n"
+    }
+  }
+
+  def getLine(c: blackbox.Context): SourceLineInfo = {
+    val p = c.enclosingPosition
+    (p.source.file.name, p.line): @nowarn // suppress, there's no clear replacement
+  }
```
jackkoenig, line 43:
```suggestion
  private[chisel3] type SourceLineInfo = (String, Int)

  private[chisel3] def formatFailureMessage(
    kind:     String,
    lineInfo: SourceLineInfo,
    cond:     Bool,
    message:  Option[Printable]
  )(
    implicit sourceInfo: SourceInfo
  ): Printable = {
    val (filename, line) = lineInfo
    val lineMsg = s"$filename:$line".replaceAll("%", "%%")
    message match {
      case Some(msg) =>
        p"$kind failed: $msg\n"
      case None => p"$kind failed at $lineMsg\n"
    }
  }

  private[chisel3] def getLine(c: blackbox.Context): SourceLineInfo = {
    val p = c.enclosingPosition
    (p.source.file.name, p.line): @nowarn // suppress, there's no clear replacement
  }
```
These were previously in a package private object, let's keep them thatway.

core/src/main/scala/chisel3/VerificationStatement.scala
```diff
@@ -229,116 +137,25 @@ object assume extends VerifPrintMacrosDoc {
...
-    id
-  }
 }
 
 object cover extends VerifPrintMacrosDoc {
 
+  /** Named class for cover statements. */
+  final class Cover private[chisel3] () extends VerificationStatement
+
+  type SourceLineInfo = (String, Int)
```
jackkoenig, line 157: Does this need to be here or can it use the one in `src/main/scala-2`?

jackkoenig (approved):
Generally looks good but see comments.

core/src/main/scala/chisel3/Printable.scala
jackkoenig: Yes please undo this, `DataImpl` is package private

PR #4451, opened 2024-10-07
Fix Boring.rwTap on instance ports

azidar:
Previous behavior would always put the new probe expression in the instance's module body, but this is incompatible with instances built with Instantiate.

Given FIRRTL now supports this, we can improve this API.

Additionally, added a few more D/I + BoringUtils tests

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

BoringUtils.rwTap can now works on a port of an `instance: Instance[..]`

- [ ] Did you review?

### DIFF
--- a/src/test/scala/chiselTests/BoringUtilsSpec.scala
+++ b/src/test/scala/chiselTests/BoringUtilsSpec.scala
@@ -249,53 +249,6 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with 
     )()
   }
 
-  it should "not create new probe port when rwTap is used on a prot" in {
-    import chisel3.experimental.hierarchy._
-    class UnitTestHarness extends Module {
-      val dut = Instantiate(new Dut)
-      probe.force(dut.widgetProbes.head, 0xffff.U)
-    }
-
-    @instantiable
-    class Widget extends Module {
-      @public val in = IO(Input(UInt(32.W)))
-      @public val out = IO(Output(UInt(32.W)))
-      out := ~in
-    }
-
-    @instantiable
-    class Dut extends Module {
-      val widgets: Seq[Instance[Widget]] = Seq.tabulate(1) { i =>
-        val widget = Instantiate(new Widget)
-        widget.in := i.U
-        printf(s"widget[${i}].in = ${i}\n")
-        printf(s"widget[${i}].out = ")
-        printf(cf"${widget.out}\n")
-        widget
-      }
-      @public val widgetProbes = widgets.map { widget =>
-        val widgetProbe = IO(RWProbe(UInt(32.W)))
-        val define = BoringUtils.rwTap(widget.out)
-        probe.define(widgetProbe, define)
-        widgetProbe
-      }
-    }
-    val str =
-      matchesAndOmits(circt.stage.ChiselStage.emitCHIRRTL(new UnitTestHarness, args))(
-        "module Widget :",
-        "input clock : Clock",
-        "input reset : Reset",
-        "input in : UInt<32>",
-        "output out : UInt<32>",
-        "node _out_T = not(in)",
-        "connect out, _out_T",
-        "module Dut :",
-        "define widgetProbes_0 = rwprobe(widgets_0.out)",
-        "public module UnitTestHarness :",
-        "force(clock, _T, dut.widgetProbes_0, UInt<32>(0hffff))"
-      )()
-  }
-
   it should "work boring upwards" in {
     import chisel3.experimental.hierarchy._
     class Bar(parentData: Data) extends RawModule {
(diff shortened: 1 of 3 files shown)


src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -573,6 +573,53 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
+    }
+
+    @instantiable
+    class Dut extends Module {
+      val widgets: Seq[Instance[Widget]] = Seq.tabulate(1) { i =>
+        val widget = Instantiate(new Widget)
+        widget.in := i.U
+        printf(s"widget[${i}].in = ${i}\n")
+        printf(s"widget[${i}].out = ")
+        printf(cf"${widget.out}\n")
```
mwachs5, line 597: are these printfs important?

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -573,6 +573,53 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
     val verilog = circt.stage.ChiselStage.emitSystemVerilog(new Top(Definition(new Foo)))
   }
 
+  it should "work to rwTap an Instance[..]'s port" in {
```
mwachs5, line 576: do we already have a test for just tap (vs rwTap)?

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
azidar: Nope

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
azidar: I don't think so, I can add it?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -212,6 +212,43 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+    }
+    matchesAndOmits(circt.stage.ChiselStage.emitCHIRRTL(new Foo, args))(
+      "module Bar :",
+      "output out : UInt<1>",
+      "module Foo :",
+      "connect sink, bar.out"
+    )()
+  }
+
+  it should "work if boring to an Instance's output port" in {
```
mwachs5, line 234:
```suggestion
  it should "work if boring to an Instance's input port" in {
```

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -573,6 +566,50 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
+    import chisel3.experimental.hierarchy._
+    class UnitTestHarness extends Module {
+      val dut = Instantiate(new Dut)
+      probe.force(dut.widgetProbes.head, 0xffff.U)
+    }
+
+    @instantiable
+    class Widget extends Module {
+      @public val in = IO(Input(UInt(32.W)))
```
mwachs5, line 578: should we test rwProbe of Input vs Output (or does the behavior not matter)?

jackkoenig:
@azidar I changed Milestone to `6.x` because BoringUtils implemented in Chisel has only existed since `v6.0.0-M1`. Feel free to change back if you were still planning to backport to `5.x` and `3.6.x`.

jackkoenig:
In fact maybe we shouldn't backport at all since presumably firtool used with `6.x` doesn't support it? Unclear.

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
azidar: This is testing a rwTap of a RWProbe output port, the other tests in `BoringUtilsTap` tests rwTap's of inputs and outputs.

mwachs5: oh yeah there is no such thing as RWProbe Input Port 👍

src/test/scala/chiselTests/BoringUtilsTapSpec.scala
```diff
@@ -573,6 +677,93 @@ class BoringUtilsTapSpec extends ChiselFlatSpec with ChiselRunners with Utils wi
...
+      "node _out_T = not(in)",
+      "connect out, _out_T",
+      "module Dut :",
+      "define widgetProbes_0 = rwprobe(widgets_0.out)",
+      "public module UnitTestHarness :",
+      "force(clock, _T, dut.widgetProbes_0, UInt<32>(0hffff))"
+    )()
+  }
+
+  it should "work to tap an Instance[..]'s port" in {
```
mwachs5, line 723: this looks good, wondering if you wouldn't want a similar test to the one above where if you `tap` to a Probe port it should work. But not blocking my review on that

mwachs5 (approved):
could add more test (unless I am missing them) for normal `tap` to a Probe port should work. And should it fail if you try to rwTap to a Probe Port? tap to a RWProbePort? can add those cases later though as they are sort of unrelated to what you are fixing, i think?

PR #4454, opened 2024-10-09
Fix addAttribute and add AttributesAnnotationSpec

unlsycn:
`chisel3.utils` is not a chisel package, change back to `chisel3.util` 
### Contributor Checklist

#### Type of Improvement

- API modification
- Bugfix

#### Desired Merge Strategy

#### Release Notes

Fix addAttribute and add AttributesAnnotationSpec.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/AttributeAnnotation.scala
+++ b/src/main/scala/chisel3/util/AttributeAnnotation.scala
@@ -17,16 +17,16 @@ object addAttribute {
     *  import chisel3._
     *  import chisel3.util.addAttribute
     *  class AttributeExample extends Module {
-    *    val io = IO(new Bundle{
-    *      val input = Input(UInt(8))
-    *      val output = Output(UInt(8))
+    *    val io = IO(new Bundle {
+    *      val input = Input(UInt(8.W))
+    *      val output = Output(UInt(8.W))
+    *    })
     *
-    *      val reg = RegNext(io.input)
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/util/AttributeAnnotation.scala
```diff
@@ -17,16 +17,16 @@ object addAttribute {
     *  import chisel3._
     *  import chisel3.util.addAttribute
     *  class AttributeExample extends Module {
-    *    val io = IO(new Bundle{
-    *      val input = Input(UInt(8))
-    *      val output = Output(UInt(8))
+    *    val io = IO(new Bundle {
```
sequencer, line 20: I think you don’t need to edit the scaladoc here?

sequencer (changes requested):

unlsycn:
I'm not very sure about the behavior of `addAttribute` and this test only checks if the attribute exists in the final Verilog while not caring about the location.

src/main/scala/chisel3/util/AttributeAnnotation.scala
unlsycn: Splitting it into a separate commit might be a better choice.

sequencer:
since this is not backported to older versions, I mark it as bugfix.

sequencer (approved):

PR #4462, opened 2024-10-10
[mill] Copy scalacOptions over to mill, add instructions to CONTRIBUTING

jackkoenig:
This is the official switch over to Mill. All developers, please start using Mill for normal development from the merging of this PR on. Please see `CONTRIBUTING.md` for initial docs and add to them as needed.

Remaining things to migrate (future PRs):
* Mdoc
* Website automations

@adkian-sifive please help me ensure that between https://github.com/chipsalliance/chisel/pull/4431 and this, we add `-Xsource:3` to the chisel build unit in the mill build.

### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [ ] Did you add appropriate documentation in `docs/src`?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- Internal or build-related (includes code refactoring/cleanup)

#### Desired Merge Strategy

- Squash

#### Release Notes

Normal development should use Mill instead of SBT from now on.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels? (Select the most appropriate one based on the "Type of Improvement")
- [ ] Did you mark the proper milestone (Bug fix: `3.6.x`, `5.x`, or `6.x` depending on impact, API modification or big change: `7.0`)?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you do one of the following when ready to merge:
  - [ ] Squash: You/ the contributor `Enable auto-merge (squash)`, clean up the commit message, and label with `Please Merge`.
  - [ ] Merge: Ensure that contributor has cleaned up their commit history, then merge with `Create a merge commit`.

PR #4466, opened 2024-10-14
[CI] Use Mill for testing

jackkoenig:
https://github.com/chipsalliance/chisel/pull/4462 broke Chisel because I accidentally deleted the scalacOption for the plugin from the Chisel module. This was missed because CI is still run with SBT. This PR will fix the issue and switch to using Mill for CI so we don't make this mistake again.

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,10 @@ jobs:
           $CHISEL_FIRTOOL_PATH/firtool -version >> $GITHUB_STEP_SUMMARY
           echo \`\`\` >> $GITHUB_STEP_SUMMARY
       - name: Test
-        run: sbt ++${{ inputs.scala }} test
+        run: |
+          ./mill firrtl[_].test
+          ./mill svsim[_].test
+          ./mill chisel[_].test
       - name: Binary compatibility
         # TODO either make this also check the plugin or decide that we don't
         #      support binary compatibility for the plugin
@@ -155,7 +158,6 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: '11'
-          cache: 'sbt'
       - name: Install CIRCT
         id: install-circt
         if: ${{ inputs.circt }}
@@ -170,7 +172,7 @@ jobs:
           dir=$(dirname $(which firtool))
           echo "CHISEL_FIRTOOL_PATH=$dir" >> "$GITHUB_ENV"
       - name: Integration Tests
-        run: sbt integrationTests/test
+        run: ./mill integrationTests[_].test
 
   # Currently just a sanity check that the benchmarking flow works
   benchmark:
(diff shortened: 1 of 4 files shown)


jackkoenig:
First testing the CI without the fix to confirm we can detect the previous breakage.

jackkoenig:
I switched the Label to "Internal" because this is really a CI change, and the bug is only visible on 1 SNAPSHOT commit, no released versions.

.github/workflows/test.yml
```diff
@@ -207,9 +209,8 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: '11'
-          cache: 'sbt'
-      - name: Unit Tests
-        run: sbt ++${{ inputs.scala }} standardLibrary/test
+      - name: Compile
+        run: ./mill stdlib[_].compile
```
jackkoenig, line 213: There are no stdlib tests...

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -813,13 +813,23 @@ private[chisel3] object Builder extends LazyLogging {
...
+        dc.value
+      } catch {
+        // If user uses -Xcheckinit there will be additional information
+        case UninitializedFieldError(msg) => err(msg)
+      }
+    // If user is not using -Xcheckinit, suggest that they do
+    if (result == null) {
+      err("Try adding -Xcheckinit to your scalacOptions to get a more useful stack trace")
+    }
     result
```
jackkoenig, line 833: This was needed to fix some tests that failed due to our now using `-Xcheckinit`. I made it such that the error suggestion works both with and without `-Xcheckinit` and either gives you the additional context from using that option or suggests you use `-Xcheckinit`.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -813,13 +813,23 @@ private[chisel3] object Builder extends LazyLogging {
...
+    val result =
+      try {
+        dc.value
+      } catch {
+        // If user uses -Xcheckinit there will be additional information
+        case UninitializedFieldError(msg) => err(msg)
+      }
+    // If user is not using -Xcheckinit, suggest that they do
+    if (result == null) {
+      err("Try adding -Xcheckinit to your scalacOptions to get a more useful stack trace")
```
seldridge, line 831: Nit: missing trailing period.

seldridge (approved):
LGTM

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: The trailing period is added by `def err` above! But had forgotten to add trailing periods to the comments

PR #4467, opened 2024-10-15
[Scala3] Redux: Convert cloneType to an extension method

adkian-sifive:
Redux of #3771. Turns out we need this for Scala3 after all

Ran into errors related to this while adding Scala3 cross-compilation support. Here's a sample of what an error with Dotty looks like:

```
[error] -- [E007] Type Mismatch Error: chisel3/core/src/main/scala/chisel3/AggregateImpl.scala:838:4 
[error] 838 |    clone
[error]     |    ^^^^^
[error]     |    Found:    (clone : chisel3.Record & chisel3.RecordImpl)
[error]     |    Required: (RecordImpl.this : chisel3.Record & chisel3.RecordImpl)
```

I suspect this is due to the tightening of the type system in Scala3, specifically changes which make `this.type` a more stricter representation of the current instance. I think this SIP might have a more detailed specification of the related changes: https://docs.scala-lang.org/sips/42.type.html#related-scala-issues-resolved-by-the-literal-types-implementation

### Contributor Checklist

#### Type of Improvement

- API Modification

#### Desired Merge Strategy

#### Release Notes
Move cloneType to an extension method in preparation for Scala3 

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/BitsImpl.scala
+++ b/core/src/main/scala/chisel3/BitsImpl.scala
@@ -21,9 +21,9 @@ private[chisel3] trait BitsImpl extends Element { self: Bits =>
   // Arguments against: generates down to a FIRRTL UInt anyways
 
   // Only used for in a few cases, hopefully to be removed
-  private[chisel3] def cloneTypeWidth(width: Width): this.type
+  private[chisel3] def cloneTypeWidth(width: Width): Bits
 
-  def cloneType: this.type = cloneTypeWidth(width)
+  override def _cloneType: Data = cloneTypeWidth(width)
 
   /** A non-ambiguous name of this `Bits` instance for use in generated Verilog names
     * Inserts the width directly after the typeName, e.g. UInt4, SInt1
@@ -167,7 +167,7 @@ private[chisel3] trait BitsImpl extends Element { self: Bits =>
   // Pad literal to that width
   protected def _padLit(that: Int): this.type
 
-  protected def _padImpl(that: Int)(implicit sourceInfo: SourceInfo): this.type = this.width match {
+  protected def _padImpl(that: Int)(implicit sourceInfo: SourceInfo): Bits = this.width match {
     case KnownWidth(w) if w >= that => this
     case _ if this.isLit            => this._padLit(that)
     case _                          => binop(sourceInfo, cloneTypeWidth(this.width.max(Width(that))), PadOp, that)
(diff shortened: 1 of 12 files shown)


jackkoenig:
> Turns out we need this for Scala3 after all

Can you explain more about _why_ we need this? I know we saw errors related to it, but is there documentation or even a thread somewhere in Scala 3 development explaining the change?

adkian-sifive:
@jackkoenig updated the comment above  with some more info

core/src/main/scala/chisel3/BitsImpl.scala
```diff
@@ -558,7 +556,7 @@ private[chisel3] trait AsyncResetImpl extends Element { self: AsyncReset =>
 
   override def toString: String = stringAccessor("AsyncReset")
 
-  def cloneType: this.type = AsyncReset().asInstanceOf[this.type]
+  override def _cloneType: Data = AsyncReset()
```
mwachs5, line 559:
```suggestion
  override def _cloneType: AsyncReset = AsyncReset()
```

? perhaps? Why is this one different than the others

core/src/main/scala/chisel3/ClockImpl.scala
```diff
@@ -13,7 +13,7 @@ private[chisel3] trait ClockImpl extends Element {
 
   override def toString: String = stringAccessor("Clock")
 
-  def cloneType: this.type = Clock().asInstanceOf[this.type]
+  override def _cloneType: Data = Clock()
```
mwachs5, line 16: why isn't this one `Clock`?

core/src/main/scala/chisel3/DataImpl.scala
```diff
@@ -1059,6 +1044,33 @@ private[chisel3] trait ObjectDataImpl {
       }
     }
   }
+
+  implicit class DataExtensions[T <: Data](self: T) {
+
+    /** Internal API; Chisel users should look at chisel3.chiselTypeOf(...).
```
mwachs5, line 1050: If this is an internal API, can we make it private to chisel?

mwachs5 (commented):
high level, why are we sometimes having the `_cloneType` return type be Data and other times we are having the explicit type?

jackkoenig:
I'm pretty sure `this.type` is fine in Scala 3, there's just a problem we run into occasionally which I've captured in a scala-cli script:
```scala
//> using scala "3.3.4"

trait Data {
  var name: String = "<default>"

  def cloneType: this.type

  override def toString: String = s"Data($name)"
}

class UInt(_name: String) extends Data {
  this.name = _name

  override def cloneType: this.type =
    new UInt(s"Clone of ${_name}").asInstanceOf[this.type]

  def withName(_name: String): this.type = {
    // Without `: this.type` we get this compile error below
    // Found:    (clone : UInt)
    // Required: (UInt.this : UInt)
    //     clone
    //     ^^^^^
    val clone: this.type = this.cloneType
    clone.name = _name
    clone
  }
}

object Main extends App {
  val u = new UInt("u")
  val v = u.withName("v")
  println(u)
  println(v)
}
```

It just requires more type annotations when using `this.type` to return a `this.type`. This should be fine for Scala 3 in Chisel.

PR #4479, opened 2024-10-17
[Module] Add afterModuleBuilt hook

fabianschuiki:
Add the `afterModuleBuilt` hook to `RawModule`, similar to the existing `atModuleBodyEnd`. This hook is called after a module has been fully constructed and its resulting component definition has been added to the builder. Thus code inside the body of `afterModuleBuilt` can get the definition of the surrounding module and instantiate it, which is interesting for generating additional collateral like unit tests alongside a module.

### Contributor Checklist

#### Type of Improvement
- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes
The new `afterModuleBuilt` hook can be used to schedule code to be executed once a module has been fully generated and its definition is available. This allows further collateral such as unit tests to be generated alongside a module.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/RawModule.scala
+++ b/core/src/main/scala/chisel3/RawModule.scala
@@ -82,6 +82,46 @@ abstract class RawModule extends BaseModule {
   }
   private val _atModuleBodyEnd = new ArrayBuffer[() => Unit]
 
+  /** Hook to invoke hardware generators after a Module has been constructed and closed.
+    *
+    * This is useful for running hardware generators after a Module's constructor has run and its Definition is available, while still having access to arguments and definitions in the constructor. The Module itself can no longer be modified at this point.
+    *
+    * An interesting application of this is the generation of unit tests whenever a module is instantiated. For example:
+    *
+    * {{{
+    *   class Example(N: int) extends RawModule {
+    *     private val someSecret: Int = ...
+    *
+    *     afterModuleBuilt {
+    *       // Executes once the surrounding module is closed.
+    *       // We can get its definition at this point and pass it to another module.
+    *       Definition(ExampleTest(this.definition, someSecret))
+    *     }
+    *   }
+    *
+    *   class ExampleTest(unitDef: Definition[Example], someSecret: Int) extends RawModule {
+    *     // Instantiate the generated module and test it.
+    *     val unit = Instance(unitDef)
+    *     ...
+    *   }
+    *
+    *   class Parent extends RawModule {
+    *     Instantiate(Example(42))
+    *   }
+    *
+    *   // Resulting modules:
+    *   // - Parent (top-level)
+    *   //   - instantiates Example
+    *   // - ExampleTest (top-level)
+    *   //   - instantiates Example
+    *   // - Example
+    * }}}
+    */
+  protected def afterModuleBuilt(gen: => Unit): Unit = {
+    _afterModuleBuilt += { () => gen }
+  }
+  private val _afterModuleBuilt = new ArrayBuffer[() => Unit]
+
   //
   // RTL construction internals
   //
@@ -252,6 +292,9 @@ abstract class RawModule extends BaseModule {
     val rhs = computeConnection(left, right)
     Builder.currentBlock.get.addSecretCommand(rhs)
   }
+
+  // Evaluate any afterModuleBuilt generators.
+  protected[chisel3] override def moduleBuilt(): Unit = _afterModuleBuilt.foreach(_())
 
   private[chisel3] def initializeInParent(): Unit = {}
 }
--- a/src/test/scala/chiselTests/RawModuleSpec.scala
+++ b/src/test/scala/chiselTests/RawModuleSpec.scala
@@ -4,7 +4,7 @@ package chiselTests
 
 import chisel3._
 import chisel3.aop.Select
-import chisel3.experimental.hierarchy.Definition
+import chisel3.experimental.hierarchy.{instantiable, public, Definition, Instance}
 import chisel3.reflect.DataMirror
 import chisel3.testers.BasicTester
 import circt.stage.ChiselStage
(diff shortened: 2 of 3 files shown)


sequencer:
Good idea! I have been thinking how to instantiating the corresponding TestBench from DUT. I think this PR gives us a acceptable solution. I wonder if there is any related API in CIRCT to elaborate Verilog for multiple top Verilog?

jackkoenig (commented):
I am interested in this API, and you're showing an interesting use case in the tests so far, but this presents a pretty big API surface area and we should make sure to work through some of the consequences:
1. Done correctly, we should be able to implement the hooking up of `ImplicitClock` and `ImplicitReset` with this API.
2. I think this should work for `Instance[_]` (like the module where `afterModuleBuilt` is itself being instantiated as an `Instance[_]`). This has pretty broad implications--it means you should not really be capturing anything from the body of the Module, but instead the type of the thunk passed to `afterModuleBuilt` probably should be `Instance[T] => Unit` so that you can access any public fields of the instance.
3. What happens if you use `afterModuleBuilt` in the top module? For implementing (1), it should be a no-op, so maybe the type should be `Option[Instance[T]] => Unit`.
4. Unfortunately I'm not sure we can prevent people capturing values from the Definition that they shouldn't (maybe we can eventually in Scala 3 with [Capture Checking](https://docs.scala-lang.org/scala3/reference/experimental/cc.html)), but perhaps this is just an advanced API and we tell users to be careful and "don't screw up" 🤷‍♀️

mwachs5:
Is this something that would let you drive the IOs of the module to default values / DontCare?

src/test/scala/chiselTests/RawModuleSpec.scala
```diff
@@ -170,4 +170,74 @@ class RawModuleSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+
+    matchesAndOmits(chirrtl)(
+      "module Foo",
+      "input port0 : UInt<1>",
+      "module Bar",
+      "input port1 : UInt<1>"
+    )()
+  }
+
+  "RawModule with afterModuleBuilt" should "be able to instantiated the surrounding module" in {
```
mwachs5, line 195:
```suggestion
  "RawModule with afterModuleBuilt" should "be able to be instantiated in a surrounding module" in {
```

is that what you meant?

jackkoenig:
> Is this something that would let you drive the IOs of the module to default values / DontCare?

I believe it should make that possible, see my comment about implementing `ImplicitClock` and `ImplicitReset` which requiring connecting them up in the instantiating module.

fabianschuiki:
Let me think through some of the stuff @jackkoenig and @mwachs5 have been pointing out.

Basically, the idea is to have a callback/lambda/block of code executed after a module's definition becomes available such that you can instantiate the module from tests. This expressly doesn't want to know about the concrete instance that triggered the module build, and it doesn't want to be able to interact with that instance in anyway (i.e. no connecting of ports). So I don't think the block should have any `Instance` argument, or know about the instantiation site at all.

It's important that the callback is able to capture values from the original module body though, mainly parameters, constructor arguments, and anything else that's lying around. However, since the module is closed, you can't connect to any of the hardware or change it in any way. You can, however, inspect its types and the like. So all this does is give you the opportunity to construct additional sibling module definitions that themselves instantiate the surrounding module. I don't think this really interacts with any of things you pointed out @jackkoenig.

Maybe the name should be something like `afterDefinitionAvailable`, `withDefinition`, `onceDefined`, or similar, to avoid confusion?

src/test/scala/chiselTests/RawModuleSpec.scala
fabianschuiki: Sorry, this should read: "be able to instantiate the surrounding module". That's the use case for deriving unit tests.

fabianschuiki:
The following is the concrete unit testing API you could build around this:
```scala
trait Testable extends RawModule {
  def unitTest(gen: => Unit): Unit = {
    val parent = this
    afterModuleBuilt {
      Definition(new RawModule with Public {
        override def desiredName = parent.desiredName + "UnitTest"
        gen
      })
    }
  }
}

class Adder(width: Int) extends RawModule {
  override def desiredName = f"Adder${width}"
  val a = IO(Input(UInt(width.W)))
  val b = IO(Input(UInt(width.W)))
  val z = IO(Output(UInt(widht.W)))
  z := a + b

  unitTest {
    val dut = Instance(this.toDefinition)
    dut.a := 1.U << (width - 1)
    dut.b := 1.U
    Assert(dut.z === 0.U, label = f"${width} bit adder should overflow")
  }
}

class Top extends RawModule {
  val a0 = Instantiate(Adder(5))
  val a1 = Instantiate(Adder(42))
}

// Expected modules:
// - Top (instantiates Adder5 and Adder 42)
// - Adder5
// - Adder42
// - Adder5UnitTest (instantiates Adder5)
// - Adder42UnitTest (instantiates Adder42)
```

src/test/scala/chiselTests/RawModuleSpec.scala
dtzSiFive, line 195:
```suggestion
  "RawModule with afterModuleBuilt" should "be able to instantiate the surrounding module" in {
```

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -253,6 +293,11 @@ abstract class RawModule extends BaseModule {
     Builder.currentBlock.get.addSecretCommand(rhs)
   }
 
+  protected[chisel3] override def moduleBuilt(): Unit = {
+    // Evaluate any afterModuleBuilt generators.
+    _afterModuleBuilt.foreach { gen => gen() }
+  }
```
seldridge, line 299: Nit:
```suggestion
  // Evaluate any afterModuleBuilt generators.
  protected[chisel3] override def moduleBuilt(): Unit = _afterModuleBuilt.foreach(_())
```

If that is too weird, then `_.apply()` may be better.

src/test/scala/chiselTests/RawModuleSpec.scala
```diff
@@ -170,4 +170,74 @@ class RawModuleSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+      port1 := port0
+
+      afterModuleBuilt {
+        val fooDef = this.toDefinition
+        Definition(new RawModule {
+          override def desiredName = "Bar1"
+          val port2 = IO(Input(Bool()))
+          val foo = Instance(fooDef)
+          foo.port0 := port2
+        })
```
seldridge, line 210: Does this _need_ to be a `Definition`? What happens if this is wrapped with a `Module.apply`?

src/test/scala/chiselTests/RawModuleSpec.scala
```diff
@@ -170,4 +170,74 @@ class RawModuleSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+    )()
+  }
+
+  "RawModule with afterModuleBuilt" should "be able to instantiate the surrounding module" in {
+    @instantiable
+    class Foo extends RawModule {
+      override def desiredName = "Foo"
+      @public val port0 = IO(Input(Bool()))
+      @public val port1 = IO(Output(Bool()))
+      port1 := port0
```
seldridge, line 201: Mega-nit, I'd golf this down to a single input port for maintainability.

seldridge (approved):
SGTM

We are starting to want to play around with this for future testing APIs. This may evolve in the future, but this is good enough for me.

A few nits about the tests. Looks good otherwise.

src/test/scala/chiselTests/RawModuleSpec.scala
fabianschuiki: I don't think so. This should just work as well with `Module`. Let me swap that around.

src/test/scala/chiselTests/RawModuleSpec.scala
fabianschuiki: :golf: :+1:

fabianschuiki:
Thanks for the feedback @seldridge! I've tweaked things a bit. Should be good to go. I don't think I have permission to merge this myself though :crying_cat_face:

PR #4483, opened 2024-10-22
[Scala3] Scala2 compatibility changes

adkian-sifive:
Minor changes that are the Scala2-specific changes needed to cross-compile with Scala3:

- Move factory objects from package.scala to live with their classes as required by Dotty
- Move reflection uses to a utility object in `scala-2/chisel3/internal/util.scala`
- Refactor Printf and VerificationStatement
- Add missing references to `this` for compatibility with Dotty

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

Minor refactoring for compatibility with Scala3. Factory objects from package.scala are moved to live with their defining classes for compatibility with Dotty. Macros and reflection moved to scala-2 API directory.

- [ ] Did you review?


core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -544,6 +544,8 @@ private[chisel3] object ir {
         _.toFirrtl.flatMap(_.update(renames))
       )
 
+    // TODO this method doesn't compile with Scala3 because it
+    // conflicts with the built-in `copy` method
     def copy(
       name:        String = name,
       components:  Seq[Component] = components,
```
jackkoenig, line 551: This was added for backwards compatibility reasons when the classes in this file were [accidentally] public, this is private now so we can feel free to change it--it'd be best to just delete this, the extra constructor above, and the companion object below.

jackkoenig: (But that need not be part of this PR)

adkian-sifive: ok i'll do it separately once this goes through

core/src/main/scala-2/chisel3/Bits.scala
```diff
@@ -182,8 +182,6 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends BitsImpl w
     * $sumWidthInt
     * @group Bitwise
     */
-  // REVIEW TODO: redundant
-  // REVIEW TODO: should these return this.type or Bits?
   final def <<(that: BigInt): Bits = macro SourceInfoWhiteboxTransform.thatArg
```
jackkoenig, line 185: These TODOs are actually somewhat relevant, or at least highlight something we need to handle in the Scala 3 upgrade. While `Bits.<<` returns `Bits` (even for `UInt` and `SInt`), `UInt.do_<<` returns `UInt`, and because this is a whitebox macro, it effectively means that `UInt.<<` returns `UInt` even though the ScalaDoc says it returns `UInt`.

That sounds a bit abstract, but consider the following:
```scala
class Foo extends Module {
  val in = IO(Input(UInt(8.W)))                                                                                                                                                                                 

  out :#= in >> 3 // This works because in >> 3 is a UInt
  out :#= (in: Bits) >> 3 // This doesn't because it returns Bits
}
```
Basically, we need to make sure to override `<<` (and related functions like `>>`) in our Scala 3 facades for `SInt` and `UInt` to make sure we return the right concrete subtypes.

core/src/main/scala-2/chisel3/VerificationStatement.scala
```diff
@@ -199,7 +201,7 @@ object stop {
       message.foreach(Printable.checkScope(_))
       when(!Module.reset.asBool) {
         message.foreach(PrintfMacrosCompat.printfWithoutReset(_))
-        pushCommand(Stop(stopId, sourceInfo, Builder.forcedClock.ref, 0))
+        pushCommand(chisel3.internal.firrtl.ir.Stop(stopId, sourceInfo, Builder.forcedClock.ref, 0))
       }
     }
     stopId
```
jackkoenig, line 207: Rather than just moving this file (and `Printf.scala`), I think it would be best to them in the same style as other things even though I know there's not much that can be shared, there are still some common types and utilities.

jackkoenig (commented):
Generally looks good but see comments.

core/src/main/scala-2/chisel3/experimental/dataview/InvertibleDataView.scala
```diff
@@ -0,0 +1,37 @@
+package chisel3.experimental.dataview
+
+import chisel3._
+import scala.reflect.runtime.universe.WeakTypeTag
+
+object InvertibleDataView {
+  private def swapArgs[A, B, C, D](f: (A, B) => Iterable[(C, D)]): (B, A) => Iterable[(D, C)] = {
```
adkian-sifive, line 7: this repeated method is not great (also exists in class DataView), maybe fix

core/src/main/scala-2/chisel3/internal/util.scala
```diff
@@ -0,0 +1,44 @@
...
+      case (Some(targetWidth), Some(thatWidth)) =>
+        if (targetWidth == thatWidth) that
+        else if (targetWidth > thatWidth) _padHandleBool(that, targetWidth)
+        else fromUInt(that.take(targetWidth))
+      case (Some(targetWidth), None) => fromUInt(_padHandleBool(that, targetWidth).take(targetWidth))
+      case (None, Some(thatWidth))   => that
+      case (None, None)              => that
+    }
+  }
+}
```
jackkoenig, line 44: Rather than splitting this out, just remove the need for the TypeTag. It's a bit shitty but you can just replace the type tag with a Boolean argument `isSigned` and compare with that.

core/src/main/scala/chisel3/internal/package.scala
```diff
@@ -69,40 +67,6 @@ package object internal {
...
-    (targetWidthOpt, that.widthOption) match {
-      case (Some(targetWidth), Some(thatWidth)) =>
-        if (targetWidth == thatWidth) that
-        else if (targetWidth > thatWidth) _padHandleBool(that, targetWidth)
-        else fromUInt(that.take(targetWidth))
-      case (Some(targetWidth), None) => fromUInt(_padHandleBool(that, targetWidth).take(targetWidth))
-      case (None, Some(thatWidth))   => that
-      case (None, None)              => that
-    }
-
```
jackkoenig, line 105: See comment elsewhere but rather than removing this, just rewrite it without the type tag (adding Boolean argument)

jackkoenig (commented):
Generally looks good but see comments

core/src/main/scala-2/chisel3/experimental/dataview/InvertibleDataView.scala
```diff
@@ -0,0 +1,37 @@
...
+        val vv = implicitly[WeakTypeTag[V]].tpe
+        val msg = s"Cannot invert '$view' as it is non-total.\n  Try providing a DataView[$vv, $tt]." +
+          s"\n  Please see https://www.chisel-lang.org/chisel3/docs/explanations/dataview."
+        throw InvalidViewException(msg)
+      }
+      implicit val sourceInfo = view.sourceInfo
+      new DataView[V, T](mkView, swapArgs(view.mapping), view.total)
+    }
+  }
+}
```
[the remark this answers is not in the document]
adkian-sifive: yeah this should've been moved over not copied, issue with rsync-ing from a different repo..

core/src/main/scala-2/chisel3/Printf.scala
```diff
@@ -13,33 +13,11 @@ import scala.language.experimental.macros
   *
   * See apply methods for use
   */
-object printf {
+object printf extends PrintfImpl {
 
   /** Named class for [[printf]]s. */
   final class Printf private[chisel3] (val pable: Printable) extends VerificationStatement
```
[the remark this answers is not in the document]
adkian-sifive: so i didn't move these types Printf, Assume, Assert cause they're referenced elsewhere as types inside objects, [notably in RawModule](https://github.com/chipsalliance/chisel/blob/abf74afc2a497947dc0ef5747d8de875f0f162ef/core/src/main/scala/chisel3/RawModule.scala#L137)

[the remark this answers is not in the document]
adkian-sifive: not sure if we want to change these paths but i guess it's not a big deal

core/src/main/scala/chisel3/BitsImpl.scala
```diff
@@ -3,8 +3,9 @@
 package chisel3
 
 import chisel3.experimental.{requireIsHardware, SourceInfo}
-import chisel3.internal.{_resizeToWidth, throwException, BaseModule}
+import chisel3.internal.{throwException, BaseModule}
 import chisel3.internal.Builder.pushOp
+import chisel3.internal._
```
jackkoenig, line 8:
```suggestion
```
Don't add a wildcard import unless we really need to (and this PR shouldn't need it).

core/src/main/scala/chisel3/PrintfImpl.scala
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+trait PrintfImpl {
```
jackkoenig, line 5:
```suggestion
private[chisel3] trait PrintfImpl {
```
This needs to be package private

jackkoenig (approved):
Approved but please fix 2 comments.

PR #4487, opened 2024-10-25
Add `withModulePrefix`

mmaloney-sf:
Introduces a new API `withModulePrefix`.

Example:

```scala
class Submod extends RawModule {
  // ...
}

class Top extends RawModule {
  val noprefix_inst = Module(new Submod)

  withModulePrefix("Foo") {
    val foo_inst = Module(new Submod)
  }

  withModulePrefix("Bar") {
    val bar_inst = Module(new Submod)
  }
}
```

This will result in the definition of the following modules:

```firrtl
  module Submod :
    skip

  module Foo_Submod :
    skip

  module Bar_Submod : 
    skip

  public module Top :

    inst noprefix_inst of Submod
    inst foo_inst of Foo_Submod
    inst bar_inst of Bar_Submod
```

Only module definitions are prefixed. Other circuit components, such as wires and registers, are unaffected.

Module prefixes nest, where `Foo_Bar_Submod` indicating that a prefix `Foo` exists in the grantparent or higher of the instance of `Submod`.

When using this API with`Instantiate`, two instantiations under two distinct module prefix blocks will result in two "copies" of the definition, one for each prefix. Conversely, two instantiations under the same module prefix block will share the same prefixed definition.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -926,6 +926,9 @@ package experimental {
           case None    => getRef.name
           case Some(c) => getRef.fullName(c)
         }
+
+    /** Returns the current nested module prefix */
+    val currentModulePrefix: String = Builder.getModulePrefix
   }
 }
 
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -928,7 +928,7 @@ package experimental {
         }
 
     /** Returns the current nested module prefix */
-    val currentModulePrefix: String = Builder.getModulePrefix
+    val modulePrefix: String = Builder.getModulePrefix
   }
 }
 
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -16,8 +16,10 @@ import _root_.firrtl.annotations.{
   Annotation,
   InstanceTarget,
   IsModule,
+  IsMember,
   ModuleName,
   ModuleTarget,
+  Target,
   SingleTargetAnnotation
 }
 import _root_.firrtl.AnnotationSeq
@@ -31,8 +33,6 @@ private[chisel3] trait ObjectModuleImpl {
   protected def _applyImpl[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = {
     // Instantiate the module definition.
     val module: T = evaluate[T](bc)
-
-    ModulePrefixAnnotation.annotate(module)
 
     // Handle connections at enclosing scope
     // We use _component because Modules that don't generate them may still have one
@@ -938,16 +938,16 @@ object withModulePrefix {
   }
 }
 
-private case class ModulePrefixAnnotation(target: IsModule, prefix: String) extends SingleTargetAnnotation[IsModule] {
-  def duplicate(n: IsModule): ModulePrefixAnnotation = this.copy(target = n)
+private case class ModulePrefixAnnotation(target: IsMember, prefix: String) extends SingleTargetAnnotation[IsMember] {
+  def duplicate(n: IsMember): ModulePrefixAnnotation = this.copy(target = n)
 }
 
 private object ModulePrefixAnnotation {
-  def annotate[T <: BaseModule](module: T): Unit = {
+  def annotate[T <: IsMember](target: T): Unit = {
       val prefix = Builder.getModulePrefix
       if (prefix != "") {
         val annotation: ChiselAnnotation = new ChiselAnnotation {
-          def toFirrtl: Annotation = ModulePrefixAnnotation(module.toTarget, prefix)
+          def toFirrtl: Annotation = ModulePrefixAnnotation(target, prefix)
         }
         chisel3.experimental.annotate(annotation)
       }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -943,11 +943,11 @@ private case class ModulePrefixAnnotation(target: IsMember, prefix: String) exte
 }
 
 private object ModulePrefixAnnotation {
-  def annotate[T <: IsMember](target: T): Unit = {
+  def annotate[T <: HasId](target: T): Unit = {
       val prefix = Builder.getModulePrefix
       if (prefix != "") {
         val annotation: ChiselAnnotation = new ChiselAnnotation {
-          def toFirrtl: Annotation = ModulePrefixAnnotation(target, prefix)
+          def toFirrtl: Annotation = ModulePrefixAnnotation(target.toTarget, prefix)
         }
         chisel3.experimental.annotate(annotation)
       }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -235,6 +235,10 @@ private[chisel3] trait ObjectModuleImpl {
     /** Explicitly Asynchronous Reset */
     case object Asynchronous extends Type
   }
+
+  def getModulePrefixList: List[String] = {
+    Builder.getModulePrefixList
+  }
 }
 
 private[chisel3] trait ModuleImpl extends RawModule with ImplicitClock with ImplicitReset {
@@ -920,7 +924,6 @@ package experimental {
           case None    => getRef.name
           case Some(c) => getRef.fullName(c)
         }
-
   }
 }
 
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -30,14 +30,9 @@ private[chisel3] trait ObjectModuleImpl {
 
   protected def _applyImpl[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = {
     // Instantiate the module definition.
-    val module = evaluate[T](bc)
+    val module: T = evaluate[T](bc)
 
-    val prefix = Builder.getModulePrefix
-    if (prefix != "") {
-      annotate(new ChiselAnnotation {
-        def toFirrtl: Annotation = ModulePrefixAnnotation(module.toTarget, prefix)
-      })
-    }
+    ModulePrefixAnnotation.annotate(module)
 
     // Handle connections at enclosing scope
     // We use _component because Modules that don't generate them may still have one
@@ -943,6 +938,18 @@ object withModulePrefix {
   }
 }
 
-case class ModulePrefixAnnotation(target: IsModule, prefix: String) extends SingleTargetAnnotation[IsModule] {
+private case class ModulePrefixAnnotation(target: IsModule, prefix: String) extends SingleTargetAnnotation[IsModule] {
   def duplicate(n: IsModule): ModulePrefixAnnotation = this.copy(target = n)
+}
+
+private object ModulePrefixAnnotation {
+  def annotate[T <: BaseModule](module: T): Unit = {
+      val prefix = Builder.getModulePrefix
+      if (prefix != "") {
+        val annotation: ChiselAnnotation = new ChiselAnnotation {
+          def toFirrtl: Annotation = ModulePrefixAnnotation(module.toTarget, prefix)
+        }
+        chisel3.experimental.annotate(annotation)
+      }
+  }
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -164,6 +164,9 @@ private[chisel3] trait ObjectModuleImpl {
   /** Returns the current Module */
   def currentModule: Option[BaseModule] = Builder.currentModule
 
+  /** Returns the current nested module prefix */
+  def currentModulePrefix: String = Builder.getModulePrefix
+
   private[chisel3] def do_pseudo_apply[T <: BaseModule](
     bc: => T
   )(
@@ -671,8 +674,9 @@ package experimental {
         // PseudoModules are not "true modules" and thus should share
         // their original modules names without uniquification
         this match {
-          case _: PseudoModule => desiredName
-          case _ => Builder.globalNamespace.name(desiredName)
+          case _: PseudoModule => Module.currentModulePrefix + desiredName
+          case _: BlackBox     => Builder.globalNamespace.name(desiredName)
+          case _               => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
         }
       } catch {
         case e: NullPointerException =>
@@ -909,5 +913,14 @@ package experimental {
           case Some(c) => getRef.fullName(c)
         }
 
+  }
+}
+
+object withModulePrefix {
+  def apply[T](prefix: String)(block: => T): T = {
+    Builder.pushModulePrefix(prefix)
+    val res = block // execute block
+    Builder.popModulePrefix() 
+    res
   }
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -12,10 +12,11 @@ import chisel3.internal.firrtl.ir._
 import chisel3.experimental.{requireIsChiselType, BaseModule, SourceInfo, UnlocatableSourceInfo}
 import chisel3.properties.{Class, Property}
 import chisel3.reflect.DataMirror
-import _root_.firrtl.annotations.{InstanceTarget, IsModule, ModuleName, ModuleTarget}
+import _root_.firrtl.annotations.{InstanceTarget, IsModule, ModuleName, ModuleTarget, SingleTargetAnnotation, Annotation}
 import _root_.firrtl.AnnotationSeq
 import chisel3.internal.plugin.autoNameRecursively
 import chisel3.util.simpleClassName
+import chisel3.experimental.{annotate, ChiselAnnotation}
 import chisel3.experimental.hierarchy.Hierarchy
 
 private[chisel3] trait ObjectModuleImpl {
@@ -23,6 +24,13 @@ private[chisel3] trait ObjectModuleImpl {
   protected def _applyImpl[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = {
     // Instantiate the module definition.
     val module = evaluate[T](bc)
+
+    val prefix = Builder.getModulePrefix
+    if (prefix != "") {
+      annotate(new ChiselAnnotation {
+        def toFirrtl: Annotation = ModulePrefixAnnotation(module.toTarget, prefix)
+      })
+    }
 
     // Handle connections at enclosing scope
     // We use _component because Modules that don't generate them may still have one
@@ -923,4 +931,8 @@ object withModulePrefix {
     Builder.popModulePrefix() 
     res
   }
+}
+
+case class ModulePrefixAnnotation(target: IsModule, prefix: String) extends SingleTargetAnnotation[IsModule] {
+  def duplicate(n: IsModule): ModulePrefixAnnotation = this.copy(target = n)
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -12,7 +12,14 @@ import chisel3.internal.firrtl.ir._
 import chisel3.experimental.{requireIsChiselType, BaseModule, SourceInfo, UnlocatableSourceInfo}
 import chisel3.properties.{Class, Property}
 import chisel3.reflect.DataMirror
-import _root_.firrtl.annotations.{InstanceTarget, IsModule, ModuleName, ModuleTarget, SingleTargetAnnotation, Annotation}
+import _root_.firrtl.annotations.{
+  Annotation,
+  InstanceTarget,
+  IsModule,
+  ModuleName,
+  ModuleTarget,
+  SingleTargetAnnotation
+}
 import _root_.firrtl.AnnotationSeq
 import chisel3.internal.plugin.autoNameRecursively
 import chisel3.util.simpleClassName
@@ -688,7 +695,7 @@ package experimental {
         this match {
           case _: PseudoModule => Module.currentModulePrefix + desiredName
           case _: BlackBox     => Builder.globalNamespace.name(desiredName)
-          case _               => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
+          case _ => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
         }
       } catch {
         case e: NullPointerException =>
@@ -931,7 +938,7 @@ object withModulePrefix {
   def apply[T](prefix: String)(block: => T): T = {
     Builder.pushModulePrefix(prefix)
     val res = block // execute block
-    Builder.popModulePrefix() 
+    Builder.popModulePrefix()
     res
   }
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -931,9 +931,13 @@ package experimental {
 
 object withModulePrefix {
   def apply[T](prefix: String)(block: => T): T = {
-    Builder.pushModulePrefix(prefix)
+    if (prefix != "") {
+      Builder.pushModulePrefix(prefix)
+    }
     val res = block // execute block
-    Builder.popModulePrefix()
+    if (prefix != "") {
+      Builder.popModulePrefix()
+    }
     res
   }
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -929,7 +929,13 @@ package experimental {
   }
 }
 
+/**
+ * Creates a block under which any generator that gets run results in a module whose name is prepended with the given prefix.
+ */
 object withModulePrefix {
+  /**
+    * @param arg prefix Prefix is the module prefix, blank means ignore.
+    */
   def apply[T](prefix: String)(block: => T): T = {
     if (prefix != "") {
       Builder.pushModulePrefix(prefix)
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -15,12 +15,12 @@ import chisel3.reflect.DataMirror
 import _root_.firrtl.annotations.{
   Annotation,
   InstanceTarget,
-  IsModule,
   IsMember,
+  IsModule,
   ModuleName,
   ModuleTarget,
-  Target,
-  SingleTargetAnnotation
+  SingleTargetAnnotation,
+  Target
 }
 import _root_.firrtl.AnnotationSeq
 import chisel3.internal.plugin.autoNameRecursively
@@ -948,12 +948,12 @@ private case class ModulePrefixAnnotation(target: IsMember, prefix: String) exte
 
 private object ModulePrefixAnnotation {
   def annotate[T <: HasId](target: T): Unit = {
-      val prefix = Builder.getModulePrefix
-      if (prefix != "") {
-        val annotation: ChiselAnnotation = new ChiselAnnotation {
-          def toFirrtl: Annotation = ModulePrefixAnnotation(target.toTarget, prefix)
-        }
-        chisel3.experimental.annotate(annotation)
+    val prefix = Builder.getModulePrefix
+    if (prefix != "") {
+      val annotation: ChiselAnnotation = new ChiselAnnotation {
+        def toFirrtl: Annotation = ModulePrefixAnnotation(target.toTarget, prefix)
       }
+      chisel3.experimental.annotate(annotation)
+    }
   }
 }
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -694,7 +694,7 @@ package experimental {
         // their original modules names without uniquification
         this match {
           case _: PseudoModule => Module.currentModulePrefix + desiredName
-          case _: BlackBox     => Builder.globalNamespace.name(desiredName)
+          case _: BaseBlackBox     => Builder.globalNamespace.name(desiredName)
           case _ => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
         }
       } catch {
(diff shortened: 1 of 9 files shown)


src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+// This has to be defined at the top-level because @instantiable doesn't work when nested.
+@instantiable
+class AddOne(width: Int) extends Module {
+  @public val in = IO(Input(UInt(width.W)))
+  @public val out = IO(Output(UInt(width.W)))
+  out := in + 1.U
+}
```
jackkoenig, line 167: It's not that it has to be at the top-level, it just cannot be inside of a class (or thunk). Stylistically, please put this inside of a companion object for `PrefixSpec` at the top.

You could alternatively use one of the existing examples in `chiselTests.experimental.hierarchy.Examples` (like `AddOneParameterized`), although I would probably put this above in the companion object.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1160,6 +1162,36 @@ private[chisel3] object Builder extends LazyLogging {
...
+    tail
+  }
+
+  // Returns the nested module prefix at this moment
+  def getModulePrefix: String = {
+    val modulePrefixStack = chiselContext.get().modulePrefixStack
+    if (modulePrefixStack.isEmpty) {
+      ""
+    } else {
+      modulePrefixStack.foldLeft("")((a, b) => b + "_" + a)
```
jackkoenig, line 1186: I'm of two minds on automatically including `_` between prefixes. On the one hand, that is almost certainly the most common use, but on the other hand, what if a user doesn't want `_` between prefixes?

I'm thinking that instead of adding them here, maybe we overload `withModulePrefix`, something like:
```scala
object withModulePrefix {
  def apply[T](prefix: String, trailingUnderscore: Boolean)(block: => T): T = ??? 
  
  def apply[T](prefix: String)(block: => T): T = apply(prefix, true)(block)
```
So that the default behavior adds the trailing `_` but the user can opt-out? WDYT?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1160,6 +1162,36 @@ private[chisel3] object Builder extends LazyLogging {
...
+    }
+  }
+
+  def getModulePrefixList: List[String] = {
+    val modulePrefixStack = chiselContext.get().modulePrefixStack
+    modulePrefixStack.toList
```
jackkoenig, line 1192: Is the `.toList` cast really needed?

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+      val np_inst = Instantiate(new AddOne(width))
+
+      foo_inst.in := in
+      bar_inst.in := foo_inst.out
+      out := bar_inst.out
+      np_inst.in := in
+    }
+
+    val chirrtl = emitCHIRRTL(new Top)
+    // println(chirrtl)
```
jackkoenig, line 114: dead code.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+      module Outer_Foo
+        inst bar of Outer_Inner_Bar
+      module Top :
+        inst foo of Outer_Foo
+        """.linesIterator.map(_.trim).toSeq
+
+    matchesAndOmits(chirrtl)(lines: _*)()
+  }
+
+  it should "Instnantiate should create distinct module definitions when instantiated with distinct prefixes" in {
```
jackkoenig, line 90:
```suggestion
  it should "Instantiate should create distinct module definitions when instantiated with distinct prefixes" in {
```

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+      }
+
+      module Foo :
+        wire a : UInt<1>
+      module Pref_Foo :
+        wire a : UInt<1>
+      module Top :
+        inst foo of Foo
+        inst pref_foo of Pref_Foo
+        """.linesIterator.map(_.trim).toSeq
```
jackkoenig, line 42:
```suggestion
     |"class":"chisel3.ModulePrefixAnnotation",
     |"target":"~Top|Pref_Foo",
     |"prefix":"Pref_"
     |module Foo :
     |module Pref_Foo :
     |module Top :
     |  inst foo of Foo
     |  inst pref_foo of Pref_Foo
     |""".stripMargin.lines
```
This is a very minor nit, but you could make this a little more concise (and make it more obvious what we're actually trying to check), no need to check for the `wire a` nor really the `{` or `}` of the annotation. Also `.stripMargin` probably would obviate the need for the `trim`. That being said, I didn't know about `linesIterator` (or `lines`), good find!

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+          "class":"chisel3.ModulePrefixAnnotation",
+          "target":"~Top|Outer_Inner_Bar",
+          "prefix":"Outer_Inner_"
+        },
+        {
+          "class":"chisel3.ModulePrefixAnnotation",
+          "target":"~Top|Outer_Foo",
+          "prefix":"Outer_"
+        }
+      ]]
```
jackkoenig, line 78: Same comment as above, I'm not sure that the specific syntax of `circuit Top :%[[` needs to be checked, etc.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+      public module Top :
+        inst np_inst of AddOne
+        inst foo_inst of Foo_AddOne
+        inst bar_inst of Bar_AddOne
+        """.linesIterator.map(_.trim).toSeq
+
+    matchesAndOmits(chirrtl)(lines: _*)("AddOne_1")
+  }
+
+  it should "Instnantiate should reference the same module definitions when instantiated with the same prefix" in {
```
jackkoenig, line 129:
```suggestion
  it should "Instantiate should reference the same module definitions when instantiated with the same prefix" in {
```

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+
+package chiselTests
+
+import chisel3._
+import chisel3.experimental.hierarchy.{instantiable, public, Definition, Instance, Instantiate}
+import circt.stage.ChiselStage.emitCHIRRTL
+
+class PrefixSpec extends ChiselFlatSpec with ChiselRunners with Utils with MatchesAndOmits {
+
+  behavior.of("withPrefix")
```
jackkoenig, line 11:
```suggestion
  behavior.of("withModulePrefix")
```

src/test/scala/chiselTests/PrefixSpec.scala
jackkoenig: Also tiny nit but the grammar here suggests you should be doing a `behavior of "Instantiate"` then just `it should "create ...`.

jackkoenig (commented):
Generally looks good, 1 question about `_`.

Also we need to define and test the behavior for `Definition` and `Instance`. `Instance` should have no interaction with the prefix but should `Definition`? My initial reaction is "no" since `Definition` is essentially a separate Chisel elaboration, but I'm unsure. Any thoughts @mwachs5, @azidar, @seldridge?

We also probably need a way to remove the prefix, e.g. `withNoModulePrefix` or `withOverrideModulePrefix("")` or something.

src/test/scala/chiselTests/PrefixSpec.scala
mmaloney-sf: Removed.

core/src/main/scala/chisel3/internal/Builder.scala
mmaloney-sf: Removed.

core/src/main/scala/chisel3/internal/Builder.scala
mmaloney-sf: It strikes me as awkward to mandate the `_` as part of the prefix. You're going to get someone who forgets it, and it's gonna make things confusing and ugly.

I wouldn't be against a global-level configuration for the separator, with `"_"` the default value for it.

src/test/scala/chiselTests/PrefixSpec.scala
mmaloney-sf: Done.

core/src/main/scala/chisel3/internal/Builder.scala
mwachs5: I don't think we need this. Users can add _ if they want, right?

mwachs5:
Well, for definitions I think you do want the module prefix baked into it somehow. Is it a Val def = withModulePrefix (Definition(new Foo)) or perhaps it's better something one adds as an argument the Definition/Module/Instantiate APIs directly?

mwachs5:
Actually now that I wrote the above comment I really think it's a more natural thing to do than withModulePrefix(...) even for Module. For the latter why would we be doing it for more than one child ever, isn't it always started at one module or definition instantiation?

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -224,6 +242,10 @@ private[chisel3] trait ObjectModuleImpl {
     /** Explicitly Asynchronous Reset */
     case object Asynchronous extends Type
   }
+
+  def getModulePrefixList: List[String] = {
+    Builder.getModulePrefixList
+  }
```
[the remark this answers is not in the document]
jackkoenig: Makes sense!

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: I did an audit of our internal code and AFAICT every single place except one[^1] the added prefix ends with `_`. I think it's reasonable to be opinionated that separators are a good thing and to default to `_`. If it comes up that some users either want a different separator, or none at all--then we can make it possible to customize the separator when invoking Chisel, and potentially make it possible to customize at specific call sites.

@mmaloney-sf one thing that having separators by default does imply though is that we need to specially handle empty Strings as not adding any prefix (i.e. don't add a separator in that case).

[^1]: And that one place it might be a mistake to not have the separator.

jackkoenig:
> Actually now that I wrote the above comment I really think it's a more natural thing to do than withModulePrefix(...) even for Module. For the latter why would we be doing it for more than one child ever, isn't it always started at one module or definition instantiation?

I agree that we should instead make it an argument to `Module`, `Definition`, and `Instantiate`, but I just want to note that it's less about more than one child ever, and more about adding 1 new API vs. modifying `n` (where `n` appears to be 3, but any other added APIs will also potentially need to add prefix).

So let's proceed with adding prefix as an optional (well optional via overloading) argument to `Module`, `Definition`, and `Instantiate`, but I want to call out some potential future work.
1. How do we deal with instantiating modules from different prefix "namespaces"? For example, say you have a library of Queues and you want to instantiate from the library rather than forcing the current prefix. We could add another overloaded form of `Module`, `Definition`, and `Instantiate` to "set" the prefix rather than push to it, or we could add something similar to `withModulePrefix` but just for that use case.
2. How do we deal with prefix for the top module? Should there be a way from within a Module class to set the prefix for yourself (and all children) or should it just be done on the command-line?

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -908,6 +931,18 @@ package experimental {
           case None    => getRef.name
           case Some(c) => getRef.fullName(c)
         }
+  }
+}
 
+object withModulePrefix {
```
mwachs5, line 937: Is this the public entry point? Scaladoc?

should this be called `withNestedModulePrefix`? Is there a `stripPrefix`?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1160,6 +1163,37 @@ private[chisel3] object Builder extends LazyLogging {
...
+
+  // Returns the nested module prefix at this moment
+  def getModulePrefix: String = {
+    val ctx = chiselContext.get()
+    val modulePrefixStack = ctx.modulePrefixStack
+    val sep = ctx.sep
+    if (modulePrefixStack.isEmpty) {
+      ""
+    } else {
+      modulePrefixStack.foldLeft("")((a, b) => b + sep + a)
```
mwachs5, line 1189:
```suggestion
      modulePrefixStack.mkString(sep)
```

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    class Top extends RawModule {
+      val foo = Module(new Foo)
+
+      withModulePrefix("Pref") {
+        val pref_foo = Module(new Foo)
+      }
+    }
+
```
mwachs5, line 25: can you please show a test that does two nested prefix blocks (make sure it's Foo_Bar not Bar_Foo and so on!)

core/src/main/scala/chisel3/internal/Builder.scala
jackkoenig: I think you would need to `.reverse` for this to work.

jackkoenig:
> > So let's proceed with adding prefix as an optional (well optional via overloading) argument to Module, Definition, and Instantiate, but I want to call out some potential future work.
> 
> > 1. How do we deal with instantiating modules from different prefix "namespaces"? For example, say you have a library of Queues and you want to instantiate from the library rather than forcing the current prefix. We could add another overloaded form of Module, Definition, and Instantiate to "set" the prefix rather than push to it, or we could add something similar to withModulePrefix but just for that use case.
> 
> Hm, very good point. I guess it's almost like you are forced to opt-in to making your component a valid 'top level' component that gets to control the prefixing.

Right, how should we do this? Maybe the answer depends on how we want to do linking.

> > 2. How do we deal with prefix for the top module? Should there be a way from within a Module class to set the prefix for yourself (and all children) or should it just be done on the command-line?
> 
> This makes me wonder, what is the behavior of doing something like:
> 
> ```
> val q = Module(withModulePrefix(new Queue))
> ```
> 
> Do you get an error? Is it the same as `withModulePrefix(Module(new Queue))`? Probably if not, you should get an error... and if so, isn't that the answer to the question 2?

I don't think it's the answer to question (2) though since we're moving to making the prefix an argument to `Module`, `Definition`, and `Instance` unless you're opinion on this has changed lol.

mwachs5:
>  don't think it's the answer to question (2) though since we're moving to making the prefix an argument to Module, Definition, and Instance unless you're opinion on this has changed lol.

Yes it has changed because i think it's not so uncommon to call 
```
withModulePrefix { HelperObjectThatActuallyCallsModule }
```

so if we need withModulePrefix anyway not sure we should expose it on the others

src/test/scala/chiselTests/PrefixSpec.scala
mmaloney-sf: This is what `it should "Allow nested module prefixes"` does.

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -908,6 +931,18 @@ package experimental {
...
+    Builder.pushModulePrefix(prefix)
+    val res = block // execute block
+    Builder.popModulePrefix()
+    res
   }
 }
+
+case class ModulePrefixAnnotation(target: IsModule, prefix: String) extends SingleTargetAnnotation[IsModule] {
+  def duplicate(n: IsModule): ModulePrefixAnnotation = this.copy(target = n)
+}
```
seldridge, line 948: Can this be made private? There's no need to make this publicly accessible. It also lets us delete it in the future.

docs/src/explanations/moduleprefix.md
```diff
@@ -0,0 +1,117 @@
...
+
+All modules defined inside of this block, whether an immediate submodule or a descendent, will be given a prefix `Foo`.
+(The prefix is separated by an underscore `_`).
+
+For example, suppose we write the following:
+
+```scala mdoc:silent
+import chisel3._
+
+class Top extends Module {
```
jackkoenig, line 28:
```suggestion
```scala mdoc:silent:reset
import chisel3._

class Top extends Module {
```
Mdoc is failing because you define top multiple times, I think you just need to reset each block.

core/src/main/scala/chisel3/experimental/hierarchy/InstantiateImpl.scala
```diff
@@ -88,9 +88,10 @@ private[chisel3] trait InstantiateImpl {
     args: K,
     f:    K => A
   ): Definition[A] = {
+    val modulePrefix = Builder.getModulePrefixList
     Builder.contextCache
       .getOrElseUpdate(
-        CacheKey(boxAllData(args), implicitly[ru.WeakTypeTag[A]]), {
+        CacheKey(boxAllData(args), implicitly[ru.WeakTypeTag[A]], modulePrefix), {
```
mwachs5, line 94: we should think about -- does/should modulePrefix need to be part of the cache key for an ExtModule? It's hard to know if the ExtModule is going to override its desiredName with Module.prefix or not.

But maybe it doesn't really matter because its not like an extmodule in firrtl has much cost to it

mwachs5:
Can we add test cases for Instantiate(SomeBlackbox) where SomeBlackbox may or may not use Module.prefix to decide its name? Not sure what the behavior *should* be but the tests should make it clear what that is

core/src/main/scala/chisel3/experimental/hierarchy/InstantiateImpl.scala
jackkoenig: This is a great point and question, and unfortunately the decision on whether or not the modulePrefix has to be made before instantiating the module (we can't like, query the user-overridden `desiredName`).

jackkoenig:
> Can we add test cases for Instantiate(SomeBlackbox) where SomeBlackbox may or may not use Module.prefix to decide its name? Not sure what the behavior _should_ be but the tests should make it clear what that is

We definitely need tests. What I think the behavior should be is that BlackBoxes (and ExtModules) by default ignore the prefix, but there is a trait you can mix in to make the module sensitive to prefix (which means `desiredName` defaults to include it and `Instantiate` includes it in the cache key).

This can be implemented pretty simply by defining the trait (not sure what to call it--`Prefixable`? `PrefixSensitive`? `CanHavePrefix`?), then match on that trait in the implementation of `BaseModule.name` and similarly using the type tag in `Instantiate`. I will leave specific suggestions in the code.

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -671,8 +688,9 @@ package experimental {
         // PseudoModules are not "true modules" and thus should share
         // their original modules names without uniquification
         this match {
-          case _: PseudoModule => desiredName
-          case _ => Builder.globalNamespace.name(desiredName)
+          case _: PseudoModule => Module.currentModulePrefix + desiredName
+          case _: BaseBlackBox => Builder.globalNamespace.name(desiredName)
+          case _ => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
```
jackkoenig, line 693:
```suggestion
          case _: PseudoModule => desiredName
          case _: Prefixable => Builder.globalNamespace.name(Module.currentModulePrefix + desiredName)
          case _ => Builder.globalNamespace.name(desiredName)
```
I don't care what we call it, but using `Prefixable` as a strawman, have `RawModule` mix the trait in, then users can decide to opt-in their BlackBoxes or go with the default behavior.

Also note I removed the prefix from `Pseudo-module`, I'm pretty sure they shouldn't be getting the prefix (but maybe I'm wrong).

core/src/main/scala/chisel3/experimental/hierarchy/InstantiateImpl.scala
jackkoenig: Because of how the diff is, I can't do this as a "suggestion", but I think lines 91-94 should basically be:
```scala
    val ttag = implicitly[ru.WeakTypeTag[A]]
    val modulePrefix = if (ttag.tpe <:< typeOf[Prefixable]) Builder.getModulePrefixList else Nil
    Builder.contextCache
      .getOrElseUpdate(
        CacheKey(boxAllData(args), ttag, modulePrefix), {
```
Again, with `Prefixable` as a strawman proposal for the opt-in trait, I'm open to better names.

mwachs5:
> we need to define and test the behavior for Definition and Instance. Instance should have no interaction with the prefix but should Definition? My initial reaction is "no" since Definition is essentially a separate Chisel elaboration, but I'm unsure. Any thoughts @mwachs5, @azidar, @seldridge?

Just to circle back on this -- Definition is *not* a separate chisel elaboration, and i think it should include the prefix for any Modules build through a definition. Extmodules can decide to use the prefix or not, Instantiate will have to assume that they do

jackkoenig:
> Just to circle back on this -- Definition is not a separate chisel elaboration, and i think it should include the prefix for any Modules build through a definition. Extmodules can decide to use the prefix or not, Instantiate will have to assume that they do

Agreed on all counts. I proposed `Prefixable` above for letting ExtModules opt-in, but that can be in a follow on PR.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+    val chirrtl = emitCHIRRTL(new Top)
+    println(chirrtl)
+
+    val lines = """
+  module Foo_AddOne
+  module Top
+        """.linesIterator.map(_.trim).toSeq
+
+    matchesAndOmits(chirrtl)(lines: _*)()
+  }
```
jackkoenig, line 200: Good test, would also like to see a test for something like this:

```scala
    class Child(defn: Definition[AddOne]) {
      val addone = Instance(defn)
    }
    class Top extends Module {
      val dfn =  Definition(new AddOne(8))

      val child = withModulePrefix("Foo") {
        Module(new Child(defn))
      }
    }
```
Show that you have `inst addone of AddOne` in `module Foo_Child`.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  it should "withModulePrefix does not affect ExtModules" in {
```
mwachs5, line 202:
```suggestion
  it should "withModulePrefix does not automatically affect ExtModules" in {
```

mwachs5: should add another test that user can override desiredName with Module.prefix or whatever the user API is and then it *can* affect the ExtModule

core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -224,6 +237,10 @@ private[chisel3] trait ObjectModuleImpl {
     /** Explicitly Asynchronous Reset */
     case object Asynchronous extends Type
   }
+
+  def getModulePrefixList: List[String] = {
```
jackkoenig, line 241: Could use ScalaDoc

core/src/main/scala/chisel3/ModuleImpl.scala
jackkoenig: Discussed offline that we are leaving this proposal to a future PR.

core/src/main/scala/chisel3/ModuleImpl.scala
jackkoenig: Agreed that this should have ScalaDoc, I am happy with the API (no need to specify that it's `nested`, that's just the API, and we can add `stripPrefix` later if needed).

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -461,6 +461,7 @@ private[chisel3] trait NamedComponent extends HasId {
 
 // Mutable global state for chisel that can appear outside a Builder context
 private[chisel3] class ChiselContext() {
+  val sep: String = "_"
```
jackkoenig, line 464: This could use a slightly more descriptive name, e.g. `modulePrefixSeparator`

docs/src/explanations/moduleprefix.md
```diff
@@ -0,0 +1,122 @@
...
+  val noprefix_sub = Instantiate(new Sub)
+}
+```
+
+In this example, we end up with four modules: `Top`, `Foo_Sub`, `Bar_Sub`, and `Sub`.
+
+When using `Definition` and `Instance`, all `Definition` calls will be affected by `withModulePrefix`.
+However, `Instance` will not be effected, since it always creates an instance of the captured definition.
+
+`ExtModule` is unaffected by `withModulePrefix`.
```
jackkoenig, line 122:
```suggestion
`BlackBox` and `ExtModule` are unaffected by `withModulePrefix`.
```
Most people use `BlackBox` so we should mention both.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+    class Sub extends ExtModule {
+    }
+
+    class Top extends Module {
+      val sub_foo = withModulePrefix("Foo") { Module(new Sub) }
+    }
+
+    val chirrtl = emitCHIRRTL(new Top)
+    //println(chirrtl)
```
jackkoenig, line 220: This is [obviously] missing the actual test so make sure to add it and delete any commented out code.

core/src/main/scala/chisel3/experimental/hierarchy/InstantiateImpl.scala
jackkoenig: I retract my suggestion above--we should always make the prefix be part of the `CacheKey` even if it leads to some unnecessary duplication of definitions for ExtModules/BlackBoxes. The reasoning is that even if we added an opt-in trait like my proposed `Prefixable`--if a user didn't use `Prefixable` but override's `desiredName` to use the prefix, we have no way to detect that they did that, but if we didn't use the module prefix as part of the key, the `Instantiable` would do the wrong thing because it would hit in the cache for the same ExtModule despite the prefix being different. Thus we have to defensively assume users could always override desiredName to incorporate the prefix, so we must always factor it in here.

src/test/scala/chiselTests/PrefixSpec.scala
jackkoenig: I agree we need this test because I think it will highlight a slight problem--`Module.currentModulePrefix` does not (I think) include a trailing separator, so how can you use it to properly generate `desiredName`?

We need tests for `Module.currentModulePrefix` and the `List` form too.

core/src/main/scala/chisel3/ModuleImpl.scala
mmaloney-sf: Whoops. Those got lost in a stash.
Re-added.

core/src/main/scala/chisel3/internal/Builder.scala
mmaloney-sf: Done.

docs/src/explanations/moduleprefix.md
mmaloney-sf: Done.

src/test/scala/chiselTests/PrefixSpec.scala
mmaloney-sf: Done.

docs/src/explanations/moduleprefix.md
```diff
@@ -0,0 +1,122 @@
+# Module Prefixing
```
mwachs5, line 1: can you add this doc to the explanations TOC (unless that is auto-generated now)? Note that is in 'order in which you should read these'

mmaloney-sf: I put it at the end, since it's a rather niche feature.

docs/src/explanations/moduleprefix.md
```diff
@@ -0,0 +1,122 @@
...
+  val noprefix_sub = Instantiate(new Sub)
+}
+```
+
+In this example, we end up with four modules: `Top`, `Foo_Sub`, `Bar_Sub`, and `Sub`.
+
+When using `Definition` and `Instance`, all `Definition` calls will be affected by `withModulePrefix`.
+However, `Instance` will not be effected, since it always creates an instance of the captured definition.
+
+`BlackBox` and `ExtModule` are unaffected by `withModulePrefix`.
```
mwachs5, line 122: this is awesome! Thanks so much for the great write up. Do you have a user API to get at the current prefix (something like `Module.prefix`?) might be nice to add to the explanation doc.

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,278 @@
...
+    val lines = """
+      extmodule Sub
+        defname = Sub
+      module Top
+        """.linesIterator.map(_.trim).toSeq
+
+    matchesAndOmits(chirrtl)(lines: _*)()
+  }
+
+  it should "Using currentModulePrefix to force the name of an extmodule" in {
```
mwachs5, line 259:
```suggestion
  it should "Using modulePrefix to force the name of an extmodule" in {
```

docs/src/explanations/moduleprefix.md
mwachs5: Lookslike this is `modulePrefix`

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    class Top extends Module {
+      val sub_foo = withModulePrefix("Foo") { Module(new Sub) }
+    }
+
+    val chirrtl = emitCHIRRTL(new Top)
+
+    val lines = """
+      extmodule Foo_Sub
```
mwachs5, line 271: cool! Thanks for adding the test. Not what i was expecting!

mwachs5 (approved):
approving with minor comments! Thanks for adding this great feature

src/test/scala/chiselTests/PrefixSpec.scala
```diff
@@ -0,0 +1,278 @@
...
+        defname = Foo_Sub
+      module Top
+        """.linesIterator.map(_.trim).toSeq
```
jackkoenig, line 274:
```suggestion
    val lines =
      """|extmodule Foo_Sub
         | defname = Foo_Sub
         |module Top
         |""".stripMargin.linesIterator.toSeq
```
This is a tiny nit, but the block comments in this file are somewhat inconsistently spaced. If you use `|` (and corresponding `.stripMargin`), then scalafmt will line them up like this for you (make sure to start the `"""|` on the following line, it lines up the `|` with each other). Then you also don't need the `map(_.trim)` then.

jackkoenig (approved):
LGTM, consider fixing nit

docs/src/explanations/moduleprefix.md
mmaloney-sf: Added.

PR #4492, opened 2024-11-01
[Scala3] Add cross build support to Mill

adkian-sifive:
The way the compilation units in the Mill build files were set up meant that differentiated module dependencies and compilation options between Scala versions could not be supported. Specifically, having these options in traits like `HasChiselPlugin` meant that they could not be optionally included in the build unit based on the cross version. 

However, this is a requirement for adding Scala3 cross compilation support, since there are compilation options (`-Ymacro-annotations`) and module dependencies (`pluginModule`) that are Scala2-specific. This PR includes a refactor of the build files and adds initial support for adding module dependencies and compilation options specific to Scala versions.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Mill build files were refactored to support differentiated module dependencies and compilation options between Scala versions

- [ ] Did you review?


build.sc
```diff
@@ -48,110 +50,23 @@ object v {
...
-    "msg=undefined in comment for method cf in class PrintableHelper:s",
-    // This is deprecated for external users but not internal use
-    "cat=deprecation&origin=firrtl\\.options\\.internal\\.WriteableCircuitAnnotation:s",
-    "cat=deprecation&origin=chisel3\\.util\\.experimental\\.BoringUtils.*:s",
-    "cat=deprecation&origin=chisel3\\.experimental\\.IntrinsicModule:s",
-    "cat=deprecation&origin=chisel3\\.ltl.*:s"
-  )
-
-  // ScalacOptions
-  val commonOptions = Seq(
```
adkian-sifive, line 79: todo: readd these for scala2

build.sc
```diff
@@ -87,71 +82,20 @@ object v {
     "-Xlint:infer-any",
     "-Xlint:missing-interpolator",
     "-language:reflectiveCalls",
-    s"-Wconf:${warnConf.mkString(",")}"
+    s"-Wconf:${scala2WarnConf.mkString(",")}"
   )
 }
```
jackkoenig, line 88: Are there some scala3CommonOptions we should be using?

build.sc
```diff
@@ -87,71 +82,20 @@ object v {
...
-          T.ctx().log.info(s"Download Successfully")
-        }
-        os.proc("tar", "xvf", tarPath, "--strip-components=1").call(T.dest)
-        T.dest
-    }
+  def isScala3(ver: String): Boolean = ver match {
+    case "3.4.2" => true
+    case "2.13.15" => false
+    case _ => false // todo throw error here?
   }
```
jackkoenig, line 99:
```suggestion
  def isScala3(ver: String): Boolean = ver.startsWith("3.")
```

From looking at other examples (e.g. scala-cli's build.sc), it's more common to just do `crossScalaVersion.startsWith("3.")`

I do think it's important that you use `crossScalaVersion` instead of `crossValue` because the former is guaranteed to always be a Scala version while the latter can be whatever we want it to be (it just happens that we're using full Scala versions right now).

build.sc
```diff
@@ -182,80 +126,131 @@ trait ChiselPublishModule extends CiReleaseModule {
...
+      case false => v.scala2CommonOptions ++ Seq(
+        "-language:reflectiveCalls",
+        "-language:existentials",
+        "-language:implicitConversions",
+        "-Yrangepos", // required by SemanticDB compiler plugin
+        "-Xsource:3",
+        "-Xsource-features:infer-override"
+      )
+      case true => Seq.empty[String]
+    }
```
jackkoenig, line 170: This is definitely a nit, but it's a bit weird to match on Boolean values vs. just good 'ole if-else (also note use of crossScalaValue) which is a safer option to use.
```suggestion
    if (utils.isScala3(crossScalaValue)) {
      Seq.empty[String]
    } else {
      v.scala2CommonOptions ++ Seq(
        "-language:reflectiveCalls",
        "-language:existentials",
        "-language:implicitConversions",
        "-Yrangepos", // required by SemanticDB compiler plugin
        "-Xsource:3",
        "-Xsource-features:infer-override"
      )
    }
```

build.sc
```diff
@@ -182,80 +126,131 @@ trait ChiselPublishModule extends CiReleaseModule {
...
-  )
+  override def scalacOptions = T {
+    utils.isScala3(crossValue) match {
+      case false => v.scala2CommonOptions ++ Seq(
+        "-Xsource:3",
+        "-Xsource-features:case-apply-copy-access"
+      )
+      case true => Seq.empty[String]
+    }
+  }
```
jackkoenig, line 201: Considering we're writing some variant of this for every single Module, I think we could simplify the typical use a bit with a common trait:
```suggestion
// Up above somewhere
trait ChiselCrossSbtModule extends CrossSbtModule {
  def isScala3: T[Boolean] = T { utils.isScala3(crossScalaValue) }
  def scala2Options: T[Seq[String]] = T { Seq.empty }
  def scala3Options: T[Seq[String]] = T { Seq.empty }
  
  override def scalacOptions = T {
    super.scalacOptions ++ (if (isScala3()) scala3Options() else scala2Options())
  }
}

  // Now using this in something like Svsim
  override def scala2Options = T {
    v.scala2CommonOptions ++ Seq(
      "-Xsource:3",
      "-Xsource-features:case-apply-copy-access"
    )
  }
  override def scala3Options = T { Seq.empty[String] }
```
One subtle useful thing here is it makes sure we include `super.scalacOptions` and makes it less likely that we accidentally forget to do that. Most of the time it doesn't matter, but it does for things using Macro2Annotations and the compiler plugin.

What do you think?

jackkoenig (commented):
Generally looks good, see comments.

One comment that is confusing me

> However, this is a requirement for adding Scala3 cross compilation support, since there are compilation options (-Ymacro-annotations) and module dependencies (pluginModule) that are Scala2-specific.

I guess the plugin module is Scala 2 specific for now, but only because we haven't implemented it for Scala 3--we do need the plugin for Scala 3. It's fine to treat it this way for now but we're gonna need it soon.

build.sc
adkian-sifive: yeah true, will change this

build.sc
adkian-sifive: I don't think I've needed any yet, but if so I can add them in with the scala3 pr

build.sc
adkian-sifive: i agree from a software perspective but i do think our future selves (and other contributors) will thank us if we optimized these build files for simple+ugly over efficient+less readable..

adkian-sifive:
Thanks for the review @jackkoenig! Re: the plugin, the comment was meant to say that we'll need it in scala3 at some point but this update makes it so that building Scala3 is not blocked on the build units requiring the plugin..

adkian-sifive:
@jackkoenig any ideas on why the integration tests are failing to get the plugin in their build unit?

build.sc
```diff
@@ -294,44 +291,33 @@ trait Core extends common.CoreModule with CrossSbtModule with ScalafmtModule {
...
-
   def scalaCompilerIvy: Dep = v.scalaCompiler(crossScalaVersion)
+
+  def ivyDeps = super.ivyDeps() ++ Agg(scalaLibraryIvy, scalaReflectIvy, scalaCompilerIvy)
 }
 
 object chisel extends Cross[Chisel](v.scalaCrossVersions)
-
-trait Chisel extends common.ChiselModule with CrossSbtModule with ScalafmtModule {
+trait Chisel extends CrossSbtModule with HasScala2MacroAnno with ScalafmtModule {
```
jackkoenig, line 315: > any ideas on why the integration tests are failing to get the plugin in their build unit?

@adkian-sifive `Chisel` also needs the compiler plugin `HasScala2Plugin`, that's why integration tests are failing. (regular CI is failing for the same reason)

adkian-sifive: yeah that's it, thanks

sequencer:
I'll take a look now, sry for the delay

panama.sc
```diff
@@ -0,0 +1,252 @@
...
+        .replaceAll("@JAVA_LIBRARY_PATH@", javaLibraryPath().mkString(","))
+        .replaceAll("@CHISEL_LIT_DIR@", chiselLitDir().toString)
+    )
+    PathRef(T.dest)
+  }
+  def run(args: String*) = T.command(
+    os.proc("lit", litConfig().path)
+      .call(T.dest, stdout = os.ProcessOutput.Readlines(line => T.ctx().log.info("[lit] " + line)))
+  )
+}
```
sequencer, line 252: How about moving inline the `LitModule`?

build.sc
```diff
@@ -87,72 +82,17 @@ object v {
...
-  val linux = operationSystem.toLowerCase.startsWith("linux")
-  val windows = operationSystem.toLowerCase.startsWith("win")
-  val amd64 = architecture.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64)$")
-  val aarch64 = architecture.equals("aarch64") | architecture.startsWith("armv8")
-
+private object utils extends Module {
   val firtoolVersion = {
     val j = _root_.upickle.default.read[Map[String, String]](os.read(millSourcePath / os.up / "etc" / "circt.json"))
     j("version").stripPrefix("firtool-")
   }
```
sequencer, line 93: how about moving this to `object v` since it defines the firtool version, which may be shared.

build.sc
```diff
@@ -87,72 +82,17 @@ object v {
...
-          )
-          T.ctx().log.info(s"Downloading jextract from ${url}")
-          mill.util.Util.download(url, os.rel / "jextract.tar.gz")
-          T.ctx().log.info(s"Download Successfully")
-        }
-        os.proc("tar", "xvf", tarPath, "--strip-components=1").call(T.dest)
-        T.dest
-    }
-  }
+  def isScala3(ver: String): Boolean = ver.startsWith("3.")
```
sequencer, line 95: And this can also be moved to `v`

build.sc
```diff
@@ -182,80 +122,138 @@ trait ChiselPublishModule extends CiReleaseModule {
...
+    v.commonText,
+    v.osLib,
+    v.json4s
+  )
 
-  def scoptIvy = v.scopt
+  def ivyDeps = if (utils.isScala3(crossScalaVersion)) {
+    commonDeps
+  } else {
+    commonDeps ++ Agg(v.dataclass)
```
sequencer, line 180: IIRC dataclass is not used anymore? It can be removed in the future API.

build.sc
```diff
@@ -182,80 +122,138 @@ trait ChiselPublishModule extends CiReleaseModule {
...
-    "-Xsource:3",
-    "-Xsource-features:case-apply-copy-access"
-  )
+  override def scalacOptions = T {
+    if (utils.isScala3(crossScalaVersion)) {
+      Seq.empty[String]
+    } else {
+      v.scala2CommonOptions ++ Seq(
+        "-Xsource:3",
+        "-Xsource-features:case-apply-copy-access"
```
sequencer, line 198: why add this feature?

sequencer (commented):
Added some nitpick, but basically LGTM. I saw it removed the `common.sc`, I used to design it to separate the definition and implementation of the build system. I do think we can remove it after making mill as the only build system support. 
I think panama-related library can also migrate to Scala3 simply, we can do it in the future.

build.sc
adkian-sifive: this is needed in Scala3 if you make the accessors private, here it avoids errors with Xsource:3

build.sc
adkian-sifive: ok i'll make a note here

jackkoenig (approved):

PR #4505, opened 2024-11-13
Change default layers to use lowercase directories

seldridge:
Change directories for default layers to use lowercase output directories. While this is independently likely a better choice, it is a somewhat pragmatic, as these directory names match the legacy names of `firtool`'s extract-test-code feature (which was a custom SiFive pass for the SFC before being open-sourced in `firtool`).  This will help us land an extract-test-code replacement without having to disturb build flows. Secondarily, and independently, lower case directories are likely better.

There are two alternative changes which were rejected:

1. The layers could be renamed to be lowercase.  Generally, it's better to have objects and classes be uppercase.

2. The automatic directory names could be mangled to be lowercase as opposed to the exact layer names.

#### Release Notes

Change directories for default layers to be lowercase.

### DIFF
--- a/core/src/main/scala/chisel3/layers/Layers.scala
+++ b/core/src/main/scala/chisel3/layers/Layers.scala
@@ -1,17 +1,34 @@
 package chisel3.layers
 
 import chisel3.experimental.UnlocatableSourceInfo
-import chisel3.layer.{Layer, LayerConfig}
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM. I think this is better default behavior

PR #4506, opened 2024-11-13
Wrap printfs in the Verification layerblock

seldridge:
Change the emission of printfs to emit these under the `Verification` layerblock.  This is done as part of replacing the `firtool` extract-test-code feature with layers and the "advanced layer sink" `firtool` feature.

The extract-test-code feature will extract asserts _and_ printfs together. This is technically a deviation from this behavior.  However, it seems better to not lump these in with asserts.  Nonetheless, this is a backwards compatible change from the perspective of extract-test-code because the `Verification` layer is always loaded if the `Assert` layer is present.  (Prints will always show up if you turn on assertions.)

In the future, we probably want to add a `Debug` layer under the `Verification` layer where these will live.  I would like to hold off on making this change until extract-test-code is gone.

#### Release Notes

Move printfs under the `Verification` layer by default.

### DIFF
--- a/src/test/scala/chiselTests/Printf.scala
+++ b/src/test/scala/chiselTests/Printf.scala
@@ -8,24 +8,20 @@ import circt.stage.ChiselStage
 class SinglePrintfTester() extends Module {
   val x = 254.U
   printf("x=%x", x)
-  stop()
 }
 
 class ASCIIPrintfTester() extends Module {
   printf((0x20 to 0x7e).map(_.toChar).mkString.replace("%", "%%"))
-  stop()
 }
 
 class MultiPrintfTester() extends Module {
   val x = 254.U
   val y = 255.U
   printf("x=%x y=%x", x, y)
-  stop()
 }
 
 class ASCIIPrintableTester extends Module {
   printf(PString((0x20 to 0x7e).map(_.toChar).mkString("")))
-  stop()
 }
 
 class ScopeTesterModule extends Module {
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/Printf.scala
```diff
@@ -8,24 +8,20 @@ import circt.stage.ChiselStage
 class SinglePrintfTester() extends Module {
   val x = 254.U
   printf("x=%x", x)
-  stop()
```
seldridge, line 11: This stop is removed so that the test can simply check if a layerblock shows up. Stops are placed under layerblocks which means that the test would pass even without the change in this PR to move printfs under layers.

FileCheck would really help here. 🙃

sequencer:
What if user specially want to put the printf into another layer?

seldridge:
> What if user specially want to put the printf into another layer?

If the user wraps them in any other layer, then the printf will go in that layer. If the user puts them in a module with layers enabled, then they will not be put in a layer.

This is using the same additional arguments that `layer.block` has for `skipIfAlreadyInBlock` and `skipIfLayersEnabled`: https://github.com/chipsalliance/chisel/pull/4506/files#diff-989569982cecd89c52b0475d7be8bba9b776eb074224f89791a1c02f6fb6c1feR79

So, a user can override this on a case-by-case basis.

If the user wants to globally change a layer to another layer or if they cannot change the printfs, then they need to use the `--remap-layers` argument to Chisel (https://github.com/chipsalliance/chisel/pull/4322). The case for this is if we have printfs in upstream Chisel and the user wants to put these into a different layer, but can't modify the source.

sequencer:
neat!

PR #4519, opened 2024-11-19
Refactor Lookupable

jackkoenig:
See Release Notes for description.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

Historically, the Lookupable API is able to change the type of fields looked up from Definitions or Instances. This enabled Module fields to be looked up as Instances, as well as user-defined types to opt-in to this same Instance-boxing behavior.

This path-dependent type changing behavior is now deprecated. Looking up Modules is also deprecated, instead, the user should cast them Instances (via `.toInstance`). It is also deprecated to mark user-defined types as `@instantiable`. Instead, users should define Lookupable for their types using the new Lookupable.product[1-5] factory methods. See the Chisel website for more details.

* Deprecate user-extension of trait Lookupable.
* Deprecate Lookupable.lookupModule. Users should use Instances instead
  of Modules.
* Deprecate Lookupable.isInstantiable. User should use new factories to
  implement Lookupable for their user-defined types instead.
* Deprecate Lookupable.SimpleLookupable.
* Add Lookupable.isLookupable factory for "simple" Lookupables.
* Add Lookupable.product1-5 factories for Lookupables for user-defined types.
* Add Lookupable for Tuple3-5 (already existed for Tuple2).
* Add private LookupableImpl which is simpler to implement.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/IsInstantiable.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/IsInstantiable.scala
@@ -11,6 +11,10 @@ trait IsInstantiable
 
 object IsInstantiable {
   implicit class IsInstantiableExtensions[T <: IsInstantiable](i: T) {
+    @deprecated(
+      "Use of @instantiable on user-defined types is deprecated. Implement Lookupable for your type instead.",
+      "Chisel 7.0.0"
+    )
     def toInstance: Instance[T] = new Instance(Proto(i))
   }
 }
--- a/core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
+++ b/core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
@@ -13,13 +13,31 @@ import chisel3.internal.firrtl.ir.{Arg, ILit, Index, LitIndex, ModuleIO, Slot, U
 import chisel3.internal.{throwException, Builder, ViewParent}
 import chisel3.internal.binding.{AggregateViewBinding, ChildBinding, CrossModuleBinding, ViewBinding, ViewWriteability}
 
-/** Represents lookup typeclass to determine how a value accessed from an original IsInstantiable
-  *   should be tweaked to return the Instance's version
-  * Sealed.
+/** Typeclass used to recontextualize values from an original Definition to an Instance
+  *
+  * Implementations for Chisel types and other common Scala types are provided in the companion object.
+  * Users can define Lookupable for their own types.
+  *
+  * @example {{{
+  * case class Foo[T <: Data](name: String, data: T)
+  * object Foo {
+  *   implicit def lookupable[T <: Data]: Lookupable.Simple[Foo[T]] =
+  *     Lookupable.product2[Foo[T], String, T](
+  *       x => (x.name, x.data),
+  *       Foo.apply
+  *     )
+  * }
+  * }}}
   */
 @implicitNotFound(
-  "@public is only legal within a class or trait marked @instantiable, and only on vals of type" +
-    " Data, BaseModule, MemBase, IsInstantiable, IsLookupable, or Instance[_], or in an Iterable, Option, Either, or Tuple2"
+  "@public is only legal within a class or trait marked @instantiable, and only on vals of types" +
+    " that have a Lookupable implementation. Chisel types like Data, BaseModule, and MemBase are supported," +
+    " as are common Scala types like String, Int, Boolean, Iterable, Option, Either, and Tuples." +
+    " Please implement Lookupable for ${B}."
+)
+@deprecatedInheritance(
+  "Users should use Lookupable factory methods instead (e.g. Lookupable.isLookupable or Lookupable.product<n>)",
+  "Chisel 7.0.0"
 )
 trait Lookupable[-B] {
   type C // Return type of the lookup
@@ -40,6 +58,24 @@ trait Lookupable[-B] {
   def definitionLookup[A](that:     A => B, definition: Definition[A]): C
   protected def getProto[A](h:      Hierarchy[A]): A = h.proto
   protected def getUnderlying[A](h: Hierarchy[A]): Underlying[A] = h.underlying
+
+  // Single method that may eventually replace instanceLookup and definitionLookup.
+  private[chisel3] def hierarchyLookup[A](that: A => B, hierarchy: Hierarchy[A]): C = {
+    hierarchy match {
+      case h: Instance[A]   => instanceLookup(that, h)
+      case h: Definition[A] => definitionLookup(that, h)
+      case h => throw new InternalErrorException(s"Match error: hierarchy=$hierarchy")
+    }
+  }
+}
+// Simplify the implementation of Lookupable by only having 1 virtual method.
+// Only the Chisel core types need separate implementations for Instance and Definition.
+private trait LookupableImpl[-B] extends Lookupable[B] {
+
+  protected def impl[A](that: A => B, hierarchy: Hierarchy[A]): C
+
+  override def instanceLookup[A](that:   A => B, instance:   Instance[A]):   C = impl(that, instance)
+  override def definitionLookup[A](that: A => B, definition: Definition[A]): C = impl(that, definition)
 }
 
 object Lookupable {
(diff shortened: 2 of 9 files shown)


core/src/main/scala/chisel3/experimental/hierarchy/core/IsInstantiable.scala
```diff
@@ -11,6 +11,10 @@ trait IsInstantiable
 
 object IsInstantiable {
   implicit class IsInstantiableExtensions[T <: IsInstantiable](i: T) {
+    @deprecated(
+      "Use of @instantiable on user-defined types is deprecated. Implement Lookupable for your type instead.",
```
mwachs5, line 15: should we point to website docs on this?

core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
```diff
@@ -13,13 +13,31 @@ import chisel3.internal.firrtl.ir.{Arg, ILit, Index, LitIndex, ModuleIO, Slot, U
...
+  *       x => (x.name, x.data),
+  *       Foo.apply
+  *     )
+  * }
+  * }}}
   */
 @implicitNotFound(
-  "@public is only legal within a class or trait marked @instantiable, and only on vals of type" +
-    " Data, BaseModule, MemBase, IsInstantiable, IsLookupable, or Instance[_], or in an Iterable, Option, Either, or Tuple2"
+  "@public is only legal within a class or trait marked @instantiable, and only on vals of types" +
```
mwachs5, line 33: i thought user can no longer mark things @instantiable...? Or they can still mark Modules and such as @instantiable to get the @public API?

core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
```diff
@@ -40,6 +58,24 @@ trait Lookupable[-B] {
   def definitionLookup[A](that:     A => B, definition: Definition[A]): C
   protected def getProto[A](h:      Hierarchy[A]): A = h.proto
   protected def getUnderlying[A](h: Hierarchy[A]): Underlying[A] = h.underlying
+
+  // Since method that may eventually replace instanceLookup and definitionLookup.
```
mwachs5, line 62: not sure about this comment's english... what does "Since method" mean?

docs/src/cookbooks/hierarchy.md
```diff
@@ -85,15 +85,16 @@ chisel3.docs.emitSystemVerilog(new AddTwoInstantiate(16))
...
+The requirements are that the field is publicly accessible, is a `val` or `lazy val`, and must have an implementation of `Lookupable`.
+
+Types that are supported by default are:
+
+1. `Data`
+2. `BaseModule`
+3. `MemBase`
+4. `IsLookupable`
+5. `Iterable`/`Option`/`Either` containing a type that meets these requirements
 6. Basic type like `String`, `Int`, `BigInt` etc.
```
mwachs5, line 98:
```suggestion
6. Basic type like `String`, `Int`, `BigInt`, `Unit`, etc.
```

docs/src/cookbooks/hierarchy.md
```diff
@@ -167,7 +170,92 @@ val chiselCircuit = (new chisel3.stage.phases.Elaborate)
...
+}
+@instantiable
+class HasUserDefinedType extends Module {
+  val inst = Module(new ChildModule)
+  val wire = Wire(UInt(8.W))
+  @public val x = UserDefinedType("foo", wire, inst.toInstance)
+}
+```
+
+We can implement the `Lookupable` type class for `UserDefinedType` in order to make it accessible.
```
mwachs5, line 202: nit 
this is getting a bit into 'explanation' over 'cookbook'

docs/src/cookbooks/hierarchy.md
```diff
@@ -167,7 +170,92 @@ val chiselCircuit = (new chisel3.stage.phases.Elaborate)
...
+
+**If any fields are `BaseModules`, you must change them to be `Instance[_]` in order to define the `Lookupable` typeclass.**
+
+For more information about typeclasses, see the [DataView section on Type Classes](https://www.chisel-lang.org/chisel3/docs/explanations/dataview#type-classes).
+
+```scala mdoc
+import chisel3.experimental.hierarchy.Lookupable
+object UserDefinedType {
+  // Use Lookupable.Simple type alias as return type.
+  implicit def lookupable[T <: Data, M <: RawModule]: Lookupable.Simple[UserDefinedType[T, M]] =
```
mwachs5, line 216: it feels like this example is more confusing than necessary with the `T` and `M` type parameters. Could you show a simpler example first with concrete types like `MyBundle` and `MyModule`? It's hard to separate in my mind which part of this is type parameters for the `Lookupable.product3` and which are for the fact that our UserDefinedType is type parameterized itself.

docs/src/cookbooks/hierarchy.md
```diff
@@ -167,7 +170,92 @@ val chiselCircuit = (new chisel3.stage.phases.Elaborate)
...
+    Lookupable.product3[UserDefinedType[T, M], String, T, Instance[M]](
+      x => (x.name, x.data, x.inst),
+      UserDefinedType.apply // Use case class-generated factory method.
```
mwachs5, line 219:
```suggestion
      // Provide the recipe for converting the UserDefinedType to a Tuple
      x => (x.name, x.data, x.inst),
      // Provide the recipe for converting a Tuple to a user defined type.
      // For case classes, you can use the built-in factory method 
      UserDefinedType.apply
```

core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
jackkoenig: It's a typo, meant "single`, will fix.

core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala
jackkoenig: They can (and should) still mark `Modules` as `@instantiable`, but cannot[^1] mark user-defined types `@instantiable`.

[^1]: Or really _should not_, I can't figure out how to enforce it on the `@instantiable` macro because it runs before type inference. It's only enforced at the point of marking a val of the user-define type `@public`

src/test/scala/chiselTests/experimental/hierarchy/Examples.scala
```diff
@@ -223,6 +223,12 @@ object Examples {
     @public val xy = (x, y)
   }
   @instantiable
+  class HasTuple5() extends Module {
+    val wire = Wire(UInt(3.W))
+    val inst = Module(new AddOne)
+    @public val tup = (3, wire, "hi", inst, List(1, 2, 3))
```
azidar, line 229: Shouldn't this be deprecated, looking up the `inst` because it is a `Module` type?

docs/src/cookbooks/hierarchy.md
azidar: Yeah I agree we should have a dead simple example, and then a second one with the type parameters.

src/test/scala/chiselTests/experimental/hierarchy/Examples.scala
jackkoenig: It is deprecated, but I've added warning suppression to the `build.sc`. I wanted to check that I implemented the tuple Lookupables correctly such that they still support the deprecated API. Follow on work will delete all of the deprecated stuff.

azidar (approved):
I'll approve, but please add a simpler example to both the documentation and the unit tests. Thanks!

docs/src/cookbooks/hierarchy.md
jackkoenig: Improved a bit I think!

docs/src/cookbooks/hierarchy.md
jackkoenig: Done, I split it into 3 sections, a simpler example, the type parameterized one, and then the "lots of fields" is its own.

PR #4532, opened 2024-11-26
Add APIs to omit module prefix separator

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

* `withModulePrefix` now optionally takes a 2nd parameter `includeSeparator` (passing `false` will omit separator).
* Overriding `localModulePrefixUseSeparator` in a Module to false will omit separator for local prefix applied by overriding `localModulePrefix` in that Module.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -241,10 +241,6 @@ private[chisel3] trait ObjectModuleImpl {
     /** Explicitly Asynchronous Reset */
     case object Asynchronous extends Type
   }
-
-  def getModulePrefixList: List[String] = {
-    Builder.getModulePrefixList
-  }
 }
 
 private[chisel3] trait ModuleImpl extends RawModule with ImplicitClock with ImplicitReset {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -241,10 +241,6 @@ private[chisel3] trait ObjectModuleImpl {
     /** Explicitly Asynchronous Reset */
     case object Asynchronous extends Type
   }
-
-  def getModulePrefixList: List[String] = {
-    Builder.getModulePrefixList
-  }
```
jackkoenig, line 247: I removed this API because it's complicated by the separator being optional and it isn't currently used (it was used internally in `Instantiate` but its use would technically be wrong with separator now being optional).

src/test/scala/chiselTests/ModulePrefixSpec.scala
```diff
@@ -357,4 +375,27 @@ class ModulePrefixSpec extends ChiselFlatSpec with ChiselRunners with Utils with
 
     matchesAndOmits(chirrtl)(lines: _*)()
   }
+
+  it should "omit the prefix if localModulePrefixUseSeparator is false" in {
+    class Foo extends RawModule {
+      override def localModulePrefix = Some("Inner")
+      override def localModulePrefixUseSeparator = false
+    }
+    class Top extends RawModule {
```
mwachs5, line 384: should we add a test if a Parent with localModulePrefixUseSeparator = false instantiates a child with it at the default `true` and confirm you get something like `ParentChild_Grandkid`?

src/test/scala/chiselTests/ModulePrefixSpec.scala
```diff
@@ -357,4 +375,27 @@ class ModulePrefixSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
+      override def localModulePrefixAppliesToSelf = false
+      val foo = Module(new Foo)
+    }
+
+    val chirrtl = emitCHIRRTL(new Top)
+    val lines =
+      """
+      module OuterInnerFoo :
+      module Top :
+        inst foo of OuterInnerFoo
```
mwachs5, line 396: I mean a test like `OuterMiddle_Inner_Foo`

mwachs5: or maybe this test would be sufficient just put `Inner`s to `true`?

mwachs5 (approved):
approving, though we could add one more test case for showing the sep override is only for one layer of hierarchy

src/test/scala/chiselTests/ModulePrefixSpec.scala
jackkoenig: Good idea will add

src/test/scala/chiselTests/ModulePrefixSpec.scala
jackkoenig: This test is checking 2 different (but related) code paths, so I'll just add a separate test for your suggestion.

src/test/scala/chiselTests/ModulePrefixSpec.scala
jackkoenig: Done.

PR #4545, opened 2024-12-03
Fix auto generated DedupGroups to be sensitive to module prefix

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

Also fix logic for naming PseudoModules which should ignore the prefix.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -682,6 +682,17 @@ package experimental {
       */
     def desiredName: String = simpleClassName(this.getClass)
 
+    /** The name that will be proposed for this module, subject to uniquification.
+      *
+      * Includes the module prefix for user-defined modules (but not for blackboxes).
+      */
+    private[chisel3] def _proposedName: String = this match {
+      // PseudoModules (e.g. Instances) and BlackBoxes have their names set by desiredName.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/ModuleImpl.scala
```diff
@@ -682,6 +682,17 @@ package experimental {
       */
     def desiredName: String = simpleClassName(this.getClass)
 
+    /** The name that will be proposed for this module, subject to uniquification.
+      *
+      * Includes the module prefix for user-defined modules (but not for blackboxes).
+      */
+    private[chisel3] def _proposedName: String = this match {
+      // PseudoModules (e.g. Instances) and BlackBoxes have their names set by desiredName.
+      case _: PseudoModule => desiredName
```
jackkoenig, line 691: The careful reviewer will notice this changed from `Module.currentModulePrefix + desiredName`. It turns out that for `Instances` (which have an underlying `PseudoModule`, the `currentModulePrefix` is always empty so it's effectively a no-op. For other PseudoModules like `ViewParent`, there is no reason to have a prefix. I tweaked the code to make it more obvious that the prefix is not actually relevant to `PseudoModules`.

mwachs5 (approved):

PR #4547, opened 2024-12-04
[test] Convert LayerSpec to use FileCheck, NFC

seldridge:
Convert `chiselTests.LayerSpec` to use FileCheck.  This is much stricter testing as the ordering of matches is now guaranteed.

### DIFF
--- a/src/test/scala/chiselTests/LayerSpec.scala
+++ b/src/test/scala/chiselTests/LayerSpec.scala
@@ -39,7 +39,6 @@ class LayerSpec extends ChiselFlatSpec with Utils with FileCheck {
     }
 
     info("CHIRRTL emission looks correct")
-    // TODO: Switch to FileCheck for this testing.  This is going to miss all sorts of ordering issues.
     generateFirrtlAndFileCheck(new Foo)(s"""|CHECK:      layer A, bind, "A" :
                                             |CHECK-NEXT:   layer B, bind, "A${sep}B" :
                                             |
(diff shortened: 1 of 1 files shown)


src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -47,19 +38,17 @@ class LayerSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
-    matchesAndOmits(chirrtl)(
-      "layer A, bind, \"A\" :",
-      "layer B, bind, \"A" ++ sep ++ "B\" :",
-      "layerblock A :",
-      "wire w : UInt<1>",
-      "layerblock B :",
-      "wire x : UInt<1>",
-      "wire y : UInt<1>"
-    )()
+    generateFirrtlAndFileCheck(new Foo)(s"""|CHECK:      layer A, bind, "A" :
```
jackkoenig, line 42:
```suggestion
    generateFirrtlAndFileCheck(new Foo)(
      s"""|CHECK:      layer A, bind, "A" :
```
I'm going to approve this because it doesn't matter that much, but if you put a newline before the `s"""` then the tabbing on all of these will be *much* better (you can see some examples below).

jackkoenig (approved):
LGTM, consider massaging tabbing.

src/test/scala/chiselTests/LayerSpec.scala
seldridge: I'd like to do this. However, `scalafmt` will remove the newline. What I may do is opt for putting the filecheck string into a value?

seldridge: Better solution, use curly braces.

seldridge: Better indentation in https://github.com/chipsalliance/chisel/pull/4547/commits/08a321d7a8ae5198eb8725eff41a912013d39433 by using curly braces.

PR #4549, opened 2024-12-04
Add partial cross-compilation for Scala 3

adkian-sifive:
Adds initial support for Scala 3 with sources and reference to 3.3.3 LTS in the Mill build config. 

~This PR depends on #4467 which is (roughly) commit 0ba3c5, which are all the scala-2/ and scala/ changes visible in this PR. @jackkoenig and I need to determine whether those changes are necessary and either merge that PR first and rebase this on main, OR figure out an alternative.~

I've added a CI test unit for Scala 3 compilation with Mill, @jackkoenig it probably needs your attention

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification

#### Desired Merge Strategy

#### Release Notes

Adds initial support for Scala 3 LTS version 3.3.3

- [ ] Did you review?


core/src/main/scala-3/chisel3/Module.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  /** A wrapper method that all Module instantiations must be wrapped in
+    * (necessary to help Chisel track internal state).
+    *
+    * @param bc the Module being created
+    *
+    * @return the input module `m` with Chisel metadata properly set
+    */
+  // TODO(adkian-sifive) the callsite here explicitly passes
+  // sourceInfo so it cannot be a contextual parameter
```
jackkoenig, line 17: This comment looks to be no longer true? But even if it were, you can always `summon[_]`.

core/src/main/scala-3/chisel3/Module.scala
```diff
@@ -0,0 +1,29 @@
...
+  def apply[T <: BaseModule](bc: => T): T = _applyImpl(bc)
+  def do_apply[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = apply(bc)
```
jackkoenig, line 19:
```suggestion
  def apply[T <: BaseModule](bc: => T)(using SourceInfo): T = _applyImpl(bc)
```
No do_apply and apply needs the context parameter.

core/src/main/scala-3/chisel3/Printf.scala
```diff
@@ -0,0 +1,59 @@
...
+      } else {
+        require(
+          x.toInt >= 32,
+          s"char ${x} to Int ${x.toInt} must be >= 32"
+        ) // TODO \xNN once FIRRTL issue #59 is resolved
+        x
+      }
+    }
+    formatIn.map(escaped).mkString("")
+  }
```
jackkoenig, line 41: Is there a reason we can't have this code in `src/main/scala` and share it?

core/src/main/scala-3/chisel3/Bits.scala
```diff
@@ -0,0 +1,648 @@
...
+  */
+sealed abstract class Bits(private[chisel3] val width: Width) extends BitsImpl with ToBoolable {
+
+  /** Tail operator
+    *
+    * @param n the number of bits to remove
+    * @return This $coll with the `n` most significant bits removed.
+    * @group Bitwise
+    */
+  def tail(n: Int)(using sourceInfo: SourceInfo): UInt = _tailImpl(n)
```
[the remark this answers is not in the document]
adkian-sifive: that works for me, but the current scalafmt version we've got in chisel cannot parse it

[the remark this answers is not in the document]
jackkoenig: Our old version of Scalafmt also has issues with VSCode+Mill, we should just update it. I have a branch, let me turn it into a PR.

core/src/main/scala-3/chisel3/Module.scala
adkian-sifive: so the comment here is referring to this callsite: https://github.com/chipsalliance/chisel/blob/dc39308e583a793c08dd124a42b5c5b8a0a1cf53/core/src/main/scala/chisel3/internal/package.scala#L133

i think we need the do_apply here? (same comment for below)

jackkoenig: That callsight should call `_applyImpl`, that's the replacement for `do_apply`. This is the same thing as the other places where we got rid of `do_` versions of things because we no longer need the macro on apply, we can just take a contextual parameter.

core/src/main/scala-3/chisel3/Module.scala
```diff
@@ -0,0 +1,29 @@
...
+    * (necessary to help Chisel track internal state).
+    *
+    * @param bc the Module being created
+    *
+    * @return the input module `m` with Chisel metadata properly set
+    */
+  // TODO(adkian-sifive) the callsite here explicitly passes
+  // sourceInfo so it cannot be a contextual parameter
+  def apply[T <: BaseModule](bc:    => T): T = _applyImpl(bc)
+  def do_apply[T <: BaseModule](bc: => T)(implicit sourceInfo: SourceInfo): T = apply(bc)
```
jackkoenig, line 19:
```suggestion
  def apply[T <: BaseModule](bc:    => T)(using sourceInfo: SourceInfo): T = _applyImpl(bc)
```
As mentioned above, change that `Module.do_apply` call in `chisel/core/src/main/scala/chisel3/internal/package.scala` to use `_applyImpl`.

core/src/main/scala-3/chisel3/VerificationStatementMacros.scala
```diff
@@ -0,0 +1,36 @@
...
+  )(
+    using sourceInfo: SourceInfo
+  ): Printable = {
+    val (filename, line) = lineInfo
+    val lineMsg = s"$filename:$line".replaceAll("%", "%%")
+    message match {
+      case Some(msg) =>
+        p"$kind failed: $msg\n"
+      case None => p"$kind failed at $lineMsg\n"
+    }
```
jackkoenig, line 34: Why is this duplicated?

core/src/main/scala-3/chisel3/experimental/hierarchy/IsA.scala
```diff
@@ -0,0 +1,48 @@
...
+  //  This function detects this case and reshapes the string to match.
+  private def modifyReplString(clz: String): String = {
+    if (clz != null) {
+      clz.split('.').toList match {
+        case "repl" :: "MdocSession" :: app :: rest => s"$app.this." + rest.mkString(".")
+        case other                                  => clz
+      }
+    } else clz
+  }
+
```
jackkoenig, line 25: This method should be in `src/main/scala`

core/src/main/scala/chisel3/BitsImpl.scala
```diff
@@ -99,7 +99,7 @@ private[chisel3] trait BitsImpl extends Element { self: Bits =>
   }
 
   protected def _applyImpl(x: UInt)(implicit sourceInfo: SourceInfo): Bool =
-    do_extract(x)
+    extract(x)
```
jackkoenig, line 102: It's probably fine, but I think all of these should be calling the `_*impl` versions

core/src/main/scala/chisel3/experimental/hierarchy/InstantiateImpl.scala
```diff
@@ -93,7 +92,7 @@ private[chisel3] trait InstantiateImpl {
     val modulePrefix = Module.currentModulePrefix
     Builder.contextCache
       .getOrElseUpdate(
-        CacheKey[A](boxAllData(args), tt, modulePrefix), {
+        CacheKey[A](boxAllData(args), tt, Seq(modulePrefix).toList), {
```
jackkoenig, line 95:
```suggestion
        CacheKey[A](boxAllData(args), tt, List(modulePrefix)), {
```

core/src/main/scala-3/chisel3/Module.scala
adkian-sifive: so making this change + changing the `Module.do_apply` to `Module._applyImpl` in internal/package.scala throws this error
```
core/src/main/scala/chisel3/internal/package.scala:133:12: method _applyImpl in trait ObjectModuleImpl cannot be accessed as a member of object chisel3.Module from package object internal in package internal
[error]  Access to protected method _applyImpl not permitted because
[error]  enclosing package object internal in package internal is not a subclass of
[error]  trait ObjectModuleImpl in package chisel3 where target is defined
[error]     Module._applyImpl(new ViewParentAPI)(UnlocatableSourceInfo)
[error]            ^
```

adkian-sifive: not really sure what subclassing requirement is not being satisfied here

adkian-sifive: i'm guessing it's cause _applyImpl is protected in ObjectModuleImpl?

jackkoenig: Right, try making it `protected[chisel3]` instead

jackkoenig (approved):
There's much more to do and I'm sure there will be issues we need to solve, but this is a huge step! Great work @adkian-sifive 

FYI @sequencer

sequencer:
Extraordinary work!
BTW, I’m trying to figure out a new types system and Scala 3 only design pattern and ask for public review before next Q1.

PR #4551, opened 2024-12-05
Fix potential name collision caused by LTL properties

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

#### Release Notes

Some nodes created by LTL properties had `suggestName` called on them, potentially leading to name collisions with user code.

- [ ] Did you review?

### DIFF
--- a/src/main/scala-2/chisel3/util/circt/LTLIntrinsics.scala
+++ b/src/main/scala-2/chisel3/util/circt/LTLIntrinsics.scala
@@ -37,7 +37,7 @@ private[chisel3] object UnaryLTLIntrinsic {
   )(
     implicit sourceInfo: SourceInfo
   ): Bool =
-    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(_in).suggestName(intrinsicName)
+    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(_in)
 }
 
 /** Base instrinsic for all binary intrinsics with `lhs`, `rhs`, and `out` ports. */
@@ -50,7 +50,7 @@ private[chisel3] object BinaryLTLIntrinsic {
   )(
     implicit sourceInfo: SourceInfo
   ): Bool =
-    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs).suggestName(intrinsicName)
+    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs)
 }
 
 /** A wrapper intrinsic for the CIRCT `ltl.and` operation. */
--- a/src/main/scala-2/chisel3/util/circt/LTLIntrinsics.scala
+++ b/src/main/scala-2/chisel3/util/circt/LTLIntrinsics.scala
@@ -37,7 +37,7 @@ private[chisel3] object UnaryLTLIntrinsic {
   )(
     implicit sourceInfo: SourceInfo
   ): Bool =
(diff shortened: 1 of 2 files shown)


src/test/scala/chiselTests/LTLSpec.scala
```diff
@@ -407,4 +407,21 @@ class LTLSpec extends AnyFlatSpec with Matchers with ChiselRunners {
     assert(assertBlockLoc < delayIntrinsicLoc)
     assert(assumeblockLoc < implicationIntrinsicLoc)
   }
+
+  it should "not produce name collisions with clock" in {
```
tmckay-sifive, line 411: I added this test to reproduce the bug. It seems like a silly test now that there is no possibility for a name collision. However, it also felt wrong to delete a test, even if it is trivially passing.

Should we keep it?

jackkoenig: We should keep it, but possibly extend it to check the names that come out

src/main/scala-2/chisel3/util/circt/LTLIntrinsics.scala
```diff
@@ -50,7 +50,7 @@ private[chisel3] object BinaryLTLIntrinsic {
   )(
     implicit sourceInfo: SourceInfo
   ): Bool =
-    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs).suggestName(intrinsicName)
+    BaseIntrinsic(f"ltl_$intrinsicName", Bool(), params)(lhs, rhs)
```
jackkoenig, line 53: Removing these suggestNames makes the naming quality a lot worse (as evidenced by the updates to the tests) so I think we want to find a different way to solve this. Can you explain a bit more about what is problematic about the collision you were seeing? Was something failing?

tmckay-sifive: The issue I was encountering was that the properties and sequences in the LTL library create nodes for the intrinsics. These nodes are named the same as the intrinsic kind. 

This can result in confusing behavior, especially for intrinsics with names that are likely to show up in user code e.g. `clock` or `delay`.

This is a case I constructed to produce the error.

```
class Test extends RawModule {
  val io = IO(Input(UInt(8.W)))

  val clockWire = Wire(Clock())

  withClock(clockWire) {
    AssertProperty(Property.eventually(io.orR))
  }

  val clock = IO(Input(Clock()))
  clockWire := clock
}
```

The assert property creates a node named `clock`, but it is not obvious that it would do so. Then, when attempting to create another node `clock` in user code, Chisel complains about the name collision.

More broadly I think the issue is that calling `suggestName` inside Chisel can make it very tough to determine why name collisions are happening.

jackkoenig: I see, so the issue is that the port comes after the property. I do think the real solution is that Chisel needs to track where the previously named thing was so we can point to it. But I see your point that this was hard to debug and we certainly want to reduce the odds that this happens to users. 

How about something a bit lighter touch--keep the `.suggestName`, but use names that are less likely to collide with signals named by users, e.g. suggest `ltl_clock` instead of `clock`?

src/test/scala/chiselTests/LTLSpec.scala
tmckay-sifive: Sounds good, will add the extended test shortly

jackkoenig (approved):
LGTM!

PR #4554, opened 2024-12-10
[choice] Adds a means to declare a choice group without use

davidbiancolin:
Per @seldridge's suggestion here I add analog of `addLayer` for `ModuleChoice` called `addGroup`. This permits defining a group in a circuit without necessarily using that group in a circuit, allowing for commonly reused top-level modules to always have these choices defined regardless of what is instantiated in their module hierarchies. 

Short of reworking the group API a bit to more explicitly register the `Case`s in the `Group` at construction time i couldn't figure out a means to do what i wanted, so i resorted to using Scala reflection. If there's a smarter way to do this i'm all ears. 

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference)

#### Release Notes
Introduce `choice.addGroup` which enables declaration of a `choice.Group` without its use (akin to `layer.addLayer`)

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/choice/package.scala
+++ b/core/src/main/scala/chisel3/choice/package.scala
@@ -3,7 +3,12 @@
 package chisel3
 
 import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal.Builder
 import chisel3.util.simpleClassName
+
+import scala.collection.mutable
+import scala.reflect.runtime.universe._
+import scala.reflect.runtime.{currentMirror => cm}
 
 /** This package contains Chisel language definitions for describing configuration options and their accepted values.
   */
@@ -13,7 +18,7 @@ package object choice {
     *
     * @example
     * {{{
-    * import chisel3.option.{Group, Case}
+    * import chisel3.choice.{Group, Case}
     * object Platform extends Group {
     *   object FPGA extends Case
     *   object ASIC extends Case
@@ -46,5 +51,28 @@ package object choice {
       * @param module Module to map to the current case.
       */
     def ->[T](module: => T): (Case, () => T) = (this, () => module)
+  }
+
+  /** Registers all options in a group with the Builder.
+    * This lets Chisel know  that this layer should be emitted into FIRRTL text.
+    *
+    * This API can be used to guarantee that a design will always have certain
+    * group defined.  This is analagous in spirit to [[layer.addLayer]].
+    */
+  def addGroup[T <: Group: TypeTag](group: T): Unit = {
+
+    val tpe = typeOf[T]
+    val classSymbol = tpe.typeSymbol.asClass
+    val classMirror = cm.reflectClass(classSymbol)
+
+    tpe.members.collect {
+      // Look only for inner objects.  Note, this is not recursive.
+      case m: ModuleSymbol if m.isStatic =>
+        val instance = cm.reflectModule(m.asModule).instance
+        // Confirms the instance is a subtype of Case
+        if (cm.classSymbol(instance.getClass).toType <:< typeOf[Case]) {
+          Builder.options += instance.asInstanceOf[Case]
+        }
+    }
   }
 }
--- a/src/test/scala/chiselTests/ModuleChoiceSpec.scala
+++ b/src/test/scala/chiselTests/ModuleChoiceSpec.scala
@@ -3,7 +3,7 @@
 package chiselTests
 
 import chisel3._
-import chisel3.choice.{Case, Group, ModuleChoice}
+import chisel3.choice.{addGroup, Case, Group, ModuleChoice}
 import chiselTests.{ChiselFlatSpec, MatchesAndOmits, Utils}
 import _root_.circt.stage.ChiselStage
 
@@ -128,4 +128,23 @@ class ModuleChoiceSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
 
   }
 
+}
+class AddGroupSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
+  it should "emit options for a registered group even if there are no consumers" in {
+    class ModuleWithoutChoice extends Module {
+      addGroup(Platform)
+      val out = IO(UInt(8.W))
+      val in = IO(UInt(8.W))
+      out := in
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(new ModuleWithoutChoice, Array("--full-stacktrace"))
+
+    info("CHIRRTL emission looks correct")
+    matchesAndOmits(chirrtl)(
+      "option Platform :",
+      "FPGA",
+      "ASIC"
+    )()
+  }
 }


src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -129,3 +129,22 @@ class ModuleChoiceSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(new ModuleWithoutChoice, Array("--full-stacktrace"))
+
+    info("CHIRRTL emission looks correct")
+    matchesAndOmits(chirrtl)(
+      "option Platform :",
+      "FPGA",
+      "ASIC"
+    )()
```
seldridge, line 148: I landed some of Jack's work to make this checking more strict. Instead of `matchesAndOmits`, can this use the following:

```suggestion
    fileCheckString(chirrtl)(
      """|CHECK: option Platform :
         |CHECK-NEXT: FPGA
         |CHECK-NEXT: ASIC
         |""".stripMargin
    )()
```

This will need to mix `chiselTests.FileCheck` into `AddGroupSpec`.

seldridge: Also, inlining `chirrtl` into the usage site can allow you to use `generateFirrtlAndFileCheck`.

src/test/scala/chiselTests/ModuleChoiceSpec.scala
```diff
@@ -129,3 +129,22 @@ class ModuleChoiceSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
...
+class AddGroupSpec extends ChiselFlatSpec with Utils with MatchesAndOmits {
+  it should "emit options for a registered group even if there are no consumers" in {
+    class ModuleWithoutChoice extends Module {
+      addGroup(Platform)
+      val out = IO(UInt(8.W))
+      val in = IO(UInt(8.W))
+      out := in
+    }
+
+    val chirrtl = ChiselStage.emitCHIRRTL(new ModuleWithoutChoice, Array("--full-stacktrace"))
```
seldridge, line 141: Nit: `Array("--full-stacktrace")` can be dropped.

core/src/main/scala/chisel3/choice/package.scala
```diff
@@ -47,4 +52,27 @@ package object choice {
...
+    tpe.members.collect {
+      // Look only for inner objects.  Note, this is not recursive.
+      case m: ModuleSymbol if m.isStatic =>
+        val instance = cm.reflectModule(m.asModule).instance
+        // Confirms the instance is a subtype of Case
+        if (cm.classSymbol(instance.getClass).toType <:< typeOf[Case]) {
+          Builder.options += instance.asInstanceOf[Case]
+        }
+    }
+  }
```
seldridge, line 77: This is clever reflection. 👍  However, it may be better to just make this information more easily accessible. Two changes would avoid the `TypeTag` (which is a Scala3 migration annoyance, I think) and any work in this function:

1. Add a `Builder` member for storing `groups` and not just options.
2. Add `mutable.LinkedHashSet` to `Group` that can be used to directly look up what `Case`s it supports. Then modify `Case` to insert into it.

seldridge (commented):
The API of `addGroup` looks great.

One suggestion on how to avoid the reflection and a nit in the test to use `FileCheck`.

core/src/main/scala/chisel3/choice/package.scala
davidbiancolin: I started doing a (2) but wasn't sure when/how to finalize the set short of doing something either janky or switching the API in someway to use a different pattern. 

But (1) seems really easy to do; the existing struct can probably just become a HashMap and we wouldn't need to reconstruct the the groups later.

davidbiancolin: Ok i got rid of the type tag but i can't figure out how to avoid it altogether unless we change the API for defining these cases (so that they are not lazy initialized).

Alternatively, we could change the `addGroup` API to to instead take a one or more cases. For example `addGroupHint(c: Case)` would suffice to workaround my problem, but it seems rather hacky to not register the whole group and breaks the direct `addLayer` analogy.

However, I doubt breaking the `Group API here is a big deal since no one should be using this? I'd change the `Group` implementation to use a more factory like pattern that would force registration on the definition of the objects.  

LMK

seldridge: Yeah, I think the laziness of the _cases_ seems wrong. Laziness of _groups_ seems fine. (You only get a group if you use it, but you get all of its cases.)

This would be a change to have the `Builder` register groups, have cases register that they are children of groups, and then pushing that change through to the converter.

However, all of this is independent of any change to the API. Given that, I will go ahead and approve this.

seldridge (approved):
LGTM

The internals of groups need to be reworked. That can be done in a follow-on (or backlogged for someone else to do).

davidbiancolin:
Also there seems to be some consistency referring to groups / cases and options in the code. Do you want me to go through and update the code and docs to match either the structures in here, or the ABI docs?

core/src/main/scala/chisel3/choice/package.scala
jackkoenig: I'm not sure if this reflection works in Scala 3 and as we are actively trying to get Scala 3 compilation into CI, I think we need to address it sooner rather than later. Can we do something similar to how `Layers` get the parents with an implicit rather than reflecting on the `Group`?

davidbiancolin: Well, can i break the existing API? If so this is pretty straight forward. Otherwise i might need one of you guys (i.e., someone with more Scala foo than i) to do it. The crux of it is that the Cases are lazily constructed so _something_ has to reference them to bind them to the group. If you're coming into the office today maybe we can chat about it?

seldridge: You can break it as much as you want. Just don't backport it to 6.x. This feature has no users and you won't be disturbing anyone.

jackkoenig:
@davidbiancolin and I discussed this offline

Unfortunately just having the `Group` as an implicit to the `Case` does not initialize the `Case` object, so to maintain the current API, the only way I can see to do it is with such reflection. I've advised David on how to split the reflection out into `src/main/scala-2`.

We discussed an alternative proposal though to switch to the style of `ChiselEnum`, basically instead of
```scala
object Platform extends Group {
  object FPGA extends Case
  object ASIC extends Case
}
```
You would write:
```scala
object Platform extends Group {
  val FPGA = Case
  val ASIC = Case
  val Special = Case("SpecialPlatform") // can manually pass the String too
}
```
This requires a simple macro to grab the val name a la `ChiselEnum`[^1], or we could just upstream `ValName` and do it that way[^2]. I think this has 3 advantages:
1. It makes it impossible to accidentally nest a `Case` instead of another `Case` which isn't valid.
2. It's more consistent with enumeration style rather than layer-style (which does allow nesting).
3. I think compile-time reflection (aka macros) is almost always preferable to runtime-reflection. It's more transparent.

What do you think @seldridge?

[^1]: https://github.com/chipsalliance/chisel/blob/4232cae4189abcebf1a7e244cb3875d85083e035/core/src/main/scala-2/chisel3/ChiselEnum.scala#L34
[^2]: https://github.com/chipsalliance/rocket-chip/blob/2fe6bb5e9c23360cac8297bb26512e4ca614f67b/macros/src/main/scala/ValName.scala#L8

seldridge:
Everything in this API can be broken as there are no users. I wouldn't worry about that. I would prefer to not introduce any new macros.

Why would something simple like having the `Case` insert themselves inside the `Group` not work? This is what I've been in favor of even if it means the API changes.

I can see a potential reason to not require the nesting and to keep the implicit, though I'm not sure if it would come up. If we have builtin groups and cases defined, then we may want the ability to add new cases to an existing group. I do expect that we will eventually have built-in groups that memories use. It is possible that we want users to be able to add cases to these groups. (Alternatively, this could be just done with a generic blackbox.)

I do agree that these are trying to describe an enum.

davidbiancolin:
> Why would something simple like having the `Case` insert themselves inside a mutable set inside the `Group` not work? 

The crux of it is `Case` cannot insert itself into a `Group` unless something references that `Case`, since they are objects and so lazily constructed. So either we have the user create that reference in the definition of group, or use a macro or reflection induce that reference without requiring the user to do so. 

The prior could look as follows:

```
object Platform extends Group {
  object FPGA extends Case
  object ASIC extends Case
  val cases = Seq(FPGA, ASIC)
}
```

But this seems like a terrible design. Alternatively, one could make case a factory:

```
object Platform extends Group {
  val ASIC =  Case("ASIC")
  val FPGA = Case("FPGA")
}
```
Then you have the doubly define names that lead one towards using a macro as done in other libraries to auto-name the cases...

```
object Platform extends Group {
  val ASIC, FPGA =  Case
}
```

core/src/main/scala/chisel3/choice/package.scala
```diff
@@ -3,8 +3,13 @@
 package chisel3
 
 import chisel3.experimental.{BaseModule, SourceInfo}
+import chisel3.internal.Builder
 import chisel3.util.simpleClassName
 
+import scala.collection.mutable
+import scala.reflect.runtime.universe._
+import scala.reflect.runtime.{currentMirror => cm}
```
adkian-sifive, line 11: `scala.reflect._` doesn't cross-compile with Scala 3, it needs to be put in `core/src/main/scala-2/chisel3/`

PR #4559, opened 2024-12-12
[svsim] Escape '$' in defines for VCS only

seldridge:
Add an abstract function to `svsim.Backend` which can be used to control how define key and values will be escaped.  Add escaping for '$' for VCS only.

This is done to work around what I think is a difference in how VCS and Verilator handle the `+define+key=value` command line argument.  In svsim, we call both like the following (this is a key only, no value):

    <tool> '+define+Foo$Bar'

While Verilator will accept this, VCS will not and instead wants:

    <tool> '+define+Foo\$Bar'

As far as I can tell, the problem seems to be the single quotes not being respected by VCS.  This is _likely_ due to the fact that the VCS entrypoint is shell and may have some oddities in how it is handling command line options.  (This is speculation.)

I expect that this could have manifested for anything that was using a Verilog system function, e.g.., `+define+RANDOM+$random`.  For that specific case, we were already escaping it internally.

The immediate need for this is to get inline layers and their `$`-ridden ABI (that I wrote...) working with VCS.  They are already working with Verilator.

#### Release Notes

Fix a bug where defines passed to VCS could be ignored by VCS if they included `$`.

### DIFF
--- a/svsim/src/main/scala/Backend.scala
+++ b/svsim/src/main/scala/Backend.scala
@@ -22,7 +22,7 @@ object CommonCompilationSettings {
     def apply(name: String) = new VerilogPreprocessorDefine(name, None)
   }
   case class VerilogPreprocessorDefine private (name: String, value: Option[String]) {
-    private[svsim] def toCommandlineArgument(backend: Backend): String = {
+    final def toCommandlineArgument(backend: Backend): String = {
       value match {
         case Some(v) => s"+define+${backend.escapeDefine(name)}=${backend.escapeDefine(v)}"
         case None    => s"+define+${backend.escapeDefine(name)}"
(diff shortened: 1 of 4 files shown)


svsim/src/main/scala/Backend.scala
```diff
@@ -22,10 +22,10 @@ object CommonCompilationSettings {
     def apply(name: String) = new VerilogPreprocessorDefine(name, None)
   }
   case class VerilogPreprocessorDefine private (name: String, value: Option[String]) {
-    private[svsim] def toCommandlineArgument: String = {
+    final def toCommandlineArgument(backend: Backend): String = {
```
jackkoenig, line 25: Did you mean to make this public?

jackkoenig (approved):
LGTM, just one question

svsim/src/main/scala/Backend.scala
seldridge: Yeah, there was an internal duplicate use of this. It seems less bad to make this public to avoid the duplication internally. I'll land this with this change in place and will re-privatize it once the internal usage is converted to ChiselSim.

PR #4587, opened 2025-01-03
Split up Decoupled.scala into multiple files, NFC

seldridge:
Break up Decoupled.scala into four files, each associated with its internal contents.  This makes it easier for developers to find Queue, Irrevocable, and ReadyValidIO.


jackkoenig (approved):
LGTM!

PR #4591, opened 2025-01-04
[scala3] Fix Scala3 difference in Probe API

seldridge:
The Probe.apply API is supposed to take a second argument of type `Layer`. However, this was incorrectly taking an argument of type `Option[Layer]`. Revert to the former.


adkian-sifive (approved):
Thanks!

PR #4605, opened 2025-01-09
[docs] Add Cookbook entry for Verilog case equality (IsX)

jackkoenig:
### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -909,7 +909,7 @@ Chisel does not support SystemVerilog `X` directly, but it is possible to check 
 import chisel3._
 import chisel3.util.circt.IsX
 
-class XSafeAssert extends Module {
+class AssertButAllowX extends Module {
   val in = IO(Input(UInt(8.W)))
 
   // Assert that in is never zero; also do not trigger assert in the presence of X.
@@ -919,7 +919,7 @@ class XSafeAssert extends Module {
 
 ```scala mdoc:invisible
 // Hidden but will make sure this actually compiles
-chisel3.docs.emitSystemVerilog(new XSafeAssert)
+chisel3.docs.emitSystemVerilog(new AssertButAllowX)
 ```
 
 ## Predictable Naming
--- a/docs/src/cookbooks/cookbook.md
+++ b/docs/src/cookbooks/cookbook.md
@@ -902,7 +902,7 @@ compile(new TooWideOrNarrowUInt(8, 4))
 Verilog has "case equality" (`===`) and inequality (`!==`) operators.
 They are typically used to ignore unknown (`X`) values in assertions.
 
-Chisel does not support SystemVerilog `X` directly, but it is possible to check if a value is `X` with `chisel3.util.circt.isX`.
+Chisel does not support Verilog `X` directly, but it is possible to check if a value is `X` with `chisel3.util.circt.isX`.
 `isX` is commonly used to guard assertions against `X` which gives similar behavior to Verilog case equality.
 
 ```scala mdoc:silent:reset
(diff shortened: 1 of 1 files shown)


docs/src/cookbooks/cookbook.md
```diff
@@ -897,6 +897,31 @@ compile(new TooWideOrNarrowUInt(8, 2))
...
+`isX` is commonly used to guard assertions against `X` which gives similar behavior to Verilog case equality.
+
+```scala mdoc:silent:reset
+import chisel3._
+import chisel3.util.circt.IsX
+
+class XSafeAssert extends Module {
+  val in = IO(Input(UInt(8.W)))
+
+  // Assert that in is never zero, but also guard against X
```
mwachs5, line 915: this comment is a little misleading, we aren't "guarding against X" we are "do not trigger assert in the presence of X"

docs/src/cookbooks/cookbook.md
```diff
@@ -897,6 +897,31 @@ compile(new TooWideOrNarrowUInt(8, 2))
...
+They are typically used to ignore unknown (`X`) values in assertions.
+
+Chisel does not support SystemVerilog `X` directly, but it is possible to check if a value is `X` with `chisel3.util.circt.isX`.
+`isX` is commonly used to guard assertions against `X` which gives similar behavior to Verilog case equality.
+
+```scala mdoc:silent:reset
+import chisel3._
+import chisel3.util.circt.IsX
+
+class XSafeAssert extends Module {
```
mwachs5, line 912:
```suggestion
class AssertButAllowX extends Module {
```

perhaps?

docs/src/cookbooks/cookbook.md
mwachs5, line 915:
```suggestion
  // Assert that `in` is never zero, but also allow X
```

jackkoenig, line 915:
```suggestion
  // Assert that in is never zero; also do not trigger assert in the presence of X.
```

docs/src/cookbooks/cookbook.md
```diff
@@ -897,6 +897,31 @@ compile(new TooWideOrNarrowUInt(8, 2))
...
+  val in = IO(Input(UInt(8.W)))
+
+  // Assert that in is never zero; also do not trigger assert in the presence of X.
+  assert(IsX(in) || in =/= 0.U, "in should never equal 0")
+}
+```
+
+```scala mdoc:invisible
+// Hidden but will make sure this actually compiles
+chisel3.docs.emitSystemVerilog(new XSafeAssert)
```
jackkoenig, line 922:
```suggestion
chisel3.docs.emitSystemVerilog(new AssertButAllowX)
```

docs/src/cookbooks/cookbook.md
jackkoenig: Sure!

docs/src/cookbooks/cookbook.md
jackkoenig: I phrased it differently, what do you think?

mwachs5 (approved):

docs/src/cookbooks/cookbook.md
```diff
@@ -897,6 +897,31 @@ compile(new TooWideOrNarrowUInt(8, 2))
 compile(new TooWideOrNarrowUInt(8, 4))
 ```
 
+## How can I use Verilog "case equality" operators in Chisel?
+
+Verilog has "case equality" (`===`) and inequality (`!==`) operators.
+They are typically used to ignore unknown (`X`) values in assertions.
+
+Chisel does not support SystemVerilog `X` directly, but it is possible to check if a value is `X` with `chisel3.util.circt.isX`.
```
tymcauley, line 905: Mega-nit, but `X` is referred to as a Verilog feature above, so this should probably be consistent with that.

```suggestion
Chisel does not support Verilog `X` directly, but it is possible to check if a value is `X` with `chisel3.util.circt.isX`.
```

jackkoenig: Agreed, thanks!

PR #4611, opened 2025-01-10
(draft) Change signature of macro-based `chisel3.Bits.unary_~` to fix IntelliJ highlighting issue

unkarjedy:
Hello, I am from the IntelliJ Scala Plugin team
We have an error report about `~` method not being resolved in IntelliJ with Scala Plugin:
https://youtrack.jetbrains.com/issue/SCL-15350/Operators-not-recognized-chisel3-library

The method is macro-based.
It declares to return `Bits`, but the macro substitutes it with `UInt`/ `Boolean` / `SInt`.
IntelliJ can't detect it because it knows nothing of the macro logic.

As I mentioned in the [comment](https://youtrack.jetbrains.com/issue/SCL-15350/Operators-not-recognized-chisel3-library#focus=Comments-27-11368838.0-0)
There is a chance that this particular issue could be partially fixed by improving the method return type to use abstract type `type Self <: Bits`.

JFTR I know nothing about chisel details
It's a draft proposal, more for a discussion initiation...

### DIFF
--- a/core/src/main/scala-2/chisel3/Bits.scala
+++ b/core/src/main/scala-2/chisel3/Bits.scala
@@ -37,6 +37,8 @@ private[chisel3] sealed trait ToBoolable extends Element {
   */
 sealed abstract class Bits(private[chisel3] val width: Width) extends BitsImpl with ToBoolable {
 
+  type Self <: Bits
+
   /** Tail operator
     *
     * @param n the number of bits to remove
@@ -170,7 +172,7 @@ sealed abstract class Bits(private[chisel3] val width: Width) extends BitsImpl w
     * @return this $coll with each bit inverted
     * @group Bitwise
     */
-  final def unary_~ : Bits = macro SourceInfoWhiteboxTransform.noArg
+  final def unary_~ : Self = macro SourceInfoWhiteboxTransform.noArg
 
   /** @group SourceInfoTransformMacro */
   def do_unary_~(implicit sourceInfo: SourceInfo): Bits = _impl_unary_~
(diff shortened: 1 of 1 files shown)


unkarjedy:
Same with `chisel3.Bits#<<` it seems

Emin017:
It works for me:
before
After fix:
after-fix

IntelliJ IDEA 2024.3.1.1 (Ultimate Edition)
Scala Plugin Version: 2024.3.605

jackkoenig:
Thanks for confirming @Emin017!

I'll try to get to applying this fix soon across the board.

sequencer:
Sorry for the delay response:

> I think @sequencer might use a plugin for his team but generally speaking it's just been kind of lingering.

However we only support fixing the D/I issue with https://github.com/chipsalliance/idealchisel/blob/master/idealchisel/src/InstanceInjector.scala
Personally, I am not inclined to modify the language for IDE support unless it does not impact the public API and passes the Mima check. In that case, I am entirely comfortable with the decision. Please leave this matter to Jack’s discretion.  

On the contrary, we should consider implementing a Chisel whitebox plugin that enables the dynamic modification of return types using macros (which should be attributed to the Scala 2 implicit function design). I concur that we need to enhance my plugin; it is currently on my to-do list, and I am hiring a Scala full-time developer to work on it.

unkarjedy:
If you have questions while developing IJ plugin, you are welcome to ask them in our discord:
https://discord.com/channels/931170831139217469/931171260824707072

PR #4614, opened 2025-01-14
[svsim] Fix initial ordering issue

### LINKED ISSUE #3962: svsim: Verilator `initial` statements are not handled correctly
**Type of issue**: Bug Report

Build a chisel module with a `Mem`, which have the option to be randomly initialized. Use svsim to simulate, and try using the random initialization by setting `+define+ENABLE_INITIAL_MEM_` and `+define+RANDOMIZE_MEM_INIT`.

A testbench is generated containing something like this:
testbench.sv:
```verilog
// Simulation
  import "DPI-C" context task simulation_body();
  initial begin
    simulation_body();
  end
```

However, all the (randomized) memory initialization also happens in initial blocks, e.g.
mymem.sv:
```verilog
 `ifdef ENABLE_INITIAL_MEM_
    reg [31:0] _RANDOM_MEM;
    initial begin	
      ...
```

The initial blocks are scheduled such that the one in the testbench is executed first, at least in my case. Since `simulation_body()` executes the entire simulation, the entire simulation runs before all the other `initial` blocks, therefore skipping any initialization.

The `simulation_body()` should run after all other initial blocks.

A quick fix, although not pretty, would be to wrap the `simulation_body()` call in `final begin .. end` instead of `initial begin .. end`, where it would be guaranteed to run after the initial blocks.

Chisel 6.2.0

**Other Information**

It's quite important to have the randomization feature working because verilator doesn't support X values. Running the entire simulation before the initial blocks (and therefore any randomization) will hide bugs.

seldridge:
Fix a bug in svsim where the entire simulation is run inside an `initial` block from a generated testbench.  This creates undefined simulation ordering with `initial` blocks in the simulated modules.  Fix this by switching to an explicit state machine inside the testbench.

Fixes #3962.

#### Release Notes

- Fix initial ordering bug in svsim.


svsim/src/test/resources/Initial.sv
```diff
@@ -0,0 +1,14 @@
...
+  reg a = 1'b0;
+
+  initial
+    a = 1'b1;
+
+  always @(posedge clock) begin
+    assert(a == 1'b1);
+  end
+
+endmodule
```
jackkoenig, line 14: I think a marginally better test would be to drive this value on an output and confirm it in the Scala testcode, but this is fine.

jackkoenig (approved):
LGTM but consider comment

svsim/src/test/resources/Initial.sv
seldridge: Nice suggestion. I switched to an output port that is checked in this commit: https://github.com/chipsalliance/chisel/pull/4614/commits/3fe2549be866264b25f92ee777ddb4034f619829

PR #4623, opened 2025-01-15
[core] Add layer block that returns colored wire

seldridge:
Change the `layer.block` API to return values.  If the value is a `Data`,
then create a wire of layer-colored probe type immediately before the
layer block.  Inside the layer block, define this value with a probe of
the return.  If the value is a non-`Data`, then return it directly.

Previously, this API would always return `Unit`.

Add a type class, to be used for layer blocks, that can be used to
post-process a layer block's return value.  Two type class generators are
provided: one that is an identity function and another that returns a
wire of layer-colored probe type.

This latter type class generator has the effect of creating code like the
following without requiring a pre-declared layer:

``` firrtl
wire a: Probe<UInt<1>>
layerblock A:
  wire b: UInt<1>
  define a = probe(b)
```

CC: @hcook 

This is stacked on https://github.com/chipsalliance/chisel/pull/4628.

#### Release Notes

Change `layer.block` API to allow for values to be returned. If a value is a `Data`, this will return a `Wire` of layer-colored probe type. If a non-`Data`, the value will be returned directly. Advanced users may override this behavior via a typeclass, `BlockReturnHandler`.

### DIFF
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -2,9 +2,11 @@
 
 package chisel3
 
-import chisel3.experimental.{SourceInfo, UnlocatableSourceInfo}
+import chisel3.Data.ProbeInfo
+import chisel3.experimental.{requireIsHardware, SourceInfo, UnlocatableSourceInfo}
 import chisel3.internal.{Builder, HasId}
 import chisel3.internal.firrtl.ir.{LayerBlock, Node}
+import chisel3.reflect.DataMirror
 import chisel3.util.simpleClassName
 import java.nio.file.{Path, Paths}
 import scala.annotation.tailrec
@@ -194,6 +196,111 @@ object layer {
       Builder.layers += layer
       currentLayer = parent
     }
+  }
+
+  /** A type class that describes how to post-process the return value from a layer block. */
+  sealed trait BlockReturnHandler[A] {
+
+    /** The type that will be returned by this handler. */
+    type R
+
+    /** Do not post-process the return value.  This is what should happen if the
+      * code early exits, i.e., if no layer block is created.
+      */
+    private[chisel3] def identity(result: A): R
+
+    /** Post-process the return value.
+      *
+      * @param placeholder the location above the layer block where commands can be inserted
+      * @param layerBlock the current layer block
+      * @param result the return value of the layer block
+      * @param sourceInfo source locator information
+      * @return the post-processed result
+      */
+    private[chisel3] def apply(placeholder: Placeholder, layerBlock: LayerBlock, result: A)(
+      implicit sourceInfo: SourceInfo
+    ): R
+  }
+
+  object BlockReturnHandler {
+
+    type Aux[A, B] = BlockReturnHandler[A] { type R = B }
+
+    /** Return a [[BlockReturnHandler]] that will always return [[Unit]]. */
+    def unit[A]: Aux[A, Unit] = new BlockReturnHandler[A] {
+
+      override type R = Unit
+
+      override private[chisel3] def identity(result: A): R = ()
+
+      override private[chisel3] def apply(
+        placeholder: Placeholder,
+        layerBlock:  LayerBlock,
+        result:      A
+      )(
+        implicit sourceInfo: SourceInfo
+      ): R = {
+        ()
+      }
+
+    }
+
+    /** Return a [[BlockReturnHandler]] that, if a layer block is created, will
+      * create a [[Wire]] of layer-colored probe type _above_ the layer block
+      * and [[probe.define]] this at the end of the layer block.  If no
+      * layer-block is created, then the result is pass-through.
+      */
+    implicit def layerColoredWire[A <: Data]: Aux[A, A] = new BlockReturnHandler[A] {
+
+      override type R = A
+
+      override private[chisel3] def identity(result: A): R = result
+
+      override private[chisel3] def apply(
+        placeholder: Placeholder,
+        layerBlock:  LayerBlock,
+        result:      A
+      )(
+        implicit sourceInfo: SourceInfo
+      ): R = {
+        // Don't allow returning non-hardware types.  This avoids problems like
+        // returning a probe type which we can't handle.
+        requireIsHardware(result)
+
+        // The result wire is either a new layer-colored probe wire or an
+        // existing layer-colored probe wire forwarded up.  If it is the former,
+        // the wire is colored with the current layer.  For the latter, the wire
+        // needs to have a compatible color.  If it has one, use it.  If it
+        // needs a color, set it on the wire.
+        val layerColoredWire = placeholder.append {
+          result.probeInfo match {
+            case None                     => Wire(probe.Probe(chiselTypeOf(result), layerBlock.layer))
+            case Some(ProbeInfo(_, None)) => Wire(probe.Probe(result.cloneType, layerBlock.layer))
+            case Some(ProbeInfo(_, Some(color))) =>
+              if (!layerBlock.layer.canWriteTo(color))
+                Builder.error(
+                  s"cannot return probe of color '${color.fullName}' from a layer block associated with layer '${layerBlock.layer.fullName}'"
+                )
+              Wire(chiselTypeOf(result))
+          }
+        }
+
+        // Similarly, if the result is a probe, then we forward it directly.  If
+        // it is a non-probe, then we need to probe it first.
+        Builder.forcedUserModule.withRegion(layerBlock.region) {
+          val source = DataMirror.hasProbeTypeModifier(result) match {
+            case true  => result
+            case false => probe.ProbeValue(result)
+          }
+          probe.define(layerColoredWire, source)
+        }
+
+        // Return the layer-colored wire _above_ the layer block.
+        layerColoredWire
+      }
+
+    }
+
   }
 
   /** Create a new layer block.  This is hardware that will be enabled
--- a/core/src/main/scala/chisel3/Layer.scala
+++ b/core/src/main/scala/chisel3/Layer.scala
@@ -311,6 +311,15 @@ object layer {
     * be the current layer which causes no layer block to be created.  (This is
     * not a _proper_ ancestor requirement.)
     *
+    * By default, the return of the layer block will be either a subtype of
+    * [[Data]] or [[Unit]], depending on the `thunk` provided.  If the `thunk`
+    * (the hardware that should be constructed inside the layer block) returns a
+    * [[Data]], then this will either return a [[Wire]] of layer-colored
+    * [[Probe]] type if a layer block was created or the underlying [[Data]] if
+    * no layer block was created.  If the `thunk` returns anything else, this
+    * will return [[Unit]].  This is controlled by the implicit argument `tc`
+    * and may be customized by advanced users to do other things.
+    *
     * @param layer the layer this block is associated with
     * @param skipIfAlreadyInBlock if true, then this will not create a layer if
     * this `block` is already inside another layerblock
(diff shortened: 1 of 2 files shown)


dtzSiFive (changes requested):
This has some problems.  The new construct isn't supported elsewhere (e.g., Select) and it's unclear what this "Region" should mean.

Marking "request changes" because it isn't suitable for use as-is, specifically the generated wire cannot be used.

seldridge:
Nice suggestion, Will.

I refactored, rebased, and force-pushed using an approach along the lines of your suggestion. This changes a `Block` to be a wrapper around a sequence of command buffers. The "append point" (the command buffer within a block to which commands will be written to) can then be changed with the standard thunk-based APIs for doing this safely.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -350,17 +354,32 @@ private[chisel3] object ir {
...
 
-    private def _closed: Boolean = _commandsBuilder == null
+    private def _closed: Boolean = _appendPoint == null
+
+    def newBuffer(): AppendPoint = {
+      val oldPoint = new AppendPoint(_appendPoint)
+      _appendPoint = ArraySeq.newBuilder[Command]
+      _commandBufferList += _appendPoint
+      oldPoint
+    }
```
jackkoenig, line 365: Discussed offline a bit that I like the direction and we could probably use this basic functionality in the future to enable certain APIs that need to declare wires outside of the current when-scope (recursively, effectively removing it from any when scoping to the top-level scope in the current module).

I do think we want to be mindful of the number and size of objects we're allocating. My hope is that we can make this "common path" of little to nothing being appended cheap and fast, while still retaining the flexibility you're adding.

I have a different idea. What if, instead of adding this complexity to Blocks, we make it such that whenever a layerblock would occur there are actually 2 Commands appended to the current commands rather than 1.
The first command is the "append point" of the previous block, the simplest implementation would be to use a Block that is empty and only contains anything if Commands are appended there. The 2nd command would then be the Block representing the Layer.

An optimization would be to turn that first command into an "EmptyCommand" that gets mutably swapped out with a Block only if appends occur. This would require making the contents of a Block an ArrayBuffer rather than an ArraySeq Builder, but that is probably a worthwhile tradeoff.

Does this make sense? Is there an obvious problem I'm missing?

seldridge: > Does this make sense? Is there an obvious problem I'm missing?

This was the original approach. That `Region` that was discussed above was a command that contained a `Block` where you could write commands later.

The problem with this is that it requires rethinking how scope checking works. Currently, a block is the same as a scope. Data is only visible to you if the current block (as viewed by the builder) is in an ancestor block. However, if we introduce a sibling block that you can write to, then that breaks this.

Hence, I went with this alternative approach which keeps the notion of a block being the same as a scope, but fracturing a block internally to be multiple command buffers.

We could add another command that contains a command buffer, but is not a block and hence not a scope.

WDYT?

seldridge:
This is now really work-in-progress. I've got an approach which switches to a `Placeholder` object wrapping a `Placeholder` IR object which duplicates some functionality of a `Block` (it has a command buffer), but it is _not_ a `Block` and will be subject to the same `Block` restrictions.

I need to dramatically clean up this patch to better incorporate the TypeClass approach that Jack worked out (shown in the `block2` API) which can be used to unify the `block` and `blockWithRegion` APIs into a single one. I'll clean this up tomorrow.

Thanks for all the great feedback. If you have any comments on the `Placeholder` approach, let me know. This will need to be updated to work with the Select APIs.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: Sorry I forgot to respond earlier but that makes sense, thank you for the explanation. 

> We could add another command that contains a command buffer, but is not a block and hence not a scope.

This matches my intuition as the best way to do it, if you could give it a try and see if it works I would appreciate it. Maybe it's a bad idea then we can revert back to this approach.

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -198,6 +199,77 @@ object layer {
...
+    /** Post-process the return value.
+      *
+      * @param placeholder the location above the layer block where commands can be inserted
+      * @param layerBlock the current layer block
+      * @param result the return value of the layer block
+      * @param sourceInfo source locator information
+      * @return the post-processed result
+      */
+    def apply(placeholder: Placeholder, layerBlock: LayerBlock, result: A)(implicit sourceInfo: SourceInfo): A
+  }
```
jackkoenig, line 214: Related to making my suggestion to make `Placeholder` private, doing so implies that this apply method should be [package] private which then implies that this trait should be sealed. I agree with all of these implications--this is extremely powerful stuff here and I'd rather keep it private and expose various things as little pieces rather than opening it all up.

FWIW I am moving Lookupable in this direction--it was a mistake to leave it unsealed so I've had to deprecate inheritance to fix that: https://github.com/chipsalliance/chisel/blob/7ba079dc51de726cfb65b7aeb3e07ad1b9dfae1c/core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala#L42. The solution for user extensibility lies in factory methods that let you implement the typeclass without having to be able to implement the virtual methods, see `Lookupable.productX` [[1]](https://github.com/chipsalliance/chisel/blob/7ba079dc51de726cfb65b7aeb3e07ad1b9dfae1c/core/src/main/scala/chisel3/experimental/hierarchy/core/Lookupable.scala#L116) and examples [[2]](https://www.chisel-lang.org/docs/cookbooks/hierarchy#how-do-i-make-case-classes-with-lots-of-fields-accessible-from-an-instance) for what I mean.

seldridge: This plan sounds good to me. I did the minimal thing of making this private in the prerequisite PR and then making the trait sealed and the `apply` method package private in https://github.com/chipsalliance/chisel/pull/4623/commits/e240e7d67ceba440ff33e69123362a289471173e. This needed a rebase, but the original commit order is unchanged.

seldridge:
@jackkoenig and @dtzSiFive: I changed this _back_ to return `(_ <: Data) | Unit` in some fixup commits. This uses a derivative of your patch, @jackkoenig.

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -196,6 +197,100 @@ object layer {
...
+        // keeps its existing layer color.
+        val layerColoredWire = placeholder.append {
+          result.probeInfo match {
+            case None            => Wire(probe.Probe(chiselTypeOf(result), layerBlock.layer))
+            case Some(probeInfo) => Wire(chiselTypeOf(result))
+          }
+        }
+
+        // Similarly, if the result is a probe, then we forward it directly.  If
+        // it is a non-probe, then we need to probe if first.
```
dtzSiFive, line 277:
```suggestion
        // it is a non-probe, then we need to probe it first.
```

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -196,6 +197,100 @@ object layer {
...
+        // the wire is colored with the current layer.  For the latter, the wire
+        // keeps its existing layer color.
+        val layerColoredWire = placeholder.append {
+          result.probeInfo match {
+            case None            => Wire(probe.Probe(chiselTypeOf(result), layerBlock.layer))
+            case Some(probeInfo) => Wire(chiselTypeOf(result))
+          }
+        }
+
+        // Similarly, if the result is a probe, then we forward it directly.  If
```
dtzSiFive, line 276: FWIW lack of probe-ness at the top level of an aggregate doesn't mean the interior doesn't have probes; similarly if they contain properties (or probes) they can't be probed.

The simplification made here (Data is either probe or not-probe and if not-probe makes sense to be probed) is one done in at least a few other places, so don't mean to snipe unduly here.  It may be suitable in practice?

I haven't put my solution hat on for this lately, so just FYI!

I'm not sure whether Chisel will presently catch this or not.  If not it probably should when constructing the `probe<T>`.  Could be useful to intercept and diagnose the problem here?

Lots of thoughts, leaving this short.

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -196,6 +197,100 @@ object layer {
...
+        // it is a non-probe, then we need to probe if first.
+        Builder.forcedUserModule.withRegion(layerBlock.region) {
```
dtzSiFive, line 278: I don't think it matters given the specific hardware created here (?), but just a note that stands out: this creates hardware assuming that changing the block is sufficient (with Placeholder and related builder-state management details fresh on our minds).  I think that's right but is important this only does things where having the rest of the state wrong is not an issue (e.g., layerStack).

core/src/main/scala/chisel3/Layer.scala
```diff
@@ -196,6 +197,100 @@ object layer {
...
+      ): R = {
+        ()
+      }
+
+    }
+
+    /** Return a [[BlockReturnHandler]] that, if a layer block is created, will
+      * create a [[Wire]] of layer-colored probe type _above_ the layer block
+      * and [[probe.define]] this at the end of the layer block.  If no
+      * layer-block is created, then the result is pass-through.
```
dtzSiFive, line 250: Cool!  This has some neat properties!

dtzSiFive (approved):
LGTM!

Can this break existing code that today "returns" something from a block (I am not sure offhand if terminating with a value will error or just be dropped today (pre-PR)) that might now be a problem?

Mostly re:handling of release flow/target version details.

dtzSiFive:
Up to you, but if this is derived from @jackkoenig's code consider including the co-authored-by information for attribution.

seldridge:
> Can this break existing code that today "returns" something from a block (I am not sure offhand if terminating with a value will error or just be dropped today (pre-PR)) that might now be a problem?

Good question. Not a problem given that the return value today is always `Unit`. If a user is trying to return something today, they can't do it through the return of the layer block and have to mutably set a variable, defined above the layer block, outside the layer block. This patch is a change to existing behavior, but not in a way which should break anything (though it is possible). E.g., this will break code that is doing a runtime check that the return type is a `Unit`:

```
val foo = layer.block(A) { Wire(Bool()) }
assert(foo.getClass() == classOf[Unit])
```

core/src/main/scala/chisel3/Layer.scala
seldridge: Fixed in: cbc862ea1

jackkoenig (approved):
LGTM! Nice work!

src/test/scala/chiselTests/LayerSpec.scala
```diff
@@ -306,6 +306,107 @@ class LayerSpec extends ChiselFlatSpec with Utils with FileCheck {
...
+          new Bundle {
+            val a = Probe(UInt(1.W), A)
+            val b = Probe(UInt(2.W), A.B)
+            val c = Probe(UInt(3.W), A.B)
+          }
+        }
+      }
+
+      // A single layer block
+      val a = layer.block(A) {
```
mwachs5, line 322: super cool! 

One thing, is it important to explain why this must be a read-only probe vs read-write? Is that a design choice or a requirement ? Is there potential future work like `layer.writableblock(A)` or is that fundamentally impossible or a "we dont need it yet so let's not do it yet"?

dtzSiFive:
> ```scala
> val foo = layer.block(A) { Wire(Bool()) }
> assert(foo.getClass() == classOf[Unit])
> ```

If the returned Data is not safe to create a probe out of, no `assert` needed.
I doubt this breaks anything currently out there, similarly to "but what if you checked the return type".

seldridge:
> If the returned Data is not safe to create a probe out of, no assert needed.

Yes, this is one place where this could break things. If the last statement is a `Data`, but is not legal to create a `Probe` type out of, this will cause a Chisel elaboration-time failure.

core/src/main/scala/chisel3/Layer.scala
seldridge: Yeah, this is a problem. I'm going to see about handling it in a follow-on.

A related problem is being able to return layer-colored `Data`. I improved this in the rebase to work more directly. See https://github.com/chipsalliance/chisel/pull/4623/files#diff-d85d136248243112b9d589c7160462146720114db92669778a6d2831ab73d3b0R343 for "up casting" from `Probe<UInt>` to `Probe<UInt, A>`. See https://github.com/chipsalliance/chisel/pull/4623/files#diff-d85d136248243112b9d589c7160462146720114db92669778a6d2831ab73d3b0R427 for checking of returning an illegal probe type and this being caught in Chisel.

For aggregates, we will need to do some deeper inspection and potentially do things like what is happening with the up-casting here on a fine-grained basis. CIRCT will have to act for the backstop for this for now...

src/test/scala/chiselTests/LayerSpec.scala
seldridge: We're not sure, yet, if we want to allow writes to layers. This patch is close enough to allowing this if we want to go that way, though. To do that, you would probably need to manually return a `RWProbe` (and this would need to be updated to keep the write flag). Something like:

``` scala
val a = layer.block(A) {
  Wire(probe.RWProbe(Bool()))
}
```

(This patch will let you do this, but you will get a read probe wire as the return. If you manually make this a RWProbe and try to drive it, CIRCT will have problems.)

Towards the problem of having data go up and down, I'm going to first see how far we can push the `Placeholder` API to allow for construction of hardware in the parent while constructing the child.

PR #4628, opened 2025-01-16
[core] Add Placeholder API

seldridge:
Add an advanced Chisel API which allows for generation of hardware at a remembered point in the design.  This API allows a user to create an object of `class Placeholder` which, if used later via its `append` method, will create hardware at the location where the `Placeholder` object was created.

This is intended to be used to support certain generator patterns where hardware must be created _before_ some other hardware, but only after the other hardware has been built.  E.g., this can be used to support creating and defining layer-colored probes after constructing a layer block.

### DIFF
--- a/core/src/main/scala/chisel3/Placeholder.scala
+++ b/core/src/main/scala/chisel3/Placeholder.scala
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.SourceInfo
+import chisel3.internal.Builder
+import chisel3.internal.firrtl.ir
+
+/** A [[Placeholder]] is an _advanced_ API for Chisel generators to generate
+  * additional hardware at a different point in the design.
+  *
+  * For example, this can be used to add a wire _before_ another wire:
+  * {{{
+  * val placeholder = new Placeholder()
+  * val a = Wire(Bool())
+  * val b = placeholder.append {
+  *   Wire(Bool())
+  * }
+  * }}}
+  *
+  * This will generate the following FIRRTL where `b` is declared _before_ `a`:
+  * {{{
+  * wire b : UInt<1>
+  * wire a : UInt<1>
+  * }}}
+  */
+private[chisel3] class Placeholder()(implicit sourceInfo: SourceInfo) {
+
+  private val state = Builder.State.save
+
+  private val placeholder = Builder.pushCommand(new ir.Placeholder(sourceInfo = sourceInfo))
+
+  /** Generate the hardware of the `thunk` and append the result at the point
+    * where this [[Placeholder]] was created.  The return value will be what the
+    * `thunk` returns which enables getting a reference to the generated
+    * hardware.
+    *
+    * @param thunk the hardware to generate and append to the [[Placeholder]]
+    * @return the return value of the `thunk`
+    */
+  def append[A](thunk: => A): A = Builder.State.guard(state) {
+    Builder.currentBlock.get.appendToPlaceholder(placeholder)(thunk)
+  }
+
+}
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -1199,6 +1199,85 @@ private[chisel3] object Builder extends LazyLogging {
   }
 
   initializeSingletons()
+
+  /** The representation of the state of the [[Builder]] at a current point in
+    * time.  This is intended to capture _enough_ information to insert hardware
+    * at another point in the circuit.
+    *
+    * @see [[chisel3.Placeholder]]
+    */
+  case class State(
+    currentModule: Option[BaseModule],
+    whenStack:     List[WhenContext],
+    blockStack:    List[Block],
+    layerStack:    List[layer.Layer],
+    prefix:        Prefix,
+    clock:         Option[Delayed[Clock]],
+    reset:         Option[Delayed[Reset]],
+    enabledLayers: List[layer.Layer]
+  )
+
+  object State {
+
+    /** Return a [[State]] suitable for doing module construction.
+      */
+    def default: State = State(
+      currentModule = Builder.currentModule,
+      whenStack = Nil,
+      blockStack = Builder.blockStack,
+      layerStack = layer.Layer.Root :: Nil,
+      prefix = Nil,
+      clock = None,
+      reset = None,
+      enabledLayers = Nil
+    )
+
+    /** Capture the current [[Builder]] state.
+      */
+    def save: State = {
+      State(
+        currentModule = Builder.currentModule,
+        whenStack = Builder.whenStack,
+        blockStack = Builder.blockStack,
+        layerStack = Builder.layerStack,
+        prefix = Builder.getPrefix,
+        clock = Builder.currentClockDelayed,
+        reset = Builder.currentResetDelayed,
+        enabledLayers = Builder.enabledLayers.toList
+      )
+    }
+
+    /** Set the [[Builder]] state to that provided by the parameter.
+      *
+      * @param state the state to set the [[Builder]] to
+      */
+    def restore(state: State) = {
+      Builder.currentModule = state.currentModule
+      Builder.whenStack = state.whenStack
+      Builder.blockStack = state.blockStack
+      Builder.layerStack = state.layerStack
+      Builder.setPrefix(state.prefix)
+      Builder.currentClock = state.clock
+      Builder.currentReset = state.reset
+      Builder.enabledLayers = enabledLayers
+    }
+
+    /** Run the `thunk` with the context provided by `state`.  Save the [[Builder]]
+      * state before the thunk and restore it afterwards.
+      *
+      * @param state change the [[Builder]] to this state when running the thunk
+      * @param thunk some hardware to generate
+      * @return whatever the `thunk` returns
+      */
+    def guard[A](state: State)(thunk: => A): A = {
+      val old = save
+      restore(state)
+      val result = thunk
+      restore(old)
+      result
+    }
+
+  }
 }
 
 /** Allows public access to the naming stack in Builder / DynamicContext, and handles invocations
--- a/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Converter.scala
@@ -2,7 +2,7 @@
 
 package chisel3.internal.firrtl
 
-import chisel3._
+import chisel3.{Placeholder => _, _}
 import chisel3.experimental._
 import chisel3.experimental.{NoSourceInfo, SourceInfo, SourceLine, UnlocatableSourceInfo}
 import chisel3.properties.Property
@@ -266,6 +266,8 @@ private[chisel3] object Converter {
       )
     case LayerBlock(info, layer, region) =>
       fir.LayerBlock(convert(info), layer, convert(region, ctx, typeAliases))
+    case Placeholder(info, block) =>
+      convert(block, ctx, typeAliases)
   }
 
   /** Convert Chisel IR Commands into FIRRTL Statements
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -337,6 +337,39 @@ private[chisel3] object ir {
   ) extends Definition
   case class DefObject(sourceInfo: SourceInfo, id: HasId, className: String) extends Definition
 
+  class Placeholder(val sourceInfo: SourceInfo) extends Command {
+    private var commandBuilder: mutable.Builder[Command, ArraySeq[Command]] = null
+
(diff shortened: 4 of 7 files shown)


core/src/main/scala/chisel3/Placeholder.scala
```diff
@@ -0,0 +1,43 @@
...
+  * }
+  * }}}
+  *
+  * This will generate the following FIRRTL where `b` is declared _before_ `a`:
+  * {{{
+  * wire b : UInt<1>
+  * wire a : UInt<1>
+  * }}}
+  */
+class Placeholder()(implicit sourceInfo: SourceInfo) {
```
jackkoenig, line 27: Mentioned to you offline, but I would rather this be a private API, at least initially. I admit I do have use cases for this in private code using Chisel, but I would rather box this up and put a bow on it as some sort of "Chisel Lazy Val" API rather than a mutable box that you can keep attaching things to.

seldridge: Made private in the rebase. Good thinking. If we ever do want such a footgun of an API made public, we can first figure out how to do use-before-def checking with changes to the Chisel Builder (or its evolved Scala3 form).

core/src/main/scala/chisel3/Placeholder.scala
```diff
@@ -0,0 +1,43 @@
...
+  /** Generate the hardware of the `thunk` and append the result at the point
+    * where this [[Placeholder]] was created.  The return value will be what the
+    * `thunk` returns which enables getting a reference to the generated
+    * hardware.
+    *
+    * @param thunk the hardware to generate and append to the [[Placeholder]]
+    * @return the return value of the `thunk`
+    */
+  def append[A](thunk: => A): A = {
+    Builder.currentBlock.get.appendToPlaceholder(placeholder)(thunk)
```
dtzSiFive, line 40: Did you mean `currentBlock`? Probably the expected behavior is that hardware generated in the placeholder belongs to the block containing the placeholder, not the current block we're appending from...?
(offhand not sure where to grab the right block, or I'd offer a concrete suggestion here)

Also, things like layerStack and whenStack will be wrong when evaluating `thunk`, probably other context bits maintained during top-down evaluation and not captured in the Command list s.t. you can jump around and insert this way.

For the when/layer stacks, these ideally would be replaced with walking up blocks to their parents (this was discussed/planned but postponed for reasons I don't have paged in).  Details like `elideLayerBlocks` are not captured, just tracked as evaluation context.  I haven't audited them all to separate those that maybe make sense as context regardless of insertion point, but offhand probably none of them (unless they'd always the same regardless...)?

That is, we're still mostly reliant on "context" state.

To help with the stacks, could require this is invoked from same block as that containing the placeholder -- but that doesn't address the other context details.

Basically, we don't have the infrastructure to do this safely.

What we have barely allows us to insert "secret connections".

Since it's now "private" (thanks, I suggested this as well...) and is probably fine for the use case envisioned for it, if we want to proceed anyway I won't stand in the way.  I'd prefer we sorted this all out instead.

dtzSiFive: Consider this test you can add to `PlaceholderSpec.scala`:

```
  they should "do something fun when used across modules" in {

    object A extends layer.Layer(layer.LayerConfig.Extract())

    class Bar extends RawModule {
      val placeholder = new Placeholder()
    }
    class Foo extends RawModule {
      val in = IO(Input(Bool()))

      val bar = Module(new Bar)
      val x = bar.placeholder.append({ val w = Wire(UInt(1.W)); w := 0.U; w })
      println(x)
      x := 1.U
    }

    generateFirrtlAndFileCheck(new Foo) {
      s"""|CHECK:      wire
          |CHECK-NEXT: asdf
          |""".stripMargin
    }

  }
```

Which generates a connection to `x` from a different module.  This is because of the issue mentioned above where the wrong block is used.

Output from print:
```
[372] circuit Foo :
[372]   layer Verification, bind, "verification" :
[372]     layer Assert, bind, "verification/assert" :
[372]     layer Assume, bind, "verification/assume" :
[372]     layer Cover, bind, "verification/cover" :
[372]   module Bar : @[src/test/scala/chisel3/PlaceholderSpec.scala 108:11]
[372]
[372]     wire x : UInt<1> @[src/test/scala/chisel3/PlaceholderSpec.scala 115:52]
[372]     connect x, UInt<1>(0h0) @[src/test/scala/chisel3/PlaceholderSpec.scala 115:67]
[372]
[372]   public module Foo : @[src/test/scala/chisel3/PlaceholderSpec.scala 111:11]
[372]     input in : UInt<1> @[src/test/scala/chisel3/PlaceholderSpec.scala 112:18]
[372]
[372]     inst bar of Bar @[src/test/scala/chisel3/PlaceholderSpec.scala 114:23]
[372]     connect x, UInt<1>(0h1) @[src/test/scala/chisel3/PlaceholderSpec.scala 117:9]
```

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -339,6 +339,27 @@ private[chisel3] object ir {
       extends Definition
   case class DefObject(sourceInfo: SourceInfo, id: HasId, className: String) extends Definition
 
+  class Placeholder(val sourceInfo: SourceInfo) extends Command {
+    private var buffer: mutable.Builder[Command, ArraySeq[Command]] = null
```
dtzSiFive, line 343: I don't know offhand, can you comment about this in comparison to what's used for the buffer in Block, for example?  Is this same but better/more explicit, or different functionally in an important way?
(`ArraySeq.newBuilder[Command]`)

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -357,6 +378,14 @@ private[chisel3] object ir {
       _commandsBuilder += c
     }
 
+    def appendToPlaceholder[A](placeholder: Placeholder)(thunk: => A): A = {
+      val oldPoint = _commandsBuilder
```
dtzSiFive, line 382: This looks a little tricksy, but is probably not likely to break presently?

I'm worried about interaction with this and use of `withRegion` where we might insert into a placeholder (because we've mutated the block to point to its buffer) instead of the intended normal block buffer.  Or anything that closes the block, so on.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
seldridge: This is just the type that is returned by `ArraySeq.newBuilder[Command]` [[1]]. The PR is trying to be clever here and initialize this to `null` to lazily initialize this. (Maybe a `lazy val` would be better than manually doing this lazily.) In order to do that, the PR needs to set the type explicitly while also making it `null`.

[1]: https://www.scala-lang.org/api/2.13.x/scala/collection/immutable/ArraySeq$.html#newBuilder[A](implicitevidence$2:scala.reflect.ClassTag[A]):scala.collection.mutable.Builder[A,scala.collection.immutable.ArraySeq[A]]

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -339,6 +339,27 @@ private[chisel3] object ir {
...
+
+  object Placeholder {
+    def unapply(placeholder: Placeholder): Option[(SourceInfo, Seq[Command])] = {
+      Some(
+        (placeholder.sourceInfo, placeholder.getBuffer.result())
+      )
+    }
+  }
+
+  class AppendPoint private[ir] (private[ir] val buffer: mutable.Builder[Command, ArraySeq[Command]])
```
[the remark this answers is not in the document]
seldridge: Nice catch. Removed in f8a6f459f.

core/src/main/scala/chisel3/internal/firrtl/Converter.scala
```diff
@@ -266,6 +266,8 @@ private[chisel3] object Converter {
       )
     case LayerBlock(info, layer, region) =>
       fir.LayerBlock(convert(info), layer, convert(region, ctx, typeAliases))
+    case Placeholder(info, block) =>
```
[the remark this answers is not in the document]
seldridge: You're talking about `convertLayer(layer: Layer): fir.Layer`?

That shouldn't need this as that is used to map over `childer: Seq[Layer]`. I.e., it's dealing with layers as opposed to layer blocks. If we ever added a notion of a placeholder for inserting layers later using this mechanism, we'd need it. (You can already insert layers by manually specifying their parent, so I don't think we would need/want this!)

Does that address your concern?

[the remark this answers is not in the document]
dtzSiFive: My mistake, apologies.  I misunderstood/misremembered.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -339,6 +339,25 @@ private[chisel3] object ir {
...
+
+  }
+
+  object Placeholder {
+    def unapply(placeholder: Placeholder): Option[(SourceInfo, Seq[Command])] = {
+      Some(
+        (placeholder.sourceInfo, placeholder.getBuffer.result())
+      )
+    }
+  }
```
jackkoenig, line 359: Do we need to check that we don't accidentally append to the buffer after calling result?
We probably also need to memoize the result since: `After calling result() the behavior of a Builder is undefined. No further methods should be called.` [[1]](https://www.scala-lang.org/api/2.13.16/scala/collection/mutable/Builder.html). It's not clear that you can even call `result()` multiple times.

jackkoenig (approved):
Please see comment about potentially protecting the `builder` better, but otherwise I think this is reasonable for making forward progress. I agree with Will about the longer term direction.

core/src/main/scala/chisel3/Placeholder.scala
seldridge: Yeah, this is a problem. Here's what's changed to address this:

1. https://github.com/chipsalliance/chisel/pull/4628/commits/4dd8f19d58410d82c6743a9df0e4f34df37a87eb adds the ability to save and restore the `Builder`'s state. This is a bit of a misnomer as it is really "state related to blocks, modules, layers, and whens", i.e., enough to do things like change the append point. You don't want to actually save the entire builder state as this includes global things.
2. The `Placeholder.append` method is updated to save/change/restore the state using a method called `guard` that takes a thunk. This should then make this approach safe.

I added two tests that show that this does actually work (allows you to construct hardware in the parent) and will error when used improperly (disallows construction of hardware in a child, i.e., a closed module). I do think that it is reasonable to relax the closed module restriction (as this is essentially a pathway to enable a cleaner `BoringUtils` as well as a Chisel-native provider API).

WDYT?

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1201,6 +1201,50 @@ private[chisel3] object Builder extends LazyLogging {
...
+      * @return whatever the `thunk` returns
+      */
+    def guard[A](state: State)(thunk: => A): A = {
+      val old = save
+      restore(state)
+      val result = thunk
+      restore(old)
+      result
+    }
+
```
jackkoenig, line 1246: Can/Should this be used in other places? Like `Module.apply`?

core/src/main/scala/chisel3/Placeholder.scala
jackkoenig: Looks reasonable enough to me, my approval remains

core/src/main/scala/chisel3/internal/Builder.scala
seldridge: Yes, probably. This is also highlighting that there is _more_ that should be saved here.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
seldridge: This is beefed up to work similarly to `Block` in https://github.com/chipsalliance/chisel/pull/4628/commits/1bc48a5828b60d417f6c8a01a87095d69ac5c10b. Thanks for the pointer to the docs. It's somewhat surprising that this is just undefined as opposed to being "read once" as I had thought.

core/src/main/scala/chisel3/internal/Builder.scala
seldridge: Much more is now saved with: https://github.com/chipsalliance/chisel/pull/4628/commits/50805afb21b536af6867e532705c44867841eeae

A follow-on PR showing this being used for `Module.apply` is here: https://github.com/chipsalliance/chisel/pull/4641

I expect that this can also be used to replace the related `Definition.apply`. I haven't looked into this, yet.

core/src/main/scala/chisel3/Placeholder.scala
seldridge: I'm looking at this specific code again. I may merge this and then loop back to change it.

The problem right now is that the `Builder` expect that it is always working with a `Block` when, after this patch, it could be inserting into a `Block` or a `Placeholder`. I think we probably want this to revert to always pointing at a command builder directly.

PR #4629, opened 2025-01-16
Add simple API for generating testharnesses inline

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Add an API to generate testharnesses inline that are emitted as additional public modules in the output.

- [ ] Did you review?


src/main/scala/chisel3/experimental/HasTests.scala
```diff
@@ -0,0 +1,71 @@
...
+trait HasTestsWithResult[TestResult] extends ElaboratesParents { module: RawModule =>
+
+  /** Given a [[Definition]] of this module, and a test body that takes an [[Instance]],
+    *  generate a testharness module that implements the test.
+    *
+    *  The default behavior is a module with the following features:
+    *
+    *  - A clock port
+    *  - A reset port with this module's reset type, or synchronous if unspecified
+    *  - The [[desiredName]] is "[[this.desiredName]] _ [[name]]".
```
tmckay-sifive, line 33:
```suggestion
    *  - The [[desiredName]] is "[[this.desiredName]] _ [[testName]]".
```

src/main/scala/chisel3/experimental/HasTests.scala
```diff
@@ -0,0 +1,71 @@
...
+        module.resetType match {
+          case t @ Module.ResetType.Asynchronous => t
+          case t @ Module.ResetType.Synchronous  => t
+          case _                                 => Module.ResetType.Synchronous
+        }
+      case _ => Module.ResetType.Synchronous
+    }
+    override val desiredName = s"${module.desiredName}_${testName}"
+    val dut = Instance(definition)
+    body(dut.asInstanceOf[Instance[module.type]])
```
tmckay-sifive, line 51: Anyone know how to get rid of this `asInstanceOf`? Normally I wouldn't mind, but users are going to have the same issue when overriding `generateTestHarness`.

```
[error]  found   : chisel3.experimental.hierarchy.core.Instance[chisel3.experimental.HasTestsWithResult[TestResult] with chisel3.RawModule]
[error]  required: chisel3.experimental.hierarchy.Instance[HasTestsWithResult.this.type]
[error]     (which expands to)  chisel3.experimental.hierarchy.core.Instance[HasTestsWithResult.this.type]
[error] Note: chisel3.experimental.HasTestsWithResult[TestResult] with chisel3.RawModule >: HasTestsWithResult.this.type, but class Instance is invariant in type P.
[error] You may wish to define P as -P instead. (SLS 4.5)
[error]     body(dut)
[error]          ^
[error] one error found
```

src/main/scala/chisel3/experimental/HasTests.scala
```diff
@@ -0,0 +1,71 @@
...
+  private lazy val moduleDefinition =
+    module.toDefinition.asInstanceOf[Definition[module.type]]
+
+  /** Generate an additional parent around this module.
+    *
+    *  @param parent generator function, should instantiate the [[Definition]]
+    */
+  def elaborateParentModule(parent: Definition[module.type] => RawModule with Public): Unit =
+    afterModuleBuilt { Definition(parent(moduleDefinition)) }
+}
```
seldridge, line 17: Can this trait be inlined into its single extension trait, `HasTestWithResult`?

Alternatively, this could be flattened in the other direction by making this a method on `RawModule`.

Pre-emptively creating inheritance hierarchies is usually an anti-pattern. (Or: more strongly, inheritance hierarchies are usually an anti-pattern _unless they are architected and/or have some theoretical basis like Scala collections or Cats [[1]].)

[1]: https://typelevel.org/cats/typeclasses.html#type-classes-in-cats

src/main/scala/chisel3/experimental/HasTests.scala
```diff
@@ -0,0 +1,71 @@
...
+    *
+    *  - A clock port
+    *  - A reset port with this module's reset type, or synchronous if unspecified
+    *  - The [[desiredName]] is "[[this.desiredName]] _ [[name]]".
+    */
+  protected def generateTestHarness(
+    testName:   String,
+    definition: Definition[module.type],
+    body:       Instance[module.type] => TestResult
+  ): RawModule with Public = new Module with Public {
```
seldridge, line 39: A `Module with Public` being immediately cast to a `RawModule with Public` seems odd.

seldridge (commented):
This took me a while to realize, but this is actually doing a separation of the "test harness" (the `body` function) from the actual "test". This three-part factoring of (1) DUT, (2) test harness, and (3) test makes sense to me.

With this, have you thought about how to share test harnesses as opposed to having them implicitly inlined in the test? Put differently, is there a way to make the test harness also a module and thereby get the benefits of using D/I on the test harness as opposed to just the DUT?

tmckay-sifive:
> With this, have you thought about how to share test harnesses as opposed to having them implicitly inlined in the test? Put differently, is there a way to make the test harness also a module and thereby get the benefits of using D/I on the test harness as opposed to just the DUT?

Good idea, I will look into that.

src/main/scala/chisel3/experimental/HasTests.scala
tmckay-sifive: The intent here is to allow the implementer to override it with something that generates a RawModule.

src/main/scala/chisel3/experimental/HasTests.scala
tmckay-sifive: Yeah it could be flattened. The use case I had in mind that lead to this separation was that the user might have a component with a bunch of tests, while still wanting to elaborate a one-off parent module that is not the same as the testharnesses for all the tests.

So, I wanted to have an open-ended `elaborateParentModule` available to the user for that purpose. But, it felt strange to have that in `HasTests`, since it's really a separate feature from unit tests; granted one that we can leverage to implement unit tests.

That's how I got here. That said, I'm not very opinionated on the organization. So, if you think it's better Scala or more Chisel-like to flatten these traits, I will happily do so.

tmckay-sifive:
> With this, have you thought about how to share test harnesses as opposed to having them implicitly inlined in the test? Put differently, is there a way to make the test harness also a module and thereby get the benefits of using D/I on the test harness as opposed to just the DUT?

I think this would require the ability to make a Definition of the testharness, with a hole in the middle for the test body. Is this possible with Definition/Instance? I could make the testharness and test-body/DUT sibling modules to accomplish something similar, but I think it's important that the DUT instance and test RTL live underneath the testharness hierarchy.

src/main/scala/chisel3/experimental/HasTests.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import chisel3._
+import chisel3.experimental.hierarchy.{Definition, Instance}
+
+/** Provides methods to build unit testharnesses inline after this module is elaborated.
+  *
+  *  @tparam TestResult the type returned from each test body generator, typically
+  *  hardware indicating completion and/or exit code to the testharness.
+  */
+trait HasTests[TestResult] { module: RawModule =>
```
tmckay-sifive, line 11: One thing I don't like about this organization is that if `TestResult` is anything except `Unit`, you need to `override def generateTestHarness` to build a testharness that uses that `TestResult`.

I don't like that they are linked in this way, so I'm going to try to clean this up.

jackkoenig:
I have feedback similar to Schuyler's. I think it's really important that we separate the definitions of TestHarnesses from the body of modules that need to be tested. Related to this, I think the type of the TestHarness (and of the TestResult) should not be set for a given module. It seems likely that a single module would want to have tests that have different testharnesses and results. Really, the type of the result should probably be tied to the type of the TestHarness, I don't think they're independent parameters.

I think we can do this with typeclasses, but perhaps we should sit down together to work through it (hard to do in a quick sketch).

A second order concern (and non-blocking)--I think it is useful to be able to create little tests inline so no issues there, but it should not be the **only** way. We need to think about _test composition_ where I could define a testbench, write a test against that interfaces, and then have multiple different modules use that test for themselves.

Another non-blocking concern, it should be easy to include or elide tests written this way. I should be able to elaborate a module and not also be required to elaborate all of the tests. That can come later though.

src/main/scala/chisel3/experimental/InlineTest.scala
```diff
@@ -0,0 +1,83 @@
+package chisel3.experimental
+
+import chisel3._
+import chisel3.experimental.hierarchy.{Definition, Instance}
+
+package object inlinetest {
```
jackkoenig, line 6: Style note, `package objects` should only be used for things that must be in an `object` but you want to be accessible within what is otherwise a package. Every top-level definition here is a regular `class`, `trait` or `object`, so there's no reason to use a package object.

Also, the file structure needs to match the package structure.

So basically, please move `inlinetest` up to the `package`, and put this code in `src/main/scala/chisel3/experimental/inlinetest/...`

tmckay-sifive: Gotcha, thanks. I'm not 100% I did what you asked, but I gave it a shot in https://github.com/chipsalliance/chisel/pull/4629/commits/5e5bdd4bf2820a2140b0e5c1f350d91d3bed4b85

src/main/scala/chisel3/experimental/inlinetest/package.scala
```diff
@@ -0,0 +1,80 @@
+package chisel3.experimental.inlinetest
```
jackkoenig, line 1: Nit, the name of this file shouldn't be `package.scala`, that is reserved for package objects, e.g. [[1]](https://github.com/chipsalliance/chisel/blob/1ac76764f2cfd2d37f5f25e4f9b42da665ffd5f0/core/src/main/scala-2/chisel3/probe/package.scala#L10).

I'm not sure what to call this file since it has basically everything in the package in it, maybe just `InlineTest.scala` (still in directory `src/main/scala/chisel3/experimental/inlinetest/`) and we can break it up later if we see fit?

src/main/scala/chisel3/experimental/inlinetest/package.scala
```diff
@@ -0,0 +1,80 @@
...
+/** An implementation of a testharness generator.
+  *
+  *  @tparam M the type of the DUT module
+  *  @tparam R the type of the result returned by the test body
+  */
+trait TestHarness[M <: RawModule, R] {
+
+  /** Generate a testharness module given the test parameters. */
+  def generate(test: TestParameters[M, R]): RawModule with Public
+}
```
jackkoenig, line 32: Is the type of `M` every really used? Or put another way--what is the use case for a TestHarness that only works for a particular kind of DUT Module? I suspect there is one, but maybe it would be good to have an example in the tests to drive the point home.

src/main/scala/chisel3/experimental/inlinetest/package.scala
```diff
@@ -0,0 +1,80 @@
...
+case class TestParameters[M <: RawModule, R](
+  /** The [[desiredName]] of the DUT module. */
+  dutName: String,
+  /** The user-provided name of the test. */
+  testName: String,
+  /** A Definition of the DUT module. */
+  dutDefinition: Definition[M],
+  /** The body for this test, returns a result. */
+  body: Instance[M] => R
+)
```
jackkoenig, line 21: This is an annoying detail but one we have to consider as library writers in Chisel.

What are the odds that we are going to have to evolve this type over time, like adding fields? I'm assuming it's high--if so, we should not make this a case class because case classes are nightmares for both source and binary compatibility.

Also, should users be able to create instances of this object or not? Maybe the constructor, factory apply, and copy methods should all be private (we won't have the latter two if we make this not a case class).

src/main/scala/chisel3/experimental/inlinetest/package.scala
```diff
@@ -0,0 +1,80 @@
...
+
+/** An implementation of a testharness generator.
+  *
+  *  @tparam M the type of the DUT module
+  *  @tparam R the type of the result returned by the test body
+  */
+trait TestHarness[M <: RawModule, R] {
```
jackkoenig, line 28: Another nit but something to think about. Neither `UnitTestHarness` nor `TestHarnessWithResultIO` extend this trait which leads me to believe `TestHarness` is the wrong name for this TestHarness-providing-typeclass. I am not sure of the right name though. `TestHarnessGenerator`? @seldridge any ideas?

jackkoenig (commented):
Generally this looks great, a few points to consider.

src/main/scala/chisel3/experimental/inlinetest/package.scala
tmckay-sifive: > What are the odds that we are going to have to evolve this type over time, like adding fields? I'm assuming it's high--

Yeah, I would expect we will at some point.

> if so, we should not make this a case class because case classes are nightmares for both source and binary compatibility.

Good to know. I made it a case class precisely because I wanted to avoid changes to the parameter list of the methods that take them.

I've changed it to a class for now.

src/main/scala/chisel3/experimental/inlinetest/package.scala
tmckay-sifive: Yeah I think `TestHarnessGenerator` makes more sense. I've changed it to that pending other opinions.

src/main/scala/chisel3/experimental/inlinetest/package.scala
tmckay-sifive: Yep I think it's useful to have when you want to reuse a testharness across DUTs where your testharness expects certain interfaces to be present on the DUT. Added an example.

tmckay-sifive:
Incidentally, the extra logic I added to the tests now either produces invalid FIR or hits a bug in firtool.

```
FIRRTL version 4.2.0
circuit test_ModuleWithTests_with_monitor :
  layer Verification, bind, "verification" :
    layer Assert, bind, "verification/assert" :
    layer Assume, bind, "verification/assume" :
    layer Cover, bind, "verification/cover" :

  module ModuleWithTests :
    input clock : Clock
    input reset : UInt<1>
    output io : { flip in : UInt<32>, out : UInt<32>}
    output monProbe : Probe<{ in : UInt<32>, out : UInt<32>}>

    define monProbe = probe(io)
    connect io.out, io.in

  module ProtocolMonitor :
    input clock : Clock
    input reset : Reset
    input io : { in : UInt<32>, out : UInt<32>}

    node _T = eq(io.in, io.out)
    node _T_1 = asUInt(reset)
    node _T_2 = eq(_T_1, UInt<1>(0h0))
    layerblock Verification :
      layerblock Assert :
        intrinsic(circt_chisel_ifelsefatal<format = "Assertion failed: in === out\n", label = "chisel3_builtin">, clock, _T, _T_2)

  public module test_ModuleWithTests_with_monitor :
    input clock : Clock
    input reset : UInt<1>

    inst dut of ModuleWithTests
    connect dut.clock, clock
    connect dut.reset, reset
    connect dut.io.in, UInt<32>(0h5)
    node _T = neq(dut.io.out, UInt<1>(0h0))
    node _T_1 = eq(reset, UInt<1>(0h0))
    layerblock Verification :
      layerblock Assert :
        intrinsic(circt_chisel_ifelsefatal<format = "Assertion failed at InlineTestSpec.scala:110\n", label = "chisel3_builtin">, clock, _T, _T_1)

    inst monitor of ProtocolMonitor
    connect monitor.clock, clock
    connect monitor.reset, reset
    connect monitor.io.out, read(dut.monProbe).out
    connect monitor.io.in, read(dut.monProbe).in
```

```
  circt.stage.phases.Exceptions$FirtoolNonZeroExitCode: /sifive/tools/llvm/circt/1.103.0/bin/firtool returned a non-zero exit code. Note that this version of Chisel (7.0.0-M2+306-941cc4b7-SNAPSHOT) was published against firtool version 1.100.0.
ExitCode:
1
STDOUT:

STDERR:
src/test/scala/chisel3/experimental/InlineTestSpec.scala:20:23: error: 'firrtl.instance' op has a wrong number of results; expected 2 but got 4
    val dut = Instance(test.dutDefinition)
                      ^
src/test/scala/chisel3/experimental/InlineTestSpec.scala:20:23: note: see current operation: %5:4 = "firrtl.instance"() <{annotations = [], layers = [], moduleName = @ModuleWithTests, name = "dut", nameKind = #firrtl<name_kind droppable_name>, portAnnotations = [[], [], [], []], portDirections = array<i1: false, true, true, true>, portNames = ["io_in", "io_out", "monProbe_in", "monProbe_out"]}> : () -> (!firrtl.uint<32>, !firrtl.uint<32>, !firrtl.probe<uint<32>>, !firrtl.probe<uint<32>>)
src/test/scala/chisel3/experimental/InlineTestSpec.scala:69:2: note: original module declared here
@instantiable
 ^
```

Not sure which yet, need to look into it.

seldridge:
@tmckay-sifive: FYI, this is fixed in CIRCT 1.104.0 and I'm seeing that test do something on my end.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  /** Generate a testharness module given the test parameters. */
+  def generate(test: TestParameters[M, R]): RawModule with Public
+}
+
+object TestHarnessGenerator {
+
+  /** The minimal implementation of a unit testharness. Has a clock input and a synchronous reset
+    *  input. Connects these to the DUT and does nothing else.
+    */
```
seldridge, line 38: Is clock and reset sufficient? @fabianschuiki has brought up some points about: (1) knowing whether or not the test is done and (2) knowing what the exit status is.

Does this motivate a moderately more heavyweight interface of:

``` scala
class UnitTestBundle extends Bundle {
  val clock = Flipped(Clock())
  val reset = Flipped(Bool())
  val done = Bool()
  val exitCode = UInt(32.W)
}
```

Alternatively, maybe there are two kinds of tests:

1. Tests that finish with either an assertion failure or a stop.
2. Tests that finish using the done/exitCode

Note: if the `stop` is dropped from (1), then (1) becomes purely a subset of (2).

seldridge (commented):
Some meandering feedback. I'm primarily looking to see if there is a way to avoid the boilerplate of the test harnesses around resetType, desired name, and the dut instantiation.

I think the main thing to consider is where this is highlighting what we are missing on the FIRRTL side in order to run these tests. We probably want to have a defined "test interface" off of which everything works. This PR isn't incompatible with that in any way.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -0,0 +1,80 @@
...
+class TestParameters[M <: RawModule, R] private[inlinetest] (
+  /** The [[desiredName]] of the DUT module. */
+  val dutName: String,
+  /** The user-provided name of the test. */
+  val testName: String,
+  /** A Definition of the DUT module. */
+  val dutDefinition: Definition[M],
+  /** The body for this test, returns a result. */
+  val body: Instance[M] => R
+)
```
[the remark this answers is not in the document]
tmckay-sifive: I originally had it that way, but according to @jackkoenig that is a problem for binary compatibility. I don't fully understand the reason for that, I'll admit.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: Yes great point. This was my main motivation for making it easy for users to override the testharness generator and result from each test. I imagine each project might have their own contract with a test driver to communicate these points.

I agree that `done`/`exitCode` is a sensible way to do that.

Since I do think most projects will settle on some testharness-testdriver contract to communicate done/pass, think it comes down to whether we prefer to ship an example implementation of this by default, or expect users to implement their own if they want it.

I landed on not adding done/pass because it felt less "presumptuous" of the users simulation environment. But I'm open to arguments that that comes at the cost of usability.

src/test/scala/chisel3/experimental/InlineTestSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+  val code = Output(UInt(8.W))
+}
+
+// Here is a testharness that consumes some kind of hardware from the test body, e.g.
+// a finish and pass/fail interface.
+object TestHarnessWithResultIO {
+  class TestHarnessModule[M <: RawModule](test: TestParameters[M, TestResultBundle]) extends Module {
+    override def resetType = Module.ResetType.Synchronous
+    override val desiredName = s"test_${test.dutName}_${test.testName}"
+    val dut = Instance(test.dutDefinition)
```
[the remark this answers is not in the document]
tmckay-sifive: I messed around with this and found it to be a bit clumsy. The main issue I face is with `resetType`. I cannot override it in the trait without specifying the self-type as `Module`. I don't want to impose that restriction. 

I'm not sure the deduplication is worth the added complexity, what do you think?

```diff
@@ -20,7 +20,18 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
   val dutDefinition: Definition[M],
   /** The body for this test, returns a result. */
   val testBody: Instance[M] => R
-)
+) {
+  def desiredTestModuleName = s"test_${dutName}_${testName}"
+}
+
+trait TestHarnessRawModule[M <: RawModule, R] extends Public { this: RawModule =>
+  def test: TestParameters[M, R]
+  override def desiredName = test.desiredTestModuleName
+}
+
+trait TestHarnessModule[M <: RawModule, R] extends TestHarnessRawModule[M, R] { this: Module =>
+  override def resetType = Module.ResetType.Synchronous
+}
 
 /** An implementation of a testharness generator.
   *
@@ -38,9 +49,9 @@ object TestHarnessGenerator {
   /** The minimal implementation of a unit testharness. Has a clock input and a synchronous reset
     *  input. Connects these to the DUT and does nothing else.
     */
-  class UnitTestHarness[M <: RawModule](test: TestParameters[M, Unit]) extends Module with Public {
+  class UnitTestHarness[M <: RawModule](val test: TestParameters[M, Unit]) extends Module with TestHarnessModule[M, Unit] {
     override def resetType = Module.ResetType.Synchronous
-    override val desiredName = s"test_${test.dutName}_${test.testName}"
+    override def desiredName = test.desiredTestModuleName
     val dut = Instance(test.dutDefinition)
     test.testBody(dut)
   }
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -0,0 +1,110 @@
...
+trait TestHarnessModule[M <: RawModule, R] extends TestHarnessRawModule[M, R] { this: Module =>
+  override def resetType = Module.ResetType.Synchronous
+}
+
+/** An implementation of a testharness generator.
+  *
+  *  @tparam M the type of the DUT module
+  *  @tparam R the type of the result returned by the test body
+  */
+trait TestHarnessGenerator[M <: RawModule, R] {
```
seldridge, line 57: Can the PR update this comment to explicitly call out that this is a type class.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -0,0 +1,110 @@
...
+/** TestHarnesses for inline tests should extend this. This trait sets the correct desiredName for
+ *  the module, instantiates the DUT, and provides methods to elaborate the test. By default, the
+ *  reset is synchronous, but this can be changed by overriding [[resetType]].
+ *
+ *  @tparam M the type of the DUT module
+ *  @tparam R the type of the result returned by the test body
+ */
+trait TestHarnessModule[M <: RawModule, R] extends TestHarnessRawModule[M, R] { this: Module =>
+  override def resetType = Module.ResetType.Synchronous
+}
```
seldridge, line 50: Mega-nit: to help with this organizationally, can the PR use an object to logically group these?

``` scala
object TestHarness {
  trait RawModule ...
  trait Module ...
}
```

seldridge (approved):
This looks great! Thank you for driving this!

We will need to sort out how to get the test-discovery and test-execution parts of this working. However, this is a great foundation on top of which we can start. 👍

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: Makes sense to me. If this is not blocking feedback, I will leave it for now.

I think it is a reasonable expectation that these testharness modules are not the top when we actually compile the simulation--there will be a test-driver layer above that generating clocks and resets, writing waves, etc. So, I think what we are looking for is an ABI between the test-driver and the testharness.

Perhaps testharness that can be `simulate`d must implement this ABI. At which point, circt-test could have its own intrinsic test-driver that works well with verilator and sane/configurable logic for clock/reset generation.

tmckay-sifive:
@jackkoenig @seldridge can we merge this?

PR #4641, opened 2025-01-23
[internal] Use State.guard in ModuleImpl

seldridge:
Change code for building a `Module` from using one-off state save/restore code to use new, common code for this.  This uses `Builder.State` methods which were developed for the `Placeholder` API, but were observed by @jackkoenig to be reusable here.

### DIFF
--- a/core/src/main/scala/chisel3/ModuleImpl.scala
+++ b/core/src/main/scala/chisel3/ModuleImpl.scala
@@ -69,63 +69,31 @@ private[chisel3] trait ObjectModuleImpl {
     }
     Builder.readyForModuleConstr = true
 
-    val parent = Builder.currentModule
-    val parentWhenStack = Builder.whenStack
-    val parentBlockStack = Builder.blockStack
-    val parentLayerStack = Builder.layerStack
+    val module = Builder.State.guard(Builder.State.default) {
+      val module: T = bc
 
-    // Save then clear clock and reset to prevent leaking scope, must be set again in the Module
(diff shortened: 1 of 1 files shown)


core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -1224,7 +1224,7 @@ private[chisel3] object Builder extends LazyLogging {
     def default: State = State(
       currentModule = Builder.currentModule,
       whenStack = Nil,
-      blockStack = Nil,
+      blockStack = Builder.blockStack,
```
seldridge, line 1227: This was a little surprising to me, but the old behavior is to leave the block stack in place and let the new module see it. This seems wrong, however, this is how it works.

I'm envisioning a larger refactor of this which makes a lot of this more straightforward. I.e., a module should have a single block created inside it. It shouldn't see the parent block stack. I haven't followed through on the ramifications of this change.

dtzSiFive: Yep! It's weird for sure.  Thanks for straightening this all out, I know I'd be happy to see the sort of "order" you're proposing brought to these bits.  But what you have already sure helps clean things up a lot, woohoo!

I can point you to further related bits (like Placeholder -> apply methods) if you'd like re:ramifications, but also you seem to be cruising so maybe best to just keep on 👍 .  Nothing that won't fall out, I don't think.

jackkoenig (approved):
It's beautiful 😭

PR #4642, opened 2025-01-24
Add UnitTest marker and test discovery utility

fabianschuiki:
Add the `UnitTest` trait as a marker for classes and objects that should be automatically discovered and constructed/executed as unit tests for hardware written in Chisel.

Also add a new `chisel3.UnitTests` main which discovers all unit tests in the classpath and generates the FIRRTL output for them. The discovery mechanism is roughly the same as implemented in ScalaTest, which scans the classpath for JAR files and directories and then tries to guess class names based on the encountered file structure. This feels pretty brittle, but has been working reliably for ScalaTest for a long time, so chances are this will work for Chisel too.

The utility offers a few options to list all available unit tests, or filter them according to inclusion or exclusion regular expressions. This allows unit tests of only the current package or a subset of the current project to be discovered and built. This is particularly useful if a Chisel project depends on libraries such as rocket-chip or by necessesity Chisel itself, whose unit tests are likely not of interest to that project.

Classes and objects can be marked as unit tests as follows:

    package magic

    class TestA extends UnitTest {
      Definition(new MyAdder(1))
      Definition(new MyAdder(42))
      // ...
    }

    object TestB extends UnitTest {
      for (i <- 1 to 4)
        Definition(new MyMultiplier(i))
      // ...
    }

The `chisel3.UnitTests` utility can then be execute, for example through mill or a standalone binary:

    # List all unit tests:
    > mill chisel[].runMain chisel3.UnitTests -l
    magic.TestA
    magic.TestB
    some.other.package.TestX

    # Produce FIRRTL output for all unit tests:
    > mill chisel[].runMain chisel3.UnitTests -o tests.fir

    # Filter tests according to a regex:
    > mill chisel[].runMain chisel3.UnitTests -l -f '^magic\.' -x TestA
    magic.TestB

As a convenience, Chisel modules can directly be marked as unit test as long as they do not have any constructor arguments. Doing so will automatically create a new `Definition` of the module without instantiating it. This can be useful for modules marked as public that serve as test harness, or by using a `FormalTest` marker to execute the module as a formal test.

    class MyTestHarness extends RawModule with UnitTest with Public {
      // ...
    }

    class MyFormalTest extends RawModule with UnitTest {
      // ...
      FormalTest(this)
    }

This is a first building block towards adding hardware unit tests to Chisel in a similar style as `AnyFlatSpec` and friends in ScalaTest, `#[test]` in Rust, lit tests in LLVM, or gtest in C++.

At a later stage, we may want to automate the execution of unit tests as formal or simulation checks, similar to how we have helper methods to run a Chisel generator all the way through to Verilog.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/UnitTest.scala
+++ b/src/main/scala/chisel3/UnitTest.scala
@@ -1,0 +1,135 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.test
+
+import chisel3.experimental.BaseModule
+import chisel3.experimental.hierarchy.Definition
+import chisel3.RawModule
+import java.io.File
+import java.util.jar.JarFile
+import scala.collection.JavaConverters._
+
+/** All classes and objects marked as [[UnitTest]] are automatically
+  * discoverable by the `DiscoverUnitTests` helper.
+  */
+trait UnitTest
+
+/** Helper to discover all subtypes of [[UnitTest]] in the class path, and call
+  * their constructors (if they are a class) or ensure that the singleton is
+  * constructed (if they are an object).
+  *
+  * This code is loosely based on the test suite discovery in scalatest, which
+  * performs the same scan over the classpath JAR files and directories, and
+  * guesses class names based on the encountered directory structure.
+  */
+private[chisel3] object DiscoverUnitTests {
+
+  /** The callback invoked for each unit test class name and unit test
+    * constructor.
+    */
+  type Callback = (String, () => Unit) => Unit
+
+  /** Discover all tests in the classpath and call `cb` for each. */
+  def apply(cb: Callback): Unit = classpath().foreach(discoverFile(_, cb))
+
+  /** Return the a sequence of files or directories on the classpath. */
+  private def classpath(): Iterable[File] = System
+    .getProperty("java.class.path")
+    .split(File.pathSeparator)
+    .map(s => if (s.trim.length == 0) "." else s)
+    .map(new File(_))
+
+  /** Discover all tests in a given file. If this is a JAR file, looks through
+    * its contents and tries to find its classes.
+    */
+  private def discoverFile(file: File, cb: Callback): Unit = file match {
+    // Unzip JAR files and process the class files they contain.
+    case _ if file.getPath.toLowerCase.endsWith(".jar") =>
+      val jarFile = new java.util.jar.JarFile(file)
+      jarFile.entries.asScala.foreach { jarEntry =>
+        val name = jarEntry.getName
+        if (!jarEntry.isDirectory && name.endsWith(".class"))
+          discoverClass(pathToClassName(name), cb)
+      }
+
+    // Recursively collect any class files contained in directories.
+    case _ if file.isDirectory =>
+      def visit(prefix: String, file: File): Unit = {
+        val name = prefix + "/" + file.getName
+        if (file.isDirectory) {
+          for (entry <- file.listFiles)
+            visit(name, entry)
+        } else if (name.endsWith(".class")) {
+          discoverClass(pathToClassName(name), cb)
+        }
+      }
+      for (entry <- file.listFiles)
+        visit("", entry)
+
+    // Ignore any other files that aren't directories.
+    case _ => ()
+  }
+
+  /** Convert a file path to a class */
+  private def pathToClassName(path: String): String =
+    path.replace('/', '.').replace('\\', '.').stripPrefix(".").stripSuffix(".class")
+
+  /** Load the given class and check whether it is a subtype of [[UnitTest]]. If
+    * it is, call the user-provided callback with a function that either calls
+    * the loaded class' constructor or ensures the loaded object is constructed.
+    */
+  private def discoverClass(className: String, cb: Callback): Unit = {
+    val clazz =
+      try {
+        classOf[UnitTest].getClassLoader.loadClass(className)
+      } catch {
+        case _: ClassNotFoundException       => return
+        case _: NoClassDefFoundError         => return
+        case _: ClassCastException           => return
+        case _: UnsupportedClassVersionError => return
+      }
+
+    // Check if it is a subtype of `UnitTest` (and also not the definition of
+    // `UnitTest` itself).
+    if (clazz == classOf[UnitTest] || !classOf[UnitTest].isAssignableFrom(clazz))
+      return
+
+    // Check if this is a `BaseModule`, in which case we implicitly wrap its
+    // constructor in a `Definition(...)` call.
+    val isModule = classOf[BaseModule].isAssignableFrom(clazz)
+
+    // Handle singleton objects by ensuring they are constructed.
+    try {
+      val field = clazz.getField("MODULE$")
+      if (isModule)
+        cb(className, () => Definition(field.get(null).asInstanceOf[BaseModule]))
+      else
+        cb(className, () => field.get(null))
+      return
+    } catch {
+      case e: NoSuchFieldException => ()
+    }
+
+    // Handle classes by calling their constructor.
+    try {
+      val ctor = clazz.getConstructor()
+      if (isModule)
+        cb(className, () => Definition(ctor.newInstance().asInstanceOf[BaseModule]))
+      else
+        cb(className, () => ctor.newInstance())
+      return
+    } catch {
+      case e: NoSuchMethodException  => ()
+      case e: IllegalAccessException => ()
+    }
+  }
+}
+
+/** A Chisel module that discovers and constructs all [[UnitTest]] subtypes
+  * discovered in the classpath. This is just here as a convenience top-level
+  * generator to collect all unit tests. In practice you would likely want to
+  * use a command line utility that offers some additional filtering capability.
+  */
+class AllUnitTests extends RawModule {
+  DiscoverUnitTests((_, gen) => gen())
+}
--- a/src/main/scala/chisel3/UnitTestMain.scala
+++ b/src/main/scala/chisel3/UnitTestMain.scala
@@ -1,0 +1,116 @@
+package chisel3
+
+import chisel3.test.DiscoverUnitTests
+import circt.stage.ChiselStage
+import java.io.File
+import java.io.PrintStream
+import scala.util.matching.Regex
+import scopt.OptionParser
+
+/** Utility to discover and generate all unit tests in the classpath. */
+object UnitTests {
+
+  /** Command line configuration options. */
+  case class Config(
+    outputFile: Option[File] = None,
+    list:       Boolean = false,
+    verbose:    Boolean = false,
+    filters:    List[Regex] = List(),
+    excludes:   List[Regex] = List()
+  )
+
+  def main(args: Array[String]): Unit = {
+    var shouldExit = false
+    val parser = new OptionParser[Config]("chisel3.UnitTests") {
+      head("Chisel Unit Test Utility")
+      help("help").abbr("h")
+
+      opt[File]('o', "output")
+        .text("Output file name (\"-\" for stdout)")
+        .action((x, c) => c.copy(outputFile = if (!x.getPath.isEmpty && x.getPath != "-") Some(x) else None))
+
+      opt[Unit]('l', "list")
+        .text("List tests instead of building them")
+        .action((_, c) => c.copy(list = true))
+
+      opt[Unit]('v', "verbose")
+        .text("Print verbose information to stderr")
+        .action((_, c) => c.copy(verbose = true))
+
+      opt[Seq[String]]('f', "filter")
+        .text("Only consider tests which match at least one filter regex")
+        .unbounded()
+        .action((x, c) => c.copy(filters = c.filters ++ x.map(_.r)))
+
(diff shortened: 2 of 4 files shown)


src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+package chisel3
+
+import chisel3.experimental.BaseModule
+import chisel3.experimental.hierarchy.Definition
+import java.io.File
+import java.util.jar.JarFile
+import scala.collection.JavaConverters._
+
+/** All classes and objects marked as `UnitTest` are automatically discoverable
+ *  by the `DiscoverUnitTests` helper. */
```
seldridge, line 12: Nit: Usually I've seen multi-line comments look like:

```suggestion
/** All classes and objects marked as `UnitTest` are automatically discoverable
  *  by the `DiscoverUnitTests` helper. 
  */
```

If you run reformat, this may be fixed: `./mill __.reformat`.

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+          for (entry <- file.listFiles)
+            visit(name, entry)
+        } else if (name.endsWith(".class")) {
+          discoverClass(pathToClassName(name), cb)
+        }
+      }
+      for (entry <- file.listFiles)
+        visit("", entry)
+    }
+  }
```
seldridge, line 61: Nit: I'd opt for either: (1) early return style or (2) a match. I.e.,:

``` scala
if (file.getPath.toLowerCase.endsWith(".jar")) {
  /** ... */
  return
}

if (!file.isDirectory) 
  return

/** Handle directory */
```

Or:

``` scala
file match {
  case jarFile if jarFile.getPath.toLowerCase.endsWith(".jar") => /** handle jar filr */
  case dir if dir.isDirectory => /** handle directory */
  case _ => 
}
```

Most Scala code tends towards (2) as it is more obvious what is going on.

seldridge: Other piece: both `discoverFile` and `visit` look to be tail recursive. You can force the compiler to check that this is definitely the case with:

``` scala
import scala.annotation

@annotation.tailrec
def foo(): Unit = ???
```

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  /** Convert a file path to a class */
+  private def pathToClassName(path: String): String =
+    path.replace('/', '.').replace('\\', '.').stripPrefix(".").stripSuffix(".class")
```
seldridge, line 65: Can the separate replacement of forward and backwards slash be implemented with a replacement of `File.pathSeparator`?

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  /** Load the given class and check whether it is a subtype of `UnitTest`. If
+   *  it is, call its constructor if it is a class, or ensure it is constructed
+   *  if it is an object. */
```
seldridge, line 69: Nit: In Scaladoc, you can use `[[Foo]]` to link against the class or trait `Foo`. This will provide a slightly better experience when looking at the API docs:

```suggestion
  /** Load the given class and check whether it is a subtype of [[UnitTest]]. If
   *  it is, call its constructor if it is a class, or ensure it is constructed
   *  if it is an object. */
```

I will tend to reserve `foo` when talking about arguments.

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+  private def discoverClass(className: String, cb: Callback): Unit = {
+    val clazz =
+      try {
+        classOf[UnitTest].getClassLoader.loadClass(className)
+      } catch {
+        case e: ClassNotFoundException       => return
+        case e: NoClassDefFoundError         => return
+        case e: ClassCastException           => return
+        case e: UnsupportedClassVersionError => return
```
seldridge, line 78: Consider:

```suggestion
        case _: ClassNotFoundException | _: NoClassDefFoundError | _: ClassCastException | _: UnsupportedClassVersionError => return
```

This sometimes gets _worse_ when you are matching on a large number of types. See what you think.

Also, at minimum `e` is captured, but not used, so use of `_` is better.

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+trait UnitTest
+
+/** Helper to discover all subtypes of `UnitTest` in the class path, and call
+ *  their constructors (if they are a class) or ensure that the singleton is
+ *  constructed (if they are an object).
+ *
+ *  This code is loosely based on the test suite discovery in scalatest, which
+ *  performs the same scan over the classpath JAR files and directories, and
+ *  guesses class names based on the encountered directory structure. */
+object DiscoverUnitTests {
```
seldridge, line 22: Does this need to be public or can it be package private (`private[chisel3]`)?

src/main/scala/chisel3/UnitTestMain.scala
```diff
@@ -0,0 +1,109 @@
...
+object UnitTests {
+  def main(args: Array[String]): Unit = {
+    var shouldExit = false
+    val parser = new OptionParser[Config]("chisel3.UnitTests") {
+      head("Chisel Unit Test Utility")
+      help("help").abbr("h")
+
+      opt[File]('o', "output")
+        .text("Output file name (\"-\" for stdout)")
+        .action((x, c) => c.copy(outputFile = if (!x.getPath.isEmpty && x.getPath != "-") Some(x) else None))
```
seldridge, line 19: Great use of including `-` as an alias to stdout.

src/main/scala/chisel3/UnitTestMain.scala
```diff
@@ -0,0 +1,109 @@
...
+      if (config.excludes.exists(_.findFirstMatchIn(className).isDefined)) {
+        Console.err.println(f"Skipping ${className} (matches exclude filter)")
+        return
+      }
+
+      // If we are just listing tests, print the class name and skip this test.
+      if (config.list) {
+        println(className)
+        return
+      }
```
seldridge, line 72: Sending this to stdout seems right.

src/main/scala/chisel3/UnitTestMain.scala
```diff
@@ -0,0 +1,109 @@
...
+      gen()
+    }
+
+    // Generate the unit tests.
+    class AllUnitTests extends RawModule {
+      DiscoverUnitTests(handler)
+    }
+    val chirrtl = ChiselStage.emitCHIRRTL(new AllUnitTests)
+    if (config.list)
+      return
```
seldridge, line 86: The fact that `list` doesn't just list tests, but also builds things and then discards the result seems a bit weird.

src/main/scala/chisel3/UnitTestMain.scala
```diff
@@ -0,0 +1,109 @@
...
+      case Some(file) =>
+        val w = new PrintWriter(file)
+        try {
+          w.write(chirrtl)
+        } finally {
+          w.close()
+        }
+      case None => print(chirrtl)
+    }
+  }
```
seldridge, line 99: This is fine. The PR could unify this handling by defining a single thing to write to and then always using that. Also, use of `PrintStream` is likely preferred. That also has the benefit of `System.Out` being an instance of `PrintStream` already.

src/main/scala/chisel3/UnitTestMain.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /** Command line configuration options. */
+  case class Config(
+    outputFile: Option[File] = None,
+    list:       Boolean = false,
+    verbose:    Boolean = false,
+    filters:    List[Regex] = List(),
+    excludes:   List[Regex] = List()
+  )
```
seldridge, line 108: Nit: I'd move this earlier in the file. 😉

src/test/scala/chiselTests/UnitTestMainSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+    if (!checkOut.isEmpty) fileCheckString(outStream.toString)(checkOut)
+    if (!checkErr.isEmpty) fileCheckString(errStream.toString)(checkErr)
+  }
+
+  def checkOutAndErr(args: String*)(checkOut: String, checkErr: String): Unit = check(args)(checkOut, checkErr)
+
+  def checkOut(args: String*)(checkOut: String): Unit = check(args)(checkOut, "")
+  def checkErr(args: String*)(checkErr: String): Unit = check(args)("", checkErr)
```
seldridge, line 26: This is nice.

FWIW, I have had difficulty defining methods like this in the past because Scalatest seems to run things multi-process, but share stdout and stderr. If you start seeing intermittent test failures, this may be the cause.

seldridge (approved):
LGTM

Nitty comments throughout and feel free to address them as you see fit.

As an aside: what is interesting about this PR is that it may have also defined a notion of a "package", i.e., a collection of modules that should all be built together.

src/main/scala/chisel3/UnitTest.scala
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
```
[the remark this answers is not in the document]
fabianschuiki: I think `chisel3.test` would be neat, but then I realized that there's no `chisel3.test` at all currently. My next thought was that if `Module` and `RawModule` are trivially imported by `import chisel3._`, maybe you should also get `UnitTest` for free to make it as simple as posisble. But that's a pretty shallow and dumb motivation :grin:. I don't mind making this `chisel3.test`, that feels appropriate.

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: It's a shame that `./mill __.reformat` doesn't reformat comments properly (at least on my machine) :cry:

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: Either way works. No need for this to be public, except maybe some fringe users wanting to discover unit tests and perform their own action on them. But we could definitely keep this internal until someone asks for it.

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: I don't think so, because on Windows we'll run into this weird situation where file paths contain `\`, while the JAR file unzipping will still produce `/` as the file separator. (At least the ScalaTest implementation suggests that this happens.)

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: The case style looks pretty readable :+1: Tail recursion doesn't work unfortunately because most recursive calls are inside loops.

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: Done!

src/main/scala/chisel3/UnitTest.scala
fabianschuiki: :+1: Good point about the `e` never being used. Collapsing stuff on one line feels a bit less legible.

src/main/scala/chisel3/UnitTestMain.scala
fabianschuiki: It ends up not building anything due to the `if (config.list) { ... return }` early return in the handler. Besides a top-level `AllUnitTests` module, which is a bit ugly. I think in case of `list` we could just skip the `emitCHIRRTL` part and run discovery directly outside a Chisel builder context.

src/main/scala/chisel3/UnitTestMain.scala
fabianschuiki: :+1: done

src/main/scala/chisel3/UnitTest.scala
seldridge: Got it. This is fine to leave as is, then.

src/test/scala/chiselTests/UnitTestMainSpec.scala
fabianschuiki: Interesting :thinking: I was hoping that the `withOut` and `withErr` would somehow setup a thread-local redirect of these streams :crossed_fingers:. Haven't seen a failure yet. But better be on the lookout for these.

fabianschuiki: ChatGPT surfaced this: https://github.com/scala/scala/blob/2.13.x/src/library/scala/Console.scala#L127-L129

src/main/scala/chisel3/UnitTestMain.scala
fabianschuiki: I've added a separate discovery run without constructing any Chisel modules in case `list` is specified :+1:

PR #4643, opened 2025-01-24
Add safer Chisel annotation API, deprecate old ones

jackkoenig:
Note that I tested that the deprecated versions of `annotate` work by updating them first, making sure all tests pass, then removing all uses of ChiselAnnotation and ChiselMultiAnnotation except in ChiselEnum. The ones in ChiselEnum have duplication checking that I could replicate with the annotations, but is annoying. I figure we'd rather just remove those (and roll out firrtl enums? 🙏) so I just left them. They also do help test the deprecated APIs.

This PR has two purposes:
1. Make it possible to check that any data being annotated are annotatable. You can see we've done that for some annotations but not all--this new API enforces it for all annotations (or at least encourages enforcement).
2. By [softly] requiring the user to record what Data are being annotated up front, we can improve the view renaming logic. Rather than having to conservatively rename tons and tons of views, we only do so for those that are actually annotated. This should be a huge performance improvement for designs that use lots of views.

(2) will be in a follow on PR.

One weirdness in the PR that I would appreciate review of is that the new API asks the caller to "report" all `InstanceIds` that will be annotated. We really only currently care about `Data` but I figure in the future we may want checks for other things. The awkwardness though, is that `InstanceId` includes `Data`, `BaseModule`, and `MemBase`, but not `Hierarchy` nor the new SRAM targets. So if we wanted more checks, we might need to change the API anyway in the future since users can't currently report `Hierarchy` or `SRAM` targets being annotated. I could take a step back and try to find a way to unify all of these, or just leave it, or switch the API to `Data` to make it more clear that that is all we're checking. WDYT?

### Contributor Checklist

#### Type of Improvement

- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

Creating annotations in Chisel now requires reporting what `InstanceIds` are going to be annotated so that Chisel can do some safety checks.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Annotation.scala
+++ b/core/src/main/scala/chisel3/Annotation.scala
@@ -3,8 +3,10 @@
 package chisel3.experimental
 
 import scala.language.existentials
+import scala.annotation.nowarn
 import chisel3.internal.Builder
-import chisel3.{Data, InstanceId, RawModule}
+import chisel3.{Data, HasTarget, InstanceId, RawModule}
+import chisel3.experimental.AnyTargetable
 import firrtl.annotations._
 import firrtl.options.Unserializable
 import firrtl.transforms.{DedupGroupAnnotation, NoDedupAnnotation}
@@ -13,6 +15,10 @@ import firrtl.transforms.{DedupGroupAnnotation, NoDedupAnnotation}
   *
   * Defines a conversion to a corresponding FIRRTL Annotation
   */
+@deprecated(
+  "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
+  "Chisel 6.7.0"
+)
 trait ChiselAnnotation {
 
   /** Conversion to FIRRTL Annotation */
@@ -23,16 +29,45 @@ trait ChiselAnnotation {
   *
   *  Defines a conversion to corresponding FIRRTL Annotation(s)
   */
+@deprecated(
+  "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
+  "Chisel 6.7.0"
+)
 trait ChiselMultiAnnotation {
   def toFirrtl: Seq[Annotation]
 }
 
+@nowarn("msg=Avoid custom annotations")
 object annotate {
-  def apply(anno: ChiselAnnotation): Unit = {
+  @deprecated(
+    "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
+    "Chisel 6.7.0"
+  )
+  def apply(anno: ChiselAnnotation): Unit =
     Builder.annotations += anno
-  }
-  def apply(annos: ChiselMultiAnnotation): Unit = {
+
+  @deprecated(
+    "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
+    "Chisel 6.7.0"
+  )
+  def apply(annos: ChiselMultiAnnotation): Unit =
     Builder.newAnnotations += annos
+
+  /** Create annotations.
+    *
+    * Avoid this API if possible.
+    *
+    * Anything being annotated must be passed as arguments so that Chisel can do safety checks.
+    * The caller is still responsible for calling .toTarget on those arguments in mkAnnos.
+    */
+  def apply(targets: AnyTargetable*)(mkAnnos: => Seq[Annotation]): Unit = {
+    targets.map(_.a).foreach {
+      case d: Data => requireIsAnnotatable(d, "Data marked with annotation")
+      case _ => ()
+    }
+    Builder.newAnnotations += new ChiselMultiAnnotation {
+      def toFirrtl: Seq[Annotation] = mkAnnos
+    }
   }
 }
 
@@ -76,7 +111,7 @@ object doNotDedup {
     * @return Unmodified signal `module`
     */
   def apply[T <: RawModule](module: T): Unit = {
-    annotate(new ChiselAnnotation { def toFirrtl: NoDedupAnnotation = NoDedupAnnotation(module.toNamed) })
+    annotate(module)(Seq(NoDedupAnnotation(module.toNamed)))
   }
 }
 
@@ -89,6 +124,6 @@ object dedupGroup {
     * @return Unmodified signal `module`
     */
   def apply[T <: BaseModule](module: T, group: String): Unit = {
-    annotate(new ChiselAnnotation { def toFirrtl: DedupGroupAnnotation = DedupGroupAnnotation(module.toTarget, group) })
+    annotate(module)(Seq(DedupGroupAnnotation(module.toTarget, group)))
   }
 }
--- a/core/src/main/scala/chisel3/ChiselEnumImpl.scala
+++ b/core/src/main/scala/chisel3/ChiselEnumImpl.scala
@@ -3,6 +3,7 @@
 package chisel3
 
 import scala.language.existentials
+import scala.annotation.nowarn
 import scala.collection.mutable
 import chisel3.experimental.{annotate, requireIsHardware, ChiselAnnotation, SourceInfo, UnlocatableSourceInfo}
 import chisel3.internal.Builder.pushOp
@@ -13,6 +14,9 @@ import chisel3.internal.binding.{Binding, ChildBinding, ConstrainedBinding}
 
 import chisel3.experimental.EnumAnnotations._
 
+// Rather than refactoring the annotation work here, we should just remove ChiselEnum annotations
+@nowarn("msg=Avoid custom annotations")
+@nowarn("msg=Enum annotations will be removed")
 private[chisel3] abstract class EnumTypeImpl(private[chisel3] val factory: ChiselEnum, selfAnnotating: Boolean = true)
     extends Element { self: EnumType =>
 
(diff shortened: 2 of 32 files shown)


core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -568,50 +568,16 @@ private[chisel3] object ir {
...
-        DefTypeAlias
-      ], Seq[Layer], Seq[DefOption], Circuit] {
-    def unapply(c: Circuit): Option[(String, Seq[Component], Seq[ChiselAnnotation], RenameMap, Seq[DefTypeAlias])] = {
-      Some((c.name, c.components, c.annotations, c.renames, c.typeAliases))
-    }
-
-    def apply(
-      name:        String,
-      components:  Seq[Component],
-      annotations: Seq[ChiselAnnotation],
```
jackkoenig, line 609: All of this was left over from previous needs to maintain binary compatibility. These APIs are all package private now so we can just delete them.

jackkoenig:
I'm thinking maybe backport this deprecation to 6.x, then I can follow up with the API removal on main and then we can merge #4644 to get the performance improvement.

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -23,16 +28,42 @@ trait ChiselAnnotation {
...
+    * The caller is still responsible for calling .toTarget on those arguments in mkAnnos.
+    */
+  def apply(targets: InstanceId*)(mkAnnos: => Seq[Annotation]): Unit = {
+    targets.foreach {
+      case d: Data => requireIsAnnotatable(d, "Data marked with annotation")
+      case _ => ()
+    }
+    // TODO record views that need to be renamed
+    Builder.annotations += (() => mkAnnos)
   }
```
seldridge, line 67: Is there a way to do this which directly passes the `targets` to the annotation construction function directly?

For this to work, an `Annotation` (or a subclass of) would need to have a constructor that can build it from an `InstanceId` (or a `Data`).

I'm primarily worried about there being no check of the connection between the `targets` and the `mkAnnos` function as a user could put in incorrect targets or ignore them entirely to get the old, legacy behavior.

Can you think of a way to untangle this?

core/src/main/scala/chisel3/experimental/EnumAnnotations.scala
```diff
@@ -16,6 +19,10 @@ object EnumAnnotations {
     def duplicate(n: Named): EnumComponentAnnotation = this.copy(target = n)
   }
 
+  @deprecated(
+    "Avoid custom annotations. If you must use annotations, new annotate.apply method that takes Data",
+    "Chisel 7.0"
```
jackkoenig, line 24: Wrong deprecation message, just deprecate the annotation.

core/src/main/scala/chisel3/Annotation.scala
jackkoenig: Discussed offline--we could untangle this, but it's a lot of extra infrastructure around stuff that users aren't really supposed to be using anyway since user-defined annotations are not supported in CIRCT. This is slightly ugly but good enough.

core/src/main/scala/chisel3/ChiselEnumImpl.scala
seldridge, line 1: So, the plan with this PR is to ignore the enum annotations and let them use the deprecated methods. We can then see about removing the annotations in a follow-on. Is that right?

src/test/scala/chiselTests/BoringUtilsSpec.scala
```diff
@@ -44,9 +44,7 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
 
   trait WireX { this: BaseModule =>
     val x = Wire(UInt(4.W))
-    chisel3.experimental.annotate(new ChiselAnnotation {
-      def toFirrtl: Annotation = DontTouchAnnotation(x.toNamed)
-    })
+    dontTouch(x)
```
seldridge, line 47: Nit: this can go all the way to:

``` scala
val x = dontTouch(Wire(UInt(4.W)))
```

core/src/main/scala/chisel3/experimental/Targetable.scala
```diff
@@ -0,0 +1,116 @@
...
+
+/** Type class for types that can be converted to a Target
+  *
+  * See [[AnyTargetable]] for type-erased containers of Targetables
+  * 
+  * @note This uses type classes instead of inheritance because Hierarchy does not constrain its type parameter
+  *       and thus not all instances of Hierarchy are Targetable.
+  */
+sealed trait Targetable[A] {
+  def toTarget(a: A): IsMember
```
seldridge, line 20: Scaladoc?

core/src/main/scala/chisel3/experimental/Targetable.scala
```diff
@@ -0,0 +1,116 @@
...
+  def toAbsoluteTarget:                           IsMember = targetable.toAbsoluteTarget(a)
+  def toRelativeTarget(root: Option[BaseModule]): IsMember = targetable.toRelativeTarget(a, root)
+  def toRelativeTargetToHierarchy(root: Option[Hierarchy[BaseModule]]): IsMember =
+    targetable.toRelativeTargetToHierarchy(a, root)
+}
+
+object AnyTargetable {
+
+  /** Implicit conversion making working with Targetables easier */
+  implicit def erase[A](a: A)(implicit targetable: Targetable[A]): AnyTargetable = {
```
seldridge, line 106: Nit: I understand that this is "erasing" the type of `A`. However, mechanically it is an implicit conversion to `AnyTargetable`. Consider a more pedestrian method name like `anyToAnyTargetable` or `toAnyTargetable`. Only for consideration.

seldridge (approved):
LGTM

There is a lot of good cleanup here.

After this, can we deprecate `ChiselAnnotation` given that this appears to remove all uses in favor of `() => Seq[Annotation]`?

core/src/main/scala/chisel3/ChiselEnumImpl.scala
jackkoenig: That's correct, I plan to backport this PR to 6.x (although with some parts backed out) and then delete the enum annotations (and deprecated annotate methods).

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -13,6 +15,10 @@ import firrtl.transforms.{DedupGroupAnnotation, NoDedupAnnotation}
   *
   * Defines a conversion to a corresponding FIRRTL Annotation
   */
+@deprecated(
+  "Avoid custom annotations. If you must use annotations, new annotate.apply method that takes Data",
```
mwachs5, line 19:
```suggestion
  "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
```

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -23,16 +29,41 @@ trait ChiselAnnotation {
   *
   *  Defines a conversion to corresponding FIRRTL Annotation(s)
   */
+@deprecated(
+  "Avoid custom annotations. If you must use annotations, new annotate.apply method that takes Data",
```
mwachs5, line 33:
```suggestion
  "Avoid custom annotations. If you must use annotations, annotate.apply method that takes Data",
```

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -23,16 +29,41 @@ trait ChiselAnnotation {
...
 
+@nowarn("msg=Avoid custom annotations")
 object annotate {
-  def apply(anno: ChiselAnnotation): Unit = {
-    Builder.annotations += anno
-  }
-  def apply(annos: ChiselMultiAnnotation): Unit = {
-    Builder.newAnnotations += annos
+  @deprecated(
+    "Avoid custom annotations. If you must use annotations, new annotate.apply method that takes Data",
```
mwachs5, line 43:
```suggestion
    "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
```

core/src/main/scala/chisel3/Annotation.scala
```diff
@@ -23,16 +29,41 @@ trait ChiselAnnotation {
...
+    "Chisel 6.7.0"
+  )
+  def apply(anno: ChiselAnnotation): Unit = apply()(Seq(anno.toFirrtl))
+
+  @deprecated(
+    "Avoid custom annotations. If you must use annotations, new annotate.apply method that takes Data",
```
mwachs5, line 49:
```suggestion
    "Avoid custom annotations. If you must use annotations, use annotate.apply method that takes Data",
```

jackkoenig:
Converted to draft because it looks like `ChiselCircuitAnnotation` leaks `ir.Circuit` which gives access to underlying annotations in a way that is problematic for this PR. Need to think of a solution.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: Backing out this change for now because `Circuit` leaks as discussed in https://github.com/chipsalliance/chisel/pull/4683. Will follow up tearing this out later.

PR #4644, opened 2025-01-24
Selective view renaming

jackkoenig:
Now depends on https://github.com/chipsalliance/chisel/pull/4717.

Because views may not map 1-1 to FIRRTL targets and that annotations are so heavily exposed to the user (esp. via Stage/Phase), we have to generate a RenameMap for any views that are annotated. Previously, we couldn't know what views might be annotated (the `.toTarget` call is after Chisel elaboration) so we had to conservatively generate a `RenameMap` for all views that don't map 1-1. This is **ludicriously** expensive.

With #4643, we now require that annotations report what Data will be annotated up front, so now we only need to rename views that are actually annotated. Much more elegant and efficient at the cost of a minor API change.

I guess technically we shouldn't merge this until the deprecated APIs are removed... 🙈 

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/experimental/dataview/package.scala
+++ b/core/src/main/scala/chisel3/experimental/dataview/package.scala
@@ -236,24 +236,27 @@ package object dataview {
       case elt: Element => view.bind(ViewBinding(elementResult(elt), writability))
       case agg: Aggregate =>
         val fullResult = elementResult ++ aggregateMappings
-
-        // We need to record any Aggregates that don't have a single-target mapping.
-        // Technically, this adds any Aggregate that is not an identity mapping,
-        // but we don't have a cheap way to check for single-target.
-        // The Builder context check is a hack to allow calling .viewAs outside of elaboration
-        // on views that are not identity but still single-target.
-        // FIXME we really just need to get rid of the need for unnammed view renaming.
-        if (Builder.inContext) {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/chisel3/experimental/dataview/package.scala
```diff
@@ -236,27 +236,30 @@ package object dataview {
...
     }
   }
 
+  // When annotating views that are not identity mappings, we need to record them for renaming
+  // Technically, this adds any Aggregate that is not an identity mapping,
+  // but we don't have a cheap way to check for single-target.
+  private[chisel3] def recordViewForRenaming(view: Data): Unit = {
+    view.topBinding match {
+      case _: ViewBinding => // No need for renaming
+      case AggregateViewBinding(mapping, _) => // Do nothing
```
seldridge, line 252: Is the trailing "Do nothing" comment correct?

seldridge (approved):
LGTM

core/src/main/scala/chisel3/experimental/dataview/package.scala
jackkoenig: Lol no it is not, not sure what I was doing there. Will fix, thanks!

PR #4649, opened 2025-01-27
Add -release:8 to scalacOptions

jackkoenig:
This wasn't previously necessary because we always published using Java 8, but since Mill now requires at least Java 11+, we now need to do this.

I tested this locally by using `import java.util.random.RandomGenerator` (added in Java 17) in our Scala code. This worked before I added the option, then once I added `-release:8` it became a compile error.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

This ensures published Chisel artifacts work with Java 8--both in generated bytecode (which was previously true because default setting is -target:8), but also in disallowing APIs added in newer JDKs.

- [ ] Did you review?

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -99,6 +99,7 @@ object v extends Module {
     "-unchecked",
     "-Werror",
     "-Ymacro-annotations",
+    "-release:8",
     "-explaintypes",
     "-Xcheckinit",
     "-Xlint:infer-any",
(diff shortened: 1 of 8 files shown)


build.mill
```diff
@@ -99,6 +99,7 @@ object v extends Module {
     "-unchecked",
     "-Werror",
     "-Ymacro-annotations",
+    "-release:8",
```
seldridge, line 102: Good to lock this in.

integration-tests/package.mill
```diff
@@ -16,6 +16,14 @@ trait IntegrationTests extends CrossSbtModule with HasScala2Plugin with Scalafmt
   def pluginModule = plugin.cross()
   def millSourcePath = super.millSourcePath / os.up
 
+  override def scalacOptions = Task {
+    if (v.isScala3(crossScalaVersion)) {
+      Seq.empty[String]
+    } else {
+      v.scala2CommonOptions
+    }
+  }
```
seldridge, line 25: This function appears three times. Is there a common utility that this should be factored into?

seldridge (approved):

integration-tests/package.mill
jackkoenig: Probably--the tricky part is it is a Task (because it needs to create `crossScalaVersion` which is a Task), so I think the only real way to reuse here is with a base trait that we mix in. This PR is just rolling out a pattern we're already using elsewhere, but I agree we should revisit the pattern.

jackkoenig: Okay I've factored stuff into `HasCommonOptions`.

PR #4650, opened 2025-01-28
[ci] Speed up tests with testForkGrouping

jackkoenig:
See docs: https://mill-build.org/mill/javalib/testing.html#_test_grouping

We could possibly do something more intelligent, but this speeds things up quite a bit on my MBP (~7.5 min to run `chisel[].test.test` vs. 15 min without this change), should speed up CI a bit too.

I tried using ScalaTest's parallelism (by overriding [def test](https://github.com/com-lihaoyi/mill/blob/07da626f5516b574586592f6083e872ead7d9752/scalalib/src/mill/scalalib/TestModule.scala#L70) passing `-P<nprocs>`) but it was slower.

I also tried `.grouped(1)` but that was much slower. `.grouped(8)` at least seems to speed things up a bit. It would be nice if the parallelism could be dynamically scheduled (the grouping is basically a static separating out tests to JVM forks so you really want to avoid putting all of the slow specs on a single JVM). That being said, we should probably just ensure that no spec is too slow.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -283,6 +283,8 @@ trait Chisel extends CrossSbtModule with HasScala2MacroAnno with HasScala2Plugin
     // Suppress Scala 3 behavior requiring explicit types on implicit definitions
     // Note this must come before the -Wconf is warningSuppression
     override def scalacOptions = Task { super.scalacOptions() :+ "-Wconf:cat=other-implicit-type:s" }
+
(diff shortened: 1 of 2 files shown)


sequencer (approved):

jackkoenig:
This needs https://github.com/chipsalliance/chisel/pull/4653

PR #4652, opened 2025-01-29
[svsim] Use 2-state variable for ports

seldridge:
Fix an apparent additional problem with svsim and VCS xprop.  For reasons that I am unable to comprehend, this change gets xprop simulations to pass.  Without them, I am continuing to see spurious clock toggles causing assertions to fire.

This is a follow-on to a77eee6.  This is a bit of a game of whack-a-mole at this point...

#### Release Notes

Use 2-state logic in svsim variables used to drive ports to (hopefully) fix lingering problems with VCS and xprop.

### DIFF
--- a/svsim/src/main/scala/Workspace.scala
+++ b/svsim/src/main/scala/Workspace.scala
@@ -97,7 +97,7 @@ final class Workspace(
       l("module ", Workspace.testbenchModuleName, ";")
       for ((port, index) <- ports) {
         if (port.isSettable) {
-      l("  reg  [$bits(", dut.instanceName, ".",  port.name, ")-1:0] ", port.name, " = '0;")
+      l("  bit  [$bits(", dut.instanceName, ".",  port.name, ")-1:0] ", port.name, " = '0;")
         } else {
       l("  wire [$bits(", dut.instanceName, ".",  port.name, ")-1:0] ", port.name, ";")
         }


svsim/src/main/scala/Workspace.scala
```diff
@@ -97,7 +97,7 @@ final class Workspace(
       l("module ", Workspace.testbenchModuleName, ";")
       for ((port, index) <- ports) {
         if (port.isSettable) {
-      l("  reg  [$bits(", dut.instanceName, ".",  port.name, ")-1:0] ", port.name, " = '0;")
+      l("  bit  [$bits(", dut.instanceName, ".",  port.name, ")-1:0] ", port.name, " = '0;")
```
seldridge, line 100: Later in this same file, the DPI functions like `setBitsImpl_*` are already using `bit` as their data type. Ignoring the hope for a VCS xprop fix, it does seem more consistent to use `bit` throughout. I can't see a reason to go from 2-state to 4-state unless the initial 4-state value is intended to be significant.

jackkoenig: Makes sense to me

seldridge:
I have a better understanding of what was going on here. This should no longer be needed (nor should the earlier PR that initialized these). The problem that this was trying to address was problems related to time-zero simulator initialization, specifically use of VCS' `+vcs+initreg` feature which can seemingly create phantom clock edges. I have this sorted out by dropping use of this feature and instead relying on normal FIRRTL randomization.

PR #4656, opened 2025-02-03
Speedup Computation for whether Aggregates containsProbe

mwachs5:
### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

#### Release Notes

Add a private containsProbe var to Aggregate and use it to speed up containsProbe checks.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/AggregateImpl.scala
+++ b/core/src/main/scala/chisel3/AggregateImpl.scala
@@ -82,6 +82,9 @@ private[chisel3] trait AggregateImpl extends Data { thiz: Aggregate =>
     */
   def getElements: Seq[Data]
 
+  /** Amortize this result, as it can be very expensive to determine this otherwise */
+  private[chisel3] lazy val containsProbe: Boolean = elementsIterator.exists(d => chisel3.internal.containsProbe(d))
+
   /** Similar to [[getElements]] but allows for more optimized use */
   private[chisel3] def elementsIterator: Iterator[Data]
 
--- a/core/src/main/scala/chisel3/AggregateImpl.scala
+++ b/core/src/main/scala/chisel3/AggregateImpl.scala
@@ -82,7 +82,7 @@ private[chisel3] trait AggregateImpl extends Data { thiz: Aggregate =>
     */
   def getElements: Seq[Data]
 
-  /** Amortize this result, as it can be very expensive to determine this otherwise */
+  /** Save this result, as it can be very expensive to determine this otherwise */
   private[chisel3] lazy val containsProbe: Boolean = elementsIterator.exists(d => chisel3.internal.containsProbe(d))
 
   /** Similar to [[getElements]] but allows for more optimized use */
--- a/core/src/main/scala/chisel3/internal/package.scala
+++ b/core/src/main/scala/chisel3/internal/package.scala
@@ -163,8 +163,7 @@ package object internal {
   }
 
   private[chisel3] def containsProbe(data: Data): Boolean = data match {
-    case a: Aggregate =>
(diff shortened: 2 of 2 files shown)


seldridge (approved):
LGTM

dtzSiFive (approved):
LGTM! Caching these recursive type properties is a good call, thanks!

Only doubt is if a Data is re-used or mutated to be probe-y, but not sure that's a thing?

jackkoenig (approved):
I think this is reasonable as is but there are a couple of possible improvements:

1. Specialize this on Vec to only check the sample element
2. Consider doing it live with some normal traversal. This is tricky to figure out the right place so I'm not sure it's worth untangling at the moment, but e.g. we do that for if something contains a flipped [[1]](https://github.com/chipsalliance/chisel/blob/62ca9bdca4648ccd1d05141c0fd048d9eae72a10/core/src/main/scala/chisel3/AggregateImpl.scala#L905). That spot won't quite work because we do `containsProbe` on unbound types, but maybe we can figure out a similar approach.

dtzSiFive:
> Only doubt is if a Data is re-used or mutated to be probe-y, but not sure that's a thing?

https://github.com/chipsalliance/chisel/blob/62ca9bdca4648ccd1d05141c0fd048d9eae72a10/core/src/main/scala/chisel3/probe/ProbeBase.scala#L35

and uses of `setProbeModifier` (which seem today to clone)?

This will increase the object size of anything that's an Aggregate to accommodate the new field -- offhand I don't know the best recommendation but IIRC @jackkoenig went around with some tricks to reduce size needed for lazy vals in common objects? Or thereabouts?

jackkoenig:
> This will increase the object size of anything that's an Aggregate to accommodate the new field -- offhand I don't know the best recommendation but IIRC @jackkoenig went around with some tricks to reduce size needed for lazy vals in common objects? Or thereabouts?

I'm not too worried about an extra `Boolean` (+ 1 bit in lazy val flags but there should be space there) in Aggregates. I'd care more if this were on all `Data` but it isn't so it's fine!

jackkoenig (changes requested):
>  val ret: T = if (!data.mustClone(prevId)) data else data.cloneType.asInstanceOf[T] 

This is a problem though, good eye @dtzSiFive

@mwachs5 please add a test for this, I'm struggling to suggest the precise check but given an Aggregate that is taking advantage of the !mustClone, it will have its `containsProbe` memoized to false but then it will be mutated to itself be a Probe.

I think there are a couple of ways to fix this.
1. Do the `.containsProbe` check after "maybe clone" part. But I don't think this is the best way because I think this mix of laziness and mutation could come up in a different place some day.
2. Change the value we're memoizing to be "containsChildProbe" or something since that can never change--then the `containsProbe` check is "am I a probe OR do I containChildProbe".

mwachs5:
i like the "contains child probe" formulation, will do that. And will add some tests specifically for this

mwachs5:
I am going with containsProbe means "is or contains a probe". Added a bunch of tests for this. I renamed the aggregate variable to be `elementsContainProbe` and optimized it for Vecs.

core/src/main/scala/chisel3/AggregateImpl.scala
```diff
@@ -233,6 +236,9 @@ private[chisel3] abstract class VecImpl[T <: Data] private[chisel3] (gen: => T,
 
   override def containsAFlipped = sample_element.containsAFlipped
 
+    /** Save this result, as it can be very expensive to determine this otherwise */
```
mwachs5, line 239: just realized my ` ./mill __.reformat` command failed, looking into it

jackkoenig (approved):
Looks great, thanks!

Changed milestone to 6.x because probes didn't exist until then.

PR #4661, opened 2025-02-04
Add publish flow for panama modules

unlsycn:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?


sequencer:
We experimentally publish the Panama project for locally experimental usage(not for all users now), but welcome for publishLocal and have a try. 

The API is subject change: we already have the Scala 3 version of panama related, since panama related projects is never published before, we might replace it without depreciation before the end of Q2.

sequencer (approved):

Avimitin:
publishLocal will invoke docJar target, which requires additional Java flags to be set:

```patch
diff --git a/circtpanamabinding/package.mill b/circtpanamabinding/package.mill
index 424db89f4..7a0cd57dc 100644
--- a/circtpanamabinding/package.mill
+++ b/circtpanamabinding/package.mill
@@ -12,7 +12,10 @@ object `package` extends RootModule with CIRCTPanamaBinding
 // Java Codegen for all declared functions.
 // All of these functions are not private API which is subject to change.
 trait CIRCTPanamaBinding extends HasJextractGeneratedSources with release.ChiselPublishModule {
+  override def javadocOptions = Task(super.javadocOptions() ++ Seq("--enable-preview", "--release", "21"))
+
   object utils extends Module {
+
     val architecture = System.getProperty("os.arch")
     val operationSystem = System.getProperty("os.name")

```

jackkoenig:
Looks like I was a few seconds too late to block merging, but that's fine, this can be done in a follow on PR.

Rather than changing this to use the published artifact which breaks the purpose of a build tool (now there is an implicit dependency that requires running publish followed by compile rather than the build tool tracking that dependency), is it possible to instead depend on the unipublish module? We will have to do some finagling here since unipublish is sort of a fake module and you'll need to get its jar on the classpath rather than its default compile classpath. An alternative approach is to figure out how the ivydeps are turned into the dependencies in the pom and insert the chisel published dependency there instead of ivydeps.

TLDR, please figure out how to use the build tool like a build tool and do not create dependencies outside of it.

sequencer:
cc @unlsycn CI failed in https://github.com/chipsalliance/chisel/actions/runs/13184372792/job/36803931937 please fix asap. 
I think we should separate compile module and add a transitive ivy dependency to follow @jackkoenig’s review.
Please do it in the another PR.

jackkoenig:
If CI's failing I'll just revert--y'all can fix and re-merge tomorrow or when you get to it.

sequencer:
sure! Thanks Jack!

PR #4665, opened 2025-02-05
[chiselsim] Add Default (non-Ephemeral) Simulator

seldridge:
Add a new simulator, DefaultSimulator, to Chiselsim.  This fills a major, long-standing hole in this API by adding a simulator that does not delete all the files that it creates after a test runs.

This uses a type class, `HasTestingDirectory`, to allow for control over where the directory is created.  A default implementation is provided which uses a nanosecond resolution timestamp to differentiate the tests. E.g., example output from running the new test included in this commit multiple times is:

    test_run_dir/
    └── DefaultSimulator
        ├── 2025-02-05T16-58-02.175175
        ├── 2025-02-05T16-58-11.941263
        └── 2025-02-05T16-58-17.721776

Note: a Scalatest-based type class implementation is intentionally _not_ provided as part of this PR.  This will be provided in a follow-on in such a way that Chiselsim does not inherit a dependency on Scalatest.  Ideally, users of Scalatest will use the future Scalatest-based type class implementation to get a subdirectory which is not a timestamp, but derived from the name of that part of the test.  I.e., a directory like:

    test_run_dir/
    └── DefaultSimulator
        ├── writes_files_to_a_permanent_directory_on_disk
        └── doesnt_delete_the_root_directory

#### Release Notes

Add `DefaultSimulator` to Chiselsim to provide a simulator which doesn't delete its build directory. This fills a long-standing hole where only the `EphemeralSimulator` was provided which will delete the build directory. The current directory structure is timestamp-based.

### DIFF
--- a/src/main/scala-2/chisel3/simulator/DefaultSimulator.scala
+++ b/src/main/scala-2/chisel3/simulator/DefaultSimulator.scala
@@ -3,11 +3,8 @@ package chisel3.simulator
 import svsim._
 import chisel3.RawModule
 import chisel3.util.simpleClassName
-import java.io.File
 import java.nio.file.{FileSystems, Files, Path}
 import java.time.LocalDateTime
-import java.time.format.DateTimeFormatter
-import scala.reflect.io.Directory
 
 /** This is a trait that can be mixed into a class to determine where
   * compilation should happen and where simulation artifacts should be written.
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM! Glad to see progress here

seldridge:
This was changed to "no release notes" since this API was removed before ever being released (even in a milestone release) with: https://github.com/chipsalliance/chisel/pull/4763

PR #4669, opened 2025-02-07
[Chiselsim] Add WithTestingDirectory

seldridge:
Add a trait, `WithTestingDirectory`, that will automatically create a test directory structure based on Scalatest information.  This provides a type class implementation of `HasTestingDirectory`, but with better Scalatest-specific directories as opposed to generic timestamp-based ones.

This necessarily introduces a provided ("compile only" to Mill) dependency on Scalatest for Chisel.  This prevents making all of Chisel pull in Scalatest and relies on the user to load it.

E.g., the included test will create the following directory structure:

```
build
└── WithTestingDirectorySpec
    └── A-test-suite-mixing-in-WithTestingDirectory
        ├── should-generate-a-directory-structure-derived-from-the-suite-and-test-name
        ├── should-generate-another-directory,-too
        ├── should-handle-CJK-characters,-e.g.,-好猫咪
        └── should-handle-emojis,-e.g.,-🚀
```

#### Release Notes

Add `WithTestingDirectory` Scalatest suite mix-in that automatically sets up a Scalatest-specific directory structure for your tests.

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -274,6 +274,8 @@ trait Chisel extends CrossSbtModule with HasScala2MacroAnno with HasScala2Plugin
 
   override def moduleDeps = super.moduleDeps ++ Seq(coreModule, svsimModule)
 
+  def compileIvyDeps = Agg(v.scalatest)
+
   object test extends SbtTests with TestModule.ScalaTest with ScalafmtModule {
     def ivyDeps = Agg(v.scalatest, v.scalacheck)
 
--- a/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
+++ b/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
@@ -32,7 +32,7 @@ trait WithTestingDirectory { self: TestSuite =>
     *
     * For different behavior, please override this in your test suite.
     */
-  def buildDir: String = "test-run-dir"
+  def buildDir: String = "build"
 
   // Assemble all the directories that should be created for this test.  This is
   // done by examining the test (via a fixture) and then setting a dynamic
--- a/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
+++ b/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
@@ -15,11 +15,11 @@ import scala.util.DynamicVariable
   * <buildDir>
   * └── <suite-name>
   *     └── <scope-1-name>
-  *         └── <scope-2-name>
-  *             └── ...
-  *                 └── <scope-n-name>
-  *                     ├── <test-1-name>
-  *                     └── <test-2-name>
+  *         └── ...
+  *             └── <scope-n-name>
+  *                 ├── <test-1-name>
+  *                 ├── ...
+  *                 └── <test-n-name>
   * }}}
   *
   * You may change the `buildDir` by overridding a method of the same name in
--- a/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
+++ b/src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator.scalatest
+
+import chisel3.simulator.HasTestingDirectory
+import java.nio.file.{FileSystems, Path}
+import org.scalatest.TestSuite
+import scala.util.DynamicVariable
+
+/** A mix-in for a Scalatest test suite that will setup the output directory for
+  * you.  E.g., this will create output directories for your tests like the
+  * following:
+  *
+  * {{{
+  * <buildDir>
+  * └── <suite-name>
+  *     └── <scope-1-name>
+  *         └── <scope-2-name>
+  *             └── ...
+  *                 └── <scope-n-name>
+  *                     ├── <test-1-name>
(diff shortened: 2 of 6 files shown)


src/test/scala/chiselTests/simulator/scalatest/WithTestingDirectorySpec.scala
```diff
@@ -0,0 +1,74 @@
...
+        file <- Seq(
+          directory.toFile.toString + "/workdir-default/Makefile",
+          directory.toFile.toString + "/primary-sources/Foo.sv"
+        )
+      ) {
+        info(s"found expected file: '$file'")
+        allFiles should contain(file)
+      }
+    }
+    it("should generate another directory, too") {
```
jackkoenig, line 52: What if the test uses a non-ASCII character like a Chinese character or an emoji?

Do we want to `java.text.Normalizer` it or just leave it be?

src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
```diff
@@ -0,0 +1,62 @@
...
+    /** Return the test name with some sanitization applied. */
+    final def getTestName = testName.value.map(_.replaceAll(" ", "_").replaceAll("\\W+", ""))
+
+    override def getDirectory(testClassName: String): Path = FileSystems
+      .getDefault()
+      .getPath(buildDir, self.suiteName +: getTestName: _*)
+
+  }
+
+}
```
jackkoenig, line 62: This is something you won't be able to test in our normal flow, but I'd like to see an enhanced error message for what happens if a user tries to use this but forgets to add ScalaTest as a dependency. You can experiment by publishing this locally and trying to call it from Scala-CLI or just making an extra build unit--but in either case try to extend this trait and use it but forget to add Scalatest as a dependency.

build.mill
```diff
@@ -274,6 +274,8 @@ trait Chisel extends CrossSbtModule with HasScala2MacroAnno with HasScala2Plugin
 
   override def moduleDeps = super.moduleDeps ++ Seq(coreModule, svsimModule)
 
+  def compileIvyDeps = Agg(v.scalatest)
+
```
jackkoenig, line 278: Just to sanity check (because in my experience, publishing stuff is finicky and especially so since unipublish is a Jack Koenig special), can you check that the scalatest dependency is showing up properly as "provided" in the unipublish POM?

src/test/scala/chiselTests/simulator/scalatest/WithTestingDirectorySpec.scala
seldridge: Maybe. Let me play with this. I'm also kind of of the opinion that maybe we don't mangle this at all. Emojis or CJK characters all seem reasonable to _not_ mangle as I think most filesystems will support those?

src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
jackkoenig: Basically I think we'll want to catch an Linkage exception or whatever somewhere and just add some more context suggesting the user add the dependency on ScalaTest.

src/test/scala/chiselTests/simulator/scalatest/WithTestingDirectorySpec.scala
jackkoenig: I like the ` ` to `_` mangling and I agree I'm not sure we should do any other mangling, just wanted to float the idea. It's also reasonable for users to avoid such characters if they do cause problems but obviously spaces are to be expected.

seldridge: Switched to mangle all whitespace to `-` in: f39550b8b This includes tests that both emoji and CJK characters (technically just Chinese (or Japanese Kanji)) work.

jackkoenig: I don't object by why `-` over `_`?

src/main/scala-2/chisel3/simulator/scalatest/WithTestingDirectory.scala
seldridge: (Summary of an offline discussion investigating this:) I looked at what the error would look like if you don't include scalatest as a dependency. Luckily, the fact that this has a self-type on `org.scalatest.TestSuite` means that you will get a compile-time error if you don't try to bring `TestSuite` into scope. If you then try to clear the self-type without providing scalatest as a dependency, you'll get a standard "scalatest is not a member of package org".

tl;dr: the self-type means that other errors will come up before you can hit this. 🎊

build.mill
seldridge: There was a missing part to include this in the unipublish flow. This is fixed via: https://github.com/chipsalliance/chisel/pull/4669/commits/c8853914dc5d4afcf82ac4ac3378878f850d56c3

With that in place, scalatest is marked as "provided" scope in the POM. If you then try to fetch this with `cs fetch -e 2.13.15 org.chipsalliance::chisel:<your version>` then this will _not_ try to fetch scalatest.

To test the inverse, I changed the dependency from a `compileIvyDeps` to an `ivyDeps` and it _would_ try to fetch it. (I was not able to show the difference through manual POM edits.)

jackkoenig (approved):
Looks awesome, nice!

PR #4670, opened 2025-02-07
Add publish flow for panama modules

unlsycn:
Revision of #4661 which eliminate implicit dependency

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification
- API deprecation
- Backend code generation
- Performance improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/panamaconverter/package.mill
+++ b/panamaconverter/package.mill
@@ -6,6 +6,7 @@ import mill.scalalib.scalafmt._
 import mill.define.Cross
 
 import build._
+import build_.release
 
 object `package` extends RootModule {
   // https://github.com/com-lihaoyi/mill/issues/3693
@@ -17,12 +18,16 @@ trait PanamaConverter
     with HasPanamaOMModule
     with CrossModuleBase
     with HasScala2Plugin
-    with ScalafmtModule {
+    with ScalafmtModule
+    with release.ChiselPublishModule {
   def millSourcePath = super.millSourcePath / os.up
 
   def panamaOMModule = panamaom.cross(crossScalaVersion)
   def chiselModule = chisel(crossScalaVersion)
   def pluginModule = plugin.cross(crossScalaVersion)
 
-  override def moduleDeps = super.moduleDeps ++ Some(chiselModule)
+  // use to be propagated to downstream modules
+  override def moduleDeps = super.moduleDeps ++ Some(unipublish)
+
+  override def compileClasspath = Task { super.compileClasspath() ++ unipublish.localClasspath() }
 }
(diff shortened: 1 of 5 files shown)


panamaconverter/package.mill
```diff
@@ -17,12 +18,13 @@ trait PanamaConverter
...
+    with ScalafmtModule
+    with release.ChiselPublishModule {
   def millSourcePath = super.millSourcePath / os.up
 
   def panamaOMModule = panamaom.cross(crossScalaVersion)
   def chiselModule = chisel(crossScalaVersion)
   def pluginModule = plugin.cross(crossScalaVersion)
 
-  override def moduleDeps = super.moduleDeps ++ Some(chiselModule)
+  override def compileModuleDeps = super.compileModuleDeps ++ Some(chiselModule)
```
sequencer, line 29: Test of panama should explicitly add `chiselModule` to `moduleDeps`

sequencer (approved):

panamaconverter/package.mill
unlsycn: Propagating dependencies manually requires the user to add chisel as a dependency to all modules that depend on panama, which is cumbersome.
Maybe we can inject the classpath to the module at compile time and use unipublish to propagate the Ivy dependency downstream.

unlsycn:
https://github.com/chipsalliance/chisel/actions/runs/13197937262/job/36844288755
The publish CI failed since it runs with jdk11 which cannot work with `--release 21`, is it possible to run CI under jdk21?

jackkoenig:
> The publish CI failed since it runs with jdk11 which cannot work with --release 21, is it possible to run CI under jdk21?

We can, but that alone won't fix the problem because it will then try to publish the Scala 3.3.4 cross versions which obviously won't work. What is the intent of adding this publishing flow? Is it just to make it possible to publish these things locally or is the intent to publish the SNAPSHOTs to Sonatype? Which projects need to be published?

sequencer:
I don’t think we should publish the panama related api to maven repo in this stage. We only use the locally published version. I think the publish ci should only publish the selected modules rather than using _ to select all.

jackkoenig:
It doesn't use `_`, the CI release plugin just grabs all Modules that mixin `CiReleaseModule`: https://github.com/ckipp01/mill-ci-release/blob/4cc52ec847feb0383480125d0be7041c25c286b4/plugin/src/io/kipp/mill/ci/release/CiReleaseModule.scala#L252.

Annoyingly, mill-ci-release is pretty inflexible on these sorts of things--as far as I can tell there is no way to customize what gets published other than making sure only the right things mixin `CiReleaseModule`. Ultimately maybe we should bring this logic in tree so we can customize it, but short term might just be to have `ChiselPublishModule` only extend `PublishModule` and then have only the plugin and unipublish mixin `CiReleaseModule`.

jackkoenig:
I pushed a fix to main

PR #4677, opened 2025-02-11
[chiselsim] Remove test class name from directory

seldridge:
Remove an argument from the `getDirectory` method of the `HasTestingDirectory` type class.  This was intended to capture the name of the class.  However, this didn't actually work.  Remove this as it is both: (1) not necessary as if users of this are likely using the Scalatest type class implementation which does support this and (2) it simplifies downstream methods as this doesn't need to be routed around.

This is changing a public method that was never in a release. Hence, there are no release notes.


jackkoenig (approved):
Yeah this is better

PR #4678, opened 2025-02-11
[chiselsim] Add, use HasSimulator type class

seldridge:
Add a new type class, `HasSimulator`, which provides a method to get a
simulator.  Provide two implementations: one for Verilator and one for
VCS.  Additionally, provide a low-priority default that maps to Verilator.

This is working towards refactoring Chiselsim to improve the
customizability of its APIs while not dramatically changing how a user
interacts with it.  Additionally, this is trying to work towards providing
Scalatest-specific implementations of `HasSimulator` which allow the
simulator to be chosen on the command line.  Much of the code for this is
currently sequestered in either ChiselSpec (part of the tests which are
not published) or in internal code.

Refactor the Chiselsim's `DefaultSimulator` to no longer create an
instance of a private Verilator simulator class.  Instead, use what is
provided by a `HasSimulator` type class implementation.

This is done to slim down `DefaultSimulator` (and transitively
`EphemeralSimulator`) to essentially nothing other than objects that
define convenience APIs for running simulation like the existing
`simulate` method.

#### Release Notes

- Add `HasSimulator` type class to Chiselsim. This allows for customization of the simulator used by `DefaultSimulator`.

### DIFF
--- a/src/main/scala/chisel3/simulator/DefaultSimulator.scala
+++ b/src/main/scala/chisel3/simulator/DefaultSimulator.scala
@@ -19,23 +19,11 @@ import java.nio.file.Files
   */
 object DefaultSimulator extends PeekPokeAPI {
 
-  private class DefaultSimulator(val workspacePath: String) extends Simulator[verilator.Backend] {
-    override val backend = verilator.Backend.initializeFromProcessEnvironment()
-    override val tag = "default"
-    override val commonCompilationSettings = CommonCompilationSettings()
-    override val backendSpecificCompilationSettings = verilator.Backend.CompilationSettings()
-  }
-
   def simulate[T <: RawModule](
     module:       => T,
     layerControl: LayerControl.Type = LayerControl.EnableAll
-  )(body: (T) => Unit)(implicit testingDirectory: HasTestingDirectory): Unit = {
-
-    val simulator = new DefaultSimulator(
-      workspacePath = Files.createDirectories(testingDirectory.getDirectory).toString
-    )
-
-    simulator.simulate(module, layerControl)({ module => body(module.wrapped) }).result
+  )(body: (T) => Unit)(implicit hasSimulator: HasSimulator, testingDirectory: HasTestingDirectory): Unit = {
+    hasSimulator.getSimulator.simulate(module, layerControl)({ module => body(module.wrapped) }).result
   }
 
 }
--- a/src/main/scala/chisel3/simulator/HasSimulator.scala
+++ b/src/main/scala/chisel3/simulator/HasSimulator.scala
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator
+
+import java.nio.file.Files
+
+/** Type class for providing a simulator. */
+trait HasSimulator {
+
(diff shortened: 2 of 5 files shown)


src/main/scala-2/chisel3/simulator/HasSimulator.scala
```diff
@@ -0,0 +1,54 @@
...
+        override val backendSpecificCompilationSettings = svsim.vcs.Backend.CompilationSettings()
+        override val workspacePath = Files.createDirectories(testingDirectory.getDirectory).toString
+      }
+  }
+
+  /** Low-priority default implementation of [[HasSimulator]] that uses Verilator.
+    * This is the default that will be used if the user does provide an
+    * alternative.
+    */
+  implicit val default: HasSimulator = verilator
```
jackkoenig, line 52:
```suggestion
  implicit def default: HasSimulator = verilator
```
Nit, there's no reason for this to be a val

src/main/scala-2/chisel3/simulator/HasSimulator.scala
```diff
@@ -0,0 +1,54 @@
...
+    */
+  def getSimulator(implicit testingDirectory: HasTestingDirectory): Simulator[_]
+
+}
+
+/** Type class implementations of [[HasSimulator]]. */
+object HasSimulator {
+
+  /** A [[HasSimulator]] implementation for a Verilator simulator. */
+  val verilator: HasSimulator = new HasSimulator {
```
jackkoenig, line 21: Similarly to the below, is there any real reason to make these `val` vs. `def`?

jackkoenig (commented):
It would be good to add a test that shows explicitly "overriding" the simulator to verilator. This obviously is no different than using the default but it shows the API. As written that would look something like:
```scala
class MyTestSpec ... {
  implicit val simulator: HasSimulator = chisel3.simulator.HasSimulator.verilator
  ...
}
```
Or in Scala 3
```scala
class MyTestSpec ... {
  given HasSimulator = chisel3.simulator.HasSimulator.verilator
  ...
}
```

I think this is reasonable, but an alternative could be to move `verilator` and `vcs` to some other object, and make them implicits. Then the override would be:
```scala
import chisel3.simulator.HasSimulator.simulators.verilator
```

I am not sure if this is better, but it is something to consider.

src/main/scala-2/chisel3/simulator/HasSimulator.scala
seldridge: No reason. Changed to a def in a312ee898.

src/main/scala-2/chisel3/simulator/HasSimulator.scala
seldridge: Changed in: a312ee898

seldridge:
> I think this is reasonable, but an alternative could be to move verilator and vcs to some other object, and make them implicits. Then the override would be:

I like this approach. I added that exact structure in https://github.com/chipsalliance/chisel/pull/4678/commits/51e838eeaa2e103f13ea581290ccf294093cd044

There is now a minimal test in: https://github.com/chipsalliance/chisel/pull/4678/commits/386d24dbe0537610c2ae2bd129b14fa9855b856c

PR #4680, opened 2025-02-12
[chiselsim] Add SimulatorAPI, reduce DefaultSimulator to nothing

seldridge:
Add a new `SimulatorAPI` trait which provides methods for running
simulations.  This will submsume the existing `DefaultSimulator.simulate`
method and extends it with an "ABI" for how to startup simulations of
Chisel-generated Verilog.

This API includes two methods:

1. simulateRaw -- This is only usable with `RawModule`s where there is
no information about the clock and reset.  This is the same as
`DefaultSimulator.simulate`.

2. simulate -- This is only usable with `Module`s where there is a
defined clock and reset.  This is the method which implements an "ABI"
startup sequence which avoids problems with simulators, time zero
behavior, and race conditions.

Remove all methods from `DefaultSimulator` and replace them with methods
provided by the `SimulatorAPI`.  This is change to the unreleased
`simulate` method for `DefaultSimulator` in that this will now require a
`Module` and uses the standard ABI bringup procedure.
`EphemeralSimulator` is left unchanged.

#### Release Notes

Add a new ChiselSim trait, `SimulatorAPI`, which adds methods for running simulations.

### DIFF
--- a/src/main/scala/chisel3/simulator/DefaultSimulator.scala
+++ b/src/main/scala/chisel3/simulator/DefaultSimulator.scala
@@ -17,13 +17,4 @@ import java.nio.file.Files
   * simulate(new MyChiselModule()) { module => ... }
   * }}}
   */
-object DefaultSimulator extends PeekPokeAPI {
-
-  def simulate[T <: RawModule](
-    module:       => T,
-    layerControl: LayerControl.Type = LayerControl.EnableAll
-  )(body: (T) => Unit)(implicit hasSimulator: HasSimulator, testingDirectory: HasTestingDirectory): Unit = {
-    hasSimulator.getSimulator.simulate(module, layerControl)({ module => body(module.wrapped) }).result
-  }
-
-}
+object DefaultSimulator extends PeekPokeAPI with SimulatorAPI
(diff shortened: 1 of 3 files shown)


seldridge:
@jackkoenig: I'm curious if you have any ideas on how to package up this API. Two options occurred to me:

1. Just do trait mix-in:

``` scala
class Foo extneds AnyFunSpec with Matchers with SimulatorAPI with PeekPokeAPI {
  /** body */
}
```

2. Wildcard import pattern:

``` scala
import chisel3.simulator.DefaultSimulator._
```

If (2), then what should the package/object be?

3. Something else?

I'd plan to keep `EphemeralSimulator` around, but deprecate it.

seldridge:
Maybe for (3) we call this `ChiselSim`. I.e., "ChiselSim" is just the collection of APIs that you use for running simulations. I'd then propose two slightly different packagings of it:

``` scala
package chisel3.simulator

trait ChiselSim extends PeekPokeAPI with SimulatorAPI
```

``` scala
package chisel3.simulator.scalatest

trait ChiselSim extends PeekPokeAPI with SimulatorAPI with WithTestingDirectory
```

Note: this can either or instead use the wildcard import approach with some slight changes to how `WithTestingDirectory` exposes its implicit in an object instead of in a trait.

PR #4682, opened 2025-02-12
Add formal contracts

fabianschuiki:
Add the `Contract` API which allows users to express the pre and post-conditions of a piece of hardware. These can be used during formal verification to prove that the hardware upholds the conditions laid out in the contract. Additionally, contracts may be used to implement a divide-and-conquer strategy to simplify larger proofs since they act like a cutpoint and an assume-guarantee substitution.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala-2/chisel3/FormalContractIntf.scala
+++ b/core/src/main/scala-2/chisel3/FormalContractIntf.scala
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.SourceInfo
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+private[chisel3] trait FormalContract$Intf extends FormalContract$VirtualMethods {
+
+  /** Create a `contract` block with one or more arguments and results. */
+  def apply(head: Data, tail: Data*): (Any => Unit) => Any =
+    macro FormalContractMacro.apply_impl
+}
+
+private[chisel3] object FormalContractMacro {
+
+  /** A macro for `FormalContract` to allow for contracts with a single argument
+    * and contracts with a tuple of arguments to be created. The macro packages
+    * the sequence of arguments up into a tuple (`Tuple1` for single arguments,
+    * and the corresponding `TupleN` for multiple arguments). Without this
+    * macro, the contracts would always operate on tuples (`Product`), forcing
+    * the user to use
+    * {{{
+    * val Tuple1(a) = FormalContract(Tuple1(b)) { case Tuple1(x) => ... }
+    * }}}
+    * for contracts with only a single argument.
+    *
+    * A call to `FormalContract(a: T)` becomes:
+    * {{{
+    * FormalContract.mapped(
+    *   Seq(a),
+    *   (values => values(0).asInstanceOf[T])
+    * )
+    * }}}
+    *
+    * A call to `FormalContract(a: A, b: B, c: C)` becomes:
+    * {{{
+    * FormalContract.mapped(
+    *   Seq(a, b, c),
+    *   (values => (
+    *     values(0).asInstanceOf[A],
+    *     values(1).asInstanceOf[B],
+    *     values(2).asInstanceOf[C]
+    *   ))
+    * )
+    * }}}
+    */
+  def apply_impl(c: whitebox.Context)(head: c.Expr[Data], tail: c.Expr[Data]*): c.Expr[(Any => Unit) => Any] = {
+    import c.universe._
+    val args = head +: tail
+    val mapping =
+      q"(values => (..${args.zipWithIndex.map { case (arg, i) => q"values(${i}).asInstanceOf[${arg.tree.tpe}]" }}))"
+    val result = q"FormalContract.mapped(Seq(..${args.map(_.tree)}), $mapping)(_)"
+    c.Expr[(Any => Unit) => Any](result)
+  }
+}
--- a/core/src/main/scala-3/chisel3/FormalContractIntf.scala
+++ b/core/src/main/scala-3/chisel3/FormalContractIntf.scala
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import scala.quoted.*
+
+private[chisel3] trait FormalContract$Intf extends FormalContract$VirtualMethods {
+
+  /** Create a `contract` block with one or more arguments and results. */
+  transparent inline def apply[T <: Data](inline args: T*): Any =
+    ${ FormalContractMacro('args) }
+}
+
+private[chisel3] object FormalContractMacro {
+
+  /** A macro for `FormalContract` that allows for contracts to be built while
+    * preserving the types of individual elements passed to the contract. This
+    * allows the types of the contract's results to match the type of the
+    * arguments passed to the contract.
+    *
+    * A call to `FormalContract(a: T)` becomes:
+    * {{{
+    * FormalContract.mapped(
+    *   Seq(a),
+    *   (values => values(0).asInstanceOf[T])
+    * )
+    * }}}
+    *
+    * A call to `FormalContract(a: A, b: B, c: C)` becomes:
+    * {{{
+    * FormalContract.mapped(
+    *   Seq(a, b, c),
+    *   (values => (
+    *     values(0).asInstanceOf[A],
+    *     values(1).asInstanceOf[B],
+    *     values(2).asInstanceOf[C]
+    *   ))
+    * )
+    * }}}
+    */
+  def apply(args: Expr[Seq[Data]])(using Type[Seq[Data]])(using Quotes): Expr[Any] = {
+    val elements = args match {
+      case Varargs(elements) => elements
+      case _                 => quotes.reflect.report.errorAndAbort("Expected a compile-time known list of arguments")
+    }
+    // In the following we use a `match case '{ _: t}` to bind the type of the
+    // left-hand side of the match to `t`, which we can then refer to within the
+    // match to cast elements to the appropriate type. This allows us to get the
+    // type of the element if there's only one, or we construct a tuple out of
+    // all the elements as a way to construct the corresponding tuple type.
+    if (elements.size == 0) {
+      quotes.reflect.report.errorAndAbort("use `FormalContract` without `()` instead")
+    } else if (elements.size == 1) {
+      // Calls of the form `apply(x: T)` require a mapping like:
+      // ```
+      // (values: Seq[Data]) => values.last.asInstanceOf[T]
+      // ```
+      elements.last match {
+        case '{ $x: t } =>
+          val mapping = '{ (values: Seq[Data]) => values.last.asInstanceOf[t] }
+          '{ FormalContract.mapped(${ Expr.ofSeq(elements) }, $mapping) }
+      }
+    } else {
+      // Calls of the form `apply(x0: T0, x1: T1, ..., xi: Ti)` require a
+      // mapping like:
+      // ```
+      // (values: Seq[Data]) => (
+      //   values(0).asInstanceOf[T0],
+      //   values(1).asInstanceOf[T1],
+      //   ...,
+      //   values(i).asInstanceOf[Ti],
+      // )
+      // ```
+      Expr.ofTupleFromSeq(elements) match {
+        case '{ $x: t } =>
+          val mapping = '{ (values: Seq[Data]) =>
+            ${
+              Expr.ofTupleFromSeq((0 until elements.size).map(i => '{ values(${ Expr(i) }) }))
+            }.asInstanceOf[t]
+          }
+          '{ FormalContract.mapped(${ Expr.ofSeq(elements) }, $mapping) }
+      }
+    }
+  }
+}
--- a/core/src/main/scala/chisel3/FormalContract.scala
+++ b/core/src/main/scala/chisel3/FormalContract.scala
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import chisel3.experimental.SourceInfo
+import chisel3.internal._
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl.ir._
+
+/** Create a `contract` block.
+  *
+  * Outside of verification uses, the arguments passed to the contract are
+  * simply returned as contract results. During formal verification, the
+  * contract may act as a cutpoint by returning symbolic values for its
+  * results and using the contract body as constraint.
+  *
+  * Within the contract body, `RequireProperty` and `EnsureProperty` can be
+  * used to describe the pre and post-conditions of the contract. During
+  * formal verification, contracts can be used to divide large formal checks
+  * into smaller pieces by first asserting some local properties hold, and
+  * then assuming those properties hold in the context of the larger proof.
+  *
+  * @example
+  * The following is a contract with no arguments.
+  * {{{
+  * FormalContract {
+  *   RequireProperty(a >= 1)
+  *   EnsureProperty(a + a >= 2)
+  * }
+  * }}}
+  *
+  * The following is a contract with a single argument. It expresses the fact
+  * that the hardware implementation `(a << 3) + a` is equivalent to `a * 9`.
+  * During formal verification, this contract is interpreted as:
+  *
+  * - `assert((a << 3) + a === a * 9)` as a proof that the contract holds.
+  * - `assume(b === a * 9)` on a new symbolic value `b` as a simplification of
+  *   other proofs.
+  *
+  * {{{
+  * val b = FormalContract((a << 3) + a) { b =>
+  *   EnsureProperty(b === a * 9)
+  * }
+  * }}}
+  *
+  * The following is a contract with multiple arguments. It expresses the fact
+  * that a carry-save adder with `p`, `q`, and `r` as inputs reduces the number
+  * of sum terms from `p + q + r` to `c + s` but doesn't change the sum of the
+  * terms. During formal verification, this contract is interpreted as:
+  *
+  * - `assert(c + s === p + q + r)` as a proof that the carry-save adder indeed
+  *   compresses the three terms down to only two terms which have the same sum.
+  * - `assume(u + v === p + q + r)` on new symbolic values `u` and `v` as a
+  *   simplification of other proofs.
+  *
+  * {{{
+  * val s = p ^ q ^ r
+  * val c = (p & q | (p ^ q) & r) << 1
+  * val (u, v) = FormalContract(c, s) { case (u, v) =>
+  *   EnsureProperty(u + v === p + q + r)
+  * }
+  * }}}
+  */
+object FormalContract extends FormalContract$Intf {
+
+  /** Create a formal contract from a sequence of expressions.
+    *
+    * The `mapping` function is used to convert the `Seq[Data]` to a
+    * user-defined type `R`, which is then passed to the contract body and
+    * returned as a result.
+    *
+    * This function is mainly intended for internal use, where the `mapping`
+    * function can be used to map the args sequence to a tuple that is more
+    * convenient for a user to work with.
+    *
+    * @example
+    * Example with identity mapping. The `Seq(a, b)` is passed to `withSeqArgs`
+    * unmodified, and the contract returns a `Seq[Data]` result.
+    * {{{
+    * def withSeqArgs(args: Seq[Data]): Unit = ...
+    * val seqResult: Seq[Data] =
+    *   FormalContract.mapping(Seq(a, b), _ => _)(withSeqArgs)
+    * }}}
+    *
+    * Example with a mapping from sequence to a tuple. The `Seq(a, b)` is mapped
+    * to a `(UInt, UInt)` tuple through `mapToTuple`, which is then passed to
+    * `withTupleArgs`, and the contract returns a corresponding `(UInt, UInt)`
+    * result.
+    * {{{
+    * def mapToTuple(args: Seq[Data]): (UInt, UInt) = {
+    *   // check number of args and types, convert to tuple
+    * }
+    * def withTupleArgs(args: (UInt, UInt)): Unit = ...
+    * val tupleResult: (UInt, UInt) =
+    *   FormalContract.mapping(Seq(a, b), mapToTuple)(withTupleArgs)
+    * }}}
+    */
+  override def mapped[R](args: Seq[Data], mapping: Seq[Data] => R)(
+    body: R => Unit
+  )(implicit sourceInfo: SourceInfo): R = {
+    // Create a sequence of contract results, one for each argument.
+    val results = args.map { arg =>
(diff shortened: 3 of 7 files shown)


core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+package chisel3
+
+import chisel3.experimental.SourceInfo
+import chisel3.internal._
+import chisel3.internal.Builder.pushCommand
+import chisel3.internal.firrtl.ir._
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+object Contract {
```
mwachs5, line 12: can this be called `FormalContract`? A contract is a very abstract term, and we also had the idea of chisel `Interface` providing a different kind of contract...

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    * into smaller pieces by first asserting some local properties hold, and
+    * then assuming those properties hold in the context of the larger proof.
+    *
+    * @example
+    * {{{
+    * // Contract with single argument: "(a<<3)+a is equivalent to a*9".
+    * // During formal verification, this is interpreted as:
+    * // - "assert((a<<3)+a === a*9)" as a proof that the contract holds
+    * // - "assume(b === a*9)" as a simplification of other proofs
+    * val b = Contract((a<<3)+a) { b =>
```
mwachs5, line 53: super nit but could we format these examples like scalafmt would format them in this repo... basically more whitespace

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    * val csa_s = p^q^r
+    * val csa_c = (p&q | (p^q)&r) << 1
+    * val (u,v) = Contract(csa_c, csa_s) { case (u,v) =>
+    *   EnsureProperty(u+v === p+q+r)
+    * }
+    * }}}
+    */
+  def apply(head: Data, tail: Data*): (Any => Unit) => Any = macro apply_impl
+
+  // This uses a macro internally to allow for contracts with a single argument
```
mwachs5, line 73: is this one of those "here be dragons" type APIs where it's technically public but we don't want peopel to call it? There is a way with scaladoc to add a `@group` macro that makes it move down in the scaladoc... can we do that here?

src/test/scala-2/chiselTests/ContractSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+      // CHECK: input a : UInt<1>
+      // CHECK: contract :
+      // CHECK:   intrinsic(circt_verif_assert, a)
+      """)
+  }
+
+  it should "support contracts with a single operand" in {
+    class Foo extends RawModule {
+      val a = IO(Input(Bool()))
+      val b = Contract(a) { x =>
```
mwachs5, line 36: i am missing the basic user understanding of when something should go into the argument list of the `Contract` or just be captured in the closure of the body.

core/src/main/scala-2/chisel3/Contract.scala
mwachs5, line 53: what is the user intuition for why you put the `(a << 3) + a` term into the argument list for the Contract but the `a*9` term is only used in the body of the contract?

mwachs5, line 53: kind of a nit but using the same variable `b` both inside and as the return value of the contract is a little confusing. What is the point of the return value `b`, how does one use that returned value?

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    * // Contract with multiple arguments:
+    * //   "carry-save adder reduces number of terms but doesn't change sum"
+    * // During formal verification, this is interpreted as:
+    * // - an assertion that the carry-save adder indeed compresses the three
+    * //   terms p/q/r down to only two terms u/v which have the same sum
+    * // - symbolic values u/v constrained to sum up to the same value as p/q/r,
+    * //   as a simplification of other proofs
+    * val csa_s = p^q^r
+    * val csa_c = (p&q | (p^q)&r) << 1
+    * val (u,v) = Contract(csa_c, csa_s) { case (u,v) =>
```
mwachs5, line 66: again, what is the point of `val (u, v)` that is returned here? what are they and what do you do with them?

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    * }}}
+    */
+  def apply()(body: => Unit)(implicit sourceInfo: SourceInfo): Unit = {
+    mapped[Unit](Seq(), _ => ())(_ => body)
+  }
+
+  /** Create a `contract` block with one or more arguments and results.
+    *
+    * Outside of verification uses, the arguments passed to the contract are
+    * simply returned as contract results. During formal verification, the
```
mwachs5, line 37: what are "the contract results"? what does that mean?

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    val contract = Builder.pushCommand(new DefContract(sourceInfo, results, args.map(_.ref(sourceInfo))))
+
+    // Map the sequence of results to the user-defined type `R`, pass it to the
+    // body to build the contract body, and also return it as the result of the
+    // contract as a whole.
+    val mapped = mapping(results)
+    Builder.forcedUserModule.withRegion(contract.region)(layer.elideBlocks(body(mapped)))
+    mapped
+  }
+}
```
jackkoenig, line 114: What about Scala 3? The code that can be shared should be in `src/main/scala`. The macro is cool and I do think it's better, but if you just did the annoying thing and write the method for 1 to N arguments (e.g. [DataProduct support for Tuples](https://github.com/chipsalliance/chisel/blob/e20069ee8fd73ffb036a2a2cc17c7b3a9956dd3c/core/src/main/scala/chisel3/experimental/dataview/DataProduct.scala#L125)), then it would work for both Scala 2 and 3.

core/src/main/scala-2/chisel3/Contract.scala
fabianschuiki: Yes definitely, I can changed that to `FormalContract` :+1:

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+    *
+    * @example
+    * {{{
+    * Contract() {
+    *   RequireProperty(a >= 1)
+    *   EnsureProperty(a + a >= 2)
+    * }
+    * }}}
+    */
+  def apply()(body: => Unit)(implicit sourceInfo: SourceInfo): Unit = {
```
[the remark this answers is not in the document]
fabianschuiki: Good point, applied that 👍

core/src/main/scala-2/chisel3/Contract.scala
fabianschuiki: The `apply` should be fine, that's what a user is supposed to call. The `apply_impl` is just the macro impl which we could hide in a group. I added the pattern as we use for `@group SourceInfoTransformMacro`.

core/src/main/scala-2/chisel3/Contract.scala
```diff
@@ -0,0 +1,114 @@
...
+      q"(values => (..${args.zipWithIndex.map { case (arg, i) => q"values(${i}).asInstanceOf[${arg.tree.tpe}]" }}))"
+    val result = q"Contract.mapped(Seq(..${args.map(_.tree)}), $mapping)(_)"
+    c.Expr[(Any => Unit) => Any](result)
+  }
+
+  /** Create a contract from a sequence of expressions.
+    *
+    * The `mapping` function is used to convert the `Seq[Data]` to a
+    * user-defined type `R`, which is then passed to the contract body and
+    * returned as a result.
```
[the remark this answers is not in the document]
fabianschuiki: Added

core/src/main/scala-2/chisel3/Contract.scala
fabianschuiki: Good call. I've move the common code into `scala/chisel3` and added an explicit `scala-3/chisel3` with the Scala 3 specific stuff.

core/src/main/scala-2/chisel3/Contract.scala
fabianschuiki: The example is a bit simplistic, but you would pass the actual implementation of your circuit, for example a radix-4 Booth multiplier, to the contract as an argument. The rationale is that you want the contract to act as a cutpoint, detaching the implementation and replacing it with a symbolic value during formal verification. You'd then use a simpler expression inside the contract to characterize your circuit, for example `a * b` for your multiplier.

fabianschuiki: The body of the contract wants to talk about the result of the contract, which is either what you pass into the argument list (during synthesis and when checking that your hardware upholds the contract), or it's a symbolic value (during formal verification when you assume that the contract holds). There isn't really a good way to do this in Scala... so the contract passes the value it produces as a result to the body as well.

core/src/main/scala-2/chisel3/Contract.scala
fabianschuiki: This is due to the cutpoint behaviour of a contract during formal verification. When doing formal verification, the contract doesn't just forward its argument list to its results, but instead creates symbolic values as its results that are then constrained by the contract body. (Which is why the body takes arguments as well: it needs to constrain either the hardware, or the symbolic value, depending on which part you're verifying.)

core/src/main/scala-3/chisel3/FormalContract.scala
```diff
@@ -0,0 +1,18 @@
...
+    ${ FormalContractMacro('args) }
+}
+
+object TestFormalContract {
+  val a = IO(Input(Clock()))
+  val b = IO(Input(UInt(42.W)))
+  FormalContract {}
+  val x: Clock = FormalContract(a) { case a => }
+  val y: (Clock, UInt) = FormalContract(a, b) { case (a, b) => }
+}
```
seldridge, line 18: Is this a test or something that should be published?

core/src/main/scala-3/chisel3/FormalContractMacro.scala
```diff
@@ -0,0 +1,77 @@
...
+    * allows the types of the contract's results to match the type of the
+    * arguments passed to the contract.
+    *
+    * A call to `FormalContract(a: T)` becomes:
+    * ```
+    * FormalContract.apply(
+    *   Seq(a),
+    *   (values => values(0).asInstanceOf[T])
+    * )
+    * ```
```
seldridge, line 20: Does `` ``` `` work in Scala3? This used to have to be `{{{` and `}}}` (as the PR's later Scala-generic examples do)?

core/src/main/scala/chisel3/FormalContractImpl.scala
```diff
@@ -0,0 +1,112 @@
...
+  * // - an assertion that the carry-save adder indeed compresses the three
+  * //   terms p/q/r down to only two terms u/v which have the same sum
+  * // - symbolic values u/v constrained to sum up to the same value as p/q/r,
+  * //   as a simplification of other proofs
+  * val csa_s = p ^ q ^ r
+  * val csa_c = (p & q | (p ^ q) & r) << 1
+  * val (u, v) = FormalContract(csa_c, csa_s) { case (u, v) =>
+  *   EnsureProperty(u + v === p + q + r)
+  * }
+  * }}}
```
seldridge, line 53: Nit: Could you break this up into three distinct examples with prose introduction as opposed to prose comments?

core/src/main/scala/chisel3/FormalContractImpl.scala
```diff
@@ -0,0 +1,112 @@
...
+  */
+private[chisel3] trait ObjectFormalContractImpl {
```
seldridge, line 55: This is fantastic Scaladoc. It's on a private trait, though. Where is it expected to show up? Is there some other place it should go?

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -349,6 +349,13 @@ case class Conditionally(info: Info, pred: Expression, conseq: Statement, alt: S
     with HasInfo
     with UseSerializer
 
+case class DefContract(info: Info, names: Seq[String], exprs: Seq[Expression], body: Statement)
+    extends Statement
+    with HasInfo
+    with UseSerializer {
+  require(names.size == exprs.size)
```
seldridge, line 356: Please include a string providing more information about this. (It's relatively self-explanatory, admittedly...)

firrtl/src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -234,6 +241,25 @@ object Serializer {
...
             indent -= 1
+          case ContractBegin(info, names, exprs) =>
+            doIndent()
+            b ++= "contract"
+            if (names.nonEmpty) {
+              b ++= " "
+              names.zipWithIndex.foreach { case (name, idx) =>
+                if (idx > 0) b ++= ", "
+                b ++= name
+              }
```
seldridge, line 252: Maybe try the following:

```suggestion
              b ++= names.mkString(" ", ", ", "")
```

Ditto for the expressions.

Scala's `mkString` is basically `llvm::interleave*`. It has two forms: a one-argument form which interleaves the argument and a three argument form which has a "start", "interleave", and "end".

There's a slight risk to this as it does create the string before passing it to the `StringBuilder`. However, I think that's probably okay here given that these are small. (I would not advise using `mkString` for interleaving newlines between modules, e.g.)

src/test/scala-2/chiselTests/FormalContractSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+        AssertProperty(a)
+      }
+    }
+    generateFirrtlAndFileCheck(new Foo)("""
+      // CHECK-LABEL: module Foo
+      // CHECK: input a : UInt<1>
+      // CHECK: contract :
+      // CHECK:   intrinsic(circt_verif_assert, a)
+      """)
+  }
```
seldridge, line 30: Please include a test that the explicit `()` version errors, assuming we _do_ want to error in that case.

src/test/scala-2/chiselTests/FormalContractSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+import chisel3.experimental.SourceLine
+import _root_.circt.stage.ChiselStage
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import Sequence._
+
+class FormalContractSpec extends AnyFlatSpec with Matchers with FileCheck {
+  it should "support contracts with no operands" in {
```
seldridge, line 17: Nit: Usually a "Flat Spec" will have some "behavior of ???" lines to help structure it. Or it will have a leading:

``` scala
"Formal Contracts" should "foo" in { ??? }

they should "bar" in { ??? }

"Formal Contracts error behavior" should "baz" in { ??? }
```

This is very nitty, but it helps with readability of what the test is doing both when reading the code and when reading what Scalatest shows as output.

seldridge (approved):
Superficial nits. Land whenever you're ready.

sequencer:
It would be super cool to add some mdoc-based documentation for the contract + hierarchical formal example!

src/test/scala-2/chiselTests/FormalContractSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+import chisel3.testers.BasicTester
+import chisel3.experimental.SourceLine
+import _root_.circt.stage.ChiselStage
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import Sequence._
+
+class FormalContractSpec extends AnyFlatSpec with Matchers with FileCheck {
```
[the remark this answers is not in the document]
jackkoenig: You should have to, I merged master so that CI will rerun to make sure we don't accidentally merge this using the older APIs, of course feel free to rebase the merge away.

core/src/main/scala-2/chisel3/FormalContract.scala
```diff
@@ -0,0 +1,35 @@
...
+  // and the corresponding `TupleN` for multiple arguments). Without this macro,
+  // the contracts would always operate on tuples (`Product`), forcing the user
+  // to use
+  //
+  //   val Tuple1(a) = FormalContract(Tuple1(b)) { case Tuple1(x) => ... }
+  //
+  // for contracts with only a single argument.
+  /** @group Macro
+    * @groupprio Macro 1001
+    */
```
mwachs5, line 26:
```suggestion
  /** @group Macro
    * @groupprio Macro 1001
    */
```

perhaps https://github.com/chipsalliance/chisel/blob/57ab35eb940224f1fd93651d87b83842c416db70/macros/src/main/scala/chisel3/SourceInfoDoc.scala#L22 ?

core/src/main/scala-3/chisel3/FormalContractMacro.scala
fabianschuiki: My bad, probably should be `{{{` and `}}}`. Used to the Markdown docs of Rust and Clang.

core/src/main/scala/chisel3/FormalContractImpl.scala
fabianschuiki: Good point, done!

core/src/main/scala/chisel3/FormalContractImpl.scala
fabianschuiki: My hope was that this would show up on Scala 2 and Scala 3 `object FormalContract` which extends `ObjectFormalContractImpl`. Let me check how doc inheritance works for private base classes/traits -- otherwise I can always factor this into a public dummy base as we do for some of the source location macro stuff.

firrtl/src/main/scala/firrtl/ir/Serializer.scala
fabianschuiki: Very nice, I wasn't aware `mkString` 👍 I don't think the same thing will work for expressions though, since they call `s` to implicitly populate `b`, so there's no string for the expression to intercept 😢

src/test/scala-2/chiselTests/FormalContractSpec.scala
fabianschuiki: Good point. I've added a `behavior of` line to this.

src/test/scala-2/chiselTests/FormalContractSpec.scala
fabianschuiki: Hmmm, that would be scala compiler error. Can we check for those?

core/src/main/scala/chisel3/FormalContractImpl.scala
jackkoenig: > I just realized that this should actually be public

This should not be public--we don't want the Scala 2 / Scala 3 split visible in the public API.

> The apply and mapped functions are supposed to be available to users of the API. 

Even wit the trait package private, `apply` and `mapped` are public on any classes/objects that mix the trait in so no problem there.

> That should also fix the doc issue.

That won't really fix the doc issue because users would still have to click through to the trait to see the doc which isn't what we want, we want it to show up on the `object FormalContract`.

I think you need to put `  /** @inheritdoc */` on each `object FormalContract` to make inheritance work. Now if that can't work from a package private trait, then you could move the Scala doc to a public trait that is nothing but the doc, but I think the implementations of `mapped` and `apply` should stick around on a private trait.

--------------------------------

**All of this being said**

I'd like to invert the public/private relationship such that the public classes and objects lives in `src/main/scala` and only package private traits that provide the public methods using macros live in `src/main/scala-2` and `src/main/scala-3`. So here's what I think you should really do.

Create a new package private trait (probably called `FormalContract$VirtualMethods`[^1]) in this file (`core/src/main/scala/chisel3/FormalContractImpl.scala`) that just defines the _virtual method_ `mapped`. Change the code in `src/main/scala-2` and `src/main/scala-3` to be package private traits (probably called `FormalContract$Intf`) that extend `FormalContract$VirtualMethods` and self-typed on `FormalContract.type`[^2] but still provide the public methods that use macros. Then back in this file, make `object FormalContract extends FormalContract$Intf` and you'll just need the ScalaDoc there. It's more clear than the old pattern IMO, I need to roll out this new approach to the rest of the code base.

[^1]: The `$` is because in Scala, an object's underlying type is named with a trailing `$` in Java bytecode. This is mostly hidden in the Scala API but leaks at times in errors. @seldridge and I discussed that the pattern `ObjectModuleImpl` is confusing that it's referring to `object module` and that `Module$Impl` would be more clear so that's the new style I'm proposing.

[^2]: `FormalContract.type` is technically the type for `object FormalContract`, but as stated above, it desugars/lowers to `FormalContract$` in Java bytecode.

src/test/scala-2/chiselTests/FormalContractSpec.scala
seldridge: You're right. Not sure how to do that. Forget it!

src/test/scala-2/chiselTests/FormalContractSpec.scala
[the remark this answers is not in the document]
fabianschuiki: Fixed 👍

src/test/scala-2/chiselTests/FormalContractSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+          AssertProperty(y)
+        }
+      }
+    }
+    generateFirrtlAndFileCheck(new Foo)("""
+      // CHECK-LABEL: module Foo
+      // CHECK: input a : UInt<1>
+      // CHECK: contract [[B:.+]] = a :
+      // CHECK:   contract [[C:.+]] = [[B]] :
+      // CHECK:     intrinsic(circt_verif_assert, [[C]])
```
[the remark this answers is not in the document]
fabianschuiki: That's a good idea. I've tweaked the tests.

core/src/main/scala/chisel3/FormalContractImpl.scala
jackkoenig: I've laid this out more explicitly and with an example in https://github.com/chipsalliance/chisel/pull/4768

fabianschuiki: Really nice. I've moved the contracts to this new pattern.

core/src/main/scala-2/chisel3/FormalContract.scala
fabianschuiki: I've managed to hide the macros properly thanks to @jackkoenig's new scala-2/3 pattern.

fabianschuiki:
Thanks for the reviews and great suggestions. Going to land this now.

PR #4683, opened 2025-02-12
Add ElaboratedCircuit and deprecate use of internal ir Circuit

jackkoenig:
When I tried out https://github.com/chipsalliance/chisel/pull/4643 internally at SiFive, I realized that despite `chisel3.internal.firrtl.ir` being a package private object, `ir.Circuit` still leaks out via `ChiselCircuitAnnotation` and it can be used freely. This leaks full access to the Chisel internal IR in a way that makes it really hard to change things.

This is my attempt to "spackle" that leak by providing the new `ElaboratedCircuit` API as the intended public interface to give to users to let them do what they want to do, while letting us muck with the internals. I think we should feel free to add and deprecate then remove APIs from `ElaboratedCircuit` as long as we take care to keep the actual internals private.

Before merging this, I intend to test it internally and redo https://github.com/chipsalliance/chisel/pull/4643 on top of it because I still want to backport the annotation deprecations as well. Technically https://github.com/chipsalliance/chisel/pull/4643 does not need this, but I think this makes the story of what changes I want to make in 7.0 a lot cleaner without having to worry about changes to `ir.Circuit` accidentally breaking user code.

### Contributor Checklist

#### Type of Improvement

- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

Change ChiselCircuitAnnotation and CircuitSerializationAnnotation to no longer be case classes to help with the transition to ElaboratedCircuit.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/ElaboratedCircuit.scala
+++ b/core/src/main/scala/chisel3/ElaboratedCircuit.scala
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+import firrtl.annotations.Annotation
+import firrtl.ir.{CircuitWithAnnos, Serializer}
+import firrtl.options.{CustomFileEmission, Unserializable}
+import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
+import chisel3.experimental.hierarchy.Definition
+import chisel3.internal.firrtl.ir.Circuit
+import chisel3.internal.firrtl.Converter
+
+/** The result of running Chisel elaboration
+  *
+  * Provides limited APIs for inspection of the resulting circuit.
+  */
+// This is an important interface so let's keep it separate from the implementation.
+sealed trait ElaboratedCircuit {
+
+  /** The name of the circuit, also the name of the top public module */
+  def name: String
+
+  /** The circuit and annotations as a string of FIRRTL IR
+    *
+    * This will include annotations passed to Chisel to build the circuit and those created during elaboration.
+    * For large circuits (> 2 GiB of text) use [[lazilySerialize]].
+    */
+  def serialize: String
+
+  /** The circuit and annotations as a string of FIRRTL IR
+    *
+    * For large circuits (> 2 GiB of text) use [[lazilySerialize]]
+    *
+    * @param annotations annotations to include in the FIRRTL IR. No other annotations will be included.
+    */
+  def serialize(annotations: Iterable[Annotation]): String
+
+  /** The circuit and annotations as a lazy buffer of strings of FIRRTL IR
+    *
+    * This will include annotations passed to Chisel to build the circuit and those created during elaboration.
+    * Serialized lazily to reduce peak memory use and support cicuits larger than 2 GiB.
+    */
+  def lazilySerialize: Iterable[String]
+
+  /** The circuit and annotations as a lazy buffer of strings of FIRRTL IR
+    *
+    * Serialized lazily to reduce peak memory use and support cicuits larger than 2 GiB.
+    *
+    * @param annotations annotations to include in the FIRRTL IR. No other annotations will be included.
+    */
+  def lazilySerialize(annotations: Iterable[Annotation]): Iterable[String]
+
+  /** The annotations created during elaboration of this circuit
+    *
+    * This does not include annotations passed to elaboration.
+    */
+  def annotations: Iterable[Annotation]
+
+  /** The Definition of the top module in the elaborated circuit */
+  def topDefinition: Definition[BaseModule]
+
+  /** The underlying circuit, for private use only */
+  private[chisel3] def _circuit: Circuit
+}
+
+private class ElaboratedCircuitImpl(circuit: Circuit, initialAnnotations: Seq[Annotation]) extends ElaboratedCircuit {
+
+  // Source locator needed for toDefinition
+  private implicit def sourceInfo: SourceInfo = UnlocatableSourceInfo
+
+  override def name: String = circuit.name
+
+  override def serialize: String = lazilySerialize.mkString
+
+  override def serialize(annotations: Iterable[Annotation]): String = lazilySerialize(annotations).mkString
+
+  override def lazilySerialize: Iterable[String] = {
+    val annotations = (initialAnnotations.view ++ circuit.firrtlAnnotations).flatMap {
+      case _: Unserializable     => None
+      case _: CustomFileEmission => None
+      case a => Some(a)
+    }.toVector
+    lazilySerialize(annotations)
+  }
+
+  override def lazilySerialize(annotations: Iterable[Annotation]): Iterable[String] = {
+    val prelude = {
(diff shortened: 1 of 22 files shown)


core/src/main/scala/chisel3/ElaboratedCircuit.scala
```diff
@@ -0,0 +1,111 @@
...
+import chisel3.experimental.{BaseModule, SourceInfo, UnlocatableSourceInfo}
+import chisel3.experimental.hierarchy.Definition
+import chisel3.internal.firrtl.ir.Circuit
+import chisel3.internal.firrtl.Converter
+
+/** The result of running Chisel elaboration
+  *
+  * Provides limited APIs for inspection of the resulting circuit.
+  */
+// This is an important interface so let's keep it separate from the implementation.
```
seldridge, line 17: Does this comment cause problems with Scaladoc?

core/src/main/scala/chisel3/ElaboratedCircuit.scala
```diff
@@ -0,0 +1,111 @@
...
+    */
+  def serialize: String
+
+  /** The circuit as a string of FIRRTL IR
+    *
+    * For large circuits (> 2 GiB of text) use [[lazilySerialize]]
+    *
+    * @param annotations alternative annotations to include in the FIRRTL IR.
+    */
+  def serialize(annotations: Iterable[Annotation]): String
```
seldridge, line 36: Is this the circuit, annotations, or both?

core/src/main/scala/chisel3/ElaboratedCircuit.scala
```diff
@@ -0,0 +1,111 @@
...
+    *
+    * This does not include annotations passed to elaboration.
+    */
+  def annotations: Iterable[Annotation]
+
+  /** Definitions of all modules in this circuit */
+  def definitions: Iterable[Definition[BaseModule]]
+
+  /** The underlying circuit, for private use only */
+  private[chisel3] def _circuit: Circuit
```
seldridge, line 63: Nit: `circuit` is probably fine given that this is private.

core/src/main/scala/chisel3/ElaboratedCircuit.scala
```diff
@@ -0,0 +1,111 @@
...
+}
+
+private class ElaboratedCircuitImpl(circuit: Circuit, initialAnnotations: Seq[Annotation]) extends ElaboratedCircuit {
+
+  // Source locator needed for toDefinition
+  private implicit def sourceInfo: SourceInfo = UnlocatableSourceInfo
+
+  def name: String = circuit.name
```
seldridge, line 71: `override def` for all these and throughout.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -567,7 +567,7 @@ private[chisel3] object ir {
 
   case class DefClass(id: Class, name: String, ports: Seq[Port], block: Block) extends Component
 
-  case class Circuit(
+  private[chisel3] case class Circuit(
```
seldridge, line 570: 🎊 🎊 🎊

docs/src/cookbooks/hierarchy.md
```diff
@@ -161,12 +161,8 @@ class Top extends Module {
...
 ```scala mdoc:passthrough
 println("```")
-val chiselCircuit = (new chisel3.stage.phases.Elaborate)
-  .transform(Seq(chisel3.stage.ChiselGeneratorAnnotation(() => new Top)))
-  .collectFirst { case chisel3.stage.ChiselCircuitAnnotation(a) =>
-    a
-  }.get
-  println(chiselCircuit)
+// Run elaboration so that the println above shows up
+circt.stage.ChiselStage.convert(new Top)
```
seldridge, line 165: Does this also need to be `println`'ed?

src/main/scala/chisel3/stage/ChiselAnnotations.scala
```diff
@@ -297,11 +297,56 @@ object ChiselGeneratorAnnotation extends HasShellOptions {
...
+    case that: ChiselCircuitAnnotation => this._circuit == that._circuit
+    case _ => false
+  }
+
+  def productArity: Int = 1
+
+  def productElement(n: Int): Any = n match {
+    case 0 => _circuit
+    case _ => throw new IndexOutOfBoundsException(s"Invalid index $n")
+  }
```
seldridge, line 328: Nice!

core/src/main/scala/chisel3/ElaboratedCircuit.scala
seldridge: After reading further, I see that this may be a useful convention throughout. Do whatever makes sense to you.

src/main/scala/chisel3/testers/TesterDriver.scala
seldridge, line 1: Ugh... this whole file just needs to go away. 🥲

seldridge (approved):
LGTM

A few nits.

core/src/main/scala/chisel3/ElaboratedCircuit.scala
jackkoenig: Both, will update all of these ScalaDocs. The pattern for "I want just the circuit" is to call this version but pass Nil

core/src/main/scala/chisel3/ElaboratedCircuit.scala
jackkoenig: I'm going to stick with this for now but since it's package private we can change it arbitrarily.

core/src/main/scala/chisel3/ElaboratedCircuit.scala
```diff
@@ -0,0 +1,111 @@
...
+    val modules = circuit.components.iterator.map(c => Converter.convert(c, typeAliases))
+    val moduleStrings = modules.flatMap { m =>
+      Serializer.lazily(m, 1) ++ Seq("\n\n")
+    }
+    prelude ++ moduleStrings
+  }
+
+  def annotations: Iterable[Annotation] = circuit.firrtlAnnotations
+
+  def definitions: Iterable[Definition[BaseModule]] = circuit.components.map(_.id.toDefinition)
```
jackkoenig, line 103: I'm changing this to just be `topDefinition` since that is sufficient for most purposes when combined with `Select`. 

It's likely we'll add a similar thing like `publicDefinitions` as well but I'll save that till we need it.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: I actually don't think this makes any difference lol. `object ir` is already package private so people already can't construct these--but the leakage of `Circuit` out via the annotations remains.

core/src/main/scala/chisel3/ElaboratedCircuit.scala
jackkoenig: It does not!

PR #4684, opened 2025-02-12
[firrtl] Add, use dramaticMessage

seldridge:
Factor out a utility for constructing distinctive messages from the CIRCT phase and into the firrtl package.  This was always the plan, however, I forgot about it for years.

This is being moved and made public so that I can use it when constructing Chiselsim error messages.

#### Release Notes

- Add `firrtl.options.StageUtils.dramaticMessage` utility for constructing distinctive error messages that do not print to stdout.

### DIFF
--- a/firrtl/src/main/scala/firrtl/options/StageUtils.scala
+++ b/firrtl/src/main/scala/firrtl/options/StageUtils.scala
@@ -22,7 +22,7 @@ object StageUtils {
     * @param message error message
     */
   def dramaticWarning(message: String): Unit = {
-    println(Console.YELLOW + dramaticMessage(header=None, body=s"Warning: $message") + Console.RESET)
+    println(Console.YELLOW + dramaticMessage(header = None, body = s"Warning: $message") + Console.RESET)
   }
 
   /** Print an error message (in red)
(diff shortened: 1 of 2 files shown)


jackkoenig (commented):
Internally, Chisel has logic to disable color use: https://github.com/chipsalliance/chisel/blob/ba49fcc524db56b00ab880ba043c8b6815d2e53d/core/src/main/scala/chisel3/internal/Error.scala#L242

Is it possible to hook into that somehow here?

seldridge:
> Is it possible to hook into that somehow here?

Probably in a follow-on. I'm only delaying it as this patch only adds the non-color version.

Did we ever try to do this in the normal way? I.e., most tools have a `--color=[when]` option where `when` can be one of `never|always|auto` where `auto` is the default and it means to "use color if stdout reports that it can handle it". (This then does what you usually want and will strip color if piped, but show it on the terminal.) It's kind of hard to get that information _into_ Chisel, but that can certainly happen with that environment variable with those three settings.

jackkoenig:
> Probably in a follow-on. I'm only delaying it as this patch only adds the non-color version.

That's reasonable.

> Did we ever try to do this in the normal way?

Normalish. Chisel detect if it should use color [[1]](https://github.com/chipsalliance/chisel/blob/14c148a38c9b8005f011fddcc3a828cb1f7d2c49/core/src/main/scala/chisel3/internal/Error.scala#L251), unfortunately the JVM doesn't let you check just stdout so it will only be color if all of stdin/out/err are "interactive" but it's something. Chisel also uses environment variable `CHISEL_USE_COLOR` to let the user specify, but yeah it doesn't take it as a CLI argument.

PR #4694, opened 2025-02-14
[chiselsim] Improve error behavior

seldridge:
Fix longstanding problems in Chiselsim where the result of a failing simulation (e.g., an assertion firing) would result in an obtuse `UnexpectedEndOfMessages` assertion.  This then forces any Chiselsim user to catch this and then figure out what went wrong.  See the unpublished ChiselSpec for an example.

Fix this with an abstract method for the Backend that forces it to describe (via a RegEx) how to find interesting parts of the log file which indicate certain failures (the aforementioned assertion failures).  This then allows for the default simulate method to substitute a _better_ assertion where possible.

This is not as good as a more advanced alternative which actually hooks into the simulation to do something when an assertion fires.  However, this is a larger effort that requires changing the Verilog that is emitted for CIRCT-compiled FIRRTL (or defining an ABI for assertions).  We definitely want to move in this direction, though.

#### Release Notes

- Improve the behavior of Chiselsim on an expectation or assertion failure.  This will no longer require the user to catch an `UnexpectedEndOfMessage` assertion and reconstruct what happened.

### DIFF
--- a/src/main/scala/chisel3/simulator/PeekPokeAPI.scala
+++ b/src/main/scala/chisel3/simulator/PeekPokeAPI.scala
@@ -5,12 +5,22 @@ import chisel3._
 
 import chisel3.experimental.{SourceInfo, SourceLine}
 import chisel3.internal.ExceptionHelpers
+import firrtl.options.StageUtils.dramaticMessage
+import scala.util.control.NoStackTrace
 
 object PeekPokeAPI extends PeekPokeAPI
 
 trait PeekPokeAPI {
   case class FailedExpectationException[T](observed: T, expected: T, message: String)
-      extends Exception(s"Failed Expectation: Observed value '$observed' != $expected. $message")
+      extends RuntimeException(
+        dramaticMessage(
+          header = Some("Failed Expectation"),
+          body = s"""|Observed value: '$observed'
+                     |Expected value: '$expected'
+                     |$message""".stripMargin
+        )
+      )
+      with NoStackTrace
   object FailedExpectationException {
     def apply[T](
       observed:     T,
--- a/src/main/scala/chisel3/simulator/Simulator.scala
+++ b/src/main/scala/chisel3/simulator/Simulator.scala
@@ -54,6 +54,31 @@ trait Simulator[T <: Backend] {
   def commonCompilationSettings:          CommonCompilationSettings
   def backendSpecificCompilationSettings: backend.CompilationSettings
 
+  /** Post process a simulation log to see if there are any backend-specific failures.
+    *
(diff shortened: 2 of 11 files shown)


src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -54,6 +54,31 @@ trait Simulator[T <: Backend] {
...
+    *
+    * @return None if no failures found or Some if they are
+    */
+  private def postProcessLog: Option[Throwable] = {
+    val log = Paths.get(workspacePath, s"workdir-${tag}", "simulation-log.txt").toFile
+    val lines = scala.io.Source.fromFile(log).getLines().zipWithIndex.filter { case (line, lineNo) => backend.assertionFailed.matches(line) }.toSeq
+    if (lines.isEmpty)
+      return None
+
+    Some(
```
jackkoenig, line 67:
```suggestion
    Option.when(lines.nonEmpty)(
```
I'm not against return but it feels egregious here 😉

src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -54,6 +54,31 @@ trait Simulator[T <: Backend] {
   def commonCompilationSettings:          CommonCompilationSettings
   def backendSpecificCompilationSettings: backend.CompilationSettings
 
+  /** Post process a simulation log to see if there are any backend-specific failures.
+    *
+    * @return None if no failures found or Some if they are
+    */
+  private def postProcessLog: Option[Throwable] = {
+    val log = Paths.get(workspacePath, s"workdir-${tag}", "simulation-log.txt").toFile
+    val lines = scala.io.Source.fromFile(log).getLines().zipWithIndex.filter { case (line, lineNo) => backend.assertionFailed.matches(line) }.toSeq
```
jackkoenig, line 63: lineNo is unused

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
```diff
@@ -5,12 +5,22 @@ import chisel3._
...
-      extends Exception(s"Failed Expectation: Observed value '$observed' != $expected. $message")
+      extends RuntimeException(
+        dramaticMessage(
+          header = Some("Failed Expectation"),
+          body = s"""|Observed value: '$observed'
+                     |Expected value: '$expected'
+                     |$message""".stripMargin
+        )
+      )
+      with NoStackTrace
```
jackkoenig, line 23: 🔥 eliminating useless stack traces is really nice.

jackkoenig (approved):
Looks great!

PR #4699, opened 2025-02-15
[chiselsim] Add FIRRTL macro control

seldridge:
Change SimulatorAPI methods to use the ChiselSettings class instead of
directly communicating layerControl.  This will enable us to lump together
more options that all are associated with Chisel/FIRRTL/the Verilog ABI.

Extend ChiselSettings with three macros which were currently being set by
ChiselRunners:

- ASSERT_VERBOSE_COND
- PRINTF_COND
- STOP_COND

Make this work in such a way that the user can specify a function that
will return the _signal_ that they want to use to set this macro in one of
several predefined ways.

This is working towards replacing this part of ChiselSpec: https://github.com/chipsalliance/chisel/blob/main/src/test/scala-2/chiselTests/ChiselSpec.scala#L60

#### Release Notes

- Add FIRRTL/Verilog ABI macro control to Chiselsim.

### DIFF
--- a/src/main/scala/chisel3/simulator/ChiselSettings.scala
+++ b/src/main/scala/chisel3/simulator/ChiselSettings.scala
@@ -149,13 +149,13 @@ object ChiselSettings {
     * must invoke this method like:
     *
     * {{{
-    * ChiselSettings.default[Foo]
+    * ChiselSettings.defaultRaw[Foo]
     * }}}
     *
     * If you invoke this method like the following, you will get an error:
     *
     * {{{
-    * ChiselSettings.default
+    * ChiselSettings.defaultRaw
     * }}}
     */
   final def defaultRaw[A <: RawModule]: ChiselSettings[A] = new ChiselSettings[A](
--- a/src/main/scala/chisel3/simulator/ChiselSettings.scala
+++ b/src/main/scala/chisel3/simulator/ChiselSettings.scala
@@ -70,9 +70,14 @@ object MacroText {
   * compilation itself or the Verilog compilation and simulation.
   *
   * @param layerControl determines which [[chisel3.layer.Layer]]s should be
+  * @param assertVerboseCond a condition that guards the printing of assert
+  * messages created from `circt_chisel_ifelsefatal` intrinsics
+  * @param printfCond a condition that guards printing of [[chisel3.printf]]s
+  * @param stopCond a condition that guards terminating the simulation (via
+  * `$fatal`) for asserts created from `circt_chisel_ifelsefatal` intrinsics
   * enabled _during Verilog elaboration_.
   */
-final class ChiselSettings[A <: RawModule](
+final class ChiselSettings[A <: RawModule] private[simulator] (
   /** Layers to turn on/off during Verilog elaboration */
   val verilogLayers:     LayerControl.Type,
   val assertVerboseCond: Option[MacroText.Type[A]],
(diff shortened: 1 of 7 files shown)


src/main/scala/chisel3/simulator/SimulatorAPI.scala
```diff
@@ -74,12 +74,12 @@ trait SimulatorAPI {
     */
   def simulate[T <: Module](
     module:                => T,
-    layerControl:          LayerControl.Type = LayerControl.EnableAll,
+    chiselSettings:        ChiselSettings[T] = ChiselSettings.default[T],
```
seldridge, line 77: Note: because the `default` factory is necessarily given a type parameter with a lower bound (`A <: RawModule]`), this necessitates giving it a type parameter on line 77 here. If I don't do this, this will infer to `Nothing` and everything blows up.

There may be a way to untangle this.

src/main/scala/chisel3/simulator/ChiselSettings.scala
```diff
@@ -11,15 +72,92 @@ package chisel3.simulator
   * @param layerControl determines which [[chisel3.layer.Layer]]s should be
   * enabled _during Verilog elaboration_.
   */
-final class ChiselSettings(
+final class ChiselSettings[A <: RawModule](
```
jackkoenig, line 75: Consider privatizing the constructor and instead using a public factory method. Not a huge deal but IMO better style and makes some possible future refactors easier.

src/main/scala/chisel3/simulator/ChiselSettings.scala
```diff
@@ -11,15 +72,92 @@ package chisel3.simulator
...
+    * }}}
+    */
+  final def default[A <: Module]: ChiselSettings[A] = new ChiselSettings[A](
+    verilogLayers = LayerControl.EnableAll,
+    assertVerboseCond = Some(MacroText.NotSignal(get = _.reset)),
+    printfCond = Some(MacroText.NotSignal(get = _.reset)),
+    stopCond = Some(MacroText.NotSignal(get = _.reset))
+  )
+
+  /** Retun a default [[ChiselSettings]] for a [[Module]].
```
jackkoenig, line 137:
```suggestion
  /** Retun a default [[ChiselSettings]] for a [[RawModule]].
```

src/main/scala/chisel3/simulator/ChiselSettings.scala
```diff
@@ -11,15 +72,92 @@ package chisel3.simulator
...
+    *
+    *  This differs from [[default]] in that it cannot set default values for
+    *  macros because a [[RawModule]] has no defined reset port.  You will
+    *  likely want to override the macros after using this factory.
+    *
+    * Note: this _requires_ that an explicit type parameter is provided.  You
+    * must invoke this method like:
+    *
+    * {{{
+    * ChiselSettings.default[Foo]
```
jackkoenig, line 147:
```suggestion
    * ChiselSettings.defaultRaw[Foo]
```

src/main/scala/chisel3/simulator/ChiselSettings.scala
```diff
@@ -11,15 +72,92 @@ package chisel3.simulator
...
+    * }}}
+    *
+    * If you invoke this method like the following, you will get an error:
+    *
+    * {{{
+    * ChiselSettings.default
```
jackkoenig, line 153:
```suggestion
    * ChiselSettings.defaultRaw
```

jackkoenig (approved):
LGTM! A few nits but I like it

src/main/scala/chisel3/simulator/ChiselSettings.scala
seldridge: Good idea. Changed to this in: 366de579b

src/main/scala/chisel3/simulator/ChiselSettings.scala
seldridge: Fixed in: dccd150f5

src/main/scala/chisel3/simulator/ChiselSettings.scala
seldridge: Fixed in: 2f6478e21

src/main/scala/chisel3/simulator/ChiselSettings.scala
seldridge: Fixed in: 2f6478e21

PR #4702, opened 2025-02-18
[svsim] End Verilator simulation on $finish

### LINKED ISSUE #4700: [Chiselsim] Exit simulation on $finish
Currently, Chiselsim will only terminate if the simulator exits or the commands that it is trying to run are done. However, different simulators handle things like `$finish` (which the Chisel `stop` API lowers to) differently. E.g., consider the following example code:

``` verilog
module Foo();

initial begin
  $display("hello");
  $finish;
  $display("world");
  $finish;
end

endmodule
```

Compiling this with Verilator (`verilator --binary Foo.sv && ./obj_dir/VFoo`) will produce:

```
hello
- Foo.sv:5: Verilog $finish
world
- Foo.sv:7: Verilog $finish
```

If you run this with VCS (`vcs Foo.sv && ./simv`) will produce:

```
hello
$finish called from file "Foo.sv", line 5.
$finish at simulation time                    0
```

The problem here is that Verilator doesn't terminate. This seems to violate the Verilog spec, based on the snippet below:

Image

That said, we can work around this by querying the `VerilatedContext.getFinish()` member function when in the `run_simulation` function (or elsewhere) to see if Verilator is reporting that the simulation is done.

This has two major benefits:

1. Simulations will actually exit instead of running until they hit a timeout. For `ChiselSpec`, this is currently 10k half-periods. We can exit simulations way earlier without wasting cycles.
2. [We are hacking around this by post-processing the log looking for finish _before_ an assert.](https://github.com/chipsalliance/chisel/blob/main/src/test/scala-2/chiselTests/ChiselSpec.scala#L104) All that checking logic can just go away if simulations actually terminate.

seldridge:
Fix a bug in how Verilator simulations work in svsim where the simulation would _not_ terminate when a `$finish` was called.  This seems to be a bug in Verilator where it will not terminate the simulation (as mandated by the Verilog spec) unless you are running with time delays.

Fixes #4700.

#### Release Notes

Fix an svsim bug, which could manifest in Chiselsim, where when a simulation calls `$finish` that it does not immediately exit for Verilator backends.

### DIFF
--- a/svsim/src/test/scala/BackendSpec.scala
+++ b/svsim/src/test/scala/BackendSpec.scala
@@ -298,6 +298,39 @@ trait BackendSpec extends AnyFunSpec with Matchers {
           }
         }
       }
+
+      it("ends the simulation on '$finish' (#4700)") {
+        workspace.reset()
+        workspace.elaborateFinishTest()
+        workspace.generateAdditionalSources()
+        simulation = workspace.compile(
+          backend
+        )(
+          workingDirectoryTag = name,
+          commonSettings = CommonCompilationSettings(),
+          backendSpecificSettings = compilationSettings,
+          customSimulationWorkingDirectory = None,
+          verbose = false
+        )
+        simulation.run(
+          verbose = false,
+          executionScriptLimit = None
+        ) { controller =>
+          val clock = controller.port("clock")
+          clock.tick(
+            inPhaseValue = 0,
+            outOfPhaseValue = 1,
+            timestepsPerPhase = 1,
+            maxCycles = 8,
+            sentinel = None
+          )
+        }
+        val re = ".*Verilog \\$finish.*".r
+        new BufferedReader(new FileReader(s"${simulation.workingDirectoryPath}/simulation-log.txt")).lines
+          .filter(re.matches(_))
+          .toArray
+          .size must be(1)
+      }
     }
   }
 }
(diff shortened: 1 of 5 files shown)


svsim/src/test/scala/BackendSpec.scala
```diff
@@ -298,6 +298,30 @@ trait BackendSpec extends AnyFunSpec with Matchers {
...
+          customSimulationWorkingDirectory = None,
+          verbose = false
+        )
+        intercept[svsim.Simulation.UnexpectedEndOfMessages.type] {
+          simulation.run(
+            verbose = false,
+            executionScriptLimit = None
+          ) { controller =>
+            controller.run(1)
+            println(controller.readLog())
```
jackkoenig, line 321: Should this be checking something instead of just printing?

jackkoenig (approved):
LGTM although it looks like the test isn't actually checking the log, it's just printing it. Or is the check the intercepted exception? A comment clarifying would be nice.

svsim/src/test/scala/BackendSpec.scala
seldridge: I thought that this was testing it properly, but it was not. The exception was coming from the simulation shutting down and then the read would fail. This isn't the best test.

I changed that in the force push to count the number of finishes in the simulation log.

PR #4706, opened 2025-02-19
[Chiselsim] Replace Chiselspec w/ ChiselSim

seldridge:
Add timeout exception and `runUntilFinished`. Use this to replace some usages of `ChiselSpec`.

This is stacked on #4699. This PR is intended to show where I am going with this and to get buy-in on how tests will look before rolling this out everywhere (to replace more widely used `ChiselSpec` methods like `assertTesterPasses` and `assertTesterFails`.

#### Release Notes

Add `Stimulus` object to `SimulatorAPI` which is intended to contain basic, common stimulus that should be applied to a circuit.  These can be chained together in series if desired. For now, only add `runUntilFinished` which is stimulus that runs for a number of cycles and then throws an exception if the simulation doesn't finish before the timeout.

### DIFF
--- a/src/main/scala/chisel3/simulator/stimulus/RunUntilFinished.scala
+++ b/src/main/scala/chisel3/simulator/stimulus/RunUntilFinished.scala
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator.stimulus
+
+import chisel3.{Clock, Module}
+import chisel3.simulator.AnySimulatedModule
+import chisel3.simulator.Exceptions
+
+/** Stimulus that will run a simulation, expecting a [[chisel3.stop]] (a Verilog
+  * `$finish`) to occur before a maximum number of cycles has elapsed.
+  */
+trait RunUntilFinished[A] extends Stimulus.Type[A] {
+
+  /** The maximum number of cycles. */
+  protected def _maxCycles: Int
+
+  /** A function that returns the clock to tick. */
+  protected def _getClock: (A) => Clock
+
+  /** Apply stimulus to the unit
+    *
+    * @param the unit to apply stimulus to
+    */
+  override final def apply(dut: A): Unit = {
+    AnySimulatedModule.current
+      .port(_getClock(dut))
+      .tick(
+        timestepsPerPhase = 1,
+        maxCycles = _maxCycles,
+        inPhaseValue = 1,
+        outOfPhaseValue = 0,
+        sentinel = None,
+        checkElapsedCycleCount = (count: BigInt) => {
+          if (count == _maxCycles)
+            throw new Exceptions.Timeout(_maxCycles, "Expected a $finish, but none received")
+        }
+      )
+  }
+
+}
+
+/** Factory of [[RunUntilFinished]] stimulus for different kinds of modules. */
+object RunUntilFinished {
+
+  /** Return stimulus for a [[Module]].  This uses the default clock of the module to apply stimulus.
+    *
+    * @param maxCycles the maximum number of cycles to run the unit for before a timeout
+    */
+  def module(maxCycles: Int): RunUntilFinished[Module] = new RunUntilFinished[Module] {
+
+    override protected final val _maxCycles = maxCycles
+
+    override protected final val _getClock = _.clock
+
+  }
+
+  /** Return stimulus for any type.  This requires the user to specify how to extract the clock from the type.
+    * @param maxCycles the maximum number of cycles to run the unit for before a timeout
+    * @param getClock a function to return a clock from the unit
+    */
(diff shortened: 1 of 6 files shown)


src/test/scala-2/chiselTests/BoringUtilsSpec.scala
```diff
@@ -34,12 +35,10 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
     BoringUtils.addSink(b, "x")
   }
 
-  behavior.of("BoringUtils.{addSink, addSource}")
+  behavior.of("BoringUtils.addSink and BoringUtils.addSource")
```
seldridge, line 38: Verilator or the Makefiles we generate have problems with some more exotic symbols. I need to circle back and replace these with simpler ones, like we do for whitespace.

src/test/scala-2/chiselTests/BoringUtilsSpec.scala
```diff
@@ -34,12 +35,10 @@ class BoringUtilsSpec extends ChiselFlatSpec with ChiselRunners with Utils with
...
 
   it should "connect two wires within a module" in {
-    runTester(
-      new ShouldntAssertTester { val dut = Module(new BoringInverter) }
-    ) should be(true)
+    simulate(new ShouldntAssertTester { val dut = Module(new BoringInverter) })(Stimulus.runUntilFinished(3))
```
jackkoenig, line 41: This is neat.

src/main/scala/chisel3/simulator/package.scala
```diff
@@ -32,7 +32,11 @@ package object simulator {
     private[simulator] val wrapped: T,
     private[simulator] val ports:   Seq[(Data, ModuleInfo.Port)],
     private[simulator] val layers:  Seq[chisel3.layer.Layer]
-  )
+  ) {
+
+    private[chisel3] lazy val portMap = ports.toMap
```
jackkoenig, line 37: Why lazy? do we ever not want it?

src/main/scala/chisel3/simulator/SimulatorAPI.scala
```diff
@@ -8,6 +8,34 @@ import java.nio.file.Files
...
+          maxCycles = maxCycles,
+          inPhaseValue = 1,
+          outOfPhaseValue = 0,
+          sentinel = None,
+          checkElapsedCycleCount = (count: BigInt) => {
+            if (count == maxCycles)
+              throw new Exceptions.Timeout(maxCycles, "Expected a $finish, but none received")
+          }
+        )
+    }
```
jackkoenig, line 35: I think this is good--we _could_ consider doing a typeclass instead of `T <: Module`. I say that because this won't be overloadable with other types so if someone wanted to use `runUnitFinished` for a type like `BasicTester` (or some future new version of that that has a `done` output like in the new inline unit testing stuff), then they'd have to give the new function a different name. Not sure if it's worthwhile and we can also add it later in a source-compatible way so not sure it matters now. Just pointing it out.

src/main/scala/chisel3/simulator/package.scala
seldridge: I had originally thought no. However, given that this will almost always execute via the macro control paths, we should just change this to be non-lazy. I noticed I can also narrow it be private to the `simulator` package, too. Both changes are applied in the rebase.

src/test/scala-2/chiselTests/BoringUtilsSpec.scala
seldridge: This is what I was primarily concerned about. Two alternatives to this API are a dedicated `simulateWithTimeout` or a default argument `timeout: BigInt` to this method. I liked the factoring in this patch better as it seemed like keeping all stimulus separate was cleaner.

src/main/scala/chisel3/simulator/SimulatorAPI.scala
seldridge: I switched to a type class style approach in https://github.com/chipsalliance/chisel/pull/4706/commits/d33e4614c70d4dda160797a8b39ce82bc30131e5. Let me know what you think.

jackkoenig (approved):
Looks great to me!

src/main/scala/chisel3/simulator/stimulus/RunUntilFinished.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    override protected final val _getClock = getClock
+
+  }
+
+  /** Return default stimulus.  This is the same as [[module]].
+    *
+    * @param maxCycles the maximum number of cycles to run the unit for before a timeout
+    */
+  def apply(maxCycles: Int) = module(maxCycles)
```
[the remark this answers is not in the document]
seldridge: Fixed in: 2ed08658b

PR #4707, opened 2025-02-19
[ci] Add -oF to tests

seldridge:
Add `-oF` to CI runs. I think I broke something with https://github.com/chipsalliance/chisel/commit/36e49577cfd46041cf70a82dbddfe7ccd10da1ae and am having difficulty debugging it.

I was originally only going to use this for testing. However, this would be useful to have generally on all CI runs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,7 @@ jobs:
           $CHISEL_FIRTOOL_PATH/firtool -version >> $GITHUB_STEP_SUMMARY
           echo \`\`\` >> $GITHUB_STEP_SUMMARY
       - name: Test
-        run: ./mill -j0 firrtl.cross[].test + svsim.cross[].test + chisel[].test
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           $CHISEL_FIRTOOL_PATH/firtool -version >> $GITHUB_STEP_SUMMARY
           echo \`\`\` >> $GITHUB_STEP_SUMMARY
       - name: Test
-        run: ./mill -j0 firrtl.cross[].test + svsim.cross[].test + chisel[].test
+        run: ./mill -j0 firrtl.cross[].test + svsim.cross[].test + chisel[].test -oF
```
jackkoenig, line 93: This is probably only gonna affect the chisel tests and not firrtl nor svsim. Probably fine but just FYI

seldridge: I didn't understand this syntax. Thanks! Changed in https://github.com/chipsalliance/chisel/pull/4707/commits/9cfb32b082f65c75ceb7538401a447f18e2d4354.

Issue #4676, opened 2025-02-11
[chiselsim] Exclude Extract Layers from Compilation if not Enabled


seldridge:
Currently, the way that extract layers are included or excluded from a Chiselsim simulation is through the `LayerControl` API which will cause the bind file (the part of the FIRRTL ABI which says that a file will be provided which, when included in the build, will enable the layer) to be included or excluded from the compilation.

However, all the files associated with the extract layer will still be included in the compilation and not used. This is problematic for two reasons:

1. It is inefficient to compile things that are known to not be used.
2. If the extract layer is being used to guard constructs which the simulator cannot support, then the compilation will fail. E.g., Verilator does not support the `s_eventually` SystemVerilog construct which can be emitted by the `ltl.eventually` API.

Change this so that the way `LayerControl` is handled by ChiselSim will instead only compile what is needed.

As a workaround, this type of feature can be put in an inline layer which will result in an `` `ifdef `` macro guarded region which will be excluded from the build.

CC: @SandeepRajendran

PR #4708, opened 2025-02-19
Dev/seldridge/switch to chisel settings

seldridge:
Factor out prerequisite pieces from: https://github.com/chipsalliance/chisel/pull/4699

I need to go muck with Layer Control to fix https://github.com/chipsalliance/chisel/issues/4676 and want to work on the latest version of this.

#### Release Notes

Add ChiselSettings class for controlling Chiselsim simulations.


jackkoenig (approved):
LGTM.

Layer control as an argument to the simulate functions has never been released right? So there's no reason to provide the old methods as deprecated?

PR #4709, opened 2025-02-19
[svsim][chiselsim] Do not compile extract layery directories

### LINKED ISSUE #4676: [chiselsim] Exclude Extract Layers from Compilation if not Enabled

seldridge:
Only compile what is needed for extract layers. Previously, all files would be compiled and the layer would be enabled using the ABI bind file. Change this to, for extract layers, only visit directories which the layer says it is writing to.

Note: this can cause problems if the user is using weird combinations of layer directory control and module-level control. However, if a user is doing this, they should know what they are doing and the limitations of the tooling.

Fixes #4676. Individual commits should be reviewed separately.

#### Release Notes

Stop compiling all extract layer collateral. Instead, only compile what is specified from those directories. This avoids a problem where LTL constructs (unsupported by Verilator) would cause compilation failure even though they would never be bound in.

### DIFF
--- a/svsim/src/main/scala/Workspace.scala
+++ b/svsim/src/main/scala/Workspace.scala
@@ -1,7 +1,9 @@
 package svsim
 
 import java.io.{BufferedReader, BufferedWriter, File, FileWriter, InputStreamReader, PrintWriter}
-import java.nio.file.{Files, Paths}
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.{FileVisitResult, FileVisitor}
+import java.nio.file.{FileVisitor, Files, Path, Paths}
 import java.lang.ProcessBuilder.Redirect
 import scala.annotation.meta.param
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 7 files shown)


svsim/src/main/scala/Workspace.scala
```diff
@@ -317,12 +319,43 @@ final class Workspace(
...
+
+        if (filter(file.toFile))
+          sourceFiles += workingDirectory.toPath().relativize(file).toString()
+        FileVisitResult.CONTINUE
+      }
+
+      override def preVisitDirectory(dir: Path, attrs: BasicFileAttributes): FileVisitResult = {
+        val filter: PartialFunction[File, Boolean] = commonSettings.directoryFilter.orElse { case _: File => true }
+
+        if (filter(dir.toFile))
```
jackkoenig, line 340:
```suggestion
    val sourceFiles = scala.collection.mutable.ArrayBuffer.empty[String]
    val fileFilter: PartialFunction[File, Boolean] = commonSettings.fileFilter.orElse { case _ => true }
    val directoryFilter: PartialFunction[File, Boolean] = commonSettings.directoryFilter.orElse { case _: File => true }
    class DirectoryVisitor extends FileVisitor[Path] {

      override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
        if (fileFilter(file.toFile))
          sourceFiles += workingDirectory.toPath().relativize(file).toString()
        FileVisitResult.CONTINUE
      }

      override def preVisitDirectory(dir: Path, attrs: BasicFileAttributes): FileVisitResult = {
        if (directoryFilter(dir.toFile))
```
This a nit/micro-optimization, but creating the filters is at least 2 object allocations each that could be memoized across all files and directories.

jackkoenig (approved):
This looks great, I especially appreciate the comments in shouldIncludeDirectory.

1 suggestion.

svsim/src/main/scala/Workspace.scala
seldridge: Good suggestion. Applied in https://github.com/chipsalliance/chisel/pull/4709/commits/00899a2e647ce7e2af786c4bcf6d2f69a426c568.

PR #4712, opened 2025-02-20
Re-land #4702

seldridge:
Re-land #4702.

The commit to fix this looks like a brittle fix as it narrowly handles the observed exception. (See: https://github.com/chipsalliance/chisel/actions/runs/13418015449/job/37483578519#step:11:1377) However, I think this is actually correct. The details are in the commit message and copied below:

Fix an issue that can occur now that the simulation will aggressively try
to shut themselves down when they see a `$finish`.  The problem occurs
because svsim is assuming that it can check if the simulation process is
alive and then shut it down.  However, this occurs in a critical region.
Just because the process is alive when checked, doesn't mean it _will_ be
alive when you try to shut it down.

I considered an alternative solutoin here which would put the simulation
into a "finish" state and then respond with a new "finish" message which
the controller would then need to handle. (It would just throw exceptions
if you tried to interact with the simulation after it ended.)  This
approach can be revived.  However, it needs to _also_ figure out what to
do with VCS.

The Verilog spec (1800-2023 Section 20.2) states:

> The `$finish` system task causes the simulator to exit and pass control
> back to the host operating system.

My read of this is that the simulation, if it is allowed to call
`$finish`, can terminate at essentially any time.  Verilator diverges from
this and will allow multiple `$finish` system tasks in the same block,
without time delays, to execute.  Based on how we construct the testbench
for Verilator, we run into this bug.  (If we use time delays, I think
Verilator will behave the same as VCS.)  Adapting the simulation to use a
"finish" message is therefore aligning the simulation with the expectation
of how Verilator behaves and is likely _not_ going to work for VCS.
Therefore, I rejected this approach.

If we want to move in this direction in the future, a change to the
testbench to use a `final` block, or some way of keeping the simulation
alive after a `$finish` to still respond to the controller, may work.

#### Release Notes

Fix an svsim bug, which could manifest in Chiselsim, where when a simulation calls `$finish` that it does not immediately exit for Verilator backends.

### DIFF
--- a/svsim/src/main/resources/simulation-driver.cpp
+++ b/svsim/src/main/resources/simulation-driver.cpp
@@ -71,7 +71,7 @@ void initTestBenchScope() { testbenchScope = svGetScope(); }
 #ifdef __cplusplus
 extern "C" {
 #endif
-extern void run_simulation(int timesteps);
+extern int run_simulation(int timesteps);
 extern void simulation_main(int argc, const char **argv);
 #ifdef __cplusplus
 }
@@ -672,7 +672,8 @@ static void processCommand() {
     if (*lineCursor != '\n') {
       failWithError("Unexpected data at end of RUN command.");
     }
-    run_simulation(time);
+    if (run_simulation(time))
+      receivedDone = true;
 
     sendAck();
     break;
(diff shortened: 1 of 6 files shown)


svsim/src/main/resources/simulation-driver.cpp
```diff
@@ -926,16 +933,18 @@ void simulation_main(int argc, char const **argv) {
...
   if(!delay) {
     testbench->eval_step();
-    return;
+    return context->gotFinish();
   }
   testbench->eval();
+  if (context->gotFinish())
+    return context->gotFinish();
   context->timeInc(delay);
+  return 0;
```
jackkoenig, line 945: I'm wondering if it's not worth also implementing vl_fatal, but perhaps checking `gotFinish` is equivalent for our purposes?

jackkoenig (approved):
LGTM, one question but feel free to land whenever

PR #4715, opened 2025-02-20
[website] Host latest and snapshot Scala CLI examples

jackkoenig:
I did not update the website with these new commands yet, want to make sure it works first and do that in follow on. It's also not clear where the website should say this, quickstart can point to latest but where to mention that you can get the snapshot? Unclear.

I did make one minor change, the Scala CLI example's first line is `//> using repository "sonatype-s01:snapshots"` which is only needed if you're building a snapshot. The current (i.e. before this PR) infrastructure strips that when publishing the example file for releases. With this PR, it will no longer strip it. It's technically not needed for published releases, but I think we should just leave it so that it's easier to switch to building from a snapshot (rather than having to figure out what the right directive is to fetch from sonatype-s01 snapshots).

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

You can now download the latest release Scala CLI example with:

```
curl -O -L https://chisel-lang.org/releases/latest/download/chisel-example.scala
```
And download for the latest snapshot with:
```
curl -O -L https://chisel-lang.org/releases/snapshot/download/chisel-example.scala
```

- [ ] Did you review?


.github/scripts/generate-scala-cli-example.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ "$#" -ne 1 ]; then
+  echo "Error: This script requires exactly 1 argument."
+  echo "Usage: $0 <chisel version>"
+  exit 1
+fi
+
+THISDIR=$(dirname "$0")
+VERSION=$1
+
+sed "s/@VERSION@/$VERSION/g" $THISDIR/chisel-example.scala
```
seldridge, line 12: Escaping would make this shellcheck-clean:

``` suggestion
sed "s/@VERSION@/$VERSION/g" "$THISDIR/chisel-example.scala"
```

website/Makefile
```diff
@@ -62,6 +65,20 @@ $(latest_scaladoc_dir)/index.html: $(latest_version) | $(latest_scaladoc_dir)
 	  rm -rf temp.jar
 	touch $@
 
+$(latest_chisel_download):
+	mkdir -p $@
+
+$(latest_chisel_download)/chisel-example.scala: $(latest_chisel_download) $(latest_version)
```
seldridge, line 71: `latest_chisel_download` is an [order-only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) (only existence matters, not timestamp):

``` suggestion
$(latest_chisel_download)/chisel-example.scala: $(latest_version) | $(latest_chisel_download) 
```

website/Makefile
```diff
@@ -62,6 +65,20 @@ $(latest_scaladoc_dir)/index.html: $(latest_version) | $(latest_scaladoc_dir)
...
+	$(eval VER := $(shell cat $(latest_version)))
+	$(generate_scala_cli) $(VER) > $@
+
+$(snapshot_chisel_download):
+	mkdir -p $@
+
+$(snapshot_chisel_download)/chisel-example.scala: $(snapshot_chisel_download)
```
seldridge, line 78: Order-only prerequisite:

``` suggestion
$(snapshot_chisel_download)/chisel-example.scala: | $(snapshot_chisel_download)
```

seldridge (commented):
SGTM to leave in the snapshot repo. No harm in that!

Thanks for this.

PR #4728, opened 2025-02-25
Use DedupGroup Phase in ChiselStage$

seldridge:
Fix a bug where the dedup group phase was only run for the ChiselStage class and not its companion object.  Add this phase.

#### Release Notes

Fix bug where dedup groups would have no effect if methods of `ChiselStage$` (the object), e.g., `ChiselStage.emitSystemVerilog`, were used.

### DIFF
--- a/src/main/scala/circt/stage/ChiselStage.scala
+++ b/src/main/scala/circt/stage/ChiselStage.scala
@@ -57,6 +57,7 @@ object ChiselStage {
     Seq(
       Dependency[chisel3.stage.phases.Elaborate],
       Dependency[chisel3.stage.phases.Convert],
+      Dependency[chisel3.stage.phases.AddDedupGroupAnnotations],
       Dependency[circt.stage.phases.AddImplicitOutputFile],
       Dependency[chisel3.stage.phases.AddImplicitOutputAnnotationFile],
       Dependency[circt.stage.phases.Checks],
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
good catch

seldridge:
The failure here is actually a bug: https://github.com/chipsalliance/chisel/issues/4730

I'm going to work around this by changing the documentation.

docs/src/cookbooks/hierarchy.md
```diff
@@ -352,11 +352,12 @@ class AddOne(val width: Int) extends Module {
   out := in + 1.U
 }
 
-class AddTwo(addOneDef: Definition[AddOne]) extends Module {
-  val i0 = Instance(addOneDef)
-  val i1 = Instance(addOneDef)
-  val in  = IO(Input(UInt(addOneDef.width.W)))
-  val out = IO(Output(UInt(addOneDef.width.W)))
+class AddTwo(addOneDef: => Definition[AddOne]) extends Module {
+  private val definition = addOneDef
```
jackkoenig, line 356: What does this change do?

seldridge: I can't remember... Let me try backing it out. That seems like it should be the same thing, except that the `Definition` will be evaluated _once_ inside `AddTwo` instead of either _once_ outside or _multiple times_ inside.

seldridge: Removed in ec59a692c.

PR #4744, opened 2025-02-27
[test] Fix color codes in WarningConfigurationSpec

seldridge:
Fix a bug in the WarningConfigurationSpec where it would fail locally due to ANSI color codes being printed.  Avoid this by switching to FileCheck and using a regular expression to ignore them.

I found this while doing some work on this test locally. I couldn't get it to pass without this.

You should be able to test locally with: 

```
./mill 'chisel[2.13.15].test.testOnly' 'chiselTests.stage.WarningConfigurationSpec'
```

I am not sure why this was not a problem in CI.


jackkoenig:
It probably doesn't fail in CI because Chisel will detect not to use color codes in CI but running locally it will not.

PR #4748, opened 2025-02-27
Add Serializer for ChiselIR

jackkoenig:
Future work TODO:
- [ ] Stop running the Converter
- [ ] Stop using firrtl.ir.Type in Properties
- [ ] Stop using firrtl.ir.Type in type aliases
- [ ] Delete the Converter
- [ ] Delete firrtl.ir.Serializer
- [ ] Lazily serialize Annotations
- [ ] Lazily serialize Module ports

### Contributor Checklist

#### Type of Improvement

- Performance improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

This reduces peak memory use when serializing to `.fir`. It heavily depends on the design, but for designs with a single very large module, we have measured 20-25% reductions in peak memory use. For most cases, the reduction will likely be more modest.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -154,7 +154,7 @@ private[chisel3] object ir {
   }
 
   case class ULit(n: BigInt, w: Width) extends LitArg(n, w) {
-    def name:     String = "UInt" + width + "(0h0" + num.toString(16) + ")"
+    def name:     String = "UInt" + width + "(0h" + num.toString(16) + ")"
     def minWidth: Int = (if (w.known) 0 else 1).max(n.bitLength)
 
     def cloneWithWidth(newWidth: Width): this.type = {
--- a/src/test/scala-2/chiselTests/BlackBox.scala
+++ b/src/test/scala-2/chiselTests/BlackBox.scala
@@ -8,6 +8,7 @@ import chisel3.experimental._
 import chisel3.reflect.DataMirror
 import chisel3.simulator.scalatest.ChiselSim
 import chisel3.simulator.stimulus.RunUntilFinished
+import chisel3.testing.scalatest.FileCheck
 import chisel3.util._
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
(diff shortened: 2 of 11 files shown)


jackkoenig:
I think this is good to go to get the ball rolling. I think the TODO items should be done in separate PRs.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -154,7 +154,7 @@ private[chisel3] object ir {
   }
 
   case class ULit(n: BigInt, w: Width) extends LitArg(n, w) {
-    def name:     String = "UInt" + width + "(0h0" + num.toString(16) + ")"
+    def name:     String = "UInt" + width + "(0h" + num.toString(16) + ")"
```
seldridge, line 157: This is the kind of thing that _could_ cause a bug. However, this looks fine. This should also cause simulations to fail if it was a problem (hopefully).

This is 💯 cleaner output, though.

src/test/scala-2/chiselTests/BlackBox.scala
```diff
@@ -8,6 +8,7 @@ import chisel3.experimental._
 import chisel3.reflect.DataMirror
 import chisel3.simulator.scalatest.ChiselSim
 import chisel3.simulator.stimulus.RunUntilFinished
+import chisel3.testing.scalatest.FileCheck
```
seldridge, line 11: 🐐

seldridge (approved):
Superficial review given the mechanical nature of this. 

Thank you for doing this tedious work.

What is the path now to deprecate the emission flows that we _don't_ want to support?

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: Yeah I could back this out--this change will be needed once I updated Properties to stop using FIRRTL IR and the converter, but it seems I backed out the rest of that change but not this. I do think we should still do this because it makes it consistent with what the serializer is doing (and ultimately we can call this from the serializer). I'll check for issues though of course.

jackkoenig:
> What is the path now to deprecate the emission flows that we don't want to support?

Deprecate the `Convert` phase any APIs that imply it (e.g. `ChiselStage.convert`). We can either backport the deprecation to 6.7.0 or just do it in 7.0, I'm not that fussed either way.

The other piece is to just make sure nothing about the Converter nor firrtl.ir.Serializer are load bearing. There's a couple of small pieces that are (types in Properties as I've mentioned).

jackkoenig:
I tested this on a large set of designs and found no Verilog diff except like a dozen literal assignments reordered (out of millions of lines of Verilog). Obviously identical by inspection.

Note that the emitted FIRRTL will differ slightly, primarily in these 3 ways:
* The current (old) emitter puts a couple of blank new lines after layerblocks, this new one does not
* This new emitter puts one extra new line after module bodies. I probably should fix this but have already qualified it as is so probably in follow on work.
* This current (old) emitter won't emit SInt literals, it converts them to UInts then wraps in `asUInt`. The new emitter just emits the SInt literal. This is a real change thus the need to check that Verilog is identical.

PR #4749, opened 2025-02-27
[testing] Add FileCheck API

seldridge:
Add an extension-method based FileCheck API.  This builds off of @jackkoenig's original FileCheck trait inside ChiselSpec and repackages it for publishing.

This makes a number of deviations from the original version:

  1. This opts for extension methods.
  2. This does not add helpers for running Chisel.
  3. This throws dedicated assertions on failures.

Taking (1) and (2) together, this is intended to _not_ duplicate methods which are already availble via ChiselStage.  Specifically, I am intending the API to be used like:

    ChiselStage.emitSystemVerilgo(new Foo).fileCheck()("CHECK: module Foo")

Things are done this way to try to make the APIs as orthogonal as possible without too many redundant ways to accomplish the same thing.

(3): is added for convenience of testing.  On failure, this provides an output that approximates what you get from lit (LLVM's Integrated Tester). This additionally provides a rerun script that you can just copy-paste into a terminal to reproduce the failure.

For ease of reproduction, both the input and the check are written to files with predictable names in a temporary directory underneath what is provided by an implementation of the `HasTestingDirectory` type class. Having both the input and check file are necessary to do easy reruns.

This is stacked on #4750.

#### Release Notes

Add `chisel3.testing.FileCheck` and `chisel3.testing.scalatest.FileCheck` trait for writing [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html)-based tests. This is useful for testing generators where you want to examine the output (FIRRTL or Verilog) as opposed to run a simulation. This is intended to complement ChiselSim.

### DIFF
--- a/src/main/scala/chisel3/testing/FileCheck.scala
+++ b/src/main/scala/chisel3/testing/FileCheck.scala
@@ -1,0 +1,133 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.testing
+
+import firrtl.options.StageUtils.dramaticMessage
+import java.io.{ByteArrayOutputStream, IOException, PrintWriter}
+import java.nio.file.{Files, StandardOpenOption}
+import scala.Console.{withErr, withOut}
+import scala.util.control.NoStackTrace
+import scala.sys.process._
+
+object FileCheck {
+
+  object Exceptions {
+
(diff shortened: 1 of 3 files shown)


jackkoenig (approved):
This is really slick!

PR #4750, opened 2025-02-27
[testing] Testing package w/ HasTestingDirectory

seldridge:
Create a new package, `chisel3.testing`, and move
`simulator.HasTestingDirectory` and the scalatest implementation of it to the new package.  This is done because the ability to get a testing directory is really a general property of "testing" and not "simulation".

This also builds towards moving the FileCheck utility out of `ChiselSpec` (i.e., only being available for tests) and into a proper home in `testing`.  It also makes sense for FileCheck to use the `HasTestingDirectory` type class as it has the same problem of needing to be given a "testing directory".


jackkoenig (approved):
I like this generalization

PR #4760, opened 2025-03-01
[docs] Add ChiselSim, FileCheck documentation

seldridge:
Add an explanation about testing in Chisel.  This discusses both the modern ChiselSim APIs and FileCheck APIs.  Both are discussed as they represent complementary testing strategies for users.

As the formal and inlinen testing comes online, this documentation is expected to be enhanced to include those.

FYI: @fabianschuiki, @tmckay-sifive


jackkoenig (approved):
Looks great! We probably need to add FileCheck to the installation page: https://www.chisel-lang.org/docs/installation but that's a separate PR IMO. It may also be worth calling out that simulation requires Verilator or VCS and FileCheck requires FileCheck but up to you.

seldridge:
@jackkoenig: I added some info blocks indicating that users need to install tools.

At least for FileCheck, there is a dedicated "FileCheck not found (did you forget to install it?)" assertion / verbose error message which should help herd users in the right direction.

PR #4763, opened 2025-03-03
[simulator] Remove DefaultSimulator

seldridge:
Remove `DefaultSimulator`.  This is fully replaced with the `ChiselSim` API and is no longer needed.  Notably, `DefaultSimulator` never made it onto an actual release of Chisel so this is safe to just remove.

Rewrite EphemeralSimulator in terms of ChiselSim APIs.  Migrate tests of DefaultSimulator to tests on ChiselSim or other simulator classes, objects, and methods.

This does not have release notes, but we should delete the release notes about `DefaultSimulator` when releasing Chisel 7.


src/main/scala/chisel3/simulator/EphemeralSimulator.scala
```diff
@@ -18,12 +18,14 @@ import java.io.File
...
+  private val chiselSim = new ChiselSim {}
+
   def simulate[T <: RawModule](
     module:       => T,
     layerControl: LayerControl.Type = LayerControl.EnableAll
   )(body: (T) => Unit): Unit = {
-    implicit val temporary = HasTestingDirectory.temporary(deleteOnExit = true)
-    DefaultSimulator.simulateRaw(
+    implicit val temporary: HasTestingDirectory = HasTestingDirectory.temporary(deleteOnExit = true)
+    chiselSim.simulateRaw(
```
jackkoenig, line 28: Is the `ChiselSim` object stateful? I assume you're sharing it between multiple runs for a reason (presumably more performant if it isn't stateful) but just want to check that's intentional.

seldridge: No mutability other than that exposed by `TestingDirectory`'s dynamic variable `testName` which is set from a fixture.

Or: other than that, it is definitely immutable as it is just packaging up `SimulatorAPI` (just methods) with `PeekPokeAPI` (methods and classes).

This should be safe.

jackkoenig (approved):

Issue #4733, opened 2025-02-26
literal.asUInt(width) always returns a Bool


tymcauley:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**

```scala
//> using scala 2.13.15
//> using repository https://s01.oss.sonatype.org/content/repositories/snapshots
//> using dep org.chipsalliance::chisel:7.0.0-M2+432-486b55dc-SNAPSHOT
//> using plugin org.chipsalliance:::chisel-plugin:7.0.0-M2+432-486b55dc-SNAPSHOT
//> using options -unchecked -deprecation -language:reflectiveCalls -feature -Xcheckinit -Xfatal-warnings -Ywarn-dead-code -Ywarn-unused -Ymacro-annotations

import chisel3._

class Foo(x: BigInt, width: Int) extends Module {
  val out1 = IO(Output(UInt(width.W)))
  val out2 = IO(Output(UInt(width.W)))
  val out3 = IO(Output(UInt(width.W)))
  out1 := x.asUInt(width)
  out2 := x.asUInt
  out3 := x.U(width.W)
}

object Main extends App {
  import _root_.circt.stage.ChiselStage
  val width = 8
  println(
    ChiselStage.emitSystemVerilog(
      gen = new Foo((1 << width) - 1, width),
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info", "-default-layer-specialization=enable")
    )
  )
}
```

Running this with `scala-cli` generates this result:

```verilog
// Generated by CIRCT firtool-1.107.0
module Foo(
  input        clock,
               reset,
  output [7:0] out1,
               out2,
               out3
);

  assign out1 = 8'h0;
  assign out2 = 8'hFF;
  assign out3 = 8'hFF;
endmodule
```

**What is the current behavior?**

Running `x.asUInt(width)` always returns a `Bool` (the same happens if you change `x` from a `BigInt` to an `Int`). If you change the `out1 := x.asUInt(width)` line above to use the `:<=` operator, you'll get this error:

```
[error] ./chisel-example.scala:13:3
[error] Cannot prove that chisel3.UInt =:= chisel3.Bool.
[error]   out1 :<= x.asUInt(width)
[error]   ^^^^^^^^^^^^^^^^^^^^^^^^
```

**What is the expected behavior?**

[...]

PR #4764, opened 2025-03-03
Warn on <literal>.asUInt|.asSInt(_: Int)

jackkoenig:
Fixes https://github.com/chipsalliance/chisel/issues/4733

Unfortunately, this is really hard to write a unit test for, but I did manually check it works (and it's using the same approach as we have been using for `.U` and `.S`). I'm also not sure that this is the _best_ fix for 4733, but it works.

FYI @tymcauley 

### Contributor Checklist

#### Type of Improvement

- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

The user probably forgot .W. Apply the same technique as used for .U|.S.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -30,10 +30,6 @@ package object chisel3 {
     * after this call using apply, ie. 0.asUInt(1)(0) due to potential for
     * confusion (the 1 is a bit length and the 0 is a bit extraction position).
     * Prefer storing the result and then extracting from it.
-    *
-    * Implementation note: the empty parameter list (like `U()`) is necessary to prevent
-    * interpreting calls that have a non-Width parameter as a chained apply, otherwise things like
-    * `0.asUInt(16)` (instead of `16.W`) compile without error and produce undesired results.
     */
   implicit class fromBigIntToLiteral(bigint: BigInt) {
 
--- a/macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala
+++ b/macros/src/main/scala/chisel3/internal/sourceinfo/SourceInfoTransform.scala
@@ -361,13 +361,13 @@ class IntLiteralApplyTransform(val c: Context) extends AutoSourceTransform {
     c.macroApplication match {
       case q"$_.$clazz($lit).$func.apply($arg)" =>
         if (
-          Set("U", "S").contains(func.toString) &&
+          Set("U", "S", "asUInt", "asSInt").contains(func.toString) &&
           Set("fromStringToLiteral", "fromIntToLiteral", "fromLongToIteral", "fromBigIntToLiteral").contains(
             clazz.toString
           )
         ) {
           val msg =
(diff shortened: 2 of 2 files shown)


mwachs5:
what is the actual change in behavior here? Without a test it's hard to tell

jackkoenig:
> what is the actual change in behavior here? Without a test it's hard to tell

```scala
val x: Int = 3
x.asUInt(8) // this is now a compile error
4.asUInt(10) // this is now a compile error
```

Full example:
```scala
class Foo(a: Int) extends Module {
  val x, y, z = IO(Output(UInt(8.W)))
  x := a.asUInt(8)
  y := 4.asUInt(8)
  z := 5.asUInt(8.W)
}
```
This warns with (but I have warnings as errors so its an error):
```
[error] /Users/koenig/work/chisel-asuint-lint/src/main/scala/chisel-example.scala:13:16: Passing an Int to .asUInt is usually a mistake: It does *not* set the width; it does a bit extraction.
[error] Did you mean .asUInt(8.W)?
[error] If you do want bit extraction, use .asUInt.extract(8) instead.
[error]   x := a.asUInt(8)
[error]                ^
[error] /Users/koenig/work/chisel-asuint-lint/src/main/scala/chisel-example.scala:14:16: Passing an Int to .asUInt is usually a mistake: It does *not* set the width; it does a bit extraction.
[error] Did you mean .asUInt(8.W)?
[error] If you do want bit extraction, use .asUInt.extract(8) instead.
[error]   y := 4.asUInt(8)
[error]                ^
[error] two errors found
```

mwachs5:
what about

```
z := VecInit(1.U, 2.U, 3.U).asUInt(8)
```

what's the argument for keeping this fine?

jackkoenig:
> what about
> 
> ```
> zz := VecInit(1.U, 2.U, 3.U).asUInt(8)
> ```
> 
> what's the argument for keeping this fine?

Typically, the number you put is a width greater than the automatic width of the argument and you'll get an out-of-bounds index error. So basically, it's not usually fine.

That doesn't apply to literals for various reasons, arguably if you don't give them a width, they shouldn't really have a width so it's weird to enforce the minimum possible width on a subsequent bit extract. And subsequent bit extracts do sometimes happen, usually in loops or more complex control flow.

The main issue though, is that it's really hard to ban a chained apply after anything. @sequencer has suggested getting rid of apply for bit extraction, we could require everyone to explicitly write `.extract(bit)` or `.extract(hi, lo)`. It certainly would avoid ambiguity issues this.

But rather than trying to tackle something so big, I figured I'd apply the fix we already have for `.U` and `.S` to `.asUInt` and `.asSInt`.

tymcauley:
Looks great to me!

To be clear, this is emitted as a _Scala_ warning, not a Chisel warning (as in [this](https://www.chisel-lang.org/docs/explanations/warnings)), correct? Presumably because we don't have access to the `Builder` in this macro?

mwachs5 (approved):

PR #4765, opened 2025-03-04
[svsim] Add clangformat target for svsim

seldridge:
The svsim project contains one non-trivial C++ file.  I need to make some relatively heavy edits to this and I would like to not worry about formatting. This adds both a mill target, `clangformat` and a simple `.clang-format` file to the project. It then formats the file.

### DIFF
--- a/svsim/package.mill
+++ b/svsim/package.mill
@@ -25,4 +25,14 @@ trait Svsim extends CrossSbtModule with HasCommonOptions with ScalafmtModule {
   object test extends SbtTests with TestModule.ScalaTest with ScalafmtModule {
     def ivyDeps = Agg(v.scalatest, v.scalacheck)
   }
+
+  def clangformat = Task {
+    os.proc(
+      "clang-format",
+      "-i",
+      "--verbose",
+      s"--style=file:${os.pwd / os.RelPath("svsim/.clang-format")}",
+      os.pwd / os.RelPath("svsim/src/main/resources/simulation-driver.cpp")
+    ).call(cwd = Task.dest)
+  }
 }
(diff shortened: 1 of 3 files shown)


svsim/package.mill
```diff
@@ -25,4 +25,14 @@ trait Svsim extends CrossSbtModule with HasCommonOptions with ScalafmtModule {
...
+
+  def clangformat = Task {
+    os.proc(
+      "clang-format",
+      "-i",
+      "--verbose",
+      s"--style=file:${os.pwd / os.RelPath("svsim/.clang-format")}",
+      os.pwd / os.RelPath("svsim/src/main/resources/simulation-driver.cpp")
+    ).call(cwd = Task.dest)
+  }
```
jackkoenig, line 37:
```suggestion
  def clangFormatConfig = Task.Source { os.pwd / "svsim" / ".clang-format" }
  
  def cppSources = Task.Sources {
    Agg(os.pwd / "svsim/src/main/resources/simulation-driver.cpp")
  }

  def clangformat: Command[Unit] = Task.Command {
    os.proc(
      "clang-format",
      "-i",
      "--verbose",
      s"--style=file:${clangFormatConfig())}",
      cppSources(): _*
    ).call(cwd = Task.dest)
  }
```
I don't think this diff is exactly right as is but something like the above

A couple of minor suggestions, use `Task.Source[s]` for source files, make the clangFormat def a Command instead of just Task (which Mill might try to cache).

seldridge: I think I have this working in: https://github.com/chipsalliance/chisel/pull/4765/commits/875680d57fc2c35cadeb80f50c5ed74238eda03e

jackkoenig (approved):
LGTM!

PR #4767, opened 2025-03-04
[svsim] Cleanup simulation-driver.cpp

seldridge:
Minor cleanup of simulation-driver.cpp:

1. Move all global mutable state into a single static struct.
2. Mark everything that is only used internally to the module as `static`.

I'm not crazy about the solution of (1). However, the order of calls necessitates something like this. What is going on:

1. `main` runs and sets up global mutable state.
2. `main` calls `simulation_main`.
3. `simulation_main` runs, dropping into Verilog.
4. Verilog calls `simulation_body` via DPI.
5. `simulation_body` needs access to things created by (1).

I'd much rather have a way to _not_ have global mutable state.

This is stacked on #4765.

### DIFF
--- a/svsim/src/main/resources/simulation-driver.cpp
+++ b/svsim/src/main/resources/simulation-driver.cpp
@@ -405,8 +405,8 @@ static int scanInt(const char **lineCursor, const char *description) {
   return (int)value;
 }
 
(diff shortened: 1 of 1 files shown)


seldridge:
I talked about this with @darthscsi offline and he said to just use a global for this. This saves on code and is simpler / more expected.

PR #4769, opened 2025-03-04
[svsim] Avoid EndOfMessages Exception

### LINKED ISSUE #4700: [Chiselsim] Exit simulation on $finish

seldridge:
Add a `final` block to svsim's Verilog which will call a DPI function to finish the simulation.  This causes the simulator to either exit (if a done command has been received) or wait until it sees _excatly_ a done command.

Because of the different ways that simulators can choose to implement time delays, the simulator _may_ have outstanding transactions that need to be finished.  Finish any outstanding transactions before blocking on a done command.  Use this new outstanding transaction state to implement the old `receivedDone` state.

This is done to get equivalent behavior between Verilator and VCS when a Verilog `$finish` is executed.

#### Release Notes

- Rework svsim to use a `final` block that will finish any outstanding transactions. This fixes a class of bugs where VCS simulations could generate `UnexpectedEndOfMessages` exceptions and need to be caught by the user. This fixes #4700 for VCS.


svsim/src/main/resources/simulation-driver.cpp
```diff
@@ -769,29 +795,25 @@ static void processCommand() {
...
-      }
+      run_simulation(timestepsPerPhase, &finish);
+      if (finish)
+        return true;
     }
 
-    cycles--; // Consume the unbalanced increment from the while condition
-    sendUintAsBits(cycles);
+    // Consume the unbalanced increment from the while condition
+    sendUintAsBits(--cycles);
```
jackkoenig, line 808: Is it really worth merging these into one line?

seldridge: I tend to like this terseness. However, I can see it being confusing.

Note: the more important change here is a change from postincrement to preincrement. This, doesn't make a difference here at all, however, I've gotten the [LLVM Coding Guidelines which basically says "never use postincrement"](https://llvm.org/docs/CodingStandards.html#prefer-preincrement) beaten into me through code reviews I will always convert something to preincrement when I see it. (Again, this doesn't make a difference as postincrement isn't slow for primitive types, but it's annoying to remember when postincrement could be expensive, so LLVM just banned it.)

After making the postincrement to preincrement change this can then be inlined and I will tend to inline anything single use (which I realize is different from others which prefer things broken out).

PR #4771, opened 2025-03-05
[testing] Cleanup default and Scalatest directory structure

seldridge:
Change the default directory that `HasTestingDirectory` uses from
`test_run_dir/chiselsim` to `build/chiselsim`.  Generally, `test_run_dir/`
is an older style and `build/` is both newer (for Chisel) and more
standard (generally).

Change the return type of the abstract method `buildDir` to be `Path`
instead of `String`.  This is better as it avoids problems with separators
which could plague Windows users.

Change the directory that Scalatest tests use from being rooted at
`build/` to being rooted at `build/chiselsim/`.  The former can have a lot
of contention with many output files and it would be better to add this
additional level.  Also, this aligns this with the new default behavior of
the non-Scalatest WithTestingDirectory.

This does not need to get release notes as these APIs have never hit a release.

CC: @mmaloney-sf: This fixes that `test_run_dir/` that you observed and switches things to `build/chiselsim/`.

### DIFF
--- a/src/main/scala/chisel3/testing/HasTestingDirectory.scala
+++ b/src/main/scala/chisel3/testing/HasTestingDirectory.scala
@@ -27,7 +27,7 @@ object HasTestingDirectory {
     * E.g., this may produce something like:
     *
     * {{{
-    * test_run_dir
+    * build
     * └── chiselsim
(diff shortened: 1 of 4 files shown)


jackkoenig (approved):
LGTM!

PR #4772, opened 2025-03-05
[scalatest] Add subdirectory control

seldridge:
Add a new overridable member function to `scalatest.TestingDirectory` which allows a user to set a subdirectory for a test.  The primary use case for this is to allow the user to set a directory derived from a seed for a test (or a timestamp).  This is not the standard behavior. Normally, you only have one output directory and you are fine to override it.  However, we have internal uses where these directories need to be split up based on seed.

#### Release Notes

- Add subdirectory customization to `chisel3.testing.scalatest.TestingDirectory` via an overridable member, `subDir`. This can be used to have multiple runs of a test output to different subdirectories, e.g., to differentiate runs of the same test by time or by seed.

### DIFF
--- a/src/main/scala/chisel3/testing/scalatest/WithTestingDirectory.scala
+++ b/src/main/scala/chisel3/testing/scalatest/WithTestingDirectory.scala
@@ -25,6 +25,26 @@ import scala.util.DynamicVariable
   * You may change the `buildDir` by overridding a method of the same name in
   * this trait.
   *
+  * If multiple tests are being run with the same directory, you may override
+  * the optional `subDir` to add hierarchy under final test directory.  This is
+  * typically used to add information about a test seed if one is used.
+  *
+  * E.g., the above directory structure can be modified to create a directory
+  * structure like the following:
+  *
+  * {{{
+  * <buildDir>
+  * └── <suite-name>
+  *     └── <scope-1-name>
+  *         └── ...
+  *             └── <scope-n-name>
+  *                 ├── <test-1-name>
+  *                 │   └── <subDir>
+  *                 ├── ...
+  *                 └── <test-n-name>
+  *                     └── <subDir>
+  * }}}
+  *
   */
 trait TestingDirectory { self: TestSuite =>
 
@@ -33,6 +53,12 @@ trait TestingDirectory { self: TestSuite =>
     * For different behavior, please override this in your test suite.
     */
   def buildDir: Path = Paths.get("build", "chiselsim")
+
+  /** An option subdirectory to put the test in _under_ theh final test directory.
+    * This is typically used to differentiate different runs of the same test,
+    * e.g., if the test is given a seed.
+    */
+  def subDir: Option[Path] = None
 
   // Assemble all the directories that should be created for this test.  This is
   // done by examining the test (via a fixture) and then setting a dynamic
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/testing/scalatest/WithTestingDirectory.scala
```diff
@@ -34,6 +54,12 @@ trait TestingDirectory { self: TestSuite =>
     */
   def buildDir: Path = Paths.get("build", "chiselsim")
 
+  /** An option subdirectory to put the test in _under_ theh final test directory.
+    * This is typically used to differentiate different runs of the same test,
+    * e.g., if the test is given a seed.
+    */
```
jackkoenig, line 60:
```suggestion
  /** An option subdirectory to put the test in _under_ the final test directory.
    *
    * This is typically used to differentiate different runs of the same test,
    * e.g., if the test is given a seed.
    */
```
In ScalaDoc it's important to separate the first line as a "short description" from the longer "full description"

src/test/scala-2/chiselTests/testing/scalatest/TestingDirectorySpec.scala
```diff
@@ -90,4 +96,21 @@ class TestingDirectorySpec extends AnyFunSpec with Matchers with SimulatorAPI wi
...
+      seed.withValue(Some("foo")) {
+        checkDirectoryStructure(
+          "build",
+          "chiselsim",
+          "TestingDirectorySpec",
+          "A-test-that-sets-a-subdirectory",
+          "should-have-a-subdirectory",
+          "foo"
+        )
+      }
```
jackkoenig, line 111: I'm a little iffy on this being done via a DynamicVariable rather than by changing the `HasTestingDirectory`. What I'm proposing works a lot better in Scala 3 with Context Functions, but works in Scala 2 as well, a couple of ways it can work:

Provide a new implicit val:
```suggestion
      // Scala 2
      implicit val testDir: HasTestingDirectory = implicitly[HasTestingDirectory].withSeed("foo")
      // Scala 3
      given HasTestingDirectory = summon[HasTestingDirectory].withSeed("foo")
      checkDirectoryStructure(
        "build",
        "chiselsim",
        "TestingDirectorySpec",
        "A-test-that-sets-a-subdirectory",
        "should-have-a-subdirectory",
        "foo"
      )
```

Or with context functions (and Scala 2's crappy version):

```suggestion
      // Scala 2
      testingDirectoryWithSeed(Some("foo")) { implicit testDir =>
      // Scala 3
      testingDirectoryWithSeed(Some("foo")) {
        checkDirectoryStructure(
          "build",
          "chiselsim",
          "TestingDirectorySpec",
          "A-test-that-sets-a-subdirectory",
          "should-have-a-subdirectory",
          "foo"
        )
      }
```

jackkoenig: I'm not opposed your approach here and it doesn't preclude changing to the approach I'm proposing in the future (especially in the Brave New World of Scala 3), just something to consider.

jackkoenig (approved):
Notwithstanding my thoughts, I'm fine with this as is

seldridge:
I had an offline discussion with Jack about this. Based on how `TestingDirectory` works today, we don't need this. This same functionality can be achieved by using a Scalatest spec which allows for nesting scopes _and putting the seed as the final scope_ (or another scope if you want that organization). This does mean you cannot use subdirectories for certain testing styles that do not allow deep scope nesting, e.g., flat spec. However, that is fine.

PR #4773, opened 2025-03-05
[chiselsim] Allow command line options to be passed to ChiselSim tests

seldridge:
This is three commits that should be reviewed independently.  This builds towards and then adds a `HasConfigMap` trait which allows for `-D<key>=<value>` options to be exposed to `chisel3.simulator.scalatest.ChiselTest`. In order to do this, and without starting to add functionality directly to `ChiselSim`, the existing `TestingDirectory` trait is refactored to be made stackable.

#### Release Notes

Add the ability for users to pass command-line options, via `-D<key>=<value>`, to any `chisel3.simulator.scalatest.ChiselTest` test for them to customize the execution of the test.

### DIFF
--- a/src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
+++ b/src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
@@ -31,14 +31,18 @@ trait HasConfigMap extends TestSuiteMixin { self: TestSuite =>
   // Implement this via a `DynamicVariable` pattern that will be set via the
   // `withFixture` method.  The `super.withFixture` function must be called to
   // make this mix-in "stackable" with other mix-ins.
-  private val _configMap: DynamicVariable[ConfigMap] = new DynamicVariable[ConfigMap](ConfigMap.empty)
+  private val _configMap: DynamicVariable[Option[ConfigMap]] = new DynamicVariable[Option[ConfigMap]](None)
   abstract override def withFixture(test: NoArgTest) = {
-    _configMap.withValue(test.configMap) {
+    _configMap.withValue(Some(test.configMap)) {
       super.withFixture(test)
     }
   }
 
-  /** Return command line options passed to the test as a `Map`. */
-  def configMap: Map[String, Any] = _configMap.value
+  /** Return the config map which contains all command line options passed to Scalatest.
+    *
+    * This is only valid during a test.  It will be `None` if used outside a
+    * test.
+    */
+  def configMap: Option[ConfigMap] = _configMap.value
 
 }
--- a/src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
+++ b/src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
@@ -42,7 +42,11 @@ trait HasConfigMap extends TestSuiteMixin { self: TestSuite =>
     *
     * This is only valid during a test.  It will be `None` if used outside a
     * test.
+    *
+    * @throws RuntimeException if called outside a Scalatest test
     */
-  def configMap: Option[ConfigMap] = _configMap.value
+  def configMap: ConfigMap = _configMap.value.getOrElse {
+    throw new RuntimeException("configMap may only be accessed inside a Scalatest test")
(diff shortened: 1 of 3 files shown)


seldridge:
Testing this a bit more, I may need another approach as the `configMap` is only valid _inside_ the test and not outside it. If you are trying to use a `-D` option to set a seed and have that used in a test name, this won't work. There are a couple of other ways to get at Scalatest's config map that I'm going to look at.

seldridge:
> 

> Testing this a bit more, I may need another approach as the configMap is only valid inside the test and not outside it. If you are trying to use a -D option to set a seed and have that used in a test name, this won't work. There are a couple of other ways to get at Scalatest's config map that I'm going to look at.

I was able to workaround this problem with https://github.com/chipsalliance/chisel/pull/4774. That provides the ability to get a new `HasTestingDirectory` with a subdirectory. This can be used to manually change the directory of calls to `simulate` (or other methods which need a `HasTestingDirectory` type class implementation).

src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.util.DynamicVariable
+
+/** A Scalatest test suite mix-in that provides access to command line options.
+  *
+  * This expose any keys/values passed using the `-D<key>=<value` command line
+  * option to the test via a `Map`.
+  *
+  * For example, you can invoke Scalatest passing the `foo=bar` option like so:
+  * {{{
+  * ./mill 'chisel[2.13.16].test.testOnly' 'fooTest' -Dfoo=bar
```
jackkoenig, line 15:
```suggestion
  * ./mill chisel[2.13.16].test.testOnly fooTest -Dfoo=bar
```

This is the nit of nits but those quotes aren't necessary

src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
```diff
@@ -0,0 +1,48 @@
...
+      super.withFixture(test)
+    }
+  }
+
+  /** Return the config map which contains all command line options passed to Scalatest.
+    *
+    * This is only valid during a test.  It will be `None` if used outside a
+    * test.
+    */
+  def configMap: Option[ConfigMap] = _configMap.value
```
jackkoenig, line 46: I understand the motivation for making it an `Option`, but I'm wondering if just returning `ConfigMap` and erroring if it isn't available might be more desirable. Is there ever a case that one would want the `None` over the exception? Maybe we provide both but make the default API error? e.g.

```suggestion
  /** Return the config map which contains all command line options passed to Scalatest.
    *
    * This is only valid during a test. It throw an exception if used outside a test.
    */
  def configMap: Option[ConfigMap] = configMapOpt.getOrElse {
    throw new Exception("ScalaTest configMap is only available within tests (e.g. it(\"works\") { <here> })")
  }

  /** Return the config map which contains all command line options passed to Scalatest.
    *
    * This is only valid during a test. It will be `None` if used outside a
    * test.
    */  
  def configMapOpt: Option[ConfigMap] = _configMap.value
```

jackkoenig (approved):
LGTM but a couple of suggestions

src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
seldridge: The logic was that it would be better to not have an exception. However, I think it's fine to only provide that. I _really_ don't want people to even be looking at this unless it's in a test. I'll change to _only_ provide the API which can throw.

seldridge: Switched to the assertion approach and added tests in: 904add2db

src/main/scala/chisel3/testing/scalatest/HasConfigMap.scala
seldridge: SGTM. Those quotes are unnecessary. Fixed in: 70c114bc9

PR #4774, opened 2025-03-05
[testing] Add HasTestingDirectory subdir factory

seldridge:
Add a new member function to `HasTestingDirectory` called `withSubdirectory`.  This is a factory of _new_ `HasTestingDirectory` objects which will return directories nested under the original.

This is intended to fit a very specific use case where a user is passing command line arguments through Scalatest's ConfigMap API (i.e., `-D<key>=<value>`) _and they want to use the information in the ConfigMap to affect the directory structure_.  Because the ConfigMap is only availalbe _inside the test_, it cannot be used to set the test name.  If a user is not passing arguments via the ConfigMap, then a user can instead structure the ScalaTest scopes in such a way to get the directory structure they want.

#### Release Notes

Add `withSubdirectory` factory to `HasTestingDirectory` type class.  This can be used to create new `HasTestingDirectory` objects with directories under the original. This can be used to affect the directory inside a Scalatest test if the information a user has is only available inside the test (e.g., Scalatest's ConfigMap).


src/main/scala/chisel3/testing/HasTestingDirectory.scala
```diff
@@ -14,6 +14,30 @@ trait HasTestingDirectory {
...
+    * val bar = foo.getSubDir("bar")
+    * }}}
+    *
+    * @param subdirectory a subdirectory
+    */
+  def withSubdirectory(subdirectory: String): HasTestingDirectory = new HasTestingDirectory {
+
+    override def getDirectory = _getDirectory.resolve(subdirectory)
+
+  }
```
jackkoenig, line 39: Your `_getDirectory` pattern is fine, but the more common approach is to alias "this" (aka self-type with no type) the outer class:
```scala
trait HasTestingDirectory { outer =>
  def withSubdirectory(subdirectory: String): HasTestingDirectory = new HasTestingDirectory {
    override def getDirectory = outer.resolve(subdirectory)
  }
}

jackkoenig (approved):
LGTM

src/main/scala/chisel3/testing/HasTestingDirectory.scala
seldridge: Good suggestion. Fixed in https://github.com/chipsalliance/chisel/pull/4774/commits/b49bd29f63743e933dda1c0456e601ea61ccf87a.

PR #4776, opened 2025-03-07
[scalatest] Add HasCliSimulator

seldridge:
Add a new trait, `HasCliSimulator`, that builds on `HasConfigMap` to provide a standardized way to get a simulator that the user can choose at scalatest test time via a `-Dsimulator=<simulator-name>` argument.

Note: this is difficult to test, however, I've tested it manually.

CC: @andrew-gouldey-sifive: I'm planning to get this into the open source and then use it internally. We will customize the simulator used internally slightly for the different Verilator or VCS default options that we use.

#### Release Notes

Add `HasCliSimulator` which can be used, along with `HasConfigMap`, to allow for tests to choose their simulator at test time via a `-D` command line option. This is added to avoid fragmentation of command line options around choosing which simulator to run.

### DIFF
--- a/src/main/scala/chisel3/testing/scalatest/HasClilSimulator.scala
+++ b/src/main/scala/chisel3/testing/scalatest/HasClilSimulator.scala
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.testing.scalatest
+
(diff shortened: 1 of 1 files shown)


jackkoenig (approved):
Looks great!

PR #4779, opened 2025-03-08
[chiselsim] Add ControlAPI w/ Waveform Enable/Disable Support

seldridge:
Add a missing ChiselSim ControlAPI.  This currently only includes APIs to
enable and disalbe waveforms.  This API was previously inaccessible
because you would need to get at an svsim Controller to use it.

Note: the enable/disable waveform APIs will have no effect unless the
simulator is compiled with waveform support.

#### Release Notes

Add `ControlAPI` to ChiselSim. This adds `enableWaves` and `disableWaves` functions. These functions _require_ compiling a simulator with waveform support.

### DIFF
--- a/src/main/scala/chisel3/simulator/ControlAPI.scala
+++ b/src/main/scala/chisel3/simulator/ControlAPI.scala
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator
+
+trait ControlAPI {
+
+  /** Enable waveform dumping
+    *
+    * This control function will enable waveforms from the moment it is applied.
+    * A simulator must be compiled with waveform dumping support for this to
+    * have an effect.
+    *
+    * Example usage:
+    * {{{
+    * enableWaves()
+    * }}}
+    */
+  def enableWaves(): Unit = {
+    AnySimulatedModule.current.controller.setTraceEnabled(true)
+  }
+
+  /** Enable waveform dumping
+    *
+    * This control function will disable waveforms from the moment it is
+    * applied.  A simulator must be compiled with waveform dumping support for
+    * this to have an effect.
+    *
+    * Example usage:
+    * {{{
+    * disableWaves()
+    * }}}
+    */
+  def disableWaves(): Unit = {
+    AnySimulatedModule.current.controller.setTraceEnabled(false)
+  }
+
+}
--- a/src/main/scala/chisel3/simulator/package.scala
+++ b/src/main/scala/chisel3/simulator/package.scala
@@ -23,7 +23,7 @@ package object simulator {
     *
     * @see [[chisel3.simulator.scalatest.ChiselSim]]
     */
-  trait ChiselSim extends PeekPokeAPI with SimulatorAPI
+  trait ChiselSim extends ControlAPI with PeekPokeAPI with SimulatorAPI
 
   /**
     * An opaque class that can be passed to `Simulation.run` to get access to a `SimulatedModule` in the simulation body.
--- a/src/main/scala/chisel3/simulator/scalatest/package.scala
+++ b/src/main/scala/chisel3/simulator/scalatest/package.scala
@@ -22,6 +22,8 @@ package object scalatest {
     *
     * @see [[chisel3.simulator.ChiselSim]]
     */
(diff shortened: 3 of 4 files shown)


src/main/scala/chisel3/simulator/ControlAPI.scala
```diff
@@ -0,0 +1,37 @@
...
+    * have an effect.
+    *
+    * Example usage:
+    * {{{
+    * enableWaves()
+    * }}}
+    */
+  def enableWaves(): Unit = {
+    AnySimulatedModule.current.controller.setTraceEnabled(true)
+  }
```
jackkoenig, line 20: Is it possible for this to error if the simulator was not compiled with waveform dumping?

jackkoenig (approved):
See question. I don't love that this is global but that's not this PRs fault, good to expose.

src/main/scala/chisel3/simulator/ControlAPI.scala
seldridge: Frustratingly, I didn't see a clean way of doing this. I plan to come back and do this as part of `svsim` cleanup.

seldridge: I have a pure-Verilog solution to do this here: https://github.com/chipsalliance/chisel/pull/4787

This produces output like:

```
# ./mill 'chisel[2.13.16].test.testOnly' 'chiselTests.simulator.scalatest.FooSpec' -- -z "terminate before 11 cycles" 
[358/365] chisel[2.13.16].test.compile
[358] [info] compiling 1 Scala source to /Users/schuylereldridge/repos/github.com/chipsalliance/chisel/out/chisel/2.13.16/test/compile.dest/classes ...
[358] [info] done compiling
[365/365] chisel[2.13.16].test.testOnly
[365] FooSpec:
[365] Foo
[365] Bar
[365] - should terminate before 11 cycles have elapsed *** FAILED ***
[365]   chisel3.simulator.Exceptions$AssertionFailed: One or more assertions failed during Chiselsim simulation
[365] ------------------------------------------------------------------------------
[365] The following assertion failures were extracted from the log file:
[365] 
[365]   lineNo  line
[365]   ----------------------------------------------------------------------------
[365]        0  [4] %Error: testbench.sv:97: Assertion failed in TOP.svsimTestbench.simulation_enableTrace: Cannot enable waves as this simulator was not compiled to support them
[365] 
[365] For more information, see the complete log file:
[365] 
[365]   build/chiselsim/FooSpec/Bar/should-terminate-before-11-cycles-have-elapsed/workdir-verilator/simulation-log.txt
[365] ------------------------------------------------------------------------------
[365]   ...
[365] Run completed in 1 second, 970 milliseconds.
[365] Total number of tests run: 1
[365] Suites: completed 1, aborted 0
[365] Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
[365] *** 1 TEST FAILED ***
[365/365] ======================= chisel[2.13.16].test.testOnly chiselTests.simulator.scalatest.FooSpec -- -z terminate before 11 cycles ====================== 4s
1 tasks failed
chisel[2.13.16].test.testOnly 1 tests failed: 
  chiselTests.simulator.scalatest.FooSpec Bar should terminate before 11 cycles have elapsed
```

kammoh:
Excited to see these improvements finally landing on svsim 🎉
Really appreciate the great work!
With the current API, it possible to enable trace generation _before_ reset procedure completes?
Specifically, I’d like to have the reset assertion and deassertion present in the generated waveform.

seldridge:
> With the current API, it possible to enable trace generation before reset procedure completes?

I'm trying to figure out the right way to expose this via `SimulatorAPI.simulate` (or not to expose it). After https://github.com/chipsalliance/chisel/commit/a7effb7da1db69adeacb1853a2e74369a4e48609, the reset initialization procedure is factored out to reusable stimulus.

So, the way to do this today is to use `simulateRaw` like:

``` scala
simulateRaw(new Foo) { case dut => 
  enableWaves()
  ResetProcedure.module()(dut) // Or ResetProcedure.any if you don't have a Module
  // user stimulus
}
```

I'm considering adding an argument like `enableWavesAtTimeZero` argument to `simulate` which would do this for you. However, I'm somewhat concerned about the `simulate` API getting overly complex. That said, going the other direction and moving everything to use `simulateRaw` and then having that replace `simulate` gets tedious as every test needs the `ResetProcedure`.

Issue #4780, opened 2025-03-09
svsim failing on windows due to no 'rm' executable found


tobias1012:
**Type of issue**: Bug Report

**Please provide the steps to reproduce the problem:**
On a windows 10 machine run `sbt test` with verilator.

**What is the current behavior?**
Tests fail with `java.io.IOException: Cannot run program "rm": CreateProcess error=2`
**What is the expected behavior?**
Tests should run

**Please tell us about your environment:**

**Other Information**
`[info]   java.io.IOException: Cannot run program "rm": CreateProcess error=2, File not found
[info]   at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1170)
[info]   at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1089)
[info]   at java.base/java.lang.Runtime.exec(Runtime.java:681)
[info]   at java.base/java.lang.Runtime.exec(Runtime.java:530)
[info]   at svsim.Workspace.reset(Workspace.scala:54)
[info]   at chisel3.simulator.Simulator._simulate(Simulator.scala:102)
[info]   at chisel3.simulator.Simulator._simulate$(Simulator.scala:97)
[info]   at chisel3.simulator.EphemeralSimulator$DefaultSimulator._simulate(EphemeralSimulator.scala:28)
[info]   at chisel3.simulator.SingleBackendSimulator.simulate(Simulator.scala:146)
[info]   at chisel3.simulator.SingleBackendSimulator.simulate$(Simulator.scala:142)
[info]   ...
[info]   Cause: java.io.IOException: CreateProcess error=2, File not found
[info]   at java.base/java.lang.ProcessImpl.create(Native Method)
[info]   at java.base/java.lang.ProcessImpl.<init>(ProcessImpl.java:500)
[info]   at java.base/java.lang.ProcessImpl.start(ProcessImpl.java:159)
[info]   at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1126)
[info]   at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1089)
[info]   at java.base/java.lang.Runtime.exec(Runtime.java:681)
[info]   at java.base/java.lang.Runtime.exec(Runtime.java:530)
[info]   at svsim.Workspace.reset(Workspace.scala:54)
[info]   at chisel3.simulator.Simulator._simulate(Simulator.scala:102)
[...]

PR #4783, opened 2025-03-11
Fix for #4780

tobias1012:
### Contributor Checklist

#### Type of Improvement
- Bugfix
- Dependency update

#### Desired Merge Strategy

#### Release Notes
Fix 'rm' executable dependency on windows.

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/Workspace.scala
+++ b/svsim/src/main/scala/Workspace.scala
@@ -5,8 +5,10 @@ import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{FileVisitResult, FileVisitor}
 import java.nio.file.{FileVisitor, Files, Path, Paths}
 import java.lang.ProcessBuilder.Redirect
+import java.util.Comparator
 import scala.annotation.meta.param
 import scala.jdk.CollectionConverters._
+import scala.sys.SystemProperties
 
 case class ModuleInfo(name: String, ports: Seq[ModuleInfo.Port]) {
   private[svsim] val instanceName = "dut"
@@ -28,7 +30,7 @@ final class Workspace(
 ) {
 
   val absolutePath =
-    if (path.startsWith("/"))
+    if (Paths.get(path).isAbsolute())
       path
     else
       s"${System.getProperty("user.dir")}/$path"
@@ -50,7 +52,15 @@ final class Workspace(
   def reset() = {
     _moduleInfo = None
 
-    val rm = Runtime.getRuntime().exec(Array("rm", "-rf", absolutePath)).waitFor()
+    // Create a path type object from the absolute path string.
+    val absolutePathObject = Paths.get(absolutePath)
+    if (Files.exists(absolutePathObject)) {
+      Files
+        .walk(absolutePathObject)
+        .sorted(Comparator.reverseOrder())
+        .forEach(Files.delete)
+    }
+
     val pathsToCreate = Seq(
       supportArtifactsPath,
       primarySourcesPath,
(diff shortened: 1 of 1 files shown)


svsim/src/main/scala/Workspace.scala
```diff
@@ -28,7 +30,7 @@ final class Workspace(
 ) {
 
   val absolutePath =
-    if (path.startsWith("/"))
+    if (Paths.get(path).isAbsolute())
```
seldridge, line 33: Nice. This is better than getting a separator and using that to check if absolute. 👍

svsim/src/main/scala/Workspace.scala
```diff
@@ -392,7 +402,13 @@ final class Workspace(
       //format: off
       // For this debug flow, we rebuild the simulation from scratch every time, to avoid issues if the simulation was originally compiled in a different environment, like using SiFive's `wake`.
       l("clean:")
-      l("\tls . | grep -v Makefile | grep -v execution-script.txt | xargs rm -rf")
+      // Add check if OS is windows, since the command syntax is different
+      if (System.getProperty("os.name").toLowerCase.contains("win")) {
+        l("\tfor /f \"delims=\" %i in ('dir /b /a-d ^| findstr /v Makefile ^| findstr /v execution-script.txt') do del \"%i\"")
+        l("\tfor /d %i in (*) do rmdir /s /q \"%i\"")
+      } else {
```
seldridge, line 409: Is this necessary if running under Windows Subsystem for Linux?

seldridge (commented):
Nice! Once question about how you're running this.

Second question: since we don't have CI, how have you tested this? I'm mostly just asking that you have tested it until we have Windows CI setup.

svsim/src/main/scala/Workspace.scala
tobias1012: This should not be necessary for WSL.
I am actually a little unsure for plain windows, as windows 10/11 does provide 'rm' alias for powershell (not cmd)
This seems to be the most portable solution.

tobias1012:
**Hi, thanks for the quick review!**

Good question—I meant to leave a note about this but forgot.

My testing was done manually with primarily ProcMon to ensure files were created.

It is not entirely clear to me When the reset function from Workspace is called. so I haven't been able to confirm the right files are deleted, on any system windows or Linux. Only that I get further in the pipeline  

I'm running everything on Windows 10 using the oss-cad-suite from YosysHQ. Currently, I'm testing by importing it into a local project. However, I also wanted to run the mill tests, but there are quite a few challenges in getting both those and the simulator working properly.

While I am confident this PR does fix the issue I originally reported, it also revealed a new bug on Windows that needs further investigation. More work is required to ensure everything functions correctly on this platform.
This bug gives the error:
`error: cannot create output directory "C:\...\chisel/build\chiselsim\ProbeSpec\Probe-force\release-reg-example\should-work-in-simulator/support-artifacts\verification?sert": Unicode-symbol does not exist`

seldridge (approved):
LGTM

tobias1012:
Yes, I found multiple places where `/` is hardcoded, I will be fixing these as I continue work on this.

svsim/src/main/scala/Workspace.scala
jackkoenig: Would the rest of the Makefile work on plain Windows? I would assume not (but I am very much assuming).

tobias1012: Hi Jack. It seems there were more issues getting this to work on Windows than expected. That being said I expect to fix this in an upcoming PR

seldridge: Also, I don't think we are actually checking the Makefile in CI at all. I use this locally and will often glancingly test it.

We probably want more CI of this, generally. I.e., a flow that builds a test _and then replays the test from the Makefile_. This is kind of tedious to do with Scalatest, but I think doable. This might be easier with a slightly different approach for some tests that was more like command line testing (think LLVM Integrated Tester / lit).

PR #4784, opened 2025-03-11
[ChiselSim] Factor reset procedure out of SimulatorAPI and into stimulus

seldridge:
Factor the reset initialization procedure out of `SimulatorAPI.simulate` and into a separate, re-usable stimulus.

#### Release Notes

Add `stimulus.ResetProcedure` that encapsulates a simple circuit initialization procedure. This is customizable for non-`Module`s by specifying a clock and reset.

### DIFF
--- a/src/main/scala/chisel3/simulator/stimulus/ResetProcedure.scala
+++ b/src/main/scala/chisel3/simulator/stimulus/ResetProcedure.scala
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator.stimulus
+
+import chisel3.{Clock, Module, RawModule, Reset}
+import chisel3.simulator.AnySimulatedModule
+
+/** Stimulus that will apply a standard reset procedure to a Chisel circuit.
+  *
+  * The reset procedure is as follows:
+  *
+  *     time 0:     bring everything up using simulator settings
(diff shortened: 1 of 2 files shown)


src/main/scala/chisel3/simulator/stimulus/ResetProcedure.scala
```diff
@@ -0,0 +1,113 @@
...
+  /** Return reset stimulus for a [[Module]]. */
+  def module[A <: Module](additionalResetCycles: Int = 0): ResetProcedure[A] = new ResetProcedure[A] {
+
+    override protected final val _additionalResetCycles = additionalResetCycles
+
+    override protected final val _getClock = _.clock
+
+    override protected final val _getReset = _.reset
+
+  }
```
jackkoenig, line 88: I think this could just be:
```scala
def module[A <: Module](additionalResetCycles: Int = 0): ResetProcedure[A] =
  any(additionalResetCycles, _.clock, _.reset)
```

seldridge: Nice catch. I'll land this and co-author you. 👍

seldridge: Landed in: 5a17cc506

PR #4786, opened 2025-03-11
Fix SInt literals to reject too small of widths

jackkoenig:
Found by @tymcauley https://github.com/ucb-bar/fixedpoint/pull/11#issuecomment-2715634001

This bug was introduced 2 years ago when we added support for zero-width literals: https://github.com/chipsalliance/chisel/pull/2932

The problem is that `BigInt.bitlength` excludes the sign bit, so we need to add 1 to it for SInts (which is what we originally did), BUT we need special case support for zero-width SInts (only when specified).

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

For users attempting to reinterpret raw 2s complement values as SInts, use `.U.asSInt`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -171,7 +171,9 @@ private[chisel3] object ir {
       val unsigned = if (n < 0) (BigInt(1) << width.get) + n else n
       s"asSInt(${ULit(unsigned, width).name})"
     }
-    def minWidth: Int = (if (w.known) 0 else 1) + n.bitLength
+
+    // Special case for 0 which can be specified to zero-width (but defaults to 1 bit).
+    def minWidth: Int = if (n == 0) (if (w.known) 0 else 1) else 1 + n.bitLength
 
     def cloneWithWidth(newWidth: Width): this.type = {
       SLit(n, newWidth).asInstanceOf[this.type]
--- a/core/src/main/scala/chisel3/internal/firrtl/IR.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/IR.scala
@@ -173,7 +173,7 @@ private[chisel3] object ir {
     }
 
     // Special case for 0 which can be specified to zero-width (but defaults to 1 bit).
-    def minWidth: Int = if (n == 0) (if (w.known) 0 else 1) else 1 + n.bitLength
+    def minWidth: Int = if (n == 0 && w.known) 0 else 1 + n.bitLength
 
     def cloneWithWidth(newWidth: Width): this.type = {
       SLit(n, newWidth).asInstanceOf[this.type]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -171,7 +171,9 @@ private[chisel3] object ir {
       val unsigned = if (n < 0) (BigInt(1) << width.get) + n else n
       s"asSInt(${ULit(unsigned, width).name})"
     }
-    def minWidth: Int = (if (w.known) 0 else 1) + n.bitLength
+
+    // Special case for 0 which can be specified to zero-width (but defaults to 1 bit).
+    def minWidth: Int = if (n == 0) (if (w.known) 0 else 1) else 1 + n.bitLength
```
seldridge, line 176: I would only consider the following change if you think it better calls out the truly special case. If `!w.known`, then minWidth behaves like anything else: 

```suggestion
    // Special case for 0 which can be specified to zero-width (but defaults to 1 bit).
    def minWidth: Int = if (n == 0 && w.known) 0 else 1 + n.bitLength
```

Only for consideration.

src/test/scala-2/chiselTests/SIntOps.scala
```diff
@@ -284,4 +284,18 @@ class SIntOpsSpec extends AnyPropSpec with Matchers with ShiftRightWidthBehavior
...
+    0.S(0.W).getWidth should be(0)
+    -1.S.getWidth should be(1)
+    1.S.getWidth should be(2)
+    // The real check.
+    -2.S.getWidth should be(2)
+    an[IllegalArgumentException] shouldBe thrownBy(-2.S(1.W))
+    0xde.S.getWidth should be(9)
+    an[IllegalArgumentException] shouldBe thrownBy(0xde.S(8.W))
+    an[IllegalArgumentException] shouldBe thrownBy(0xde.S(4.W))
+  }
```
seldridge, line 300: These tests look right! I haven't thought through the other ramifications of this change, other than that the existing tests in the repo all pass.

seldridge (approved):

core/src/main/scala/chisel3/internal/firrtl/IR.scala
jackkoenig: I do actually think that's better, thanks! I didn't love the nested if.

src/test/scala-2/chiselTests/SIntOps.scala
jackkoenig: Yeah I'm confident it's correct. Note that even before this change `0xde.S.getWidth == 9`, it's just that we were accepting 1-bit less than the correct answer if the user specified it which would then coincidentally do the right thing in _most_ circumstances due to the FIRRTL we emitted actually having the correct width semantics. Thus things would work as long as you didn't introspect on the width and do things with it.

tymcauley:
Awesome, thanks for the super-fast fix @jackkoenig!

PR #4787, opened 2025-03-12
[svsim] Error if waves enabled in unsupported sim

seldridge:
Error, in Verilog, if an svsim simulator tries to enable waves in a simulation that wasn't compiled to support it.

#### Release Notes

Change svsim to error if a controller tries to enable waves in a simulation which wasn't compiled to support them.


jackkoenig:
Neat!

seldridge:
I included a test, but only for Verilator. I was not able to make this work cleanly for both Verilator and VCS testing due to the architecture of `svsim`. The Scala compiler gets confused around the type of the backend settings and I wasn't able to massage it to make it work. I would like to punt this to when I can rework `svsim` to use a more standard Scala software architecture.

jackkoenig (approved):
LGTM!

PR #4792, opened 2025-03-13
[svsim] Use SNPS-specific license in VCS

### LINKED ISSUE #4785: LM_LICENSE_FILE environment variable required for ChiselSim using VCS
**Type of issue**: Bug Report

Run any ChiselSim test using `chisel3.simulator.HasSimulator.simulators.vcs` as the simulator backend, but don't define the `LM_LICENSE_FILE` environment variable.

This throws an exception from here: https://github.com/chipsalliance/chisel/blob/a7effb7da1db69adeacb1853a2e74369a4e48609/src/main/scala/chisel3/simulator/HasSimulator.scala#L64-L68

The code searching for this environment variable comes from here: https://github.com/chipsalliance/chisel/blob/a7effb7da1db69adeacb1853a2e74369a4e48609/svsim/src/main/scala/vcs/Backend.scala#L73

ChiselSim shouldn't require this particular environment variable for VCS simulations.

version: `7.0.0-M2+529-5f2e70c3-SNAPSHOT`
OS: RHEL 8

**Other Information**

It appears that if `SNPSLMD_LICENSE_FILE` is defined, VCS will use that and ignore `LM_LICENSE_FILE`.

That being said, does ChiselSim (or rather svsim) need to be prescriptive about this? What's the need for svsim to grab this environment variable and re-export it during the simulation?

This was discussed here: https://github.com/chipsalliance/chisel/pull/3121#discussion_r1144281412

Enable more uses to use the VCS backend for ChiselSim.

seldridge:
Fix the logic for grabbing environment variables for the replay Makefile when running VCS tests.  The VCS manual says that you grab the Synopsys-specific license file (SNPSLMD_LICENSE_FILE) first and only use the generic license file (LM_LICENSE_FILE) if that fails.  The existing code was wrong in that it only grabbed the latter.  There is no requirement that the latter is available and that is only a fallback if the former doesn't exist.

Also, add some lightweight infrastructure to cleanup the representation of VCS license files.

I tested this internally and all the tests passed.  (I had to change the test to queue for licenses, though...)

Fixes #4785.

CC: @tymcauley 

#### Release Notes

In svsim, use Synopsys-specific license if available. Otherwise, fallback to the default, generic license file. This is used for generating the rerun Makefile. Previously, this would always use the generic license file which is not _required_ that it must be set. (This aligns svsim with the VCS manual on what the expected environment should look like.)

### DIFF
--- a/src/main/scala/chisel3/simulator/HasSimulator.scala
+++ b/src/main/scala/chisel3/simulator/HasSimulator.scala
@@ -61,11 +61,7 @@ object HasSimulator {
     ): HasSimulator = new HasSimulator {
       override def getSimulator(implicit testingDirectory: HasTestingDirectory): Simulator[svsim.vcs.Backend] =
         new Simulator[svsim.vcs.Backend] {
-          override val backend = svsim.vcs.Backend.initializeFromProcessEnvironment().getOrElse {
-            throw new Exception(
-              "Unable to load VCS from the environment. (Did you forget to set VCS_HOME and LM_LICENSE_FILE?)"
-            )
-          }
+          override val backend = svsim.vcs.Backend.initializeFromProcessEnvironment()
           override val tag = "vcs"
           override val commonCompilationSettings = compilationSettings
           override val backendSpecificCompilationSettings = vcsSettings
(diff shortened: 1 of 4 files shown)


tymcauley:
Looks great to me, thanks @seldridge!

PR #4793, opened 2025-03-14
[svsim] Add coverage collection to VCS backend

seldridge:
Add VCS backend options to enable coverage collection.  The options for this are taken directly from our internal flow.  I don't think that all these options need to be turned on (e.g., @girishpai indicated that `USE_UNR_ONLY_CONSTRAINTS` isn't load bearing).  However, I'd rather migrate things wholesale and then refactor later.

This includes a very simple test that a coverage database directory is created.  This database is not created if coverage is turned off.

CC: @andrew-gouldey-sifive

#### Release Notes

Add options to svsim VCS backend to allow for coverage collection.

### DIFF
--- a/svsim/src/main/scala/vcs/Backend.scala
+++ b/svsim/src/main/scala/vcs/Backend.scala
@@ -154,12 +154,36 @@ object Backend {
 
   }
 
+  /** Settings for controlling VCS toggle coverage.
+    *
+    * These options map to the `-cm_tgl` option.  Consult the Synopsys VCS user
+    * guide for documentation.
+    */
+  final case class ToggleCoverageSettings(
+    assign:              Boolean = false,
+    portsonly:           Boolean = false,
+    fullintf:            Boolean = false,
+    mda:                 Boolean = false,
+    count:               Boolean = false,
+    structarr:           Boolean = false,
+    modportarr:          Boolean = false,
+    union_excl:          Boolean = false,
+    union_adv:           Boolean = false,
+    unencrypted_signals: Boolean = false,
+    old:                 Boolean = false
+  ) extends PlusSeparated {
+
+    override def name = "cm_tgl"
+
+  }
+
   case class CompilationSettings(
     xProp:                       Option[CompilationSettings.XProp] = None,
     randomlyInitializeRegisters: Boolean = false,
     traceSettings:               CompilationSettings.TraceSettings = CompilationSettings.TraceSettings(),
     simulationSettings:          SimulationSettings = SimulationSettings(),
     coverageSettings:            CoverageSettings = CoverageSettings(),
+    toggleCoverageSettings:      ToggleCoverageSettings = ToggleCoverageSettings(),
     licenceExpireWarningTimeout: Option[Int] = None,
     archOverride:                Option[String] = None,
     waitForLicenseIfUnavailable: Boolean = false
--- a/svsim/src/main/scala/vcs/Backend.scala
+++ b/svsim/src/main/scala/vcs/Backend.scala
@@ -177,6 +177,20 @@ object Backend {
 
   }
 
+  /** Settings for controlling VCS branch coverage.
+    *
+    * These options map to the `-cm_branch` option.  Consult the Synopsys VCS
+    * user guide for documentation of this option.
+    */
+  final case class BranchCoverageSettings(
+    values:               Boolean = false,
+    ignoreMissingDefault: Boolean = false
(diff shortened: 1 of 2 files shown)


svsim/src/main/scala/vcs/Backend.scala
```diff
@@ -96,11 +96,60 @@ object Backend {
...
+  // Note: This case class is being clever and re-using the names of the
+  // elements as the names of the options.  This makes these very terse, but
+  // they then match, exactly, the documentation in the VCS manual.
+  final case class CoverageSettings(
+    line:   Boolean = false,
+    cond:   Boolean = false,
+    fsm:    Boolean = false,
+    tgl:    Boolean = false,
+    branch: Boolean = false,
+    assert: Boolean = false
```
jackkoenig, line 120: Given in all likelihood there will be other options added, I would not make this a case class because then adding any argument breaks source (and binary) compatibility. Using a private case class is fine but I'd make the public APIs simpler.

svsim/src/main/scala/vcs/Backend.scala
```diff
@@ -96,11 +96,60 @@ object Backend {
...
+  ) {
+    def compileFlags: Seq[String] = productElementNames.zip(productIterator).flatMap {
+      case (_, false)   => None
+      case (name, true) => Some(name)
+    } match {
+      case coverageTypes if !coverageTypes.isEmpty =>
```
jackkoenig, line 126: It's a bit nitty but I'd just assign to a val named `coverageTypes` and then have an `if` rather than matching.

svsim/src/main/scala/vcs/Backend.scala
```diff
@@ -96,11 +96,60 @@ object Backend {
...
+        // if you want all these options this should be handled with an omnibus
+        // setting.
+        (Seq(
+          "-cm"
+        ) :+ coverageTypes.mkString("+") :+
+          "+define+FIRRTL_BEFORE_INITIAL=// VCS coverage off" :+
+          "+define+FIRRTL_AFTER_INITIAL=// VCS coverage on" :+
+          "+define+USE_UNR_ONLY_CONSTRAINTS") ++ Option.when(line || tgl)("-cm_seqnoconst") ++
+          Option.when(tgl)(Seq("-cm_tgl", "portsonly")).toSeq.flatten ++
+          Option.when(branch)(Seq("-cm_branch", "ignoreMissingDefault")).toSeq.flatten
```
jackkoenig, line 142: This confuses me a bit, so if `tgl` is `true`, we you get both: `-cm +tgl` and `-cm_cgl portsonly`?

seldridge: It confused me, too... However, this is the set of internal options.

How about this? I'll break these out into individual options and then bridge them internally.

seldridge: I backed this out and the options now are transparently mapped to their VCS options in https://github.com/chipsalliance/chisel/pull/4793/commits/9559742ab5bd4042c3a75b36ba692458a795a3cc.

A consideration is that we generally do not expose options this way for Chisel/firtool. Those, we happily make them `Array[String]` and let the user get it correct. With svsim, we already started down the path of type-safe options. Given the large amount of VCS options, we can either add these on demand as needed, or we could switch to the `Array[String]` approach for backend options.

There is some benefit to the `Array[String]` approach in that backends no longer have different option types. However, it's more stringy.

svsim/src/main/scala/vcs/Backend.scala
seldridge: Fixed in: https://github.com/chipsalliance/chisel/pull/4793/commits/a8b4257e1297c647a0539ab8156513a9f2fef997

svsim/src/main/scala/vcs/Backend.scala
jackkoenig: I honestly would prefer less string-y arguments for Chisel and firtool, but am certainly not going to let the perfect be the enemy of the good on doing that for VCS coverage options and just leaving the Chisel and firtool options as `Array[String]` for the time being.

One tricky bit about structured types is that you probably want some form of parsing into your type which using `Array[String]` allows you to gloss over (this was a nice thing about how we/you did options annotations), but again, I still think this is good as is.

jackkoenig (approved):
LGTM!

seldridge:
I will do a rebase and then land the three logical commits here individually.

PR #4796, opened 2025-03-14
[svsim][chiselsim] Add plusarg support

seldridge:
Add the ability to set `$value$plusarg` and `$test$plusargs` directly from
ChiselSim simulate APIs.  These can be specified using the lower level
options to the simulator or via the implicit common/backend-specific
options.  However, this new API is the intended way that these should be
set by the user going forward.

Add the ability to pass plusargs to svsim simulations.

Note: there is some ugliness here in that simulation options are passed to functions that are doing "compilation".  This is a software architecture problem in svsim as evidenced by backend compilation options including runtime options (for VCS).  This all needs to be cleaned up separately.

#### Release Notes

Add support for Verilog `$value$plusargs` and `$test$plusargs` to svsim and ChiselSim.

### DIFF
--- a/svsim/src/main/scala/Backend.scala
+++ b/svsim/src/main/scala/Backend.scala
@@ -13,6 +13,38 @@ object CommonSettingsModifications {
 
 // -- Compilation Settings
 
+/** Backend-independent simulation runtime settings
+  *
+  * @note Use [[CommonSimulationSettings$]] methods to create objects of this
+  * class.
+  */
+final class CommonSimulationSettings private[svsim] (
+  val plusArgs: Seq[PlusArg]
+) {
+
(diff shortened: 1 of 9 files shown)


svsim/src/main/scala/Backend.scala
```diff
@@ -77,6 +110,15 @@ object CommonCompilationSettings {
   }
 }
 
+final class PlusArg(
```
jackkoenig, line 113: Could use a small ScalaDoc

jackkoenig (approved):
LGTM, one minor suggestion

svsim/src/main/scala/Backend.scala
seldridge: Fixed in the rebase.

PR #4797, opened 2025-03-15
[chiselsim] Add CLI FSDB, VPD support, enable waves (VCD, too)

seldridge:
Change the existing VCD waveform CLI API to _also_ enable waves. There was enough infrastructure there to do this and this is actually a way better API.

Add FSDB and VPD CLI emission. These require VCS and will throw an `IllegalArgumentException` if you try to use them with Verilator.

#### Release Notes

- Add ChiselSim traits which can be used to add Scalatest CLI support for compiling in FSDB and VPD waveform support into a test.
- Modify existing Scalatest CLI traits to both compile _and enable_ waves. These can be accessed with `-DemitVcd=1`, `-DemitVpd=1`, and `-DemitFsdb=1`.


jackkoenig (approved):
LGTM!

PR #4799, opened 2025-03-15
Replace autoNameRecursively with chisel3.withName

jackkoenig:
`chisel3.internal.plugin.autoNameRecursively` has always felt like a bit of a mouthful to me, especially in stack traces as it is _everywhere_ since it is inserted by the naming plugin. So in celebration of Chisel 7.0, I think it's time to simply the name to just `chisel3.withName`. This API also is pretty reasonable for users to go ahead and use, and I suspect in the near future we should start encouraging (or requiring?) it over `.suggestName`--but that is potentially more controversial so I'll leave that to a future PR (that will also need docs).

I've left `chisel3.internal.plugin.autoNameRecursively` around but deprecated so that people won't get weird linkage errors if they use an older version of the plugin--but they will get lots of deprecation warnings which should encourage them to bump the version of the plugin to match their version of Chisel.

I micro-optimized the arguments to the `Product` version of `withName` (note that it's not called `withNameProduct` since we can overload here). Rather than `List[Option[String]]` which requires creating lots of little objects to call the function, I made if `varargs` of `String` where empty String is the equivalent of `None` so that it's implemented with an Array of String. This probably doesn't matter much in practice but will save some object allocations.

### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

- Squash

#### Release Notes

Also use varargs of String instead of List[Option[String]] for the
Product version of withName.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/package.scala
+++ b/core/src/main/scala/chisel3/package.scala
@@ -442,4 +442,55 @@ package object chisel3 {
     private[chisel3] def apply(t: SramTarget): HasTarget = Impl(t)
 
   }
+
+  private def _withNameImpl[T](prevId: Long, name: String, nameMe: T): T = {
+    chisel3.internal.Builder.nameRecursively(
+      name,
+      nameMe,
+      (id: chisel3.internal.HasId, n: String) => {
+        // Name override only if result was created in this scope
+        if (id._id > prevId) {
+          id.forceAutoSeed(n)
+        }
+      }
+    )
+    nameMe
+  }
+
+  /** Use Chisel's naming algorithm to name the returned value
+    *
+    * This will name typical "nameable" things like Chisel [[Data]], but only if the object is created in the thunk.
+    * @param name The name to use
+    * @param nameMe A thunk that returns the thing to be named
+    * @tparam T The type of the thing to be named
+    * @return The thing, possibly now named
+    */
+  def withName[T](name: String)(nameMe: => T): T = {
+    // The _id of the most recently constructed HasId
+    val prevId = Builder.idGen.value
+    val result = nameMe
+    _withNameImpl(prevId, name, result)
+  }
+
+  /** Use Chisel's naming algorithm to name values within the returned Product value
+    *
+    * This will name typical "nameable" things like Chisel [[Data]], but only for objects crated in the thunk.
+    * @param names The names to use corresponding to interesting fields of the Product, empty Strings means no name
+    * @param nameMe The [[scala.Product]] to be named
+    * @tparam T The type of the thing to be named
+    * @return The thing, with members possibly named
+    */
+  def withNames[T <: Product](names: String*)(nameMe: => T): T = {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/chisel3/package.scala
```diff
@@ -442,4 +442,51 @@ package object chisel3 {
...
+
+  /** Use Chisel's naming algorithm to name values within the returned Product value
+    *
+    * This will name typical "nameable" things like Chisel [[Data]], but only for objects crated in the thunk.
+    * @param names The names to use corresponding to interesting fields of the Product, empty Strings means no name
+    * @param nameMe The [[scala.Product]] to be named
+    * @tparam T The type of the thing to be named
+    * @return The thing, with members possibly named
+    */
+  def withName[T <: Product](names: String*)(nameMe: => T): T = {
```
jackkoenig, line 483: Or maybe this should be `withNames`? Does it matter?

core/src/main/scala/chisel3/package.scala
```diff
@@ -442,4 +442,51 @@ package object chisel3 {
...
+    // The _id of the most recently constructed HasId
+    val prevId = Builder.idGen.value
+    val result = nameMe
+    for ((name, t) <- names.iterator.zip(result.productIterator) if name.nonEmpty) {
+      _withNameImpl(prevId, name, t)
+    }
```
seldridge, line 489: Should this check that the length of names and product match?

core/src/main/scala/chisel3/package.scala
seldridge: `withNames` seems better only because the parameter is "names". I can see it going either way.

core/src/main/scala/chisel3/package.scala
seldridge: It looks like it didn't check this before.

seldridge (approved):
LGTM

The original behavior of naming a product _not_ checking that the lengths matched isn't good and that is then carried forward here. That seems fine. We may want to change it in the future. (Do you ever want to name _part_ of a `Product` or to pass _more_ names and have it just ignore the later ones?)

core/src/main/scala/chisel3/package.scala
jackkoenig: Good point, I'll add a check.

core/src/main/scala/chisel3/package.scala
```diff
@@ -442,4 +442,55 @@ package object chisel3 {
...
+    * @return The thing, with members possibly named
+    */
+  def withNames[T <: Product](names: String*)(nameMe: => T): T = {
+    // The _id of the most recently constructed HasId
+    val prevId = Builder.idGen.value
+    val result = nameMe
+    require(
+      names.size == result.productArity,
+      s"Number of names must match number of fields in Product, got ${names.size} and ${result.productArity}."
+    )
```
seldridge, line 490: 👍

PR #4808, opened 2025-03-17
[chiselsim] Chisel/firtool update type classes

seldridge:
Add more type classes that can be used to modify the Chisel and `firtool` options of ChiselSim's `SimulatorAPI` methods.

While the number of type classes here is getting comical, these are being added to fill two remaining gaps.  We need the ability to control, via command line arguments, certain options.  With Scalatest, the only way to do this is via `-D` options.  However, the way to use `-D` options is via the `HasCliOptions` API which is implemented via this type class pattern.

#### Release Notes

Provide infrastructure for allowing modifications to Chisel and `firtool` command line arguments when running ChiselSim tests, e.g., via the Scalatest command line.


jackkoenig (approved):
LGTM

PR #4815, opened 2025-03-19
[chiselsim] Add subdirectory to SimulatorAPI fns

seldridge:
Add a subdirectory parameter to `simulate` and `simulateRaw` methods. This is intended for a bit of a corner case where a user wants to run simulate multiple times within a single test or they want to customize the test directory based on Cli arguments.

#### Release Notes

Add `subdirectory` parameters to ChiselSim `simulate` and `simulateRaw` APIs. These complement the existing `HasTestingDirectory.withSubdirectory` API which is slightly cumbersome to use in Scala 2 due to implicit resolution rules.


jackkoenig (approved):
LGTM!

PR #4821, opened 2025-03-21
Add support for modifiers to format specifiers in printing

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

This includes modifiers defined in the SystemVerilog spec like minimum width (e.g. `%0d`) and set width (e.g. `%8d`). It does **not** include non-standard extensions like left-justification and zero-padding for decimal.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -4,6 +4,7 @@ package chisel3
 
 import chisel3.experimental.SourceInfo
 import chisel3.internal.firrtl.ir.Component
+import chisel3.FirrtlFormat.FormatWidth
 
 import scala.collection.mutable
 
@@ -179,9 +180,11 @@ case class PString(str: String) extends Printable {
 
 /** Superclass for Firrtl format specifiers for Bits */
 sealed abstract class FirrtlFormat(private[chisel3] val specifier: Char) extends Printable {
-  def bits: Bits
+  def bits:   Bits
+  def unpack: (String, Bits)
   def unpack(ctx: Component)(implicit info: SourceInfo): (String, Iterable[String]) = {
-    (s"%$specifier", List(bits.ref.fullName(ctx)))
+    val (str, bits) = unpack
+    (str, List(bits.ref.fullName(ctx)))
   }
 
   def unpackArgs: Seq[Bits] = List(bits)
@@ -192,9 +195,34 @@ object FirrtlFormat {
   def unapply(x: Char): Option[Char] =
     Option(x).filter(x => legalSpecifiers contains x)
 
+  /** Width modifiers for format specifiers */
+  sealed trait FormatWidth {
+    def toFormatString: String
+  }
+  object FormatWidth {
+
+    /** Display using the minimum number of characters */
+    case object Minimum extends FormatWidth {
+      def toFormatString: String = "0"
+    }
+
+    /** Pad the display to the number of characters needed to display the maximum possible value for the signal width and formatting */
+    case object Automatic extends FormatWidth {
+      def toFormatString: String = ""
+    }
+
+    /** Display using the specified number of characters */
+    case class Fixed(value: Int) extends FormatWidth {
+      require(value != 0, s"For minimum width, use FormatWidth.Minimum, not FormatWidth.Fixed(0)")
+      require(value > 0, s"Width.Fixed must be positive, got $value")
+      def toFormatString: String = value.toString
+    }
+  }
+
   /** Helper for constructing Firrtl Formats
     * Accepts data to simplify pack
     */
+  @deprecated("Use FirrtlFormat.parse instead", "Chisel 7.0.0")
   def apply(specifier: String, data: Data): FirrtlFormat = {
     val bits = data match {
       case b: Bits => b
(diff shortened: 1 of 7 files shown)


docs/src/explanations/printing.md
```diff
@@ -69,6 +69,27 @@ printf(cf"hierarchical path = $HierarchicalName\n") // hierarchical path = <veri
...
+* Leading `-` will left-justify the value.
+* A non-negative field width will override the default Verilog sizing of the value.
+* Specifying a field width of `0` will always display the value with the minimum width (no zero nor space padding).
+* A leading `0` combined with a field width will pad the value with `0` to the field width.
+
+```scala mdoc:compile-only
+val myUInt = WireInit(UInt(64.W), 33.U)
+printf(cf"myUInt = $myUInt%d")  // myUInt =         33!
+printf(cf"myUInt = $myUInt%0d") // myUInt = 33!
+printf(cf"myUInt = $myUInt%-d") // myUInt = 33!
```
terpstra, line 89: Is the above comment right?

terpstra: Also, shouldn't you include `%4d` as an example as well?

tymcauley:
I am irrationally excited for this feature. Thank you!

core/src/main/scala/chisel3/Printable.scala
```diff
@@ -187,35 +193,47 @@ object FirrtlFormat {
...
+          case _   => throw new Exception(s"Unreachable! specifier = '$spec'")
+        }
+      case bad => throw new Exception(s"Illegal format specifier '$bad'!")
     }
   }
 }
 
 /** Format bits as Decimal */
-case class Decimal(bits: Bits) extends FirrtlFormat('d')
+case class Decimal(bits: Bits, modifier: String = "") extends FirrtlFormat('d')
```
jackkoenig, line 227: I've realized that this should use richer types, we don't want people writing things like `Decimal(foo, "-4")`, it should be something like `Decimal(foo, Aligment.Left, MinimumDisplayWidth(4))`

jackkoenig:
@seldridge and I are waffling a bit on this. New short term plan is to just have firtool start emitting `%0<n>` by default and then we can figure out how we want to handle width specifiers later. Some challenges:
* Left justification (`-%`) is not defined in the SystemVerilog spec so simulators implement it in different ways (not all will left justify hexadecimal, and even when they do, they're inconsistent on the size of the field).
  * We could just not support left justification, otherwise we probably want to specify its behavior ourselves which will require custom implementations so that the behavior is consistent across simulators
* Since we want `%0<n>` to be the default behavior (meaning no space or zero padding to the width of the maximum possible value), how should we let users specify when they do want that padding automatically based on the width of the `Bits` being displayed?
  * I'm soft pitching `*` for this, such that `printf(cf"foo = $foo%*d\n")` or `printf("foo = %*d\n", foo)` would give the "autopadding" behavior.
  * Another alternative is just to provide no pithy syntax for this but instead require the user to use the Printable types directly, e.g. `printf(cf"foo = ${Decimal(foo, width = AutoPad)}\n")`. The issue is that we added `cf"..."` in the first place because users hated having to write `printf(p"foo = ${Decimal}\n")` in the first place. There's just a delicate balance between how much do we define our own format-specifiers vs. drawing on Verilog or other languages.

terpstra:
You realize that switching to `%0d` by default is a breaking change? There are almost certainly downstream tools that rely on the fixed-with padding.

jackkoenig:
> You realize that switching to %0d by default is a breaking change? There are almost certainly downstream tools that rely on the fixed-with padding.

Discussed offline and we may reconsider, but we do think that it's better default behavior if we can swing it.

jackkoenig:
This PR is back and ready for review, although now it's just supporting width specifiers in the SystemVerilog spec. I removed support for left-justification and 0-padding of decimal formats as those are both non-standard extensions that are implemented differently by different simulators. Eventually we would like to support such things, but we want to make sure that Chisel-emitted Verilog will always behave the same way, no matter the simulator.

docs/src/explanations/printing.md
```diff
@@ -72,6 +72,24 @@ printf(cf"100$Percent\n") // 100%
 printf(cf"100%%\n") // equivalent to the above
 ```
 
+#### Format modifiers
+
+Chisel supports standard Verilog-style modifiers for `%d`, `%x`, `%b`, and `%c` between the `%` and the format specifier.
```
seldridge, line 77: This isn't supported on `%c`.

docs/src/explanations/printing.md
```diff
@@ -144,6 +162,8 @@ Chisel provides `printf` in a similar style to its C namesake. It accepts a doub
 | `%%` | literal percent |
 | `%m` | hierarchical name |
 
+`%d`, `%x`, `%b`, and `%c` support the modifiers described in the [Format modifiers](#format-modifiers) section above.
```
seldridge, line 165: `%c` doesn't support this, right?

src/test/scala-2/chiselTests/PrintableSpec.scala
```diff
@@ -323,4 +326,51 @@ class PrintableSpec extends AnyFlatSpec with Matchers with FileCheck {
...
+    FirrtlFormat.parse("%0b", x) should be(Right(Binary(x, FormatWidth.Minimum)))
+    FirrtlFormat.parse("%23d", x) should be(Right(Decimal(x, FormatWidth.Fixed(23))))
+    FirrtlFormat.parse("%23x", x) should be(Right(Hexadecimal(x, FormatWidth.Fixed(23))))
+    FirrtlFormat.parse("%23b", x) should be(Right(Binary(x, FormatWidth.Fixed(23))))
+  }
+  it should "not support illegal format specifiers" in {
+    val x = 123.U
+    an[Exception] should be thrownBy {
+      FirrtlFormat("a", x)
+    }
```
seldridge, line 364: This could be checked as this returns a `Left`.

seldridge (approved):
LGTM

tymcauley:
> > You realize that switching to %0d by default is a breaking change? There are almost certainly downstream tools that rely on the fixed-with padding.
> 
> Discussed offline and we may reconsider, but we do think that it's better default behavior if we can swing it.

How about a command-line argument to select between default-`FormatWidth.Automatic` and default-`FormatWidth.Minimum`? If you want to go with the breaking change, you could make _that_ setting choose `FormatWidth.Minimum` if unspecified, but users can always opt back into default-`FormatWidth.Automatic` if they have tooling which depends on that.

seldridge:
>>> You realize that switching to %0d by default is a breaking change? There are almost certainly downstream tools that rely on the fixed-with padding.
>>
>> Discussed offline and we may reconsider, but we do think that it's better default behavior if we can swing it.
>
> How about a command-line argument to select between default-FormatWidth.Automatic and default-FormatWidth.Minimum? If you want to go with the breaking change, you could make that setting choose FormatWidth.Minimum if unspecified, but users can always opt back into default-FormatWidth. Automatic if they have tooling which depends on that.

FWIW, I already tried this internally and broke tools that parsed the output of printfs and were brittle to this change. 

While this pains me, the ship may have sailed on this, even though Chisel has the wrong default. We will see about changing this in the future, because `%0d` is a much saner output. However, this will have to be done carefully.

docs/src/explanations/printing.md
jackkoenig: Good catch, didn't update this fully.

docs/src/explanations/printing.md
jackkoenig: Correct, good catch.

PR #4824, opened 2025-03-23
peek/poke/expect for all Data types, include Records and Vecs

kammoh:
This PR changes the "testable" implicit classes in `PeekPokeAPI` to allow for a unified `peek()`, `poke()`, and `expect` API for all data types, including Records, Vecs, Enums, and nested aggregates.

The implementation does not look very pretty, but I could not figure out a cleaner way to do it.

This is part of a series of PRs trying to integrate some `chiseltest` features, partially based on #4209.

### Contributor Checklist

#### Type of Improvement
- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?


src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+            case TestOp.Mul => a * b
+            case _          => throw new Exception("Invalid operation")
+          }
+          val expectedCmp = a.compare(b) match {
+            case -1 => dut.CmpResult.LT
+            case 0  => dut.CmpResult.EQ
+            case 1  => dut.CmpResult.GT
+          }
+
+          dut.io.out.valid.expect(true.B)
```
jackkoenig, line 126: Would it be possible to split this into a bunch of smaller tests of just individual types? Having the peeks and pokes of all types together kind of muddles what's being tested a bit.

src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+              _.b -> 2.U,
+              _.v1 -> Vec.Lit(Seq.fill(vecDim)(3.U(w.W)): _*),
+              _.v2 -> Vec.Lit(Seq.fill(vecDim)(4.U(w.W)): _*)
+            )
+          )
+          dut.io.in.valid.poke(true)
+          dut.io.op.poke(TestOp.Add)
+          dut.clock.step()
+
+          dut.io.out.bits.c.expect(5.U)
```
jackkoenig, line 190: If this is just testing that the expectation is correct, could the rest of the test be simpler?

jackkoenig (commented):
Overall I think this looks really good. Sorry for the delay on reviewing, I'm not super familiar with the PeekPokeAPI so had to look over it again so I could review 😅 

I don't want to sandbag so feel free to push back, but my only real complaint is a nit that the tests are a big big rather than being split into small pieces for like each type being tested. Would it be too much trouble to split those up?

kammoh:
Hi @jackkoenig,
Thank you for the review and your feedback!
I'll try to cleanup and reorganize the tests, maybe break it up into multiple files.
I also have some doubts about what needs to be exposed as the public API at this point and also what needs to be supported for _all_ types of `Data`. I'm also thinking about the usability and extensibility of the API for downstream libraries (e.g. FixedPoint and FloatingPoint data). 
Also please let me know if you have any thoughts on the the `expect` functions that have a message generator argument.
I'm also testing this with my own libraries and implementations to make sure everything turns out consistent and works as expected.

kammoh:
The other issue is the ability to extend for the multi-threaded version of `PeekPokeAPI`. It seems to me that the current structure makes code reuse difficult.

kammoh:
All those `.asInstanceOf[T]` is scary and ugly, but I don't know how else we can do this.
Any thoughts @jackkoenig @seldridge ?

src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
kammoh: It is also indirectly testing `poke`s.

src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
kammoh: I think the most reliable way to test peeks and pokes is to apply `poke()`s to the inputs, let the circuit do its thing, and then `peek()` at or `expect()` the outputs. I really think we need even more testing/coverage, not less.

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
```diff
@@ -127,58 +304,236 @@ trait PeekPokeAPI {
...
+        case (dat: Record, exp: Record) =>
+          new TestableRecord(dat).expect(exp, buildMsgFn _)
+        case (dat: Vec[_], exp: Vec[_]) =>
+          new TestableVec(dat).expectVec(
+            exp,
+            (obs: Data, _: Data, idx: Int) => {
+              require(obs.getClass == exp.getClass, s"Type mismatch: ${obs.getClass} != ${exp.getClass}")
+              buildMessage(obs.asInstanceOf[T], expected.asInstanceOf[T]) + s". First mismatch at index $idx."
+            }
+          )
```
jackkoenig, line 493: I'm a little confused why there is special logic for Vec here but not for Record? I see Records are checked in the methods on `TestableRecord` so I think it's fine there's no extra checks, but Vecs are tested in `TestableVec` so why are there additional checks here?

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
```diff
@@ -127,58 +304,236 @@ trait PeekPokeAPI {
...
+  implicit class TestableData[T <: Data](val data: T) extends PeekPokable[T] {
+
+    def peek(): T = {
+      data match {
+        case x: Bool     => new TestableBool(x).peek().asInstanceOf[T]
+        case x: UInt     => new TestableUInt(x).peek().asInstanceOf[T]
+        case x: SInt     => new TestableSInt(x).peek().asInstanceOf[T]
+        case x: EnumType => new TestableEnum(x).peek().asInstanceOf[T]
+        case x: Record   => new TestableRecord(x).peek().asInstanceOf[T]
+        case x: Vec[_]   => new TestableVec(x).peek().asInstanceOf[T]
```
jackkoenig, line 460: Are these the `asInstanceOfs` you're concerned about? I actually think there are fine--we know the type is correct because we have matched on the concrete runtime type, so barring a bug in `TestableBool` (or whatever for the given type), the cast will always work. The issue is just convincing the compiler that the type is correct--which it is.

src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
jackkoenig: I'm not saying have less testing or coverage, I'm just saying to do it in smaller chunks. For example, one could organize this by type and have basically a type parametered passthrough module (or have a register in there to delay a cycle), and then for each type and peek and poke method have a test that pokes the input and then checks the output.

My only real qualm is the "various" in `it("should peek and poke various data types correctly")`, I'd rather see:
```scala
it("should peek and poke Bool correctly")
it("should peek and poke UInt correctly")
it("should peek and poke ChiselEnum correctly")
... etc.
```
I'm not going to sandbag over this, I just think that would be better

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
kammoh: It's probably not needed, but I wasn't 100% sure. The type generics for the extension classes differ between `Vec` and `Record`:
```scala
class TestableRecord[T <: Record](val data: T)  // type parameter is the type itself, so methods accepting T are type-checked
```

vs
```scala
class TestableVec[T <: Data](val data: Vec[T]) extends PeekPokable[Vec[T]]  // type parameter is the element type and is gone after type erasure
```
That's the only way I could get `Vec[T]` working. Please let me know if you think there's a cleaner way.

I cleaned up the code a little bit (using a weird workaround for `Vec` in `_expect` and `_poke`), so the special cases are now tucked away.

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
kammoh: Yes, my primary concern was really `Vec`, but I agree that all should be fine.

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
```diff
@@ -342,22 +341,15 @@ object PeekPokeAPI {
...
     def expect(expected: T, buildMessage: (T, T, String) => String, allowPartial: Boolean)(
       implicit sourceInfo: SourceInfo
     ): Unit = {
       require(DataMirror.checkTypeEquivalence(data, expected), "Type mismatch")
 
-      // FIXME: I can't understand why but _not_ getting the peeked value as a `val` beforehand results in infinite recursion
-      val peekedValue = peek()
+      // Not peeking the value beforehand or using `def` or `lazy val` results in a mysterious infinite recursion and StackOverflowError
+      // The value is not used if the expected value matches and incurs an overhead
+      // TODO: dig deeper amd see if we can avoid this
```
kammoh, line 351: @jackkoenig Any thoughts on why this happens?

src/main/scala/chisel3/simulator/PeekPokeAPI.scala
```diff
@@ -429,7 +415,12 @@ object PeekPokeAPI {
         expected.length == data.length,
         s"Vec length mismatch: Data port has ${data.length} elements while the expected value is of length ${expected.length}"
       )
+
+      // Not peeking the value beforehand or using `def` or `lazy val` results in a mysterious infinite recursion and StackOverflowError
+      // The value is not used if the expected value matches and incurs an overhead
+      // TODO: dig deeper amd see if we can avoid this
       val peekedValue = data.peek()
```
kammoh, line 422: And here

src/test/scala-2/chiselTests/simulator/PeekPokeApiSpec.scala
kammoh: Oh I see what you mean and do agree with you. I cleaned up the tests as much as I could. Please let me know if you think it still needs more cleanup.

jackkoenig (approved):
I attempted to resolve the merge conflicts, although there is an mdoc failure, feel free to push over my merge commit if you prefer. @kammoh if you can take a look at the failure, that'd be great, if not, I'll take a look when I get a chance.

PR #4825, opened 2025-03-24
[chiselsim] Add RunUntilSuccess stimulus

seldridge:
Add new stimulus, `RunUntilSuccess`, which can be used to run a simulation until a user-specified "success" port asserts.  This is added as a complementary testing approach to `RunUntilFinished` whish is needed for some environments where the test harness cannot terminate on its own or the test harness is re-used in a separate, non-ChiselSim testing context where it cannot terminate on its own.

#### Release Notes

Add `RunUntilSuccess` stimulus for ChiselSim. This works like `RunUntilFinished` except that it is checking for a user-specified "success" port to assert.

### DIFF
--- a/src/main/scala/chisel3/simulator/stimulus/RunUntilSuccess.scala
+++ b/src/main/scala/chisel3/simulator/stimulus/RunUntilSuccess.scala
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator.stimulus
+
+import chisel3.{Bool, Clock, Module}
+import chisel3.simulator.AnySimulatedModule
+import chisel3.simulator.Exceptions
+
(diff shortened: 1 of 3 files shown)


src/main/scala/chisel3/simulator/stimulus/RunUntilSuccess.scala
```diff
@@ -0,0 +1,99 @@
...
+import chisel3.simulator.Exceptions
+
+/** Stimulus that will run a simulation expecting a "success" port to assert.
+  *
+  * If the specified "success" port does not assert, then an
+  * [[Exceptions.Timeout]] will be thrown.
+  *
+  * @see [[RunUntilFinished]]
+  */
+trait RunUntilSuccess[A] extends Stimulus.Type[A] {
```
jackkoenig, line 16: Should this be sealed?

PR #4826, opened 2025-03-24
Add command line options to control elaboration of inline tests

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

- Provide control over when inline tests are elaborated
    - Add `--include-tests-module` command line argument to enable inline test elaboration for modules matching a glob
    - Add `--include-tests-name` command line argument to enable inline test elaboration for tests matching a glob
    - Add `HasTests.elaborateTests` to dynamically enable test elaboration for a module (dynamic per-test elaboration can be done by wrapping calls to `test` in a conditional block)

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -486,11 +486,12 @@ private[chisel3] class DynamicContext(
   val sourceRoots:         Seq[File],
   val defaultNamespace:    Option[Namespace],
   // Definitions from other scopes in the same elaboration, use allDefinitions below
-  val loggerOptions: LoggerOptions,
-  val definitions:   ArrayBuffer[Definition[_]],
-  val contextCache:  BuilderContextCache,
-  val layerMap:      Map[layer.Layer, layer.Layer],
-  val elaborateInlineTests: Boolean,
+  val loggerOptions:               LoggerOptions,
+  val definitions:                 ArrayBuffer[Definition[_]],
+  val contextCache:                BuilderContextCache,
+  val layerMap:                    Map[layer.Layer, layer.Layer],
+  val includeInlineTestsForModule: Seq[String],
+  val includeInlineTestsWithName:  Seq[String]
 ) {
   val importedDefinitionAnnos = annotationSeq.collect { case a: ImportDefinitionAnnotation[_] => a }
 
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -486,12 +486,11 @@ private[chisel3] class DynamicContext(
   val sourceRoots:         Seq[File],
   val defaultNamespace:    Option[Namespace],
   // Definitions from other scopes in the same elaboration, use allDefinitions below
-  val loggerOptions:               LoggerOptions,
-  val definitions:                 ArrayBuffer[Definition[_]],
-  val contextCache:                BuilderContextCache,
-  val layerMap:                    Map[layer.Layer, layer.Layer],
-  val includeInlineTestsForModule: Seq[String],
-  val includeInlineTestsWithName:  Seq[String]
+  val loggerOptions:      LoggerOptions,
+  val definitions:        ArrayBuffer[Definition[_]],
+  val contextCache:       BuilderContextCache,
+  val layerMap:           Map[layer.Layer, layer.Layer],
+  val inlineTestIncluder: InlineTestIncluder
 ) {
   val importedDefinitionAnnos = annotationSeq.collect { case a: ImportDefinitionAnnotation[_] => a }
 
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -489,7 +489,8 @@ private[chisel3] class DynamicContext(
   val loggerOptions: LoggerOptions,
   val definitions:   ArrayBuffer[Definition[_]],
   val contextCache:  BuilderContextCache,
-  val layerMap:      Map[layer.Layer, layer.Layer]
+  val layerMap:      Map[layer.Layer, layer.Layer],
+  val elaborateInlineTests: Boolean,
 ) {
   val importedDefinitionAnnos = annotationSeq.collect { case a: ImportDefinitionAnnotation[_] => a }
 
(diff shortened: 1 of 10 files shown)


src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -95,6 +95,21 @@ object TestHarnessGenerator {
...
 trait HasTests[M <: RawModule] { module: M =>
 
+  /** Whether inline tests will be elaborated as a top-level definition to the circuit. */
+  protected def elaborateTests: Boolean = true
+
+  private val builderContext = internal.Builder.captureContext()
+
+  private val thisModuleShouldElaborateTests =
+    builderContext.includeInlineTestsForModule.exists { glob =>
+      glob.replace("*", ".*").r.matches(desiredName)
```
jackkoenig, line 105: Please use `java.nio.file.PathMatcher` instead of converting globs to regex. Converting to regex has all kinds of issues (e.g. people can write a regex and it will "work"). `java.nio.file.PathMatcher` has native glob support, it was added for Paths but it just kind of works on raw Strings and it ships with the JVM so 🤷‍♀️ .

I used it for warning configuration: https://github.com/chipsalliance/chisel/blob/477900b9c72c89821f239fd800d495b077ddaf4e/core/src/main/scala/chisel3/internal/Error.scala#L160

The TLDR is you do something like `FileSystems.getDefault().getPathMatcher("glob:" + <your glob string>)` will give you a `PathMatcher` then `.matches(Paths.get(<arbitrary string>))` to check if the glob matches.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -486,10 +486,12 @@ private[chisel3] class DynamicContext(
...
-  val loggerOptions: LoggerOptions,
-  val definitions:   ArrayBuffer[Definition[_]],
-  val contextCache:  BuilderContextCache,
-  val layerMap:      Map[layer.Layer, layer.Layer]
+  val loggerOptions:               LoggerOptions,
+  val definitions:                 ArrayBuffer[Definition[_]],
+  val contextCache:                BuilderContextCache,
+  val layerMap:                    Map[layer.Layer, layer.Layer],
+  val includeInlineTestsForModule: Seq[String],
+  val includeInlineTestsWithName:  Seq[String]
```
jackkoenig, line 494:
```suggestion
  val includeInlineTestsForModule: Seq[PathMatcher],
  val includeInlineTestsWithName:  Seq[PathMatcher]
```
These should be `PathMatchers`

jackkoenig: On second thought, maybe these should be collapsed into a single type:
```scala
class InlineTestIncluder(...) {
  def includeTestsForModule(name: String): Boolean = ???
  def includeTestsWithName(name: String): Boolean = ???
}
```

The reason I'm suggesting this is a code organization suggestion. It seems to me that the default behavior should be to include no tests, but what if the user passes only `--include-tests-module=Foo` or only passes `--include-tests-name=smoke*`. It seems to me that when only 1 argument is passed, then the other should default to `*`. How to express this logic is tricky, but it may be easiest to do by encapsulating the logic in a single class that is created from `Seq[Annotation]` so that it can consume all of the annotations created from the command-line and figure out what to do.

It would also let us hide the `PathMatcher` logic because it is a little bit weird feeling but it implements globs correctly so it's good to use--but probably good to hide.

tmckay-sifive: I agree that this seems like the most sensible default behavior.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -95,6 +95,14 @@ object TestHarnessGenerator {
   */
 trait HasTests[M <: RawModule] { module: M =>
 
+  /** Whether inline tests will be elaborated as a top-level definition to the circuit. */
+  protected def elaborateTests: Boolean = true
+
+  private val builderContext = internal.Builder.captureContext()
+
+  private def shouldElaborateTest(testName: String) =
+    builderContext.inlineTestIncluder.shouldElaborateTest(module.desiredName, testName)
```
jackkoenig, line 104:
```suggestion
  private val inlineTestIncluder = internal.Builder.captureContext().inlineTestIncluder

  private def shouldElaborateTest(testName: String) =
    inlineTestIncluder.shouldElaborateTest(module.desiredName, testName)
```
Captring the entire Builder context here feels a bit extreme, I think we should just grab the test includer.

jackkoenig (approved):
LGTM with 1 tweak

seldridge (commented):
This is super cool!

core/src/main/scala/chisel3/internal/InlineTestIncluder.scala
```diff
@@ -0,0 +1,38 @@
...
+    val matcher = filesystem.getPathMatcher(s"glob:$glob")
+    matcher.matches(Paths.get(path))
+  }
+
+  def shouldElaborateTest(moduleDesiredName: String, testName: String): Boolean = {
+    val (resolvedModuleGlobs, resolvedTestNameGlobs) = (includeModuleGlobs, includeTestNameGlobs) match {
+      case x @ (Nil, Nil) => x
+      case (Nil, ts)      => (Seq("*"), ts)
+      case (ms, Nil)      => (ms, Seq("*"))
+      case x              => x
```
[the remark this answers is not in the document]
jackkoenig: Great catch this either needs to be `List` (instead of `Seq`) or match on `Seq()` for empty Seqs.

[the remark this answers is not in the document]
tmckay-sifive: Fixed--what is the problem with matching on `Nil`?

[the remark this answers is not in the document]
jackkoenig: `Seq` is the super type, `Nil` is concretely an empty List, so while `case Nil` would match `List()` or even `Seq()` (because that happens to return `Nil`), it would not match `Vector()` which `case Seq()` still would.

Although now that I check this, it doesn't seem to be true anymore?
```
$ scala-cli -S 2.13
Welcome to Scala 2.13.16 (OpenJDK 64-Bit Server VM, Java 17.0.7).
Type in expressions for evaluation. Or try :help.

scala> def func[A](xs: Seq[A]): Unit = xs match {
     |   case Nil => println(s"$xs matches Nil")
     |   case Seq() => println(s"$xs matches Seq()")
     |   case _ => println(s"$xs matches neither")
     | }
def func[A](xs: Seq[A]): Unit

scala> func(Nil)
List() matches Nil

scala> func(List())
List() matches Nil

scala> func(Seq())
List() matches Nil

scala> func(Vector())
Vector() matches Nil
```

This definitely _used_ to be the case, but it does make sense that an empty Vector should `==` an empty List...

PR #4832, opened 2025-03-26
Standardize inline testharness interface to clock, reset, finish, and success

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

- Inline testharnesses now have clock, reset, finish, and success IOs
    - Test is completed when `finish` is asserted
    - `success` is evaluated when `finish` is asserted; 1 indicates pass

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
+++ b/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
@@ -28,35 +28,34 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
 
 /** Contains traits that implement behavior common to generators for unit test testharness modules. */
 object TestHarness {
-  import chisel3.{Module => ChiselModule, RawModule => ChiselRawModule}
-
-  /** TestHarnesses for inline tests without clock and reset IOs should extend this. This
-    * trait sets the correct desiredName for the module, instantiates the DUT, and provides
-    * methods to elaborate the test.
-    *
-    *  @tparam M the type of the DUT module
-    *  @tparam R the type of the result returned by the test body
-    */
-  trait RawModule[M <: ChiselRawModule, R] extends Public { this: ChiselRawModule =>
-    def test: TestParameters[M, R]
-    override def desiredName = test.desiredTestModuleName
-    val dut = Instance(test.dutDefinition)
-    final def elaborateTest(): R = test.testBody(dut)
-  }
+  import chisel3.{Module => ChiselModule}
 
   /** TestHarnesses for inline tests should extend this. This trait sets the correct desiredName for
-   *  the module, instantiates the DUT, and provides methods to elaborate the test. By default, the
+   *  the module, instantiates the DUT, and provides methods to generate the test. By default, the
    *  reset is synchronous, but this can be changed by overriding [[resetType]].
    *
    *  @tparam M the type of the DUT module
    *  @tparam R the type of the result returned by the test body
    */
-  trait Module[M <: ChiselRawModule, R] extends RawModule[M, R] { this: ChiselModule =>
+  class Module[M <: RawModule, R](test: TestParameters[M, R]) extends ChiselModule with Public {
+    override def desiredName = test.desiredTestModuleName
+
     override def resetType = test.resetType match {
       case Some(rt @ Module.ResetType.Synchronous)  => rt
       case Some(rt @ Module.ResetType.Asynchronous) => rt
       case _                                        => Module.ResetType.Synchronous
     }
+
+    final val finish = IO(Output(Bool()))
+    final val success = IO(Output(Bool()))
+
+    // Handle the base case where a test has no result. In this case, we expect
+    // the test to end the simulation and signal pass/fail.
+    finish := false.B
+    success := true.B
+
+    protected val dut = Instance(test.dutDefinition)
+    protected val testResult = test.testBody(dut)
   }
 }
 
@@ -69,22 +68,20 @@ object TestHarness {
 trait TestHarnessGenerator[M <: RawModule, R] {
 
   /** Generate a testharness module given the test parameters. */
-  def generate(test: TestParameters[M, R]): RawModule with Public
+  def generate(test: TestParameters[M, R]): TestHarness.Module[M, R]
 }
 
 object TestHarnessGenerator {
 
-  /** The minimal implementation of a unit testharness. Has a clock input and a synchronous reset
-    *  input. Connects these to the DUT and does nothing else.
-    */
-  class UnitTestHarness[M <: RawModule](val test: TestParameters[M, Unit])
-      extends Module
-      with TestHarness.Module[M, Unit] {
-    elaborateTest()
-  }
+  /** Factory for a TestHarnessGenerator typeclass. */
+  def apply[M <: RawModule, R](gen: TestParameters[M, R] => TestHarness.Module[M, R]) =
+    new TestHarnessGenerator[M, R] {
+      override def generate(test: TestParameters[M, R]) = gen(test)
+    }
 
-  implicit def unitTestHarness[M <: RawModule]: TestHarnessGenerator[M, Unit] = new TestHarnessGenerator[M, Unit] {
-    override def generate(test: TestParameters[M, Unit]) = new UnitTestHarness(test)
+  /** Typeclass for base [[TestHarness.Module]] generator. */
+  implicit def baseTestHarnessGenerator[M <: RawModule]: TestHarnessGenerator[M, Unit] = {
+    TestHarnessGenerator(new TestHarness.Module[M, Unit](_))
   }
 }
 
--- a/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
+++ b/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
@@ -11,7 +11,7 @@ import chisel3.experimental.hierarchy.{Definition, Instance}
   *  @tparam M the type of the DUT module
   *  @tparam R the type of the result returned by the test body
   */
-class TestParameters[M <: RawModule, R] private[inlinetest] (
+final class TestParameters[M <: RawModule, R] private[inlinetest] (
   /** The [[desiredName]] of the DUT module. */
   val dutName: String,
   /** The user-provided name of the test. */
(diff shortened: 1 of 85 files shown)


src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -28,35 +28,34 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
   /** TestHarnesses for inline tests should extend this. This trait sets the correct desiredName for
-   *  the module, instantiates the DUT, and provides methods to elaborate the test. By default, the
+   *  the module, instantiates the DUT, and provides methods to generate the test. By default, the
    *  reset is synchronous, but this can be changed by overriding [[resetType]].
    *
    *  @tparam M the type of the DUT module
    *  @tparam R the type of the result returned by the test body
    */
-  trait Module[M <: ChiselRawModule, R] extends RawModule[M, R] { this: ChiselModule =>
+  class Module[M <: RawModule, R](test: TestParameters[M, R]) extends ChiselModule with Public {
```
jackkoenig, line 40: Is it really worth it to alias the class name `Module` vs. just calling this a `TestHarness`? (this would imply also moving it out of `object TestHarness`).

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -28,35 +28,34 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
+    override def desiredName = test.desiredTestModuleName
+
```
jackkoenig, line 42: Also I see this is just a `class`, not an `abstract class`, do you expect users to instantiate it directly or should they always be extending it?

jackkoenig:
Marked as `No Release Notes` instead of `API Modification` because there is not yet a release with this feature.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: Agreed, this organization made more sense when we also has `TestHarness.RawModule`

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: I made it just `class` so that chisel could provide a default generator that instantiates that base testharness module. To do this `inlinetest` needs to be able to instantiate it.

That said, I would want to discourage users from instantiating it. So, I can make it `abstract` and just instantiate an anonymous concrete class.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
-  trait RawModule[M <: ChiselRawModule, R] extends Public { this: ChiselRawModule =>
-    def test: TestParameters[M, R]
-    override def desiredName = test.desiredTestModuleName
-    val dut = Instance(test.dutDefinition)
-    final def elaborateTest(): R = test.testBody(dut)
-  }
-
-  /** TestHarnesses for inline tests should extend this. This trait sets the correct desiredName for
-   *  the module, instantiates the DUT, and provides methods to elaborate the test. By default, the
+/** TestHarnesses for inline tests should extend this. This trait sets the correct desiredName for
```
jackkoenig, line 29:
```suggestion
/** TestHarnesses for inline tests should extend this. This abstract class sets the correct desiredName for
```

jackkoenig (approved):
LGTM! Merging https://github.com/chipsalliance/chisel/pull/4834 caused a conflict.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
+  final val success = IO(Output(Bool()))
+
+  // Handle the base case where a test has no result. In this case, we expect
+  // the test to end the simulation and signal pass/fail.
+  finish := false.B
+  success := true.B
+
+  protected val dut = Instance(test.dutDefinition)
+  protected val testResult = test.testBody(dut)
 }
```
seldridge, line 55: This needs to prevent further IO creation. There are two ways to do that:

1. Use `FixedIORawModule`:

``` scala
final class TestHarnessIO extends Bundle {
  final val clock = IO(Input(Clock()))
  final val reset = IO(Input(Bool())) // Or Reset?
  final val finish = IO(Output(Bool()))
  final val success = IO(Output(Bool()))
}

class TestHarness(params...) extends FixedIORawModule(new TestHarnessIO) {
  ...
}
```

2. Use the somewhat obscure `endIOCreation()` API ad the end of this module to prevent additional IO from being created. (`FixedIORawModule` already does this.)

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
    *  @tparam R the type of the result returned by the test body
    */
-  trait Module[M <: ChiselRawModule, R] extends RawModule[M, R] { this: ChiselModule =>
-    override def resetType = test.resetType match {
-      case Some(rt @ Module.ResetType.Synchronous)  => rt
-      case Some(rt @ Module.ResetType.Asynchronous) => rt
-      case _                                        => Module.ResetType.Synchronous
-    }
+abstract class TestHarness[M <: RawModule, R](test: TestParameters[M, R]) extends Module with Public {
+  override def desiredName = test.desiredTestModuleName
```
seldridge, line 37:
```suggestion
  override final def desiredName = test.desiredTestModuleName
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
+
+  override def resetType = test.resetType match {
+    case Some(rt @ Module.ResetType.Synchronous)  => rt
+    case Some(rt @ Module.ResetType.Asynchronous) => rt
+    case _                                        => Module.ResetType.Synchronous
```
seldridge, line 42:
```suggestion
  override final def resetType = test.resetType match {
    case Some(rt @ Module.ResetType.Synchronous)  => rt
    case Some(rt @ Module.ResetType.Asynchronous) => rt
    case _                                        => Module.ResetType.Synchronous
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
+  final val finish = IO(Output(Bool()))
+  final val success = IO(Output(Bool()))
+
+  // Handle the base case where a test has no result. In this case, we expect
+  // the test to end the simulation and signal pass/fail.
+  finish := false.B
+  success := true.B
+
+  protected val dut = Instance(test.dutDefinition)
+  protected val testResult = test.testBody(dut)
```
seldridge, line 54:
```suggestion
  protected final val dut = Instance(test.dutDefinition)
  protected final val testResult = test.testBody(dut)
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -26,38 +26,32 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
+  override def desiredName = test.desiredTestModuleName
+
+  override def resetType = test.resetType match {
+    case Some(rt @ Module.ResetType.Synchronous)  => rt
+    case Some(rt @ Module.ResetType.Asynchronous) => rt
+    case _                                        => Module.ResetType.Synchronous
   }
+
+  final val finish = IO(Output(Bool()))
+  final val success = IO(Output(Bool()))
```
seldridge, line 46: Please include Scaladoc indicating what these mean.

Random thought: should these be protected? Are these ever supposed to be accessible from _outside_ the test harness?

tmckay-sifive: > Please include Scaladoc indicating what these mean.

Yep, will do.

> Random thought: should these be protected? Are these ever supposed to be accessible from outside the test harness?

I left them public so that there is a Chisel interface for these modules in addition to the FIR/SV interface from `Public`. For example, this is used to make a canned ChiselSim stimulus in https://github.com/chipsalliance/chisel/pull/4855

https://github.com/chipsalliance/chisel/blob/0f604d397ea9357d30383ea99f34c5e59a82f690/src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala#L10-L20

seldridge: Sorry, no, you are right, these need to be public.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -28,35 +28,34 @@ class TestParameters[M <: RawModule, R] private[inlinetest] (
...
-    *  @tparam M the type of the DUT module
-    *  @tparam R the type of the result returned by the test body
-    */
-  trait RawModule[M <: ChiselRawModule, R] extends Public { this: ChiselRawModule =>
-    def test: TestParameters[M, R]
-    override def desiredName = test.desiredTestModuleName
-    val dut = Instance(test.dutDefinition)
-    final def elaborateTest(): R = test.testBody(dut)
-  }
+  import chisel3.{Module => ChiselModule}
```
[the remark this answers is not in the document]
tmckay-sifive: Incidentally, I fixed this in https://github.com/chipsalliance/chisel/pull/4834

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: I intentionally left this non-final to allow people to override the default reset type when the testharness must infer it. Do you think that API is worth keeping open or is synchronous a good default?

tmckay-sifive: The only use case for this that I can think of is that a class extending `TestHarness` that wants to use async reset logic for the test hardware.

tmckay-sifive:
Broke out https://github.com/chipsalliance/chisel/pull/4832/commits/e280752a56bb383bee18fcef1b36157e8b2f45ce into https://github.com/chipsalliance/chisel/pull/4874. I need `TestHarness` to be a `Module` so that classes extending can use `clock` and `reset` implicitly.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: For now I've settled on inferring synchronous if we must. I think that makes the most sense since we are expecting the simulation to drive a synchronous reset for one cycle. If a user truly needs async logic, they can construct it in a `withReset` block or submodule.

PR #4834, opened 2025-03-26
Remove type parameter from chisel3.experimental.inlinetest.HasTests

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Desired Merge Strategy

#### Release Notes

- Remove type parameters from `HasTests`

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
+++ b/src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
@@ -93,7 +93,8 @@ object TestHarnessGenerator {
   *  @tparam TestResult the type returned from each test body generator, typically
   *  hardware indicating completion and/or exit code to the testharness.
   */
-trait HasTests[M <: RawModule] { module: M =>
+trait HasTests { module: RawModule =>
+  type M = module.type
 
   /** A Definition of the DUT to be used for each of the tests. */
   private lazy val moduleDefinition =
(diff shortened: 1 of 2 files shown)


jackkoenig:
Marked as No Release Notes instead of API Modification because there is not yet a release with this feature.

jackkoenig (commented):
Are we sure we want this? I'd like to see a test that works against an interface, e.g.
```scala
@instantiable
trait MyModuleIntf extends RawModule with HasTests {
  @public val io = ???

  test("test1")(MyModuleIntf.test1(_))
}
object MyModuleIntf {
  def test1(inst: Instance[MyModuleIntf]): Unit = ???

  def test2(inst: Instance[MyModuleIntf]): Unit = ???
}

class MyModule extends MyModuleIntf {
  def test("test2")(MyModuleIntf.test2(_))
}
```

The idea here is to make sure that the self typing is good enough and that we don't need F-bounded polymorphism (the type parameter that you're deleting).

jackkoenig:
Another polymorphism problem to think about is what if you have two `HasTests` traits you want to inherit from that are different and each define their own inline tests for their particular interfaces? I can't imagine many cases where the module would operate correctly under such circumstances though... but I do wonder how the typing works out. FWIW the self-typing is probably _better_ for this than the type parameter.

jackkoenig (approved):
Looks great!

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -93,7 +93,8 @@ object TestHarnessGenerator {
   *  @tparam TestResult the type returned from each test body generator, typically
   *  hardware indicating completion and/or exit code to the testharness.
   */
-trait HasTests[M <: RawModule] { module: M =>
+trait HasTests { module: RawModule =>
+  type M = module.type
```
seldridge, line 97: Given that this is now defined, can this be used to replace existing `module.type` in this trait?

seldridge (commented):
LGTM

PR #4835, opened 2025-03-26
[chiselsim] Add randomization to settings

seldridge:
Add the ability to control the existing FIRRTL randomization settings via an additional parameter to `chisel3.simulator.Settings`.  Set the default behavior for this to always randomize.  This is a reasonable default for Verilator or any other 2-state simulator.  For 4-state simulators, users may want to deviate and use the provided `uniniitialized` randomization.

This commit intentionally does not try to struggle with setting per-simulator settings that are different from each other.  While I think this could make sense (you would like to use randomization for Verilator and uninitialized w/ x-prop for VCS), I think that the logic necessary to handling this is going to be too opaque for most users.  I would rather keep this as a single setting on simulate and any simulator you use will then simulate the same thing.  If you want to do something different, then it is on the user.

Note: I don't expect most users to care about running different simulators, in spite of the fact that we provide the `Cli.Simulator` trait (and intentionally do not mix it into `ChiselSim`).  If users _do_ care about running different simulators, then they would be better served by enumerating these as explicit tests.  I.e., I am not super concerned about the use case of writing a test with one randomization mode and then expecting the randomization mode to morph if the user changes the simulator on the command line.

#### Release Notes

Add randomization control to ChiselSim settings. This changes the default randomization behavior to turn on FIRRTL randomization. If users would like to leave things uninitialized (a reasonable VCS setting), then there is a provided `Randomization.uninitialized` setting that can be used. Note: be careful using uninitialized with Verilator as without additional Verilator command line arguments, state will initialize to zero.

### DIFF
--- a/src/main/scala/chisel3/simulator/Randomization.scala
+++ b/src/main/scala/chisel3/simulator/Randomization.scala
@@ -84,7 +84,7 @@ object Randomization {
     registers = true,
     memories = true,
     delay = Some(1),
-    randomValue = Some("$random")
+    randomValue = Some("$urandom")
   )
 
 }
--- a/src/main/scala/chisel3/simulator/Randomization.scala
+++ b/src/main/scala/chisel3/simulator/Randomization.scala
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.simulator
+
+import svsim.CommonCompilationSettings.VerilogPreprocessorDefine
+
+/** A description of how a Chisel circuit should be randomized
+  *
+  * @param registers if true, randomize the initial state of registers
+  * @param memories if true, randomize the initial state of memories
+  * @param garbageAssign
+  * @param invalidAssign
+  * @param delay an optional delay value to apply to the randomization.  This
+  * will cause the randomization to be applied this many Verilog time units
+  * after the simulation starts.
+  * @throws IllegalArgumentException if register and memory randomization are
+  * both disabled and delay or randomValue are non-empty
+  */
+final class Randomization(
+  val registers:   Boolean,
+  val memories:    Boolean,
+  val delay:       Option[Int],
+  val randomValue: Option[String]
+) {
+
+  require(
+    registers || memories || (delay.isEmpty && randomValue.isEmpty),
+    "when register and memory randomization is disabled, then `delay` and `randomValue` should be empty as they have no effect"
+  )
+
+  require(
+    delay.isEmpty || delay.get != 0,
+    "a delay of zero is illegal as this can conflict with initial blocks and simulator-specific time-zero behavior"
+  )
+
+  /** Create a copy of this [[Randomization]] changing some parameters */
+  def copy(
+    registers:   Boolean = registers,
+    memories:    Boolean = memories,
+    delay:       Option[Int] = delay,
+    randomValue: Option[String] = randomValue
+  ) = new Randomization(registers, memories, delay, randomValue)
+
+  /** Convert this to Verilog preprocessor defines */
+  private[simulator] def toPreprocessorDefines: Seq[VerilogPreprocessorDefine] = {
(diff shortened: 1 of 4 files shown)


src/main/scala/chisel3/simulator/Randomization.scala
```diff
@@ -0,0 +1,90 @@
...
+    * Non-two-state values (i.e., `x`)
+    *
+    * @note The FIRRTL/Verilog ABI for randomization is undocumented in the
+    * FIRRTL ABI specification.
+    */
+  def random = new Randomization(
+    registers = true,
+    memories = true,
+    delay = Some(1),
+    randomValue = Some("$random")
```
jackkoenig, line 87: Should we consider defaulting to `$urandom`? I'm not an expert in this but I feel like that's what every DV engineer has always told me. As a historical note, I don't think Verilator supported `$urandom` back in the day but it does now.

src/test/scala-2/chiselTests/simulator/scalatest/ChiselSimSpec.scala
```diff
@@ -285,6 +286,95 @@ class ChiselSimSpec extends AnyFunSpec with Matchers with ChiselSim with FileChe
...
+        info(s"register is not zero, has value $regValue")
+        regValue should not be (0)
+
+        val cmemValue = dut.cm.peekValue().asBigInt
+        info(s"combinational read memory index zero is not zero, has value $cmemValue")
+        cmemValue should not be (0)
+
+        val smemValue = dut.sm.peekValue().asBigInt
+        info(s"sequential read memory index zero is not zero, has value $smemValue")
+        cmemValue should not be (0)
```
jackkoenig, line 329: I suspect the randomization is stable, but if it isn't, we have a about a 1 in 1.4 billion chance of this failing randomly every time we run this test 🙂.

jackkoenig (approved):
Looks good, consider comment about `$urandom`.

src/main/scala/chisel3/simulator/Randomization.scala
seldridge: Maybe.

What I would honestly like is to take our internal `$urandom` + `%m`-based randomization seeding for `$srandom` and move that into upstream. Unfortunately, that relies on VCS-specific options for getting the global seed. `$urandom` is probably a saner default, though.

This is pushing towards that CIRCT Verilog  library that we've been discussing off-and-on.

src/test/scala-2/chiselTests/simulator/scalatest/ChiselSimSpec.scala
seldridge: I actually wanted to increase it, but rand into a CIRCT "bug" where the randomization address is one bit too large for memories. Once I look at that, I'll increase the bit width to get the probability much lower.

seldridge: Ack, the CIRCT bug is reachable here: https://github.com/chipsalliance/chisel/actions/runs/14095137789/job/39480847729?pr=4835#step:11:1957

src/main/scala/chisel3/simulator/Randomization.scala
seldridge: Switched to `$urandom` in: https://github.com/chipsalliance/chisel/pull/4835/commits/8cf4e7ae0c83bc975b8ea0f33e9f9b70e1efeb55

src/test/scala-2/chiselTests/simulator/scalatest/ChiselSimSpec.scala
seldridge: I decreased the size of the `Queue` in the failing test to workaround this. I'll get a CIRCT fix in for this this week. If users run into this there are several workarounds, with the most direct being to disable that Verilator warning.

seldridge: Tracking issue here: https://github.com/llvm/circt/issues/8350

PR #4838, opened 2025-03-27
[test] Revert to original RecordSpec, NFC

seldridge:
Revert a small change in dd04d997d which was done to workaround CIRCT producing width warnings with Verilator.  This will be fixed in CIRCT 1.111.0.

I'm re-opening this as the nightly CI didn't seem to be kicking in.


seldridge:
This will fail until the nightly gets refreshed.

PR #4840, opened 2025-03-27
Fix layer.enable applying to all later modules

### LINKED ISSUE #4839: `layer.enable` incorrectly applies to submodules
Consider the following:

``` scala
//> using scala "2.13.16"
//> using dep "org.chipsalliance::chisel::7.0.0-M2+599-7965f8e4-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::7.0.0-M2+599-7965f8e4-SNAPSHOT"

import chisel3._

class Bar extends RawModule

class Foo extends RawModule {

  layer.enable(layers.Verification)

  val b = Module(new Bar)

}

object Main extends App {

  println(
    ChiselStage.emitCHIRRTL(new Foo)
  )

}
```

This produces the following FIRRTL:

``` firrtl
FIRRTL version 4.2.0
circuit Foo :
  layer Verification, bind, "verification" :
    layer Assert, bind, "verification/assert" :
    layer Assume, bind, "verification/assume" :
    layer Cover, bind, "verification/cover" :

  module Bar enablelayer Verification :

    skip


  public module Foo enablelayer Verification :

    inst b of Bar
```

`module Bar` should not have the `Verification` layer enabled.

While this should be harmless, it can cause errors if combined with `ModuleChoice` and external modules. This can then create multiple instance choices which have different layers enabled (because external modules cannot have layers enabled). This then results in CIRCT catching the problem: https://github.com/llvm/circt/issues/8353

seldridge:
Fix a bug where use of the `layer.enable` API would incorrect cause a propagation of layers being enabled to all subsequent modules.  This was due to a bug in how the state save/restore logic worked.

Fixes #4839.

#### Release Notes

Fix a bug where `layer.enable` would apply to all later modules.

### DIFF
--- a/src/test/scala-2/chiselTests/LayerSpec.scala
+++ b/src/test/scala-2/chiselTests/LayerSpec.scala
@@ -216,6 +216,23 @@ class LayerSpec extends AnyFlatSpec with Matchers with FileCheck {
 
   }
 
+  they should ", when enabled, not propagate to child modules" in {
+
+    class Bar extends RawModule
+
+    class Foo extends RawModule {
+      layer.enable(A)
+      val bar = Module(new Bar)
+    }
+
+    ChiselStage.emitCHIRRTL(new Foo).fileCheck() {
+      """|CHECK: module Bar :
+         |CHECK: module Foo enablelayer A :
+         |""".stripMargin
+    }
+
+  }
+
   they should "work correctly with Definition/Instance" in {
 
     @instantiable
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -1227,7 +1227,8 @@ private[chisel3] object Builder extends LazyLogging {
       Builder.setPrefix(state.prefix)
       Builder.currentClock = state.clock
       Builder.currentReset = state.reset
-      Builder.enabledLayers = enabledLayers
+      Builder.enabledLayers.clear()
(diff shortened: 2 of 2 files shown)


jackkoenig (approved):

src/test/scala-2/chiselTests/LayerSpec.scala
```diff
@@ -216,6 +216,23 @@ class LayerSpec extends AnyFlatSpec with Matchers with FileCheck {
...
+  they should ", when enabled, not propagate to child modules" in {
+
+    class Bar extends RawModule
+
+    class Foo extends RawModule {
+      layer.enable(A)
+      val bar = Module(new Bar)
+    }
+
+    ChiselStage.emitCHIRRTL(new Foo, Array("--full-stacktrace")).fileCheck() {
```
jackkoenig, line 228:
```suggestion
    ChiselStage.emitCHIRRTL(new Foo).fileCheck() {
```
Unnecessary (but benign) argument.

rwy7 (approved):

src/test/scala-2/chiselTests/LayerSpec.scala
seldridge: Thanks for catching this! I don't like when these stay in the tests. Fixed in: https://github.com/chipsalliance/chisel/pull/4840/commits/607aa6daf6a89764463f08933e9665956e8ea182

Issue #4837, opened 2025-03-27
Error occurs when using the VCS backend for simulation in SVSIM


jcvclouds:
In the Makefile file generated by SVSIM (VCS backend), there is a macro named **SVSIM_BACKEND_ENGAGES_IN_ASLR_SHENANIGANS**, which will cause the following error:

```
[info] - compiles an example testbench and runs tests *** FAILED ***
[info]   java.lang.Exception: Expected ' ' but got 'ronologic VCS simulator copyright 1991-2023'
[info]   at svsim.Simulation$Controller.mustRead$1(Simulation.scala:159)
[info]   at svsim.Simulation$Controller.readNextAvailableMessage(Simulation.scala:162)
[info]   at svsim.Simulation$Controller.processNextMessage(Simulation.scala:227)
[info]   at svsim.Simulation$Controller.expectNextMessage(Simulation.scala:237)
[info]   at svsim.Simulation$Controller.<init>(Simulation.scala:347)
[info]   at svsim.Simulation.run(Simulation.scala:53)
```

When I modify the source code of svsim (`svsim/src/main/scala/vcs/Backend.scala`) to remove this macro from the generated Makefile, the simulation can run normally.

I saw the following comments for this macro in `svsim/src/main/scala/Backend.scala`:

VCS first checks whether address-space layout randomization (ASLR) is enabled, and if it is, _helpfully_ relaunches this executable with ASLR disabled. Unfortunately, this causes code executed prior to `simulation_main` to be executed twice, which is problematic, especially since we redirect `stdin` and `stdout`.

Is this caused by a VCS version issue?

seldridge:
Can you post a full test if possible and I can try to run it?

Also, can you post information about what version of Chisel you are using.

I'm surprised that that define is what is making the difference. The error message is acting like there is an invalid response received and the response looks like VCS responding with the version which seems off.

jcvclouds:
Hi @seldridge, thank you for your prompt response.

I used the GCD test provided in the svsim source code directory (`svsim/src/test/scala/BackendSpec.scala`) and did not modify any source code.

The corresponding chisel version is the latest commit from the main branch, which I have published locally for use (I think recent chisel versions should have this issue).

After encountering this issue, I tried debugging and removed the macro from the svsim source code. After re-publishing the chisel to the local Ivy repository, I found that the test was working properly.

The following information may help you analyze this issue:
* System: CentOS 7.9.2009
[...]

PR #4845, opened 2025-03-30
[svsim] Use -no_save instead of ASLR workarounds

seldridge:
For the internal installation of VCS and RHEL, our simulations do not work due to VCS' save/restore functionality which will launch a simulation, detect if ASLR is present, and then kill the simulation and relaunch it with ASLR disabled.  The workaround employed here has been empirically shown to be problematic for at least one user (#4837).  Drop the ASLR workaround and add the runtime option `-no_save` which turns off VCS save/restore functionality.  svsim doesn't support this and it's ill advised to use non-standard workarounds.

#### Release Notes

Fix problems for some users when using svsim with VCS where a simulation would fail to start due to non-standard ASLR workarounds employed. Change svsim to run VCS simulations with the `-no_save` option to avoid VCS simulation relaunches on machines with ASLR enabled.

### DIFF
--- a/svsim/src/main/resources/simulation-driver.cpp
+++ b/svsim/src/main/resources/simulation-driver.cpp
@@ -37,10 +37,6 @@
 #include <strings.h>
 #include <unistd.h>
 
-#ifdef SVSIM_BACKEND_ENGAGES_IN_ASLR_SHENANIGANS
-#include <sys/personality.h>
-#endif
-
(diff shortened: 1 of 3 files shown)


seldridge:
CC: @jackkoenig: This will require some small patches internally to add `-no_save` to VCS' simulation-time arguments. I tested this on one test internally, using the old `ChiselsimScalatestTester`, and it worked.

PR #4849, opened 2025-04-01
Add SimulationTest marker

fabianschuiki:
The FIRRTL spec is about to define a `simulation` construct to indicate that a module should be executed as a simulation test. Add a user-facing `SimulationTest` class that can be used to mark a module as to be executed as a simulation test. This would commonly be used inside a test harness module to mark the surrounding module as a test. For example:

    class TestHarness extends RawModule {
      SimulationTest(this)
    }

The simulation test can be given an optional name and a map of parameters. The parameters are identical to the ones accepted by the `FormalTest` marker. Multiple simulation test markers may be added to a single module, which may be useful if a test should be run with different sets of user-defined parameters.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?


core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -576,12 +576,14 @@ private[chisel3] object ir {
...
 
-  case class DefFormalTest(
+  case class DefTestMarker(
+    kind:       String,
     name:       String,
     module:     BaseModule,
     params:     MapTestParam,
     sourceInfo: SourceInfo
   ) extends Component {
+    require(kind == "formal" || kind == "simulation")
```
seldridge: The PR could use an enumeration pattern here:

``` scala
object TestKind {
  sealed trait Type {
    def toString: String
  }
  case object Formal extends Type {
    override def toString = "formal"
  }
  case object Simulation extends Type {
    override def toString = "simulation"
  }
}
```

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -826,12 +826,13 @@ case class ArrayTestParam(value: Seq[TestParam]) extends TestParam with UseSeria
...
 
-/** Formal Test
+/** Formal or Simulation Test
   */
 @deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")
-case class FormalTest(info: Info, name: String, moduleName: String, params: MapTestParam)
+case class TestMarker(kind: String, info: Info, name: String, moduleName: String, params: MapTestParam)
     extends DefModule
     with UseSerializer {
+  require(kind == "formal" || kind == "simulation")
```
seldridge: Same comment as earlier about using an enumeration pattern.

core/src/main/scala/chisel3/TestMarker.scala
```diff
@@ -0,0 +1,137 @@
...
+    *   )
+    * }
+    * }}}
+    */
+  def apply(
+    module: BaseModule,
+    params: MapTestParam = MapTestParam(Map.empty),
+    name:   String = ""
+  )(implicit sourceInfo: SourceInfo): Unit = {
+    val proposedName = if (name != "") {
```
seldridge: Maybe `name.isEmpty`?

core/src/main/scala/chisel3/TestMarker.scala
```diff
@@ -0,0 +1,137 @@
...
+    *       ))
+    *     ))
+    *   )
+    * }
+    * }}}
+    */
+  def apply(
+    module: BaseModule,
+    params: MapTestParam = MapTestParam(Map.empty),
+    name:   String = ""
```
seldridge, line 119: My knee jerk reaction here is to use `Option[String]`. However, that then means you can have `Some("")` which doesn't make sense. This is fine.

jackkoenig:
> @jackkoenig: Can the binary compatibility be waived for this as it is changing FIRRTL IR case classes only?

Yeah that's fine

PR #4862, opened 2025-04-05
[firrtl] Drop Circuit from Target/Named

seldridge:
Remove the circuit from both `Target` and `Named`.  For the latter, this means that `CircuitName` is deleted.

Carry this to its logical conclusion.  The only parts that do require a change:

1. The RenameMap wants to know if a path is renamed to something "absolute".  It does this check by looking to see if a module name matches the circuit name.  (Note: this check is incorrect because it assumes that a circuit has a single root.  However, this has not been the case since D/I was implemented.)  Work around this by adding a "circuit name" parameter to the RenameMap so that this check can still be done.  Note: large code paths within the rename map that are checking for module or instance path renames should be completely unreachable.  The only user of the RenameMap at this point in-tree is `DataView` and this is not going to do renames involving paths.

2. The LayerControl API needs to know what the circuit name is since this is part of its ABI.  Previously, the circuit name was known because of a, now deleted, method on `RawModule`.  Change this to use the module name itself when determining the ABI setting (file name or macro name).  This should always be correct because the method to generate the ABI setting because ChiselSim is always given some top module that it wants to simulate.  This may be problematic for multi-rooted circuits like inline directed tests.

3. Users of the Trace API do not have a trivial access to the circuit name, i.e., the top module.  However, if you are a user of this, you should know what your top name is without relying on the returned `CompleteTarget`.  See the `TraceSpec` for an example of this.

See the associated FIRRTL spec change here: https://github.com/chipsalliance/firrtl-spec/commit/160bafef053f1fa28e8592ebf54400e81b90e210

#### Release Notes

- Remove the "circuit name" from all annotations and annotation-adjacent code. This has the effect of serializing all annotations using new target syntax that does _not_ have a circuit name. E.g., whereas a target may have previously looked like: `~Foo|Bar>a`, it will now look like: `~|Bar>a`. This is a change that is already compatible with CIRCT and requires no update to your FIRRTL compiler.
- Users that create a rename map will now need to pass a "circuit name" parameter to the rename map.
- `RawModule` no longer has a `circuitName` member.

### DIFF
--- a/core/src/main/scala/chisel3/internal/Builder.scala
+++ b/core/src/main/scala/chisel3/internal/Builder.scala
@@ -12,7 +12,7 @@ import chisel3.internal.binding._
 import chisel3.internal.firrtl.ir._
 import chisel3.internal.firrtl.Converter
 import chisel3.internal.naming._
-import _root_.firrtl.annotations.{Annotation, CircuitName, ComponentName, IsMember, ModuleName, Named, ReferenceTarget}
+import _root_.firrtl.annotations.{Annotation, ComponentName, IsMember, ModuleName, Named, ReferenceTarget}
 import _root_.firrtl.annotations.AnnotationUtils.validComponentName
 import _root_.firrtl.{annoSeqToSeq, AnnotationSeq}
 import _root_.firrtl.renamemap.MutableRenameMap
@@ -344,20 +344,6 @@ private[chisel3] trait HasId extends chisel3.InstanceId {
     case Some(p)          => p.name
     case None             => throwException(s"$instanceName doesn't have a parent")
   }
-  def circuitName: String = _parent match {
-    case None =>
-      // Only modules have circuits
-      this match {
-        case b: BaseModule =>
-          b._circuit match {
-            case Some(c) => c.circuitName
-            case None    => instanceName
-          }
-        case _ => instanceName
-      }
-    case Some(ViewParent) => reifyParent.circuitName
-    case Some(p)          => p.circuitName
-  }
 }
 
 /** Holds the implementation of toNamed for Data and MemBase */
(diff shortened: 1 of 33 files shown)


firrtl/src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -139,11 +133,10 @@ object Target {
   def deserialize(s: String): Target = {
     val regex = """(?=[~|>/:.\[])"""
     s.split(regex)
-      .foldLeft(GenericTarget(None, None, Vector.empty)) { (t, tokenString) =>
+      .foldLeft(GenericTarget(None, Vector.empty)) { (t, tokenString) =>
         val value = tokenString.tail
         tokenString(0) match {
-          case '~' if t.circuitOpt.isEmpty && t.moduleOpt.isEmpty && t.tokens.isEmpty =>
-            if (value == "???") t else t.copy(circuitOpt = Some(value))
+          case '~' if t.moduleOpt.isEmpty && t.tokens.isEmpty => t
```
seldridge, line 139: Note: the effect of this is that this will correctly tokenize targets which both have and do not have circuits. If there is a circuit, it will just be discarded. This is why the Trace API continues to work even though CIRCT will produce annotations that include a circuit.

core/src/main/scala/chisel3/internal/Builder.scala
```diff
@@ -344,20 +344,6 @@ private[chisel3] trait HasId extends chisel3.InstanceId {
     case Some(p)          => p.name
     case None             => throwException(s"$instanceName doesn't have a parent")
   }
-  def circuitName: String = _parent match {
```
jackkoenig, line 347: It would probably still be possible to keep this, possibly deprecated. I'm not sure of the value but worth considering--maybe if it comes up as an issue in `7.0.0-RC1` we can restore it.

jackkoenig (approved):
This makes me a bit nervous as it's a pretty substantial change. We could instead do something more modest--keeping circuit around in Chisel but just dropping it when we serialize. Everything in package `firrtl` is notionally deprecated so hopefully this won't affect too many users but we shall see in 7.0.0-RC1.

In any case, I understand the need so LGTM.

PR #4864, opened 2025-04-06
[docs] Add Probe Documentation

seldridge:
Add documentation of probes.  This includes light documentation of `BoringUtils` which is closely related.

### DIFF
--- a/docs/src/explanations.md
+++ b/docs/src/explanations.md
@@ -40,5 +40,6 @@ read these documents in the following order:
 * [Warnings](explanations/warnings)
 * [Deep Dive into Legacy Connection Operators](explanations/connection-operators)
 * [Properties](explanations/properties)
+* [Probes](explanations/probes)
 * [Layers](explanations/layers)
 * [Module Prefixing](explanations/moduleprefix)
--- a/docs/src/explanations/probes.md
+++ b/docs/src/explanations/probes.md
@@ -248,10 +248,10 @@ import chisel3.probe.{Probe, ProbeValue}
 class Foo extends Module {
 
   val d = IO(Input(UInt(32.W)))
-  val q = IO(Output(chiselTypeOf(d)))
-  val r_probe = IO(Output(Probe(chiselTypeOf(d))))
+  val q = IO(Output(UInt(32.W)))
+  val r_probe = IO(Output(Probe(UInt(32.W))))
 
-  private val r = Reg(chiselTypeOf(d))
+  private val r = Reg(UInt(32.W))
 
   q :<= r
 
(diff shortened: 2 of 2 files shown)


dtzSiFive:
> Mecahnistically

Typo

dtzSiFive:
> prots

ports

dtzSiFive:
This is great, thank you very much! This looks solid and I like the
examples and explanations.  Especially for some tricky areas!


The clone type stuff maybe should be somewhere else? But I appreciate the
kinda "things folks using probes tend to ask/run into" vibe so I say this
only bc this seems important beyond probes, ya know? Not that it's wrong
here.  Indeed this is likely a minor paint point for folks for now, so
thanks for putting this here as inevitably it would come up.

Also I think it's a bit inaccurate/incomplete, although maybe not worth
capturing the nuances of the current implementation -- very quickly the
best answer is to just go read it (or try it) vs documenting how it really
works.  But anyway (and these are kinda related):

1)none of them clone probe type information that isn't only applied to the
top-most level (err the level given to th clone?), iirc.  That is they all
at best will reapply top level probe-ness.  It may not be easy to apply
elsewhere from user code? But this matters for things like making leaf
fields rwprobe, for example.

2)you can and IMHO we should sometimes, create types that have probe fields
in them.  The description given assumes probes are only /modifications/ of
an existing chisel type.  Probe information is not stripped if it's part of
the type proper, only if it was added later (and it's still not stripped
just not re-added).

I'm good with not changing it, tricky needle to thread.

I'm a little surprised the ABI details are in the Chisel documentation? But
they're there for layers as well, works for me.  Something to remember
should they change, but all for well informed users.

Thanks again, appreciate it!  Module my nits and comments, generally LGTM!

~Will

seldridge:
Thanks for the review, @dtzSiFive!

Typos are fixed in: https://github.com/chipsalliance/chisel/pull/4864/commits/a9e378b01382b4645d72178da3486aa17232f64e

Based on your comments, I'd rather not have something partially useful related to cloning types. I'm removed that commentary or any example use of `chiselTypeOf` here: https://github.com/chipsalliance/chisel/pull/4864/commits/cb0b50e639d24af92f07c882c247cf53d6d31813

I think we can re-add this, but it would be better in the "cookbook" section. I.e., it's more of a Q&A of: "How do I clone a type and preserve probes?" or "How do I clone a type and strip probes?" Also, this is more of a library writer problem than a Chisel user problem and it's fine if that is left out for now.

@dtzSiFive wrote:
> I'm a little surprised the ABI details are in the Chisel documentation? But
> they're there for layers as well, works for me.  Something to remember
> should they change, but all for well informed users.

Realize that I also wrote the layers documentation, so the consistency should be unsurprising.

My thought process here is that a Chisel user should be able to use Chisel essentially independently of knowledge of FIRRTL. For that reason, they need to have some knowledge of the FIRRTL/Verilog ABI in order to do their job. There's already missing places where this comes up / is left unsaid, e.g., around the port ABIs involving aggregates. Users pick up on this eventually, but I think it should be more explicitly stated.

When showing this, the PR tries to show what is minimally necessary and avoid some of the exactness of the ABI. E.g., there is this ABI, it looks like this, here's an example, and you can probably intuit the rest, but here is a link to the FIRRTL ABI if you need to know everything.

One question @dtzSiFive: Is there anything missing in the "Why Input Probes and Not Allowed" section?

PR #4875, opened 2025-04-09
Add SimulationTime (%T) format specifier

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Unlike `%t` in Verilog, SimulationTime (`%T`) does not take an argument.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Printable.scala
+++ b/core/src/main/scala/chisel3/Printable.scala
@@ -72,7 +72,7 @@ sealed abstract class Printable {
 
 object Printable {
 
-  private[chisel3] def isNoArgSpecifier(c: Char): Boolean = c == '%' || c == 'm'
+  private[chisel3] def isNoArgSpecifier(c: Char): Boolean = c == '%' || c == 'm' || c == 'T'
 
   /** Pack standard printf fmt, args* style into Printable
     */
@@ -102,7 +102,7 @@ object Printable {
     while (iter < fmt.size) {
       // Encountered % which is either
       // 1. Describing a format specifier.
-      // 2. %% or %m
+      // 2. %%, %m, or %T
       // 3. Dangling percent - most likely due to a typo - intended literal percent or forgot the specifier.
       // Try to give meaningful error reports
       if (fmt(iter) == '%') {
@@ -193,10 +193,10 @@ object Printable {
   /** Resolve Printables that are resolved at Chisel-time */
   private[chisel3] def resolve(pable: Printable, ctx: Component)(implicit info: SourceInfo): Printable =
     pable.map {
-      case Name(data)             => PString(data.ref.name)
-      case FullName(data)         => PString(data.ref.fullName(ctx))
-      case HierarchicalModuleName => PString("{{HierarchicalModuleName}}")
-      case other                  => other
+      case Name(data)     => PString(data.ref.name)
+      case FullName(data) => PString(data.ref.fullName(ctx))
+      case s: SpecialFirrtlSubstitution => PString(s.substitutionString)
+      case other => other
     }
 }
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/chisel3/Printable.scala
```diff
@@ -377,11 +377,29 @@ case object Percent extends Printable {
   final def unpack:     (String, Seq[Data]) = ("%%", List.empty)
 }
 
+/** Printable with special representation in FIRRTL */
+sealed trait SpecialFirrtlSubstitution extends Printable {
+  def substitutionString: String
+}
```
jackkoenig, line 383: @seldridge is this what you had in mind? Or should the `{{ }}` be part of this String rather than above?

I'm not sure how I feel about adding this public API since it's really just to make an implementation slightly simpler--of two minds.

seldridge: I would make this protected with a private converter that adds the braces:

``` scala
sealed trait SpecialFirrtlSubstitution extends Printable {
  protected def substitutionString: String
  private [chisel3] toFirrtlString = "{{" + substitutionString + "}}"
}
```

There is another alternative which I've used elsewhere which derives the string automatically from the class name. I did this here for VCS flags: https://github.com/chipsalliance/chisel/blob/2bb07daf45c597c4bb7bafbc89280a5d2a0bf1ce/svsim/src/main/scala/vcs/Backend.scala#L228

E.g.:

``` scala
sealed trait SpecialFirrtlSubstitution { this: Singleton with Printable => 
  private [chisel3] substitutionString: String = "{{" + this.getClass.getSimpleName + "}}"
}

case object HierarchicalModuleName extends Printable with SpecialFirrtlSubstitution
```

This is perhaps a bit too cute, however, the motivation with the VCS options was that I wanted the names to match _exactly_ what was in the VCS docs. The case for matching the FIRRTL docs less strong, though forcing the linkage by name helps with understandability.

Do what makes sense to you.

docs/src/explanations/printing.md
```diff
@@ -62,12 +62,16 @@ printf(cf"myUInt = $myUInt%c") // myUInt = !
 There are special values you can include in your `cf` interpolated string:
 
 * `HierarchicalModuleName` (`%m`): The hierarchical name of the current module
+* `SimulationTime` (`%T`): The current simulation time (unlike Verilog's `%t`, this does not take an argument)
```
seldridge, line 65: I like `%T` (capital T) as this is different from Verilog.

seldridge (approved):
LGTM

Consider the comment and see what you think.

docs/src/explanations/printing.md
jackkoenig: That was my thinking. Technically `%T` is legal in SystemVerilog, but despite ChatGPT swearing to me that it has different semantics than `%t` (it tried to claim that `%T` includes the unit when printing lol), `%T` in Verilog is the same as `%t` but obviously much less frequently used.

core/src/main/scala/chisel3/Printable.scala
jackkoenig: Yeah I like that a lot better, will do, thanks!

docs/src/explanations/printing.md
seldridge: Ack... yeah, I guess `%T` is the same as `%t`. I can't find any references in 1800-2023 about the capital version other than that it lists them as being the same. After looking more, ever format specifier has both lowercase and uppercase versions. It doesn't seem to clarify on any difference other than they are alternatives. I would expect that simulators will implement `%X`/`%H` as capital hex. However, this wouldn't make sense for a radix or format that doesn't include Latin characters. 😕

PR #4878, opened 2025-04-09
Deprecate everything in package firrtl

jackkoenig:
Previously, we tried to avoid this by deprecating any imports from the firrtl package, but this missed things. Instead, we deprecate every public class, object, and trait.

Also delete the previous attempt using the DeprecateSFCComponent in the compiler plugin. It is now redundant.

I tried to write a [POSIX] sed script to do this but I was really struggling with having it take any number of `abstract|final|case|sealed` so instead I asked an AI and it generated a stupid but effective way to do it:

`deprecate.sed`:
```
/^sealed abstract class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^abstract class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^final class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^case class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^final case class/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^trait/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^sealed trait/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^object/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")

/^case object/i \
@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")
```

```bash
find firrtl/src/main/scala -name '*.scala' | xargs sed -i '' -f deprecate.sed
```

Naturally the initial version missed a bunch of things so I expanded it till it seemed to work.

This does not get things inside of package objects, and I don't think you can deprecate them, so instead I handled those manually, it was only like 6 things.

### Contributor Checklist

#### Type of Improvement

- API deprecation

#### Desired Merge Strategy

- Squash

#### Release Notes

If there is anything you cannot do without accessing APIs in `package firrtl` directly, please [open an issue on the Chisel Github repository](https://github.com/chipsalliance/chisel/issues/new/choose) to explain your use case so we can see about adding APIs to Chisel directly.

- [ ] Did you review?


seldridge (approved):
LGTM

Nice to see the compiler plugin go.

PR #4882, opened 2025-04-10
Add SimLog for printing to files

jackkoenig:
This is implemented using FIRRTL's fprintf API.

This is called `SimLog` because it's "Simulation Logging". I don't want it to say anything about "Files" since it's a common type that includes StdErr for now and eventually might even include Strings or other things one could log to (maybe even DPI???). It also enables us to eventually add `ElabLog` for "Elaboration Logging" which would be generation-time logging.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

You can now print to files:
```scala
val fd = SimLog.file("logfile.txt")
fd.printf(cf"foo = $foo%0d\n")
```

- [ ] Did you review?


core/src/main/scala/chisel3/FileDescriptor.scala
```diff
@@ -0,0 +1,97 @@
...
+    fmt:  String,
+    data: Bits*
+  )(
+    implicit sourceInfo: SourceInfo
+  ): chisel3.printf.Printf =
+    printfWithoutReset(Printable.pack(fmt, data: _*))
+}
+
+// Uses firrtl fprintf with a String filename
+private class StringLogFile(filename: String) extends FileDescriptor {
```
jackkoenig, line 78:
```suggestion
private case class StringLogFile(filename: String) extends FileDescriptor {
```
This should be a case class so that it has the right `==` behavior (and makes debug printing nicer).

core/src/main/scala/chisel3/FileDescriptor.scala
```diff
@@ -0,0 +1,97 @@
...
+  override protected def fileDescriptor: Option[String] = Some(filename)
+}
+
+// Defaults to firrtl printf
+private object DefaultDescriptor extends FileDescriptor {
```
jackkoenig, line 83:
```suggestion
private case object DefaultDescriptor extends FileDescriptor {
```

firrtl/src/main/scala/firrtl/ir/IR.scala
```diff
@@ -435,6 +435,20 @@ case class Print(
     with IsDeclaration
     with UseSerializer
 
+@deprecated("All APIs in package firrtl are deprecated.", "Chisel 7.0.0")
+case class Fprint(
+  val info:   Info,
+  val fd:     String,
```
seldridge, line 441: Nit: this is really a filename and not a file descriptor.

core/src/main/scala/chisel3/internal/firrtl/IR.scala
```diff
@@ -492,7 +492,8 @@ private[chisel3] object ir {
 
   case class Port(id: Data, dir: SpecifiedDirection, sourceInfo: SourceInfo)
 
-  case class Printf(id: printf.Printf, sourceInfo: SourceInfo, clock: Arg, pable: Printable) extends Definition
+  case class Printf(id: printf.Printf, sourceInfo: SourceInfo, fd: Option[String], clock: Arg, pable: Printable)
+      extends Definition
```
seldridge, line 496: Nit again: It's not an `fd`, but a `filename`.

src/test/scala-2/chiselTests/FileDescriptorSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.annotation.nowarn
+
+class FileDescriptorSpec extends AnyFlatSpec with Matchers with FileCheck {
+
+  behavior.of("FileDescriptor")
+
+  it should "support simple printing to files" in {
+    class MyModule extends Module {
+      val fd = FileDescriptor("logfile.log")
+      fd.printf(cf"An exact string")
```
terpstra, line 21: 👍

terpstra (approved):
Nice!

core/src/main/scala/chisel3/FileDescriptor.scala
```diff
@@ -0,0 +1,97 @@
...
+    *
+    * @see [[Printable]] documentation
+    * @param pable [[Printable]] to print
+    */
+  def printf(pable: Printable)(implicit sourceInfo: SourceInfo): chisel3.printf.Printf = {
+    this.printfWithReset(pable)(sourceInfo)
+  }
+
+  // The actually underlying file descriptor, None means just use printf
+  protected def fileDescriptor: Option[String]
```
seldridge, line 38: To move towards unifying this with stdout/stderr, introducing a new enum of: `File(name: String) | StdErr` would be a way to start. This would also avoid the question of "what is the default" when this is `None`. This _should_ unify fprintf with printf from the Chisel side and enable moving in the direction of `LogFile` or something like that?

core/src/main/scala/chisel3/FileDescriptor.scala
```diff
@@ -0,0 +1,97 @@
...
+
+// Uses firrtl fprintf with a String filename
+private class StringLogFile(filename: String) extends FileDescriptor {
+  override protected def fileDescriptor: Option[String] = Some(filename)
+}
+
+// Defaults to firrtl printf
+private object DefaultDescriptor extends FileDescriptor {
+  override protected def fileDescriptor: Option[String] = None
+}
```
seldridge, line 85: This is really `StdErr` based on how FIRRTL compilers have historically compiled printfs, right?

src/test/scala-2/chiselTests/FileDescriptorSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+    ChiselStage
+      .emitCHIRRTL(new MyModule(FileDescriptor("logfile.log")))
+      .fileCheck()(
+        """CHECK: fprint(clock, UInt<1>(0h1), "logfile.log", "in = %0d\n", in)"""
+      )
+    ChiselStage
+      .emitCHIRRTL(new MyModule(FileDescriptor.Default))
+      .fileCheck()(
+        """CHECK: printf(clock, UInt<1>(0h1), "in = %0d\n", in)"""
+      )
```
seldridge, line 57: This unification with `printf` is right on the money.

seldridge (commented):
Quick review. The main thing is that I don't like talking about "file descriptors" when we're we're not talking about [File Descriptors](https://en.wikipedia.org/wiki/File_descriptor). Every place that this is used, it's really just an object that stores a filename. (You can argue that this _is_ a file descriptor, however, that always makes me think of an integer as this has such strong OS connotations. If tied to this, I would suggest "file handle".)

That said, I think there's a nice unification that moves away from using `filedescriptor: Option[String]` and instead uses a sum type for this of `File(name: String) | StdErr` which should provide a tighter unification of printf and fprintf.

I do like that this already unifies printf and fprintf via the same API.

src/test/scala-2/chiselTests/SimLogSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+  it should "support writing to a file in simulation" in {
+    class MyModule extends Module {
+      val (count, done) = chisel3.util.Counter(0 until 4)
+      val fd = SimLog.file("logfile.log")
+      fd.printf(cf"count = $count%0d\n")
+      when(done) { stop() }
+    }
+    simulate(new MyModule)(RunUntilFinished(5))
+    val testdir = implicitly[HasTestingDirectory].getDirectory
+    val logfile = testdir.resolve("workdir-verilator").resolve("logfile.log")
```
seldridge, line 69: This should delete the file before running the test to prevent a stale file from being read.

core/src/main/scala/chisel3/SimLog.scala
```diff
@@ -0,0 +1,97 @@
...
+  )(
+    implicit sourceInfo: SourceInfo
+  ): chisel3.printf.Printf = {
+    val clock = Builder.forcedClock
+    val printfId = new chisel3.printf.Printf(pable)
+
+    Printable.checkScope(pable)
+
+    layer.block(layers.Verification, skipIfAlreadyInBlock = true, skipIfLayersEnabled = true) {
+      Builder.pushCommand(Printf(printfId, sourceInfo, this._filename, clock.ref, pable))
```
seldridge, line 63: Nit: the `this` looks unnecessary:

```suggestion
      Builder.pushCommand(Printf(printfId, sourceInfo, _filename, clock.ref, pable))
```

core/src/main/scala-2/chisel3/SimLogIntf.scala
```diff
@@ -0,0 +1,51 @@
...
+    *
+    * ===Format Specifiers===
+    *
+    * Format specifiers are prefixed by `%`. If you wish to print a literal `%`, use `%%`.
+    *   - `%d` - Decimal
+    *   - `%x` - Hexadecimal
+    *   - `%b` - Binary
+    *   - `%c` - 8-bit Character
+    *   - `%n` - Name of a signal
+    *   - `%N` - Full name of a leaf signal (in an aggregate)
```
seldridge, line 44: Should this mention the new format specifiers?

seldridge (approved):
LGTM!

core/src/main/scala-2/chisel3/SimLogIntf.scala
jackkoenig: It should, I was just moving it so didn't really notice. I'll update it!

PR #4889, opened 2025-04-14
[Scala3] Add ChiselComponent plugin

adkian-sifive:
ChiselComponent naming plugin with 80% of Scala 2 plugin's functionality. Supports:
- Chisel-style grabbing of Data names from typed syntax trees and transforming `val` definitions to call `withName` method
- Detecting Options and Iterables typed on Data for naming, and naming them accordingly
- Checking nested Iterables for Data subtypes and naming them

Not supported yet:
- Chisel constructs defined in either simple or complex tuple unapply types. 
- Handling ClassDef source locators: this feels out of place here, I'll likely refactor this out in future work

Simple unapply methods look like `val (a, b) = (Wire(UInt(8.W)), Wire(UInt(8.W))`, while complex ones involve calls to constructor functions that return tuples. Desugaring tuple unapplys works quite differently in Scala 3 than in Scala 2 - specifically, tuple ValDef constructors are desugared into a indices of a temporary node. For example:

```
val (a, b) = (Wire(UInt(8.W)), Wire(UInt(8.W))
val temp = TupleExpr(...)
val a = temp._1
val b = temp._2
```
Since RHS now make a reference to a temporary node, ChiselComponent's ValDef transform cannot add a `withName` to the RHS of val definitions as in the Scala 2 version which uses a product version of `withName`.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Adds initial Scala3 naming plugin

- [ ] Did you review?


core/src/main/scala/chisel3/experimental/prefix.scala
```diff
@@ -55,6 +55,14 @@ object prefix {
...
     ret
   }
+
+  def apply[T](name: Any)(f: => T): T = {
+    name match {
+      case x: HasId  => applyHasId[T](x)(f)
+      case x: String => applyString[T](x)(f)
+      case _ => f
+    }
+  }
```
jackkoenig, line 65: This change is fine for development, but I do not want to merge this. It is less type safe and strictly inferior to just overloading apply with the correct concrete subtypes.

If we must have something temporarily, we can have `applyHasId` and `applyString` _in addition to_ the overloaded `apply` methods, but there must be a way to make the Scala 3 plugin do the right thing.

adkian-sifive: yeah it's not great - i'll move these into their own methods. it's on my todo list to ask around on how to get types resolved in the `appliedTo` calls

plugin/src/main/resources/plugin.properties
```diff
@@ -0,0 +1 @@
+pluginClass=chisel3.internal.plugin.ChiselComponent
```
jackkoenig, line 1: I think we need to differentiate resources for Scala 2 and Scala 3

jackkoenig (commented):
Looks good to get going, do you want to port tests in this PR or in a follow on?

adkian-sifive:
I'll move the tests over in a follow up, after Unapply naming is functional and a couple other changes (VerifStatement macro apply stubs) go through

jackkoenig (approved):

PR #4903, opened 2025-04-23
Support reset-initializing the lastGrant register in RRArbiter

tymcauley:
The default behavior remains the same: don't initialize the `lastGrant` register in `LockingRRArbiter`. See #267 for previous discussions on this.

As noted in that old PR, this register is frequently in the control path of a design, where X-propagation can lead to simulation errors. There are nice ways to work around this uninitialized register in RTL simulations (namely the `RANDOMIZE_REG_INIT` text macro in the emitted SystemVerilog). However I've run into issues with non-resettable registers in the control path during gate-level simulations, where it's a bit harder to control the random initialization of registers. Having the option to resolve that problem with an RTL change would be very useful.

A few other notes about this PR:

- I chose an API where a `Boolean` lets the user opt in to resetting this register to 0. We could also use an `Option[UInt]` argument (which defaults to `None`), which would let users reset the register to other values. My gut says the `Option[UInt]` style is more complexity than necessary, and passing `true` to this argument to get reset behavior is a simpler API than passing `Some(0.U)`.
- There aren't any tests for the arbiters. Also no documentation for the locking arbiters. I might have time to dig into that in the near future, but I first wanted to check if there's any prior code for that lying around somewhere I'm not aware of.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Add the `initLastGrant` argument to the round-robin arbiter classes (`RRArbiter` and `LockingRRArbiter`) to support reset-initializing a register which might be in the control path of designs.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/util/Arbiter.scala
+++ b/src/main/scala/chisel3/util/Arbiter.scala
@@ -75,10 +75,18 @@ abstract class LockingArbiterLike[T <: Data](gen: T, n: Int, count: Int, needsLo
   }
 }
 
-class LockingRRArbiter[T <: Data](gen: T, n: Int, count: Int, needsLock: Option[T => Bool] = None)
-    extends LockingArbiterLike[T](gen, n, count, needsLock) {
-  // this register is not initialized on purpose, see #267
-  lazy val lastGrant = RegEnable(io.chosen, io.out.fire)
+class LockingRRArbiter[T <: Data](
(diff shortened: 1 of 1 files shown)


jackkoenig (approved):
With all the caveats from #267 that this is a bandaid and not really a fix, I think it's fine

PR #4910, opened 2025-04-30
[svsim] Add 'additionalArguments' to allow adding arbitrary flags to the Verilator backend

kammoh:
This allows adding arbitrary flags to the executed verilator command.
An example use case for me was adding "--trace-max-array" to be able to trace arrays in the generated waveforms.
Exposing all verilator flag in `CompilationSettings` seems impossible so this could be a way to allow for different use cases without cluttering up the Chisel side.

### Contributor Checklist

#### Type of Improvement
- Feature (or new API)

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/svsim/src/main/scala/verilator/Backend.scala
+++ b/svsim/src/main/scala/verilator/Backend.scala
@@ -20,7 +20,8 @@ object Backend {
     outputSplitCFuncs:          Option[Int] = None,
     disabledWarnings:           Seq[String] = Seq(),
     disableFatalExitOnWarnings: Boolean = false,
-    enableAllAssertions:        Boolean = false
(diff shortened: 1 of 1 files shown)


seldridge (commented):
Can the PR instead add any option that it needs on-demand?

I.e., if this is. done to add `--max-trace-array`, then add that.

I realize that this approach will make each backend's compilation options get comically large. However, it's more typesafe, it avoids having two ways to do things (`outputSplitCFuncs` can then be specified two ways), and it's consistent with the direction that the project has started down.

So, add whatever you want! If it turns into 800 options, I'm fine with that. 😅

kammoh:
Fair enough. Replaced with #4911

PR #4911, opened 2025-04-30
[svsim] Trace options for the Verilator backend

kammoh:
+ Adds the following options for generating traces using the Verilator backend:
  - `maxArraySize`: Maximum array depth for tracing
  - `maxWidth`: Maximum bit width for tracing
  - `traceDepth`: Depth of tracing

The corresponding Verilator flags are only added if the above integer values are > 0.

+ Adds `TraceStyle.Fst` for generating FST traces. 

+ Removes the _unused_ `filename` field from `TraceStyle`, which seemed very confusing. The trace file name (stem) is only configured through `CommonSimulationSettings.traceFileStem`.

I guess, technically an API change due to the change of the `TraceStyle.Vcd` case class?

Replaces #4910

### Contributor Checklist

#### Type of Improvement
- Feature (or new API)
- API modification

#### Desired Merge Strategy

- [ ] Did you review?


kammoh:
One question is if the default values for `maxArraySize` and `maxWidth` should be non-zero. Not being able to see arrays or wide signals in the waveform could be confusing to new-commers (or people less familiar with Verilator). I don't think _performance_ is major concern as waveforms are only used for debugging, and from my experience any extra overhead is unnoticable.

svsim/src/main/scala/Simulation.scala
```diff
@@ -5,14 +5,33 @@ package svsim
...
+  protected final def getProjectRootOrCwd: Path =
+    sys.props
+      .get("chisel.project.root")
+      .orElse(sys.env.get("MILL_WORKSPACE_ROOT"))
+      .orElse(sys.env.get("PWD"))
+      .map(Paths.get("").toAbsolutePath.resolve(_))
+      .getOrElse(Paths.get(""))
+      .toAbsolutePath
+
+  def workingDirectory: Path = {
```
jackkoenig, line 27: This looks like a better API to use generally should we deprecate `val workingDirectoryPath` and suggest people use this?

kammoh: I can't think why not, so I moved `getProjectRootOrCwd` to the `Simulation` object so it's accessable elsewhere (and public).
I also added `CHISEL_PROJECT_ROOT` env variable as an alternative way of specifying/detecting the project root.
Then I used it in `chisel3.testing.scalatest`'s implicit implementation of `HasTestingDirectory`.
Now `scalatest.ChiselSim` build directory in my mill projects are created in the workspace folder (`./build`)  instead of being hidden under `./out/module_name/tests/testOnly.dest/sandbox/build`.

seldridge (commented):
This generally seems great to me. I have some nits about how to structure the options for Verilator to be more flexible.

svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -8,9 +8,54 @@ import scala.collection.mutable
...
+        Option.when(traceUnderscore)("--trace-underscore") ++
+        Option.when(maxArraySize > 0)(s"--trace-max-array-size=$maxArraySize") ++
+        Option.when(maxWidth > 0)(s"--trace-max-width=$maxWidth") ++
+        Option.when(traceDepth > 0)(s"--trace-depth=$traceDepth")
+
+      def traceUnderscore: Boolean
+      def maxArraySize:    Int
+      def maxWidth:        Int
+      def traceDepth:      Int
+    }
```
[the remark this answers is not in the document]
kammoh: Thanks, I updated TraceStyle based on your suggestions. Please let me know your thoughts about this version.

svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -8,9 +8,55 @@ import scala.collection.mutable
...
-    sealed trait TraceStyle
-    object TraceStyle {
-      case class Vcd(traceUnderscore: Boolean = false, filename: String = "") extends TraceStyle
+
+    sealed trait TraceKind {
+      private[Backend] def toCompileFlags: Seq[String]
+    }
+
+    object TraceKind {
+
```
seldridge, line 17: This is a mega-nit, but I prefer enums like this:

```suggestion
    object TraceKind {
      sealed trait Type {
        private[Backend] def toCompileFlags: Seq[String]
      }

```

svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -8,9 +8,55 @@ import scala.collection.mutable
...
+      /**
+        * VCD tracing
+        */
```
seldridge, line 20: Mega-nit (throughout):

```suggestion
      /** VCD tracing */
```

svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -8,9 +8,55 @@ import scala.collection.mutable
...
+        final def toCompileFlags: Seq[String] =
+          Seq("--trace-fst") ++
+            Option.when(traceThreads > 0)(Seq("--trace-threads", traceThreads.toString)).toSeq.flatten
+      }
+
+      /**
+        * FST tracing
+        * 
+        * @param traceThreads If set to a positive value, enables FST waveform creation using `traceThreads` separate threads
+        */
```
seldridge, line 36:
```suggestion
      /** FST tracing
        * 
        * @param traceThreads If set to a positive value, enables FST waveform creation using `traceThreads` separate threads
        */
```

svsim/src/main/scala/verilator/Backend.scala
```diff
@@ -8,9 +8,55 @@ import scala.collection.mutable
...
+      case class Fst(traceThreads: Int) extends TraceKind with FstTraceKind
+
+      case object Fst extends TraceKind with FstTraceKind { override val traceThreads = 0 }
```
seldridge, line 39: This duplication feels a little baroque. I'd opt for just a single `case class Fst` with a `traceThreads: Option[Int]` parameter.

seldridge (approved):
LGTM modulo nits.

svsim/src/main/scala/verilator/Backend.scala
kammoh: The difference in use with the `Vcd` option didn't feel right, so I resorted to this. But I agree it's a bit weird.
I'll change all `Int` parameters to `Option[Int]` then.

svsim/src/main/scala/verilator/Backend.scala
kammoh: fixed

PR #4913, opened 2025-05-02
[docs] Add a blog post about type classes

jackkoenig:
#### Type of Improvement

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/website/blog/2025/05-05-arbiter-type-class/index.md
+++ b/website/blog/2025/05-05-arbiter-type-class/index.md
@@ -1,0 +1,280 @@
+---
+authors:
+  - jackkoenig
+tags: [design patterns]
+slug: arbiter-type-class
+description: Type classes and hardware design.
+---
+
+# Type Classes and Hardware Design
+
+When giving talks about Chisel, I always say that Chisel is intended for writing _reusable hardware generators_.
+The thesis is that, by virtue of being embedded in Scala, Chisel can take advantage of powerful programming language features to make hardware design easier.
+How to actually do this; however, is usually left as an exercise for the audience.
+
+In this blog post, we'll explore how one might use the *type class* pattern to build Chisel generators.
+
+## A Specialized Arbiter for a Specific Protocol
+
+Let's first build a simple fixed-priority arbiter that works for a specific kind of client:
+
+```scala
+class ArbiterClient extends Bundle {
+  val request = Bool()
+  val grant = Flipped(Bool())
+}
+
+class PriorityArbiter(nClients: Int) extends Module {
+  val clients = IO(Vec(nClients, Flipped(new ArbiterClient)))
+
+  val requests = clients.map(_.request)
+  val granted = PriorityEncoderOH(requests)
+
+  for ((client, grant) <- clients.zip(granted)) {
+    client.grant := grant
+  }
+}
+```
+
+<!-- truncate -->
+
+`ArbiterClient` is written from the perspective of the client--`request` is in the default orientation, intended to be driven by the client, and `grant` is flipped, intended to be driven by the arbiter.
+
+This works fine if you only ever need to arbitrate `ArbiterClient` bundles, but it doesn't generalize.
+For example, how would you use this arbiter with a `ready-valid` interface like `Decoupled`?
+
+```scala
+val needToArbitrate = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter = Module(new PriorityArbiter(needToArbitrate.size))
+
+for ((client, decoupled) <- arbiter.clients.zip(needToArbitrate)) {
+  client.request := decoupled.valid
+  decoupled.ready := client.grant
+}
+```
+
+This works, but its a bit clunky.
+Considering the implementation of `PriorityArbiter`, is only 6 lines of Scala, it's not ideal that it takes another 4 lines to use it.
+
+## Generalizing the Arbiter
+
+What we really want is an arbiter that is generic to the type of the client.
+For those familiar with _Object-Oriented Programming_, you might be tempted to try to use inheritance to define a common interface for clients:
+
+```scala
+trait Arbitrable {
+  def request: Bool
+  def grant: Bool
+}
+
+class ArbiterClient extends Bundle with Arbitrable {
+  val request = Bool()
+  val grant = Flipped(Bool())
+}
+```
+
+However, for our example above, this would be a bit difficult.
+`Decoupled` is defined within Chisel itself--how can a user make Chisel's `Decoupled` inherit from `Arbitrable`?
+
+Instead, we can try something different.
+We could make the Arbiter generic to the type of the client and then use higher-order functions to extract the request and grant signals.
+
+```scala
+class GenericPriorityArbiter[A <: Data](
+    nClients: Int,
+    clientType: A
+  )(
+    requestFn: A => Bool,
+    grantFn: (A, Bool) => Unit) extends Module {
+  val clients = IO(Vec(nClients, Flipped(clientType)))
+
+  val requests = clients.map(requestFn(_))
+  val granted = PriorityEncoderOH(requests)
+
+  for ((client, grant) <- clients.zip(granted)) {
+    grantFn(client, grant)
+  }
+}
+```
+
+You may notice this looks quite similar to the original `PriorityArbiter` in its implementation.
+It uses two parameter lists in order to help the Scala type inferencer derive the types of the functions of the type of the client--we could do this with one parameter list but then it would require explicitly passing the type of the client.
+
+Now we can use it for both our `ArbiterClient` and `Decoupled` interfaces.
+
+```scala
+val clients1 = IO(Vec(4, Flipped(new ArbiterClient)))
+val arbiter1 = Module(
+  new GenericPriorityArbiter(4, new ArbiterClient)(_.request, (c, g) => c.grant := g)
+)
+arbiter1.clients :<>= clients1
+
+val clients2 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter2 = Module(
+  new GenericPriorityArbiter(4, Decoupled(UInt(8.W)))(_.valid, (d, g) => d.ready := g)
+)
+arbiter2.clients :<>= clients2
+```
+
+This is still a bit clunky--we have to pass two additional arguments to the arbiter.
+
+Another thing to consider is that these additional arguments are the same for all instances of a given type.
+For example, if we were to arbitrate another set of `Decoupled` interfaces, we would have to pass the same functions again:
+
+```scala
+val clients3 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter3 = Module(
+  new GenericPriorityArbiter(4, Flipped(Decoupled(UInt(8.W))))(_.valid, (d, g) => d.ready := g)
+) // the two function arguments are the same as above.
+arbiter2.clients :<>= clients3
+```
+
+## Introducing a Type Class
+
+To clean this up even more, we can introduce a type class that captures the "arbitrable" pattern:
+
+```scala
+trait Arbitrable[A] {
+  def request(a: A): Bool
+  def grant(a: A, value: Bool): Unit
+}
+```
+
+Effectively, we have taken the two arguments to the arbiter and turned them into methods on the type class.
+This looks similar to the proposed object-oriented version of `Arbitrable` above, but note how it is parameterized by the type of the client and accepts the client as an argument.
+
+We can then provide instances of this type class for specific types. For example, for `ArbiterClient` and `Decoupled`:
+
+```scala
+class ArbiterClientArbitrable extends Arbitrable[ArbiterClient] {
+  def request(a: ArbiterClient) = a.request
+  def grant(a: ArbiterClient, value: Bool) = a.grant := value
+}
+
+class DecoupledArbitrable[T <: Data] extends Arbitrable[DecoupledIO[T]] {
+  def request(a: DecoupledIO[T]) = a.valid
+  def grant(a: DecoupledIO[T], value: Bool) = a.ready := value
+}
+```
+
+Then, we can refactor the arbiter to use the type class:
+
+```scala
+class GenericPriorityArbiter[A <: Data](nClients: Int, clientType: A, arbitrable: Arbitrable[A]) extends Module {
+  val clients = IO(Vec(nClients, Flipped(clientType)))
+
+  val requests = clients.map(arbitrable.request(_))
+  val granted = PriorityEncoderOH(requests)
+
+  for ((client, grant) <- clients.zip(granted)) {
+    arbitrable.grant(client, grant)
+  }
+}
+```
+
+This makes instantiating the arbiter a hair nicer:
+
+```scala
+val clients1 = IO(Vec(4, Flipped(new ArbiterClient)))
+val arbiter1 = Module(new GenericPriorityArbiter(4, new ArbiterClient, new ArbiterClientArbitrable))
+arbiter1.clients :<>= clients1
+
+val clients2 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter2 = Module(new GenericPriorityArbiter(4, Decoupled(UInt(8.W)), new DecoupledArbitrable[UInt]))
+arbiter2.clients :<>= clients2
+```
+
+At least we aren't repeating logic anymore, instead we get to just refer to the type class instance.
+
+However, we can do even better.
+
+## Implicit Type Class Instances
+
+Scala has a powerful feature called **implicit resolution**.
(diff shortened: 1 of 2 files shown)


website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  val requests = clients.map(_.request)
+  val granted = PriorityEncoderOH(requests)
+
+  for ((client, grant) <- clients.zip(granted)) {
+    client.grant := grant
+  }
+}
+```
+
+<!-- truncate -->
```
mwachs5, line 39: what does this do?

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+val arbiter = Module(new PriorityArbiter(needToArbitrate.size))
+
+for ((client, decoupled) <- arbiter.clients.zip(needToArbitrate)) {
+  client.request := decoupled.valid
+  decoupled.ready := client.grant
+}
+```
+
+This works, but its a bit clunky.
+Considering the implementation of `PriorityArbiter`, is only 6 lines of Scala, it's not ideal that it takes another 4 lines to use it.
```
mwachs5, line 57:
```suggestion
Considering the implementation of `PriorityArbiter` is only 6 lines of Scala, it's not ideal that it takes another 4 lines to use it.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  val request = Bool()
+  val grant = Flipped(Bool())
+}
+```
+
+However, for our example above, this would be a bit difficult.
+`Decoupled` is defined within Chisel itself--how can a user make Chisel's `Decoupled` inherit from `Arbitrable`?
+
+Instead, we can try something different.
+We could make the Arbiter generic to the type of the client and then use higher-order functions to extract the request and grant signals.
```
mwachs5, line 80:
```suggestion
We could make the Arbiter generic to the type of the client and then require additional function arguments that tell how extract the request and grant signals from the particular client type we are using.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+```scala
+class GenericPriorityArbiter[A <: Data](
+    nClients: Int,
+    clientType: A
+  )(
+    requestFn: A => Bool,
+    grantFn: (A, Bool) => Unit) extends Module {
```
mwachs5, line 88:
```suggestion
    /** Function that indicates how to connect request from type A */
    requestFn: A => Bool,
    /** Function that indicates how to connect the grant from type A */
    grantFn: (A, Bool) => Unit) extends Module {
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  val granted = PriorityEncoderOH(requests)
+
+  for ((client, grant) <- clients.zip(granted)) {
+    grantFn(client, grant)
+  }
+}
+```
+
+You may notice this looks quite similar to the original `PriorityArbiter` in its implementation.
+It uses two parameter lists in order to help the Scala type inferencer derive the types of the functions of the type of the client--we could do this with one parameter list but then it would require explicitly passing the type of the client.
```
mwachs5, line 101:
```suggestion
> NOTE: It uses two parameter lists in order to help the Scala type inferencer derive the types of the functions of the type of the client--we could do this with one parameter list but then it would require explicitly passing the type of the client.
```

maybe? this is sort of an aside

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+Now we can use it for both our `ArbiterClient` and `Decoupled` interfaces.
+
+```scala
+val clients1 = IO(Vec(4, Flipped(new ArbiterClient)))
+val arbiter1 = Module(
+  new GenericPriorityArbiter(4, new ArbiterClient)(_.request, (c, g) => c.grant := g)
```
mwachs5, line 108: not sure if it might be helpful to show a version of this with the arguments explicitly called out by name. Generally when showing big code blocks it can be helpful to try to draw people's attention to the most relevant bits. Then we could say something like "we can make this shorter / more scala-y by writing it like this :... but it's still pretty clunky"

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+val clients3 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter3 = Module(
+  new GenericPriorityArbiter(4, Flipped(Decoupled(UInt(8.W))))(_.valid, (d, g) => d.ready := g)
+) // the two function arguments are the same as above.
+arbiter2.clients :<>= clients3
+```
+
+## Introducing a Type Class
+
+To clean this up even more, we can introduce a type class that captures the "arbitrable" pattern:
```
mwachs5, line 134:
```suggestion
To clean this up even more, we can introduce a _type class_ that captures the "arbitrable" pattern:
```

As we were talking about before, to me this just looks like a `trait`, maybe a "trait that is parameterized by some type A". You are defining what a type class is with one example, and I'm missing the specific thing about it that you're calling out as the type class pattern.

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+```scala
+trait Arbitrable[A] {
+  def request(a: A): Bool
+  def grant(a: A, value: Bool): Unit
+}
+```
+
+Effectively, we have taken the two arguments to the arbiter and turned them into methods on the type class.
+This looks similar to the proposed object-oriented version of `Arbitrable` above, but note how it is parameterized by the type of the client and accepts the client as an argument.
```
mwachs5, line 144:
```suggestion
This looks similar to the proposed object-oriented version of `Arbitrable` above, but note how it is parameterized by the type of the client and each function defined in the trait accepts the client as an argument.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+We can then provide instances of this type class for specific types. For example, for `ArbiterClient` and `Decoupled`:
```
mwachs5, line 146:
```suggestion
We can then define subclasses (or implementations?) of this type class for specific types. For example, for `ArbiterClient` and `Decoupled`:
```
Why do we call this an "instance". This is just a class. one would have to instantiate it to make it an instance. I think later we will make default implicit instances, but we haven't done that yet, right?

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  def grant(a: ArbiterClient, value: Bool) = a.grant := value
+}
+
+class DecoupledArbitrable[T <: Data] extends Arbitrable[DecoupledIO[T]] {
+  def request(a: DecoupledIO[T]) = a.valid
+  def grant(a: DecoupledIO[T], value: Bool) = a.ready := value
+}
+```
+
+Then, we can refactor the arbiter to use the type class:
```
mwachs5, line 160:
```suggestion
Then, we can refactor the arbiter to use the type class so that we're getting one 'package' of functions for type A, not having to pass them individually:
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+val clients1 = IO(Vec(4, Flipped(new ArbiterClient)))
+val arbiter1 = Module(new GenericPriorityArbiter(4, new ArbiterClient, new ArbiterClientArbitrable))
+arbiter1.clients :<>= clients1
+
+val clients2 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter2 = Module(new GenericPriorityArbiter(4, Decoupled(UInt(8.W)), new DecoupledArbitrable[UInt]))
+arbiter2.clients :<>= clients2
+```
+
+At least we aren't repeating logic anymore, instead we get to just refer to the type class instance.
```
mwachs5, line 187:
```suggestion
At least we aren't repeating logic anymore, instead we get to reuse the code for making the type class.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+However, we can do even better.
+
```
mwachs5, line 190: I'm sort of wondering why we even need to mention "type class" at all, up until this point. Like you could describe the pattern first, then say "by the way, this pattern of making a trait that is parameterized by type, and all its methods are parameterized by type, we call 'type class'"

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+## Implicit Type Class Instances
+
+Scala has a powerful feature called **implicit resolution**.
+This allows us to avoid passing around the type class instance explicitly.
```
mwachs5, line 194:
```suggestion
This allows us to avoid figuring out what type class we need to instantiate at every call site.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+Instead, we can define the type class instance as an implicit value and the compiler will automatically find it for us.
```
mwachs5, line 195:
```suggestion
Instead, we can define a default function to use when a specific type class is needed, and the compiler will automatically find it for us. We do this by making the argument to the function implicit, then making sure the implicit value of the type class is in scope.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+Let us rewrite our typeclass instances as implicit values:
```
mwachs5, line 197:
```suggestion
Let us instantiate implicit values of our type class instances. This tells the compiler, "if you need an `Arbitrable[ArbiterClient]`, use this one."
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+```scala
+implicit val arbiterClientArbitrable: Arbitrable[ArbiterClient] =
```
mwachs5, line 200:
```suggestion
// We could make a def, but since this function is the same every time, we just make this a `val`.
implicit val arbiterClientArbitrable: Arbitrable[ArbiterClient] =
```

Would it be worth showing this as a `def` first, or showing the `Decoupled` one first, since i think the `def` is the more obvious follow-on from what you've been showing, the `val` is sort of an optimization

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  new Arbitrable[ArbiterClient] {
+    def request(a: ArbiterClient) = a.request
+    def grant(a: ArbiterClient, value: Bool) = a.grant := value
+  }
+
+// In chisel3.util, the type is DecoupledIO while we construct instances of it with Decoupled.
+// Note that this is a def because DecoupledIO itself takes a type parameter.
```
mwachs5, line 207:
```suggestion
// Note that this is a def because DecoupledIO itself takes a type parameter,
// so we can't reuse the same one for every call-site.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+implicit def decoupledArbitrable[T <: Data]: Arbitrable[DecoupledIO[T]] =
+  new Arbitrable[DecoupledIO[T]] {
+    def request(a: DecoupledIO[T]) = a.valid
+    def grant(a: DecoupledIO[T], value: Bool) = a.ready := value
+  }
+```
+
+And then we can refactor the arbiter to use the implicit type class instance:
```
mwachs5, line 215:
```suggestion
Now we can refactor the arbiter to make its `arbitrable` argument `implicit`:
```

I might actually put this part first. Show that we could make this an implicit argument and still pass it manually, BUT if we create the implicit instances, then the compiler can do that final step for us.

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+val clients1 = IO(Vec(4, Flipped(new ArbiterClient)))
+val arbiter1 = Module(new GenericPriorityArbiter(4, new ArbiterClient))
+arbiter1.clients :<>= clients1
+
+val clients2 = IO(Vec(4, Flipped(Decoupled(UInt(8.W)))))
+val arbiter2 = Module(new GenericPriorityArbiter(4, Decoupled(UInt(8.W))))
+arbiter2.clients :<>= clients2
+```
+
+This is much cleaner and more readable.
```
mwachs5, line 242:
```suggestion
This is much cleaner and more readable. Even more importantly, it makes it the responsibility of the library
writer to determine how to make a certain type `Arbitrable`, not everyone who instantiates an arbiter.
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+
+Scala also has special syntax for implicit typeclass instances:
+
+```scala
+class GenericPriorityArbiter[A <: Data : Arbitrable](nClients: Int, clientType: A) extends Module {
```
mwachs5, line 247:
```suggestion
// Equivalent to:
// class GenericPriorityArbiter[A <: Data](nClients: Int, clientType: A)(implicit arbitrable: Arbitrable[A]) extends Module {
class GenericPriorityArbiter[A <: Data : Arbitrable](nClients: Int, clientType: A) extends Module {
```

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+  ...
+}
+```
+
+This is equivalent to the previous definition, but is more concise.
+Note that unlike the version with the implicit argument, this one does not bind a variable name for the implicit argument.
```
mwachs5, line 253: is this really relevant for the blog post? is this something people are going to need to do?

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+Note that Scala has rules for _implicit resolution_ for how to find the type class instance for a given type.
+As a general rule, you should define implicit type class instances in the companion object of the type they are for, or in the companion object for the type class itself.
+
+For example, since `DecoupledIO` is defined in Chisel itself, you could define the implicit value in the companion object for `Arbitrable`:
+```scala
+object Arbitrable {
+  implicit def decoupledArbitrable[T <: Data]: Arbitrable[DecoupledIO[T]] = ...
+}
+```
+
```
mwachs5, line 269: Or if you are defining a new type you want to make arbitrable, you could include it in the package object for that type

Isn't this the more common case? Can we show it?

website/blog/2025/05-05-arbiter-type-class/index.md
```diff
@@ -0,0 +1,280 @@
...
+For more information, see [further reading](#further-reading) below.
+
+## Conclusion
+
+This example only scratches the surface of what type classes can do in Chisel and Scala.
+Whenever you find yourself passing around the same bits of logic repeatedly, think about whether a type class could capture that pattern.
```
mwachs5, line 275:
```suggestion
Whenever you find yourself passing functions around repeatedly, or are struggling with an inheritance pattern, think about whether a type class could capture that pattern.
```

mwachs5:
not sure what the milestone should be, when did we introduce the blog to the website

website/blog/2025/05-05-arbiter-type-class/index.md
seldridge: It's a directive to tell Docusaurus to only show what is above in the "collapsed" view on the webiste. I didn't use this for the only other blog post: https://www.chisel-lang.org/blog

If I put that between the first and second paragraphs, then it would only show the first paragraph and display a "Read More" button to expand. For an example, you can see the "Read More" behavior on the Docusaurus blog here: https://docusaurus.io/blog

jackkoenig: And if you don't have it, Docusaurus prints a warning (and since this post is long, the Blog page looks a bit silly)

website/blog/2025/05-05-arbiter-type-class/index.md
jackkoenig: Yeah I think that's a good suggestion.

PR #4923, opened 2025-05-12
[chiselsim] Set +incdir on all primary sources

seldridge:
Set `+incdir` for every primary source directory.  Previously, only the
root primary source directory would be set.

This change is done to allow for the FIRRTL ABI related to bind layers to
be changed to _not_ use an FIRRTL compiler output directory-relative path
for `` `include `` [[1]].  This is necessary because some flows are
rearranging the directory structure of Chisel-generated Verilog, which
includes layers, and paths relative to the root will break.

This is a compromise between complexity of the build system, ABI
simplicitly, and ease of directory manipulation.  This has the effect of
users with custom build flows will need to recognize that they need to set
the `+incdir` recursively based on all output directories that were
created.  Practically, this is easy to do.

[1]: https://github.com/llvm/circt/pull/8474

### DIFF
--- a/src/main/scala/chisel3/simulator/Simulator.scala
+++ b/src/main/scala/chisel3/simulator/Simulator.scala
@@ -2,7 +2,9 @@ package chisel3.simulator
 
 import chisel3.{Data, RawModule}
 import firrtl.options.StageUtils.dramaticMessage
-import java.nio.file.Paths
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.{FileVisitResult, FileVisitor, Files, Path, Paths}
+import scala.collection.mutable
 import scala.util.{Failure, Success, Try}
 import scala.util.control.NoStackTrace
 import svsim._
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -123,12 +125,36 @@ trait Simulator[T <: Backend] {
...
+      override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+        FileVisitResult.CONTINUE
+      }
+
+      override def preVisitDirectory(dir: Path, attrs: BasicFileAttributes): FileVisitResult = {
+        FileVisitResult.CONTINUE
+      }
+
+      override def postVisitDirectory(dir: Path, ioe: java.io.IOException): FileVisitResult = {
+        primarySourcesDirectories += dir.toString()
```
jackkoenig, line 141:
```suggestion
        primarySourcesDirectories += dir.toString
```
Nit but I don't think the parentheses are necessary (would be an error if `toString` were a Scala API, but not for Java APIs).

jackkoenig (approved):

src/main/scala/chisel3/simulator/Simulator.scala
seldridge: 👍 Fixed in: https://github.com/chipsalliance/chisel/pull/4923/commits/a92bbc03c39eefb8acd1a3cfb5ec6d132aef7a54

kammoh:
This seems to have broken ChiselSim Verilator simulations for me:

```
%Error: blahblah.sv:4:12: Cannot find include file: 'layers-blahblah-Verification.sv'
```

The file is generated under a `verification` subfolder, but the `+incdir` only adds the higher-level `primary-sources` directory.

PR #4925, opened 2025-05-13
Fail boring into module with endIOCreation set

azidar:
### Contributor Checklist

#### Type of Improvement

- API modification

#### Release Notes

Setting `endIOCreation` now also errors if any IOs are created using `BoringUtils`.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -858,7 +858,8 @@ package experimental {
     private[chisel3] def createSecretIO[A <: Data](data: => A)(implicit sourceInfo: SourceInfo): A = {
       val iodef = data
       requireIsChiselType(iodef, "io type")
-      require(!isFullyClosed, "Cannot create secret ports if module is fully closed")
+      if (isFullyClosed) Builder.error(s"Cannot create secret ports into fully closed ${this.name} (from ${Builder.currentModule.get.name})")
+      if (!isIOCreationAllowed) Builder.error(s"Cannot create secret ports into ${this.name} (from ${Builder.currentModule.get.name}) if IO creation is not allowed")
 
       Module.assignCompatDir(iodef)
       iodef.bind(SecretPortBinding(this), iodef.specifiedDirection)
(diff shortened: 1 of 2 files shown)


src/test/scala-2/chiselTests/BoringUtilsSpec.scala
```diff
@@ -689,4 +689,21 @@ class BoringUtilsSpec extends AnyFlatSpec with Matchers with LogUtils with FileC
...
+
+    class Baz extends RawModule {
+      val c = Wire(Property[Int]())
+      endIOCreation()
+    }
+
+    val e = intercept[Exception] {
+      circt.stage.ChiselStage.emitCHIRRTL(new Bar, args)
+    }
+    e.getMessage should include("Cannot create secret ports into Baz (from Bar) if IO creation is not allowed")
```
mwachs5, line 707: this is a user level API that's doing the wrong thing, but users presumably don't know about "secret ports". Can we make this error message reference the bore? Can we say "bore or tap"?

core/src/main/scala/chisel3/Module.scala
```diff
@@ -858,7 +858,8 @@ package experimental {
     private[chisel3] def createSecretIO[A <: Data](data: => A)(implicit sourceInfo: SourceInfo): A = {
       val iodef = data
       requireIsChiselType(iodef, "io type")
-      require(!isFullyClosed, "Cannot create secret ports if module is fully closed")
+      if (isFullyClosed) Builder.error(s"Cannot create secret ports into fully closed ${this.name} (from ${Builder.currentModule.get.name})")
+      if (!isIOCreationAllowed) Builder.error(s"Cannot create secret ports into ${this.name} (from ${Builder.currentModule.get.name}) if IO creation is not allowed")
```
mwachs5, line 862:
```suggestion
      if (!isIOCreationAllowed) Builder.error(s"Cannot create bore or tap into ${this.name} (from ${Builder.currentModule.get.name}) if IO creation is not allowed")
```

or do a try-catch to give the user the right error message in boring utils?

src/test/scala-2/chiselTests/BoringUtilsSpec.scala
mwachs5: presumably this test needs to be updated

Also is there a reason not to update the older message too? It would be a change, but the same argument applies. Is there any other user-level way of creating "secret ports"?

azidar: I believe the only way to create "secret ports" is through BoringUtils.

PR #4933, opened 2025-05-21
[core] Add Bool "implies" method

seldridge:
Add a metho to Bool for logical implication.  Use a prose method
name ("implies") as opposed to a terse operator to avoid
problems/confusion that was brought up in original PR review.

Fixes #1500.

#### Release Notes

Add `implies` method to `Bool` for logical implication.


fabianschuiki (approved):
LGTM! Nice test :smiley:

PR #4942, opened 2025-05-28
[chiselsim] Add frequency control to stimulus

seldridge:
Add the ability to better control the frequency of a simulation in
ChiselSim stimulus.  Previously, this was left up to the simulator to
define and could result in some weird behaviors: Verilator defaults to
`1ps/1ps` (which would cause simulations to seemingly simulate very fast)
and VCS defaults to `1s/1s` (which would cause simulations to seemingly
simulate very slowly).

Change the default behavior to simulate at 1GHz (by using a default
timescale of `1ns/100ps` and making the clock period 10 precision units).
Additionally, make the first two steps of the reset procedure longer to
keep things nicely aligned in waveforms.

Due to how Verilator and non-Verilator (e.g., VCS) simulators are handled
differently by svsim, this requires some changes to align the two
implementations of the `run_simulation` function.  For Verilator, this is
implemented in C++ and calls Verilator's `timeInc(Int)` function.  This
will increment time by the precision.  E.g., if the timescale is
`1ns/100ps`, then `#2` will increment by 200ps.  For non-Verilator, this
function is implemented in Verilog.  Add an argument to an svsim function
that generates this Verilog to increment in multiples of precision.

After this PR, Verilator and VCS will generate the same VCD (checked by
inspection in a waveform).

Note 1: the change to `run_simulation` in Verilog is much easier in
1800-2023 as this has `$timeunit` and `$timeprecision` system functions.
However, VCS only supports 1800-2012.

Note 2: this is being done because some gate-level sims are being done
using ChiselSim and accurate power requires realistic frequencies. If
users don't care about this (or aren't integrating "real" models of
hardware) then this change shouldn't matter or affect users other than
seeing the units in the favorite waveform viewer change.

Here is the waveform I was inspecting:

image

### DIFF
--- a/src/main/scala/chisel3/simulator/PeekPokeAPI.scala
+++ b/src/main/scala/chisel3/simulator/PeekPokeAPI.scala
@@ -36,7 +36,12 @@ trait PeekPokeAPI {
   }
 
   implicit class testableClock(clock: Clock) {
-    def step(cycles: Int = 1): Unit = {
+    def step(cycles: Int = 1, period: Int = 10): Unit = {
+      require(
+        period >= 2,
+        s"specified period, '${period}', must be 2 or greater because an integer half period must be non-zero"
+      )
+
       val module = AnySimulatedModule.current
       module.willEvaluate()
       if (cycles == 0) {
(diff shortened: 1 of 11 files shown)


seldridge:
Note 3: @jackkoenig brought up some good points with this that it would be better to figure out how to _not_ have to pass the period to the stimulus. I.e., there should be a way to do this which has information about the number of precision ticks for each clock already available and the tick methods just do what the user wants. Unfortunately, I am not seeing a great way to make this work right now.

PR #4946, opened 2025-06-04
ChiselSim API for inline testing

tmckay-sifive:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

- Add a ChiselSim API for simulating inline tests

- [ ] Did you review?


svsim/src/main/scala/Workspace.scala
```diff
@@ -65,7 +65,7 @@ final class Workspace(
     else
       s"${System.getProperty("user.dir")}/$path"
 
-  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
+  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes. artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
```
tmckay-sifive:
```suggestion
  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes. These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -55,85 +181,65 @@ final class TestResultBundle extends Bundle {
...
-   * - [[finish]]: the test shall be considered complete on the first positive edge of [[finish]].
+   *  that of the DUT, or is [[Synchronous]] if it must be inferred.
    *
    *  @tparam M the type of the DUT module
-   *  @tparam R the type of the result returned by the test body
    */
-abstract class TestHarness[M <: RawModule, R](test: TestParameters[M, R])
-    extends FixedIOModule(new TestResultBundle)
+abstract class TestHarness[M <: RawModule](test: TestParameters[M])
+    extends FixedIOModule(new TestHarnessInterface)
```
tmckay-sifive, line 168: I removed the `R` type parameter (the thing returned from a test-body and consumed by the testharness).

Instead of leaving this API wide open and expecting users to implement their own test-testharness interfaces, I added a new `TestConfiguration` case class.

The idea is that as the test body elaborates it builds a `TestConfiguration`, which is more or less "how the testharness should handle this test."

Based on what that is (e.g. "run for 10 cycles", "end when this signal is asserted"), the testharness knows how to wiggle the finish/success pins.

In the future, this might also include things like "expect some assertion to fire" if we want to support tests that expect to fail.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,152 @@ import scala.collection.mutable
...
+  finishCondition:  Option[Bool],
+  successCondition: Option[Bool],
+  failureMessage:   Option[Printable]
+) {
+  private[inlinetest] def driveInterface(testName: String, intf: TestHarnessInterface) = {
+    intf.finish := finishCondition.getOrElse(false.B)
+    intf.success := successCondition.getOrElse(true.B)
+    failureMessage.foreach { failureMessage =>
+      when(intf.finish && intf.success) {
+        printf(cf"${testName} failed: ${failureMessage}")
```
mikeurbach, line 37: this says `failed`, but it prints when `intf.success` is true... am i missing something?

tmckay-sifive: No not missing anything. Also, I forgot I had this feature here--need to add a test for it.

src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -118,12 +141,69 @@ trait Simulator[T <: Backend] {
...
+          () => module,
+          args = chiselOptsModifications(chiselOpts).toSeq,
+          firtoolArgs = firtoolOptsModifications(firtoolOpts).toSeq
+        )
+    _simulate(workspace, elaboratedModule, settings)(body)
+  }
+
+  final def simulateTests[T <: RawModule with HasTests, U](
+    module:           => T,
+    includeTestGlobs: Seq[String],
```
seldridge: Nit: for consistency:

```suggestion
    includeTestGlobs: Array[String],
```

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
```diff
@@ -0,0 +1,77 @@
...
+import firrtl.options.StageUtils.dramaticMessage
+
+class InlineTestSignaledFailureException private[simulator]
+    extends RuntimeException(
+      dramaticMessage(
+        header = Some(s"The test finished and signaled failure"),
+        body = ""
+      )
+    )
+    with NoStackTrace
```
seldridge: This should be moved into `Simulator.scala`'s `Exceptions` object. I also think this is a generic `TestFailed` exception and not specific to inline unit tests.

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    clock.tick(
+      timestepsPerPhase = 1,
+      maxCycles = 1,
+      inPhaseValue = 1,
+      outOfPhaseValue = 0,
+      sentinel = None
+    )
+
+    reset.set(0)
```
seldridge: This looks like `stimulus.ResetProcedure`. Reuse that. That will require adding a `period` parameter to this `apply` method. However, that's fine.

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    while (finish.get().asBigInt == 0) {
+      clock.tick(
+        timestepsPerPhase = 1,
+        maxCycles = 1,
+        inPhaseValue = 1,
+        outOfPhaseValue = 0,
+        sentinel = None
+      )
+      cycleCount += 1
```
seldridge: Doing this tick-by-tick may be very inefficient for ChiselSim because each tick requires IPC. Instead, can this be rewritten in terms of a sentinel:

```suggestion
    while (finish.get().asBigInt == 0) {
      clock.tick(
        timestepsPerPhase = 1,
        maxCycles = 1,
        inPhaseValue = 1,
        outOfPhaseValue = 0,
        sentinel = Some(finish, 1)
      )
      cycleCount += 1
```

There is also a `checkElapsedCycleCount` parameter that you can use here.

seldridge (commented):
Some comments about the stimulus.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
seldridge, line 1: This is not for this PR, but we need to move this out of `experimental`.

src/test/scala-2/chiselTests/experimental/InlineTestSpec.scala
```diff
@@ -91,54 +95,54 @@ class ModuleWithTests(
...
+  test("passing") { instance =>
     instance.io.in := 3.U(ioWidth.W)
-    assert(instance.io.out === 3.U): Unit
+    assert(instance.io.out === 3.U)
+    TestConfiguration.runForCycles(10)
   }
 }
 
-class InlineTestSpec extends AnyFlatSpec with FileCheck {
+class InlineTestSpec extends AnyFlatSpec with FileCheck with ChiselSim {
```
seldridge: Mix-in `Matchers` here (which was already imported) and this will make the testing a lot easier. Inline examples follow:
```suggestion
class InlineTestSpec extends AnyFlatSpec with ChiselSim with FileCheck with Matchers {
```

src/test/scala-2/chiselTests/experimental/InlineTestSpec.scala
```diff
@@ -402,12 +388,111 @@ class InlineTestSpec extends AnyFlatSpec with FileCheck {
...
+    case _ => ()
+  }
+
+  it should "simulate and pass if finish asserted with success=1" in {
+    val results = simulateTests(
+      new ModuleWithTests,
+      tests = TestChoice.Name("passing"),
+      timeout = 100
+    )
+    assert(results.size == 1, "Expected exactly one test result")
```
seldridge: Avoid `assert` where possible. Instead use the matchers:

```suggestion
    results.size should be(1)
```

The other, major benefit of this is that you don't need to write messages.

Ditto throughout.

src/test/scala-2/chiselTests/experimental/InlineTestSpec.scala
```diff
@@ -402,12 +388,111 @@ class InlineTestSpec extends AnyFlatSpec with FileCheck {
...
+    assertPass(results.head.result)
```
seldridge: With the matchers mixed in, this can be a lot simpler and you can avoid the helper method:

```suggestion
    results.head.result should be(TestResult.Success)
```

Note: usually this matches are not matching on a singleton and instead matching on a type. To write this that way, you would use:

```suggestion
    results.head.result shouldBe a[TestResult.Success.type]
```

Ditto throughout.

svsim/src/main/scala/Workspace.scala
```diff
@@ -352,6 +352,31 @@ final class Workspace(
...
+    val sourcePrimarySources = new File(this.primarySourcesPath)
+    if (sourcePrimarySources.exists()) {
+      Files
+        .walk(sourcePrimarySources.toPath)
+        .filter(Files.isRegularFile(_))
+        .forEach { source =>
+          val relativePath = sourcePrimarySources.toPath.relativize(source)
+          val targetPath = Paths.get(newPrimarySources.getPath, relativePath.toString)
+          targetPath.getParent.toFile.mkdirs()
+          Files.createSymbolicLink(targetPath, source)
```
seldridge: Mega-nit: "targetPath" and "source" are the exact opposite of the Java docs for this. Java calls the first argument "link" and the second "target". However, this is muddling that.

seldridge: This is creating a symlink which uses an absolute path. This will create problems for sandboxed build systems (it's also slightly less than ideal generally). According to the Java docs for this, if the second parameter is relative, then the link will be relative. Can this be changed so that it is relative?

src/test/scala-2/chiselTests/experimental/InlineTestSpec.scala
```diff
@@ -402,12 +388,111 @@ class InlineTestSpec extends AnyFlatSpec with FileCheck {
...
+  }
+
+  it should "simulate and fail if finish asserted with success=0" in {
+    val results = simulateTests(
+      new ModuleWithTests,
+      tests = TestChoice.Name("failing"),
+      timeout = 100
+    )
+    assert(results.size == 1, "Expected exactly one test result")
+    assertFail("test signaled failure")(results.head.result)
```
seldridge: As another alternative, you can combine these into one match:

```suggestion
    results.map(_.result) should contain theSameElementsInOrderAs Seq(TestResult.Failure("test signaled failure"))
```

https://www.scalatest.org/user_guide/using_matchers#workingWithSequences is a great read that gives you a lot of powerful ways to write these things.

src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -118,12 +141,69 @@ trait Simulator[T <: Backend] {
...
+    chiselOpts:       Array[String] = Array.empty,
+    firtoolOpts:      Array[String] = Array.empty,
+    settings:         Settings[TestHarness[T]] = Settings.defaultRaw[TestHarness[T]]
+  )(body: (SimulatedModule[TestHarness[T]]) => U)(
+    implicit chiselOptsModifications: ChiselOptionsModifications,
+    firtoolOptsModifications:         FirtoolOptionsModifications,
+    commonSettingsModifications:      svsim.CommonSettingsModifications,
+    backendSettingsModifications:     svsim.BackendSettingsModifications
+  ): Seq[SimulatedTest] = {
```
seldridge: One big deviation from `simulate` is that `simulate` will assert if the test fails while `simulateTests` will not. This is going to be a big footgun for users as they need to remember to check that all their tests pass. Instead, we should aggressively error if any test fails.

Instead, aggregate all the failing tests into a single exception with a very verbose message that shows all the tests that are failing.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -178,40 +279,44 @@ trait HasTests { module: RawModule =>
...
     *  testharness has clock and synchronous reset IOs and contains the test
     *  body.
     *
     *  @param testBody the circuit to elaborate inside the testharness
     */
-  protected final def test[R](
+  protected final def test(
     testName: String
-  )(testBody: Instance[M] => R)(implicit testHarnessGenerator: TestHarnessGenerator[M, R]): Unit = {
+  )(testBody: Instance[M] => TestConfiguration)(implicit testHarnessGenerator: TestHarnessGenerator[M]): Unit = {
```
seldridge, line 298: At first glance, it seems like the test configuration is a parameter of the `test` method itself and not the return of the lambda. I.e., it seems more natural for this to be:

```
protected def test(testName: String, config: TestConfiguration)(body: Instance[R] => Unit)
```

However, the PR is using this to force the communication of how the success and finish are hooked up.

I don't see anything wrong here, just this is unexpected.

seldridge (commented):
This is looking fantastic. I am having a lot of fun playing with this locally.

Address the feedback and tag me for another review.

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
tmckay-sifive: Nice, this is perceptibly faster now.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
tmckay-sifive: I agree it's a bit unexpected, but it is useful to have it as a return type so that it can reference hardware in the test body (e.g. some sentinel value to watch for finishing the test).

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,158 @@ import scala.collection.mutable
...
+    new TestConfiguration(finishCondition = None, successCondition = None, failureMessage = None)
+
+  def runForCycles(nCycles: Int): TestConfiguration = {
+    val counter = Counter(nCycles)
+    counter.inc()
+    new TestConfiguration(
+      finishCondition = Some(counter.value === (nCycles - 1).U),
+      successCondition = None,
+      failureMessage = None
+    )
```
seldridge: I think this is the same:

```suggestion
    val (_, done) = Counter(true.B, nCycles)
    new TestConfiguration(
      finishCondition = Some(done),
      successCondition = None,
      failureMessage = None
    )
```

Slightly more terse:

```suggestion
    new TestConfiguration(
      finishCondition = Some(Counter(true.B, nCycles)._2),
      successCondition = None,
      failureMessage = None
    )
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,158 @@ import scala.collection.mutable
...
+  val result = outcome match {
+    case SimulationOutcome.Success                    => TestResult.Success
+    case SimulationOutcome.Assertion(simulatorOutput) => TestResult.Failure(simulatorOutput)
+    case SimulationOutcome.Timeout(n)                 => TestResult.Failure(s"timeout reached after ${n} timesteps")
+    case SimulationOutcome.SignaledFailure            => TestResult.Failure(s"test signaled failure")
+  }
+  val success = result match {
+    case TestResult.Success => true
+    case _                  => false
+  }
```
seldridge: Nit. I've grown to like using `isInstanceOf` for this pattern:

```suggestion
  val success = result.isInstanceOf[TestResult.Success]
```

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,158 @@ import scala.collection.mutable
...
+
+    /** Run tests matching this name. */
+    def apply(names: String*): Names = new Names(names.toArray)
+  }
+
+  object Name {
+
+    /** Run tests matching this name. */
+    def apply(name: String) = new Names(Array(name))
+  }
```
seldridge: This should be unnecessary given the variadic version.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,158 @@ import scala.collection.mutable
...
+
+  object Globs {
+    def apply(globs: String*): Globs = new Globs(globs.toArray)
+  }
+
+  object Glob {
+
+    /** Run tests matching a glob. */
+    def apply(glob: String) = new Globs(Array(glob))
+  }
```
seldridge: This should be unnecessary given the variadic version.

src/main/scala/chisel3/experimental/inlinetest/InlineTest.scala
```diff
@@ -6,26 +6,158 @@ import scala.collection.mutable
...
+
+  /** Run tests matching any of these names. */
+  case class Names(names: Array[String]) extends Type {
+    override def globs = names
+  }
```
seldridge: There's no conceptual difference between `Names` and `Globs` in terms of mechanics. I would just drop `Name`.

src/main/scala/chisel3/simulator/Simulator.scala
```diff
@@ -118,12 +161,80 @@ trait Simulator[T <: Backend] {
...
+      }
 
+    val failures = results.filter(!_.success)
+    if (failures.nonEmpty) {
+      val failedTests = failures.size
+      val passedTests = results.size - failedTests
+      val moduleName = results.head.dutName
+      val failedTestNames = failures.map(_.testName).mkString(", ")
+      val aggregatedMessage =
+        s"${moduleName} tests: ${passedTests} succeeded, ${failedTests} failed (${failedTestNames})"
```
seldridge: I'd probably use YAML-esque for this to make it easier to see what happened:

```
  passed: ...
  failed: ...
  failures: 
    - message 1
    - message 2 
```

We can play around with this.

src/main/scala/chisel3/simulator/SimulatorAPI.scala
```diff
@@ -102,4 +103,53 @@ trait SimulatorAPI {
...
+    * @param chiselOpts command line options to pass to Chisel
+    * @param firtoolOpts command line options to pass to firtool
+    * @param settings ChiselSim-related settings used for simulation
+    * @param additionalResetCycles a number of _additional_ cycles to assert
+    * reset for
+    * @param subdirectory an optional subdirectory for the test.  This will be a
+    * subdirectory under what is provided by `testingDirectory`.
+    * @param stimulus directed stimulus to use
+    * @param testingDirectory a type class implementation that can be used to
+    * change the behavior of where files will be created
```
seldridge: The parameters are no longer correct.

seldridge: Looking at this, the `additionalResetCycles` is something that may be necessary to make work as you can construct a test which cannot work unless you reset it for longer than one cycle.

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
```diff
@@ -0,0 +1,58 @@
...
+    val clock = module.port(dut.clock)
+    val reset = module.port(dut.reset)
+    val finish = module.port(dut.io.finish)
+    val success = module.port(dut.io.success)
+
+    ResetProcedure.module(_period)(dut)
+
+    var cycleCount = 0
+
+    while (finish.get().asBigInt == 0) {
```
seldridge: I think the `while` loop doesn't do anything now?

src/main/scala/chisel3/simulator/stimulus/InlineTestStimulus.scala
```diff
@@ -0,0 +1,58 @@
...
+  override final def apply(dut: TestHarness[_]): Unit = {
+    val module = AnySimulatedModule.current
+    val controller = module.controller
+
+    val clock = module.port(dut.clock)
+    val reset = module.port(dut.reset)
+    val finish = module.port(dut.io.finish)
+    val success = module.port(dut.io.success)
+
+    ResetProcedure.module(_period)(dut)
```
seldridge: Can you plumb `additionalResetCycles` into this?

seldridge (approved):
LGTM. Please land with the comments addressed.

The one semi-big thing which needs to get sorted out is the reporting. Though, this can happen in follow-ons. There's a pretty big difference between an assertion failing in an inline test vs. an assertion failing in a non-inline test. E.g., if a test fails an assertion inline you see:

```
[365] - should simulate and fail early if assertion raised *** FAILED ***
[365]   chisel3.simulator.Exceptions$TestsFailed: One or more tests failed during simulation
[365] ------------------------------------------------------------------------------
[365] ModuleWithTests tests: 0 succeeded, 1 failed (assertion)
[365] ------------------------------------------------------------------------------
[365]   ...
```

However, this would include much more information if not inline: what assertion failed, the line numbers, etc. The infrastructure for showing this is all there, it just needs to be plumbed through to this output.

Also, we should probably err on the side of providing too much information. E.g., just binning by assertion assert or timeout is likely not enough.

This isn't blocking and I am confident that this will get sorted out as we use it.

svsim/src/main/scala/Workspace.scala
```diff
@@ -65,7 +65,7 @@ final class Workspace(
     else
       s"${System.getProperty("user.dir")}/$path"
 
-  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
+  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes. These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
```
tmckay-sifive:
```suggestion
  /** A directory where the user can store additional artifacts which are relevant to the primary sources (for instance, artifacts related to the generation of primary sources). These artifacts have no impact on the simulation, but it may be useful to group them with the other files generated by svsim for debugging purposes.
```

PR #4947, opened 2025-06-04
[Scala3] Add Bundle plugin

adkian-sifive:
Chisel's Bundle transform for Scala 3. 

### Contributor Checklist

- [X] Did you add appropriate documentation in `docs/src`?

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

#### Release Notes

Adds Bundle plugin

- [ ] Did you review?


core/src/main/scala-3/chisel3/BundleSelect.scala
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+extension (b: Bundle) {
```
jackkoenig, line 15: Would be good to include a comment mentioning that this implements Selectable.

plugin/src/main/resources/plugin.properties
```diff
@@ -1 +1 @@
-pluginClass=chisel3.internal.plugin.ChiselComponent
\ No newline at end of file
+pluginClass=chisel3.internal.plugin.ChiselPlugins
```
jackkoenig: Does this not cause problems for the Scala 2 plugin? Can you fork this into `src/main/resources-2` and `src/main/resources-3`?

Alternatively we could just keep the name `ChiselComponent`, but if you really want to call the Scala 3 version something else, that's fine, but the Scala 2 version has to work too.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,227 @@
...
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.core.Decorators.toTermName
+import dotty.tools.dotc.core.Definitions
+
+import scala.collection.mutable
+
+object BundleHelpers {
+  def cloneTypeFull(tree: Tree)(using Context): Tree = {
+    val cloneSym =
+      requiredMethod("chisel3.reflect.DataMirror.internal.chiselTypeClone")
```
jackkoenig: I get the desire to factor this out into a helper, but am I right in assuming that `requiredMethod` is expensive and we don't want to be invoking every time we call this? I suspect that's why `ChiselNamingPhase` has all of the `required*` things be uninitialized vars set once `prepareForUnit`.

Same all of the other utilities below.

You could still make these utilities, but you might want to make these `required*` things be arguments to the functions.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,227 @@
...
+              _ => tpd.Literal(Constant(true))
+            )
+          )
+        } else None
+
+      // TODO
+      // val autoTypenameOpt =
+      //   if (BundleHelpers.isAutoTypenamed(record.symbol)) {
+      //     BundleHelpers.generateAutoTypename(record, thiz, conArgs.map(_.flatten))
+      //   } else None
```
jackkoenig: Is this still TODO?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,227 @@
...
+            Flags.Method | Flags.Override | Flags.Protected,
+            defn.BooleanType
+          )
+          Some(
+            tpd.DefDef(
+              isPluginSym.asTerm,
+              _ => tpd.Literal(Constant(true))
+            )
+          )
+        } else None
```
jackkoenig:
```suggestion
        Option.when(isBundle) {
          val isPluginSym = newSymbol(
            record.symbol,
            Names.termName("_usingPlugin"),
            Flags.Method | Flags.Override | Flags.Protected,
            defn.BooleanType
          )
          Some(
            tpd.DefDef(
              isPluginSym.asTerm,
              _ => tpd.Literal(Constant(true))
            )
          )
        }
```
Style nit but `Option.when` is great, please use it especially when multiline.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,227 @@
...
+
+      record match {
+        case td @ TypeDef(name, tmpl: tpd.Template) => {
+          val newDefs =
+            elementsImplOpt.toList ++ usingPluginOpt.toList ++ cloneTypeImplOpt.toList
+          val newTemplate =
+            if (tmpl.body.size >= 1)
+              cpy.Template(tmpl)(body = newDefs ++ tmpl.body)
+            else
+              cpy.Template(tmpl)(body = newDefs)
```
jackkoenig: This is a tiny nit, but it's better to prepend to Lists than to append, and you can prepend or append anything iterable (Options included so I'd write newDefs as:
```scala
val newDefs = elementsImplOpt ++: usingPluginOpt ++: cloneTypeImplOpt.toList
```

Similar thing for `tmpl.body`, it's probably better to use `++:`.

That being said, do we really want these new defs at the start of the definition rather than the end?

And do we really need the body size check? Shouldn't it work even if empty?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,227 @@
...
+      val cloned: tpd.Tree =
+        if (ChiselTypeHelpers.isData(vp.tpt.tpe)) cloneTypeFull(select) else select
+
+      if (vp.tpt.tpe.isRepeatedParam)
+        tpd.SeqLiteral(List(cloned), cloned)
+      else cloned
+    }))
+  }
+
+  private def makeArray(values: List[Tree])(using Context): Tree = {
```
jackkoenig: Nit but this is making a Vector, not an Array

jackkoenig (commented):
This looks really good, a few comments, and it seems CI isn't working so need to get that fixed.

plugin/src/main/resources/plugin.properties
adkian-sifive: Scala 2 and Scala 3 plugin property files don't conflict cause Scala 3 expects a different name (plugin.properties) and in a different format from Scala 2's XML. All this is pretty poorly documented

I'm having issues splitting the resources into resources-2 and resources-3 - Scala 3 does not see the plugin.properties file when it's in `src/main/resources-3`. I thought I saw somewhere in Mill docs that this should work but I can't find it - do you have any pointers handy?

adkian-sifive: Specifically I'm getting these warnings:

```
No such entry: plugin.properties
[warn] No plugin in path chisel3/out/plugin/cross/3.3.4/jar.dest/out.jar
```

 and then it fails to compile

jackkoenig: > Scala 2 and Scala 3 plugin property files don't conflict cause Scala 3 expects a different name (plugin.properties) and in a different format from Scala 2's XML. All this is pretty poorly documented

Ah got it, thanks for the explanation.

But yeah we still want to split the resources since there's no reason to ship the metadata for Scala 3 in the Scala 2 plugin and vice versa.

> I'm having issues splitting the resources into resources-2 and resources-3 

This split is not a property of Scala itself, it's something we'll have to do ourselves in Mill. It seems that it does such a split automatically for sources but not for resources. Basically in `plugin/package.mill` we should `override def resources` and have it use the Scala major version in the resource directories.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: After much benchmarking of these "required" methods from dotty and looking into their implementation in the compiler it's safe to conclude that these method calls are cached and won't be helped much by refactoring. For example, here's a snippet from benchmarking of`requiredModule` (the heaviest of the required calls)

```
requiredModule took 0.1 ms (total 0.1 ms)
requiredModule took 0.0 ms (total 0.1 ms)
requiredModule took 0.0 ms (total 0.1 ms)
requiredModule took 0.0 ms (total 0.1 ms)
```

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: The body size guard I copy-pasted from dotty internals so not exactly sure why it's needed but when I remove it I get a cryptic crash from pickler
```
[error] java.lang.AssertionError: assertion failed: unresolved symbols: value j (line 31) #19714, value k (line 36) #19719, value i (line 29) #19712, value bar in class BundleFooBar (line 11) #9152, value j (line 24) #19320, value i (line 35) #19718, value j (line 37) #19720, value foo in class BundleFooBar (line 10) #9151, value k (line 23) #19319, value i (line 22) #19318, value k (line 30) #19713 when pickling chisel3/src/test/scala/chiselTests/Basic.scala 
```

Don't really wanna dig into this atm so I'll leave it as is for now

plugin/src/main/resources/plugin.properties
adkian-sifive: Handled here: #4971

core/src/main/scala-2/chisel3/SelectableCompat.scala
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3
+
+/** Does nothing in Scala 2; this stub trait is for maintaining
+  * compatibility with Scala 3's selectable, required for Scala 3
+  * Records.
```
jackkoenig:
```suggestion
  * compatibility with Scala 3's selectable, required to support
  * structurally-typed Bundles in Scala 3.
```

core/src/main/scala-3/chisel3/BundleSelect.scala
```diff
@@ -0,0 +1,16 @@
...
+
+// Unlike in Scala 2, in Scala 3 the compiler no longer infers the
+// refined structural type for an anonymous subclass - it infers the
+// unrefined type for anonymous Bundles, which will always be
+// Bundle.
+//
+// This extension method allows accessing elements of an
+// anonymous Bundle even if it's refined type is not inferred -
+// without it the compiler will try to grab the element from class
+// Bundle itself and will obviously error.
```
jackkoenig: Should this be a ScalaDoc?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+      )
+      Some(tpd.DefDef(cloneTypeSym.asTerm, _ => newExpr))
+    }
+  }
+
+  def extractConArgs(
+    record:   tpd.TypeDef,
+    thiz:     tpd.This,
+    isBundle: Boolean
+  )(using Context): Option[List[List[tpd.Tree]]] = {
```
jackkoenig: A comment on many of these functions would be useful, but this one especially. What does the return value mean? I'm assuming `List[List[_]]` is because of multiple parameter lists, I'm assuming None means the primary constructor is private? If so please write that.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+    if (primaryConstructorOpt.isEmpty) {
+      report.warning("Unable to determine primary constructor!", record.sourcePos)
+      return None
+    }
+
+    val constructor = primaryConstructorOpt.get
+    val paramLookup = paramAccessors.map(sym => sym.name.toString -> sym).toMap
+    if (constructor.symbol.is(Flags.Private)) {
+      None
+    } else {
```
jackkoenig: Perhaps a good place for `Option.when(!constructor.symbol.is(Flags.Private)) {`?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+      Some(constructor.termParamss.map(_.collect { case vp =>
```
jackkoenig:
```suggestion
      Some(constructor.termParamss.map(_.map { vp =>
```
I don't think you're really doing a `collect` here, just a `map`.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+        val p: Symbol = paramLookup(vp.name.toString)
+        val select =
+          if (
+            record.symbol.asClass.paramAccessors
+              .find(_.name == vp.name)
+              .forall(_.is(Flags.Private))
+          ) None
+          else Some(tpd.Select(thiz, p.name))
+        if (select.isEmpty) return None
```
jackkoenig: I really don't like the `return None` here--in fact, I'm surprised it's allowed, I thought Scala 3 dropped non-local return.

Furthermore, you're constructing a datastructure of all of the classes paramAccessors and then linearly searching it for every single constructor argument which is `O(n^2)`.

Can you hoist this check outside of the nested loop? Then inside the loop just assume none of these are private, it will make the code more clear.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+      val recordTpe = requiredClassRef("chisel3.Record")
+      val cloneTypeSym = newSymbol(
+        record.symbol,
+        Names.termName("_cloneTypeImpl"),
+        Flags.Method | Flags.Override | Flags.Protected,
+        MethodType(Nil, Nil, recordTpe)
+      )
+      Some(tpd.DefDef(cloneTypeSym.asTerm, _ => newExpr))
+    }
+  }
```
jackkoenig: You're basically propagating Optionality from conArgsOpts through to the return value in this function, rather than doing that, do that outside:
```suggestion
  def generateAutoCloneType(
    record:     tpd.TypeDef,
    thiz:       tpd.This,
    conArgs: List[List[tpd.Tree]],
    isBundle:   Boolean
  )(using Context): tpd.DefDef = {
    val newExpr = tpd.New(record.symbol.typeRef, conArgs.flatten)
    val recordTpe = requiredClassRef("chisel3.Record")
    val cloneTypeSym = newSymbol(
      record.symbol,
      Names.termName("_cloneTypeImpl"),
      Flags.Method | Flags.Override | Flags.Protected,
      MethodType(Nil, Nil, recordTpe)
    )
    tpd.DefDef(cloneTypeSym.asTerm, _ => newExpr)
  }
```

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+      m.isPublic
+      && (
+        ChiselTypeHelpers.isData(m.info)
+          || ChiselTypeHelpers.isBoxedData(m.info)
+      )
+    }
+
+    val currentFields: List[Tree] = bundleSym.info.decls.toList.collect {
+      case m if isBundleDataField(m) =>
+        val name = m.name.show.trim
```
jackkoenig: Please add a comment for why you are trimming.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chisel3.internal.plugin
+
+import dotty.tools.dotc.*
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.ast.untpd
+import dotty.tools.dotc.ast.tpd.*
```
jackkoenig: Rather than wildcard importing from this, I think it would be better to call the methods with `tpd`, e.g. `tpd.ref`

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  private def makeVector(values: List[Tree])(using Context): Tree = {
+    val elemTpe = defn.TupleClass.typeRef.appliedTo(List(defn.StringType, defn.AnyType))
+    val vectorModule = ref(requiredModule("scala.collection.immutable.Vector").termRef)
+
+    val vec = Apply(
+      TypeApply(Select(vectorModule, nme.apply), List(TypeTree(elemTpe))),
+      List(SeqLiteral(values, TypeTree(elemTpe)))
+    )
+    vec
```
jackkoenig:
```suggestion
    Apply(
      TypeApply(Select(vectorModule, nme.apply), List(TypeTree(elemTpe))),
      List(SeqLiteral(values, TypeTree(elemTpe)))
    )
```

jackkoenig: Also where does `nme` come from? I think it might be good to have specific imports for some of these things being used rather than wildcard imports.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,226 @@
...
+      && !record.symbol.flags.is(Flags.Abstract)
+    ) {
+      val isBundle: Boolean = ChiselTypeHelpers.isBundle(record.tpe)
+      val thiz:     tpd.This = tpd.This(record.symbol.asClass)
+
+      // =========== Generate _cloneTypeImpl ==================
+      val conArgs: Option[List[List[tpd.Tree]]] =
+        BundleHelpers.extractConArgs(record, thiz, isBundle)
+      val cloneTypeImplOpt =
+        BundleHelpers.generateAutoCloneType(record, thiz, conArgs, isBundle)
```
jackkoenig:
```suggestion
      val cloneTypeImplOpt = conArgs.map(args => 
        BundleHelpers.generateAutoCloneType(record, thiz, args, isBundle)
      )
```
Related to my comment above.

But also, if we can't generate cloneTYpe because the constructor is private, we need to print the same error message that the Scala 2 plugin generates:
```
[error] Private bundle constructors cannot automatically be cloned, try making it package private
[error] class MyBundle private (x: Int) extends Bundle {
[error]                ^^^^^^^^^^^^^^^^^^^^^^^^^
```

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: good catch, thanks

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: I don't actually think we need this here?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: Yeah this was a bit of a hail mary attempt. Cleaned it up now

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,228 @@
...
+    val paramAccessors = record.symbol.primaryConstructor.paramSymss.flatten
+
+    if (primaryConstructorOpt.isEmpty) {
+      report.warning("Unable to determine primary constructor!", record.sourcePos)
+      return None
+    }
+
+    val constructor = primaryConstructorOpt.get
+    val paramLookup = paramAccessors.map(sym => sym.name.toString -> sym).toMap
+    val symAccessors = (x: Names.TermName) => record.symbol.asClass.paramAccessors.find(_.name == x)
```
jackkoenig: Is this not still n^2? Assuming `paramAccessors` is a `Seq` or somethhing like that then each lookup is linear. Should we instead create a Map?

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
```diff
@@ -0,0 +1,228 @@
...
+      report.error(msg, constructor.srcPos)
+      return None
+    }
+
+    constructor.termParamss.map(_.map {
+      case param if symAccessors(param.name).forall(_.is(Flags.Private)) =>
+        report.warning("Private Bundle constructor parameters render the constructor uncloneable")
+        return None
+      case _ =>
+    })
```
jackkoenig:
```suggestion
    constructor.termParamss.foreach(_.foreach {
      case param if symAccessors(param.name).forall(_.is(Flags.Private)) =>
        report.warning("Private Bundle constructor parameters render the constructor uncloneable")
        return None
      case _ =>
    })
```
No value returned by the maps so use a foreach to avoid allocating collections.

Also, report also needs to include a source position, at least the constructor if not the exact argument.

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: it is indeed, I added a map

plugin/src/main/scala-3/chisel3/internal/plugin/BundlePhase.scala
adkian-sifive: looks great now
```
[warn] 17 |class ArbiterIO[T <: Data](private val gen: T, val n: Int) extends Bundle {
[warn]    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]    |      Private Bundle constructor parameters render the constructor uncloneable
```

jackkoenig (approved):
🎉

PR #4949, opened 2025-06-08
Add API to include comments in the FIRRTL

jackkoenig:
I was working on something complicated and this would help a lot with development.

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

This is for debugging generated FIRRTL only--comments are ignored by the FIRRTL compiler.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
+++ b/core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
@@ -297,6 +297,17 @@ private[chisel3] object Serializer {
       serializeIntrinsic(ctx, info, None, None, intrinsic, args, params, typeAliases)
     case i @ DefIntrinsicExpr(info, intrinsic, id, args, params) =>
       serializeIntrinsic(ctx, info, Some(i.name), Some(id), intrinsic, args, params, typeAliases)
+    case FirrtlComment(text) =>
+      // Because of split, iterator will always be non-empty even if text is empty
+      val it = text.split("\n").iterator
+      while ({
+        val line = it.next()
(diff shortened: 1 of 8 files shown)


core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
```diff
@@ -297,6 +297,17 @@ private[chisel3] object Serializer {
...
+      // Because of split, iterator will always be non-empty even if text is empty
+      val it = text.split("\n").iterator
+      while ({
+        val line = it.next()
+        b ++= "; "; b ++= line
+        if (it.hasNext) {
+          newLineAndIndent()
+        }
+        it.hasNext
+      }) ()
```
seldridge, line 310: I know this is the correct approach given that Scala dropped `do--while`. It still looks weird to me, though... 🫠

seldridge (approved):
Nice. This seems useful for debugging generators.

In the future, I'm very open to figuring out how to have comments that are preserved to Verilog. This is a different beast, though.

Also in the future, consider an option that completely disables all comment emission from Chisel. This is an API which can cause problems for build caching depending on how people start to use this.

core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
jackkoenig: I don't love it but it is the most direct way to write this and I do prefer it to duplicating logic for first line vs. the rest 🤷‍♂️

PR #4953, opened 2025-06-13
Add OverrideSerializationClass for custom annotation serialization

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- Feature (or new API)

#### Desired Merge Strategy

- Squash

#### Release Notes

Annotations can now mixin `OverrideSerializationClass` to customize the "class" field in annotation serialization. This allows annotation classes to reside in a different package than firtool is expecting for the `class` field. Note that this breaks automatic deserialization which does not matter to firtool but would matter for annotations communicating information between Chisel generator invocations.

- [ ] Did you review?


firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -149,13 +176,42 @@ object JsonProtocol extends LazyLogging {
...
-      .distinct
+  private def getTagsAndOverrides(annos: Seq[Annotation]): (List[Class[_]], Map[Class[_], String]) = {
+    val tags = mutable.ListBuffer.empty[Class[_]]
+    val seenTags = mutable.Set.empty[Class[_]]
+    def addTag(clazz: Class[_]): Unit = {
+      if (!seenTags(clazz)) {
+        tags += clazz
+        seenTags += clazz
+      }
+    }
```
seldridge, line 187: I didn't realize that `ListBuffer` has constant time prepend _and append_. This looked like a performance issue, but it isn't.

firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -149,13 +176,42 @@ object JsonProtocol extends LazyLogging {
...
+
+    val tagOverride = mutable.Map.empty[Class[_], String]
+
+    for (anno <- annos) {
+      addTag(anno.getClass)
+      anno match {
+        case anno: HasSerializationHints =>
+          anno.typeHints.foreach(addTag(_))
+        case _ => ()
+      }
```
seldridge, line 197: Consider using an `isInstanceOf` over a match here given that this is exactly one check:

```suggestion
      if (anno.isInstanceOf[HasSerializationHints]) {
        anno.typeHints.foreach(addTag(_))
      }
```

Ditto below. This formulation saves on indentation and may be slightly easier to read. Non-blocking and for consideration only.

firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -149,13 +176,42 @@ object JsonProtocol extends LazyLogging {
...
+          tagOverride.get(anno.getClass) match {
+            case Some(existing) =>
+              if (existing != anno.serializationClassOverride) {
+                throw new Exception(
+                  s"Class ${anno.getClass.getName} has multiple serialization class overrides: $existing, ${anno.serializationClassOverride}"
+                )
+              }
+            case None =>
+              tagOverride += anno.getClass -> anno.serializationClassOverride
+          }
```
seldridge: Possibly terser to use `put` to override it and then examine the return `Option` to see if the existing value got booted:

```suggestion
          tagOverride.put(anno.getClass, anno.serializationClassOverride).foreach {
            case existing => 
                throw new Exception(
                  s"Class ${anno.getClass.getName} has multiple serialization class overrides: $existing, ${anno.serializationClassOverride}"            
          }
```

There's also a non-foreach version using an if that may be cleaner:

```suggestion
          val existing = tagOverride.put(anno.getClass, anno.serializationClassOverride)
          if (existing.isDefined) {
            throw new Exception(
              s"Class ${anno.getClass.getName} has multiple serialization class overrides: ${existing.get}, ${anno.serializationClassOverride}"
            )
          }
```

firrtl/src/test/scala/firrtl/JsonProtocolSpec.scala
```diff
@@ -100,4 +103,35 @@ class JsonProtocolSpec extends AnyFlatSpec {
...
+  }
+
+  "JsonProtocol.serializeRecover" should "emit even annotations that cannot be serialized" in {
+    case class MyAnno(x: Int) extends NoTargetAnnotation
+    val target = ModuleTarget("Foo").ref("x")
+    val annos = MyAnno(3) :: DontTouchAnnotation(target) :: Nil
+    val res = JsonProtocol.serializeRecover(annos)
+    res should include(""""class":"firrtl.annotations.UnserializeableAnnotation",""")
+    res should include(""""error":"Classes defined in method bodies are not supported.",""")
+    res should include(""""content":"MyAnno(3)"""")
```
seldridge, line 123: FileCheck would be better. 😉 However, I don't think FileCheck is available in this compilation unit.

firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: Yeah, and `.toList` is constant time (copy-on-write to original ListBuffer object). It's how Scala makes List relatively fast all things considered.

firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: I applied this but had to back it out because you also would then need an `.asInstanceOf`--I think the match is cleaner--alas.

PR #4969, opened 2025-07-09
[chiselsim] Stop using CIRCT filelists

seldridge:
Change ChiselSim to stop using undocumented CIRCT-generated filelists to
know what files to include in simulator compile.  Specifically, don't rely
on the existence of `filelist.f` or `firrtl_black_box_resource_files.f`.

This is needed because the latter file was removed from (the unreleased)
CIRCT 1.125.0 [[1]] for reasons that it wasn't supposed to be load
bearing.  Removing the reliance on the former is just good to do.

Note: this may result in some breakages as users may have been
inadvertently relying on this behavior to exclude some generated files
from their compilation.  E.g., a user could generate some Chisel-time
metadata file which gets put into the build area, but then is excluded
from the compile.  To work around this, users should dump these files in
an alternative area.

[1]: https://github.com/llvm/circt/commit/c7e10d7c35f860c406a342ef54d21c84a132c7a3

### DIFF
--- a/src/main/scala/chisel3/simulator/package.scala
+++ b/src/main/scala/chisel3/simulator/package.scala
@@ -219,31 +219,15 @@ package object simulator {
             movedFiles += abs
         }
       }
-      def moveFiles(filelist: Path) =
-        // Extract all lines (files) from the filelist.
-        Files
-          .lines(filelist)
-          .map(Paths.get(_))
-          .forEach(moveFile)
 
-      // Move a file in a filelist which may not exist.  Do nothing if the
-      // filelist does not exist.
-      def maybeMoveFiles(filelist: Path): Unit = filelist match {
-        case _ if Files.exists(filelist) => moveFiles(filelist)
-        case _                           =>
-      }
-
-      // Move files indicated by 'filelist.f' (which must exist).  Move files
-      // indicated by a black box filelist (which may exist).
-      moveFiles(supportArtifactsPath.resolve("filelist.f"))
-      maybeMoveFiles(supportArtifactsPath.resolve("firrtl_black_box_resource_files.f"))
-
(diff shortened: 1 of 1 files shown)


jackkoenig (approved):

src/main/scala/chisel3/simulator/package.scala
```diff
@@ -219,31 +219,15 @@ package object simulator {
...
-      // ABI-meaningful.
+      // Move all files from the build flow except for some specifically
+      // excluded files:
+      //   - Any filelist like file
+      val skip_re = "^.*\\.f$".r
       Files
         .walk(supportArtifactsPath)
         .filter(_.toFile.isFile)
-        .filter(_.getFileName.toString.startsWith("layers-"))
+        .filter(f => !skip_re.matches(f.getFileName.toString))
```
jackkoenig, line 231: It might be marginally cleaner and faster to do this based on the file extension rather than a regex but it ain't that deep.

seldridge: I actually started with an `endsWith`, but backed off because I thought we would need to ignore more than just `.f`. I remembered that `--dump-fir` could be a problem as that dumps `.fir` files in the build area.

Instead of the simpler approach, I kept the regular expression and beefed it up to also ignore `.fir`. This should justify its use more as I don't know of a clean way to skip by file extension other than iterating through them all (which seems kind of jank).

seldridge:
There may be a bit of whack-a-mole with this file exclusion approach. It would be better if we could break up the builds into distinct Chisel elaboration outputs and distinct FIRRTL compilation outputs. I'm signing up to deal with any fallout of this, though.

I ran this internally and didn't see any problems other than needing to also exclude `.fir` files. 🤞

PR #4972, opened 2025-07-10
Upgrade to Mill 1.0.0

jackkoenig:
Mill 1.0.0 is here 🎉. We badly need this so we can publish Snapshots to Maven Central.

Lots to be aware of for Chisel developers:
* Mill files are now Scala 3.7.1
* `ivy` has been replaced with `mvn` (which makes more sense)
* `Agg` is gone, just use `Seq`
* Reading from files need to be guarded in `BuildCtx.watchValue` due to sandboxing
* `VcsVersion` has not yet been published for Mill 1.0.0 (not even the RCs) so I just re-implemented it
* `docs` is lousy with doing things outside of the `out/` folder which is bad, but rather than tweaking all of that, I just guarded with `BuildCtx.withFilesystemCheckerDisabled`, we should try to remove those.
* We can probably get rid of the `cross` business but I will leave that for follow on work.

### Contributor Checklist

#### Type of Improvement

#### Desired Merge Strategy

- Squash

#### Release Notes

No SNAPSHOT publishing yet, that will be follow on work.

- [ ] Did you review?


build.mill
```diff
@@ -130,6 +137,38 @@ object v extends Module {
...
+          previousTagNoV // really the current tag
+        } else {
+          val currentCommit = {
+            val proc = os.call(("git", "rev-parse", "HEAD"), cwd = rootDir, stderr = os.Pipe)
+            proc.out.trim().take(8) // include first 8 characters of commit hash
+          }
+          s"$previousTagNoV+$commitsSincePreviousTag-$currentCommit-SNAPSHOT"
+        }
+    }
+  }
```
seldridge, line 171: Not actionable in this PR: doing a simpler `git --describe --tag` as opposed to rolling our own here would be good. Granted, this differs slightly from your format as it adds the `g` before the `$currentCommit` and appends `-SNAPSHOT`:

```
# git describe --tag
v7.0.0-RC2-4-gece296e71
```

core/package.mill
```diff
@@ -3,20 +3,20 @@ package build.core
 import mill._
 import mill.scalalib._
 import mill.scalalib.scalafmt._
-import mill.define.Cross
-import de.tobiasroeser.mill.vcs.version.VcsVersion // pulled in by mill-ci-release
+import mill.api.Cross
+//import de.tobiasroeser.mill.vcs.version.VcsVersion // pulled in by mill-ci-release
```
seldridge: Delete instead of comment out.

plugin/package.mill
```diff
@@ -35,21 +34,21 @@ trait Plugin extends CrossSbtModule with ScalafmtModule with release.ChiselPubli
...
   override def scalacOptions = super.scalacOptions() ++ releaseArg(crossScalaVersion)
 
-  def scalaLibraryIvy = v.scalaLibrary(crossScalaVersion)
-  def scalaReflectIvy = v.scalaReflect(crossScalaVersion)
-  def scala2CompilerIvy: Dep = v.scala2Compiler(crossScalaVersion)
-  def scala3CompilerIvy: Dep = v.scala3Compiler(crossScalaVersion)
+  def scalaLibraryMvn = v.scalaLibrary(crossScalaVersion)
+  def scalaReflectMvn = v.scalaReflect(crossScalaVersion)
+  def scala2CompilerMvn: Dep = v.scala2Compiler(crossScalaVersion)
+  def scala3CompilerMvn: Dep = v.scala3Compiler(crossScalaVersion)
```
seldridge, line 40: Thanks for updating the method names even though this was not necessary. 👍

seldridge (approved):
LGTM

I don't understand the actual `mill` changes, however, I reviewed for consistent application of mechanical changes.

core/package.mill
jackkoenig: Whoops missed that, thanks

build.mill
jackkoenig: I prefer our versioning to the default output of `git describe` as it more closely matches what other Scala projects do.

That being said I think a regex to convert the output of `git describe --tag` would perhaps be simpler.

plugin/package.mill
jackkoenig: FWIW it was sort of a happy accident, I updated these with `sed` 😂

PR #4978, opened 2025-07-15
[panama] Remove OMMap, dropped in CIRCT 1.126.0

seldridge:
Remove dropped OM Map features that will appear in CIRCT 1.126.0.  This
was removed in this patch [[1]].

[1]: https://github.com/llvm/circt/commit/2beb8e783f16085026c21a28a585993e18dc99c1


seldridge:
I think this can be made directly against `main` and not have to go to `ci/ci-circt-nightly`. (It could go to either, but it's more straightforward to drop this on `main` if it works.)

jackkoenig (approved):
SGTM, I wouldn't call this an API Modification as this code is not published so isn't part of the public API as far as release notes are concerned.

PR #4982, opened 2025-07-17
[core] Switch to FIRRTL 6.0.0 inline layer ABI

seldridge:
The forthcoming FIRRTL 6.0.0 makes some changes to the ABI for inline layers [[1]].  Specifically, the module name was dropped and the delimiter _after_ "layer" was changed from '_' to '$'.  E.g., what was previously:

```
layer_Foo$A
```

Is now:

```
layer$A
```

Propagate this change into Chisel by updating the way that inline layers self-report their ABI.  This is needed to keep Chisel compatible with CIRCT 1.126.0.

Note: this needs a nightly which has the CIRCT changes here [[2]] and this is currently not available.

[1]: https://github.com/chipsalliance/firrtl-spec/commit/b88817bdde55a10a5bcb432180d26f8ba93da5cf
[2]: https://github.com/llvm/circt/commit/dc0940687a7621ad68de47849dad0f5a4a8e6287

### DIFF
--- a/docs/src/explanations/layers.md
+++ b/docs/src/explanations/layers.md
@@ -208,17 +208,17 @@ only a child layer's file will automatically include its parent layer's file.
 
 Inline layers have their layer blocks guarded with conditional compilation
 directives.  To enable an inline layer, set a preprocessor define when compiling
-your design.  The preprocessor define begins with `layer_` and then includes the
-circuit name and all layer names delimited with dollar signs (`$`).  Parent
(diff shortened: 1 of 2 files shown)


seldridge:
I amended (without editing) the last commit in this PR in order to refresh CI. Now that a nightly is available, this should fix the nightly failure observed here: https://github.com/chipsalliance/chisel/actions/runs/16369009004

PR #4983, opened 2025-07-18
[core] Add knownLayers API to BlackBox, ExtModule

seldridge:
Add an API to add `knownlayer` specifications to FIRRTL `extmodule`s.
This is a mechanism to declare that an external module was compiled with
support for specific layers.  This then allows a FIRRTL comopiler compiler
to know that it should automatically enable these layers if needed.

#### Release Notes

Add `knownLayers` API to `BlackBox` and `ExternalModule`. This allows for external modules to declare that they will contain layers, e.g., that they contain debug collateral. This allows a FIRRTL compiler using them to know that they _should_ enable these layers in concert with the parent circuit.

### DIFF
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -3,6 +3,7 @@
 package chisel3
 
 import chisel3.experimental.{BaseModule, Param}
+import chisel3.layer.Layer
 import chisel3.internal.BaseBlackBox
 import chisel3.internal.Builder.pushCommand
 import chisel3.internal.firrtl.ir._
@@ -17,6 +18,18 @@ package internal {
     // is not possible to properly name EmptyExtModule created from
     // Defintions. See unit test SeparateElaborationSpec #4.a
     private[chisel3] def _isImportedDefinition: Boolean = false
+
+    /** Return the layes which are supported by this `BlackBox`.
+      *
+      * E.g., if this was a [[BlackBox]] that points at Verilog built from
+      * _another_ Chisel elaboration, then this would be the layers that were
+      * defined in that circuit.
+      *
+      * @note This will cause the emitted FIRRTL to include the `knownlayer`
+      * keyword on the `extmodule` declaration.
+      */
+    def knownLayers: Seq[Layer]
+    knownLayers.foreach(layer.addLayer)
   }
 }
 
@@ -70,9 +83,15 @@ package experimental {
     * }}}
     * @note The parameters API is experimental and may change
     */
-  abstract class ExtModule(val params: Map[String, Param] = Map.empty[String, Param]) extends BaseBlackBox {
+  abstract class ExtModule(
+    val params:                     Map[String, Param] = Map.empty[String, Param],
+    override final val knownLayers: Seq[Layer] = Seq.empty[Layer]
+  ) extends BaseBlackBox {
     private[chisel3] override def generateComponent(): Option[Component] = {
       require(!_closed, "Can't generate module more than once")
+
+      evaluateAtModuleBodyEnd()
+
       _closed = true
 
       // Ports are named in the same way as regular Modules
@@ -81,7 +100,7 @@ package experimental {
       val firrtlPorts = getModulePorts.map { case port =>
         Port(port, port.specifiedDirection, UnlocatableSourceInfo)
       }
-      val component = DefBlackBox(this, name, firrtlPorts, SpecifiedDirection.Unspecified, params)
+      val component = DefBlackBox(this, name, firrtlPorts, SpecifiedDirection.Unspecified, params, knownLayers)
       _component = Some(component)
       _component
     }
@@ -127,7 +146,10 @@ package experimental {
   * }}}
   * @note The parameters API is experimental and may change
   */
-abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param]) extends BaseBlackBox {
+abstract class BlackBox(
+  val params:                     Map[String, Param] = Map.empty[String, Param],
+  override final val knownLayers: Seq[Layer] = Seq.empty[Layer]
+) extends BaseBlackBox {
 
   // Find a Record port named "io" for purposes of stripping the prefix
   private[chisel3] lazy val _io: Option[Record] =
@@ -147,6 +169,9 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     require(portsSize == 1, "BlackBox must only have one IO, called `io`")
 
     require(!_closed, "Can't generate module more than once")
+
+    evaluateAtModuleBodyEnd()
+
     _closed = true
 
     val io = _io.get
@@ -166,7 +191,7 @@ abstract class BlackBox(val params: Map[String, Param] = Map.empty[String, Param
     val firrtlPorts = namedPorts.map { namedPort =>
       Port(namedPort._2, namedPort._2.specifiedDirection, UnlocatableSourceInfo)
     }
-    val component = DefBlackBox(this, name, firrtlPorts, io.specifiedDirection, params)
+    val component = DefBlackBox(this, name, firrtlPorts, io.specifiedDirection, params, knownLayers)
     _component = Some(component)
     _component
   }
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -10,6 +10,8 @@ import chisel3.internal.firrtl.ir._
 import chisel3.internal.throwException
 import chisel3.experimental.{SourceInfo, UnlocatableSourceInfo}
 
+import scala.collection.mutable
+
 package internal {
 
   private[chisel3] abstract class BaseBlackBox extends BaseModule {
@@ -19,7 +21,7 @@ package internal {
     // Defintions. See unit test SeparateElaborationSpec #4.a
     private[chisel3] def _isImportedDefinition: Boolean = false
 
-    /** Return the layes which are supported by this `BlackBox`.
+    /** User-provided information about what layers are known to this `BlackBox`.
       *
       * E.g., if this was a [[BlackBox]] that points at Verilog built from
       * _another_ Chisel elaboration, then this would be the layers that were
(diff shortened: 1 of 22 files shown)


core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -17,6 +18,17 @@ package internal {
     // is not possible to properly name EmptyExtModule created from
     // Defintions. See unit test SeparateElaborationSpec #4.a
     private[chisel3] def _isImportedDefinition: Boolean = false
+
+    /** Return the layes that exist inside this `BlackBox`.
```
rwy7:
```suggestion
    /** Return the layers which are supported by this `BlackBox`.
```

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -296,6 +296,14 @@ abstract class RawModule extends BaseModule {
...
 
   private[chisel3] def initializeInParent(): Unit = {}
+
+  /** Record the layers that are known to this [[RawModule]] at the time it was
+    * created.
+    *
+    * This is needed if this [[RawModule]] is ever converted to a [[BlackBox]]
+    * or [[ExtModule]].
+    */
+  final private[chisel3] def knownLayers: Seq[layer.Layer] = Builder.layers.toSeq
```
rwy7: This will be re-evaluated on every use, right?

seldridge: Good catch. This is probably wrong and needs to be something which sets the `knownLayers` at module close based on the state of the `Builder`.

rwy7: I think that's probably OK. I was worried we were stashing the layers BEFORE running the module's body.

core/src/main/scala/chisel3/BlackBox.scala
seldridge: 👍

seldridge: Fixed in: https://github.com/chipsalliance/chisel/pull/4983/commits/85d9ff6497ef8b9da5c249e71972a374076ec60f

core/src/main/scala/chisel3/RawModule.scala
seldridge: This is changed in https://github.com/chipsalliance/chisel/pull/4983/commits/c1e68bec5382aa58594b408ebdf1e02db45e2087 along with a lot of other stuff related to recorded `knownLayers` for more things. This is a higher blast radius than I anticipated.

I also changed the default layers getting added _before_ the main entry point is evaluated so that the behavior of the default layers won't be different for external modules vs. other modules.

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -296,6 +296,19 @@ abstract class RawModule extends BaseModule {
   protected[chisel3] override def moduleBuilt(): Unit = _afterModuleBuilt.foreach(_())
 
   private[chisel3] def initializeInParent(): Unit = {}
+
+  final var _knownLayers: Option[Seq[layer.Layer]] = None
+  atModuleBodyEnd {
```
mwachs5: are there any sneaky APIs that *add* layers with `atModuleBodyEnd`? Can bored probes impact this? Should we at least add some sort of check that no one tries to add a new layer after this _knownLayers was evaluated?

core/src/main/scala/chisel3/BlackBox.scala
```diff
@@ -17,6 +18,18 @@ package internal {
...
+      *
+      * E.g., if this was a [[BlackBox]] that points at Verilog built from
+      * _another_ Chisel elaboration, then this would be the layers that were
+      * defined in that circuit.
+      *
+      * @note This will cause the emitted FIRRTL to include the `knownlayer`
+      * keyword on the `extmodule` declaration.
+      */
+    def knownLayers: Seq[Layer]
+    knownLayers.foreach(layer.addLayer)
```
jackkoenig, line 61: If the user implements knownLayers with a `val` you will get a `NullPointerException`, this should use `atModuleBodyEnd` or otherwise be in the Module "teardown" logic, i.e. in `generateComponent()`.

core/src/main/scala/chisel3/RawModule.scala
seldridge: Boring could definitely cause this to be incorrect. I can probably show that this is a problem if it's not updated there.

The thing that's really annoying about this is that really _only_ `ExtModule`/`BlackBox` should have known layers. However, because there are APIs that let you create an `Instance` from a `Definition` without examining the `DesignAnnotation` (which has layer information) then this forces everything up to `BaseModule` to need to record this.

It perhaps makes more sense this way, just this is surprisingly invasive just to get the separate elaboration examples working correctly.

@jackkoenig: Do you have any ideas on how to architect this more cleanly?

jackkoenig: > However, because there are APIs that let you create an Instance from a Definition without examining the DesignAnnotation (which has layer information) then this forces everything up to BaseModule to need to record this.

I think this is a key observation that applies to a lot of things. We have Modules and then we have auxiliary information stored in the Builder that is largely lost once you exit the build phase for that particular circuit. Definitions act like small circuits and copy things to the Builder of the outer elaboration (if there is one).

I think what we probably need is to actually lean more in to `Definitions` as circuit. Instead of making `knownLayers` a public method on `BaseModule`, can it instead be on `Definition`? And then could `Definitions` capture more information from their `DynamicContexts` (in this case obviously the layers)?

core/src/main/scala/chisel3/RawModule.scala
```diff
@@ -296,6 +296,19 @@ abstract class RawModule extends BaseModule {
   protected[chisel3] override def moduleBuilt(): Unit = _afterModuleBuilt.foreach(_())
 
   private[chisel3] def initializeInParent(): Unit = {}
+
+  final var _knownLayers: Option[Seq[layer.Layer]] = None
```
jackkoenig: Side note and probably moot since ideally we can get rid of this, but if we do need to keep it, it needs to be private.

seldridge: This is now a bit cleaner with a private, mutable `val _layers: LinkedHashSet[Layer]` and a `def layers: Seq[Layer]` which requires that the module is closed.

This helps address the problem that @mwachs5 brought up of needing to worry about boring into a module and then creating a definition of it. Currently, you cannot probe and create anything layer colored, so this is impossible. Though, we will need to add this when boring allows for this.

Any other library doing boring-like things (diplomacy) will need to be updated to call these APIs (and they will likely need to be made publicly accessible).

core/src/main/scala/chisel3/BlackBox.scala
seldridge: I'm surprised I wasn't hitting this as both `ExtModule` and `BlackBox` define `knownLayers` as a constructor parameter `val`.

If I keep this, but make these `final`, would that be fine?

I'm only pushing back on `atModuleBodyEnd` as it's currently defined on `RawModule` and not on `BaseModule`.

jackkoenig: > I'm surprised I wasn't hitting this as both ExtModule and BlackBox define knownLayers as a constructor parameter val.

Ah that is why you wouldn't hit it, constructor vals are the one way to make the val initialize before the supertype constructor is run.

> If I keep this, but make these final, would that be fine?

Yes that is fine!

core/src/main/scala/chisel3/RawModule.scala
seldridge: The problem with `Definition` is that it's currently too generic. I.e., you can have a `Definition[A]` for any `A`. Hence, it's weird to have this live on `Definition[A]`, when it's really a property of `A`. (I generally agree with this, though, and we should move towards having a `Definition` be more tied to a module/extmodule and less generic.)

core/src/main/scala/chisel3/RawModule.scala
seldridge: This is now removed from being specified on `RawModule`. It still is inherited from `BaseModule`, but is now `private [chisel3]`. Layers on a module are updated by `addLayer`.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -953,6 +954,14 @@ package experimental {
     localModulePrefix.foreach { prefix =>
       Builder.pushModulePrefix(prefix, localModulePrefixUseSeparator)
     }
+
+    /** Mutable record of what layers were created in this module or its children */
+    private[chisel3] val _layers: LinkedHashSet[Layer] = LinkedHashSet.empty[Layer]
+
+    private[chisel3] def layers: Seq[Layer] = {
+      require(isClosed, "Can't get layers before module is closed")
+      return chisel3.layers.defaultLayers ++ _layers.toSeq
```
jackkoenig:
```suggestion
     chisel3.layers.defaultLayers ++ _layers.toSeq
```

src/test/scala-2/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
```diff
@@ -503,4 +503,45 @@ class SeparateElaborationSpec extends AnyFunSpec with Matchers with Utils with T
...
+  describe("(5): Known layers") {
+
+    it("(5.a): should be added to imported definitions") {
+
+      val testDir = implicitly[HasTestingDirectory].getDirectory.toString
+
+      val annotations = (new ChiselStage).execute(
+        Array("--target-dir", s"$testDir/Bar", "--target", "chirrtl"),
+        Seq(ChiselGeneratorAnnotation(() => new Bar))
+      )
```
jackkoenig: Is there any reason to use `class ChiselStage` instead of just `ChiselStage.emitCHIRRTLFile`? (it does take `args`).

jackkoenig (approved):
LGTM with a couple of thoughts for consideration

src/test/scala-2/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
seldridge: I thought we didn't have a method that returns an `AnnotationSeq`, but `emitCHIRRTLFile` looks like it does this. 👍

core/src/main/scala/chisel3/Module.scala
seldridge: Thanks! I'm bouncing between too many languages...

seldridge: Fixed in: aaa462043

src/test/scala-2/chiselTests/experimental/hierarchy/SeparateElaborationSpec.scala
seldridge: Fixed in: aaa462043

seldridge:
☝️ I talked to @jackkoenig about this offline and he's good with this approach.

Issue #5011, opened 2025-09-02
Select API's allCurrentInstancesIn affect later currentInstancesIn


fabianschuiki:
I ran into the following issue where a call to `Select.unsafe.allCurrentInstancesIn` affects the result of later `Select.unsafe.currentInstancesIn`. Minimal example:

```scala
// scala-cli select_issue.scala

//> using scala "2.13.12"
//> using dep "org.chipsalliance::chisel:7.0.0-RC4"
//> using plugin "org.chipsalliance:::chisel-plugin:7.0.0-RC4"
//> using options "-unchecked" "-deprecation" "-language:reflectiveCalls" "-feature" "-Xcheckinit" "-Xfatal-warnings" "-Ywarn-dead-code" "-Ywarn-unused" "-Ymacro-annotations"

import chisel3._
import chisel3.experimental.hierarchy._
import chisel3.aop.Select
// _root_ disambiguates from package chisel3.util.circt if user imports chisel3.util._
import _root_.circt.stage.ChiselStage

class Foo extends Module {
  val x = Instantiate(new Bar)
  val y = Instantiate(new Bar)

  println(f"first currentInstancesIn:")
  Select.unsafe.currentInstancesIn(this).foreach { i => println(s"- $i") }

  println(f"allCurrentInstancesIn:")
  Select.unsafe.allCurrentInstancesIn(this).foreach { i => println(s"- $i") }

  println(f"second currentInstancesIn:")
  Select.unsafe.currentInstancesIn(this).foreach { i => println(s"- $i") }
}

@instantiable
class Bar extends Module {
  val u = Instantiate(new Crumpet)
  val v = Instantiate(new Crumpet)
}

@instantiable
class Crumpet extends Module

object Main extends App {
  println(
    ChiselStage.emitSystemVerilog(
      gen = new Foo,
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

This produces the following output:
```
first currentInstancesIn:
- Instance(Clone(ModuleClone(Bar@61f3fbb8)))
- Instance(Clone(ModuleClone(Bar@61f3fbb8)))
allCurrentInstancesIn:
- Instance(Clone(ModuleClone(Bar@61f3fbb8)))
- Instance(Clone(ModuleClone(Bar@61f3fbb8)))
- Instance(Clone(InstanceClone(Crumpet@6124287a)))
- Instance(Clone(InstanceClone(Crumpet@6124287a)))
[...]

PR #5012, opened 2025-09-02
Fix construction of PseudoModules to have correct parent

jackkoenig:
This is mostly used when constructing PseudoModules when looking up Instances within other Instances or Definitions. Previously, we would construct the PseudoModule then manually set the _parent field. The problem with this is that _parent is used during construction of the object and defaults to Builder.currentModule. Instead, we properly set the Builder state to hold the desired _parent and then construct the PseudoModule.

This also cleaned up tech debt in Module.do_pseudo_apply which duplicated logic from the standard Module.apply path in a way that had already gotten out of sync and likely would result in other very subtle bugs.

Fixes https://github.com/chipsalliance/chisel/issues/5011

### Contributor Checklist

#### Type of Improvement

- Bugfix

#### Desired Merge Strategy

- Squash

#### Release Notes

This fixes subtle bugs where accessing children of an `Instance` (whether directly or by using `Select` APIs) could change the result of future `Select` to be incorrect. For example, a call to `Select.unsafe.allCurrentInstancesIn` could cause `Select.unsafe.currentInstancesIn` to also incorrectly include grandchildren and all other transitive children.

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -151,21 +151,24 @@ object Module extends Module$Intf {
   def currentModulePrefix: String = Builder.getModulePrefix
 
   private[chisel3] def do_pseudo_apply[T <: BaseModule](
-    bc: => T
+    bc:     => T,
+    parent: Option[BaseModule]
   )(
     implicit sourceInfo: SourceInfo
   ): T = {
-    val parent = Builder.currentModule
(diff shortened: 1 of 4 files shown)


mikeurbach (approved):
Nice, looks like a good solution to me. I hadn't seen `Builder.State` APIs yet but that looks great, glad to see it used here to simplify `do_pseudo_apply`.

PR #5018, opened 2025-09-09
[core] Add Temporal inline layer, trait

seldridge:
Add a new `Temporal` inline layers that are nested under each of the
builtin Assert, Assume, and Cover layers.  These layers exist because we
have observed situations where assertions which _should_ be legal in all
designs are unsupported by specific tools, though they are legal
properties by the SystemVerilog spec.

E.g., `s_eventually` is entirely unsupported by Verilator and has only
limited support in VCS.  The motivation for this PR is to have a location
where we can put these assertions and guard them off with tools that do
not support them.

Because this is expected to be problematic for all of the existing builtin
layers, this is added to each of Assert, Assume, and Cover.  Additionally,
as users have user-defined layers, the pattern of creating this `Temporal`
layer is factored into a trait, `HasTemporalInlineLayer` so that it can be
reused by others.

#### Release Notes

- Add a `Temporal` layer under each of the builtin `Assert`, `Assume`, and
  `Cover` layers.  This is intended to be used to guard complex or unsupported
  properties due to spotty simulator support.
- Add `HasTemporalInlineLayer` trait to allow users to easily add a recognizable
  `Temporal` child layer to their user-defined layers.

### DIFF
--- a/core/src/main/scala/chisel3/layers/Layers.scala
+++ b/core/src/main/scala/chisel3/layers/Layers.scala
@@ -4,6 +4,30 @@ import chisel3.experimental.UnlocatableSourceInfo
 import chisel3.layer.{CustomOutputDir, Layer, LayerConfig}
 import java.nio.file.Paths
 
+/** Trait that adds a `Temporal` layer inside another layer.
+  *
+  * This temporal layer can used to guard statements which are unsupported,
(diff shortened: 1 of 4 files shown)


seldridge:
The name could be improved. Bikeshed comments are welcome, too.

PR #5024, opened 2025-09-17
[chiselsim] Drop Temporal layers for Verilator Cli

seldridge:
Disable Temporal layers (either internal or those that use the
`HasTemporalInlineLayer` trait) when use ChiselSim with Verilator via the
`Cli.Simulator` trait.

This required adding a number of additional functions to the `CliOption` class
that can be used to control what happens when an option is _not_ set.  This was
a big hole in this API as it created a discontinuity between arguments like
`-Dsimulator=verilator` which would run settings modification functions and the
absence of any simulator being specified which would _not_ run any settings
modifications.  If the settings modifications were non-identity, then this was
incongruous.

Closes #5023.

#### Release Notes

Disable Temporal layers (either internal or those that use the
`HasTemporalInlineLayer` trait) when use ChiselSim with Verilator via the
`Cli.Simulator` trait.

### DIFF
--- a/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
+++ b/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
@@ -66,13 +66,13 @@ object HasCliOptions {
 
     @deprecated("avoid use of copy", "Chisel 7.1.0")
     def copy[A](
-      name:                  String = name,
-      help:                  String = help,
-      convert:               (String) => A = convert,
-      updateChiselOptions:   (A, Array[String]) => Array[String] = updateChiselOptions,
-      updateFirtoolOptions:  (A, Array[String]) => Array[String] = updateFirtoolOptions,
-      updateCommonSettings:  (A, CommonCompilationSettings) => CommonCompilationSettings = updateCommonSettings,
-      updateBackendSettings: (A, Backend.Settings) => Backend.Settings = updateBackendSettings
+      name:                  String,
+      help:                  String,
+      convert:               (String) => A,
+      updateChiselOptions:   (A, Array[String]) => Array[String],
+      updateFirtoolOptions:  (A, Array[String]) => Array[String],
+      updateCommonSettings:  (A, CommonCompilationSettings) => CommonCompilationSettings,
+      updateBackendSettings: (A, Backend.Settings) => Backend.Settings
     ): CliOption[A] = CliOption[A](
       name = name,
       help = help,
--- a/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
+++ b/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
@@ -45,6 +45,7 @@ object HasCliOptions {
 
   object CliOption {
 
+    @deprecated("Use newer CliOption case class apply", "Chisel 7.1.0")
     def apply[A](
       name:                  String,
       help:                  String,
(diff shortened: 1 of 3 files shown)


src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
```diff
@@ -30,17 +30,70 @@ object HasCliOptions {
...
+        updateBackendSettings = updateBackendSettings,
+        updateUnsetChiselOptions = identity,
+        updateUnsetFirtoolOptions = identity,
+        updateUnsetCommonSettings = identity,
+        updateUnsetBackendSettings = identity
+      )
+    }
+
+    @deprecated("Use newer CliOption case class unapply", "Chisel 7.1.0")
+    def unapply[A](cliOption: CliOption[A]): Option[
```
jackkoenig: FWIW I don't think there can be a "newer" unapply, there can only be one. We can just suggest people to not use unapply though.

src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
```diff
@@ -30,17 +30,70 @@ object HasCliOptions {
...
+    convert:                    (String) => A,
+    updateChiselOptions:        (A, Array[String]) => Array[String],
+    updateFirtoolOptions:       (A, Array[String]) => Array[String],
+    updateCommonSettings:       (A, CommonCompilationSettings) => CommonCompilationSettings,
+    updateBackendSettings:      (A, Backend.Settings) => Backend.Settings,
+    updateUnsetChiselOptions:   (Array[String]) => Array[String],
+    updateUnsetFirtoolOptions:  (Array[String]) => Array[String],
+    updateUnsetCommonSettings:  (CommonCompilationSettings) => CommonCompilationSettings,
+    updateUnsetBackendSettings: (Backend.Settings) => Backend.Settings
   )
```
jackkoenig: One source and binary incompatibility is this needs the old constructor:
```scala
def this(
(
    name:                  String,
    help:                  String,
    convert:               (String) => A,
    updateChiselOptions:   (A, Array[String]) => Array[String],
    updateFirtoolOptions:  (A, Array[String]) => Array[String],
    updateCommonSettings:  (A, CommonCompilationSettings) => CommonCompilationSettings,
    updateBackendSettings: (A, Backend.Settings) => Backend.Settings
) = this(name, help, convert, updateChiselOptions, updateFirtoolOptions, updateCommonSettings, updateBackendSettings, identity, identity, identity, identity)
```

src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
```diff
@@ -30,17 +30,70 @@ object HasCliOptions {
     * be converted to.
     */
   case class CliOption[A](
```
jackkoenig, line 32: I forgot you also need to deprecate copy on this, same sort of thing as `unapply`, there can only be one and we just deprecate suggesting not using it.

src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
seldridge: Cleaned up in: 5db7f12ce

src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
seldridge: Thanks for catching this. Fixed in: 5db7f12ce

src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
seldridge: Fixed in: 5db7f12ce This was used in one place related to creating `flag` options and then copying them. I added parameters to the `flag` method (with default values for new parameters) to avoid the need for the copy.

PR #5029, opened 2025-09-19
Fix binary compatibility for HasCliOptions.CliOption

jackkoenig:
You can see the failures in https://github.com/chipsalliance/chisel/pull/5028

### Contributor Checklist

#### Type of Improvement

- Bugfix

Marked No Release Notes because this binary compatibility issue is not yet released.

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
+++ b/src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
@@ -29,7 +29,7 @@ object HasCliOptions {
     * @tparam the internal type of the option.  This is what the `<value>` will
     * be converted to.
     */
-  case class CliOption[A](
+  case class CliOption[A] private (
     name:                       String,
     help:                       String,
     convert:                    (String) => A,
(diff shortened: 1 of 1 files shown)


src/main/scala/chisel3/simulator/scalatest/HasCliOptions.scala
```diff
@@ -216,6 +216,35 @@ object HasCliOptions {
...
+        help = help,
+        updateChiselOptions = identity,
+        updateFirtoolOptions = identity,
+        updateCommonSettings = identity,
+        updateBackendSettings = identity,
+        updateUnsetChiselOptions = identity,
+        updateUnsetFirtoolOptions = identity,
+        updateUnsetCommonSettings = identity,
+        updateUnsetBackendSettings = identity
+      )
```
jackkoenig, line 301: I'm actually surprised this works, since the full `def flag` below has default arguments for everything but `name` and `help` I would assume there's an ambiguity issue. Maybe it only materializes if you try to call this method?

I did test that by adding `addOption(CliOption.flag("foo", "bar"))` to `trait EmitVcd`, but that didn't trip it. So I guess we go with it? Another option is to make this package private so it only exists for binary compatibility reasons but you can't call it.

Anyway like I said, a _little_ surprised this works, might dig a little bit more tomorrow to understand why.

seldridge (commented):
This generally looks good. I had something similar originally, but it wouldn't compile with Scala 3 (I think?). I had to back out the default options.

jackkoenig:
> This generally looks good. I had something similar originally, but it wouldn't compile with Scala 3 (I think?). I had to back out the default options.

I figured it out, whereas Scala 2 won't generate a copy method if you have any copy method, Scala 3 will still generate one unless the copy method has the exact arguments. So the solution is to implement a private version with the correct arguments.

I also made the constructor private to help reduce impact of adding any more arguments.

seldridge (approved):
Nice. Thanks for figuring this out, Jack.

PR #5036, opened 2025-09-29
ChiselEnum convenience functions; support string interaction

maartenboersma:
### Contributor Checklist

- [X] Did you add Scaladoc to every public function/method?
- [X] Did you add at least one test demonstrating the PR?

#### Type of Improvement

- Feature (or new API)
- API modification: yes, I propose to override the `toString` method on `ChiselEnum`. I wonder if anybody out there depends on the current behavior. If so, we could rename to `listValues`

#### Desired Merge Strategy

#### Release Notes

Convenience functions to interact between ChiselEnum and String. Allows RTL readability improvements, especially if Enum values are not yet known at compile time (e.g. decode tables in separate files)

- [ ] Did you review?

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -30,3 +30,4 @@ newlines.sometimesBeforeColonInMethodReturnType = true
 optIn.annotationNewlines = true
 
 rewrite.rules = [SortImports, PreferCurlyFors, AvoidInfix]
+rewrite.avoidInfix.excludeFilters."+" = ["nameContains"]
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -30,4 +30,3 @@ newlines.sometimesBeforeColonInMethodReturnType = true
 optIn.annotationNewlines = true
 
 rewrite.rules = [SortImports, PreferCurlyFors, AvoidInfix]
-rewrite.avoidInfix.excludeFilters."+" = ["nameContains"]
--- a/core/src/main/scala/chisel3/ChiselEnum.scala
+++ b/core/src/main/scala/chisel3/ChiselEnum.scala
@@ -282,6 +282,11 @@ abstract class ChiselEnum extends ChiselEnumIntf {
 
   def apply(): Type = new Type
 
+  /** Return the Enum value of which the name exactly matches the specified String. If @name does not match a valid Enum, fail */
+  def apply(name: String): Type =
+    allWithNames.collectFirst { case (enumValue, enumName) if enumName == name => enumValue }
+      .getOrElse(throwException(s"Enum value $name is not defined"))
+
   private def castImpl(
     n:    UInt,
     warn: Boolean
--- a/core/src/main/scala/chisel3/ChiselEnum.scala
+++ b/core/src/main/scala/chisel3/ChiselEnum.scala
@@ -248,8 +248,12 @@ abstract class ChiselEnum extends ChiselEnumIntf {
   /** All Enum values with their names */
   def allWithNames: Seq[(Type, String)] = all.zip(allNames)
 
-  /** All Enum values with their names, printed one per line. Compatible with gtkwave filter file */
-  override def toString(): String = allWithNames.map(e => s"${e._1.litValue} ${e._2}").mkString("", "\n", "\n")
+  /** Print Enum type name, followed by with all Enum values with their names to a string like this:
+   *  `Opcodes(add=0, sub=1, mul=2, div=3)`
+   */
+  override def toString: String =
+    getClass.getSimpleName.init +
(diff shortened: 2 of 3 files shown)


core/src/main/scala/chisel3/ChiselEnum.scala
```diff
@@ -229,10 +239,18 @@ abstract class ChiselEnum extends ChiselEnumIntf {
...
-  /* Accessor for Seq of names in enumRecords */
+
+  /** All Enum names */
   def allNames: Seq[String] = enumNames
 
+  /** All Enum values with their names */
+  def allWithNames: Seq[(Type, String)] = all.zip(allNames)
+
+  /** All Enum values with their names, printed one per line. Compatible with gtkwave filter file */
+  override def toString(): String = allWithNames.map(e => s"${e._1.litValue} ${e._2}").mkString("", "\n", "\n")
```
jackkoenig: Nit, don't add `()` here (that implies mutation in canonical Scala style).

More important discussion--I took a look at how this prints and I don't think this is the right `.toString`.

For reference, for the following example:
```scala
object EnumExample extends ChiselEnum {
  val e0, e1, e2 = Value

  val e100 = Value(100.U)
  val e101 = Value(101.U)
}
```
This `toString` is:
```
0 e0
1 e1
2 e2
100 e100
101 e101
```
Old (which I agree is bad so no issues replacing) is:
```
chiselTests.EnumExample$@4cd7e993
```
Note that the `.toString` for a specific value (e.g. `EnumExample.e100`) is:
```
EnumExample(100=e100)
```
(I'd prefer if this order were swapped, `e100=100`, I think it's backwards but oh well). 

I think the "right" .toString would be the String concatenation of all of the values:
```
EnumExample(0=e0, 1=e1, 2=e2, 100=e100, 101=e101)
```
That would be more consistent with normal Scala toStrings (which typically do not include newlines).

I think this functionality of printing the the gtkwave compatible style is good but belongs in a method on `DataMirror`. Maybe something like `DataMirror.gtkwaveEnumMapping` or something, what do you think?

core/src/main/scala/chisel3/ChiselEnum.scala
```diff
@@ -118,6 +121,13 @@ abstract class EnumType(private[chisel3] val factory: ChiselEnum) extends Elemen
     implicit sourceInfo: SourceInfo
   ): Bool = isOneOf(u1 +: u2.toSeq)
 
+  /** Creates circuitry that outputs True iff the Enum is equal to one of the values that has `s` in its name
+    *
+    * @param s the substring to search for in the Enum value's name
+    */
+  def contains(s: String)(implicit sourceInfo: SourceInfo): Bool =
+    isOneOf(factory.allWithNames.filter(m => m._2 contains s).map(m => m._1))
```
jackkoenig: I feel like contains is not quite the right name for this, I don't know exactly what `contains` should mean on a enum value but I think this is more of like `nameContains`. Since it is analogous to `String.contains` but for the `name` of the enum rather than its value.

maartenboersma: I agree. Method name change pushed as you suggested.

core/src/main/scala/chisel3/ChiselEnum.scala
maartenboersma: The .toString method now does what you suggested (+ test added). I agree this makes sense.

I moved the previous functionality into a new method called `asTable`. I think we should not explicitly have gtkwave in any of the chisel method names. Gtkwave is just one of the waveform viewers out there.

In our environment, the Enum mappings are also parsed by other tools (simulation environment) - not just gtkwave. We chose the gtkwave filter format because we see no need to invent "yet another format".

I do like the asTable method to be a member of `ChiselEnum`, not hide it in `DataMirror`. I think dumping the ChiselEnum mappings in a machine-readable way is a pretty generic thing.

.scalafmt.conf
```diff
@@ -30,3 +30,4 @@ newlines.sometimesBeforeColonInMethodReturnType = true
 optIn.annotationNewlines = true
 
 rewrite.rules = [SortImports, PreferCurlyFors, AvoidInfix]
+rewrite.avoidInfix.excludeFilters."+" = ["nameContains"]
```
jackkoenig: While I kind of like infix with named functions, the Scala community soured on it a while ago and I think we should follow their lead.
```suggestion
```

maartenboersma: OK. In the test code, I replaced `a nameContains "b"` by `a.nameContains("b")` and reverted the scalafmt exception.

jackkoenig (approved):

PR #5041, opened 2025-10-01
[core] Add initial domain support

seldridge:
Add support for creating domains and domain types.  Add a built-in
ClockDomain type.

This is very basic support initially and many things will not work, namely
connections or compilation of these to anything meaningful.

#### Release Notes

Add initial support for creating domains. These are intended to model things like clock, reset, and power domains and their associations to ports in the design.

### DIFF
--- a/core/src/main/scala/chisel3/Module.scala
+++ b/core/src/main/scala/chisel3/Module.scala
@@ -578,8 +578,8 @@ package experimental {
 
     private val _associations = new HashMap[Data, LinkedHashSet[domain.Type]]()
 
-    protected[chisel3] def getAssociations: Map[Data, Set[domain.Type]] =
-      _associations.view.mapValues(_.toSet).toMap
+    /** Return true if this module has any port associations. */
+    private[chisel3] def hasAsssociations: Boolean = _associations.nonEmpty
 
     // getPorts unfortunately already used for tester compatibility
     protected[chisel3] def getModulePorts: Seq[Data] = {
--- a/core/src/main/scala/chisel3/BlackBox.scala
+++ b/core/src/main/scala/chisel3/BlackBox.scala
@@ -220,7 +220,7 @@ abstract class BlackBox(
     // Note: BlackBoxes, because they have a single `io` cannot currently
     // support associations because associations require having multiple ports.
     // If this restriction is lifted, then this code should be updated.
-    require(getAssociations.isEmpty, "BlackBoxes cannot support associations at this time, use an ExtModule")
+    require(!hasAsssociations, "BlackBoxes cannot support associations at this time, use an ExtModule")
     val firrtlPorts = namedPorts.map { namedPort =>
       Port(namedPort._2, namedPort._2.specifiedDirection, Seq.empty, UnlocatableSourceInfo)
     }


core/src/main/scala/chisel3/domain/Domain.scala
```diff
@@ -0,0 +1,75 @@
...
+  *   )
+  * }
+  * }}}
+  *
+  * @see [[chisel3.domains.ClockDomain]]
+  */
+abstract class Domain()(implicit val sourceInfo: SourceInfo) { self: Singleton =>
+
+  // The name that will be used when generating FIRRTL.
+  private[chisel3] def name: String = simpleClassName(this.getClass())
```
jackkoenig:
```suggestion
  private[chisel3] def name: String = simpleClassName(this.getClass)
```
Nit but we typically call `.getClass` without `()`.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -576,16 +576,23 @@ package experimental {
 
     private val _ports = new ArrayBuffer[(Data, SourceInfo)]()
 
+    private val _associations = new HashMap[Data, LinkedHashSet[Data]]()
+
+    protected[chisel3] def getAssociations: scala.collection.immutable.Map[Data, LinkedHashSet[Data]] =
+      _associations.toMap
```
jackkoenig: Is this really needed or can we just use `_associations`?

core/src/main/scala/chisel3/Module.scala
```diff
@@ -608,6 +615,22 @@ package experimental {
...
     protected def portsSize: Int = _ports.size
 
+    /* Associate a port of this module with one or more domains. */
+    final def associate(port: Data, domain: Data, domains: Data*)(implicit si: SourceInfo): Unit = {
+      val allDomains = Seq(domain) ++ domains
+      if (!portsContains(port)) {
+        Builder.error(s"""Unable to associate port '$port' to domains '${allDomains.mkString(
+            ", "
+          )}' because the port does not exist in this module""")(si)
+        println(getModulePorts)
```
jackkoenig:
```suggestion
```
stray println

core/src/main/scala/chisel3/Module.scala
```diff
@@ -608,6 +615,22 @@ package experimental {
 
     protected def portsSize: Int = _ports.size
 
+    /* Associate a port of this module with one or more domains. */
+    final def associate(port: Data, domain: Data, domains: Data*)(implicit si: SourceInfo): Unit = {
+      val allDomains = Seq(domain) ++ domains
+      if (!portsContains(port)) {
+        Builder.error(s"""Unable to associate port '$port' to domains '${allDomains.mkString(
+            ", "
+          )}' because the port does not exist in this module""")(si)
```
jackkoenig: This is getting formatted weirdly, can you massage it a bit, e.g. assign the result of `allDomains.mkString` to a val.

core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
```diff
@@ -29,10 +29,12 @@ private[chisel3] object Serializer {
...
-    // If the name starts with a digit, then escape it with backticks.
-    case _ if name.head.isDigit => s"`$name`"
-    case _                      => name
+  private def legalize(name: String): String = {
+    name match {
+      // If the name starts with a digit, then escape it with backticks.
+      case _ if name.head.isDigit => s"`$name`"
+      case _                      => name
+    }
   }
```
jackkoenig: This change feels pointless but also is fine

core/src/main/scala/chisel3/Module.scala
```diff
@@ -576,16 +576,23 @@ package experimental {
...
       _ports.iterator.collect { case (d: Data, _) => d }.toSeq
     }
 
     // gets Ports along with there source locators
-    private[chisel3] def getModulePortsAndLocators: Seq[(Data, SourceInfo)] = {
+    private[chisel3] def getModulePortsAndLocators: Seq[(Data, SourceInfo, Seq[Data])] = {
       require(_closed, "Can't get ports before module close")
-      _ports.toSeq
+      _ports.toSeq.map { case (port, info) =>
+        (port, info, _associations.get(port).map(_.toSeq).getOrElse(Seq.empty[Data]))
```
jackkoenig:
```suggestion
        (port, info, _associations.get(port).fold(Seq.empty[Data])(_.toSeq)
```
I personally have taken to `.fold(default)(f)` instead of `.map(f).getOrElse(default)` but I'll defer to you on which you think is better.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -608,6 +615,22 @@ package experimental {
 
     protected def portsSize: Int = _ports.size
 
+    /* Associate a port of this module with one or more domains. */
+    final def associate(port: Data, domain: Data, domains: Data*)(implicit si: SourceInfo): Unit = {
```
jackkoenig: So we've often used this `T, T*` as a trick to ensure one or more, but it makes certain things really awkward, e.g. trying to work with a Seq, you have to do:
```scala
associate(port, myDomains(0), myDomains.tail: _*)
```
I think it would be better to overload with:
```scala
final def associate(port: Data, domain: Data)(implicit si: SourceInfo): Unit = associate(port, Seq(domain))
final def associate(port: Data, domains: Seq[Data])(implicit si: SourceInfo): Unit = ...
```
Or just have a single version with varags and accept it's a runtime error if called with no domains?

Or I guess you could overload with
```
final def associate(port: Data, domain: Data, domains: Data*)(implicit si: SourceInfo): Unit = associate(port, Seq(domain) ++ domains)
final def associate(port: Data, domains: Seq[Data])(implicit si: SourceInfo): Unit = ...
```

core/src/main/scala/chisel3/domain/Type.scala
```diff
@@ -0,0 +1,43 @@
...
+package chisel3.domain
+
+import chisel3.{fromIntToLiteral, Data, Element, Printable, UInt, UnknownWidth, Width}
+import chisel3.experimental.SourceInfo
+import chisel3.internal.{throwException, Builder}
+
+/** A [[Data]] that is used to communicate information of a specific domain
+  * kind.
+  */
+final class Type private (val domain: Domain) extends Element { self =>
```
jackkoenig: Does a `domain.Type` really need to be an Element or Data at all?

seldridge: They do need to be ports, so I think it would either be changing the type that `IO` accepts or adding a new IO creation method just for `domain.Type`. Either of these is doable.

The more critical thing would be that I expect libraries that assume everything in Chisel can be shoved into a `Wire`, like diplomacy, will want to put `domain.Type` into a wire. We might be able to avoid this with changes to libraries so that they don't use a single `Wire`. (This is related to what I was talking to you offline about assuming everything can go into a `Bundle` is making things hard.) WDYT?

jackkoenig: Ah I see, then I think `Element` is right. If these need to be ports and wired through then I think they run into the same needs as other things like `Properties`. As in the larger conversation about `Bundles`, maybe there's a better way out there but there has to be some way to compose these things together and `Bundle` (well `Record`) remains the only existing avenue.

src/test/scala-2/chiselTests/DomainSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+        ("name", Field.String),
+        ("voltage", Field.Integer),
+        ("alwaysOn", Field.Boolean)
+      )
+
+    }
+
+    class Foo extends RawModule {
+      val A = IO(Input(domain.Type(ClockDomain)))
+      val B = IO(Input(domain.Type(PowerDomain)))
```
jackkoenig: This feels a bit inconsistent with the rest of Chisel, what about:
```suggestion
      val A = IO(Input(ClockDomain()))
      val B = IO(Input(PowerDomain()))
```
or perhaps
```suggestion
      val A = IO(Input(ClockDomain.Type()))
      val B = IO(Input(PowerDomain.Type()))
```

core/src/main/scala/chisel3/Module.scala
seldridge: Really needed...

```
[391] [error] core/src/main/scala/chisel3/Module.scala:582:7: type mismatch;
[391] [error]  found   : scala.collection.mutable.HashMap[chisel3.Data,scala.collection.mutable.LinkedHashSet[chisel3.Data]]
[391] [error]  required: scala.collection.immutable.Map[chisel3.Data,scala.collection.mutable.LinkedHashSet[chisel3.Data]]
[391] [error]       _associations
[391] [error]       ^
```

core/src/main/scala/chisel3/domain/Domain.scala
seldridge: Fixed in: 0b78e3209

core/src/main/scala/chisel3/Module.scala
seldridge: Whoops. Thanks! Fixed in: 3cde25f30

core/src/main/scala/chisel3/Module.scala
seldridge: Fixed in: f01a4b8ce

core/src/main/scala/chisel3/internal/firrtl/Serializer.scala
seldridge: Not intentional. I think I was using this for debugging at one point and didn't change it back (do a self-review). Fixed in: d207464b7

core/src/main/scala/chisel3/Module.scala
seldridge: Interesting. Sure, I can try that (here: 44aece60c). This originally didn't make sense to me, but it is mathematically well-formed.

core/src/main/scala/chisel3/Module.scala
seldridge: I may have been overzealous here. Switching to a simple `require` in: a528589c0

src/test/scala-2/chiselTests/DomainSpec.scala
seldridge: This is a nice idea. This did feel somewhat awkward... I switched to your suggestion in: c8eb82305

WDYT about `.Type()` vs `.type()`?

core/src/main/scala/chisel3/Module.scala
jackkoenig: I should have been more clear, is `getAssociations` even needed? It's only used in 1 place, and only to check if it's empty.

core/src/main/scala/chisel3/Module.scala
```diff
@@ -608,6 +615,22 @@ package experimental {
 
     protected def portsSize: Int = _ports.size
 
+    /* Associate a port of this module with one or more domains. */
+    final def associate(port: Data, domains: Data*)(implicit si: SourceInfo): Unit = {
```
jackkoenig: I'm not sure why I didn't notice this the first time, can the type of domains be tighter than `Data`? Can it be `domain.Type`?

core/src/main/scala/chisel3/Module.scala
seldridge: I'd rather not expose a mutable data structure if the API (even if it is an internal API) is that it it must not be mutated.

Is there a better way to return a constant "view" or constant "reference" to an immutable structure?

core/src/main/scala/chisel3/Module.scala
seldridge: Great catch. This also has implications for things like `_associations`. I'll pull on that thread.

seldridge: This is tightened up in: https://github.com/chipsalliance/chisel/pull/5041/commits/432854b7cc1f71f88dca68e26917caaae293ae6d

core/src/main/scala/chisel3/Module.scala
jackkoenig: You could expose different types based on what you need that aren't necessarily the data structure, e.g. `hasAssociations: Boolean` would be enough for current use (but I'm predicting won't be enough forever).

Another strategy is to return `scala.collection.Map` which is the supertype of `immutable.Map` and `mutable.Map`.

Note that even as is, you're returning an immutable map but the values are mutable Sets so you still have a wide open mutable API.

If you need to check the associations you could do something like `hasAssociation(data: Data, domain: Data): Boolean = _assocations.get(data).fold(false)(_.contains(domain))`

Or we could memoize the immutable Map (and also make sure to make the values immutable), bonus points if we can free the mutable data structure

jackkoenig: Oh I missed that you are turning these Sets into Seqs, another option is to box this functionality in a class that hides the details, e.g.

```scala
private[chisel3] class Associations(_associations: mutable.HashMap[Data, mutable.LinkedHashSet[domain.Type]]) {
  def isEmpty: Boolean = _associations.isEmpty

  def get(key: Data): Seq[domain.Type] = _associations.get(key).fold(Seq.empty)(_.toSeq)
}
```
Anyway I'm not going to block on this, it's package private anyway so it's not like it's leaking to users.

jackkoenig (approved):

core/src/main/scala/chisel3/Module.scala
seldridge: Ah, sure. I went the simple route of `getAssociations: Map[Data, Set[domain.Type]]`. The actual order should be useful/relevant for an end user. https://github.com/chipsalliance/chisel/pull/5041/commits/19b68279e50c748f7d1d8a245efdcbb82859d988

seldridge: Switched again to `hasAssociations` in: https://github.com/chipsalliance/chisel/pull/5041/commits/77add105d06d6337f1ce4079b77451e1ad4cc634

PR #5048, opened 2025-10-06
[Scala3] Add apply methods to assert, assume, cover

adkian-sifive:
Add apply methods that don't propagate source line info for Scala 3

### Contributor Checklist

#### Type of Improvement

- Feature (or new API)
- API modification

#### Desired Merge Strategy

#### Release Notes

Added Scala 3 assert, assume, cover apply methods for Scala 3

- [ ] Did you review?

### DIFF
--- a/core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
+++ b/core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
@@ -6,7 +6,8 @@ import chisel3.experimental.{BaseModule, SourceInfo}
 
 trait VerifPrintMacrosDoc
 
-val emptySourceLine: VerifStmtMacrosCompat.SourceLineInfo = ("???", 0)
+private val emptySourceLine: VerifStmtMacrosCompat.SourceLineInfo =
+  ("[Source line unavailable: see https://github.com/chipsalliance/chisel/issues/5049]", 0)
 
 private[chisel3] trait Assert$Intf extends VerifPrintMacrosDoc { self: assert.type =>
 
--- a/core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
+++ b/core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
@@ -2,14 +2,65 @@
 
 package chisel3
 
-private[chisel3] trait Assert$Intf { self: assert.type =>
-  // TODO implement
+import chisel3.experimental.{BaseModule, SourceInfo}
+
+trait VerifPrintMacrosDoc
+
+val emptySourceLine: VerifStmtMacrosCompat.SourceLineInfo = ("???", 0)
+
+private[chisel3] trait Assert$Intf extends VerifPrintMacrosDoc { self: assert.type =>
+
+  def apply(
+    cond:    Bool,
+    message: String,
+    data:    Bits*
+  )(
+    implicit sourceInfo: SourceInfo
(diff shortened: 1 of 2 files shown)


core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
```diff
@@ -2,14 +2,65 @@
 
 package chisel3
 
-private[chisel3] trait Assert$Intf { self: assert.type =>
-  // TODO implement
+import chisel3.experimental.{BaseModule, SourceInfo}
+
+trait VerifPrintMacrosDoc
+
+val emptySourceLine: VerifStmtMacrosCompat.SourceLineInfo = ("???", 0)
```
seldridge: I talked with @adkian-sifive offline about this already.

I'm okay with this as an expedient path. However, please do three things:

1. Open a tracking issue on this repo about adding the missing macro to fix this.
2. Change the "???" to reference the issue number of (1).
3. Open an internal tracking ticket / include this in the internal tracking of the Scala 3 migration.

Ideally, there is _some_ test that will fail with this being incorrect and that would be good as it prevents us releasing a Scala 3 version without fixing this.

seldridge:
```suggestion
private val emptySourceLine: VerifStmtMacrosCompat.SourceLineInfo = ("???", 0)
```

core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
```diff
@@ -2,14 +2,65 @@
...
+  def apply(
+    cond:    Bool,
+    message: Printable
+  )(
+    implicit sourceInfo: SourceInfo
+  ): Assert = VerifStmtMacrosCompat.assert._applyWithSourceLinePrintable(cond, emptySourceLine, Some(message))
+
+  def apply(cond: Bool)(implicit sourceInfo: SourceInfo): Assert =
+    VerifStmtMacrosCompat.assert._applyWithSourceLinePrintable(cond, emptySourceLine, None)
 }
```
seldridge, line 35: Does this need to be two methods, one that is variadic and one that isn't? Can this just be one method with branching in the body based on `data.nonEmpty`?

seldridge: Same question for later methods.

core/src/main/scala-3/chisel3/VerificationStatementMacros.scala
```diff
@@ -4,6 +4,9 @@ package chisel3
 
 import chisel3.experimental.SourceInfo
 import chisel3.internal._
+import chisel3.internal.firrtl.ir._
```
seldridge, line 6: Nit: is this emitting that much from this that it needs a wildcard? Can this use a more targeted import?

core/src/main/scala-3/chisel3/VerificationStatementIntf.scala
adkian-sifive: Yeah there's a bunch of cleanup/consolidation possible here that I was tempted to do (maybe even with Scala 3's Union types?). Decided to keep this as-is for now so that it's laid out the same as in Scala 2 which should make it a bit easier to reason about when all this eventually gets replaced with the macro versions, which I think is a better point to work on cleaning it up

